PR #8, opened 2026-05-20
Handle Kubernetes 1.36 pod certificate requests

howardjohn:
Kubernetes 1.36 stopped putting the subject key in spec.pkixPublicKey for PodCertificateRequest and now sends a stub PKCS#10 request instead. The old controller parsed the deprecated field unconditionally, so new PCRs had pkixPublicKey=null and failed with an ASN.1 'sequence truncated' error before any certificate could be issued.

Add a shared helper that extracts the public key from spec.stubPKCS10Request first and falls back to spec.pkixPublicKey for older clusters. Both pod identity and service DNS signers use that path, so their behavior stays consistent across Kubernetes versions. Bump the Kubernetes dependencies to expose the new field and update controller-runtime to the matching generation.

Without this, the `ate-system` pod all fail to start due to missing certificates.

- [x] Tests pass


ahmedtd (dismissed):

BenTheElder (dismissed):
Thanks John! You can tell we've been testing with 1.35 in kind and GKE so far 😅

.github/workflows/pr-workflow.yaml
```diff
@@ -45,7 +45,11 @@ jobs:
     - name: Install Agent Substrate
       run: hack/install-ate-kind.sh --deploy-ate-system
     - name: Wait For Ready
-      run: kubectl wait pod --all --for=condition=Ready --namespace=ate-system --timeout=300s
```
BenTheElder: I have a more complete fix for this open already in #6, let me get that merged ...

BenTheElder: that's in, can you rebase / drop commit 3? shouldn't flake on this now ..

hack/install-ate-kind.sh
```diff
@@ -22,7 +22,7 @@ ROOT=$(git rev-parse --show-toplevel)
 # override reading dev env
 export NO_DEV_ENV=true
 # we will push images to the local registry
-export KO_DOCKER_REPO=localhost:5001
+export KO_DOCKER_REPO=${KO_DOCKER_REPO:-localhost:5001}
```
BenTheElder, line 25: stray change?

aside: are you using a "real" remote registry? curious what this looks like :-)

howardjohn: I just run a local registry on localhost:5000 not localhost:5001 and its nice to re-use it. I can back this out if not desired

BenTheElder: Ahh, Yeah, localhost:5000 conflicts on macOS (airplay receiver service) so we switched to 5001 everywhere.

Let's keep it, thanks.

Issue #12, opened 2026-05-20
[RFC][Architecture] Control Plane Persistence Layer: Scalability, Durability, and Latency Evaluation


MushuEE:
> **Updated 2026-06-16**: revised candidate set based on permissive-license filter; added Garnet, FoundationDB, TiKV, KeyDB as un-evaluated candidates; removed Aerospike, Cloud Spanner / CockroachDB, Cloud Bigtable, DynamoDB / ScyllaDB. Original benchmark data retained as historical reference.

## 1. Objective & Context

Agent Substrate decouples the high-frequency lifecycle of **Actors** (sandboxed agent workloads) from **Workers** (physical Kubernetes Pods). To meet our North Star target of **sub-100ms workload activation (wakeup) latency** at a scale of **1 billion registered actors** and **1,000+ concurrent wakeups/second**, Substrate bypasses the standard Kubernetes API server for real-time scheduling and state routing, offloading dynamic state to a dedicated high-performance database.

The storage tier's own latency budget — the portion of the 100ms whole-wakeup that the persistence layer alone may consume — is roughly **<10ms whole-path** for the actor state-machine operations. That budget is the architectural constraint that drives every candidate-selection decision below; it rules out any backend that cannot fit consistently inside it at target scale.

This RFC evaluates open-source, permissively-licensed database candidates for Substrate's dynamic instance state, summarizing benchmark results, production operations concerns, and data structure redesign requirements.

---

<details>
  <summary><strong>Considerations for Any Solution</strong> — evaluation template for current and future candidates</summary>

The candidate matrix and deep-dive sections below apply these criteria. Future contributors proposing additional backends should structure their analysis around the same checklist.

### Critical requirements (must-have)

- **Low latency.** P99 write latency must fit inside the storage-tier sub-budget of the 100ms wakeup target — concretely, <10ms whole-path under realistic load on cloud block storage (PD-SSD class). Sub-millisecond P50 for hot-path reads is strongly preferred.
  *Question for any candidate*: what is the measured P99 write latency under burst load, not best-case benchmarks? How does that change with `fsync` policy and replication mode?

- **Permissive open-source license.** BSD, MIT, Apache 2.0, or equivalent (see §2 below for the filter rationale). Stewardship by a foundation or community that prevents Redis/Aerospike-style relicense risk is strongly preferred over solo-vendor governance.
  *Question for any candidate*: who controls the license trajectory? What's the relicense-risk profile in the next 24 months?

[...]

PR #13, opened 2026-05-20
feat(scheduler): optimize worker scheduling to O(1) using Valkey/Redi…

MushuEE:
## Description

Addresses #12 

This PR optimizes Substrate's critical scheduling path by replacing the legacy **$O(N)$ worker range scan bottleneck** with an **$O(1)$ constant-time idle worker Set queue** backed by Valkey/Redis. 

---

## Architectural Rationale & Changes

Previously, when an actor was resumed, the scheduler step (`AssignWorkerStep.Execute` in `workflow_resume.go`) would fetch **all** registered workers in the pool via `store.ListWorkers(ctx)`. For a pool of 10,000 workers, this required serial keyspace scans over all master shards, loading and unmarshaling 10,000 JSON blocks on every single wakeup request. 

This PR completely eliminates that bottleneck by shifting worker state queue management into the database layer:

### 1. Interface Enhancements (`store.Interface`)
* Added `ClaimIdleWorker(ctx, namespace, pool, actorID, actorNamespace, actorTemplate)` to the database contract. 

### 2. Set-Based Idle Pool Management (`ateredis.go`)
We leverage Valkey/Redis's high-performance in-memory `Set` indexing to track available capacity:
* **Atomic Selection ($O(1)$)**: The new `ClaimIdleWorker` executes a single, atomic **`SPOP`** operation on the `pool:<namespace>:<pool>:idle_workers` Set to claim a random free worker in constant time.
* **Zero Scheduling Collisions**: Because `SPOP` is an atomic server-side operation, concurrent scheduler instances are guaranteed to pop unique worker IDs. This completely eliminates optimistic lock collisions and database retries on concurrent wakeups.
* **Automated Lifecycle Hooks**:
  * Newly registered workers are added to the idle set during `CreateWorker`.
  * Deleted worker pods are cleanly purged from the set during `DeleteWorker`.
  * Suspended workers are returned to the set when their `ActorId` transitions to empty in `UpdateWorker`.

### 3. Handling Redis Cluster Slot Restrictions (`CROSSSLOT` fix)
During integration testing, multi-key transaction pipelines (trying to write a worker record and mutate the idle set in a single transaction block) failed with `CROSSSLOT` errors because the keys hashed to different clustered slots. 

We solved this by **splitting the operations sequentially** outside the transactions:
* The worker metadata record remains safely protected by optimistic locking version checks (`WATCH` transactions).
* The indexing mutations (`SAdd`/`SRem` on the idle Set) are executed as separate, independent commands immediately upon transaction success. This ensures 100% compatibility with clustered production Valkey.

---

## Verification & Tests Completed

* **Compilation**: Fully verified with `go vet` (0 errors/warnings).
* **Unit & Integration Tests**: Re-ran the entire store package test suite; **100% of tests passed flawlessly**:
  ```text
  go test -v ./cmd/servers/ateapi/store/...
  PASS
  ok      github.com/agent-substrate/substrate/cmd/servers/ateapi/store/ateredis  0.168s

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_resume.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume.go
@@ -86,7 +86,7 @@ func (s *AssignWorkerStep) IsComplete(ctx context.Context, input *ResumeInput, s
 func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, state *ResumeState) error {
 	var assignedWorker *ateapipb.Worker
 
-	// Re-use previously assigned worker if available.
+	// Re-use previously assigned worker if available (e.g., on retry of a failed resume).
 	if state.Actor.GetAteomPodName() != "" {
 		worker, err := s.store.GetWorker(ctx, state.Actor.GetAteomPodNamespace(), state.ActorTemplate.Spec.WorkerPoolRef.Name, state.Actor.GetAteomPodName())
 		if err == nil && worker.GetActorId() == input.ActorID {
--- a/cmd/ateapi/internal/controlapi/workflow_resume.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"math/rand"
 	"time"
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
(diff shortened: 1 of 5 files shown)


MushuEE:
Fixed format issue using `make fmt`

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -85,38 +84,34 @@ func (s *AssignWorkerStep) IsComplete(ctx context.Context, input *ResumeInput, s
...
-		return fmt.Errorf("while listing workers: %w", err)
-	}
-
 	var assignedWorker *ateapipb.Worker
 
-	// Check if we already have a worker assigned from a previous failed attempt
-	for _, worker := range workers {
-		if worker.GetActorId() == input.ActorID && worker.GetWorkerPool() == state.ActorTemplate.Spec.WorkerPoolRef.Name && worker.GetWorkerNamespace() == state.ActorTemplate.Spec.WorkerPoolRef.Namespace {
+	// Re-use previously assigned worker if available.
+	if state.Actor.GetAteomPodName() != "" {
```
ahmedtd, line 90: This will never be non-empty during resume (except maybe on retry of a failed resume?)

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -246,15 +258,32 @@ func (s *Persistence) UpdateWorker(ctx context.Context, worker *ateapipb.Worker,
 		return fmt.Errorf("while executing update worker transaction: %w", err)
 	}
 
+	// Run SAdd sequentially outside the transaction to avoid cluster slot restrictions.
+	if shouldAddToIdle {
+		setKey := fmt.Sprintf("pool:%s:%s:idle_workers", worker.GetWorkerNamespace(), worker.GetWorkerPool())
```
ahmedtd, line 263: One of our design goals is to let all ActorTemplates share the same WorkerPool to maximize efficiency.  In that case, this will mean that we are sending all restore traffic through a single hash bucket (and thus, one single valkey node).

We are going to need to solve this problem, and I'm having difficulty seeing how it could be solved within redis.

However, this is heaps better than the current strategy of "read every worker all the time".

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -246,15 +258,32 @@ func (s *Persistence) UpdateWorker(ctx context.Context, worker *ateapipb.Worker,
 		return fmt.Errorf("while executing update worker transaction: %w", err)
 	}
 
+	// Run SAdd sequentially outside the transaction to avoid cluster slot restrictions.
+	if shouldAddToIdle {
```
ahmedtd, line 262: I think this needs to be a three phase process.

1) Mark the worker as free, but not yet returned to the free set.
2) Return to the free set
3) Make the worker as fully free.

Otherwise, a crash between steps 1 and 2 will permanently leak the worker from consideration.

With the three-phase approach, we can have an additional background thread sweeping workers that are "free, but not yet returned to the free set" back to the free set.

MushuEE: I think we do have a re-sync process? Will need to double check. But I will make this change. Thanks Taahir.

cmd/ateapi/internal/store/ateredis/ateredis.go
MushuEE: 100% I have a much more advanced change that handles locality but because it is much more ossifying I felt this was a decent intermediate step while the datamodel and DB choices were finalized.

cmd/ateapi/internal/store/ateredis/ateredis.go
MushuEE: To solve this without a complex 3-phase state machine, I’ve implemented a self-healing reconciliation approach: (Felt better than introducing a new `STATUS_FREEING` etc state. IIUC eventual consistency here is likely fine.

Push/Pull Reconciliation:
- Push: Whenever a pod update event fires, the K8s informer actively calls EnsureWorkerIdle O(1) to sync any missing idle workers (on that pod)
- Pull: A background loop sweeps worker records (typically <5,000 pods, so extremely cheap) every 1 min to heal any leaked idle workers.

cmd/ateapi/internal/controlapi/workflow_resume.go
MushuEE: clarified the inline comment to specify that this branch is active and crucial for handling retries of a failed resume attempt (e.g., if a previous attempt failed during the restore/boot phase).

cmd/ateapi/internal/store/ateredis/ateredis.go
MushuEE: lmk if that flow holds

MushuEE:
Because of #135 I think we should hold off on this because I don't think we need to introduce a new function to store.go and likely in the next week that interface will settle.

PR #14, opened 2026-05-20
Remove merge markers from LICENSE

thockin:
This is my fault.  I hang my head in shame.


BenTheElder (approved):

thockin:
self-merge while we figure out roles and members

PR #17, opened 2026-05-20
cleanup: use a cache for JWT/CA instead of reading from disk every time

yuval-k:
Small PR to fix re-reading the JWT keys and CAs every-time.

For sessionIDJWTPoolFile, sessionIDCAPoolFile check modification time, and if not changed, use cached value.

- [X] Tests pass


ahmedtd:
This looks good, but actually I think we probably don't even need to check mtimes --- it's probably sufficient to just check if the keys have changed every 5 minutes or so.  Root key rotations are not part of the normal operation of the system.

yuval-k:
hi @ahmedtd can you allow CI workflows please?

thockin:
CI approved

Issue #18, opened 2026-05-20
Design: How should substrate use (or not use) namespaces in a Kubernetes cluster?


thockin:
## Current state:

Not a ton of thought has gone into the current design.

### ActorTemplates

CRDs are cluster-global, but the CRs they describe can be namespaced.  `ActorTemplate` is a namespaced CR.  The substrate code today pays attention to all `ActorTemplate`s, regardless of namespace.  Most of the UX involves specifying templates as a tuple of {namespace, name}.  For example: `kubectl ate create actor foo --template bar/my-template` creates an actor named "foo" from the `ActorTemplate` named "my-template" in k8s namespace "bar".

### WorkerPools

Like `ActorTemplate`, `WorkerPool` is a namespaced CRD which is ultimately manifested as k8s pods in that same namespace.

### Control plane

To run Agent Substrate, we need a namespace in which to run a bunch of stuff, such as the control plane.

## How should substrate use k8s namespaces?

Kubernetes namespaces are generally used for separating teams or “tenants”, environments (staging vs. prod), service producers, for managing quota and charge-backs, and many other reasons.

So now we need to answer a fundamental question - is a substrate a cluster level concept  (i.e. any cluster can have at most 1 substrate) or can a single cluster run multiple instances of substrate?

It **seems** obvious (to me) that a single cluster should be able to run multiple instances of the substrate.  So how would one deploy two instances of it in a cluster? Obviously, those would go into different namespaces, but exactly how to map those is not obvious.

"Workers" are analogous to Kubernetes Nodes.  Kubernetes Nodes are not namespaced, which means that Pods can share a Node even if the Pods are from different Namespaces.  From that point of view it seems natural that Workers shouldn't be namespace, either. Especially if we want cluster-wide worker-pools that are fungible across actors.  But do we?  

If a cluster can be "multi-tenant" (with varying levels of hard/soft isolation possible), then a cluster-wide ANYTHING will be trouble.  In fact, it's probably not controversial to say that many of the “cluster scoped” (non-namespaced) concepts in Kubernetes are considered problematic in hindsight.

[...]

PR #20, opened 2026-05-20
Support actortemplate secret env

### LINKED ISSUE #15: Support Secret env sources in ActorTemplate containers
## Summary

ActorTemplate container env currently accepts Kubernetes `corev1.EnvVar` in the CRD schema, so users can write `valueFrom.configMapKeyRef` and `valueFrom.secretKeyRef` in manifests. At runtime, Substrate flattens env vars into its atelet workload protobuf, which historically carried only literal `name`/`value` pairs. That means `valueFrom` entries are accepted by the API but are not usable by actors.

Add support for resolving `valueFrom.configMapKeyRef` and `valueFrom.secretKeyRef` for ActorTemplate container env before handing the workload spec to atelet.

## Why this matters

- Secret-backed env vars are the standard Kubernetes pattern for API keys and credentials. Without this, demos and real workloads either inline secrets as plain env values in rendered manifests or need custom workaround scripts.
- ConfigMap-backed env vars are the standard Kubernetes pattern for non-secret runtime configuration.
- The current behavior is surprising because the CRD schema exposes Kubernetes EnvVar, including `valueFrom`, but Substrate silently loses the source and sends an empty literal value to atelet.
- Supporting these two common `valueFrom` sources makes ActorTemplate behavior closer to what Kubernetes users expect while keeping unsupported sources explicit.

## Proposed scope

- Support literal `env.value` as before.
- Support `env.valueFrom.configMapKeyRef` from the ActorTemplate namespace.
- Support `env.valueFrom.secretKeyRef` from the ActorTemplate namespace.
- Preserve Kubernetes `optional: true` behavior by skipping the env var if the referenced object or key is missing.
- Return a clear error for missing required objects/keys.
- Return a clear error for unsupported sources such as `fieldRef`, `resourceFieldRef`, and `fileKeyRef`.

## Security / design note

Resolving Secret and ConfigMap references in ate-api means ate-api needs read access to those resources. The straightforward implementation grants the ate-api service account `get` access to ConfigMaps and Secrets so ActorTemplates work without custom per-namespace RBAC. This is a conscious tradeoff for now, but it should remain a security TODO/design choice.

Possible future designs include per-namespace delegated RBAC, admission-time validation/materialization, or another scoped credential model that avoids broad Secret read access in the control plane.

## Acceptance criteria

- ActorTemplate env with `valueFrom.secretKeyRef` reaches the actor process with the referenced Secret value.
- ActorTemplate env with `valueFrom.configMapKeyRef` reaches the actor process with the referenced ConfigMap value.
- Missing required refs fail clearly instead of becoming empty env vars.
- Optional missing refs are skipped.
- Unsupported `valueFrom` sources fail clearly.
- Docs call out supported sources and the security/design tradeoff.

EItanya:
A couple of TODOs/Discussions which I think are important to have based on these changes.
1. Do we need a caching client for this to avoid calling the apiserver every time?
2. Where/how do we want to resolve secrets generally? I think there are good arguments to be made for the control-plane OR the "node" component. In my mind this also related to: https://github.com/agent-substrate/substrate/issues/18

Fixes #15 

- [ x ] Tests pass


manifests/ate-install/ate-api-server.yaml
```diff
@@ -22,6 +22,11 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["get", "watch", "list"]
+# TODO: Decide whether env source resolution should keep using ate-api's
+# broad read access or move to a more scoped delegation/admission model.
+- apiGroups: [""]
```
ahmedtd: We should not give a wide grant like this to ate-api-server by default.  Instead, each demo should just grant ate-api-server access to the configmaps and secrets it needs.

cmd/ateapi/internal/controlapi/workload_spec.go
```diff
@@ -0,0 +1,229 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func workloadSpecFromActorTemplate(ctx context.Context, kubeClient kubernetes.Interface, actorTemplate *atev1alpha1.ActorTemplate) (*ateletpb.WorkloadSpec, error) {
+	workloadSpec := &ateletpb.WorkloadSpec{
+		PauseImage: actorTemplate.Spec.PauseImage,
+	}
+	resolver := envResolver{
```
ahmedtd, line 39: We are pessimizing Actor resumption by putting one or more k8s reads in the hot path.  We need to move these off the hot path, maybe by using a shared cache of secrets and configmaps with a 1-minute expiry.

That way, only the first resumption is slowed down.  And in the future, we could proactively fill and refresh the cache by watching ActorTemplates.

ahmedtd (commented):
Can you also refactor the API type so we no longer use the K8s environment variable type, but instead one defined alongside ActorTemplate?  It can still have the same shape as the k8s type.

cmd/ateapi/internal/controlapi/workload_spec.go
ahmedtd: In the future, once some sort of Kubernetes RBAC support for conditional authorization lands, we could migrate to letting ate-api-server read (and inform on) any configmap or secret with a specific label.  That way, we could always just read the secret/configmap from the informer cache.

cmd/ateapi/internal/controlapi/workload_spec.go
```diff
@@ -0,0 +1,229 @@
...
+			Image:   ctr.Image,
+			Command: ctr.Command,
+		}
+		for _, env := range ctr.Env {
+			ateletEnv, err := resolver.resolve(ctx, ctr.Name, env)
+			if err != nil {
+				return nil, err
+			}
+			if ateletEnv != nil {
+				ateletCtr.Env = append(ateletCtr.Env, ateletEnv)
```
ahmedtd, line 57: Our RPC interceptor in atelet and ateom log the request proto --- we need to ensure that secrets are not written to logs.

manifests/ate-install/ate-api-server.yaml
mtaufen: +1.

Some options we can consider:
- initially, just grant individual secrets one by one to ate-api-server as Taahir suggested, this is probably the fastest option that has reasonable scope...
- Can we somehow authorize based on label selectors so secrets need to be labeled before ate-api-server can read them? They can be checked in authz: https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/4601-authorize-with-selectors/README.md, but since these are configured via a kube-apiserver config file rather than the API, it complicates the substrate installation process if we do it. 
- Should we do something similar to ReferenceGrant in Gateway API? (too bad we never generalized this in K8s)

cmd/ateapi/internal/controlapi/workload_spec.go
mtaufen: We should do log sanitization as a follow-up too... Filed https://github.com/agent-substrate/substrate/issues/103

cmd/ateapi/internal/controlapi/workload_spec.go
mtaufen: Note we can already check label selectors in authz today: https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/4601-authorize-with-selectors/README.md. But I think it requires a custom authorizer.

manifests/ate-install/ate-api-server.yaml
mtaufen: Another temporary option until we have something better might be to designate a special secrets namespace (or multiple) for ate-api-server to read secrets from, and only grant it access to read secrets from that namespace.

cmd/ateapi/internal/controlapi/workload_spec.go
```diff
@@ -0,0 +1,229 @@
...
+		if apierrors.IsNotFound(err) {
+			if isOptional(ref.Optional) {
+				return "", false, nil
+			}
+			return "", false, status.Errorf(codes.FailedPrecondition, "%s references missing ConfigMap %s/%s", envID, r.namespace, ref.Name)
+		}
+		return "", false, fmt.Errorf("while resolving %s configMapKeyRef %s/%s: %w", envID, r.namespace, ref.Name, err)
+	}
+
+	value, ok := configMap.Data[ref.Key]
```
mtaufen: ConfigMaps also have a BinaryData map, perhaps we should handle that too? Does downward API support BinaryData in env vars in K8s? (maybe not?)

cmd/ateapi/internal/controlapi/workload_spec.go
```diff
@@ -0,0 +1,229 @@
...
+
+	configMap, err := r.configMap(ctx, ref.Name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			if isOptional(ref.Optional) {
+				return "", false, nil
+			}
+			return "", false, status.Errorf(codes.FailedPrecondition, "%s references missing ConfigMap %s/%s", envID, r.namespace, ref.Name)
+		}
+		return "", false, fmt.Errorf("while resolving %s configMapKeyRef %s/%s: %w", envID, r.namespace, ref.Name, err)
```
mtaufen: I noticed some errors are using status.Errorf with a grpc error code, and others are just fmt.Errorf. Should we always be using status.Errorf? Or maybe we should just use normal Go errors and transform them for grpc in a central place?

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -157,7 +158,8 @@ func (s *AssignWorkerStep) findFreeWorker(workers []*ateapipb.Worker, workerPool
 }
 
 type CallAteletRestoreStep struct {
-	dialer *AteletDialer
+	dialer     *AteletDialer
+	kubeClient kubernetes.Interface
 }
 
 func (s *CallAteletRestoreStep) Name() string { return "CallAteletRestore" }
```
mtaufen, line 167: On resume from a snapshot, as @fvoznika pointed out in https://github.com/agent-substrate/substrate/issues/15#issuecomment-4521119107, old env vars are still in process memory and new env vars get put in a file the process needs to optionally refresh from. Can we make this more obvious in the API somehow? It's difficult to debug unless you know that and will probably confuse users.

manifests/ate-install/ate-api-server.yaml
EItanya: I will drop the wide grant from `ate-install/ate-api-server.yaml` and move ConfigMap/Secret RBAC into the per-demo manifests so each demo only grants what it actually references. 

I think we need to solve this from the install/tenancy perspective, but we can punt that decision a little bit past this particular PR I think

cmd/ateapi/internal/controlapi/workload_spec.go
EItanya: Agreed on getting these off the hot path. Before rolling our own TTL cache, could we reuse [`sigs.k8s.io/controller-runtime/pkg/cache`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache)? It already gives us informer-backed reads with the option to scope by namespace or label selector, which lines up well with your follow-up about watching ActorTemplates / informing on labeled secrets once conditional authz lands. WDYT — reuse controller-runtime cache, or build a smaller bespoke TTL cache to keep the dependency surface small?

A downside of the controller-runtime cache could be extra dependencies, as well as a watch dependency which can be more difficult from an RBAC tracking perspective.

cmd/ateapi/internal/controlapi/workload_spec.go
EItanya: Makes sense, do you want that to happen in this PR or in a follow-up?

cmd/ateapi/internal/controlapi/workload_spec.go
EItanya: From what I can tell it's not

cmd/ateapi/internal/controlapi/workload_spec.go
EItanya: Good point — I want to clean this up in this PR. Two directions I see:

1. Use plain `fmt.Errorf` everywhere internally and map to gRPC codes in a single place at the RPC boundary (e.g., a `controlapi.toRPCError` helper or a server interceptor that recognizes a small set of sentinel errors / typed errors).
2. Keep `status.Errorf` at the origin so we don't lose code fidelity through wrapping, and just be consistent about using it everywhere it's appropriate.

I lean toward (1) since most of these errors aren't inherently gRPC-shaped — the workflow steps could be reused outside an RPC handler — but happy to go with (2) if you prefer keeping codes close to the source. Preference?

cmd/ateapi/internal/controlapi/workflow_resume.go
EItanya: Agreed, this is a real footgun. A few API-surface ideas, ordered roughly by invasiveness:

1. **Status visibility.** Surface the resolved env on `Actor.status` (or a hash/generation of it) at restore time, plus a `LastEnvResolvedAt` timestamp. Users get a clear "this is the env the snapshot was resumed with" signal even if the in-process values are stale.
2. **Snapshot-vs-resume divergence condition.** When we detect that the resolved env differs from what was captured at snapshot time, set a condition like `EnvDriftedSinceSnapshot=True` with the affected keys. That makes the "old values still in memory" gotcha discoverable from `kubectl describe` instead of from process introspection.
3. **Explicit refresh contract on the API.** Document on `ActorTemplate` (and probably the Restore RPC) that env values are materialized into the new env file at restore but the process must opt in to re-reading. Optionally expose a knob on the template like `envRefreshPolicy: { OnRestore | Never }` so the behavior is declared rather than implicit.
4. **Signal to the process.** Longer-term, atelet could deliver a signal (SIGUSR1 / a fd notification) on restore so SDKs can wire up automatic refresh without polling the env file.

(1) + (3) feel like the minimum cost to stop confusing users. (2) is the most user-visible. (4) probably belongs to a separate SDK-side issue. @mtaufen @fvoznika which combination do you want?

cmd/ateapi/internal/controlapi/workload_spec.go
BenTheElder: x-ref: https://github.com/kubernetes/kubernetes/issues/139132

cmd/ateapi/internal/controlapi/workload_spec.go
ahmedtd: I think we need to make sure we are not logging the secrets in this PR.  It could be a ham-fisted approach such as just completely blanking out the environment variables field when logging the request proto.

cmd/ateapi/internal/controlapi/workflow_resume.go
mtaufen: Agree 1 + 3 (status and document it) seem like the best start. (2) seems redundant w/ (1)? 

WDYT about an env hash in addition to timestamp? Or maybe a list of env var keys that have been updated?

cmd/ateapi/internal/controlapi/workload_spec.go
mtaufen: (1) was also my natural inclination. @ahmedtd @juli4n do you have an opinion either way on error handling approach?

cmd/ateapi/internal/controlapi/workload_spec.go
mtaufen: Ack. Can we add a comment pointing to that K8s issue?

cmd/ateapi/internal/controlapi/workflow_resume.go
EItanya: Implemented in aeafd2c, incorporating Tim’s guidance from issue #15. `Actor` now records `last_resolved_env_sha256` and `last_resolved_env_at` on successful resume. If an actor has a snapshot and the previously recorded env hash differs from the newly resolved env hash, resume bypasses snapshot restore, boots from the current spec, and clears the stale `last_snapshot` reference. Added a functional test that resumes with ConfigMap env V1, suspends, updates the ConfigMap to V2, then verifies the second resume uses `Run` instead of `Restore`.

cmd/ateapi/internal/controlapi/workload_spec.go
EItanya: Done in aeafd2c. Added an inline comment beside the ConfigMap `Data` lookup pointing at https://github.com/kubernetes/kubernetes/issues/139132 so we can revisit `BinaryData` handling if Kubernetes changes `ConfigMapKeyRef` semantics.

cmd/ateapi/internal/controlapi/workload_spec.go
EItanya: Fixed in aeafd2c. The shared unary interceptor now logs sanitized proto clones for both requests and responses and recursively clears any `env` fields before logging, so atelet `Run`/`Restore` request env values are not emitted. Added `TestServerUnaryInterceptorRedactsEnvFromProtoRequestLogs` to cover this.

cmd/ateapi/internal/controlapi/workload_spec.go
EItanya: Done in aeafd2c. I kept expected user/template validation failures as `FailedPrecondition`, and changed unexpected Kubernetes read failures during env resolution to `status.Errorf(codes.Internal, ...)` so this path consistently returns gRPC status errors instead of leaking raw `fmt.Errorf` errors.

Falven:
I disagree with silently discarding snapshots just because a ConfigMap or Secret changed. At a minimum, it should be a setting where the operator decides the default behavior. This way operators can have a say: “for this template, config freshness is more important than memory continuity.”

EItanya:
> I disagree with silently discarding snapshots just because a ConfigMap or Secret changed. At a minimum, it should be a setting where the operator decides the default behavior.

That was the decided outcome from the discussion in #15

EItanya:
I updated against latest `main`, ready for another review

cmd/ateapi/internal/controlapi/workload_spec.go
mtaufen: Is the controller-runtime cache compatible with direct GETs instead of a watch, since we're currently telling users to grant per-secret permissions?

thockin:
Procedural point - can you squash and get rid of merge commits?

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -30,12 +30,6 @@ const (
 	PhaseFailed            PhaseType = "Failed"
 )
 
-type EnvVar struct {
```
thockin, line 33: I'm curious why it's valuable to move this to a different file?

docs/api-guide.md
```diff
@@ -42,6 +42,8 @@ The `ActorTemplate` defines the code, environment, and state-management policies
 | `pauseImage` | `string` | **Required.** The image used for the sandbox root (e.g. `gcr.io/gke-release/pause`). |
 | `runsc` | `RunscConfig` | **Required.** Multi-platform configuration for fetching the gVisor binary. |
 
+Container environment variables support literal `value` entries and `valueFrom.secretKeyRef`. Secret references are resolved by `ate-api-server` from the `ActorTemplate` namespace when an actor cold-starts or first resumes; the resolved values are sent to atelet but are not serialized into the public Actor API. Other Kubernetes `valueFrom` sources are not supported yet. Secret changes do not automatically restart actors or invalidate snapshots; rotating a Secret requires an explicit actor or template lifecycle action.
```
thockin: Is this actually right?  Assuming all "starts" are from-golden-snapshot, the secret is captured at the time the golden is created.  It will only ever be updated if the golden is recreated.

Am I missing it?

cmd/ateapi/internal/controlapi/workload_spec.go
```diff
@@ -0,0 +1,166 @@
...
+			Name:  env.Name,
+			Value: env.Value,
+		}, nil
+	}
+
+	if env.Value != "" {
+		return nil, status.Errorf(codes.FailedPrecondition, "%s sets both value and valueFrom", envID)
+	}
+
+	sources := valueFromSourceCount(env.ValueFrom)
```
thockin: We need to be doing this at API time rather than at runtime e.g. `// +kubebuilder:validation:MinProperties=1` and `// +kubebuilder:validation:MaxProperties=1`

thockin (commented):
This LGTM mostly, except the hot-path read.

I see we are not really doing much API validation of CRDs - we will have to do that, which means we should not need to do it in the runtime code.

The point about no reading k8s API on the hot-path is a major open.

EItanya:
> The point about no reading k8s API on the hot-path is a major open.

100%

In my opinion this PR has already been open for a while and finally seems to be getting to an agreeable point. Since the scale of this will not be important immediately after merging, can I do the caching in a fast-follow?

pkg/api/v1alpha1/actortemplate_types.go
EItanya: Bad merge, when I originally created this PR there was no type here

cmd/ateapi/internal/controlapi/workload_spec.go
EItanya: No it uses `watch` iirc. Also see the following comment: https://github.com/agent-substrate/substrate/pull/20#issuecomment-4632226505

internal/ateinterceptors/ateinterceptors.go
```diff
@@ -54,3 +56,39 @@ func ServerUnaryInterceptor(ctx context.Context, req any, info *grpc.UnaryServer
...
+	}
+
+	clone := proto.Clone(msg)
+	clearEnvFields(clone.ProtoReflect())
+	return clone
+}
+
+func clearEnvFields(msg protoreflect.Message) {
+	msg.Range(func(fd protoreflect.FieldDescriptor, value protoreflect.Value) bool {
+		if fd.Name() == "env" {
```
juli4n, line 73: This will be hard to maintain / evolve, and protos have a way to declare this with the debug_redact = true annotation. How about annotating the field with that in atelet.proto and then just change the code here if the proto option is set?

EItanya: Please see the following issue https://github.com/agent-substrate/substrate/issues/103

I agree with you 100%, but as I mentioned above, this PR has been open a long time and the scope keeps expanding and contracting. Can we merge as is with a fast-follow from me?

cmd/ateapi/internal/controlapi/workload_spec.go
```diff
@@ -0,0 +1,150 @@
...
+func (r *envResolver) resolve(ctx context.Context, containerName string, env atev1alpha1.EnvVar) (*ateletpb.EnvEntry, error) {
+	envID := fmt.Sprintf("container %q env %q", containerName, env.Name)
+	if env.ValueFrom == nil {
+		return &ateletpb.EnvEntry{
+			Name:  env.Name,
+			Value: env.Value,
+		}, nil
+	}
+
+	if env.Value != "" {
```
juli4n: Nit: If we want to be defensive (this is also enforced by CRD validations), how about we check the invariant at the very beginning of the method and fail fast? 

```
if env.Value != "" && env.ValueFrom != nil {
		return nil, status.Errorf(codes.FailedPrecondition, "%s sets both value and valueFrom", envID)
	}
```

cmd/ateapi/internal/controlapi/workload_spec_test.go
```diff
@@ -0,0 +1,194 @@
...
+					Name:    "main",
+					Image:   "main",
+					Command: []string{"/main"},
+					Env: []atev1alpha1.EnvVar{
+						{
+							Name:  "LITERAL",
+							Value: "plain",
+						},
+						{
+							Name: "ANTHROPIC_API_KEY",
```
juli4n: Nit: Let's use values which are not used in demos / manifest / production code as that makes it easier (removes false positives) to search replace when we *have* to change those. Maybe just "SOME_KEY"

cmd/ateapi/internal/controlapi/workload_spec.go
juli4n: We only need to resolve the env secrets in the ResumeActor call that boots up a new actor (today this is either the golden actor initial ResumeActor call or the first ResumeActor{boot:true} call of a regular actor (I'm not sure how important this use case is). Do we need a cache at all? Doing an extra Secret Get call during golden actor creation doesn't hurt much.

cmd/ateapi/internal/controlapi/workload_spec.go
juli4n: We should have guidelines for this, but today the model is: a) if you are reporting an error a you have an opinion about how that error maps to gRPC status, report it using the right code, otherwise return a regular go (which will bubble up and be mapped to INTERNAL). Also, if you are calling code that can return grpc status errors, and you are adding context to the error, wrap it.

internal/ateinterceptors/ateinterceptors.go
juli4n: Doing this as a follow up PR sounds very reasonable.

cmd/ateapi/internal/controlapi/workload_spec.go
Falven: > We only need to resolve the env secrets in the ResumeActor call that boots up a new actor (today this is either the golden actor initial ResumeActor call or the first ResumeActor{boot:true} call of a regular actor (I'm not sure how important this use case is). Do we need a cache at all? Doing an extra Secret Get call during golden actor creation doesn't hurt much.

+1 on reduced complexity/no cache here.

Falven (approved):

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -59,6 +53,55 @@ type Container struct {
...
+// subset of sources that the substrate control plane resolves: literal values,
+// and Secret keys.
+//
+// +kubebuilder:validation:XValidation:rule="!(has(self.value) && has(self.valueFrom))",message="value and valueFrom are mutually exclusive"
+type EnvVar struct {
+	// Name of the environment variable. Must be a C_IDENTIFIER.
+	// +required
+	Name string `json:"name"`
+
+	// Variable value. Defaults to "". Mutually exclusive with ValueFrom.
```
thockin, line 67: We should document the behavioral difference vs k8s proper wrt. variable interpolation in env values.  These are not interpolated, IIUC, while Kubernetes env vars are.

cmd/ateapi/internal/controlapi/workload_spec.go
thockin: This is a good point - golden startup is not such a big deal.  But unless I am misunderstanding, this is still called on every resume?

cmd/ateapi/internal/controlapi/workload_spec.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (r *envResolver) resolveValueFrom(ctx context.Context, envID string, valueFrom *atev1alpha1.EnvVarSource) (string, bool, error) {
+	if ref := valueFrom.SecretKeyRef; ref != nil {
+		return r.resolveSecretKeyRef(ctx, envID, ref)
+	}
+	return "", false, status.Errorf(codes.FailedPrecondition, "%s uses unsupported valueFrom source; only secretKeyRef is supported", envID)
+}
+
+func (r *envResolver) resolveSecretKeyRef(ctx context.Context, envID string, ref *atev1alpha1.SecretKeySelector) (string, bool, error) {
+	if ref.Name == "" {
```
thockin, line 107: Please at least drop a TODO that this should be handled by API validation rather than being done here

EItanya: So this is already done via CEL validation today, so this is basically just fallback. Would your rather I removed this inline validation step altogether?

EItanya:
The GitHub UI is freaking out and won't let me reply to this comment about golden startup:
https://github.com/agent-substrate/substrate/pull/20#discussion_r3374721412

That is correct that it's done for every resume, but I think we need to decide what we want to happen here, and I think the scope of it is big enough that it warrants a follow-up. Here are the solutions I can imagine:
1. Cache with a TTL which lives across calls and makes the lookups faster
2. Store the values inside of the snapshot for later grabs. (This means that snapshot storage will have secrets)

cmd/ateapi/internal/controlapi/workload_spec.go
thockin: Am I missing where this CEL is being done?

thockin:
> 1. Cache with a TTL which lives across calls and makes the lookups faster

Plausible

> 2. Store the values inside of the snapshot for later grabs. (This means that snapshot storage will have secrets)

Let's not do that!

Option 3: Distinguish cold-start from actual resume and populate those only on cold.  That said, I suspect we will want an optional mode which is does not use golden snapshots, so we will need to fall back on 1.

I'm a little anxious about committing this without a plan in place for how we intend to handle this.  It's the first case of having to handle this, so I apologize for the latency on it.

cmd/ateapi/internal/controlapi/workload_spec.go
thockin: I found it.  I apologize for missing it.  I am looking at adding validation for everything else.

Yes, I think the server logic should NOT re-handle validation.  It seems innocuous now but will spiral quickly.

EItanya:
> > 2. Store the values inside of the snapshot for later grabs. (This means that snapshot storage will have secrets)
> 
> Let's not do that!

Agreed I wasn't a big fan of storing it in the snapshot.

> Option 3: Distinguish cold-start from actual resume and populate those only on cold. That said, I suspect we will want an optional mode which is does not use golden snapshots, so we will need to fall back on 1.

The API doesn't allow this today, it takes in the same API object regardless, I think we could reason about that change but it would require changing the `atelet` protos to have different `"specs"` depending on start vs. resume.

From the `gvisor` code, it seems like all that is needed on restore is the container name, so I can propose proto changes to make this path work. That's probably the cleanest path forward

thockin:
> From the gvisor code, it seems like all that is needed on restore is the container name, so I can propose proto changes to make this path work. That's probably the cleanest path forward

But it won't solve the case of non-golden, which seems inevitable to me.

What if we just keep a cache with a 30 second TTL and synchronous refresh, and leave a TODO to make the refresh more intelligent.  Later, when we are sure this is a pattern we are REALLY keeping, we can do async refresh.

EItanya:
> > From the gvisor code, it seems like all that is needed on restore is the container name, so I can propose proto changes to make this path work. That's probably the cleanest path forward
> 
> But it won't solve the case of non-golden, which seems inevitable to me.
> 
> What if we just keep a cache with a 30 second TTL and synchronous refresh, and leave a TODO to make the refresh more intelligent. Later, when we are sure this is a pattern we are REALLY keeping, we can do async refresh.

I implemented this in https://github.com/agent-substrate/substrate/pull/20/commits/d30e4f9da7084c0cb04e13e6a5ff40cc7847bf30

cmd/ateapi/internal/controlapi/workload_spec.go
```diff
@@ -0,0 +1,205 @@
...
+
+type envResolver struct {
+	kubeClient kubernetes.Interface
+	namespace  string
+	cache      *envSecretCache
+}
+
+func (r *envResolver) resolve(ctx context.Context, containerName string, env atev1alpha1.EnvVar) (*ateletpb.EnvEntry, error) {
+	envID := fmt.Sprintf("container %q env %q", containerName, env.Name)
+	if env.Value != "" && env.ValueFrom != nil {
```
thockin, line 74: This can also be done at API validation, but if it is the only thing I flag, it can be a followup.  I have a PR adding more validation almost ready

thockin (commented):
Approve, I will rebase on this and send some validation cleanup

thockin:
```
diff --git a/manifests/ate-install/generated/ate.dev_actortemplates.yaml b/manifests/ate-install/generated/ate.dev_actortemplates.yaml
17
index 59682eb..bb84b50 100644
18
--- a/manifests/ate-install/generated/ate.dev_actortemplates.yaml
19
+++ b/manifests/ate-install/generated/ate.dev_actortemplates.yaml
20
@@ -71,9 +71,9 @@ spec:
21
                       items:
22
                         description: |-
23
                           EnvVar represents an environment variable supplied to a container in an
24
-                          ActorTemplate. It mirrors the shape of corev1.EnvVar but only models the
25
-                          subset of sources that the substrate control plane resolves: literal values,
26
-                          and Secret keys.
27
+                          ActorTemplate. It models only a subset of Kubernetes Pod env behavior:
28
+                          literal values are not expanded with Kubernetes-style $(VAR) references,
29
+                          envFrom is not supported, and valueFrom currently supports only secretKeyRef.
30
                         properties:
31
                           name:
32
                             description: Name of the environment variable. Must be
33
Run ./hack/update/go-generate.sh to update generated files.
```

PR #22, opened 2026-05-20
fix(ategcs): stop calling CreateBucket on every PutObject

brandonrjacobs:
## Summary

`s3Client.PutObject` called `CreateBucket` before every upload. Convenient against local dev backends like rustfs/minio where buckets may not pre-exist, but against managed S3-compatible backends the caller typically lacks `s3:CreateBucket`, so each snapshot upload paid for a 403 in added latency and audit-log noise.

For local kind dev, add a one-shot Job alongside the rustfs Deployment that creates the `ate-snapshots` bucket once at install time.

## Validation

Verified locally with a kind cluster:

- Applied namespace + `manifests/ate-install/kind/rustfs.yaml`
- `rustfs-bucket-init` Job retried once while rustfs came up, then created `ate-snapshots`
- `aws s3api list-buckets` against rustfs returns `ate-snapshots`

### DIFF
--- a/manifests/ate-install/kind/rustfs.yaml
+++ b/manifests/ate-install/kind/rustfs.yaml
@@ -102,7 +102,7 @@ spec:
       restartPolicy: OnFailure
       containers:
       - name: create-bucket
-        image: amazon/aws-cli:2.17.0
+        image: amazon/aws-cli:2.17.0@sha256:643507c10ada7964ca6157b3d799f030b90577643da9955d319a77399ed80d73
         env:
         - name: AWS_ACCESS_KEY_ID
           value: rustfsadmin
--- a/manifests/ate-install/kind/rustfs.yaml
+++ b/manifests/ate-install/kind/rustfs.yaml
@@ -89,3 +89,45 @@ spec:
       - name: data
         persistentVolumeClaim:
           claimName: rustfs-data
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: rustfs-bucket-init
+  namespace: ate-system
+spec:
+  backoffLimit: 10
+  template:
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: create-bucket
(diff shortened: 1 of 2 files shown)


BenTheElder:
Hmm, perhaps we should make the kind environment also pre-create the bucket on install and drop this entirely.

In general we want the CI environment to look like production where possible.

brandonrjacobs:
> Hmm, perhaps we should make the kind environment also pre-create the bucket on install and drop this entirely.
> 
> In general we want the CI environment to look like production where possible.

Yeah that would make sense; i wasn't sure what the general idea was but that works as well if we want to go that route.

brandonrjacobs:
@BenTheElder I dropped the create call and added it to kind setup if we want to go that way instead.

manifests/ate-install/kind/rustfs.yaml
```diff
@@ -89,3 +89,45 @@ spec:
...
+  name: rustfs-bucket-init
+  namespace: ate-system
+spec:
+  backoffLimit: 10
+  template:
+    spec:
+      restartPolicy: OnFailure
+      containers:
+      - name: create-bucket
+        image: amazon/aws-cli:2.17.0
```
BenTheElder: Let's pin the digest like the other images:

```suggestion
        image: amazon/aws-cli:2.17.0@sha256:643507c10ada7964ca6157b3d799f030b90577643da9955d319a77399ed80d73
```

BenTheElder (commented):
Thank you!

Unfortunately as far as I know [we require the CLA](https://github.com/agent-substrate/substrate/blob/18c8639f65b32b53825538bc43b89a689379afb1/CONTRIBUTING.md#sign-our-contributor-license-agreement) currently.

Otherwise aside from one minor comment I am +1

brandonrjacobs:
Working on getting that resolved!

brandonrjacobs:
CLA Signed

BenTheElder:
Thanks!

PR #25, opened 2026-05-20
declare initial kubernetes version support intent

BenTheElder:
Discussed this morning with other maintainers.

We know we need to test more with 1.36 (see #8), this is just the declaration of intent to start.


thockin:
LGTM

PR #29, opened 2026-05-20
Add trace ID to all log entries

ahmedtd:
This commit adds a small log handler that reads the trace id out of the context, if present, and includes it as a log attribute.  Additionally, ateom is integrated with tracing (although it has no exporter because it has no network connectivity).

This allows easy correlation of logs from a single parent request across ate-api-server, atelet, and ateom.

I have also fixed the potential for log tearing from the ateom runsc log exporter by having it reuse the same slog.Handler as the main program.

### DIFF
--- a/cmd/servers/ateapi/controlapi/service.go
+++ b/cmd/servers/ateapi/controlapi/service.go
@@ -15,17 +15,10 @@
 package controlapi
 
 import (
-	"context"
-	"errors"
-	"log/slog"
-
 	"github.com/agent-substrate/substrate/cmd/servers/ateapi/store"
 	listersv1alpha1 "github.com/agent-substrate/substrate/pkg/client/listers/api/v1alpha1"
 
 	"github.com/agent-substrate/substrate/proto/ateapipb"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 )
 
 // Service implements ateapipb.Control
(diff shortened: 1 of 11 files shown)


cmd/servers/ateapi/controlapi/service.go
```diff
@@ -50,25 +43,3 @@ func NewService(persistence store.Interface, actorTemplateLister listersv1alpha1
 
 	return s
 }
-
-// StatusErrorInterceptor searches the error chain for a gRPC status error.
```
juli4n, line 54: As discussed over chat, let's keep this.

internal/contextlogging/contextlogging.go
```diff
@@ -0,0 +1,61 @@
...
+		internal: internal,
+	}
+}
+
+func (h *ContextHandler) Enabled(ctx context.Context, lvl slog.Level) bool {
+	return h.internal.Enabled(ctx, lvl)
+}
+
+func (h *ContextHandler) Handle(ctx context.Context, rec slog.Record) error {
+	rpcIDVals := metadata.ValueFromIncomingContext(ctx, "ate-rpc-id")
```
juli4n: What's this used for?

ahmedtd: ooops, leftover.  I had been planning to also generate and log an RPC ID, but it turns out that trace ID is always populated.

ahmedtd: Removed.

internal/contextlogging/contextlogging.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (h *ContextHandler) Enabled(ctx context.Context, lvl slog.Level) bool {
+	return h.internal.Enabled(ctx, lvl)
+}
+
+func (h *ContextHandler) Handle(ctx context.Context, rec slog.Record) error {
+	spanContext := trace.SpanContextFromContext(ctx)
+	if spanContext.HasTraceID() {
+		traceID := spanContext.TraceID()
+		rec.AddAttrs(slog.String("ate.dev/trace-id", hex.EncodeToString(traceID[:])))
```
juli4n: Why not just traceID here? It's already coding to hexa

internal/ateinterceptors/ateinterceptors.go
```diff
@@ -0,0 +1,56 @@
...
+			GRPCStatus() *status.Status
+		}
+
+		if errors.As(err, &statusErr) {
+			st := statusErr.GRPCStatus()
+			return nil, status.Error(st.Code(), st.Message())
+		}
+
+		// No status error found in chain.
+		return nil, status.Error(codes.Internal, "internal server error")
```
juli4n, line 52: I thought you wanted to include the whole error message here?

juli4n (approved):

internal/ateinterceptors/ateinterceptors.go
ahmedtd: I think this is fine for now.  Let's see if it actually becomes a pain point.

PR #32, opened 2026-05-21
docs: clarify kind prereqs and fix stale install-script references

dims:
- README Quickstart now pins kind >= v0.31.0. Earlier releases ship a default Kubernetes node image without the PodCertificateRequest API the install scripts depend on, so kube-apiserver never reaches readiness on a fresh `brew install kind` machine.
- internal/e2e/readme.md referenced a `hack/install.sh` that does not exist. Point readers at the two real entry points instead (hack/install-ate.sh for GKE, hack/install-ate-kind.sh for kind).
- demos/counter/README.md claimed a GCS bucket was required, but the kind overlay provisions an in-cluster rustfs S3 store. Document both targets so the demo can run without provisioning GCP.

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ goals in the near term.
 
 To quickly set up the complete environment:
 
-1. Make sure you have [Go](https://go.dev/doc/install), [`kubectl`](https://kubernetes.io/docs/tasks/tools/), and [`docker`](https://www.docker.com/) installed and configured on your dev machine. We will automatically manage other dependencies via Go, including [`kind`](https://kind.sigs.k8s.io/).
+1. Make sure you have [Go](https://go.dev/doc/install), [`kubectl`](https://kubernetes.io/docs/tasks/tools/), and [`docker`](https://www.docker.com/) installed and configured on your dev machine. We will automatically manage other dependencies via Go, including [`kind`](https://kind.sigs.k8s.io/). On Docker Desktop, allocate at least **12 GiB of memory** to the Docker VM (Settings → Resources → Memory) — the full bring-up otherwise OOMs.
 
 2. Run the following steps:
 ```shell
--- a/demos/counter/README.md
+++ b/demos/counter/README.md
@@ -6,9 +6,20 @@ It deploys a simple Go HTTP server (`counter.go`) that increments a counter on e
 
 ## Prerequisites
 
-- A k8s cluster with Agent Substrate installed (`./hack/install-ate.sh --deploy-ate-system`).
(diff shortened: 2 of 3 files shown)


BenTheElder:
Thanks, one thing: I'd like to eliminate those pre-reqs as much as possible https://github.com/agent-substrate/substrate/pull/46

README.md
```diff
@@ -87,7 +87,10 @@ goals in the near term.
 
 To quickly set up the complete environment:
 
-1. Make sure you have [`kind`](https://kind.sigs.k8s.io/), [`kubectl`](https://kubernetes.io/docs/tasks/tools/), and [`docker`](https://www.docker.com/) installed and configured on your dev machine.
+1. Make sure the following tools are installed and configured on your development machine:
+   - [`kind`](https://kind.sigs.k8s.io/) **v0.31.0 or newer** (required for the `PodCertificateRequest` API the install scripts use; matches the CI-pinned `go install sigs.k8s.io/kind@v0.31.0`).
```
BenTheElder: not required or used anymore

README.md
```diff
@@ -87,7 +87,10 @@ goals in the near term.
...
+   - [`kubectl`](https://kubernetes.io/docs/tasks/tools/).
+   - [`docker`](https://www.docker.com/) with **at least 12 GiB of memory** allocated to the VM (Docker Desktop: Settings → Resources → Memory).
```
BenTheElder: are we using 12GiB? 

must be the image cache ... we're going to need to think about tiers there

README.md
dims: @BenTheElder we still use ( https://github.com/agent-substrate/substrate/blob/2f2292f17109c99121cd82a93cc9b93dbaef00ac/.github/workflows/pr-workflow.yaml#L42 ) 

should we reflect that? (and drop the note about `PodCertificateRequest` API)

README.md
dims: 12 seemed to work :) i did not try going lower

README.md
BenTheElder: If you DIY install we don't inherently depend on kind, but our dev scripts will depend on the version in the tools module in the future, I think we don't need to specify the version here. Thanks.

I also think we will stop depending on the default node image in the future, but I want to ship some upstream kind fixes first.

README.md
BenTheElder: Hmm OK, I want to avoid overstating it if we can, this is also likely going to be very fluid as we iterate, most of it is probably various in-memory storage currently that doesn't strictly need to all be in memory in the future ...

PR #33, opened 2026-05-21
hack: tear down the kind overlay symmetrically on --delete-all

dims:
`install-ate-kind.sh --deploy-ate-system` installs the kind overlay (rustfs and the `otel-system` namespace) via a kustomize build of `manifests/ate-install/kind`, but the matching teardown only removed `kind/rustfs.yaml`, and only when `ATE_INSTALL_RUSTFS=true` was set in the environment.

Teardown now deletes the same kustomize set whenever `ATE_INSTALL_KIND=true` (already exported by `install-ate-kind.sh`). The GKE path is unchanged; `ATE_INSTALL_RUSTFS` is unused after this change.

- [x] Tests pass

### DIFF
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -313,10 +313,12 @@ deploy_atenet() {
 
 delete_ate_system() {
   log_step "delete_ate_system"
-  if [[ "${ATE_INSTALL_RUSTFS:-false}" == "true" ]]; then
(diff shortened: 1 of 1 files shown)


dims:
cc @BenTheElder

BenTheElder (approved):
thanks!

PR #34, opened 2026-05-21
manifests: add metrics pipeline to the kind otel-collector

dims:
atelet and ate-api-server push traces and metrics via OTLP. The kind collector declared only a `traces:` pipeline, so each metric push returned `Unimplemented`; the GKE managed collector accepts both.

Add a matching `metrics:` pipeline, re-exported via a Prometheus endpoint on `:8889`. Server-side configuration is unchanged.

- [x] Tests pass

### DIFF
--- a/manifests/ate-install/kind/otel-collector.yaml
+++ b/manifests/ate-install/kind/otel-collector.yaml
@@ -38,6 +38,8 @@ data:
         endpoint: jaeger.otel-system.svc.cluster.local:4317
         tls:
           insecure: true
+      prometheus:
+        endpoint: 0.0.0.0:8889
       debug:
         verbosity: detailed
     service:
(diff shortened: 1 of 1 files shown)


manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -38,6 +38,8 @@ data:
         endpoint: jaeger.otel-system.svc.cluster.local:4317
         tls:
           insecure: true
+      prometheus:
+        endpoint: 0.0.0.0:8889
       debug:
         verbosity: detailed
```
juli4n, line 44: Not strictly related to your change, but we should probably switch to "normal" to not spam the logs.

juli4n (changes requested):

manifests/ate-install/kind/otel-collector.yaml
juli4n: Let's not block this PR on this. We can send a separate PR.

juli4n (approved):

BenTheElder (approved):

PR #35, opened 2026-05-21
hack: pre-fetch runsc on the kind node before deploy_ate_system

dims:
The atelet `Run` RPC fetches the gVisor runsc binary (~79 MiB) from public GCS on first invocation, under a 28s per-actor workflow deadline (`cmd/servers/ateapi/controlapi/workflow.go`). Slower networks miss the deadline; the ActorTemplate then stays in `phase: ResumeGoldenActor` while atelet restarts the download on every reconcile.

When `ATE_INSTALL_KIND=true`, parse the runsc URL and sha256 from `manifests/counter/counter.yaml.tmpl`, download once on the host, and copy the binary into `/run/ateom-gvisor/static-files/` on each node returned by `kind get nodes`. atelet's existing cache lookup short-circuits subsequent activations. The GKE install path is unchanged.

- [x] Tests pass

### DIFF
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -206,8 +206,73 @@ deploy_crds() {
   run_ko apply -f manifests/ate-install/generated
 }
 
+# When running on kind, pre-populate every node's atelet hostPath cache with
+# the gVisor runsc binary referenced by the demo templates. The atelet `Run`
+# RPC fetches runsc from public GCS the first time it is invoked, which can
+# exceed the per-actor lock TTL (~28s) on networks where a ~79 MiB anonymous
+# GCS pull is slow. Pre-fetching here removes that race for the documented
+# kind quickstart. No effect on the GKE install path.
+ensure_runsc_cached_on_kind() {
+  [[ "${ATE_INSTALL_KIND:-false}" == "true" ]] || return 0
+  log_step "ensure_runsc_cached_on_kind"
+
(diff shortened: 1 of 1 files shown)


BenTheElder:
Hmm, this seems like awkward coupling, we don't just want to support counter demo and you can have different versions for different actors / snapshots.

cc @ahmedtd

EItanya:
Can't we just bake it into the `Dockerfile`?

BenTheElder:
> Can't we just bake it into the Dockerfile?

We could build a base image and bake it in, a bit annoying since we're using stock images and ko (and not hosting any images currently)

The real issue though is mentioned above ... we intentionally support having varying versions of this binary. So baking it in doesn't really solve the problem, except for this particular demo.

It's intended that a snapshot can be paired with a specific version of the binary, which is why we're having to read it from the counter demo.

ahmedtd:
Yes, the constraint we're operating under here is that gVisor snapshots are tightly coupled to the version of the binary that made them, there is neither forward- nor backward-compatibility.

We could have atelet download the "most likely" runsc versions during startup, which would be a less-coupled way to get the same effect.

Also, we should probably remove or rework the deadline imposed on the resume and suspend workflows --- @juli4n .  If we are using redis locks for correctness, we should continue to refresh the lock while the workflow runs.  But I think we probably don't need the locks for correctness.

dims:
i'll close this then, we can figure out some other way :)

PR #36, opened 2026-05-21
manifests: switch demos to the off-GCP pause image (use community assets!)

dims:
`api/v1alpha1/actortemplate_types.go:64-67` documents the GKE `pause` digest as the on-GCP choice and `registry.k8s.io/pause:3.10.2@…` as the off-GCP choice. All four demo templates currently pin the on-GCP digest, which puts the local kind quickstart on a Google-hosted registry by default.

The API doc comment keeps both references; on-GCP deployments can still pin the GKE digest explicitly.

- [ ] Tests pass


BenTheElder (approved):
funnily enough I had this review comment before it went public but we just didn't think it was worth changing at the time :-)

PR #39, opened 2026-05-21
Fix atelet otel kustomize

### LINKED ISSUE #38: Atelet metrics are not exported to OTEL collector when running on Kind
We kustomize ate-api-server-deployment to point to OTEL_EXPORTER_OTLP_ENDPOINT, but we are missing atelet DaemonSet.

juli4n:
We were doing it for `ate-apiserver`, but not `atelet`. This results in `atelet` failing to ingest its own metrics.

Fixes https://github.com/agent-substrate/substrate/issues/38


BenTheElder (approved):
Thanks!!

PR #40, opened 2026-05-21
fix(ategcs): surface GCS upload error during checkpoint

azr:
## Symptom

`atelet`'s `Checkpoint` RPC returns `err: nil` even when the GCS upload it performs silently failed. The destination bucket stays empty (no `checkpoint.img.zstd`, no `pages.img.zstd`), and every subsequent actor `Resume` then fails with `storage: object doesn't exist`, which looks like a Resume bug but is actually a checkpoint-upload bug.

## Root cause

[`internal/ategcs/ategcs.go:51`](https://github.com/agent-substrate/substrate/blob/main/internal/ategcs/ategcs.go#L51) — `gcsClient.PutObject`:

```go
wc := g.client.Bucket(bucket).Object(object).NewWriter(ctx)
defer wc.Close()
_, err := io.Copy(wc, reader)
return err
```

For the GCS Go client, `io.Copy` only writes into a client-side buffer; the actual HTTP PUT (and any API-level error like 403 from an OAuth scope mismatch) is returned from `Writer.Close()`. The `defer wc.Close()` here discards exactly that error, so PutObject returns `nil` even when the upload was rejected.

## Fix

Return the error from `wc.Close()`. On `io.Copy` failure still best-effort-close the writer but report the copy error, matching the existing `slog.InfoContext("Dropped error ...")` pattern used elsewhere in this file. Minimal, localised change; no API surface change.

## How I noticed this

Evaluating Substrate on a GKE Standard 1.35.3 cluster (with `--enable-kubernetes-unstable-apis=certificates.k8s.io/v1beta1/{podcertificaterequests,clustertrustbundles}`) where the node pool's OAuth scope was `devstorage.read_only`. That's my misconfig — pods can't write to GCS — but the user-visible symptom was confusing: `atelet.Checkpoint` reported success in ~282ms, `gs://ate-snapshots-.../` was empty, and the first user-facing error came from a completely unrelated actor's Resume call. Surfacing the upload error at the checkpoint call site makes the misconfig immediately diagnosable.

## Verification

- `GOOS=linux go build ./...` passes locally.
- No new dependencies, no new logging frameworks, single file touched.

Leaving as **draft** for maintainer eyes before marking ready — happy to add a unit test against a fake `ObjectStorage` if you'd like, but didn't want to bloat the diff before getting a signal on the approach.

### DIFF
--- a/internal/ategcs/ategcs.go
+++ b/internal/ategcs/ategcs.go
@@ -16,6 +16,7 @@ package ategcs
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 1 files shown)


BenTheElder (approved):
thanks!

PR #51, opened 2026-05-21
ActorTemplate: Enforce that all images must be pinned

### LINKED ISSUE #10: Prevent user from creating ActorTemplates with un-pinned images
Today, if you put a non-pinned image reference in `Spec.Containers[].Image`, the workload will start up, but the image will always be a cache miss in the atelet image pull cache.

However, if the actual image pointed by the tag changes, this will invalidate all gvisor snapshots that were based on the old image.

We should using a ValidatingAdmissionPolicy to ensure that all images in an ActorTemplate are pinned to a particular SHA, and we should prevent mutation of the ActorTemplate once it is created.

ahmedtd:
This commit adds a ValidatingAdmissionPolicy that checks that all images in an ActorTemplate must be pinned to a specific hash.

Fixes #10

### DIFF
--- a/pkg/api/v1alpha1/actortemplate_types.go
+++ b/pkg/api/v1alpha1/actortemplate_types.go
@@ -37,6 +37,8 @@ type Container struct {
 	Name string `json:"name"`
 
 	// Image to use for the worker replicas.
+	//
+	// +kubebuilder:validation:XValidation:rule="self.contains('@')",message="All images must be pinned (changing the image invalidates snapshots)"
 	Image string `json:"image,omitempty"`
 
 	// Entrypoint array. Not executed within a shell.
@@ -67,11 +69,13 @@ type ActorTemplateSpec struct {
 	//   - [2] registry.k8s.io/pause:3.10.2@sha256:f548e0e8e3dc1896ca956272154dde3314e8cc4fde0a57577ee9fa1c63f5baf4
 	//
 	// +required
+	// +kubebuilder:validation:XValidation:rule="self.contains('@')",message="All images must be pinned (changing the image invalidates snapshots)"
 	PauseImage string `json:"pauseImage,omitempty"`
 
 	// Containers is the workload definition.
 	//
 	// +optional
+	// +kubebuilder:validation:MaxItems=10
 	Containers []Container `json:"containers,omitempty"`
 
 	// Snapshots configuration for the actor.
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -343,11 +343,11 @@ func createTemplate(t *testing.T, tc *testContext, ns string) {
 					SHA256Hash: "a397be1abc2420d26bce6c70e6e2ff96c73aaaab929756c56f5e2089ea842b63",
 				},
 			},
-			PauseImage: "pause",
+			PauseImage: "pause@sha256:abc",
 			Containers: []atev1alpha1.Container{
 				{
 					Name:    "main",
-					Image:   "main",
+					Image:   "main@sha256:abc",
 					Command: []string{"/main"},
 				},
 			},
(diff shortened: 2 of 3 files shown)


manifests/ate-install/actortemplate-vap.yaml
```diff
@@ -0,0 +1,39 @@
+#  Copyright 2026 Google LLC
```
juli4n: Have you considered adding this validation as an XValidation in the CRD itself? The advantage is that it gets coupled to the CRD so there is no way to delete it. I don't know this, but I suspect error messages will be more precise too as the error is specific to the field.

BenTheElder: Good point. We need VAP for cross-field work, but this is scoped to a single field.

https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2876-crd-validation-expression-language/README.md

ahmedtd: Done, it seems to work.

juli4n (dismissed):
You'll need to rebase on main and move this as the CRD types are under pkg/api/v1alpha1 now.

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -31,6 +31,8 @@ const (
 )
 
 // A single application container that you want to run within a WorkerPool.
+//
+// +kubebuilder:validation:XValidation:rule="self.image.contains('@')"
```
BenTheElder: `./hack/update/go-generate.sh` should update the CRD yaml to pick this up

BenTheElder (dismissed):

juli4n (dismissed):

BenTheElder:
https://github.com/agent-substrate/substrate/actions/runs/26594400905/job/78361379429?pr=51

> envtest start failed: unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "actortemplates.ate.dev": CustomResourceDefinition.apiextensions.k8s.io "actortemplates.ate.dev" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[containers].items.x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared), spec.validation.openAPIV3Schema.properties[spec].properties[containers].items.x-kubernetes-validations[0].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema, spec.validation.openAPIV3Schema: Forbidden: x-kubernetes-validations estimated rule cost total for entire OpenAPIv3 schema exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)]

juli4n:
> https://github.com/agent-substrate/substrate/actions/runs/26594400905/job/78361379429?pr=51
> 
> > envtest start failed: unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "actortemplates.ate.dev": CustomResourceDefinition.apiextensions.k8s.io "actortemplates.ate.dev" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[containers].items.x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared), spec.validation.openAPIV3Schema.properties[spec].properties[containers].items.x-kubernetes-validations[0].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema, spec.validation.openAPIV3Schema: Forbidden: x-kubernetes-validations estimated rule cost total for entire OpenAPIv3 schema exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)]

Should we set a limit in the number of containers on a given actor template spec?

ahmedtd:
Uh, I thought I tested this locally, and saw it correctly rejecting images without '@' symbols. Maybe I had forgotten to remove the VAP?

juli4n (dismissed):

ahmedtd:
OK, I've limited us to 10 containers.  I also applied the check to the pause image, which has the same problem.

PR #56, opened 2026-05-21
feat: add script to update tool dependencies in hack/tools directory

BenTheElder:
We are just using tools modules (plural) to manage installing go binaries with binned versions easily.
We are NOT trying to patch transitive dependencies in those tools, versus just updating those tools themselves.

Add a helper script to update them as intended, and add and AGENTS.md to tell agents to use the script.

- [x] Tests pass

### DIFF
--- a/hack/update-tool.sh
+++ b/hack/update-tool.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


hack/update-tool.sh
```diff
@@ -0,0 +1,82 @@
...
+  exit 1
+fi
+
+if [ ! -f "${TOOL_DIR}/go.mod" ]; then
+  echo "Error: ${TOOL_DIR}/go.mod does not exist" >&2
+  exit 1
+fi
+
+# Extract tool import paths declared in go.mod
+tool_paths=()
```
juli4n, line 54: Nit: can this somehow be simplified using go mod edit -json + jq?

BenTheElder: found another solution that is simpler than both, pushing

juli4n (approved):

PR #57, opened 2026-05-21
fix failing kind demo commands in README

### LINKED ISSUE #55: kind demo commands in README failed
## Expected Behavior
No errors

## Actual Behavior
```shell
$ kubectl port-forward -n ate-system svc/atenet-router 8000:80 &
curl -X POST -H "Host: my-counter-1.actors.resources.substrate.ate.dev" -i http://localhost:8000/

[1] 2974034
curl: (7) Failed to connect to localhost port 8000 after 0 ms: Could not connect to server
```

Run command in README
```shell
kubectl port-forward -n ate-system svc/atenet-router 8000:80 &
curl -X POST -H "Host: my-counter-1.actors.resources.substrate.ate.dev" -i http://localhost:8000/
```

## Specifications

- Version:
- Platform:

troychiu:
Fixes #55

I was following the demo but ran into the issues described in #55. The command is failing because the curl command runs before the port-forwarding is ready.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,13 @@ go install ./cmd/kubectl-ate
 
 # create a counter actor and demo it
 kubectl ate create actor my-counter-1 --template ate-demo-counter/counter
-kubectl port-forward -n ate-system svc/atenet-router 8000:80 &
+
+# port-forward the network router to bind to local port `8000`
+kubectl port-forward -n ate-system svc/atenet-router 8000:80
+```
+
+3. In a **separate terminal**, send an HTTP request to increment the counter:
+```shell
 curl -X POST -H "Host: my-counter-1.actors.resources.substrate.ate.dev" -i http://localhost:8000/
 ```
 


README.md
```diff
@@ -105,7 +105,11 @@ go install ./cmd/kubectl-ate
 
 # create a counter actor and demo it
 kubectl ate create actor my-counter-1 --template ate-demo-counter/counter
-kubectl port-forward -n ate-system svc/atenet-router 8000:80 &
+
+# port-forward the network router to bind to local port `8000`:
+kubectl port-forward -n ate-system svc/atenet-router 8000:80
```
BenTheElder, line 110: Copy pasting and running this entire snippet still won't work, since it will stop at this command not exiting.


I think instead we might need to split this into multiple code boxes and make it clear that you can copy paste up until this command at once, but the curl request is a follow-up.

We can specify to run the curl command in a second shell.

troychiu: That makes sense! Updated. Let me know if this looks okay to you.

BenTheElder (approved):
thanks!

PR #58, opened 2026-05-22
Reorganize code

### LINKED ISSUE #49: Reorganize code layout
The current code layout grew fast and organically, and doesn't follow any rule / invariant.

We have most Substrate binaries under /cmd/<component>/... although each one structures its packages in a different way. We also have api/, pkg/ internal/ etc.

This proposal tries to introduce some rules so we can have:

* Packages that can be imported by downstream projects in a safe way (no breakages).
* Clean separation of production code vs dev tooling.
* A common place for internal shared packages.
* A common place for internal per component packages.

A good starting point could be:

* `cmd/<component>/main.go` for the component main packages / entrypoints.
* `internal/<component>` for component specific specific packages. 
* `internal` for internal shared packages (across components). 
* `pkg` for the shared public packages.
* `hack/` dev scripts.
* `tools/<tool-name>`: dev tools (e.g. benchmarking, load testing).

juli4n:
This PR has three commits, the first one introduces a tool to move / patch files within the repo to align with the rules enumerated in https://github.com/agent-substrate/substrate/issues/49.

The second commit was automatically generated by running the tool (and go generate). This way we can rebase on top of inflight PRs, making the migration hopefully less painful.

The third one deletes the one-off tool.

Once the migration is done the tool will be removed so this PR should probably be not squashed (i.e. rebase and merge) to ease the rollback.

Fixes https://github.com/agent-substrate/substrate/issues/49


dberkov (dismissed):

thockin:
Notes:

./tools is ambiguous WRT hack/tools and the only thing in tools/ is "setup" which is GCP specific (rename to setup-gcp?).  What else might go in ./tools ?  This is my only blocking question.

./pkg/api and ./pkg/proto are ambiguous -- they are both APIs, but for different layers.  OK for now, but I expect we will want clarity on that eventually.  Similarly ./pkg/client.  Non-blocking

juli4n:
> Notes:
> 
> ./tools is ambiguous WRT hack/tools and the only thing in tools/ is "setup" which is GCP specific (rename to setup-gcp?). What else might go in ./tools ? This is my only blocking question.

Renamed tools/setup to tools/setup-gcp.

> 
> ./pkg/api and ./pkg/proto are ambiguous -- they are both APIs, but for different layers. OK for now, but I expect we will want clarity on that eventually. Similarly ./pkg/client. Non-blocking

Agreed.

I also added a third commit that deletes the tool at the tip.

thockin:
LGTM

thockin:
Fire in the hole!

PR #60, opened 2026-05-22
Reoganize demo folders

dberkov:
Reorganize demo folders by moving manifests files inside the demo folder from shared manifest folder


juli4n (commented):
Should we also move the other two demos? (agent-secret and claude-code-multiplex)

dberkov:
@juli4n - yep, you are right, I missed those 2 other demos

juli4n (approved):

PR #61, opened 2026-05-22
Make agent-secret demo route correctly: inject session header + fix DNS suffix

dims:
Two compact changes that together make the `agent-secret` demo run as documented:

- **`feat(atenet/router)`** — inject the resolved actor ID as an `x-agentset-session` request header on every successful resume, so workloads can identify their own actor without parsing the rewritten `:authority`.
- **`fix(demos/agent-secret)`** — replace the stale `substrate.k8s.io` DNS suffix with `substrate.ate.dev` (the live router suffix per `internal/resources/actor.go:27`) in three curl examples and two source comments.

Running the demo's README verbatim today fails in two ways:

1. The `Host` header uses the legacy `substrate.k8s.io` suffix and is rejected by the router with a 404.
2. Even with the correct suffix, the router rewrites `:authority` to the worker pod IP before the workload sees it. The workload's fallback prefix-extraction then grabs the first octet of the IP (`10` for `10.244.0.30`), `SuspendActor(actor_id="10")` returns `NotFound`, and the actor never self-suspends.

The header-injection change is also useful beyond this demo: any future workload that needs its own actor identity now has a stable header to read it from, instead of either reverse-engineering `:authority` or relying on an environment-variable mount.

Testing:
```
go build -buildvcs=false ./cmd/servers/atenet/... ./demos/agent-secret/...    # exit 0
go vet                  ./cmd/servers/atenet/app/router/... ./demos/agent-secret/...   # exit 0
go test -count=1        ./cmd/servers/atenet/app/router/...                            # ok 0.787s
```

End-to-end on a local kind cluster:
- `curl -H "Host: my-agent.actors.resources.substrate.ate.dev" http://localhost:8000` (no other headers) — workload now logs `DEBUG: Identified ActorID: [my-agent]`, the response reads `Session: my-agent` (was `Session: 10`), and the actor flips to `STATUS_SUSPENDED` ~9 s after the response.
- RAM-state preservation (same `SECRET-NNNN` returned across resume cycles, including worker-pod migration) is unchanged.

### DIFF
--- a/cmd/servers/atenet/app/router/extproc_out.go
+++ b/cmd/servers/atenet/app/router/extproc_out.go
@@ -43,6 +43,21 @@ func addAuthorityMutation(auth string, mut *extprocv3.HeaderMutation) {
 	)
 }
 
+// sessionHeader is set on every successfully-routed request so that the
+// workload can identify its own actor without parsing the rewritten :authority.
+const sessionHeader = "x-agentset-session"
+
+func addSessionHeaderMutation(actorID string, mut *extprocv3.HeaderMutation) {
+	mut.SetHeaders = append(mut.SetHeaders,
+		&corev3.HeaderValueOption{
+			Header: &corev3.HeaderValue{
+				Key:      sessionHeader,
+				RawValue: []byte(actorID),
+			},
+		},
+	)
+}
+
 func immediateResponse(statusCode envoy_type.StatusCode, message string) *extproc.ProcessingResponse {
 	return &extproc.ProcessingResponse{
 		Response: &extproc.ProcessingResponse_ImmediateResponse{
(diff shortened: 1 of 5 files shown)


cmd/servers/atenet/app/router/extproc_out.go
```diff
@@ -43,6 +43,21 @@ func addAuthorityMutation(auth string, mut *extprocv3.HeaderMutation) {
...
 
+// sessionHeader is set on every successfully-routed request so that the
+// workload can identify its own actor without parsing the rewritten :authority.
+const sessionHeader = "x-agentset-session"
+
+func addSessionHeaderMutation(actorID string, mut *extprocv3.HeaderMutation) {
+	mut.SetHeaders = append(mut.SetHeaders,
+		&corev3.HeaderValueOption{
+			Header: &corev3.HeaderValue{
+				Key:      sessionHeader,
```
dberkov, line 54: what will be a behavior  if caller passes same header in the request?

dberkov:
The agent-secret demo was developed while ago, before atenet module was developed and recently migrated to this repository. 
This is a historical reason why it checks for "x-agentset-session" header. 

Not sure it worth adding new header, since actorId can be extracted directly from host name.

dims:
> Not sure it worth adding new header, since actorId can be extracted directly from host name.

Ack, will close this PR. If there's a better way to fix this demo, let's do that!

PR #63, opened 2026-05-22
chore(hack): add golangci-lint to developer and CI workflows

dims:
Adds golangci-lint v2.12.2 using the existing per-tool-module pattern. The new verifier is auto-discovered by `hack/verify-all.sh`, so the existing `verify` step in `pr-workflow.yaml` runs it — no workflow YAML changes.

- Tool module + wrappers: `hack/tools/golangci-lint/`, `hack/{,verify/,update/}golangci-lint.sh`.
- `.golangci.yaml` with the standard linter set (`errcheck`, `govet`, `ineffassign`, `staticcheck`, `unused`).
- `make lint` and `make lint-fix`.
- Small code adjustments so the first run is clean: two dead-code deletions, two empty-branch fixes, one return-bool simplification, and inline `//nolint` for six audit-tracked or intentional sites.


hack/update/golangci-lint.sh
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+set -o errexit -o nounset -o pipefail
+
+ROOT="$(git rev-parse --show-toplevel)"
+cd "${ROOT}"
+
+# See hack/verify/golangci-lint.sh for why the tool path is resolved under
+# the host GOOS and then re-exec'd under GOOS=linux.
+BIN="$("${ROOT}"/hack/run-tool.sh --print-bin-path golangci-lint)"
```
thockin: As many false positives as we get sometimes, I don't know if we want to have an "update"?

Makefile
```diff
@@ -69,6 +69,16 @@ fmt:
...
 	@./hack/verify/gofmt.sh
 
+.PHONY: lint lint-fix
+
+# Runs golangci-lint and fails on any reported issues.
+lint:
+	@./hack/verify/golangci-lint.sh
+
+# Runs golangci-lint with --fix, applying auto-fixes in place.
+lint-fix:
```
thockin: Same as other - do we really want this?

hack/golangci-lint.sh
thockin, line 1: This file is never used?

demos/claude-code-multiplex/ui/server.go
```diff
@@ -493,7 +493,7 @@ func handleGiveTask(w http.ResponseWriter, r *http.Request) {
 }
 
 func main() {
-	rand.Seed(time.Now().UnixNano())
+	rand.Seed(time.Now().UnixNano()) //nolint:staticcheck // SA1019: TODO remove; seeding is a no-op since Go 1.20.
```
thockin: Why not just remove this line?

cmd/servers/atenet/app/router/router.go
```diff
@@ -286,7 +286,7 @@ func (s *RouterServer) Run(ctx context.Context) error {
 		g.Go(func() error {
 			go func() {
 				if err := httpServer.Serve(listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
-					// silent shutdown
+					slog.WarnContext(ctx, "status HTTP server exited unexpectedly", slog.Any("err", err))
```
thockin: I think this should be Error

hack/update/golangci-lint.sh
dims: Dropping

Makefile
dims: Dropping

hack/golangci-lint.sh
dims: Dropping

demos/claude-code-multiplex/ui/server.go
dims: agree. doing!

cmd/servers/atenet/app/router/router.go
dims: yep

thockin:
LGTM

PR #66, opened 2026-05-23
fix(ateom-gvisor): make eth0 move/restore idempotent + roll back on error

dims:
`ateom-gvisor` moves the pod's `eth0` interface into the actor's interior netns before invoking `runsc run`/`runsc restore`. The current code assumes `eth0` is in the pod netns at the start of every `RunWorkload`/`RestoreWorkload`. If a previous call fails mid-flight, `eth0` is left stranded in the interior netns and every subsequent call on the same worker fails with `eth0: Link not found` until the pod is restarted.

This change adds two complementary safeguards:

1. **`ensureEth0InPodNetns`** at the top of both entry points. Moves `eth0` back to the pod netns when it is currently in the interior; no-op otherwise. Recovers from a previously-aborted call.
2. **Deferred rollback** after the eth0-into-interior move. Returns `eth0` to the pod netns if the current call returns an error.

Hammer `CreateActor` + `ResumeActor` + `SuspendActor` cycles against a worker pod. Before this change, every second iteration trips the original failure path and the third leaves the pod unusable. After this change, the failure does not recur.

### DIFF
--- a/cmd/ateom-gvisor/main.go
+++ b/cmd/ateom-gvisor/main.go
@@ -195,7 +195,36 @@ func NewService(interiorNetNS netns.NsHandle, eth0LinkInfo *SaveLinkInfo, actorL
 	return svc
 }
 
-func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkloadRequest) (*ateompb.RunWorkloadResponse, error) {
+// ensureEth0InPodNetns moves eth0 back to the pod netns if a prior
+// Run/Restore left it stuck in the interior netns. Idempotent: returns
+// nil if eth0 is already in the pod netns or absent from both.
+func ensureEth0InPodNetns(ctx context.Context, s *AteomService) error {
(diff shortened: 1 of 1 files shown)


dims:
@AlexBulankou is this you asking?

AlexBulankou:
@dims the [question](https://github.com/agent-substrate/substrate/pull/66#issuecomment-4545292683) was not from me, but from my bot on my standing instructions. It identified the differences in how your own PR #75 handled state of failure and recovery compared to this PR. 
The reason it brought it up is that standing request to this bot was to comment on PRs where it identified any inconsistencies between technical approaches across multiple PRs in progress to ensure consistency across the codebase.
Let me know if you believe the comment is out of place or misses your approach altogether, this would be helpful context to the bot.

ahmedtd (approved):

ahmedtd:
@dims This looks good.  Can you rebase?

dims:
> [Davanum Srinivas (@dims)](https://github.com/dims) This looks good. Can you rebase?

On it!

PR #67, opened 2026-05-23
feat(install): build + push ateom-gvisor image during kind bootstrap

dims:
`ateom-gvisor` is the per-worker-pod sidecar referenced via `WorkerPool.spec.ateomImage`.

Its source lives under `cmd/ateom-gvisor`, but no `Deployment` or `DaemonSet` manifest under `manifests/ate-install/` references it. The `ko resolve` pipeline that builds every other binary therefore never builds `ateom-gvisor`.

After a fresh `hack/install-ate-kind.sh --deploy-ate-system`, operators creating a `WorkerPool` have to either:

- run a packaged demo (whose template's `ko://` reference side-effects the build), or
- invoke `ko publish ./cmd/ateom-gvisor` by hand and substitute the resulting digest into their manifest.

Add `publish_ateom_image` to `hack/install-ate.sh`. It runs:

```
ko publish --base-import-paths ./cmd/ateom-gvisor
```

and writes the resulting `<repo>@sha256:<digest>` to `.ate-kind/ateom-image`.

Wiring:

- Called from the end of `deploy_ate_system` when `ATE_INSTALL_KIND=true`.
- Also exposed as `--publish-ateom-image` for standalone rebuilds.
- `.ate-kind/` added to `.gitignore`.

### DIFF
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -68,6 +68,7 @@ function usage() {
   echo "  --deploy-atelet                        Deploy atelet only"
   echo "  --deploy-ate-apiserver                 Deploy ate-api-server only"
   echo "  --deploy-atenet                        Deploy atenet only"
+  echo "  --publish-ateom-image                  Build + push ateom-gvisor (kind only); writes digest to .ate-kind/ateom-image"
   echo ""
   echo "To create individual resources used by ate-system (Note: These are"
   echo "called automatically by --deploy-ate-system):"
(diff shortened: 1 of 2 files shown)


hack/install-ate.sh
```diff
@@ -245,6 +246,32 @@ deploy_ate_system() {
...
+# Build + push the ateom-gvisor image and persist its digest under
+# .ate-kind/ateom-image. ateom-gvisor is the per-worker-pod sidecar
+# referenced via WorkerPool.spec.ateomImage; it has no Deployment
+# manifest under manifests/ate-install/, so the ko-resolve pipeline that
+# builds every other binary never touches it. Without this step,
+# operators creating a WorkerPool after a fresh --deploy-ate-system have
+# to invoke ko publish by hand.
+#
+# Only runs in kind mode; non-kind installs continue to manage their own
+# image-publishing flow against the cluster's registry.
```
BenTheElder, line 269: Don't we have the same problem there? We're using a registry for kind installs.

PR #68, opened 2026-05-23
fix(router): map gRPC errors to proper HTTP codes (#31)

### LINKED ISSUE #31: Router error messages are not great
```
$ curl -X POST -H "Host: ctr6.actors.resources.substrate.ate.dev" -i http://localhost:8888/
content-type: 
content-length: 96
date: Thu, 21 May 2026 02:40:40 GMT
server: envoy

error resuming actor ctr6: rpc error: code = FailedPrecondition desc = no free workers available

$ curl -X POST -H "Host: ctr7.actors.resources.substrate.ate.dev" -i http://localhost:8888/
HTTP/1.1 404 Not Found
content-type: 
content-length: 9
date: Thu, 21 May 2026 02:40:46 GMT
server: envoy

not found
```

mchmarny:
Fixes #31.

The Envoy `ext_proc` handler wrapped resume errors with `%w`, so the gRPC encoding leaked into the response body, and every non-NotFound failure returned 500. The 404 path returned `"not found"` without indicating what was missing.

This change adds a small mapping layer that translates gRPC status codes to HTTP status codes with short, client-safe bodies:

| gRPC code            | HTTP | Body                                       |
|----------------------|------|--------------------------------------------|
| `NotFound`           | 404  | `actor "<id>" not found`                   |
| `FailedPrecondition` | 503  | `actor "<id>" unavailable: <desc>` *       |
| `Unavailable`        | 503  | `actor "<id>" unavailable`                 |
| `DeadlineExceeded`   | 504  | `actor "<id>" request timed out`           |
| `PermissionDenied`   | 403  | `actor "<id>" access denied`               |
| `Unauthenticated`    | 401  | `actor "<id>" authentication required`     |
| `ResourceExhausted`  | 429  | `actor "<id>" rate limited`                |
| anything else        | 500  | `error resuming actor "<id>"`              |

\* `FailedPrecondition` is the only code whose gRPC `desc` is preserved verbatim; it carries actionable client-facing context like `"no free workers available"` and is not security-sensitive. Other codes deliberately drop the `desc` to avoid leaking server-side detail.

The original error is preserved via `Unwrap` so existing `slog.ErrorContext` logging keeps full fidelity.

Two smaller follow-ons made in the same change:

* `ActorResumer` no longer translates `codes.NotFound` into a router-layer sentinel. gRPC errors now propagate untouched so the HTTP boundary owns the mapping. The retry behavior is unchanged (the backoff already exits on any non-`Aborted` error).
* The "bad worker IP" branch (returned actor with an unparseable IP) now responds with an opaque `actor "<id>" routing failed` instead of the previous body that echoed back the invalid IP. The full IP is still logged server-side. This is deliberate; the IP is internal infrastructure detail and shouldn't appear in a client response.

### Before / after

Before:

```shell
error resuming actor ctr6: rpc error: code = FailedPrecondition desc = no free workers available
```

```shell
HTTP/1.1 404 Not Found
not found
```

After:

```shell
HTTP/1.1 503 Service Unavailable
actor "ctr6" unavailable: no free workers available
```

```shell
HTTP/1.1 404 Not Found
invalid host "ctr6.example.com": invalid actor_id: must end with actors.resources.substrate.ate.dev, got "ctr6.example.com"
```

### Test plan

- [x] `go test -race ./cmd/atenet/internal/app/router/...` — all pass
- [x] `go vet ./...` — clean
- [x] `make verify-fmt` — clean
- [x] `hack/verify-all.sh` (boilerplate, go-generate, gofmt, licenses, mod-tidy) — all pass
- [x] `make build-atenet` — succeeds
- New tests: 9-case table for `mapResumeError`, plus constructor and contract tests in `errors_test.go`; existing `extproc_test.go` cases expanded to cover the new gRPC code mappings end-to-end through `handleRequestHeaders`.

cc @thockin, is this what you had in mind in #31? PTAL

### DIFF
--- a/cmd/atenet/internal/app/router/errors.go
+++ b/cmd/atenet/internal/app/router/errors.go
@@ -24,7 +24,7 @@ import (
 
 // newReqError builds a reqError whose body is the formatted message and no
 // wrapped cause. Set the cause field directly when one is available.
-func newReqError(code envoy_type.StatusCode, format string, args ...any) *reqError {
+func newReqError(code envoy_type.StatusCode, format string, args ...any) error {
 	return &reqError{
 		msg:        fmt.Sprintf(format, args...),
 		statusCode: int(code),
@@ -32,13 +32,13 @@ func newReqError(code envoy_type.StatusCode, format string, args ...any) *reqErr
 }
 
 // actorNotFoundErr returns a 404 reqError identifying the missing actor.
-func actorNotFoundErr(actorID string) *reqError {
+func actorNotFoundErr(actorID string) error {
 	return newReqError(envoy_type.StatusCode_NotFound, "actor %q not found", actorID)
 }
 
 // invalidHostErr returns a 404 reqError explaining why the request host was
 // rejected. The cause is preserved for log inspection via Unwrap.
-func invalidHostErr(host string, cause error) *reqError {
+func invalidHostErr(host string, cause error) error {
 	return &reqError{
 		msg:        fmt.Sprintf("invalid host %q: %v", host, cause),
 		cause:      cause,
@@ -53,34 +53,40 @@ func invalidHostErr(host string, cause error) *reqError {
 //
 // Unrecognised errors collapse to 500 with a generic body to avoid leaking
 // server-side detail (stack traces, internal IDs) to clients.
-func mapResumeError(actorID string, err error) *reqError {
+func mapResumeError(actorID string, err error) error {
 	if err == nil {
 		return nil
 	}
 
-	var re *reqError
+	re := &reqError{cause: err}
 	switch status.Code(err) {
 	case codes.NotFound:
-		re = actorNotFoundErr(actorID)
+		re.statusCode = int(envoy_type.StatusCode_NotFound)
+		re.msg = fmt.Sprintf("actor %q not found", actorID)
 	case codes.FailedPrecondition:
 		// Preserve the gRPC description for FailedPrecondition only: it carries
 		// actionable client-facing context (e.g. "no free workers available")
 		// and is not security-sensitive.
-		re = newReqError(envoy_type.StatusCode_ServiceUnavailable,
-			"actor %q unavailable: %s", actorID, status.Convert(err).Message())
+		re.statusCode = int(envoy_type.StatusCode_ServiceUnavailable)
+		re.msg = fmt.Sprintf("actor %q unavailable: %s", actorID, status.Convert(err).Message())
 	case codes.Unavailable:
-		re = newReqError(envoy_type.StatusCode_ServiceUnavailable, "actor %q unavailable", actorID)
+		re.statusCode = int(envoy_type.StatusCode_ServiceUnavailable)
+		re.msg = fmt.Sprintf("actor %q unavailable", actorID)
 	case codes.DeadlineExceeded:
-		re = newReqError(envoy_type.StatusCode_GatewayTimeout, "actor %q request timed out", actorID)
+		re.statusCode = int(envoy_type.StatusCode_GatewayTimeout)
+		re.msg = fmt.Sprintf("actor %q request timed out", actorID)
 	case codes.PermissionDenied:
-		re = newReqError(envoy_type.StatusCode_Forbidden, "actor %q access denied", actorID)
+		re.statusCode = int(envoy_type.StatusCode_Forbidden)
+		re.msg = fmt.Sprintf("actor %q access denied", actorID)
 	case codes.Unauthenticated:
-		re = newReqError(envoy_type.StatusCode_Unauthorized, "actor %q authentication required", actorID)
+		re.statusCode = int(envoy_type.StatusCode_Unauthorized)
+		re.msg = fmt.Sprintf("actor %q authentication required", actorID)
 	case codes.ResourceExhausted:
-		re = newReqError(envoy_type.StatusCode_TooManyRequests, "actor %q rate limited", actorID)
+		re.statusCode = int(envoy_type.StatusCode_TooManyRequests)
+		re.msg = fmt.Sprintf("actor %q rate limited", actorID)
 	default:
-		re = newReqError(envoy_type.StatusCode_InternalServerError, "error resuming actor %q", actorID)
+		re.statusCode = int(envoy_type.StatusCode_InternalServerError)
+		re.msg = fmt.Sprintf("error resuming actor %q", actorID)
 	}
-	re.cause = err
 	return re
 }
--- a/cmd/atenet/internal/app/router/errors.go
+++ b/cmd/atenet/internal/app/router/errors.go
@@ -1,0 +1,86 @@
+//  Copyright 2026 Google LLC
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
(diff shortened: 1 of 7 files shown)


cmd/atenet/internal/app/router/errors.go
```diff
@@ -0,0 +1,91 @@
...
+	return &reqError{
+		msg:        fmt.Sprintf(format, args...),
+		statusCode: int(code),
+	}
+}
+
+// actorNotFoundErr returns a 404 reqError identifying the missing actor.
+func actorNotFoundErr(actorID string) *reqError {
+	return newReqError(envoy_type.StatusCode_NotFound, "actor %q not found", actorID)
+}
```
mchmarny, line 37: Self-review: `actorNotFoundErr` is defined and tested but never called from production code — the `codes.NotFound` branch in `clientMessageForResume` inlines `fmt.Sprintf("actor %q not found", ...)` instead of using it. Two ways out: (a) use this constructor in that branch (and drop the inline message), or (b) delete `actorNotFoundErr` and its test. Slight preference for (a) since it gives the constructor an actual callsite.

cmd/atenet/internal/app/router/errors.go
```diff
@@ -0,0 +1,91 @@
...
+
+	msg, code := clientMessageForResume(actorID, err)
+	return &reqError{
+		msg:        msg,
+		cause:      err,
+		statusCode: int(code),
+	}
+}
+
+func clientMessageForResume(actorID string, err error) (string, envoy_type.StatusCode) {
```
mchmarny: nit: `clientMessageForResume` has exactly one caller (`mapResumeError`) and no test of its own — the tests go through `mapResumeError`. Inlining the switch saves a layer without losing coverage. Happy either way.

cmd/atenet/internal/app/router/errors_test.go
```diff
@@ -0,0 +1,190 @@
...
+	t.Parallel()
+
+	const actorID = "ctr6"
+
+	tests := []struct {
+		name     string
+		err      error
+		wantCode envoy_type.StatusCode
+		wantBody string
+		wantWrap bool // whether errors.Is(result, original) should hold
```
mchmarny: nit: every case in the table sets `wantWrap: true`, so the field and the `tc.wantWrap &&` guard below add no signal. Can drop both and check `errors.Is` unconditionally.

cmd/atenet/internal/app/router/errors_test.go
```diff
@@ -0,0 +1,190 @@
...
+// existing handleRequestHeaders branch (errors.As(err, &reqErr)) keeps working.
+func TestMapResumeError_IsReqError(t *testing.T) {
+	t.Parallel()
+
+	err := mapResumeError("x", status.Error(codes.NotFound, "x"))
+	var reqErr *reqError
+	if !errors.As(err, &reqErr) {
+		t.Fatalf("errors.As(*reqError) = false, want true; err type = %T", err)
+	}
+	_ = fmt.Sprintf("%v", reqErr) // sanity: stringable
```
mchmarny: nit: `_ = fmt.Sprintf("%v", reqErr)` doesn't assert anything — `reqErr` is already known to be `*reqError` from the `errors.As` above, and `%v` on a non-nil pointer can't panic. Safe to drop, along with the `"fmt"` import once gone.

mchmarny (commented):
Self-review pass — one mild structural finding (`actorNotFoundErr` is unused production-side) and three test-code nits. None block merge; I'll wait for maintainer feedback before churning the PR.

cmd/atenet/internal/app/router/errors.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+
+	envoy_type "github.com/envoyproxy/go-control-plane/envoy/type/v3"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// newReqError builds a reqError whose body is the formatted message and no
+// wrapped cause. Set the cause field directly when one is available.
+func newReqError(code envoy_type.StatusCode, format string, args ...any) *reqError {
```
bowei: We should probably return `error` type vs returning the concrete type.

cmd/atenet/internal/app/router/errors.go
```diff
@@ -0,0 +1,86 @@
...
+	return &reqError{
+		msg:        fmt.Sprintf(format, args...),
+		statusCode: int(code),
+	}
+}
+
+// actorNotFoundErr returns a 404 reqError identifying the missing actor.
+func actorNotFoundErr(actorID string) *reqError {
```
bowei: Same comment above.

cmd/atenet/internal/app/router/errors.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// mapResumeError translates an ActorResumer error into a client-facing
+// reqError. It maps gRPC status codes to appropriate HTTP status codes and
+// short, human-readable bodies. The original error is preserved via Unwrap
+// so callers can still inspect it via errors.Is / errors.As when logging.
+//
+// Unrecognised errors collapse to 500 with a generic body to avoid leaking
+// server-side detail (stack traces, internal IDs) to clients.
+func mapResumeError(actorID string, err error) *reqError {
```
bowei: see above

bowei (commented):
Some nits. Thanks for the change!

cmd/atenet/internal/app/router/errors.go
mchmarny: Done in d91a4a0 — `newReqError` now returns `error`.

cmd/atenet/internal/app/router/errors.go
mchmarny: Done in d91a4a0 — `actorNotFoundErr` now returns `error`.

cmd/atenet/internal/app/router/errors.go
mchmarny: Done in d91a4a0 — `mapResumeError` (and `invalidHostErr`) now return `error`. Tests switched to `errors.As` to access the struct fields.

cmd/atenet/internal/app/router/errors.go
mchmarny: Resolved in 0b8af6e — `mapResumeError` now calls `actorNotFoundErr` for the `codes.NotFound` branch (option a).

cmd/atenet/internal/app/router/errors.go
mchmarny: Resolved in 0b8af6e — `clientMessageForResume` inlined into `mapResumeError`.

cmd/atenet/internal/app/router/errors_test.go
mchmarny: Resolved in 0b8af6e — `wantWrap` field and its guard removed; `errors.Is(got, tc.err)` is now checked unconditionally for every case.

cmd/atenet/internal/app/router/errors_test.go
mchmarny: Resolved in 0b8af6e — dead `fmt.Sprintf("%v", reqErr)` and `"fmt"` import dropped.

mchmarny:
> Some nits. Thanks for the change!

@bowei, PTAL
All 3 should be resolved with `d91a4a0`

bowei (approved):
looks good

PR #70, opened 2026-05-23
Add shellcheck and fix issues

thockin:
Not too many were found.  Let's keep it that way.

### DIFF
--- a/hack/install-ate-kind.sh
+++ b/hack/install-ate-kind.sh
@@ -1,4 +1,5 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
 # Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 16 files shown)


dims (approved):
LGTM

One question though: do we want to standardize in `#!/usr/bin/env bash` in all scripts?

thockin:
> One question though: do we want to standardize in #!/usr/bin/env bash in all scripts?

Probably.  I can add a commit here.

hack/install-ate-kind.sh
```diff
@@ -19,16 +20,19 @@ set -o pipefail
 
 ROOT=$(git rev-parse --show-toplevel)
 
+# shellcheck disable=SC2155 # safe initialization
```
BenTheElder, line 23: this shouldn't be failing SC2155? is this stale?
not going to block over this nit but.

Issue #23, opened 2026-05-20
Handle Pod Eviction/Deletion gracefully


juli4n:
Today Substrate doesn't have any special handling of worker pod deletions. If the worker pod assigned to a RUNNING actor is deleted, the system gets into a corrupted state where the control plane still thinks it's RUNNING. Moreover, we don't do any graceful suspension of the actor.

Additionally, the way we sync worker pods into the control plane Valkey is fragile and can drift.

This issue is to design a more robust handling of pod deletion, where we hook into the deletion process (via a finalizer) and perform a graceful worker pod deletion. This involves suspending the actor. It's unclear to me if it's needed to *also* resume it on a different worker pod, sort of a live actor migration, but I'll assume that just suspending is enough for now.

The model I have in mind is one where we externalize the registration and de-registration of worker pods to a dedicated controller running centrally inside `atecontroller`. I'll refer to this controller as the "wp-controller".

**Worker Pod Creation**

* The user creates or updates a `WorkerPool` custom resource.
* The `WorkerPool` controller creates a Kubernetes Deployment. Pods get a finalizer (e.g., `ate.gke.io/worker-cleanup`).
* The wp-controller observes the newly created Pod object and immediately calls `RegisterWorker(NOT_READY)` on `ate-apiserver`.

**Registration**

* Kubernetes schedules the worker pod, assigns it an IP address, and starts the `ateom` container.
* Kubelet periodically probes worker pod readiness. Once `ateom` completes boot and verifies internal runtime health, the probe passes and kubelet marks the Pod condition as `Ready = True`.
* The wp-controller observes the `Ready = True` Pod update and calls `MarkWorkerReady(IP)` on `ate-apiserver` to transition the worker to `READY`.

**Running**

* The worker is now in `FREE` state.
* When an actor needs to be resumed, the `ateapi` server picks this worker and assigns the actor to it.
* The worker handles requests for that actor until the actor is suspended.

**Deregistration / Deletion (Controller-Driven)**

* Kubernetes initiates a Pod deletion (e.g., scale down, node drain), setting `deletionTimestamp` on the Pod and sending `SIGTERM` to the `ateom` container (we will need to set a generous grace period on the worker pod).
* `ateom` intercepts `SIGTERM` and sheds incoming HTTP traffic by returning `503 Retry-After`.
* The wp-controller observes the `deletionTimestamp != nil` update on the Pod and drives the graceful suspension:
[...]

PR #75, opened 2026-05-24
ateapi/syncer: release actor when host pod is deleted

dims:
Today, if a worker pod that hosts an active actor is forcibly destroyed, substrate does not migrate the actor. The actor's AteomPodName still points at the dead pod and the router times out forwarding to a dead IP.

WorkerPoolSyncer's pod-informer DeleteFunc and soft-delete branch now check whether the worker being removed is bound to an actor, and if so, reset that actor to `STATUS_SUSPENDED` before the worker row goes away.

The new helper releaseActorOnDeadWorker reads the actor, only acts if the actor still claims this worker (a concurrent SuspendActor or DeleteActor that already advanced state is respected), clears the pod-binding fields and InProgressSnapshot, preserves LastSnapshot, and writes via version-checked UpdateActor. On ErrPersistenceRetry we drop the attempt and let the next informer event retry — no separate lock, no in-handler retry budget.

`STATUS_SUSPENDED` is reused rather than introducing a new state. The post-orphan invariant is identical and reusing keeps the proto, printer, dialer, and router contracts unchanged. The next request through atenet triggers an implicit resume; findFreeWorker picks any free worker in the pool; LastSnapshot is restored.

Test plan: `TestSyncer_DeleteBoundWorker_ClearsActor` covers RUNNING actor → pod delete → SUSPENDED with cleared bind fields, InProgressSnapshot dropped, LastSnapshot preserved. All existing tests in `cmd/ateapi/internal/controlapi/` still pass.


cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -136,3 +142,41 @@ func (s *WorkerPoolSyncer) syncWorkerToStore(ctx context.Context, pod *corev1.Po
 func isWorkerEligible(pod *corev1.Pod) bool {
 	return pod.Status.PodIP != ""
 }
+
+// releaseActorOnDeadWorker resets the actor bound to a vanishing worker pod
+// back to STATUS_SUSPENDED so the next request reassigns it. Optimistic
+// version on UpdateActor protects against a concurrent SuspendActor /
+// ResumeActor; on contention we drop this attempt and rely on the next
```
juli4n: The caller unconditionally deletes the worker so next releaseActorOnDeadWorker will just short-circuit because the worker is not found, right?

I think maybe we can just mention that this is best effort for now (until we do https://github.com/agent-substrate/substrate/issues/23)? This PR makes things better, but I don't think we can make it handle all failure modes until we have finalizers + controller.

We can maybe also add a link to https://github.com/agent-substrate/substrate/issues/23 from here.

dims: Yes. Ack. Let me add comments to reflect best effort and link to 23 as well.

PR #77, opened 2026-05-26
ci: also run pr-workflow on push to main

dims:
The existing trigger is pull_request only, so when a PR merges no CI runs against the resulting main-branch commit. A bad squash, a post-rebase test failure, or a flake-masked regression isn't surfaced until the next PR opens, at which point bisecting "is it my change or main?" wastes contributor time.

Add a push trigger restricted to main so the same run-tests and e2e-test jobs run against the merged commit. No new jobs, no new permissions, no change to PR behavior.

- [ ] Tests pass

### DIFF
--- a/.github/workflows/pr-workflow.yaml
+++ b/.github/workflows/pr-workflow.yaml
@@ -15,6 +15,8 @@
 name: pr-workflow
 on:
   pull_request:
+  push:
+    branches: [main]
 jobs:
   run-tests:
     runs-on: ubuntu-latest


BenTheElder (approved):
Will need a manual rebase, but LGTM

dims:
@BenTheElder rebased!

BenTheElder (approved):

BenTheElder:
Hmm, I like having the result in main, but I think we're effectively requiring this before merge currently. 

It seems the merge rules are requiring rebase before merge after any other PR merges, unless I'm missing something.

cc @dberkov @thockin 

This might be redundant now.

BenTheElder:
Followed up with @thockin, the merge requirement was relaxed.

PR #80, opened 2026-05-26
ci: add dependabot config

dims:
Weekly checks for four ecosystems present in the repo: gomod (grouped into a single PR to keep review noise low), github-actions, and the two Dockerfiles under benchmarking/locust and demos/claude-code-multiplex/workload.

- [ ] Tests pass

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,24 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


BenTheElder:
So we actually removed dependabot config before the initial public release because it interacted very poorly with the repo and the PRs generally failed presubmits.

- doesn't go-mod-tidy tools modules
- doesn't update LICENSES for the root module

For the locust benchmark utils, I don't think we're ready to auto-update them until we have CI coverage to see if the change is safe.

PR #82, opened 2026-05-26
Fix/version stamping

dims:
Adds a pkg/version package and wires -ldflags -X through the Makefile + hack/install-ate.sh. All six binaries (kubectl-ate, atenet, ateapi, atelet, ateom-gvisor, podcertcontroller) now respond to --version and print something!

Please share what you think ... exactly we should print :) 

- [ ] Tests pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,16 @@ KO := hack/run-tool.sh ko
 BINDIR := bin/
 ATECTL := $(BINDIR)/kubectl-ate
 
+# Version stamping. Override on the make command line to pin
+# (e.g. `make VERSION=v0.5.0 build`).
+VERSION    ?= $(shell git describe --tags --always --dirty 2>/dev/null || echo dev)
+COMMIT     ?= $(shell git rev-parse HEAD 2>/dev/null || echo unknown)
+BUILD_DATE ?= $(shell date -u +%Y-%m-%dT%H:%M:%SZ)
+VERSION_PKG := github.com/agent-substrate/substrate/internal/version
+LDFLAGS := -X $(VERSION_PKG).Version=$(VERSION) \
+           -X $(VERSION_PKG).Commit=$(COMMIT) \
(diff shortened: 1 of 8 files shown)


internal/version/version.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2026 Google LLC
```
juli4n, line 1: This package is internal to the substrate project so it should live under internal/ instead of pkg. Could you move it there?

dims: on it

BenTheElder:
> Adds a pkg/version package and wires -ldflags -X through the Makefile + hack/install-ate.sh. All six binaries (kubectl-ate, atenet, ateapi, atelet, ateom-gvisor, podcertcontroller) now respond to --version and print something!

For a greenfield effort, did we consider https://pkg.go.dev/runtime/debug#BuildInfo instead of depending on ldflags?

thockin:
Edit - it does use BuildInfo *and* ldflags

thockin:
@dims rebase and I am LGTM on it

dims:
@thockin Done! (rebased)

Makefile
```diff
@@ -34,22 +44,22 @@ build: build-images build-atectl
...
 .PHONY: build-images
 build-images:
-	$(KO) build ./cmd/ateapi
-	$(KO) build ./cmd/atelet
-	$(KO) build ./cmd/podcertcontroller
-	$(KO) build ./cmd/atenet
+	$(KO) build --ldflags "$(LDFLAGS)" ./cmd/ateapi
+	$(KO) build --ldflags "$(LDFLAGS)" ./cmd/atelet
+	$(KO) build --ldflags "$(LDFLAGS)" ./cmd/podcertcontroller
+	$(KO) build --ldflags "$(LDFLAGS)" ./cmd/atenet
```
BenTheElder, line 50: wouldn't we also need this independently computed for hack/install-ate.sh?

PR #84, opened 2026-05-26
chore: enhance .gitignore with common patterns

ameukam:

- [x] Tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,34 @@ __pycache__/
 
 # Local environment variables
 .ate-dev-env.sh
+
+# OS artifacts
+.DS_Store
+Thumbs.db
+
+# Editors and IDEs
+.idea/
+.vscode/
+*.swp
+*.swo
+*~
+
+# AI IDEs and CLIs
+.aider*
+.claude/
+.cline/
+.codex/
+.continue/
+.copilot/
+.cursor/
+.gemini/
+.goose/
+.kiro/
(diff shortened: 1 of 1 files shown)


ahmedtd (dismissed):

.gitignore
```diff
@@ -5,11 +5,50 @@
 /atelet
 /bin/
 
+# Build outputs (SBOMs, image refs)
+/_output/
+
+# Go test and coverage
+*.test
+*.out
+coverage.*
```
BenTheElder: nothing should be writing these to the tree

.gitignore
```diff
@@ -5,11 +5,50 @@
 /atelet
 /bin/
 
+# Build outputs (SBOMs, image refs)
+/_output/
```
BenTheElder: we use bin/, are we using _output anywhere?

ameukam: I don't think so. Removed

.gitignore
ameukam: copied from https://github.com/github/gitignore/blob/main/Go.gitignore.
I'll remove them

BenTheElder: Let's do that, and we can instruct agents to write binaries under bin/

We don't want surprises and we have already reserved bin/ and a few other binary output paths

ameukam: Done.

BenTheElder (approved):

PR #91, opened 2026-05-27
fix(memorypullcache): surface pull failures with the image ref

azr:
## Symptom

When `memorypullcache.Fetch` fails (network, auth, 404, ...), the log shows `Cache miss ref=...` followed by silence. The eventual user-visible error comes from a higher layer with no ref attached, so the actual offending image is invisible in operator logs.

## Cause

`remote.Image`, `img.Size`, and `io.ReadAll` errors were wrapped as `in remote.Image: %w` (etc.) and returned without ever being logged or naming the ref.

## Fix

Wrap each error with the ref and log it at WARN. Same shape as the existing surfacing in #40.


azr:
Closing: no longer pursuing this; substrate is ruled out for the sandbox work this came from. Happy to reopen if the error-surfacing improvement is still wanted.

PR #92, opened 2026-05-27
fix(ateompath): validate ateom socket path length

azr:
## Symptom

When `<namespace>:<pod-name>` exceeds 107 bytes, `ateom-gvisor` exits with `bind: invalid argument` and the controller surfaces a generic `DeadlineExceeded`.

## Cause

Linux limits unix-socket paths to 107 bytes (`sockaddr_un.sun_path`). A pod's ReplicaSet hash can jiggle between 8 and 10 chars across rollouts, so a demo that fits at 8 silently breaks at 10 with no clear error.

## Fix

Add `ValidateAteomSocketPath` (with unit tests) and call it from `ateom-gvisor` before the bind, so the operator sees the real cause in the pod log instead of a kernel EINVAL.

### DIFF
--- a/internal/ateompath/ateompath.go
+++ b/internal/ateompath/ateompath.go
@@ -16,8 +16,14 @@
 package ateompath
 
 import (
+	"fmt"
 	"path/filepath"
 )
+
+// MaxUnixSocketPathLen is the practical Linux limit for unix-domain socket
+// paths. The kernel's sockaddr_un.sun_path is 108 bytes including the trailing
+// NUL, leaving 107 usable bytes. Bind fails with EINVAL above this.
+const MaxUnixSocketPathLen = 107
 
 const (
 	// The base path.  This is both the path of the root shared folder on the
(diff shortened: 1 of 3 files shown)


internal/ateompath/ateompath.go
```diff
@@ -49,6 +55,22 @@ func AteomSocketPath(ateomNamespace, ateomName string) string {
...
 
+// ValidateAteomSocketPath returns a descriptive error when the socket path
+// derived from ateomNamespace and ateomName would exceed Linux's unix-socket
+// limit. Calling net.Listen("unix", ...) with an over-limit path otherwise
+// fails with the cryptic "bind: invalid argument".
+func ValidateAteomSocketPath(ateomNamespace, ateomName string) error {
+	p := AteomSocketPath(ateomNamespace, ateomName)
+	if len(p) > MaxUnixSocketPathLen {
+		return fmt.Errorf(
+			"ateom socket path %q is %d bytes, exceeds Linux unix-socket limit of %d: shorten the namespace or pod name (%d + %d = %d chars used for namespace + name)",
```
BenTheElder, line 66: long-term I think we should probably fix this instead of suggesting the user work around it, perhaps using a cheap hash like fnv instead of the actual namespace + pod name. cc @ahmedtd

BenTheElder (approved):

PR #96, opened 2026-05-27
[WIP] feat: GPU passthrough for actors via gVisor nvproxy + cuda-checkpoint

dims:
## Summary

End-to-end GPU passthrough for substrate actors via gVisor's nvproxy + cuda-checkpoint path.

## Changes

- **CRD** — `pkg/api/v1alpha1/actortemplate_types.go::Container` grows an optional `Resources *ContainerResources` carrying `GPU *GPUResource{Count, Device, DriverCapabilities, DriverVersion}`. `zz_generated.deepcopy.go` and `manifests/.../ate.dev_actortemplates.yaml` regenerated via `go generate ./...`.

- **Protos** — `internal/proto/ateletpb` and `internal/proto/ateompb` `Container` messages gain a `GpuSpec gpu` field mirroring the CRD shape.

- **ateapi → atelet** — `cmd/ateapi/internal/controlapi/gpu.go` adds `toAteletGpuSpec(*v1alpha1.ContainerResources) *ateletpb.GpuSpec`; the resume and suspend workflows populate it on each `ateletpb.Container`.

- **atelet → ateom** — `cmd/atelet/main.go` projects `ateletpb.GpuSpec → ateompb.GpuSpec` via `toAteomGpuSpec`. `cmd/atelet/oci.go::prepareOCIDirectory` gains a `gpu *ateletpb.GpuSpec` parameter and `addGPUToOCISpec()` helper that injects `/dev/nvidia*` device nodes (host major/minor) into `Linux.Devices` and bind-mounts `/usr/local/bin/cuda-checkpoint` + the wrapper script when the workload requests GPU. Pause containers pass `nil`.

- **ateom-gvisor** — `cmd/ateom-gvisor/runsc.go` gains a `gpu *ateompb.GpuSpec` field on the `runsc` struct (populated via `firstGPUSpec` from the workload's containers). `gpuGlobalFlags()` emits `--nvproxy [--nvproxy-driver-version=X] [--nvproxy-allowed-driver-capabilities=...]` on `runsc create`/`checkpoint`/`restore`. `gpuSaveRestoreFlags()` is gated to the **root container only** (the supervisor sub-container restore must not re-invoke the wrapper, and gVisor's nvproxy auto-registers cuda-checkpoint internally on `release-20260520.0`).

- **Wrapper** — `hack/cuda-checkpoint-wrapper.sh` (20 lines). Idempotent `cuda-checkpoint --toggle` over every CUDA-touching PID found in `/proc/*/maps`. Only skips `$$` (self) — **not** PID 1, because inside a substrate sandbox the workload is PID 1. Used by the bare-metal `validate-bare.sh` in the demo dir; substrate proper leans on gVisor's internal cuda-checkpoint registration.

- **Tests** — `cmd/ateom-gvisor/runsc_test.go` (linux build tag) covers `gpuGlobalFlags`, `gpuSaveRestoreFlags`, `firstGPUSpec`. The duration-string regression (`30s`, not `30000`) is asserted explicitly. `go test ./pkg/api/v1alpha1/...` clean.

## Constraints (from the gVisor source crawl)

- NVIDIA driver must be R570+ (cuda-checkpoint NVML support) AND appear in `runsc nvproxy list-supported-drivers`. release-20260520.0 supports 16 versions across 535/550/570/580/590.
- Driver version must match across checkpoint and restore — migration across driver upgrades is rejected.
- x86_64 only. cuda-checkpoint isn't supported on arm64.
- Operational gotcha (documented in the demo README): on hosts with `nvidia-persistenced` running, replace `/run/nvidia-persistenced/socket` with a regular file — gVisor's gofer can't bind-mount Unix sockets, and `nvidia-container-cli` hard-codes that mount.

## Test plan

- [x] `go vet ./pkg/... ./internal/proto/... ./cmd/atelet/... ./cmd/ateapi/...`
- [x] `GOOS=linux go build ./cmd/atelet ./cmd/ateapi ./cmd/ateom-gvisor`
- [x] `go test ./pkg/api/v1alpha1/...`


BenTheElder (commented):
If this is only nvidia GPUs, we should probably name it appropriately instead of "GPU"

What about other CDI Devices? TPUs? At the very least we probably want to leave API shape for this.

hack/cuda-checkpoint-wrapper.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
BenTheElder, line 1: Currently I don't think we're shipping any other `hack/` script to prod. We should probably move this (or replace it with a binary)

elezar (commented):
I would have a similar question to @BenTheElder.

Why do we not use CDI devices as the general mechanism and have NVIDIA GPUs be one example of that?

PR #97, opened 2026-05-27
Add govulncheck CI workflow

ameukam:


ameukam:
cc @BenTheElder @juli4n

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,30 @@
...
+  contents: read
+
+jobs:
+  govulncheck:
+    runs-on: ubuntu-latest
+    steps:
+      - id: govulncheck
+        uses: golang/govulncheck-action@v1
+        with:
+          go-version-input: '1.26'
```
jbpratt:
```suggestion
        with:
          go-version-file: 'go.mod'
```
👀

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,30 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+name: govulncheck
+on:
+  pull_request:
```
BenTheElder: I think it might be more useful to run periodically on the merged code, since vulns can appear even with no pull request, and probably aren't related to the pull request?

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  govulncheck:
+    runs-on: ubuntu-latest
+    steps:
+      - id: govulncheck
+        uses: golang/govulncheck-action@v1
```
BenTheElder, line 31: does this do symbol level by default, or only package level?

.github/workflows/govulncheck.yaml
ameukam: TIL!
Fixed

.github/workflows/govulncheck.yaml
ameukam: Added a cron.

.github/workflows/govulncheck.yaml
ameukam: I believe it does symbol level by default.

.github/workflows/govulncheck.yaml
BenTheElder: I think we should drop the PR flow as well, so it doesn't block PRs while we remediate.

.github/workflows/govulncheck.yaml
BenTheElder: OK, I'd like to make sure it is symbol level to cut down noise.

BenTheElder: AIUI in Kubernetes it is only running at the package level.

.github/workflows/govulncheck.yaml
ameukam: Done

.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,31 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+name: govulncheck
+on:
+  schedule:
+    - cron: "37 4 * * 1"
```
BenTheElder, line 21: let's also run on push to main

.github/workflows/govulncheck.yaml
ameukam: Govulncheck does symbol-level analysis by default, it traces call graphs and only reports vulns in functions the code actually calls. From the [docs](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck): "It uses static analysis of source code or a binary's symbol table to narrow down reports to only those that could affect the application."

The GHA is running with the default behavior AIUI.

PR #100, opened 2026-05-28
ateom: avoid unix socket path length limits

BenTheElder:
Pod names alone can easily exceed the max length (253 > 107).

Add a symlink with the full pod name for debugging.

Follow-up #92 

- [x] Tests pass

---

TODO: I'm not sure about the symlink. It's annoying mapping pod => hash(pod) otherwise though. We could put a file inside the directory with the pod name instead, but I think we should probably put the full pod name _somewhere_ in the structure so it's easier to inspect these.

### DIFF
--- a/internal/ateompath/ateompath.go
+++ b/internal/ateompath/ateompath.go
@@ -16,14 +16,8 @@
 package ateompath
 
 import (
-	"fmt"
 	"path/filepath"
 )
-
-// MaxUnixSocketPathLen is the practical Linux limit for unix-domain socket
-// paths. The kernel's sockaddr_un.sun_path is 108 bytes including the trailing
-// NUL, leaving 107 usable bytes. Bind fails with EINVAL above this.
-const MaxUnixSocketPathLen = 107
 
 const (
 	// The base path.  This is both the path of the root shared folder on the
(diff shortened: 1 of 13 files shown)


BenTheElder:
> If it's a concern, cheapest fix is os.Remove(debugAteomDir) (ignoring os.IsNotExist) before the Symlink — symmetric with the RemoveAll(sockPath) pattern a few lines down.

Agreed. Done.

This is a host mount, which you can discover from the install manifests.

internal/ateompath/ateompath.go
```diff
@@ -41,6 +37,33 @@ func RunSCBinaryPath(sha256 string) string {
 }
 
 func AteomPath(ateomNamespace, ateomName string) string {
```
bowei: It would be good to put an example of the what this name looks like (full path).

internal/ateompath/ateompath.go
```diff
@@ -41,6 +37,33 @@ func RunSCBinaryPath(sha256 string) string {
...
+func hashPodName(name string) string {
+	hasher := fnv.New64a()
+	hasher.Write([]byte(name))
+	return base64.RawURLEncoding.EncodeToString(hasher.Sum(nil))
+}
+
+// DebugFriendlyAteomPath returns the path to the ateom directory using the pod name
+// and namespace (without hashing), this is useful for debugging.
+// Ateom will create a symlink at this path pointing to AteomPath.
+func DebugFriendlyAteomPath(ateomNamespace, ateomName string) string {
```
bowei: Suggest that you name this SymlinkAteomPath, as this is the specific use of the name. The fact that it is DebugFriendly is one of the requirements...

BenTheElder: The symlink path is only meant for debug and the other path should be the primary path. I was thinking we probably don't want callers to depend on this path for much. Maybe not.

BenTheElder:
Technically we could also solve this with Chdir and using a short relative path but I went with this approach because we'd have to be careful to lock to an OS thread without CLONE_FS and isolate it in atelet or else race on process-wide chdir.

Thinking about it more, maybe we want to make the symlink the shortened path and use that just for socket binding.

internal/ateompath/ateompath.go
ahmedtd: Can we use the pod UID for the ateom path?  That will always fit in the correct length, and is reasonably discoverable.

BenTheElder: I like it. Will update.

BenTheElder:
EDIT: Squashed down.

The diff is a bit more involved ... but I think this is still better architecturally.

ahmedtd (approved):

PR #101, opened 2026-05-28
fix(atelet): prevent path traversal in OCI tar extraction

konippi:
Resolves the three TODO comments in `cmd/atelet/oci.go` that called for a constrained filesystem to prevent path traversal, symlink escape, and hardlink escape during OCI tar extraction.

Uses `os.Root` (Go 1.24+) to confine all file operations to the rootfs directory. Adds `validateTarName()` as defence in depth.

- [x] Tests pass

### DIFF
--- a/cmd/atelet/oci.go
+++ b/cmd/atelet/oci.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/agent-substrate/substrate/internal/ateompath"
 	"github.com/agent-substrate/substrate/internal/memorypullcache"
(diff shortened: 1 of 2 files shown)


konippi:
e2e-test failure is unrelated to this change — `TestDemo3` timed out waiting for `ResumeGoldenActor` (gVisor restore) after 90s. The atelet image built and rolled out successfully; `run-tests` (unit tests + verify) passed. This looks like a flaky timeout in the kind CI environment. Happy to rebase and re-run if needed.

BenTheElder:
Hmm, I haven't seen this flake yet and we've tested quite a few PRs in this environment. Will rerun it.

BenTheElder:
These tests are also possible to run locally.

BenTheElder:
The repeated e2e seems likely to be related, flakes have not been observed with kind on other PRs and it has failed twice in a row here. You can test the counter demo locally by following the README.

konippi:
@BenTheElder 
Thanks for the re-run and the nudge to test locally — I was able to reproduce and found the root cause.

`mutate.Extract` produces entries with absolute paths (e.g. `/ko-app/counter`), and my `validateTarName` was rejecting them via `filepath.IsLocal`. The old code handled this implicitly through `filepath.Join`. Fixed by stripping the leading `/` before validation. Verified locally on kind.

BenTheElder:
The other PR is ... "WIP" ... this is not a concern.

BenTheElder:
I checked this out locally for review, it looks like we need to handle "last entry wins" still for symlinks being replaced by real files and vice versa, wrote a follow-up commit at:

https://github.com/agent-substrate/substrate/compare/main...BenTheElder:substrate:tar-fix?expand=1

BenTheElder:
That appears to be a pre-existing issue, so I think we can handle it as a follow-up.

PR #110, opened 2026-05-28
Replace actor eth0 move with veth networking

### LINKED ISSUE #122: Replace actor eth0 move with worker-owned veth networking
Today `ateom-gvisor` gives an active actor network connectivity by moving the worker pod's real Kubernetes `eth0` into the actor/gVisor network namespace. That works for basic actor execution, but it has an important side effect: while the actor is running, the worker pod no longer owns its pod network interface.

This makes the worker pod partially disappear from the normal Kubernetes networking model during actor activation. It also prevents us from cleanly adding pod-local networking components, such as an egress proxy or policy enforcement sidecar, because those components remain in the worker pod network namespace while actor traffic leaves through an interface that has been moved elsewhere.

We should replace the `eth0` move approach with an explicit veth pair between the worker pod namespace and the actor/gVisor namespace.

## Rationale

Keeping the pod's real `eth0` in the worker namespace gives us a stable networking boundary:

* The worker pod keeps normal Kubernetes network connectivity while an actor is running.
* Sidecars or worker-local helper processes can remain reachable and can participate in actor networking.
* Actor networking becomes explicit and owned by Substrate instead of depending on moving the CNI-provided interface.
* This creates the foundation for transparent egress capture and policy enforcement.
* It avoids coupling actor lifecycle to destructive mutation of the pod's primary network interface.

This is especially important for planned transparent egress policy work. We want actor traffic to be captured without requiring application opt-in via `HTTP_PROXY` or `HTTPS_PROXY`. To do that cleanly, actor traffic needs to cross a worker-owned interface where Substrate can install forwarding, NAT, and later proxy-capture rules.

## Proposed approach

For each active actor, `ateom-gvisor` should create a point-to-point veth pair:

* Worker pod namespace side: `ateom0`, for example `10.200.0.1/30`.
* Actor/gVisor namespace side: renamed to `eth0`, for example `10.200.0.2/30`.
* Actor default route points to the worker-side veth IP.
* The Kubernetes-provided pod `eth0` remains in the worker pod namespace.

For compatibility with the current routing model, the initial implementation can install temporary nftables rules:

* Masquerade actor egress behind the worker pod IP.
* DNAT inbound traffic from the worker pod IP and actor service port to the actor veth IP.
* Enable forwarding between the actor veth and pod `eth0`.

These rules are intended as a compatibility bridge, not the final egress policy implementation. The later egress work should replace broad actor egress NAT with transparent capture into AgentGateway and default-deny policy rules.

## Expected outcome

After this change:

* Actors still start, checkpoint, restore, and receive inbound traffic as before.
* The worker pod retains network connectivity during actor execution.
* `ateom-gvisor` no longer needs to scrape, move, or restore the pod's real `eth0`.
* Substrate has a cleaner networking foundation for transparent egress capture, policy enforcement, and future worker-local AgentGateway integration.

## Validation

The implementation should be validated with:

* Existing focused Go tests for `ateom-gvisor` and related networking/server boot packages.
* The demo e2e flow that exercises actor startup, golden snapshot creation, restore, and inbound routing.
* A runtime check that the worker pod still has its Kubernetes `eth0` while an actor is active.
* A runtime check that actor traffic routes through the veth pair and inbound traffic still reaches the actor.

EItanya:
Fixes #122

## Summary

This replaces the `ateom-gvisor` networking path that moved the worker pod's Kubernetes-provided `eth0` into the actor/gVisor network namespace.

Instead, the worker pod keeps its real `eth0`, and `ateom-gvisor` creates a point-to-point veth pair between the worker pod namespace and the actor namespace. The actor-side peer is renamed to `eth0`, receives the actor-side address, and uses the worker-side veth as its default gateway.

The PR also adds temporary nftables compatibility rules so existing inbound and outbound behavior continues to work while preserving the worker pod's own network connectivity.

## Why

Moving the pod's real `eth0` makes the worker pod lose normal Kubernetes network connectivity while an actor is active. That blocks pod-local networking components, including the planned transparent egress capture and AgentGateway integration, because those components remain in the worker pod namespace while actor traffic leaves through an interface that was moved elsewhere.

Keeping `eth0` in the worker pod namespace gives Substrate a stable worker-owned networking boundary for future transparent egress policy enforcement.

## Validation

* [x] `go test ./cmd/ateom-gvisor ./cmd/atelet ./internal/ateompath ./internal/controllers`
* [x] `go test ./cmd/ateom-gvisor ./internal/serverboot`
* [x] `NO_DEV_ENV=true BUCKET_NAME=ate-snapshots KO_DOCKER_REPO=localhost:5001 KUBECTL_CONTEXT=kind-kind ./hack/run-e2e.sh ./internal/e2e/suites/demo -run TestDemo3 -count=1`

## Checklist

* [x] Tests pass
* [x] Appropriate changes to documentation are included in the PR


cmd/ateom-gvisor/main.go
```diff
@@ -485,6 +354,395 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+	forward := c.AddChain(&nftables.Chain{
+		Name:     "forward",
+		Table:    table,
+		Type:     nftables.ChainTypeFilter,
+		Hooknum:  nftables.ChainHookForward,
+		Priority: nftables.ChainPriorityFilter,
+		Policy:   &acceptPolicy,
+	})
+
+	c.AddRule(&nftables.Rule{
```
ahmedtd: Nit: can you put the programs for each chain right below the corresponding chain definition?

cmd/ateom-gvisor/main.go
```diff
@@ -485,6 +354,395 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+		Priority: nftables.ChainPriorityNATSource,
+	})
+	acceptPolicy := nftables.ChainPolicyAccept
+	forward := c.AddChain(&nftables.Chain{
+		Name:     "forward",
+		Table:    table,
+		Type:     nftables.ChainTypeFilter,
+		Hooknum:  nftables.ChainHookForward,
+		Priority: nftables.ChainPriorityFilter,
+		Policy:   &acceptPolicy,
```
ahmedtd: Nit: You can use ptr.To: https://pkg.go.dev/k8s.io/utils/ptr#To

cmd/ateom-gvisor/main.go
```diff
@@ -485,6 +354,395 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+	})
+
+	c.AddRule(&nftables.Rule{
+		Table: table,
+		Chain: postrouting,
+		Exprs: append(ipSourceEqual(actorVethIP), &expr.Masq{}),
+	})
+	preroutingExprs := append(ipDestinationEqual(podIP.String()), tcpDestinationPortEqual(80)...)
```
ahmedtd, line 641: Can you leave a few TODOs here:

* We need to handle inbound UDP as well (actors may run QUIC servers).
* We need to handle multiple, configurable inbound ports.  (I think maybe that will require multiple rules on the prerouting chain?  Or maybe a more complicated rule that tests against multiple ports).

cmd/ateom-gvisor/main.go
```diff
@@ -485,6 +354,395 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+
+	actorLink, err := netlink.LinkByName(actorVethTempName)
+	if err != nil {
+		return fmt.Errorf("while getting actor veth peer: %w", err)
+	}
+	if err := netlink.LinkSetNsFd(actorLink, int(s.interiorNetNS)); err != nil {
+		return fmt.Errorf("while moving actor veth peer into interior netns: %w", err)
+	}
+
+	if err := netNSDo(ctx, s.interiorNetNS, configureActorVeth); err != nil {
```
ahmedtd, line 427: This is really the only part that needs to be done on run / restore, I think, since gVisor wipes the routes from the links in the interior namespace.  Everything else could be done on startup, at the same time we create the namespace.

(Except, eventually, I guess we will have per-actor egress rules we also program into nftables, so maybe wiping everything on each run makes more sense)

cmd/ateom-gvisor/main.go
```diff
@@ -50,6 +53,18 @@ var (
...
 
+const (
+	hostVethName      = "ateom0"
+	actorVethName     = "eth0"
+	actorVethTempName = "ateom1"
+	hostVethCIDR      = "10.200.0.1/30"
+	actorVethCIDR     = "10.200.0.2/30"
+	actorVethGateway  = "10.200.0.1"
+	actorVethIP       = "10.200.0.2"
+	defaultActorPort  = "80"
```
ahmedtd: This is unused.

ahmedtd (dismissed):

cmd/ateom-gvisor/main.go
```diff
@@ -485,6 +354,395 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+	// ateom0 as the gateway, and moves only the veth peer into the actor netns.
+	// The actor side renames that peer to eth0 and installs a default route via
+	// the worker-side veth address. This replaces the old behavior of moving the
+	// Kubernetes-provided eth0 out of the worker pod.
+	//
+	// The nftables rules installed here are a compatibility bridge for the
+	// current router assumptions: actor egress is masqueraded behind the worker
+	// pod IP, and inbound traffic to the worker pod's HTTP port is DNAT'd to the
+	// actor veth IP. Later transparent egress capture will replace the broad
+	// egress NAT with AgentGateway-bound capture rules.
```
BenTheElder: ... Have we decided to use AgentGateway? cc @bowei 

AFAICT this is the first reference in this project.

cmd/ateom-gvisor/main.go
BenTheElder: We could instead DNAT the IP without filtering on ports and avoid the port config, since we don't need network traffic except for the actor currently. 

EDIT: Spoke with Taahir, let's not :-)

internal/serverboot/serverboot.go
```diff
@@ -51,10 +51,6 @@ type TracingOptions struct {
 	// Sampler is required. ateapi typically uses ParentBased(AlwaysSample);
 	// atelet/ateom-gvisor use ParentBased(NeverSample).
 	Sampler sdktrace.Sampler
-	// NoExporter skips the OTLP exporter. Used by binaries with no
-	// network egress (ateom-gvisor, after eth0 moves into the gvisor
-	// netns).
-	NoExporter bool
```
BenTheElder, line 81: I think we should consider proxying otel through atelet if we need it for aetom. 

aetom-gvisor itself not having network traffic is a nice property.

cmd/ateom-gvisor/main.go
EItanya: This was a mistaken comment, I will fix. My agent decided to hallucinate a bit based on previous stuff I was doing.

internal/serverboot/serverboot.go
ahmedtd: I think we have been stockholming ourselves into thinking it's a nice property :P

Certainly, at least for me, I wanted ateom to be contactable over standard gRPC from ate-api-server.  I only gave it up because I wasn't confident in being able to figure out veth + nftables before our release deadline.

As we potentially support more different types of ateoms (kata, etc), I think it might be likely that we want ateom to handle aspects of image pulls, etc.  We will obviously still need to share the pull cache node-wide, using the shared /run/ateom-gvisor directory.

BenTheElder: [We chatted, Agreed]

cmd/ateom-gvisor/main.go
EItanya: Agreed, but this also depends on the outcome of #123 and how the different `ateom-*` conversation plays out. Either way I think it's ok for now

BenTheElder (commented):
needs a `make update` or the LICENSES script

thanks for working on this :-)

EItanya:
Up to date with `main` and new vendoring strategy, please review again

cmd/ateom-gvisor/main.go
```diff
@@ -485,6 +353,399 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+	//
+	// This function is intentionally idempotent so it can run before setup, after
+	// checkpoint, and from setup failure cleanup without requiring the caller to
+	// know how far network initialization progressed.
+	if err := removeActorNftablesRules(); err != nil {
+		return err
+	}
+
+	if link, err := netlink.LinkByName(hostVethName); err == nil {
+		if err := netlink.LinkDel(link); err != nil {
```
nybidari, line 521: If the host side veth clean up fails, we do not cleanup the actor's veth and return. I think we should log a warning and still go ahead and cleanup the interior network ns.

cmd/ateom-gvisor/main.go
```diff
@@ -485,6 +353,399 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+	// default-deny rules.
+	if err := removeActorNftablesRules(); err != nil {
+		return err
+	}
+
+	c := &nftables.Conn{}
+	table := &nftables.Table{
+		Family: nftables.TableFamilyIPv4,
+		Name:   actorNftTableName,
+	}
```
nybidari, line 627: We should also leave a TODO/comment about supporting IPv6.

cmd/ateom-gvisor/main.go
EItanya: Done in 2e5e6c7. Host-side veth lookup/delete failures are now recorded and logged as warnings, but cleanup still continues into the actor netns. The function returns the joined cleanup error afterward if any cleanup step failed.

cmd/ateom-gvisor/main.go
EItanya: Done in 2e5e6c7. Added a TODO near the nftables setup documenting that the current compatibility networking path is IPv4-only and needs IPv6 veth addressing, forwarding, and nftables support for dual-stack pods.

internal/serverboot/serverboot.go
```diff
@@ -96,16 +70,14 @@ func InitTracing(ctx context.Context, opts TracingOptions) (*sdktrace.TracerProv
...
-	if !opts.NoExporter {
-		exporter, err := otlptracegrpc.New(ctx,
-			// GKE managed traces doesn't support validating the TLS certs of the collector.
-			otlptracegrpc.WithInsecure(),
-		)
-		if err != nil {
-			return nil, fmt.Errorf("create OTLP exporter: %w", err)
-		}
-		tpOpts = append(tpOpts, sdktrace.WithBatcher(exporter))
+	exporter, err := otlptracegrpc.New(ctx,
```
ahmedtd, line 95: I think I need to do some experiments to convince myself that the kernel will not allow the same TCP/UDP ephemeral source port to be chosen by both ateom and the guest simultaneously.  But I don't think that blocks merge; we know we need veth to insulate us from the differences between different CNI implementations.

At most, it would mean that we still can't make network connections from ateom.

ahmedtd (dismissed):

bowei:
Needs rebase...

EItanya:
> Needs rebase...

Why does it need rebase? I don't see any conflicts according to GitHub.

cmd/ateom-gvisor/main.go
```diff
@@ -34,30 +33,40 @@ import (
...
+	podName      = flag.String("pod-name", "", "The name of the current pod")
 
 	reapLock sync.RWMutex
 )
 
+const (
+	hostVethName      = "ateom0"
+	actorVethName     = "eth0"
+	actorVethTempName = "ateom1"
+	hostVethCIDR      = "10.200.0.1/30"
```
bowei: Do we want to use link local addresses here? 169.254.x.x? Private addresses like 10.x.x.x would alias many user VPCs.

EItanya: Should we make this configurable in the future?

cmd/ateom-gvisor/main.go
```diff
@@ -393,60 +319,17 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
-		return nil, fmt.Errorf("while moving eth0 into interior network namespace: %w", err)
-	}
-	// Roll eth0 back to the pod netns if any subsequent step errors,
-	// otherwise the worker pod is bricked for the next actor.
 	defer func() {
 		if retErr != nil {
-			if cleanupErr := ensureEth0InPodNetns(ctx, s); cleanupErr != nil {
-				slog.WarnContext(ctx, "Failed to roll back eth0 after Restore failure", "err", cleanupErr)
+			if cleanupErr := s.cleanupActorNetwork(ctx); cleanupErr != nil {
+				slog.WarnContext(ctx, "Failed to clean up actor network after Restore failure", "err", cleanupErr)
```
bowei: I think this should cause Ateom to exit if there are any issues with cleanup of the Worker. In the worst case scenarios, this just results in decreased capacity as the Worker is recycled. Otherwise we potentially have issues with stale state persisting through successive Actor runs.

cmd/ateom-gvisor/main.go
bowei: Yes, file an issue to track. I'm not certain we have a system configuration story for parameters like this, so it that may need some decisions. For now, can we make it link-local, say 169.254.17.x/30? I know that use of 10.x.x.x will run into problems out of the gate.

cmd/ateom-gvisor/main.go
bowei: Can you reference this const or remove?

bowei:
I know there are comments discussing the networking setup, but I think after fixing the remaining comments (mark them as resolved) and filing the follow up issues, we can get this one in. We should expect that things will change as we make progress on the project.

EItanya:
> I know there are comments discussing the networking setup, but I think after fixing the remaining comments (mark them as resolved) and filing the follow up issues, we can get this one in. We should expect that things will change as we make progress on the project.

Fixed the last link local comment, should be ready to go

internal/serverboot/serverboot.go
bowei: I would like to understand the requirement that Ateom itself needs to have an ingress path to it network wise. This is not relevant to this review but it would be good to write down why.

bowei (approved):

PR #132, opened 2026-05-31
Update/verify dep vendoring

### LINKED ISSUE #19: Should we vendor our dependencies?
It's sort of a pain, but it provides a level of transparency we don't have without it.

thockin:
This adds an update/verify script pair to check `go mod vendor` in the main module.  Should we do this in sub-modules too?

Fixes #19


dims:
> Should we do this in sub-modules too?

yes please!

thockin:
Now vendors all go.mod directories

thockin:
I broke up the large commits into script-changes and run-the-script commits.  the "run" commits are just the output of running the script.

dims (approved):
LGTM

the individual vendor/ for each tool ... we'll get used to it.

thockin:
rebased

BenTheElder:
> This adds an update/verify script pair to check go mod vendor in the main module. Should we do this in sub-modules too?

I'm not convinced. That adds a LOT more bloat, and reviewing tools-module vendor is not nearly as actionable.

We're using go mod as an implementation detail of obtaining those tools, we could just as well fetch binaries.

BenTheElder:
Also, we shouldn't be dropping anything from LICENSES here, what's going on there?

LICENSES/github.com/hashicorp/go-reap/reap_stub_test.go
```diff
@@ -1,27 +0,0 @@
-// +build windows solaris
```
BenTheElder, line 1: The code in LICENSES was here because the library is MPL and requires redistributing the sources. I think this commit broke something.

BenTheElder:
Github won't show me the full diff, even commit by commit.

dims:
@BenTheElder i switched to `diffshub` for large commits

BenTheElder:
Also, the tools modules are supposed to be updated by updating the version of the tool and letting the other deps be auto selected to the versions specified by the to upstream. We have a script for that. Which avoids breakage caused by patching those tools in unexpected ways.

So updating the tools will create a huge churn each time as ~all transitive deps update at once.

I would rather work out binary fetch for each than pretend to review their vendor.

If the concern is those tools stop building because the module proxy is missing a dependency, we should use binaries. If those are a concern, well we have bottom turtles we need to download for our images too.

thockin:
One of the motivations for vendoring is "I am at the airport.  I quickly clone the repo and get on the plane.  I can do work."  If I can't compile, that's a problem?

BenTheElder:
We don't vendor the base images though.
The modules will be in your system module cache, just run a build before you leave.

LICENSES/github.com/hashicorp/go-reap/reap_stub_test.go
thockin: Running `go mod vendor` causes go-licenses to NOT create these files.  No logs, no warnings.  Adding `--include_tests` does not bring it back (though we should probably add that).

Given that these are tests, they are not linked into our project, so I think the need for them is dubious.

thockin:
`just run a build before you leave` is insufficient.  You at least need to run `hack/update-all`.  Anyway, I think I agree with you.  @dims?

I moved the commits to vendor tools to the end of this series.  Other than that, I think it's good.

thockin:
In fact, I pushed without the tools now.  Let's debate that in a followup.

BenTheElder (approved):

PR #134, opened 2026-05-31
Atelet runs only on nodes hosting ateoms (#9)

### LINKED ISSUE #9: Atelet should only run on nodes where ateoms are running.
Today, atelet is deployed as a DaemonSet, which means it runs on all nodes in the cluster, taking into account taints and tolerations.

Ateom pods are managed as Deployments today, though we may eventually switch to a custom workload controller.

We should find a solution that places atelet pods only on nodes where ateoms are or will be running.

This could look like:
* Manage atelets from the same workload controller as ateoms, so that it can place atelet pods on the correct nodes
* Come up with a convention for tainting certain nodes as only for ateoms, not regular workloads.

eliranw:
## Summary

Closes #9. Atelet currently runs on every node as a DaemonSet. This makes it run **only on nodes that currently host ateom pods**.

A new `AteletNodeReconciler` (Pod-keyed) watches ateom pods and maintains a substrate-owned `ate.dev/atelet=true` label on each Node currently hosting an ateom. The atelet DaemonSet now carries `nodeSelector: ate.dev/atelet=true`, so its footprint follows ateom placement.

### Mechanism

- **Reactive labeling.** On each ateom-pod event, the reconciler lists pods on the affected node (via a cached `spec.nodeName` field index) and SSA-applies the desired node state.
- **Per-pool refcounting via annotations.** Each node carries one `ate.dev/claim.<workerpool-uid>` annotation per WorkerPool occupying it. The label is present iff at least one claim exists; SSA's field-ownership prunes claims/label as pods leave. Multiple WorkerPools can safely share a node.
- **Init container.** Every ateom pod gets a `wait-for-atelet` init container that TCP-probes `$HOST_IP:8085` until the local atelet is serving. This absorbs the scheduling→atelet-ready gap and makes ateoms robust to atelet restarts/upgrades mid-life.
- **Finalizer.** A `ate.dev/release-node-claims` finalizer on WorkerPool holds deletion until its claims are released, so claims can't leak.
- **RBAC.** Adds `nodes: get;list;watch;patch` and `pods: get;list;watch` to the controller ClusterRole (regenerated from kubebuilder markers).

### Why reactive (not a placement policy)

Substrate doesn't decide *which* nodes ateoms run on — the kube-scheduler does, freely. The reconciler just records that choice as a label. This keeps the change small and avoids inventing a node-selection policy; richer schemes (worker classes, capacity reservations) can layer on later.

## Test plan

- [x] envtest: single pod → node gets label + claim
- [x] envtest: two pools on one node → two claims, one label
- [x] envtest: delete one pool's pod with another present → only its claim removed, label stays
- [x] envtest: delete last pod → claim and label both removed (SSA prune)
- [x] envtest: WorkerPool deletion held by finalizer until claim released, then completes
- [x] envtest: ateom Deployment carries the `wait-for-atelet` init container
- [x] `make verify` (tests, gofmt, lint, codegen, licenses, go-mod-tidy, boilerplate)

## Migration / rollout

This replaces the atelet DS's "run on all nodes" behavior with "run on labeled nodes only." Recommended two-phase rollout for existing clusters:

1. Deploy the new controller image + RBAC first. It observes existing ateom pods and labels their nodes.
2. Apply the atelet DS manifest change. The rolling update drops atelet from unlabeled nodes (no ateoms there) and keeps it on labeled ones (no disruption).

On a **fresh** cluster the atelet DaemonSet runs zero pods until the first WorkerPool's ateoms are scheduled — expected, not a broken install (`kubectl rollout status` on a 0-desired DaemonSet returns success immediately). The init container means even a single-shot `kubectl apply -k` is safe: ateoms wait rather than crash.

### DIFF
--- a/cmd/atecontroller/main.go
+++ b/cmd/atecontroller/main.go
@@ -88,6 +88,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = (&controllers.AteletNodeReconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "AteletNode")
+		os.Exit(1)
+	}
+
 	//+kubebuilder:scaffold:builder
 
 	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
--- a/internal/controllers/ateletnode_controller.go
+++ b/internal/controllers/ateletnode_controller.go
@@ -33,28 +33,22 @@ import (
 )
 
 const (
-	// AteletNodeLabel is the substrate-owned label that gates atelet DS
-	// scheduling. Present on a Node iff at least one ateom pod is
-	// currently scheduled to that Node.
 	AteletNodeLabel      = "ate.dev/atelet"
 	AteletNodeLabelValue = "true"
 
-	// AteletNodeClaimAnnoPrefix is the prefix for per-pool claim
-	// annotations on Nodes. Full key: ate.dev/claim.<workerpool-uid>.
+	// Per-pool claim annotation. Full key: ate.dev/claim.<workerpool-uid>.
 	AteletNodeClaimAnnoPrefix = "ate.dev/claim."
 
-	// AteletNodeFieldOwner is the SSA field owner for substrate-managed
-	// Node fields (the label and claim annotations).
 	AteletNodeFieldOwner = "atelet-node-controller"
-
-	// PodNodeNameIndex is the field-indexer key for Pod.Spec.NodeName.
-	// Required for List(client.MatchingFields{PodNodeNameIndex: ...}).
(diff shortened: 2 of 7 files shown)


Gentoli:
Node label based on-demand scheduling seems easy at first but can be complicated.

For example, on cpu capacity, if an ateom pod scheduled makes the node full, then the new atelet needs to either evict pods to schedule or it cannot start, making ateom hang.

eliranw:
@Gentoli I agree with your concern.
I can add a `priorityClassName: system-node-critical` (plus modest resource requests) to the atelet DS, so it can preempt to claim room, taking this pattern from projects like gpu-operator.
The atelet-first (or pro-active) alternative is a bigger change: an admin (or a component inside substrate) might label nodes for atelet and as you note those nodes have to actually fit what the ateoms need (GPU, memory, etc.), plus we might want a readiness signal so ateoms don’t schedule before atelet is serving.

PR #136, opened 2026-06-01
Feature/flexible persistence via ListOptions

### LINKED ISSUE #135: RFC: Flexible & Non-Ossifying Control Plane Persistence Layer
# RFC: Flexible & Non-Ossifying Control Plane Persistence Layer

**Authors**: MushuEE / Agent Substrate Maintainers
**Status**: Draft / Proposed

---

## What's the Goal?

As Agent Substrate matures, we want to make sure our Control Plane persistence layer (`store.Interface`) doesn't paint us into a corner. 

Specifically, we want to achieve three things:
1. **Pluggable Databases**: Allow users/deployers to bring their own database of choice (PostgreSQL, Spanner, Aerospike, DynamoDB) without rewriting the entire API server.
2. **Locality-Aware Scheduling (Hot Snapshot Cache)**: Enable the scheduler to query for workers that already have a target snapshot hot-cached, with a clean fallback to the general pool if none are warm.
3. **Non-Ossifying API Surface**: Add these advanced query capabilities without bloating our Go interfaces or hardcoding specific filters into the persistence method signatures.

---

## The Problem Today

Right now, our `store.Interface` listing methods are completely rigid and all-or-nothing:
```go
ListWorkers(ctx context.Context) ([]*ateapipb.Worker, error)
ListActors(ctx context.Context) ([]*ateapipb.Actor, error)
```

This introduces severe issues as the project scales:
* **No indexing/filtering in the DB**: To find a free worker in a specific pool, the scheduler is forced to retrieve *every single worker* from the database and filter them in-memory in Go.
* **Zero flexibility**: If we want to query workers by Node name, size, or status, we either have to add new custom query methods to the interface (bloating the API surface) or continue doing extremely expensive in-memory scans in the control plane.
* **Database lock-in**: Alternative databases (like SQL or DynamoDB) cannot leverage their indexing engines (e.g. B-Trees) to optimize queries because the interface gives them zero filtering context.

---

## The Proposal: Kubernetes-style `ListOptions`

Rather than hardcoding specific query parameters, we introduce a generic, extensible `ListOptions` block to list operations:

```go
// ListOptions holds parameters for filtering List results.
type ListOptions struct {
	// FieldSelector filters results by matching specific fields exactly (e.g., "worker_pool=cpu-pool").
	FieldSelector map[string]string
}
```

This allows our listing methods to support generic query filtering:
```go
ListWorkers(ctx context.Context, opts ListOptions) ([]*ateapipb.Worker, error)
ListActors(ctx context.Context, opts ListOptions) ([]*ateapipb.Actor, error)
```

---

## Technical Arguments

<details>
  <summary><b>1. Non-Ossifying API Surface</b></summary>

Using a map-based `FieldSelector` completely decouples the query signature from the resource schemas:
* If a new field (e.g., `node_name`, `size`, `gpu_type`) is added to the `Worker` protobuf tomorrow, **the Go interface does not change.** 
* New clients can start passing `opts.FieldSelector["node_name"] = "k8s-node-1"` immediately.
* The database implementation simply adds a new `case` statement to its internal matcher to support it.

</details>

<details>
  <summary><b>2. Enabling Production-Grade Pluggable DBs</b></summary>

This design allows pluggable database drivers to translate these selectors directly into native indexing queries instead of scanning the entire database:

* **PostgreSQL / Spanner**: Can map the `FieldSelector` map directly to parameterized SQL `WHERE` clauses (leveraging B-Tree indexes):
  ```sql
  SELECT * FROM workers WHERE worker_pool = $1 AND actor_id = $2;
  ```
* **DynamoDB**: Can map the selector directly to Key Condition Expressions or Filter Expressions.
* **Redis (Default)**: Maintains simple, low-overhead client-side filtering on top of Master key scans for development/testing, but can easily be optimized to perform set lookups (e.g., `SPOP` or `SRANDMEMBER` on `set:idle:cpu-pool`) for specific common selector combinations.

</details>

<details>
  <summary><b>3. High-Performance Locality & Snapshot Scheduling</b></summary>

A major performance booster for Agent Substrate is scheduling actors onto workers that already have the snapshot hot-cached in memory or locally on disk. 

With `ListOptions`, the scheduler can execute a clean fallback loop:

```go
// 1. Try to find a free worker with the snapshot hot-cached
warmWorkers, err := s.store.ListWorkers(ctx, store.ListOptions{
	FieldSelector: map[string]string{
		"worker_pool":     "cpu-pool",
		"actor_id":        "", // Free workers only
		"cached_snapshot": targetSnapshotURI,
	},
})
if err == nil && len(warmWorkers) > 0 {
	return warmWorkers[0], nil // Hot-cache hit!
}

// 2. Fallback: Find any idle worker in the pool
return s.store.ListWorkers(ctx, store.ListOptions{
	FieldSelector: map[string]string{
		"worker_pool": "cpu-pool",
		"actor_id":    "",
	},
})
```

Under the hood, a custom SQL database or advanced Redis driver can intercept the first query and check a specialized snapshot-to-worker index table to return matches in $O(1)$ time.

</details>

## Conclusion

Adding `ListOptions` strikes a balance between **minimal invasiveness today** and **complete architectural readiness for high-performance custom databases tomorrow**. It solves the scaling issues of in-memory filtering without forcing massive, premature changes to the codebase. As workers are very likely to need to support structural drift, such as scaling (size/cpu), and other worker/actor specifics will evolve on a deployment basis, this should enable some flexibility.

MushuEE:
Fixes #135

Decouples the control plane persistence layer (`store.Interface`) from rigid listing constraints by introducing a generic, non-ossifying query filtering surface.
### Problem
Currently, the persistence listing methods (`ListWorkers` and `ListActors`) are all-or-nothing. To filter workers by pool or namespace, the scheduler must fetch all records from the database and perform in-memory filtering in Go. This approach:
1. Scales poorly at high worker counts.
2. Coupled the storage API surface tightly to structural changes.
3. Prevents pluggable databases (like SQL or DynamoDB) from leveraging B-Tree indices to execute $O(1)$ query-time filtering.
### Solution
1. **Introduced `store.ListOptions`**: Added a generic `ListOptions` struct with a `FieldSelector map[string]string` parameter to the storage `Interface` listing methods.
2. **Non-Ossifying Mapping Strategy**: Implemented allocation-free, fast switch-based field mapping helpers (`matchesWorker`, `matchesActor`) in the `ateredis` package. This lets the storage engine filter items internally before returning them.
3. **Scheduler Integration**: Updated the API service handlers and scheduling workflow (`AssignWorkerStep`) to pass the options cleanly.
4. **Validation**: Added robust unit tests verifying exact-match field selector filtering (e.g., by pool, empty `actor_id` to claim ready workers, and actor `status`).
### Architectural Advantages
* **Decoupled API Surface**: Adding new worker fields (e.g., `node_name`, `gpu_type`) does not require changing the `store.Interface` signatures.
* **Database Pluggability**: Relational database drivers (like PostgreSQL or Spanner) can translate the `FieldSelector` directly into indexed SQL `WHERE` clauses.
* **Locality-Aware Ready-Checks**: Allows the scheduler to query for free workers with specific snapshot locality hints (warm-cache lookups) with a clean fallback to the general pool.

### DIFF
--- a/cmd/ateapi/internal/store/ateredis/ateredis.go
+++ b/cmd/ateapi/internal/store/ateredis/ateredis.go
@@ -353,7 +353,7 @@ func (s *Persistence) UpdateActor(ctx context.Context, actor *ateapipb.Actor, ex
 	return nil
 }
 
-func (s *Persistence) ListWorkers(ctx context.Context) ([]*ateapipb.Worker, error) {
+func (s *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) ([]*ateapipb.Worker, error) {
 	var result []*ateapipb.Worker
 	var mu sync.Mutex
 
@@ -377,6 +377,10 @@ func (s *Persistence) ListWorkers(ctx context.Context) ([]*ateapipb.Worker, erro
 				return fmt.Errorf("in protojson.Unmarshal: %w", err)
 			}
 
+			if !matchesWorker(worker, opts) {
+				continue
+			}
+
 			mu.Lock()
 			result = append(result, worker)
 			mu.Unlock()
@@ -393,7 +397,7 @@ func (s *Persistence) ListWorkers(ctx context.Context) ([]*ateapipb.Worker, erro
 	return result, nil
 }
 
-func (s *Persistence) ListActors(ctx context.Context) ([]*ateapipb.Actor, error) {
+func (s *Persistence) ListActors(ctx context.Context, opts store.ListOptions) ([]*ateapipb.Actor, error) {
 	var result []*ateapipb.Actor
 	var mu sync.Mutex
 
@@ -416,6 +420,10 @@ func (s *Persistence) ListActors(ctx context.Context) ([]*ateapipb.Actor, error)
 				return fmt.Errorf("in protojson.Unmarshal: %w", err)
 			}
 
+			if !matchesActor(actor, opts) {
+				continue
+			}
+
 			mu.Lock()
 			result = append(result, actor)
 			mu.Unlock()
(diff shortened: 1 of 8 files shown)


cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -429,6 +437,88 @@ func (s *Persistence) ListActors(ctx context.Context) ([]*ateapipb.Actor, error)
 	return result, nil
 }
 
+func matchesWorker(w *ateapipb.Worker, opts store.ListOptions) bool {
+	if len(opts.FieldSelector) == 0 {
+		return true
+	}
+	for k, v := range opts.FieldSelector {
+		switch k {
+		case "worker_namespace":
```
troychiu, line 446: Since the storage interface (`store.Interface`) is intended to be pluggable, multiple database backends must agree on the exact string keys used in ListOptions.FieldSelector. Would it make sense to define these field selector keys as exported constants in the `store` package?

MushuEE: I assumed this would be governed by the proto representation. Though I am now looking at introducing labels: https://github.com/agent-substrate/substrate/issues/186

For now however I may hold off changes as I am unsure if we want to enable this level of filtering at the client level or at the database level.

PR #137, opened 2026-06-01
Don't use K8s corev1.EnvVar

ahmedtd:
ActorTemplate should not refer directly to a K8s type, although we want to keep the same shape of the API for features we support.  Define an equivalent EnvVar that only has the things we currently support.


EItanya:
I already do this in #20?

dims (approved):

ahmedtd:
> I already do this in #20?

I think there's still some points of discussion happening on your env secrets PR, and I wanted to make sure that we remove this type ASAP.   Apologies for the churn.

PR #144, opened 2026-06-02
lint: enable more golangci-lint plugins

ameukam:
Enable contextcheck, gocritic, and exhaustive to catch detached contexts, bug patterns, and incomplete controller state-machine switches.

Assisted by GPT 5.5

- [x] Tests pass

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -33,6 +33,8 @@ linters:
     - contextcheck
     # High-signal bug-pattern checks without enabling style-only churn.
     - gocritic
+    # CRD phase/state-machine switches should handle every declared state.
+    - exhaustive
 
   settings:
     errcheck:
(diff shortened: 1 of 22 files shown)


ameukam:
cc @BenTheElder

troychiu (commented):
Can we also enable formatters which group imports by prefix, such as goimports?

thockin:
This looks overall good, but:
a) needs a rebase
b) can I arm-twist you into enabling one per commit so we can see which fixes go to which lints?

BenTheElder:
Let's hold on the formatters part.

ameukam:
> This looks overall good, but: a) needs a rebase b) can I arm-twist you into enabling one per commit so we can see which fixes go to which lints?

@thockin it's done

PR #145, opened 2026-06-02
feat(observability): add Prometheus to the kind dev stack

### LINKED ISSUE #106: Install a prometheus server on Kind
So we can collect metrics from core components such as ate-apiserver, atelet, atelet, etc.

konippi:
Adds a Prometheus server to the kind dev stack (in `otel-system`, alongside the OpenTelemetry Collector and Jaeger). It scrapes component `/metrics` via the existing `prometheus.io/scrape` annotations on `ate-api-server` and `atelet`, scoped to the `ate-system`/`otel-system` namespaces.

Fixes #106

- [x] Tests pass

### DIFF
--- a/manifests/ate-install/kind/prometheus.yaml
+++ b/manifests/ate-install/kind/prometheus.yaml
@@ -1,0 +1,201 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


manifests/ate-install/kind/prometheus.yaml
```diff
@@ -0,0 +1,201 @@
...
+  namespace: otel-system
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: prometheus-config
+  namespace: otel-system
+data:
+  prometheus.yml: |
+    global:
```
juli4n, line 68: Was this taken from some other place?  (e.g. prometheus installation examples? In that case, can we add a reference to it?

konippi: Yes — adapted from the prometheus-community Helm chart's [`kubernetes-pods` job](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml); added a reference comment.

PR #146, opened 2026-06-02
Fix stale kind cluster setup for certs and valkey 

npolshakova:

- [x] Tests pass

Had a kind cluster up for a couple days and ran into these issues:

1. Cert expires and `./hack/install-ate-kind.sh --deploy-ate-system` setup script won't refresh the certs

```
  while listing actors in db: while iterating all redis master: tls: failed to verify certificate: x509: certificate has expired or is not yet valid: current time 2026-06-02T14:26:03Z is after 2026-05-30T18:46:15Z
```

This adds a new `--refresh-local-certs` flag to recreate local CA roots used by the podcertificate controller, local session identity JWT/CA pools, CA bundle used by API server to trust Valkey TLS and restarts the pods that mount generated certs.

2. Valkey remembers old pod IPs after the StatefulSet restarts. In the `ate-api-server-deployment` it kept timing out since the old IPs were being used. 

The new `--reset-local-valkey` deletes `job/valkey-cluster-init` and Valkey PVCs with label `app=valkey-cluster` so the Valkey Cluster can be initialized from scratch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,22 @@ Similarly, you can deploy or cleanup specific Agent Substrate components using t
 ./hack/install-ate-kind.sh --delete-all
 ```
 
+For local `kind` clusters that have been running for several days, generated
+certificate material or Valkey's persisted cluster metadata can become stale.
+Use the local recovery flags before redeploying the system:
+
+```bash
+# Regenerate local cert/JWT prerequisites and restart cert consumers.
+./hack/install-ate-kind.sh --refresh-local-certs --deploy-ate-system
+
+# Delete local Valkey state and reinitialize the cluster. This removes local
+# actor and worker control-plane state.
+./hack/install-ate-kind.sh --reset-local-valkey --deploy-ate-system
+
+# If both expired certs and stale Valkey pod IPs are present:
+./hack/install-ate-kind.sh --refresh-local-certs --reset-local-valkey --deploy-ate-system
+```
+
 #### Tearing down resources (GCP)
 
 If you need to delete the resources created by the setup script, you can use the provided script `hack/teardown.sh`. This script will delete resources in the reverse order of creation and handles partial failures gracefully.
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -62,6 +62,8 @@ function usage() {
   echo "  --deploy-ate-system                    Deploy core system (CRDs, atelet, apiserver)"
   echo "  --delete-ate-system                    Delete core system"
   echo "  --delete-all                           Delete core system and all registered demos"
+  echo "  --refresh-local-certs                  Regenerate local cert/JWT prerequisites and restart cert consumers"
+  echo "  --reset-local-valkey                   Delete local kind Valkey state so the cluster can be reinitialized"
   echo ""
   echo "Infrastructure components:"
   echo ""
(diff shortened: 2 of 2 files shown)


npolshakova:
@BenTheElder @dberkov saw you had some recent changes in the setup scripts. Any thoughts about adding this for us with long-lived kind clusters 😅

hack/install-ate.sh
```diff
@@ -62,6 +62,8 @@ function usage() {
   echo "  --deploy-ate-system                    Deploy core system (CRDs, atelet, apiserver)"
   echo "  --delete-ate-system                    Delete core system"
   echo "  --delete-all                           Delete core system and all registered demos"
+  echo "  --refresh-local-certs                  Regenerate local cert/JWT prerequisites and restart cert consumers"
```
BenTheElder, line 65: We should probably figure out how to fix this generally so it doesn't require manual intervention, kind aside, cc @MushuEE @juli4n (anyone is welcome to look at this please, I just know these two are looking at valkey)

BenTheElder (commented):
Heh, I think we're all iterating so much with clean installs every time.

I've also been wondering if we should drop the number of replicas etc at least on kind to make startup faster and use less resources ... but there's a tradeoff since that's our main source of continuous test coverage right now (and we expect "real" clusters) to need to deal with shards (which affects ate-apiserver).

BenTheElder:
x-ref https://github.com/agent-substrate/substrate/issues/225

README.md
```diff
@@ -168,6 +168,22 @@ Similarly, you can deploy or cleanup specific Agent Substrate components using t
 ./hack/install-ate-kind.sh --delete-all
 ```
 
+For local `kind` clusters that have been running for several days, generated
```
dberkov, line 171: it is a chance the issue exists for non `kind` version too. Yesterday I tried to re-deploy ate-api and it failed. It was complaining on expired token at startup. Deleting everything and installing from scratch solved the issue.

hack/install-ate.sh
npolshakova: Yeah I agree- can we change `manifests/ate-install/valkey.yaml` to either deal with the stale IPs or cert expiry? (maybe announce the dns names and have the init job use hostnames?) @MushuEE @juli4n

README.md
npolshakova: Yeah if there's work being done on the valkey side, I think we should fix this generally instead of just adding this to the script (although I'll probably keep this change locally for now to make it less painful for me)

hack/install-ate.sh
BenTheElder: Yeah, maybe even something hacky like a cronjob in the short term? 😂

MushuEE: I captured some of the risks in the cert space: 

- https://github.com/MushuEE/substrate/blob/main/docs/valkey/risk-register.md#r-6-ca-bundle-rotation-hazard
- https://github.com/MushuEE/substrate/blob/main/docs/valkey/risk-register.md#r-7-no-pod-certificate-monitoring

MushuEE: https://github.com/MushuEE/substrate/blob/main/docs/valkey/recovery-procedures.md#r-10-recover-from-cert-expiry

npolshakova:
Closing in favor of https://github.com/agent-substrate/substrate/pull/267 and maybe future valkey changes

PR #152, opened 2026-06-02
Use pflag instead of Go's flag lib

thockin:
Better to get ahead of this early.

### DIFF
--- a/cmd/ateom-gvisor/main.go
+++ b/cmd/ateom-gvisor/main.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"context"
-	"flag"
 	"fmt"
 	"log/slog"
 	"net"
@@ -36,6 +35,7 @@ import (
 	"github.com/agent-substrate/substrate/internal/serverboot"
 	"github.com/agent-substrate/substrate/internal/version"
 	"github.com/hashicorp/go-reap"
+	"github.com/spf13/pflag"
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
@@ -45,19 +45,20 @@ import (
 )
 
 var (
-	podUID = flag.String("pod-uid", "", "The UID of the current pod")
+	podUID = pflag.String("pod-uid", "", "The UID of the current pod")
 
-	showVersion = flag.Bool("version", false, "Print version and exit.")
+	showVersion = pflag.Bool("version", false, "Print version and exit.")
 
 	reapLock sync.RWMutex
 )
 
 func main() {
-	flag.Parse()
+	pflag.Parse()
 	if *showVersion {
 		fmt.Println(version.String())
 		return
 	}
+
 	ctx := context.Background()
 
 	if err := do(ctx); err != nil {
--- a/demos/counter/counter.go
+++ b/demos/counter/counter.go
@@ -21,7 +21,6 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"encoding/base64"
-	"flag"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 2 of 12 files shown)


dims:
@thockin need to fix these?

https://github.com/agent-substrate/substrate/blob/main/internal/controllers/workerpool_controller.go#L143-L144

dims:
- hack/run-e2e.sh:106 — ... -args -e2e ... → --e2e
- hack/run-e2e.sh:103 — extra_e2e_args+=("-kube-context" ...) → --kube-context
- .github/workflows/pr-workflow.yaml:49 — hack/run-e2e.sh -v -args -no-color → --no-color
- internal/e2e/testmain.go:45 — the skip message still tells users go test ./internal/e2e/... -args -e2e (single dash, now broken) → --e2e.
- internal/e2e/README.md:5,14 and hack/run-e2e.sh help text (lines 40-47) — examples show single-dash -e2e/-no-color/-kube-config/-kube-context.

internal/e2e/testmain.go
```diff
@@ -16,9 +16,11 @@ package e2e
 
 import (
 	"context"
-	"flag"
+	goflag "flag"
```
troychiu, line 19: Is this renaming required?

thockin: Go's test flags are defined in terms of `flag`.  It's sad that Go's stdlib is so good in some places and so bad in others.

thockin:
Working on the e2e flags.  I missed those

dims (approved):

troychiu (approved):

ahmedtd:
Why do we want to do this?  What does pflag give us that we need?

thockin:
POSIX-ish flags.

Go's flag library treats `--foo` and `-foo` as the same thing, whereas every other UNIXy flag system knows `-foo` means `-f -o -o`.  pflag is effectively part of the de facto extended standard library.

demos/sandbox/client/main.go
```diff
@@ -19,8 +19,8 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/json"
-	"flag"
 	"fmt"
+	"github.com/spf13/pflag"
```
dberkov: nit: could you please move it to independent section from go imports

demos/counter/counter.go
```diff
@@ -21,8 +21,8 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"encoding/base64"
-	"flag"
 	"fmt"
+	"github.com/spf13/pflag"
```
dberkov: nit: could you please move it to independent section from go imports

cmd/ateom-gvisor/main.go
```diff
@@ -27,6 +26,8 @@ import (
 	"sort"
 	"sync"
 
+	"github.com/spf13/pflag"
```
dberkov: What is a reason for this import to be in a separate block? 

Same style used in all files. Therefore could you please update all the files if it was non intentionally.

dberkov (changes requested):

ahmedtd:
> POSIX-ish flags.
> 
> Go's flag library treats `--foo` and `-foo` as the same thing, whereas every other UNIXy flag system knows `-foo` means `-f -o -o`. pflag is effectively part of the de facto extended standard library.

No, I understand what pflag does.  I don't understand what not having posix-ish flags is blocking.  We are not a posix binary, we have no system conventions to fit into. 

> knows `-foo` means `-f -o -o`

Personally, I think this is a misfeature in almost all cases, but especially for server binaries.

cmd/ateom-gvisor/main.go
thockin: This is whatever `goimports` did will fix to be more consistent :)

thockin:
This may come down to taste -- even if we never use short flags, we are a Linux system and Go's flag conventions don't fit.  In addition to that, pflag has a richer set of types it supports, such as slices of strings, which is why Kubernetes and so many other projects have adopted pflag.

dberkov (approved):

dberkov:
@ahmedtd 
>Personally, I think this is a misfeature in almost all cases, but especially for server binaries.

k8s uses pflag in all the command lines and servers (ex: https://github.com/kubernetes/kubernetes/blob/90b2754235c7927987d22f695a04b519a8ad04cf/cmd/kubelet/app/server.go#L40)

PR #155, opened 2026-06-03
feat(serverboot): set service.instance.id on the telemetry resource

### LINKED ISSUE #154: Add per-instance identity to the telemetry resource
`internal/serverboot` builds the OTel resource with only `service.name`, so instances of the same service are indistinguishable. `atelet` runs as a DaemonSet (one per node), so its OTLP metrics share an identical resource and collide in the backend and can't be broken down per pod. `OTEL_RESOURCE_ATTRIBUTES` is also ignored today (no `WithFromEnv`), so the standard Kubernetes way of injecting pod identity has no effect.

`service.instance.id` is Recommended by the OTel resource semconv and isn't set by the SDK automatically.

Proposal: build the shared tracer/meter resource with `service.instance.id` (UUID fallback) and `WithFromEnv` last so env overrides defaults; have the Deployments/DaemonSet inject pod identity via the Downward API (`service.instance.id=$(POD_UID)`, `k8s.namespace.name/pod.name/pod.uid`).

Noticed while working on the Kind Prometheus PR (#145, for #106): scraped metrics get per-pod identity from the pod label, but OTLP-exported metrics have none.

konippi:
Fixes #154

Give `serverboot`'s OTel resource a per-instance identity so multi-instance components (e.g. the `atelet` DaemonSet) don't collide in the backend, and honor `OTEL_RESOURCE_ATTRIBUTES`. The shared tracer/meter resource now carries `service.instance.id` (UUID fallback) with `WithFromEnv` last so env overrides defaults (`ErrPartialResource`/`ErrSchemaURLConflict` are non-fatal), and the `ate-api-server` Deployment / `atelet` DaemonSet inject pod identity via the Downward API (`service.instance.id=$(POD_UID)`, `k8s.namespace.name/pod.name/pod.uid`).

- [x] Tests pass

### DIFF
--- a/internal/serverboot/serverboot.go
+++ b/internal/serverboot/serverboot.go
@@ -20,12 +20,14 @@ package serverboot
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net/http"
 	"os"
 
 	"github.com/agent-substrate/substrate/internal/contextlogging"
+	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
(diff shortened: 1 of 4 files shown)


HavenXia (approved):
Thanks for the instance id fix,  without it multiple instances of one service collapse onto a single backend time series.

`atenet-router` also runs `serverboot` in #157 and now emits metrics; without this env block it'll fall back to a random UUID instance id on every restart. I'll add the matching block in #157

PR #162, opened 2026-06-03
Fix `make` with LDFLAGS

thockin:
`$(KO)` != `$(GO)`
    
ko does not support any way to pass build flags directly.  We have to
use GOFLAGS.  Go does not fully shell-parse GOFLAGS, it just does simple
whitespace splitting and whole-value quoting.
    
E.g. this fails for good reason:
    GOFLAGS="-ldflags=-X=foo=foo -X=bar=bar"
...becomes:
    [ "go", "build", "-ldflags=-X=foo=foo", "-X=bar=bar" ]
...where `-X` is passed to `go build`
    
We can try:
    GOFLAGS="-ldflags='-X=foo=foo -X=bar=bar'"
...but that becomes
    [ "go", "build", "-ldflags='-X=foo=foo", "-X=bar=bar'" ]
...which is unfortunate.
    
Go does some quote handling, but:
    GOFLAGS='-ldflags="-X=foo=foo -X=bar=bar"'
...still becomes:
    [ "go", "build", "-ldflags=\"-X=foo=foo", "-X=bar=bar\"" ]
...which is surprising.
    
We can try:
    GOFLAGS="-ldflags=-X=foo=foo -ldflags=-X=bar=bar"
...but go just takes the last value of that flag rather than the accumulation.
    
What does work is:
    GOFLAGS='"-ldflags=-X=foo=foo -X=bar=bar"'
...with the WHOLE flag quoted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,12 +29,8 @@ ATECTL := $(BINDIR)/kubectl-ate
 # Version stamping. Override on the make command line to pin
 # (e.g. `make VERSION=v0.5.0 build`).
 VERSION    ?= $(shell git describe --tags --always --dirty 2>/dev/null || echo dev)
-COMMIT     ?= $(shell git rev-parse HEAD 2>/dev/null || echo unknown)
-BUILD_DATE ?= $(shell date -u +%Y-%m-%dT%H:%M:%SZ)
 VERSION_PKG := github.com/agent-substrate/substrate/internal/version
-LDFLAGS := -X $(VERSION_PKG).Version=$(VERSION) \
-           -X $(VERSION_PKG).Commit=$(COMMIT) \
-           -X $(VERSION_PKG).BuildDate=$(BUILD_DATE)
+LDFLAGS := -X=$(VERSION_PKG).Version=$(VERSION)
 
 .PHONY: all
 all: build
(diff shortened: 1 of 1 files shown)


dims (approved):

BenTheElder:
How does this behave with the install scripts, which do not use the makefile at all AFAIK?

BenTheElder:
Do we _really_ need ldflags? Why not just use the Go-buil- in version stamping? Kubernetes uses ldflags because this functionality didn't exist.

thockin:
There's no way I can see to set a version string in Go's build info?

BenTheElder:
Go automatically grabs the version from git

thockin:
vcs.revision is the commit, we want a symbolic name for releases

thockin:
Found a way to make GOFLAGS work.

BenTheElder:
>vcs.revision is the commit, we want a symbolic name for releases

True, but that's in ~`info, err := debug.ReadBuildInfo(); info.Main.Version`

thockin:
> info.Main.Version

Where does that value come from?  There are no docs I can find on it?

BenTheElder:
> Where does that value come from? There are no docs I can find on it?

It's the module version, for the main module in the binary, which indirectly comes from vcs. 
On a tagged release it will just be the tag.

I want to avoid having a complex bash build system again, and go gives us reasonable metadata built in.

https://pkg.go.dev/runtime/debug#BuildInfo

thockin:
> On a tagged release it will just be the tag.

I tried adding a tag and got nothing.  Go's docs here are awful.

thockin:
As far as I can tell there is no way to set the Main.Version on a local build.  This PR uses an age-old technique.

BenTheElder (commented):
ACK ... I still don't see this working correctly when using e.g. hack/install-ate.sh though?

Maybe return to a ko.sh and pass the ldflags there (so both make and the install scripts can invoke it)

bowei (approved):

PR #163, opened 2026-06-03
fix(build): stamp ko images via .ko.yaml, drop unsupported ko --ldflags

dims:
UGH! i made a mistake..  ko build has no --ldflags flag (only ko run/delete tolerate unknown flags), so `make build-images` and `make build-demos` aborted with "unknown flag: --ldflags". It went unnoticed because no CI workflow runs `make build*`: e2e installs via hack/install-ate.sh, which builds images with `ko apply`/`ko resolve`. Only the plain go-build targets (kubectl-ate, atenet) actually stamped anything.

- [ ] Tests pass

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -18,6 +18,13 @@ defaultPlatforms:
   - linux/amd64
   - linux/arm64
 
+# ldflags for ko-built images ({{.Git.*}}/{{.Date}} templating). Version uses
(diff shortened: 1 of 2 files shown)


thockin:
compare to #162 ?

dims:
@thockin beat me to it! https://github.com/agent-substrate/substrate/pull/162

PR #165, opened 2026-06-03
Tidy up the messages when deleting a kind cluster

thockin:
If the registry exists and was not created by us, say so, but if it did not exist, don't waste our breath.


dims (approved):

BenTheElder (approved):
Thanks

Really, I need to reconcile this against the kind script upstream (which never deletes the registry) and make both smarter

ko raised a new issue about insistently dualstack attempting ipv6 first and then falling flat when dialing localhost and ipv6 isn't forwarded.

Assuming the host is always dualstack is an ugly shortcut.

thockin:
If you approve this, Ben, you should merge it

Issue #166, opened 2026-06-03
Avoid re-untarring actor image rootfs on every restore


Falven:
## Problem

While testing larger actor images, I noticed that `ResumeActor` can spend most of its time rebuilding the OCI rootfs, not restoring the checkpoint.

In one downstream test with a coding-agent style image:

```text
Control/ResumeActor:        ~15-20s
prepareOCIDirectory:        ~15-20s
untar image rootfs:          ~15-20s
RestoreWorkload:             ~268ms
```

So the actual `runsc restore` path was relatively fast, but `atelet` spent almost the whole request recreating the image rootfs.

This is probably fine for the small demo images, but it becomes painful for larger agent images. For example, the Claude Code multiplex demo uses a Node/Claude Code image, and downstream coding-agent images tend to include a lot of stuff.

## Current behavior

As far as I understand the current path:

1. `atelet` downloads the checkpoint files.
2. `atelet` calls `prepareOCIBundles`.
3. `prepareOCIDirectory` removes and recreates the actor rootfs.
4. The image is fetched/extracted.
5. The full image rootfs is untarred into the actor bundle.
6. `ateom.RestoreWorkload` calls `runsc restore`.

The memory pull cache helps with image fetch/cache behavior for small images, but it does not avoid repeatedly materializing the same rootfs.

## Question / proposal

Would it make sense for `atelet` to cache an extracted rootfs per immutable image digest on the node?

Rough shape:

```text
first use of image digest on node:
  pull image
  extract rootfs into a node-local cache

later restore using same image digest:
  create actor rootfs from cached rootfs
  use hardlinks/reflinks/overlay/snapshotter/etc.
  restore actor checkpoint
```

The important part is that the cache key should be the resolved image digest, not a tag.

This seems related to the earlier discussion that even if image pulling moved around, some host-local shared image/cache state is probably still needed for performance.

## Why I think this is worth tracking

For larger actor images, image dieting helps, but it does not change the scaling behavior:

```text
without rootfs reuse:
  every restore pays image extraction cost

with rootfs reuse:
  first restore on a node pays extraction cost
  later restores mostly pay actor-specific restore cost
```

This feels like a useful step toward separating:

- immutable image/rootfs state
- actor-specific memory/filesystem checkpoint state
- working-space state

## Open questions

- Is an extracted-rootfs cache compatible with the current gVisor restore assumptions?
- Is hardlink/reflink/overlay the simplest first implementation?
[...]

PR #179, opened 2026-06-04
feat(router): make atenet timeouts configurable

Falven:
## Overview
This PR makes the atenet router timeout budgets configurable while preserving the current defaults.

## Motivation
Some actor resume paths can exceed the current hard-coded ext_proc and route budgets. When that happens, Envoy can time out request processing before the router finishes the resume path. Operators should be able to tune these budgets per environment without recompiling the router.

## Implementation
- Add `--extproc-timeout` with the existing `5s` default for Envoy ext_proc gRPC and message processing.
- Add `--actor-resume-timeout` with the existing `15s` default for actor resume calls made during request routing.
- Add `--route-timeout` with the existing `10s` default; `0s` disables the Envoy route timeout.
- Thread configured durations through router startup, xDS, ext_proc, and ActorResumer.
- Add router tests for defaults, overrides, disabled route timeout, and actor resume deadline propagation.

Related to #166.

## Verification
- [x] `GOTOOLCHAIN=go1.25.0+auto GOWORK=off go test ./cmd/atenet/internal/app/router`

### DIFF
--- a/cmd/atenet/internal/app/router/extproc.go
+++ b/cmd/atenet/internal/app/router/extproc.go
@@ -39,12 +39,12 @@ type ExtProcServer struct {
 	resumer   *ActorResumer
 }
 
-func NewExtProcServer(port int, apiClient ateapipb.ControlClient) *ExtProcServer {
+func NewExtProcServer(port int, apiClient ateapipb.ControlClient, actorResumeTimeout time.Duration) *ExtProcServer {
 	return &ExtProcServer{
 		port:      port,
 		apiClient: apiClient,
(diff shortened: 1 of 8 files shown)


bowei:
Can you group the timeouts into its own struct to keep things tidy?

something like 

```
type routerTimeoutConfig struct {
  ... put the timeout values here
}
```

This will also let us easily see if some timeouts are dupe or can be derived from other values.

bowei:
Hi, we moved some of the route files to flatten the hierarchy. You will need to rebase (these should just moved).

PR #182, opened 2026-06-04
feat(atelet): inject ATE_ACTOR_ID into actor process env

### LINKED ISSUE #178: Add ActorID to the injected  actor's process
In some cases, it might be useful for the actor to know its own ID. Currently, the only way to determine this is by stripping the incoming request from atenet and extracting the ID from the host name.

Might be expose actor's ID via environment variable or a file on the rootfs?

fkautz:
Fixes #178

Exposes an actor's own ID to its workload via an `ATE_ACTOR_ID`
environment variable, so actors no longer need to strip the incoming
atenet request and parse the `Host` header to learn their identity.

### What changed
- `cmd/atelet/oci.go`: new `buildActorEnv` helper assembles the actor
  container's OCI process env (default `PATH` → template-supplied env →
  substrate-managed identity vars). `prepareOCIDirectory` now uses it.
- The injected value is **authoritative**: any template-supplied entry
  whose name collides with the reserved `ATE_ACTOR_ID` is dropped before
  substrate's value is appended. This is a security requirement, not
  cosmetic — `getenv(3)` returns the *first* match in glibc/most runtimes,
  so a surviving template-supplied duplicate would win at runtime and let
  a template spoof its actor ID and impersonate another actor.
- `cmd/atelet/oci_test.go`: `TestBuildActorEnv` covers PATH-first,
  template env preserved, the spoofed duplicate dropped, exactly one
  `ATE_ACTOR_ID`, and the authoritative value present.
- `docs/api-guide.md`: documents the variable for actor authors, next to
  the Uniform DNS / `Host`-header section.

### Naming
- Uses the repo's `ATE_*` prefix (`ATE_STORAGE_BACKEND`,
  `ATE_API_REDIS_*`) for consistency.

### Notes / follow-ups
- The rootfs-file variant mentioned in #178 (e.g. `/run/ate/actor-id`)
  is deliberately deferred; the env var resolves the core need.

- [x] Tests pass

### DIFF
--- a/cmd/atelet/oci.go
+++ b/cmd/atelet/oci.go
@@ -62,10 +62,7 @@ func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryP
 		return fmt.Errorf("in untar: %w", err)
 	}
 
-	envVars := []string{
-		"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
-	}
-	envVars = append(envVars, env...)
+	envVars := buildActorEnv(actorID, env)
 
 	ociSpec := &specs.Spec{
 		Process: &specs.Process{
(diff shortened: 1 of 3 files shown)


fkautz:
Closing this PR: the approach here (injecting `ATE_ACTOR_ID` into the actor's process environment) **does not work for the common case**, verified end-to-end on a local kind cluster with gVisor.

### Why it fails
An actor's environment lives in its **process memory**, which gVisor captures in the **golden snapshot** at `ActorTemplate` creation time. On the normal activation path — `runsc restore` from that snapshot — the checkpointed memory is reloaded as-is, and the per-actor `config.json`'s `process.env` is **not** re-applied. So every actor restored from a given template's golden snapshot sees the **golden actor's** ID, not its own.

It only works on a cold boot (`runsc run` / `--boot`), which is not the path Substrate optimizes for.

### Evidence (kind + gVisor, reading `/proc/1/environ` inside the live sandbox)

| Actor | Resume path | `config.json` says | Live `/proc/1/environ` |
|---|---|---|---|
| `sandbox-restore-1` | restore from golden (normal) | `ATE_ACTOR_ID=sandbox-restore-1` | ❌ `ATE_ACTOR_ID=<golden-uuid>` |
| `sandbox-boot-1` | `--boot` from scratch | `ATE_ACTOR_ID=sandbox-boot-1` | ✅ `ATE_ACTOR_ID=sandbox-boot-1` |

The env var is injected into `config.json` correctly, but it's frozen at the golden ID after restore — worse than missing, since it's identical (and misleading) for every actor of the template.

### Why an env var fundamentally can't do this
Per-actor identity can't ride in process memory when all actors fork from one shared golden memory image. It has to come from a channel re-evaluated **after** restore. (The "file on the rootfs" idea from #178 has the same problem if the file is baked into the rootfs — the gVisor overlay is frozen at checkpoint too.)

### Replacement
A follow-up will implement two channels that survive restore, both verified live on the same cluster:
- **Bind-mounted per-actor file `/run/ate/actor-id`** — external bind mounts are re-attached from the restore-time `config.json` (same mechanism as `/etc/resolv.conf`), so the content is correct per-actor after restore.
- **Trusted `x-ate-actor-id` request header** from `atenet` — per-request, set by the router, overwrites any client-supplied value (spoof-proof).

Closing in favor of that approach. Tracking issue: #178.

PR #184, opened 2026-06-04
feat(atelet): expose actor identity via /run/ate/actor-id

### LINKED ISSUE #178: Add ActorID to the injected  actor's process

fkautz:
Fixes #178

Gives an actor a reliable way to learn its own ID without parsing the
`Host` header: atelet bind-mounts a read-only, per-actor identity
directory at `/run/ate`, containing the file `actor-id`.

Per the discussion on #178, the mount is a *directory* rather than a
single file so that further late-bound data (more downward-API fields,
config, eventually credentials) can land as sibling files without
changing the mount shape.

### Why not an environment variable
An env var (or a file baked into the image) cannot carry per-actor
identity. Both live in the checkpointed process memory / overlay, which
is captured once in the golden snapshot. On the normal activation path
(`runsc restore` from that snapshot), every actor of a template would
report the golden actor's ID instead of its own. This was verified on a
live cluster: the env approach reports the golden ID after restore, while
this mount approach reports the correct per-actor ID.

### What changed (atelet)
- Populates a per-actor identity directory (currently just `actor-id`)
  and bind-mounts it read-only at `/run/ate` when assembling each app
  container's OCI bundle. The bundle is regenerated on every resume, and
  external bind mounts are re-attached from the restore-time
  `config.json`, so the value stays correct per-actor through a snapshot
  restore. The pause container gets no identity mount. The directory is
  cleared and recreated with the other per-actor dirs, and `actor-id` is
  written atomically (temp file + fsync + rename + parent fsync) since
  the directory is visible to a possibly-running actor.
- The mount point is created via `os.Root`, so a symlink planted in the
  image cannot redirect the write outside the extracted rootfs.
- Path-input validation at the RPC boundary comes from #206 (merged);
  the copy of it this branch previously carried was deduplicated in the
  rebase, so this PR is now purely the identity feature.

### Testing
- Unit tests: identity mount presence/absence, symlink-escape refusal,
  atomic-write semantics (overwrite on resume, no leftover temp files),
  and the router-client helpers.
- E2E (`internal/e2e/suites/identity`): restores two actors from one
  golden snapshot and asserts each observes its own ID and not the golden
  ID, which is the one property unit tests and `config.json` inspection
  cannot catch. It introduces a minimal `probe` actor fixture (`/whoami`)
  and a router HTTP client for the harness, and passes against a live kind
  cluster on current main.

### Usage (for actor authors)
Read `/run/ate/actor-id` fresh on use rather than caching it at startup.
The file holds the raw actor ID with no trailing newline. Documented in
`docs/api-guide.md`.

- [x] Tests pass

### DIFF
--- a/cmd/atelet/oci.go
+++ b/cmd/atelet/oci.go
@@ -34,7 +34,24 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 )
 
-func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryPullCache, actorTemplateNamespace, actorTemplateName, actorID, containerName, ref string, args []string, env []string, annotations map[string]string, netns string) error {
+const (
+	// IdentityMountPath is the in-actor directory at which atelet bind-mounts
+	// the actor's identity data. Workloads read the files inside it (at
+	// request time, not cached at startup) to learn about themselves. It is
+	// delivered as a per-actor bind mount rather than environment variables
+	// because env lives in the checkpointed process memory and would be
+	// frozen at the golden snapshot's values after a restore; a bind mount is
+	// re-attached per-actor on every resume. A directory (rather than a
+	// single-file mount) so further identity data can be added without
+	// changing the mount shape.
+	IdentityMountPath = "/run/ate"
+
+	// ActorIDFileName is the file inside IdentityMountPath holding the
+	// actor's own ID, raw with no trailing newline.
+	ActorIDFileName = "actor-id"
+)
+
+func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryPullCache, actorTemplateNamespace, actorTemplateName, actorID, containerName, ref string, args []string, env []string, annotations map[string]string, netns string, identityDir string) error {
 	tracer := otel.Tracer("prepareOCIDirectory")
 
 	ctx, span := tracer.Start(ctx, "prepareOCIDirectory")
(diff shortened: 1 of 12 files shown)


MushuEE:
Do we want to expose more to the actor? Like labels? Or do we want only immutable metadata?

cmd/atelet/oci.go
```diff
@@ -112,35 +185,7 @@ func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryP
 			Readonly: false,
```
MushuEE, line 194: For sandboxed workloads, is it recommended from a security perspective to have read-only file systems? @mtaufen ? Would we need to mount /run as a separate, writeable tmpfs volume in the OCI spec mounts, write the file into that tmpfs source directory on the host, and mount it into the guest if we harden this?

fkautz: I think this is a good idea, but perhaps should be in a separate PR? For this PR, the actor-id is its own single-file `ro` bind (`oci.go:134-141`), so `Root.Readonly: true` doesn't affect it.

I can open an issue if you want.

fkautz:
> Do we want to expose more to the actor? Like labels? Or do we want only immutable metadata?

My intention for the probe is to expose more information so we can test the environment. That'll be in future patches.

I'm not the right person to answer the broader question i.r.t. what should be exposed through the mount. We could migrate it to a directory and populate it.

fkautz:
Since the last review:

- **The mount is now the directory `/run/ate`** (read-only), with the ID in `actor-id` inside it, per the #178 discussion. This means additional metadata can be added as sibling files later without changing the mount shape. @MushuEE this is also the answer to your earlier question: labels or other immutable (or slow-changing) metadata could land here as new files, and what exactly gets exposed is being worked out in #178.
- Follow-ups from a second review pass: the identity directory is now cleared and recreated in `resetActorDirs` with the other per-actor dirs, and `actor-id` is written atomically (temp file + fsync + rename) since the directory is visible to a possibly-running actor.
- Rebased onto current main.

Verified end-to-end on a kind cluster: the identity e2e restores two actors from one golden snapshot and each reads its own ID (not the golden's) from `/run/ate/actor-id`. The workflow run needs a maintainer approval for CI to confirm.

cmd/atelet/oci.go
mtaufen: Read-only root filesystem is usually a good idea.

fkautz:
rebased onto main and squashed to two commits (feature, e2e verification). nothing functionally new since the last review except review feedback; notable diffs from what was previously pushed:

- the DNS-1123 validation this branch carried was deduplicated into #206, which merged; this PR is now purely the identity feature
- actor-id is written atomically and the identity dir is reset with the other per-actor dirs
- probe deploy no longer shells out, and the probe reports identity read errors so e2e failures explain themselves

re-verified after the rebase: unit tests, make verify/test, and the kind e2e identity suite (two actors from one golden snapshot each observing their own id) against current main.

cmd/atelet/oci.go
fkautz: filed #235 to track this as an open question. one detail from my earlier reply has since changed shape: the identity mount is now a read-only *directory* at /run/ate rather than a single file, but the conclusion holds, it's independent of rootfs writability either way.

MushuEE: Thanks for opening that 🙏

ahmedtd:
I think we can go ahead and merge this.  I'm currently working on the actor token/cert injection support, which will probably make some changes to this code:

* Make the location configurable in the ActorTemplate
* Use https://pkg.go.dev/k8s.io/kubernetes/pkg/volume/util#AtomicWriter to handle atomically updating the whole directory
* Refresh credentials as they approach expiration.

But this is a good base to build on.  Thanks for sending it.

PR #187, opened 2026-06-05
Add goimports to .golangci.yaml and fix lint errors

troychiu:
Add goimports to `.golangci.yaml` to follow Go import organization best practice and fix corresponding lint error.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -107,6 +107,13 @@ linters:
         linters: [staticcheck]
         text: '^ST1019'
 
+formatters:
+  enable:
+    - goimports
+  settings:
+    goimports:
+      local-prefixes: github.com/agent-substrate/substrate
+
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
--- a/cmd/ateapi/internal/controlapi/create_actor.go
+++ b/cmd/ateapi/internal/controlapi/create_actor.go
@@ -19,12 +19,13 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
-	"github.com/agent-substrate/substrate/internal/resources"
-	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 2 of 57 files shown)


ameukam:
LGTM
cc @BenTheElder @juli4n

BenTheElder (commented):
Lint errors that require manually fixing but are in fact auto-formatted are ... pretty annoying, it just forces people to manually do what could've been automated. 

Can we consider goimports in the gofmt flow instead?

cc @thockin

troychiu:
> Lint errors that require manually fixing but are in fact auto-formatted are ... pretty annoying, it just forces people to manually do what could've been automated.

I'm not sure if I fully understood your point, but using `golangci-lint run --fix` formats them automatically.

BenTheElder:
> I'm not sure if I fully understood your point, but using golangci-lint run --fix formats them automatically.

Hmm, does that change anything else?

We have `make update` which does gofmt etc. I'd want it to happen when I call this, but it also seems semi-redundant to gofmt (can we dedupe?)

thockin:
There was a reason I did gofmt and not goimports, but I can't recall it.

I do not like the idea of --fix being the norm because I have frequently disagreed with the linters.

Is the goimports linters JUST imports or is it all of gofmt, too.

The advantage of update/gofmt.sh is that it does exactly one thing.  I don't care if that calls goimports instead (modulo the thing I can't remember)

.golangci.yaml
```diff
@@ -107,6 +107,13 @@ linters:
         linters: [staticcheck]
         text: '^ST1019'
 
+formatters:
+  enable:
+    - goimports
+  settings:
+    goimports:
+      local-prefixes: github.com/agent-substrate/substrate
```
thockin, line 115: The problem I have with this is that I don't know how to configure my editor (neovim) to get a per-repo configuration for this.  AFAICT I either configure it globally (so all prefixes are "local" in all projects) or I am forced to call something to fix it.

What I really want is that when I save in editor it runs goimports (which works today) with THIS REPO's `-local` flags.  I don't know how to do that.

What I DON'T WANT is for my editor to save code one way, then I send a PR, then I have to run ./hack/update/goimports on every PR.

thockin:
What's next with this PR?

PR #188, opened 2026-06-05
Add Skill to validate k8s CRDs.

juli4n:
Include a verbatim copy of the k8s conventions into the skill. I tried the skill locally and it generated a pretty decent summary of findings. I hope this will help to catch violations to the KRM style guide as we evolve our k8s APIs.

### DIFF
--- a/.agents/skills/review-crds/SKILL.md
+++ b/.agents/skills/review-crds/SKILL.md
@@ -1,0 +1,35 @@
+---
+name: review-crds
+description: Reviews CRDs for compliance with Kubernetes API conventions.
+---
+
+# Kubernetes API Conventions Skill
+
+## Task
+
+Review any CRDs in Agent Substrate to ensure they follow the conventions in `references/api-conventions.md`.
+
+## Output Format
+
+Your output should be a list of violations conforming to the following JSON format:
+
+```json
+[
+  {
+    "location": "",
+    "rule": "",
+    "fix": ""
+  }
+]
+```
+
+- location: file path and line range that violated the rule, formatted like this: `path/to/file/under/review:1-10`. The path is relative to the repo root.
+- rule: Exact quote of violated rule from `references/api-conventions.md`. If there is no rule, this should be set to the string `OPINION`.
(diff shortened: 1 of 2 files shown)


BenTheElder (commented):
see also: https://github.com/kubernetes-sigs/kube-api-linter

mtaufen:
We can add both, it's good for the agents to have an understanding of the conventions while working.

.agents/skills/review-crds/SKILL.md
```diff
@@ -0,0 +1,23 @@
+---
+name: review-crds
+description: Guides the agent to follow Kubernetes API conventions for OSS standards.
```
mtaufen: nit: "Reviews CRDs for compliance with Kubernetes API conventions" would be more tool-centric language. Whether this guides the agent depends on if the agent is triggered to open the skill.

.agents/skills/review-crds/SKILL.md
```diff
@@ -0,0 +1,23 @@
...
+---
+
+# Kubernetes API Conventions Skill
+
+## Purpose
```
mtaufen: Purpose may be best left to the description in frontmatter as it's somewhat redundant. Since we want a structured response, recommend asking for JSON. We could format the SKILL something like this:

## Task

Review any CRDs in Agent Substrate to ensure they follow the conventions in `references/api-conventions.md`. 

## Output Format

Your output should be a list of violations conforming to the following JSON format:

```json
[ 
  {
    "location":"",
    "rule": "",
    "fix": ""
  }
]
```

- location: file path and line range that violated the rule, formatted like this: `path/to/file/under/review:1-10`. The path is relative to the repo root.
- rule: Exact quote of violated rule from references/api-conventions.md. If there is no rule, this should be set to the string `OPINION`.
- fix: Your suggestion of how to fix the issue.

## Hints
- Most CRDs are defined in `pkg/api/v1alpha1`.

juli4n: I like it. Done.

mtaufen (approved):

PR #189, opened 2026-06-05
feat(actortemplate): add per-template GoldenSnapshotWait

carlossg:
Adds an optional `goldenSnapshotWait` field to `ActorTemplateSpec` so each workload can control how long the controller waits between resuming the golden actor and taking its snapshot. Slow-booting workloads (e.g. Chromium under gVisor) need more than the 20-second default before their memory state is consistent enough to checkpoint; with this field they can set their own wait without forcing every template in the cluster to use the slowest boot budget.

When unset, the controller falls back to the existing 20-second default.

## Changes

- `ActorTemplateSpec.GoldenSnapshotWait *metav1.Duration` (optional)
- Controller uses spec value when set, otherwise `defaultGoldenSnapshotWait = 20s`
- CRD manifest + deepcopy regenerated via `go generate`

- [x] Tests pass


EItanya (commented):
Looks great, I was running into this issue as well

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -96,6 +96,13 @@ type ActorTemplateSpec struct {
 	//
 	// +required
 	Runsc RunscConfig `json:"runsc,omitempty"`
+
+	// GoldenSnapshotWait is how long to wait after the golden actor resumes
+	// before taking the snapshot. Slow-booting workloads
+	// may need longer than the 20-second default.
+	//
+	// +optional
+	GoldenSnapshotWait *metav1.Duration `json:"goldenSnapshotWait,omitempty"`
```
bowei, line 105: Should this go inside SnapshotConfig

snapshotConfig.goldenSnapshotTimeout

carlossg: as a new section? I can do that

thockin:
Should we be using k8s probes or something for this?

carlossg:
already handled in #487

PR #190, opened 2026-06-05
Switch ateom gVisor state path

EItanya:
## Summary

  Switch the shared atelet/ateom gVisor state directory from `/run/ateom-gvisor` to `/var/lib/ateom-
  gvisor`.

  ## Why

  This path is not just a transient socket location. It is the shared host/container state root for
  downloaded `runsc` binaries, per-ateom sockets, OCI bundles, runsc state, pidfiles, checkpoint state,
  and restore state.

  `/run` is normally volatile runtime storage and may be tmpfs-backed, which makes it a poor fit for
  large or service-owned local state. Moving this under `/var/lib` better matches Linux filesystem
  conventions and avoids treating gVisor runtime/checkpoint artifacts as ephemeral runtime-only files.

Additionally I was running into many permissions issues with apps trying to use various folders mounted under `/root` before this change.

  ## Changes

  - Update `ateompath.BasePath` to `/var/lib/ateom-gvisor`.
  - Update the WorkerPool controller to use `ateompath.BasePath` instead of hard-coding the mount path.
  - Update the atelet install manifest to mount the new hostPath.

  ## Impact

  This keeps `atelet`, `ateom-gvisor`, and dynamically created WorkerPool Deployments aligned on the
  same shared state path. Without that alignment, the system can fail to find sockets, bundles, runsc
  state, or checkpoint/restore files.

### DIFF
--- a/internal/ateompath/ateompath.go
+++ b/internal/ateompath/ateompath.go
@@ -22,7 +22,7 @@ import (
 const (
 	// The base path.  This is both the path of the root shared folder on the
 	// host filesystem, and when it is mounted into ateom and atelet containers.
-	BasePath = "/run/ateom-gvisor"
+	BasePath = "/var/lib/ateom-gvisor"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


ahmedtd (approved):

BenTheElder (approved):
Thanks!

PR #191, opened 2026-06-05
Update the variable name used in logging to be actual Actor Template Namespace

HavenXia:
It was using a wrong name `Actor Namespace` before, which doesn't exist yet.

- [x] Tests pass

### DIFF
--- a/cmd/ateom-gvisor/internal/ateom/logger.go
+++ b/cmd/ateom-gvisor/internal/ateom/logger.go
@@ -62,14 +62,14 @@ func NewActorLogger(w io.Writer, isOnGCE bool) *ActorLogger {
 }
 
 // EmitLifecycleLog logs a synthetic actor lifecycle event.
-func (al *ActorLogger) EmitLifecycleLog(msg, actorID, actorTemplate, actorNamespace string) {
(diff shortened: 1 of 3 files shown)


cmd/ateom-gvisor/internal/ateom/logger.go
```diff
@@ -62,14 +62,14 @@ func NewActorLogger(w io.Writer, isOnGCE bool) *ActorLogger {
...
+func (al *ActorLogger) EmitLifecycleLog(msg, actorID, actorTemplate, actorTemplateNamespace string) {
 	envelope := map[string]any{
 		"time":    time.Now().Format(time.RFC3339Nano),
 		"message": msg,
 		al.labelsKey: map[string]string{
-			"ate.dev/actor_id":        actorID,
-			"ate.dev/actor_template":  actorTemplate,
-			"ate.dev/actor_namespace": actorNamespace,
+			"ate.dev/actor_id":                 actorID,
+			"ate.dev/actor_template":           actorTemplate,
```
juli4n: Should we call this ate.dev/actor_template_name to be consistent with namespace?

HavenXia: Gotcha, updated.

juli4n (approved):

Issue #103, opened 2026-05-28
Add log sanitization


mtaufen:
We already noticed one place where secrets could leak in logs: https://github.com/agent-substrate/substrate/pull/20/changes#r3313203264. We should add a sanitizer based on field tagging for printing structs/protos and basic regexp redaction for common credential formats as a backup.

juli4n:
Protobuf has `debug_redact = true` to hint that a field can have sensitive content (e.g. key material). IIUC we are slog-ging the proto message directly, which will traverse the proto struct using reflection to convert it into a json-value. 

How about annotating all sensitive fields in our services with `debug_redact` and then implementing a small utility function to redact fields with this annotation? I couldn't find official protobuf/go support for this, but something like https://github.com/ntk221/redact

Then we could instrument `internal/contextlogging` which defines the slog handler we use in all servers to sanitize the value if it's a protobuf Message.

mtaufen:
From what I remember, there were also some ideas in
https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1753-logs-sanitization,
it was implemented in alpha but later deprecated and removed (I think due
to performance concerns?):
https://groups.google.com/g/kubernetes-dev/c/xhQuwdd2Smw/m/L_EyHKbDAAAJ?e=48417069,
in favor of https://github.com/kubernetes/enhancements/issues/1933 (
https://github.com/kubernetes/enhancements/tree/master/keps/sig-security/1933-secret-logging-static-analysis)
(static analysis to prevent secret leaks).

PR #197, opened 2026-06-07
fix(router): stop logging secrets in ext_proc request logs

konippi:
Part of #103.

Stops the router (ext_proc) request path from writing secrets to logs.

This PR was prepared in part with the assistance of generative AI.

- [x] Tests pass

### DIFF
--- a/cmd/atenet/internal/router/extproc.go
+++ b/cmd/atenet/internal/router/extproc.go
@@ -108,7 +108,7 @@ func (s *ExtProcServer) Process(stream extprocv3.ExternalProcessor_ProcessServer
 		default:
 			// No modification for other processing states, but log because this should
 			// not be called.
-			slog.Error("Unexpected request type", slog.Any("reqType", reqType))
+			slog.Error("Unexpected request type", slog.String("reqType", fmt.Sprintf("%T", reqType)))
 			resp.Response = &extprocv3.ProcessingResponse_RequestHeaders{
 				RequestHeaders: &extprocv3.HeadersResponse{
 					Response: &extprocv3.CommonResponse{},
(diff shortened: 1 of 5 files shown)


bowei:
The ext_proc changes look ok. If you split that into its own PR, I can review and merge that.

konippi:
@bowei
Thanks! Scoped this down to just the ext_proc changes.

The other two pieces are split out:
- KubernetesClaims slog allowlist → its own PR (#224).
- gRPC interceptor payload logging → dropped, since #20 already redacts `env` via sanitizeForLog.

cmd/atenet/internal/app/router/extproc_in.go
```diff
@@ -47,7 +47,7 @@ func newRequestMetadata(headers []*corev3.HeaderValue) *requestMetadata {
 
 		headersMap[k] = val
 		if k == ":path" {
-			path = val
+			path, _, _ = strings.Cut(val, "?")
```
bowei: Do you want to use url.Parse to avoid potential edge cases?

konippi: I tried both. For `:path`, cutting at the first `?` removes exactly the query, which is the secret vector here.

`url.Parse` doesn't improve that for this use and adds downsides:
- It errors on e.g. control chars, so it needs a fallback — and a fallback that logs the raw value would leak the query. `strings.Cut` is total, so there's no error path that can fall through to logging the secret.
- It reinterprets `//host/path` as authority and silently drops it.

Given that, my inclination would be to keep `strings.Cut` — but I'm very open to using `url.Parse` if you'd prefer. Let me know what you think.

bowei: Actually, do we need this change?

I think interpreting the `path:` coming from Envoy would add some additional non-obvious semantics to this code. If we maintain that we pass the Envoy argument through unchanged, this will be much simpler to explain how this works.

bowei: Is there a security issue that you are trying to fix?

bowei (commented):
One comment on the path change. Otherwise looks ok

cmd/atenet/internal/app/router/extproc_in.go
konippi: Yes — the `:path` query string can carry secrets (tokens, OAuth codes, etc.; CWE-598), and `m.path` gets persisted into the statusz QueryRecorder, which is served unauthenticated. That's the leak I'm closing.

You're right that the parser shouldn't reinterpret `:path`. I've reverted `newRequestMetadata` to pass Envoy's `:path` through verbatim and moved the query-stripping to the sink (`redactPath` in `AddRouterRequest`) — same "redact at the sink" approach as `sanitizeForLog`.

thockin:
ping to @bowei

bowei (approved):

Issue #131, opened 2026-05-31
Need a glossary


thockin:
We have a lot of Jargon emerging and we need to define terms clearly, and consider whether we like those terms or need to change them.  In particular (top of my head):

Actor (I think we need a better word, personally.  Some older docs (internal) still say "session")

Ateom (I dislike it because it is hard to type, hard to say, and autocorrect keeps changing it)

Atelet (I like it but autocorrect keeps giving me "platelet")

BenTheElder:
Renaming things, and documenting things seems like two distinct issues.

Can we split out any rename proposals from "we should document the names we have"?

I think we have some of this in https://github.com/agent-substrate/substrate/blob/main/docs/architecture.md currently

npolshakova:
This is slightly adjacent to the glossary ask, but I think it would also be useful to have one or two diagrams that show how the major terms relate to each other.

The architecture README is already helpful, and the `atenet` README has a good overview of the networking pieces, but I found myself wanting a single diagram that connects the core concepts: `ActorTemplate`, `Actor`, `WorkerPool`, `Worker`, `ate-api`, `ate-net`, and `ateom`.

For example, maybe something like this that I’ve been using to mentally sketch out the components:

```mermaid
flowchart TD
  REQ["Request to<br/><code>&lt;actor&gt;.actors.resources.substrate.ate.dev</code>"]

  subgraph K8s["Kubernetes objects"]
    direction TB

    WP["WorkerPool CRD<br/>desired warm capacity"]
    DEP["Deployment"]
    PODS["Warm Worker Pods<br/>ateom runtime"]

    DNS["ate-net dns"]
    ROUTER["ate-net router<br/>(Envoy + ext_proc)"]

    WP -->|"reconciled into"| DEP
    DEP -->|"starts"| PODS

    DNS -->|"resolves actor DNS to router"| ROUTER
    ROUTER -->|"proxy request to actor pod IP"| PODS
  end

  subgraph Substrate["Substrate control plane / state store"]
    direction TB

    AT["ActorTemplate CRD<br/>actor image + config"]
    ACT["Actor record<br/>status + snapshot refs"]
    W["Worker record<br/>idle / busy + pod IP"]
    API["ate-api<br/>scheduler + lifecycle API"]

    API -->|"CreateActor"| ACT
    ACT -->|"template"| AT
    API -->|"pick / claim worker"| W
  end

  STORE["Snapshot storage<br/>GCS/S3"]

[...]

PR #200, opened 2026-06-08
docs: add a glossary of core terms

fkautz:
Addresses #131

Adds `docs/glossary.md`: a single place that defines the terms used across
Agent Substrate and shows how they relate.

### What it covers
- Resources (ActorTemplate, WorkerPool), control-plane records (Actor, Worker),
  components (ate-api-server, atecontroller, atelet, ateom, atenet,
  podcertcontroller, kubectl-ate), runtime/sandbox concepts (gVisor, runsc,
  pause container, checkpoint/restore), snapshots (golden, last, storage), and
  the Uniform DNS Mesh.
- Three UML diagrams: a sequence diagram (activation/resume flow), a class
  diagram (resource/ownership model, grouped into Kubernetes-object and
  control-plane-record packages), and a state-machine diagram (actor lifecycle).
- Linked from the README documentation section.

### Scope
Per the issue thread, this documents the names as they exist in the codebase;
rename proposals are intentionally left to a separate discussion, so this PR is
documentation-only (not a renaming proposal).

### Accuracy
Every term and diagram was checked against the code and existing docs. Project
terminology and casing are reused where things are already named (for example
"Uniform DNS Mesh", the atelet "Herder", and the "interior gVisor" coordinator),
and component responsibilities, RPC names, relationships, and the actor status
lifecycle map to their sources.

- [x] Tests pass (documentation-only change)

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -1,105 +1,16 @@
 # Agent Substrate Glossary
 
-This document defines the core terms used across Agent Substrate and shows how
-they relate to each other. It documents the names as they exist in the codebase;
-proposals to rename any of these terms are tracked separately (see issue #131).
+This document defines the core terms used across Agent Substrate.
 
-For deeper detail, see the [Architecture](architecture.md) and
+For how the pieces fit together, see the [Architecture](architecture.md) and
 [API Guide](api-guide.md).
-
-## How the pieces fit together
-
-**Activation flow (UML sequence diagram).** A request reaches an actor through
-the networking stack, which resumes it onto a worker if it is suspended:
-
-```mermaid
-sequenceDiagram
-    actor Client
-    participant DNS as atenet DNS
-    participant Router as atenet router
-    participant API as ate-api-server
-    participant Atelet as atelet
-    participant Ateom as ateom
-    participant Store as snapshot storage
-
-    Client->>DNS: resolve actor DNS name
-    DNS-->>Client: router address
-    Client->>Router: HTTP request (Host = actor)
-    Router->>API: ResumeActor(actorID)
-    API->>Atelet: Restore
-    Store-->>Atelet: download snapshot
-    Atelet->>Ateom: RestoreWorkload
-    Note over Ateom: runsc restore
-    Ateom-->>Atelet: ready
-    Atelet-->>API: worker pod IP
-    API-->>Router: worker pod IP
-    Router->>Ateom: proxy request to worker pod
-    Ateom-->>Router: response
-    Router-->>Client: response
-    Note over API,Store: later, when idle: SuspendActor checkpoints back to storage and frees the worker
-```
-
-**Resource model (UML class diagram).** The CRDs and control-plane records
-behind that flow, with their relationships and multiplicities:
-
-```mermaid
-classDiagram
-    namespace KubernetesObjects {
-        class ActorTemplate {
-            <<CRD>>
-        }
-        class WorkerPool {
-            <<CRD>>
-        }
-        class Deployment
-        class WorkerPod {
-            ateom
-            runsc
-        }
-    }
-
-    namespace ControlPlaneRecords {
-        class Actor {
-            <<record>>
-            status
-            snapshotRefs
-        }
-        class Worker {
-            <<record>>
-            actorId
-            podIP
-        }
-    }
-
-    ActorTemplate "1" --> "1" WorkerPool : workerPoolRef
-    WorkerPool ..> Deployment : reconciled by atecontroller
-    Deployment "1" *-- "*" WorkerPod : manages
-    Actor ..> ActorTemplate : derived from
-    Actor "0..1" --> "0..1" Worker : runs on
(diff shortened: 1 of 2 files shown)


docs/glossary.md
```diff
@@ -0,0 +1,204 @@
+# Agent Substrate Glossary
+
+This document defines the core terms used across Agent Substrate and shows how
+they relate to each other. It documents the names as they exist in the codebase;
+proposals to rename any of these terms are tracked separately (see issue #131).
+
+For deeper detail, see the [Architecture](architecture.md) and
+[API Guide](api-guide.md).
+
+## How the pieces fit together
```
bowei: Bit of a nit, but my expectation for a glossary would be just a list of terms and their definitions. The addition of the sequence diagram and resource model feels like they belong somewhere else. 

This is mostly from the standpoint of discoverability when people are making modifications to the system as it's not immediately obvious that there will be diagrams in the glossary to change.

Suggest finding a different home for the diagrams?

docs/glossary.md
```diff
@@ -0,0 +1,204 @@
...
+- **WorkerPool**: declares warm compute capacity, a fleet of pre-started worker
+  pods. It is reconciled into a Kubernetes `Deployment` by the
+  [atecontroller](#components).
+
+## Records (dynamic state, in the control-plane store)
+
+These are not Kubernetes objects; they live in the control-plane database
+(Valkey/Redis) because they change too frequently for etcd.
+
+- **Actor**: a single instance derived from an `ActorTemplate`, identified by a
```
bowei, line 25: We should be careful about where we are defining things. For a glossary, I would expect it to remain at a high level vs referencing things like enum states etc as this will be hard to maintain consistency across the entire codebase. In addition, having multiple sources of definitions may confuse the agent coding depending on which source they end up referencing.

Can we keep the glossary definitions to be summaries at the high level? More low level descriptions should be kept as close to the code definitions as possible (i.e. in docstrings) so there are less consistency issues.

This comment applies in general to the other entries...

docs/glossary.md
```diff
@@ -0,0 +1,204 @@
...
+  DNS-1123 actor ID. It is the unit that is suspended and resumed, and it moves
+  between workers over its lifetime. An Actor record tracks status
+  (`SUSPENDED` → `RESUMING` → `RUNNING` → `SUSPENDING` → `SUSPENDED`) and its
+  snapshot references. (Some older internal docs call this a "session".)
```
bowei: We should try to avoid reference older internal docs, at least not yet before our first real release!

docs/glossary.md
```diff
@@ -0,0 +1,204 @@
...
+  24h) via two signers: a service-DNS server identity and a pod/workload client
+  identity (equivalent to KSA tokens). Components use these as their TLS identity
+  to authenticate connections to one another (mutual TLS).
+
+- **kubectl-ate**: a `kubectl` plugin CLI for managing the Actor lifecycle
+  (create, resume, suspend, delete, logs) and listing Workers.
+
+## Runtime concepts
+
+- **gVisor**: the sandboxing technology Substrate uses to isolate and snapshot
```
bowei: No clear we need to define this?

docs/glossary.md
```diff
@@ -0,0 +1,204 @@
...
+
+- **podcertcontroller**: issues short-lived pod certificates (lifetime capped at
+  24h) via two signers: a service-DNS server identity and a pod/workload client
+  identity (equivalent to KSA tokens). Components use these as their TLS identity
+  to authenticate connections to one another (mutual TLS).
+
+- **kubectl-ate**: a `kubectl` plugin CLI for managing the Actor lifecycle
+  (create, resume, suspend, delete, logs) and listing Workers.
+
+## Runtime concepts
```
bowei: I don't know if these terms need a special definition...

We should focus the glossary on terms that are unique to Substrate and will need clarification for people external to the project.

docs/glossary.md
fkautz: Agreed. Pulled the diagrams out in `4dce2ee` so the glossary is just terms now. I'll relocate them to the architecture docs in a follow-up rather than dropping them.

docs/glossary.md
fkautz: Makes sense. Cut the state sequence and RPC/enum detail throughout, leaving short summaries. Filed #236 to move that detail into docstrings near the code. Applied your "in general" note to the other entries too.

docs/glossary.md
fkautz: Removed. The "session" mention (and the older-docs reference) is gone from the intro entirely.

docs/glossary.md
fkautz: Agreed on both. Cut gVisor, runsc, and "pause container" since they're external tech, not Substrate vocabulary. The one relevant fact, that suspend/resume builds on gVisor's checkpoint/restore, is now folded into the Suspend/Resume definitions.

thockin:
I am sure we will refine this over time, but LGTM

PR #201, opened 2026-06-08
Remove unused ports block from API

thockin:
Noticed this while digging around


juli4n (approved):

bowei:
Note: we may need this back soon, but let's remove this for now.

PR #205, opened 2026-06-08
Add more CRD validation

thockin:
It's not all done but the fields we are playing with now are mostly done.


pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -55,18 +64,29 @@ type Container struct {
 // literal values are not expanded with Kubernetes-style $(VAR) references,
 // envFrom is not supported, and valueFrom currently supports only secretKeyRef.
 //
-// +kubebuilder:validation:XValidation:rule="!(has(self.value) && has(self.valueFrom))",message="value and valueFrom are mutually exclusive"
+// +kubebuilder:validation:ExactlyOneOf={value, valueFrom}
```
juli4n, line 67: How does this interplay with `omitempty` in `Value`?  Can you add a test where we set `Env` to an empty string?

thockin: You are correct!

Kubectl works because it never serializes to a Go type, which is than builtin types here.  In CRD `optional` is satisfied by the key being present at all, whereas builtins require an actual value (since builtin validation happens on the deserialized data).

This means we have to change Value to a pointer.  Will add that here.

thockin:
@EItanya specifically for the `Value` change.  I missed that in review!

EItanya:
> Eitan Yarmush (@EItanya) specifically for the `Value` change. I missed that in review!

Good catch, should we think about using the linter: https://github.com/kubernetes-sigs/kube-api-linter

juli4n (approved):

juli4n:
> > Eitan Yarmush (Eitan Yarmush (@EItanya)) specifically for the `Value` change. I missed that in review!
> 
> Good catch, should we think about using the linter: https://github.com/kubernetes-sigs/kube-api-linter

This came up a few days ago as well in another PR. Filed https://github.com/agent-substrate/substrate/issues/207 to track.

PR #206, opened 2026-06-09
fix(atelet): validate RPC inputs that become host filesystem paths

### LINKED ISSUE #202: Validate atelet RPC inputs that become host filesystem paths
## Summary

atelet builds host filesystem paths from several `Run`/`Restore`/`Checkpoint` request fields without validating them. Since atelet listens on an insecure hostPort (bypassing the API server), any reachable caller can smuggle a path separator or `..` through these fields and make atelet read / `RemoveAll` / write outside the intended actor tree, or collide bundles.

## Unvalidated fields

1. **`target_ateom_uid`** → `/run/netns/ateom:<uid>` and `.../ateoms/<uid>/ateom.sock`. Separators or `..` redirect the netns/socket paths.
   - `internal/ateompath/ateompath.go:45,56` (`AteomSocketPath`, `AteomNetNSPath`)

2. **`spec.containers[].name`** → `OCIBundlePath(ns, tmpl, id, name)`, where atelet runs `os.RemoveAll` / `os.MkdirAll` / writes `config.json`. `/` or `..` escapes the bundle dir.
   - `pkg/api/v1alpha1/actortemplate_types.go:43` (no constraint on `Container.Name`)
   - `internal/ateompath/ateompath.go:96` (`OCIBundlePath`)

3. **Reserved / duplicate container names** — `prepareOCIBundles` builds the `pause` bundle and app bundles (keyed by name) concurrently. An app container named `pause` writes the same path as the sandbox-infra container (nondeterministic race / corruption); two app containers with the same name clobber each other.
   - `cmd/atelet/main.go` (`prepareOCIBundles`), `cmd/ateom-gvisor/main.go:459` (hard-coded `pause`)

## Suggested fix

Validate all path-constructing inputs at the three RPC boundaries before any path is built:
- `target_ateom_uid` and each container name as DNS-1123 labels (UUIDs and normal names qualify; separators/`..` rejected).
- Reject the reserved name `pause` and duplicate container names.
- Optional defense-in-depth: `+kubebuilder:validation:Pattern` on `Container.Name`.

> Note: PR #184 already validates namespace/template/actor-id this way; this extends the same mechanism to the remaining path-constructing fields.

fkautz:
Fixes #202

atelet's gRPC server runs on an unauthenticated hostPort
(`manifests/ate-install/atelet.yaml`), so its `Run`/`Restore`/`Checkpoint`
RPCs must treat every request field as untrusted. Several of those fields
are turned into host filesystem paths and then deleted, created, written,
or executed, none of them validated first. A caller that can reach the
port can smuggle a path separator or `..` through them and make atelet
operate outside the intended actor tree.

### Unvalidated inputs (each an independent escape route)
- **`actor_template_namespace` / `name` / `id`** are concatenated by
  `ateompath.ActorPath` into the per-actor dir that `resetActorDirs` runs
  `os.RemoveAll` / `os.MkdirAll` on.
- **`target_ateom_uid`** builds the netns path (`/run/netns/ateom:<uid>`)
  and the ateom control socket (`.../ateoms/<uid>/ateom.sock`).
- **`spec.containers[].name`** becomes an OCI bundle path component
  (`os.RemoveAll` / `MkdirAll` / `config.json` write).
- **`runsc.sha256_hash`** builds `static-files/runsc-<hash>`. The cache-hit
  `os.Stat` early return hands that path to ateom to **execute** before the
  existing hex check runs, so an escaped path could execute an arbitrary
  on-disk binary as the sandbox runtime.

### Where the validators live
The field rules are exported from **`internal/resources`**
(`ValidateActorRef`, `ValidateAteomUID`, `ValidateContainerNames`,
`ValidateRunscHash`, `ValidateSnapshotURIPrefix`), alongside the existing
`ValidateActorID`. Every one of these values originates upstream in
user-supplied API/CRD data (the runsc hash and snapshot location come from
the `ActorTemplate` spec), so the API server and controller can apply the
same rules at admission time instead of failing at actor-activation time.
`ValidateContainerNames` takes names rather than the atelet proto so
`resources` stays proto-free. Unifying `ValidateActorID` with
`ValidateActorRef` is deferred: it changes ate-api-server's client-facing
error responses.

The per-RPC glue stays in `cmd/atelet`: each of `Run`/`Checkpoint`/`Restore`
calls its own typed validator (`validateRunRequest` etc.) as the first
statement at the boundary, before any path is constructed.

### What is validated
- namespace and actor ID as DNS-1123 labels, template as a DNS-1123
  subdomain, matching how these are named; additionally the combined
  `<ns>:<tmpl>:<id>` path component is capped at 255 bytes, since
  individually valid DNS names can join to 381 bytes, over common
  filesystems' filename limit;
- `target_ateom_uid` as a DNS-1123 label (pod UIDs are UUIDs, which
  qualify);
- each container name as a DNS-1123 label, additionally rejecting the
  reserved name `pause` (collides with the sandbox-infra bundle and races
  its concurrent writer) and duplicate names (same bundle path, mutual
  corruption);
- `runsc.sha256_hash` as exactly 64 hex chars, checked in `fetchRunsc`
  before the path is built and before the cache-hit early return;
- `snapshot_uri_prefix` (Checkpoint/Restore) as a parseable URI naming a
  bucket, with query/fragment/userinfo rejected: object names are appended
  by concatenation and the storage layer re-parses using only host and
  path, so those components would silently redirect the upload/download to
  a different object. The scheme is deliberately not restricted; that is
  storage-backend policy.

Validation failures return `codes.InvalidArgument` with the reason.
Previously the interceptor surfaced them as `codes.Internal` with the
message masked as "internal server error".

DNS-1123 is a deliberate fit. These values already *are* Kubernetes names,
so every legitimate value passes, while the charset forbids `/`, `.` (for
labels), and `..`, which is exactly what enables traversal. For a
cooperating caller the change is behavior-neutral: inputs newly rejected
previously failed later or misbehaved silently.

### Testing
- Table tests for each validator in `internal/resources`, targeting the
  exported functions (valid values, separators, `..`, empty, case, length,
  the label-vs-subdomain distinction, the combined-length cap, the
  `pause`/duplicate rules, and the URI component rejections).
- Per-request validator tests in `cmd/atelet` pin which rules each RPC
  enforces, one table per validator.
- `TestRPCBoundariesReject` exercises `Run`/`Checkpoint`/`Restore` directly
  and asserts a traversal value is rejected with `InvalidArgument`, which
  guards against future removal or reordering of the validation call.
- `TestFetchRunscRejectsBadHash` plants a real cache file at the path an
  invalid hash resolves to and asserts `fetchRunsc` still errors. It passes
  with the current code and fails if validation is moved after the
  `os.Stat` early return, so it proves the ordering rather than just that
  validation runs.

- [x] Tests pass
- [x] No documentation changes needed

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -645,14 +645,20 @@ func validateCheckpointRequest(req *ateletpb.CheckpointRequest) error {
 	if err := validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec()); err != nil {
 		return err
 	}
-	return validateSnapshotURIPrefix(req.GetSnapshotUriPrefix())
+	if err := validateSnapshotURIPrefix(req.GetSnapshotUriPrefix()); err != nil {
+		return err
+	}
+	return nil
 }
 
 func validateRestoreRequest(req *ateletpb.RestoreRequest) error {
 	if err := validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec()); err != nil {
 		return err
 	}
-	return validateSnapshotURIPrefix(req.GetSnapshotUriPrefix())
+	if err := validateSnapshotURIPrefix(req.GetSnapshotUriPrefix()); err != nil {
+		return err
+	}
+	return nil
 }
 
 // validateActorRequest is the shared core for the fields common to all three
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -282,7 +283,7 @@ func (s *AteomHerder) fetchRunsc(ctx context.Context, cfg *ateletpb.RunscConfig)
 }
 
 func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (*ateletpb.RunResponse, error) {
-	if err := validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec()); err != nil {
+	if err := validateRunRequest(req); err != nil {
 		return nil, err
 	}
 
@@ -359,7 +360,7 @@ func recordSnapshotSize(ctx context.Context, kind, path, atNamespace, atName str
 }
 
 func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRequest) (*ateletpb.CheckpointResponse, error) {
-	if err := validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec()); err != nil {
+	if err := validateCheckpointRequest(req); err != nil {
 		return nil, err
 	}
 
@@ -426,7 +427,7 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 }
 
 func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest) (*ateletpb.RestoreResponse, error) {
-	if err := validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec()); err != nil {
+	if err := validateRestoreRequest(req); err != nil {
 		return nil, err
 	}
 
@@ -629,11 +630,33 @@ func (d *AteomDialer) DialAteomPod(ctx context.Context, podUID string) (*grpc.Cl
 	return conn, nil
 }
 
-// validateActorRequest validates the request fields that atelet turns into host
-// filesystem paths. atelet listens on an insecure hostPort, so any reachable
-// caller could otherwise smuggle a path separator or ".." through them and make
-// atelet read/RemoveAll/write outside the intended directory tree, or collide
-// bundles. Validate at the RPC boundary, before any path is built.
+// validateRunRequest, validateCheckpointRequest, and validateRestoreRequest
+// validate everything in their request that atelet turns into host filesystem
+// paths, plus the request-specific fields. atelet listens on an insecure
+// hostPort, so any reachable caller could otherwise smuggle a path separator
+// or ".." through these fields and make atelet read/RemoveAll/write outside
+// the intended directory tree, or collide bundles. Each RPC validates at its
+// boundary, before any path is built.
+func validateRunRequest(req *ateletpb.RunRequest) error {
+	return validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec())
+}
+
+func validateCheckpointRequest(req *ateletpb.CheckpointRequest) error {
+	if err := validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec()); err != nil {
+		return err
+	}
+	return validateSnapshotURIPrefix(req.GetSnapshotUriPrefix())
+}
+
+func validateRestoreRequest(req *ateletpb.RestoreRequest) error {
+	if err := validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec()); err != nil {
+		return err
+	}
+	return validateSnapshotURIPrefix(req.GetSnapshotUriPrefix())
+}
+
+// validateActorRequest is the shared core for the fields common to all three
+// RPCs.
 func validateActorRequest(namespace, template, actorID, targetAteomUID string, spec *ateletpb.WorkloadSpec) error {
 	if err := validateActorRef(namespace, template, actorID); err != nil {
 		return err
@@ -642,6 +665,26 @@ func validateActorRequest(namespace, template, actorID, targetAteomUID string, s
 		return err
 	}
 	return validateContainers(spec)
+}
+
+// validateSnapshotURIPrefix ensures the checkpoint/restore snapshot location
+// is a well-formed URI with a bucket, so a bad prefix fails fast at the RPC
+// boundary instead of deep inside an object-storage call. It deliberately
+// does not restrict the scheme: the storage layer (ategcs.parseGCSURL) only
+// uses the host (bucket) and path, and which schemes are acceptable is a
+// storage-backend policy, not a per-RPC one. The local paths used for
+// snapshot upload/download are derived from the separately validated actor
+// ref, not from this URI, so this is a sanity check rather than a
+// path-traversal guard.
+func validateSnapshotURIPrefix(prefix string) error {
+	u, err := url.Parse(prefix)
+	if err != nil {
+		return fmt.Errorf("invalid snapshot URI prefix %q: %v", prefix, err)
+	}
+	if u.Host == "" {
+		return fmt.Errorf("invalid snapshot URI prefix %q: missing bucket", prefix)
+	}
+	return nil
 }
 
 // validateActorRef ensures every component of the per-actor directory tree is a
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -52,8 +52,10 @@ import (
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc"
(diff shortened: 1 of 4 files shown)


cmd/atelet/main.go
```diff
@@ -611,6 +629,96 @@ func (d *AteomDialer) DialAteomPod(ctx context.Context, podUID string) (*grpc.Cl
...
+}
+
+// validateActorRef ensures every component of the per-actor directory tree is a
+// valid DNS-1123 name. namespace+template+actorID are concatenated by
+// ateompath.ActorPath into a host path on which atelet runs os.RemoveAll and
+// os.MkdirAll, so all three must be validated. Checking only one would still
+// leave a traversal window via the others. Template names are DNS-1123
+// subdomains (dots allowed); namespaces and actor IDs are labels.
+func validateActorRef(namespace, template, actorID string) error {
+	for _, f := range []struct {
```
dberkov: nit: could you pleas rewrite it in format like 
if errs := validation.IsDNS1123Label(namespace); len(errs) > 0 {
  return fmt.Errorf(...)
}

it will be much easy to read the code.

cmd/atelet/main.go
```diff
@@ -611,6 +629,96 @@ func (d *AteomDialer) DialAteomPod(ctx context.Context, podUID string) (*grpc.Cl
...
+// before it is used to build the on-disk binary path (static-files/runsc-<hash>)
+// and, on a cache hit, returned for ateom to execute. Without this, a hash
+// containing path separators or ".." could point the cache-hit early return
+// (and the download target) at an arbitrary binary outside the static-files dir.
+func validateRunscHash(sha256Hash string) error {
+	if len(sha256Hash) != 64 {
+		return fmt.Errorf("invalid runsc sha256 hash: want 64 hex chars, got %d", len(sha256Hash))
+	}
+	for _, r := range sha256Hash {
+		isHex := (r >= '0' && r <= '9') || (r >= 'a' && r <= 'f') || (r >= 'A' && r <= 'F')
```
dberkov: nit: have you considered to use regex instead?

cmd/atelet/main.go
```diff
@@ -276,6 +282,10 @@ func (s *AteomHerder) fetchRunsc(ctx context.Context, cfg *ateletpb.RunscConfig)
 }
 
 func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (*ateletpb.RunResponse, error) {
+	if err := validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec()); err != nil {
```
dberkov: nit: could you please write validation method per request type, like validateRunRequest(). Each request has non common parameters that needs to be validated.

cmd/atelet/main_test.go
```diff
@@ -0,0 +1,236 @@
+// Copyright 2026 Google LLC
```
dberkov, line 1: thanks for writing unit tests for the new functions. Ideally we not supposed to write unit tests for private methods. How complicated to write unit tests for public methods instead? 
You can keep the unit tests of public method to concentrate on the params validations only in this PR, however it will be a good start for adding more unit tests later.

cmd/atelet/main.go
fkautz: done in 936ee6b, unrolled into three plain if blocks.

cmd/atelet/main.go
fkautz: went with `hex.DecodeString` instead (936ee6b). it's the same decoder the digest comparison in fetchRunsc already uses.

cmd/atelet/main.go
fkautz: done in e81ba71. each RPC now has a typed validator (validateRunRequest / validateCheckpointRequest / validateRestoreRequest) wrapping a shared core for the common fields. the checkpoint/restore ones also validate the one non-common field, snapshot_uri_prefix: it must parse and name a bucket. deliberately not restricting the scheme since that's storage-backend policy, parseGCSURL only consumes host and path.

cmd/atelet/main_test.go
fkautz: not a problem, my preference is to test both external and internal APIs since they are both contracts (turtles all the way down?). TestRPCBoundariesReject does test the package's publicly exported entry points (all three RPC handlers), and TestFetchRunscRejectsBadHash checks validation ordering on an internal method, but agree that it should be more exhaustive. Happy to take that on in a different PR.

One observation, we can't fully validate the happy path of the RPC without mocking the full downstream stack (object storage, actor dirs, image pulls, a live ateom), which is why there are combinatorial tables on the helpers.

While double checking this I also found we have two implementations of the actor-id rule: ate-api-server uses a regex (`resources.ValidateActorID`) and this PR uses k8s.io/apimachinery's DNS-1123 validators. I'll refactor this PR to move the new validators into `internal/resources` and export them, so they have a public contract and the table tests target it. Unifying `ValidateActorID` with these is better as a follow-up since it touches ate-api-server's client-facing error responses.

Thoughts on this as an approach? If it looks good I'll push that refactor shortly and open issues for the deferred work.

cmd/atelet/main.go
```diff
@@ -629,11 +630,33 @@ func (d *AteomDialer) DialAteomPod(ctx context.Context, podUID string) (*grpc.Cl
...
+// boundary, before any path is built.
+func validateRunRequest(req *ateletpb.RunRequest) error {
+	return validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec())
+}
+
+func validateCheckpointRequest(req *ateletpb.CheckpointRequest) error {
+	if err := validateActorRequest(req.GetActorTemplateNamespace(), req.GetActorTemplateName(), req.GetActorId(), req.GetTargetAteomUid(), req.GetSpec()); err != nil {
+		return err
+	}
+	return validateSnapshotURIPrefix(req.GetSnapshotUriPrefix())
```
dberkov: nit:
if err := validateSnapshotURIPrefix(req.GetSnapshotUriPrefix()); err !=  nil {
  return err
}

return nil.

It will allow more cleaner way to  add more validation later. Similar comment for validateRestoreRequest()

cmd/atelet/main_test.go
dberkov: could you please create an issue so we can track it.

cmd/atelet/main_test.go
```diff
@@ -169,6 +169,75 @@ func TestValidateActorRequest(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+// TestPerRequestValidators pins the request-specific rules: Checkpoint and
+// Restore must reject a bad snapshot URI prefix even when every common field
+// is valid, and Run (which has no URI field) must accept the same common
+// fields.
+func TestPerRequestValidators(t *testing.T) {
```
dberkov: nit: could you please split out to independent TestXXX methods. Lets assume it will be bug in both validateRunRequest & validateCheckpointRequest.

Entire test will fail and it will not be clear enough what validation has an issue, without reading the logs

cmd/atelet/main_test.go
```diff
@@ -169,6 +169,75 @@ func TestValidateActorRequest(t *testing.T) {
...
+	okSpec := &ateletpb.WorkloadSpec{Containers: []*ateletpb.Container{{Name: "worker"}}}
+
+	if err := validateRunRequest(&ateletpb.RunRequest{
+		ActorTemplateNamespace: okNS, ActorTemplateName: okTmpl, ActorId: okID,
+		TargetAteomUid: okUID, Spec: okSpec,
+	}); err != nil {
+		t.Errorf("validateRunRequest with valid fields: %v", err)
+	}
+
+	if err := validateCheckpointRequest(&ateletpb.CheckpointRequest{
```
dberkov: could you please use "could you please use tests := []struct {}" style for testing multiple test cases

dberkov (commented):
.

cmd/atelet/main.go
fkautz: done in 2cc5021, both validateCheckpointRequest and validateRestoreRequest.

cmd/atelet/main_test.go
fkautz: done in 2cc5021. one test per request validator (TestValidateRunRequest / TestValidateCheckpointRequest / TestValidateRestoreRequest), so in your scenario the output is two distinct top-level failures, each naming its validator. the subtest name carries the broken rule.

cmd/atelet/main_test.go
fkautz: done in 2cc5021, same shape across all three: a tests := []struct table where each row mutates one field of a shared valid-request builder.

cmd/atelet/main_test.go
fkautz: filed #216 for unifying ValidateActorID with the new validators and #217 for expanding test coverage at the RPC layer. the refactor from the thread above is now on the branch (34dae11): the validators live in internal/resources as exported functions and the table tests target that public contract.

juli4n: ValidateActorID intentionally avoids using the apimachinery validator, because we need to expose a validation function *and* expose a regex for valid actor IDs, so atenet can creating DNS configuration. I think there is value in having a single authoritative definition of what an actor id is, in a single place.

dberkov (approved):

PR #221, opened 2026-06-11
feat: Add request parking to the atenet router

### LINKED ISSUE #27: Router needs to park requests and wait for capacity
Currently if no worker is available, the router returns 500 immediately.

```
$ curl -X POST -H "Host: ctr6.actors.resources.substrate.ate.dev" -i http://localhost:8888/
content-type: 
content-length: 96
date: Wed, 20 May 2026 22:46:44 GMT
server: envoy

error resuming actor ctr6: rpc error: code = FailedPrecondition desc = no free workers available
```

omeryahud:
When a request targets a suspended actor, the router resumes it via the control plane before routing. A momentarily saturated worker pool makes `ResumeActor` return `FailedPrecondition` ("no free workers available"), which the router previously turned straight into a 503. In an oversubscribed system that shortage might pass quickly as another actor suspends and frees its worker, so failing fast was wasteful.

Park such requests instead: retry the resume on `FailedPrecondition` until the actor becomes routable or a bounded wait elapses, capped by a fixed-capacity admission lot that sheds excess load. On budget expiry the underlying capacity error is surfaced so the HTTP boundary maps it faithfully. singleflight still collapses concurrent waiters for the same actor into one resume RPC. Parking can be disabled to preserve the legacy fail-fast behavior.

Fixes #27

- [x] Tests pass


docs/request-parking.md
```diff
@@ -0,0 +1,96 @@
...
+
+Only transient capacity (`FailedPrecondition`) and concurrency (`Aborted`)
+conditions are parked. Errors that will not resolve by waiting are returned
+immediately, preserving prior semantics:
+
+| Resume result                         | Behavior                          |
+| ------------------------------------- | --------------------------------- |
+| `OK`                                  | Route to worker                   |
+| `Aborted` (concurrent resume)         | Retry (always)                    |
+| `FailedPrecondition` (no free worker) | **Park & retry** (when enabled)   |
```
thockin: FailedPrecondition can mean multiple things, I think we should think more carefully about what error code we want for this specific case.  NotFound?  ResourceExhausted?

docs/request-parking.md
```diff
@@ -0,0 +1,96 @@
...
+| `NotFound`                            | Fail fast → `404`                 |
+| `Unavailable`                         | Fail fast → `503`                 |
+| `DeadlineExceeded`                    | Fail fast → `504`                 |
+| `PermissionDenied` / `Unauthenticated`| Fail fast → `403` / `401`         |
+
+When parking is **disabled** (`--parking-enabled=false`), the router preserves
```
thockin: Is this different than just setting `--parking-max-parked=0` ?

cmd/atenet/internal/router/resumer.go
```diff
@@ -25,36 +27,109 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+// legacyResumeBudget is the total time the resumer spends retrying a resume when
+// request parking is disabled. It preserves the historical fail-fast-on-capacity
+// behavior (only concurrent-update conflicts are retried).
+const legacyResumeBudget = 15 * time.Second
```
thockin: I know it's not the scope of this PR, but we should not be carrying legacy yet :)

Also, the use of "Aborted" for concurrency is suspect...

thockin (commented):
Took a quick spin through this, emphasis on quick.

docs/request-parking.md
```diff
@@ -0,0 +1,96 @@
+# Request Parking (atenet router)
```
bowei, line 1: [aside] we should figure out where the small design doc snippets should be organized.

@thockin @juli4n -- regarding general repo organization.

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,155 @@
...
+	"errors"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Default request-parking parameters. See parkingConfig for the meaning of each
+// field; these are also the flag defaults wired up in NewCmd.
+const (
+	defaultParkingEnabled   = true
```
bowei: Do we need this flag if to max parked is set to 0?

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,155 @@
...
+	defaultParkingMaxWait   = 30 * time.Second
+	defaultParkingMaxParked = 2048
+)
+
+// Parking-wait outcome labels, recorded on the parking.wait.duration histogram.
+const (
+	parkOutcomeServed   = "served"   // resume succeeded and the request was routed
+	parkOutcomeTimeout  = "timeout"  // the request's deadline elapsed while parked
+	parkOutcomeCanceled = "canceled" // the client disconnected while parked
+	parkOutcomeError    = "error"    // resume failed (including park-budget exhaustion)
```
bowei: I think we should track budget exhaustion explicitly

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,155 @@
...
+//
+// When a request targets a suspended actor, the router resumes it via the
+// control plane before routing. If the worker pool is momentarily saturated the
+// control plane returns FailedPrecondition ("no free workers available"). With
+// parking enabled the router holds ("parks") such a request and keeps retrying
+// the resume until the actor becomes routable or maxWait elapses, instead of
+// failing the request immediately. maxParked bounds how many requests may be
+// parked at once so the router sheds load rather than queueing without bound.
+type parkingConfig struct {
+	enabled   bool
```
bowei: see comment on maxParked ==0

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,155 @@
...
+	return &parkingLot{cfg: cfg, metrics: m}
+}
+
+// enter attempts to reserve a parking slot. On success it returns a release
+// func and ok=true; the caller MUST invoke release exactly once (passing the
+// request outcome, e.g. parkOutcomeServed) when the resume attempt completes.
+// ok=false means the lot is full and the request should be shed without
+// waiting. When parking is disabled every request is admitted and no slot
+// accounting or metrics are recorded.
+func (l *parkingLot) enter(ctx context.Context) (release func(outcome string), ok bool) {
```
bowei: we should type the `outcome` as a go enum (string typed) instead of raw string

e.g.

type parkingLotOutcome string

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,155 @@
...
+		return func(string) {}, true
+	}
+
+	for {
+		cur := atomic.LoadInt64(&l.active)
+		if cur >= int64(l.cfg.maxParked) {
+			l.metrics.recordRejected(ctx)
+			return nil, false
+		}
+		if atomic.CompareAndSwapInt64(&l.active, cur, cur+1) {
```
bowei: Can we use mutex based exclusion for now instead of raw atomics?

At some point under a lot of load, we may want to switch to atomic vars, but I would like to take that step when we understand that this is worth taking on the complexity. This is not a statement about correctness of the current code, but it does complicate any future extensions or adjustments to the semantics of the parking lot code.

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// enter attempts to reserve a parking slot. On success it returns a release
+// func and ok=true; the caller MUST invoke release exactly once (passing the
+// request outcome, e.g. parkOutcomeServed) when the resume attempt completes.
+// ok=false means the lot is full and the request should be shed without
+// waiting. When parking is disabled every request is admitted and no slot
+// accounting or metrics are recorded.
+func (l *parkingLot) enter(ctx context.Context) (release func(outcome string), ok bool) {
+	if l == nil || !l.cfg.enabled {
```
bowei: This might be overkill --

You have a nil behavior and a cfg.enabled behavior.

Also see my comment about maxParked == 0 case.

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,155 @@
...
+		Active:    l.activeCount(),
+		MaxParked: l.cfg.maxParked,
+		MaxWait:   l.cfg.maxWait.String(),
+	}
+}
+
+// parkOutcome classifies a completed resume attempt for the wait-duration
+// metric. A budget-exhausted park surfaces the underlying capacity error and is
+// reported as parkOutcomeError.
+func parkOutcome(err error) string {
```
bowei: make outcome typed..

bowei (commented):
Reviewing in parts. I just took a look at parking.go

omeryahud:
Thanks @bowei & @thockin !
I'll address your comments and let you know once this is ready for another review

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,160 @@
...
+	mu     sync.Mutex
+	active int // current number of occupied slots; guarded by mu
+}
+
+func newParkingLot(cfg parkingConfig, m *parkingMetrics) *parkingLot {
+	return &parkingLot{cfg: cfg, metrics: m}
+}
+
+// enter attempts to reserve a parking slot. On success it returns a release
+// func and ok=true; the caller MUST invoke release exactly once (passing the
```
bowei, line 150: nit: I think since you use sync.Once, calling release() multiple times technically doesn't matter -- only the first instance does anything. To be clear, code that is sloppy and calls it more than once is probably a signal of other badness.

cmd/atenet/internal/router/resumer.go
bowei: +1

cmd/atenet/internal/router/resumer.go
```diff
@@ -28,21 +30,109 @@ import (
...
+
 // ActorResumer coordinates safe, deduplicated resumption of actors.
 type ActorResumer struct {
 	apiClient ateapipb.ControlClient
 	flight    singleflight.Group
+
+	// parkEnabled makes transient worker-pool saturation (FailedPrecondition)
+	// retryable, so a request is parked and retried until budget rather than
+	// failing immediately.
+	parkEnabled bool
```
bowei, line 75: I'm thinking instead of having extra flag, user can disable if park max == 0? What do you think? This should do the same thing?

cmd/atenet/internal/router/resumer.go
```diff
@@ -28,21 +30,109 @@ import (
...
-func NewActorResumer(apiClient ateapipb.ControlClient) *ActorResumer {
-	return &ActorResumer{
+// resumerOption configures an ActorResumer.
+type resumerOption func(*ActorResumer)
+
+// withParking configures parking behavior. When enabled, FailedPrecondition
+// ("no free workers available") becomes retryable and the resume is retried for
+// up to maxWait; a non-positive maxWait keeps the default budget. When disabled,
+// the resumer preserves its legacy fail-fast-on-capacity behavior.
+func withParking(enabled bool, maxWait time.Duration) resumerOption {
```
bowei: Given there is only one option right now, this is somewhat overkill. Are we planning on having more options?

cmd/atenet/internal/router/resumer.go
```diff
@@ -28,21 +30,109 @@ import (
...
+
+// retryable reports whether err warrants another resume attempt while the
+// request remains parked. A concurrent-resume conflict (Aborted) is always
+// retried. Transient pool saturation (FailedPrecondition, "no free workers
+// available") is retried only when parking is enabled, turning a momentary
+// shortage into a bounded wait instead of an immediate failure. All other codes
+// (NotFound, Unavailable, DeadlineExceeded, ...) are returned to the caller so
+// the HTTP boundary can map them with full fidelity.
+func (r *ActorResumer) retryable(err error) bool {
+	switch status.Code(err) {
```
bowei, line 124: would prefer if you did this:

if ! r.parkEnabled { return false }

then we don't have to worry about maintaining this in the `switch`

cmd/atenet/internal/router/resumer_test.go
```diff
@@ -168,3 +169,114 @@ func TestActorResumer_ResumeActor(t *testing.T) {
 		}
 	})
 }
+
+func TestActorResumer_Parking(t *testing.T) {
+	const testActorID = "actor-park"
```
bowei: [minor] use the const block

```
const (
 ...
)
```

cmd/atenet/internal/router/resumer_test.go
```diff
@@ -168,3 +169,114 @@ func TestActorResumer_ResumeActor(t *testing.T) {
...
+		mock := &resumerMockClient{
+			resumeFn: func(ctx context.Context, in *ateapipb.ResumeActorRequest, opts ...grpc.CallOption) (*ateapipb.ResumeActorResponse, error) {
+				mu.Lock()
+				calls++
+				mu.Unlock()
+				return nil, status.Error(codes.FailedPrecondition, "no free workers available")
+			},
+		}
+
+		// Budget large enough for a few ~500ms-spaced retries before it elapses;
```
bowei: Can we mock the clock make be with https://go.dev/blog/synctest?

Also, for unit test purposes, we should make the timeouts very short to avoid ending up with tests that take a bit of time to run.

cmd/atenet/internal/router/router.go
```diff
@@ -88,6 +88,12 @@ type RouterConfig struct {
 	AteapiCAFile     string
 	AteapiServerName string
 	AteapiTokenFile  string
+
+	// Request parking: hold and retry requests whose actor cannot be served
+	// immediately due to transient worker-pool saturation, instead of failing
+	// fast. A non-positive ParkingMaxParked disables parking. See parkingConfig.
+	ParkingMaxWait   time.Duration
```
bowei: I think you call this `budget` in other places. Do we want to align the names?

I suggest:

* `ParkedRequestBudget`
* `ParkedRequestMax`

demos/parking/load.sh
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env bash
```
bowei, line 1: I would prefer if this was written as an e2e test.

You can add the e2e in a follow up PR.

cmd/atenet/internal/router/errors.go
```diff
@@ -46,6 +46,13 @@ func invalidHostErr(host string, cause error) error {
 	}
 }
 
+// parkingFullErr returns a 503 reqError signaling that the router's parking lot
+// is at capacity, so the request was shed without waiting. Clients should retry.
+func parkingFullErr(actorID string) error {
```
bowei, line 67: Leave a TODO -- this will need to include the atespace and actorID is being renamed to actorName.

bowei, line 67: Leave a TODO -- this will need to include the atespace and actorID is being renamed to actorName.

cmd/atenet/internal/router/extproc.go
```diff
@@ -148,8 +150,20 @@ func (s *ExtProcServer) handleRequestHeaders(
...
 	}
 
+	// Admit the request to the parking lot before resuming. While resume is
+	// in-flight the request occupies a slot; if the actor's worker pool is
+	// momentarily saturated the resumer parks (retries) here rather than failing
+	// fast. A full lot sheds the request immediately so the router applies
+	// backpressure instead of queueing without bound.
+	release, ok := s.parking.enter(ctx)
+	if !ok {
+		return nil, metadata, "", "", "", parkingFullErr(actorID)
```
bowei: add a log here that the parklot is full

bowei (commented):
Generally lgtm, we should be able to merge after the comments are addressed.

cmd/atenet/internal/router.go
```diff
@@ -62,6 +62,8 @@ func NewRouterCmd() *cobra.Command {
 	cmd.Flags().StringVar(&cfg.AteapiCAFile, "ateapi-ca-file", ateapiauth.DefaultServiceAccountCAFile, "PEM file with CAs trusted to verify the ateapi server cert. Required for jwt.")
 	cmd.Flags().StringVar(&cfg.AteapiServerName, "ateapi-server-name", "", "SNI / hostname expected on the ateapi server cert. Optional.")
 	cmd.Flags().StringVar(&cfg.AteapiTokenFile, "ateapi-token-file", ateapiauth.DefaultServiceAccountTokenFile, "Projected SA token file used as Bearer credential. Required for jwt.")
+	cmd.Flags().DurationVar(&cfg.ParkingMaxWait, "parking-max-wait", 30*time.Second, "Maximum time a request may be parked (held and retried) waiting for its actor to become routable")
+	cmd.Flags().IntVar(&cfg.ParkingMaxParked, "parking-max-parked", 2048, "Maximum number of requests that may be parked simultaneously; excess requests are shed with 503. 0 disables parking (requests fail fast on worker-pool saturation)")
```
keithmattix: 👍🏾 park by default seems reasonable

cmd/atenet/internal/router/extproc.go
```diff
@@ -148,8 +150,20 @@ func (s *ExtProcServer) handleRequestHeaders(
...
+	// fast. A full lot sheds the request immediately so the router applies
+	// backpressure instead of queueing without bound.
+	release, ok := s.parking.enter(ctx)
+	if !ok {
+		return nil, metadata, "", "", "", parkingFullErr(actorName)
+	}
+
 	slog.InfoContext(ctx, "ResumeActor", slog.String("atespace", atespace), slog.String("actor", actorName))
 	actor, err := s.resumer.ResumeActor(ctx, atespace, actorName)
+	release(parkOutcomeFor(err))
```
keithmattix, line 165: nit: feels like something that should be deferred so future early returns or blocking code don't leave requests forever parked

cmd/atenet/internal/router/extproc_test.go
```diff
@@ -252,3 +255,52 @@ func TestExtProcHeadersEvaluation(t *testing.T) {
...
+		resumeFn: func(ctx context.Context, in *ateapipb.ResumeActorRequest, opts ...grpc.CallOption) (*ateapipb.ResumeActorResponse, error) {
+			resumeCalled = true
+			return &ateapipb.ResumeActorResponse{Actor: &ateapipb.Actor{AteomPodIp: "10.0.0.1"}}, nil
+		},
+	}
+
+	// A 1-slot lot with the slot already occupied deterministically simulates a
+	// full lot without needing a concurrent in-flight request.
+	s := NewExtProcServer(50051, clientMock, nil, parkingConfig{maxWait: time.Second, maxParked: 1}, nil)
+	occupy, ok := s.parking.enter(context.Background())
```
keithmattix: nit: isn't occupy a callback to free the slot? The naming is a bit confusing

cmd/atenet/internal/router/metrics.go
```diff
@@ -27,6 +30,14 @@ const (
 	// atenet.router.route.duration measures the latency from when the ext_proc handler receives a request
 	// (Envoy -> EPP) until the target worker endpoint is resolved
 	routeDurationMetricName = "atenet.router.route.duration"
+
+	// Request-parking instruments. parking.active is the live count of parked
+	// requests; parking.wait.duration is how long each request stayed parked
+	// (labeled by outcome); parking.rejected counts requests shed because the
+	// parking lot was full.
+	parkingActiveMetricName   = "atenet.router.parking.active"
+	parkingWaitMetricName     = "atenet.router.parking.wait.duration"
```
keithmattix, line 39: ~~The cardinality of this feels like it could be explosive if it's every request. Maybe a histogram across all requests might be more scalable?~~

Saw that this is what's implemented. Might be good to add a suffix (e.g _duration_ms) or something but looks good wrt scale

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+// parkOutcome is the terminal disposition of a parked request. It is recorded
+// as the `outcome` label on the parking.wait.duration histogram.
+type parkOutcome string
+
+// Park-wait outcomes, recorded on the parking.wait.duration histogram.
+const (
+	parkOutcomeServed          parkOutcome = "served"           // resume succeeded and the request was routed
+	parkOutcomeBudgetExhausted parkOutcome = "budget_exhausted" // the park budget elapsed while still blocked on a retryable condition
```
keithmattix, line 53: Might be good to have a metric for this as well, separate from rejected since that's only shed requests

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,160 @@
...
+// control plane before routing. If the worker pool is momentarily saturated the
+// control plane returns FailedPrecondition ("no free workers available"). With
+// parking enabled the router holds ("parks") such a request and keeps retrying
+// the resume until the actor becomes routable or maxWait elapses, instead of
+// failing the request immediately. maxParked bounds how many requests may be
+// parked at once so the router sheds load rather than queueing without bound;
+// a non-positive maxParked disables parking entirely.
+type parkingConfig struct {
+	maxWait   time.Duration
+	maxParked int
```
keithmattix: Super nit: would uint be better?

cmd/atenet/internal/router/parking.go
```diff
@@ -0,0 +1,160 @@
...
+	l.mu.Unlock()
+
+	start := time.Now()
+	l.metrics.addActive(ctx, 1)
+
+	var once sync.Once
+	return func(outcome parkOutcome) {
+		once.Do(func() {
+			l.mu.Lock()
+			l.active--
```
keithmattix: nit: maybe add a lower bound to ensure this doesn't go negative. I know that would only happen because of other bugs though

cmd/atenet/internal/router/resumer.go
keithmattix: +1

omeryahud:
Will be addressing comments next week (after k8s code freeze)

omeryahud:
@bowei @thockin @keithmattix 
Hi guys, this is ready for another review.

I've reworked it quite a bit, and still have a couple of minor things to flesh out, including an e2e test I want to add.

Please let me know what you think

cmd/atenet/internal/router/cmd.go
```diff
@@ -60,6 +60,12 @@ func NewRouterCmd() *cobra.Command {
...
+	cmd.Flags().IntVar(&cfg.ParkedRequestMax, "parked-request-max", defaultParkedRequestMax, "Maximum number of requests that may be parked simultaneously; excess requests are shed with 503. 0 disables parking (requests fail fast on worker-pool saturation)")
+	cmd.Flags().DurationVar(&cfg.ParkedRequestRetryInterval, "parked-request-retry-interval", defaultParkedRequestRetryInterval, "Delay before a parked request's first resume retry")
+	cmd.Flags().Float64Var(&cfg.ParkedRequestRetryFactor, "parked-request-retry-factor", defaultParkedRequestRetryFactor, "Multiplier applied to the retry delay after each attempt; must be >= 1")
+	cmd.Flags().Float64Var(&cfg.ParkedRequestRetryJitter, "parked-request-retry-jitter", defaultParkedRequestRetryJitter, "Random fraction in [0, 1) added to each retry delay to de-synchronize parked requests")
+	cmd.Flags().IntVar(&cfg.ExtProcMaxRequests, "extproc-max-requests", defaultExtProcMaxRequests, "Circuit-breaker max_requests for Envoy's ext_proc cluster; every parked request holds one slot for its full wait, so this must be >= --parked-request-max (the excess is fast-path headroom)")
```
bowei: Does this make sense to calculate from parked-request-max vs as its own flag?

cmd/atenet/internal/router/config.go
```diff
@@ -55,4 +56,32 @@ type routerConfig struct {
...
+	ParkedRequestMax           int
+	ParkedRequestRetryInterval time.Duration
+	ParkedRequestRetryFactor   float64
+	ParkedRequestRetryJitter   float64
+
+	// ExtProcMaxRequests is the circuit-breaker max_requests Envoy applies to
+	// the ext_proc cluster. Every parked request holds one slot for its entire
+	// wait, so this must be >= ParkedRequestMax (validated at startup); the
+	// excess is fast-path headroom for requests to already-running actors.
+	ExtProcMaxRequests int
```
bowei, line 70: See statement above -- does this need to be separate from the ParkedRequestMax or we can do something that calculates it? Like set it to 120% of ParkedRequestMax?

cmd/atenet/internal/router/config.go
```diff
@@ -55,4 +56,32 @@ type routerConfig struct {
 	OtlpCollectorAddress string
 
 	Auth authConfig
+
+	// Request parking: hold and retry requests whose actor cannot be served
+	// immediately due to transient worker-pool saturation, instead of failing
+	// fast. A non-positive ParkedRequestMax disables parking. See parkingConfig.
+	ParkedRequestBudget        time.Duration
```
bowei: Let's put the parkedxxx flags together like this:

```
type ParkedRequestConfig struct {
  Budget
  Max
  RetryInterval
  ...
}
```

This will make the prefix in the code and easier to deal with.

cmd/atenet/internal/router/config.go
```diff
@@ -55,4 +56,32 @@ type routerConfig struct {
...
+	// ExtProcMaxRequests is the circuit-breaker max_requests Envoy applies to
+	// the ext_proc cluster. Every parked request holds one slot for its entire
+	// wait, so this must be >= ParkedRequestMax (validated at startup); the
+	// excess is fast-path headroom for requests to already-running actors.
+	ExtProcMaxRequests int
+}
+
+// validate rejects flag combinations that would make the router misbehave
+// rather than merely differ.
+func (c routerConfig) validate() error {
```
bowei, line 95: put the validate on the parkRequestConfig (see comment above)

cmd/atenet/internal/router/extproc_test.go
bowei: yes, this should be called `release`

bowei (commented):
just a few quick things

omeryahud:
@bowei Thanks for the comments! I've addressed them all, including e2e test

PR #224, opened 2026-06-11
fix(k8sjwt): redact sensitive claims from logs

konippi:
Part of #103.

`KubernetesClaims` now implements `slog.LogValuer` via a value receiver, so a logged copy cannot bypass redaction. Only non-sensitive claims (issuer, subject, audiences, timestamps, jti, namespace, service account, pod, node) are emitted; the secret name/UID and the resource UIDs are excluded.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/k8sjwt/k8sjwt.go
+++ b/cmd/ateapi/internal/k8sjwt/k8sjwt.go
@@ -110,6 +110,27 @@ type KubernetesClaims struct {
 	WarnAfter time.Time
 }
 
+var _ slog.LogValuer = KubernetesClaims{}
+
+// Allowlist of non-sensitive claims; the value receiver stops a logged copy from bypassing it.
+func (c KubernetesClaims) LogValue() slog.Value {
+	attrs := []slog.Attr{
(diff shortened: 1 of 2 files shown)


EItanya (dismissed):
I wasn't aware of the `slog.LogValuer` interface before, I like the approach :)

konippi:
Rebased onto main to resolve the conflict from #270.  Thanks for the review @EItanya!

mtaufen:
Why are these fields considered sensitive? What is the attack that relies on them?

thockin:
ping on this

konippi:
Honestly, "sensitive" was the wrong word here. None of these are credentials (the signed token never even lands in `KubernetesClaims`), and the UIDs and `secret` bound-ref are just identifiers, so I can't point to a real attack. What it actually buys us is future-proofing: a value-receiver `LogValuer` makes the struct allowlist-by-default, so a genuinely sensitive field added later won't silently leak through reflection — and #103's proto-redaction won't cover a plain struct like this one. Happy to keep it as that small safety net (retitled, with the harmless IDs allowlisted), or close it in favor of #103 — your call.

Issue #119, opened 2026-05-29
[Feature] Implementation of Actor State Machine: Lifecycle and Transitions


dberkov:
## Issue Description

This issue tracks the formal implementation of the actor state machine within the substrate. 

To achieve high density and efficiency, we must distinguish between different "resting" states for actors - specifically focusing on the trade-offs between resume speed, data durability, and resource costs. The implementation must specifically address a potential NIC saturation problem. A simplistic approach of uploading full memory and filesystem images over the network during every suspend/resume cycle would quickly overwhelm the host VM's network interface card (NIC). Consequently, the state machine implementation must prioritize keeping data on the local host to maintain required performance and density.

The "story" of an actor follows a journey from active execution through various stages of dormancy or failure, driven by both user actions and system timeouts.

## Actor State Definitions

| State | Description | Durability & Performance |
| :--- | :--- | :--- |
| **RUNNING** | The actor is currently active and executing on a worker. | Active CPU/RAM usage. |
| **PAUSED** | The actor is offloaded from the CPU, with snapshots stored locally on the node VM. | Fast resume; lower durability (data lost if node fails); cheaper (local, already provisioned storage, no NIC); we can optimize this (e.g., save to peers opportunistically (TODO add issue link)). |
| **SUSPENDED** | The actor is offloaded from the CPU, and snapshots are persisted to durable storage. | Higher SLO/durability (e.g., saved to blob storage); expensive (external storage and NIC); requires retention policies; slower resume due to data transfer; can be symbolically tagged. |
| **CRASHED** | A failure state reached if the host node fails while an actor is Running or Paused. | Requires manual recovery or revert to a known good state. |

## Snapshot layers

```mermaid
---
config:
  treemap:
    showValues: false
---
treemap-beta
"Actor's process"
    "memory" : 40
    "local storage"
        "rootfs"
            "durabledir" : 20
            "oci bundle" : 80:::noborder

classDef noborder stroke-width:0px;
```

- **memory** - a snapshot of an actor's process memory. A snapshot cannot be used for restoration if the OCI image of the actor has been changed (e.g., upgraded).
- **rootfs** - an entire boot disk snapshot with all files. A snapshot cannot be used for restoration if the OCI image of the actor has been changed (e.g., upgraded).
- **durabledir** - a durable directory on rootfs that persists across resumes and participates in snapshots. Can be used for restoration even if the OCI image has been changed.

## Actor Lifecycle Story

```mermaid
---
[...]

Issue #16, opened 2026-05-20
Some improvements to ActorTemplate


juli4n:
There are a few cases where updating / deleting ActorTemplate resources can lead to a broken system.

* Deleting an `ActorTemplate` that is currently in use by an actor breaks the `ResumeActor` and `SuspendActor` RPCs.
* If an `ActorTemplate` is updated, the golden snapshot becomes stale and actors using the template diverge. Although `ActorTemplate`s are conceptually immutable, this is not enforced today. 
* Deleting an `ActorTemplate` leaks its associated golden actor.

We should discuss whether we want to allow an actor to outlive its template. Assuming we don't, a possible path forward would be:

* The Substrate controller manages `ActorTemplate` resources.
* `ActorTemplate`s are immutable (enforced via CEL validation).
* Each `ActorTemplate` owns the lifecycle of a “golden actor”. The golden actor is created when the template is created, and deleted when the template is deleted.
* An `ActorTemplate` cannot be deleted if there are actors using it (this will require `ate-apiserver` to expose an efficient way to check if there are active actors referencing a given template).

ArnaudBger:
I’m wondering about the upgrade/recovery story for long-lived actors.

Let’s say an Actor is running an Agent Session Loop, and that loop is responsible for calling an external LLM provider API.

If the provider API changes incompatibly, does that mean actors created or snapshotted before the change could resume with an outdated in-memory Agent Session Loop that no longer works?

In that case, even if ActorTemplates become immutable, I’m not sure what the recovery path is for existing actors.
Is the intended model:

old actors must continue using the exact template/image they were created from,
actors should be recreated from a newer template,
or there should be some mechanism to “bump” / migrate an actor to a newer or previous template?
Put differently: how does the current design expect users to handle external API drift when actor state may outlive the template version that originally produced it?

thockin:
@ArnaudBger I think this topic is not pinned down yet, so let's talk about what semantic we think makes sense.  I would love for us all to agree on a single semantic, but I suspect there will end up being some configuration involved.

> old actors must continue using the exact template/image they were created from,
> actors should be recreated from a newer template,
>or there should be some mechanism to “bump” / migrate an actor to a newer or previous template?

My gut instinct is that there MUST be a way for a developer to update a template and then roll that out across the fleet of instances of that template.

I suspect we really want to define something like a set of versions of a given template.  For example (not well thought out yet):

* Each `ActorTemplate` (AT) has N "Versions" (ATVs).
[...]

PR #226, opened 2026-06-11
Allow tag-based ActorTemplate images

### LINKED ISSUE #223: Allow tag-based ActorTemplate images by persisting resolved digests with snapshots
## Problem

ActorTemplate currently requires all workload images to be digest-pinned:

- `spec.pauseImage`
- every `spec.containers[].image`

The validation message says this is required because changing the image invalidates snapshots. That concern is real: on restore, atelet recreates OCI rootfs bundles before calling `runsc restore`, so restoring a checkpoint against a different base image can be incorrect or fail.

However, requiring users to provide digest-pinned image references makes demos and day-to-day development awkward. Users often naturally write `repo/image:tag`, and tooling has to resolve tags before applying ActorTemplates.

## Proposal

Allow ActorTemplate images to be specified by tag, but make snapshot restore deterministic by persisting the resolved image reference alongside snapshot data.

When atelet pulls an image for `Run` or `Restore`, it should record the exact digest-resolved image reference that was used to build the OCI rootfs. When atelet later checkpoints the actor, it should upload that image-resolution metadata under the snapshot prefix, alongside the existing checkpoint files.

On subsequent restore from that snapshot, atelet should read the snapshot metadata and use the stored digest-resolved references when rebuilding OCI bundles, instead of resolving the current ActorTemplate tags again.

This preserves the safety property the digest requirement was trying to enforce, without forcing users to pre-resolve every image in the ActorTemplate.

## Current code shape

Today:

- The CRD validation requires `@` in both `Container.Image` and `ActorTemplateSpec.PauseImage`.
- ateapi materializes an atelet `WorkloadSpec` directly from the current ActorTemplate.
- atelet `Restore` downloads checkpoint files, then calls `prepareOCIBundles` using that current `WorkloadSpec`.
- `memorypullcache.Fetch` only uses its digest-keyed cache when the requested ref already contains a digest.
- snapshot storage currently contains `checkpoint.img.zstd`, `pages.img.zstd`, and `pages_meta.img.zstd`; there is no snapshot manifest containing image metadata.

## Implementation sketch

1. Relax ActorTemplate image validation:
   - remove the `self.contains(@)` XValidation from `spec.pauseImage`
   - remove the same validation from `spec.containers[].image`
   - update generated CRDs, Helm CRDs, tests, docs, and demos

2. Extend the image pull path:
   - change `memorypullcache.Fetch` or a wrapper around it to return the resolved digest reference in addition to the rootfs tar stream
   - for tag refs, resolve to the platform-specific image digest used by `remote.Image(..., remote.WithPlatform(...))`
   - for already-pinned refs, preserve the caller’s pinned digest semantics where appropriate, especially for multi-arch/index digests

3. Persist local runtime image metadata:
   - when `prepareOCIDirectory` prepares a bundle, write metadata recording:
     - container name (`pause` or workload container name)
     - original image ref from the WorkloadSpec
     - resolved digest ref actually used for the rootfs
   - checkpoint should upload this metadata before deleting local actor dirs

4. Add a snapshot manifest:
   - store a small JSON object under the snapshot prefix, for example `snapshot-manifest.json`
   - include a version field for future compatibility
   - include pause image and per-container resolved refs keyed by container name

5. Use the manifest on restore:
   - before `prepareOCIBundles`, fetch the snapshot manifest if present
   - rewrite the restore WorkloadSpec image refs to the manifest’s resolved refs
   - fail clearly if the manifest is missing for snapshots that require tag resolution, or define an explicit backwards-compatibility fallback for old snapshots

6. Keep checkpoint accurate:
   - do not resolve tags at checkpoint time from the current ActorTemplate
   - checkpoint should persist the image refs that were used when the actor was actually started/restored

## Acceptance criteria

- ActorTemplates can be created with tag-based image refs, e.g. `busybox:latest`.
- The first run resolves tags and records digest-pinned refs.
- Golden snapshots persist the resolved refs used to create the golden actor.
- Restoring from a golden snapshot or actor snapshot uses the stored digest refs, not whatever the tag points to later.
- Existing digest-pinned ActorTemplates continue to work.
- Existing snapshots either continue to restore through a documented fallback or fail with a clear actionable error.
- Unit tests cover:
  - validation accepts tag refs
  - resolved refs are persisted
  - restore prefers snapshot manifest refs over current ActorTemplate refs
  - already-pinned multi-arch refs preserve the intended digest behavior

EItanya:
## Summary

Fixes #223.

- relax ActorTemplate image validation so tags are accepted
- persist resolved digest image refs in a versioned `snapshot-manifest.json`
- restore snapshots using manifest-pinned image refs instead of re-resolving current tags
- normalize object-storage missing-object errors for manifest fallback handling
- reserve workload container name `pause` at the API level because atelet uses it for sandbox infrastructure

## Testing

- `go test ./cmd/atelet ./internal/memorypullcache ./pkg/api/v1alpha1`
- `GOCACHE=$PWD/.cache/gocache hack/update/go-generate.sh`
- `hack/verify/crd-chart.sh`
- `git diff --check`

Full `go test ./...` was not clean in this sandbox: default run hit `NO_COLOR`-sensitive color tests; rerun with `NO_COLOR` unset and `GOCACHE=/tmp/substrate-go-cache` then failed on sandbox-restricted local TCP listeners (`socket: operation not permitted`) and `setup-envtest`.

### DIFF
--- a/cmd/atelet/internal/ategcs/gcs.go
+++ b/cmd/atelet/internal/ategcs/gcs.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 
 	"cloud.google.com/go/storage"
 )
@@ -32,7 +33,14 @@ func NewGCSClient(client *storage.Client) ObjectStorage {
 }
 
 func (g *gcsClient) GetObject(ctx context.Context, bucket, object string) (io.ReadCloser, error) {
-	return g.client.Bucket(bucket).Object(object).NewReader(ctx)
+	rc, err := g.client.Bucket(bucket).Object(object).NewReader(ctx)
+	if err != nil {
+		if errors.Is(err, storage.ErrObjectNotExist) {
+			return nil, fmt.Errorf("%w: %w", fs.ErrNotExist, err)
+		}
+		return nil, err
+	}
+	return rc, nil
 }
 
 func (g *gcsClient) PutObject(ctx context.Context, bucket, object string, reader io.Reader) error {
--- a/cmd/atelet/internal/ategcs/objects.go
+++ b/cmd/atelet/internal/ategcs/objects.go
@@ -15,6 +15,7 @@
 package ategcs
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -55,6 +56,17 @@ func FetchFromGCS(ctx context.Context, client ObjectStorage, gsURL string) ([]by
 	}
 
 	return content, nil
+}
+
+func SendBytesToGCS(ctx context.Context, client ObjectStorage, gsURL string, content []byte) error {
+	bucket, object, err := parseGCSURL(gsURL)
+	if err != nil {
+		return fmt.Errorf("while parsing URL: %w", err)
+	}
+	if err := client.PutObject(ctx, bucket, object, bytes.NewReader(content)); err != nil {
+		return fmt.Errorf("while putting object: %w", err)
+	}
+	return nil
 }
 
 func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, localFilePath string) (err error) {
--- a/cmd/atelet/internal/ategcs/s3.go
+++ b/cmd/atelet/internal/ategcs/s3.go
@@ -16,10 +16,14 @@ package ategcs
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"io"
+	"io/fs"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/smithy-go"
 )
 
 type s3Client struct {
@@ -36,6 +40,10 @@ func (s *s3Client) GetObject(ctx context.Context, bucket, object string) (io.Rea
 		Key:    aws.String(object),
 	})
 	if err != nil {
+		var apiErr smithy.APIError
+		if errors.As(err, &apiErr) && apiErr.ErrorCode() == "NoSuchKey" {
+			return nil, fmt.Errorf("%w: %w", fs.ErrNotExist, err)
+		}
 		return nil, err
 	}
 	return output.Body, nil
--- a/cmd/atelet/internal/memorypullcache/memorypullcache.go
+++ b/cmd/atelet/internal/memorypullcache/memorypullcache.go
@@ -42,6 +42,11 @@ type MemoryPullCache struct {
 	cache *lru.Cache
 }
 
+type PullResult struct {
+	RootFSTar   io.ReadCloser
+	ResolvedRef string
+}
+
 func NewMemoryPullCache(ctx context.Context, gcpAuthenticator authn.Authenticator, localhostRegistryReplacement string) (*MemoryPullCache, error) {
 	c := &MemoryPullCache{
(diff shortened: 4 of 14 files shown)


dberkov:
>relax ActorTemplate image validation so tags are accepted

To ensure OCI image immutability, ActorTemplate currently requires an explicit full SHA256. Using labels instead would jeopardize resume operations for SUSPENDED actors if the OCI image associated with that label changes.

As part of #119, a "homedir" snapshot concept is being introduced to allow resume operations to persist even when an OCI image does not match. Additionally, [the proposal in #16](https://github.com/agent-substrate/substrate/issues/16#issuecomment-4585262887) addresses an upgrade mode intended to support surviving OCI image transitions.

I recommend finalizing the upgrade flow first. This will clarify whether ActorTemplate should be treated as a mutable or immutable object before we implement further behavioral changes.

@juli4n - FYI

EItanya:
> To ensure OCI image immutability, ActorTemplate currently requires an explicit full SHA256. Using labels instead would jeopardize resume operations for SUSPENDED actors if the OCI image associated with that label changes.

FWIW this doesn't actually get rid of that contract, it just sets that SHA256 into the contract on first pull rather than requiring the user to set it. It makes the system lock it rather than requiring the user search out that information. I understand the technical limitation, I was just trying to make the user workflow a tiny bit easier. I have found for myself that this adds friction.

> As part of https://github.com/agent-substrate/substrate/issues/119, a "homedir" snapshot concept is being introduced to allow resume operations to persist even when an OCI image does not match. Additionally, https://github.com/agent-substrate/substrate/issues/16#issuecomment-4585262887 addresses an upgrade mode intended to support surviving OCI image transitions.

I haven't read through #119 yet, but I will try to do so today. Using a homedir approach for flexible snapshots make sense, especially for agents who rely mostly on filesystems that way.

BenTheElder:
> FWIW this doesn't actually get rid of that contract, it just sets that SHA256 into the contract on first pull rather than requiring the user to set it. It makes the system lock it rather than requiring the user search out that information. I understand the technical limitation, I was just trying to make the user workflow a tiny bit easier. I have found for myself that this adds friction.

Yeah, the only downside is this might be surprising, since you specified a tag but it gets locked in. 

Right now it's explicit that you must use an immutable reference which is annoying but should have unsurprising behavior.

Would we re-resolve on a new goldensnapshot?

If we were going full KRM, I would probably reflect the resolved image in the actortemplate status for visibility.

We could instead offer tooling to create actortemplates which does the resolve on creation of the actortemplate.

BenTheElder:
I'm just thinking about how many times I've seen users trip over `imagePullPolicy` defaulting depending on if the image is explicitly or implicitly `:latest` in kubernetes verus literally any other reference.

I think a lot of users do not have a good grasp on tag lifecycle so while I'm not opposed in general we should think about how to make the behavior least-surprising. Requiring tag resolution be done explicitly up front is annoying but has ~unsurprising runtime behavior.

EItanya:
I'm going to resolve conflicts now

>  I'm just thinking about how many times I've seen users trip over imagePullPolicy defaulting depending on if the image is explicitly or implicitly :latest in kubernetes verus literally any other reference.

> I think a lot of users do not have a good grasp on tag lifecycle so while I'm not opposed in general we should think about how to make the behavior least-surprising. Requiring tag resolution be done explicitly up front is annoying but has ~unsurprising runtime behavior.

I agree with all of this, which is exactly why I'm resolving once onto the `ActorTemplate` status. Users can still use `sha` if they want more specific behavior.

PR #227, opened 2026-06-12
Actor Pause/Resume flow

### LINKED ISSUE #119: [Feature] Implementation of Actor State Machine: Lifecycle and Transitions

dberkov:
Implements  `PAUSED` state for issue #119.

The snapshot files are kept locally on node VM in a separate folder. At resume time, scheduler uses a node VM hint and picks up a worker from the same node where files were stored at suspend time.

The local file management solution is temporary and will be replaced once @msau42 introduces a new component that is supposed to manage files on the node VM.

- [X] Tests pass
- [X] Manual tests with counter demo
```
>kubectl ate create actor my-counter-1 --template ate-demo-counter/counter
>curl -X POST -H "Host: my-counter-1.actors.resources.substrate.ate.dev" http://localhost:8000
hello from: 169.254.17.2 | preserved memory count: 1
>curl -X POST -H "Host: my-counter-1.actors.resources.substrate.ate.dev" http://localhost:8000
hello from: 169.254.17.2 | preserved memory count: 2
>kubectl ate pause actor my-counter-1 -o json
> curl -X POST -H "Host: my-counter-1.actors.resources.substrate.ate.dev" http://localhost:8000
hello from: 169.254.17.2 | preserved memory count: 3
```

### Breaking change
This PR introduces a breaking change in the Actor proto. All existing actor  needs to be recreated, prior testing PAUSE functionality.

### DIFF
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -1124,14 +1124,112 @@ func TestSuspendActor(t *testing.T) {
 			ActorTemplateNamespace: ns,
 			ActorTemplateName:      "tmpl1",
 			Status:                 ateapipb.Actor_STATUS_SUSPENDED,
-			LastSnapshot:           fmt.Sprintf("gs://my-bucket/%s/tmpl1/%s/", ns, id),
+			LatestSnapshotInfo: &ateapipb.SnapshotInfo{
+				Type: ateapipb.SnapshotType_SNAPSHOT_TYPE_EXTERNAL,
+				Data: &ateapipb.SnapshotInfo_External{
+					External: &ateapipb.ExternalSnapshotInfo{
+						SnapshotUriPrefix: fmt.Sprintf("gs://fake-fake-fake/%s/", id),
+					},
+				},
+			},
 		},
 	}
 
-	if diff := cmp.Diff(want, getResp, protocmp.Transform(), protocmp.IgnoreFields(&ateapipb.Actor{}, "version", "last_snapshot", "ateom_pod_uid")); diff != "" {
+	if diff := cmp.Diff(want, getResp,
+		protocmp.Transform(),
+		protocmp.IgnoreFields(&ateapipb.Actor{}, "version"),
+		protocmp.IgnoreFields(&ateapipb.Actor{}, "ateom_pod_uid"),
+		protocmp.FilterField(&ateapipb.ExternalSnapshotInfo{}, "snapshot_uri_prefix", cmp.Comparer(func(x, y string) bool {
+			return strings.HasPrefix(y, x)
+		})),
+	); diff != "" {
 		t.Errorf("GetActor response mismatch (-want +got):\n%s", diff)
 	}
+}
 
+// TestPauseActor tests the full workflow of pausing a running actor.
+// Workflow:
+// 1. Creates a mock ActorTemplate.
+// 2. Creates a mock Atelet Pod on 'node1'.
+// 3. Creates a mock worker Pod on 'node1'.
+// 4. Waits for the WorkerPoolSyncer to mirror the worker to Redis.
+// 5. Creates an actor.
+// 6. Calls ResumeActor to transition it to RUNNING.
+// 7. Calls PauseActor RPC.
+// 8. Verifies that the fake Atelet received the Pause call.
+func TestPauseActor(t *testing.T) {
+	ns := namespaceForTest("ns-suspend")
+	tc := setupTest(t, ns)
+	defer tc.cleanup()
+
+	createTemplate(t, tc, ns)
+
+	createWorkerPod(t, tc, ns, "worker-1", "node1")
+
+	_, err := tc.client.CreateActor(context.Background(), &ateapipb.CreateActorRequest{
+		ActorTemplateNamespace: ns,
+		ActorTemplateName:      "tmpl1",
+		ActorId:                "id1",
+	})
+	if err != nil {
+		t.Fatalf("CreateActor failed: %v", err)
+	}
+	id := "id1"
+
+	// Resume first to make it running
+	_, err = tc.client.ResumeActor(context.Background(), &ateapipb.ResumeActorRequest{
+		ActorId: id,
+	})
+	if err != nil {
+		t.Fatalf("ResumeActor failed: %v", err)
+	}
+
+	// Suspend
+	_, err = tc.client.PauseActor(context.Background(), &ateapipb.PauseActorRequest{
+		ActorId: id,
+	})
+	if err != nil {
+		t.Fatalf("PauseActor failed: %v", err)
+	}
+
+	if !tc.fakeAtelet.CheckpointCalled {
+		t.Errorf("expected atelet Checkpoint to be called")
+	}
+
+	getResp, err := tc.client.GetActor(context.Background(), &ateapipb.GetActorRequest{
+		ActorId: id,
+	})
+	if err != nil {
+		t.Fatalf("GetActor failed: %v", err)
+	}
+	want := &ateapipb.GetActorResponse{
+		Actor: &ateapipb.Actor{
+			ActorId:                id,
+			ActorTemplateNamespace: ns,
+			ActorTemplateName:      "tmpl1",
+			Status:                 ateapipb.Actor_STATUS_PAUSED,
+			LatestSnapshotInfo: &ateapipb.SnapshotInfo{
+				Type: ateapipb.SnapshotType_SNAPSHOT_TYPE_LOCAL,
+				Data: &ateapipb.SnapshotInfo_Local{
+					Local: &ateapipb.LocalSnapshotInfo{
+						SnapshotPrefix:            "id1",
+						NodeVmsWithLocalSnapshots: []string{"node1"},
+					},
+				},
+			},
+		},
+	}
+
+	if diff := cmp.Diff(want, getResp,
+		protocmp.Transform(),
+		protocmp.IgnoreFields(&ateapipb.Actor{}, "version"),
+		protocmp.IgnoreFields(&ateapipb.Actor{}, "ateom_pod_uid"),
+		protocmp.FilterField(&ateapipb.LocalSnapshotInfo{}, "snapshot_prefix", cmp.Comparer(func(x, y string) bool {
+			return strings.HasPrefix(y, x)
+		})),
+	); diff != "" {
+		t.Errorf("GetActor response mismatch (-want +got):\n%s", diff)
+	}
 }
 
 // TestValidation tests the negative validation cases for all gRPC methods.
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -868,6 +868,7 @@ func TestResumeActor(t *testing.T) {
 		ActorTemplate:   "tmpl1",
 		ActorId:         id,
 		Ip:              "127.0.0.1",
+		NodeName:        "node1",
 	}
 
 	if diff := cmp.Diff(wantWorker, actorWorker, protocmp.Transform(), protocmp.IgnoreFields(&ateapipb.Worker{}, "version"), protocmp.IgnoreFields(&ateapipb.Worker{}, "worker_pod_uid")); diff != "" {
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -1158,7 +1158,7 @@ func TestSuspendActor(t *testing.T) {
 // 7. Calls PauseActor RPC.
 // 8. Verifies that the fake Atelet received the Pause call.
 func TestPauseActor(t *testing.T) {
-	ns := namespaceForTest("ns-suspend")
+	ns := namespaceForTest("ns-pause")
 	tc := setupTest(t, ns)
 	defer tc.cleanup()
 
@@ -1184,7 +1184,7 @@ func TestPauseActor(t *testing.T) {
 		t.Fatalf("ResumeActor failed: %v", err)
 	}
 
-	// Suspend
+	// Pause
 	_, err = tc.client.PauseActor(context.Background(), &ateapipb.PauseActorRequest{
 		ActorId: id,
 	})
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -17,6 +17,7 @@ package controlapi
 import (
 	"context"
 	"fmt"
+	"log"
 	"net"
 	"os"
 	"os/exec"
@@ -60,9 +61,11 @@ var (
 
 func TestMain(m *testing.M) {
 	cmd := exec.Command("bash", "../../../../hack/run-tool.sh", "setup-envtest", "use", "--print", "path")
+	var stderr strings.Builder
+	cmd.Stderr = &stderr
 	out, err := cmd.Output()
 	if err != nil {
-		os.Exit(1)
+		log.Fatalf("setup-envtest: %v (stderr: %s)", err, stderr.String())
 	}
 	binaryAssetsDirectory := strings.TrimSpace(string(out))
 
@@ -73,19 +76,19 @@ func TestMain(m *testing.M) {
 
 	cfg, err = testEnv.Start()
 	if err != nil {
-		os.Exit(1)
+		log.Fatalf("testEnv.Start: %v", err)
 	}
 
 	// Create ate-system namespace
 	k8sClient, err := kubernetes.NewForConfig(cfg)
 	if err != nil {
-		os.Exit(1)
+		log.Fatalf("kubernetes.NewForConfig: %v", err)
 	}
 	_, err = k8sClient.CoreV1().Namespaces().Create(context.Background(), &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{Name: "ate-system"},
 	}, metav1.CreateOptions{})
 	if err != nil && !strings.Contains(err.Error(), "already exists") {
-		os.Exit(1)
+		log.Fatalf("create ate-system namespace: %v", err)
 	}
 
 	// Create shared Atelet Pod
@@ -106,14 +109,14 @@ func TestMain(m *testing.M) {
 	}
 	createdAtelet, err := k8sClient.CoreV1().Pods("ate-system").Create(context.Background(), ateletPod, metav1.CreateOptions{})
 	if err != nil && !strings.Contains(err.Error(), "already exists") {
-		os.Exit(1)
+		log.Fatalf("create atelet pod: %v", err)
 	}
 	if err == nil {
 		createdAtelet.Status.PodIPs = []corev1.PodIP{{IP: "127.0.0.1"}}
 		createdAtelet.Status.Phase = corev1.PodRunning
 		_, err = k8sClient.CoreV1().Pods("ate-system").UpdateStatus(context.Background(), createdAtelet, metav1.UpdateOptions{})
 		if err != nil {
-			os.Exit(1)
+			log.Fatalf("update atelet pod status: %v", err)
 		}
 	}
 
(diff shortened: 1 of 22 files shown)


internal/ateompath/ateompath.go
```diff
@@ -135,3 +142,10 @@ func PIDFilePath(actorTemplateNamespace, actorTemplateName, actorID, containerNa
 		containerName+".pid",
 	)
 }
+
+func LocalCheckpointDir(actorTemplateNamespace, actorTemplateName, actorID string) string {
```
BenTheElder: Dead code? It seems only the plural version is called.

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+		}
+
+		// 2. Safely clear ActiveWorker now that the worker object in DB is freed
+		latestActor, err = s.store.GetActor(ctx, input.ActorID)
+		if err != nil {
+			return err
+		}
+		latestActor.Status = ateapipb.Actor_STATUS_PAUSED
+		// TODO(dberkov) - what if we still don't know the node name? Maybe move to CRASHED status?
```
BenTheElder, line 232: Yeah ... this seems like a problem, we should probably bail out if we can't get the node name?

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func (s *MarkPausingStep) Name() string { return "MarkPausing" }
+func (s *MarkPausingStep) IsComplete(ctx context.Context, input *PauseInput, state *PauseState) (bool, error) {
+	// Fast forward if we've already marked our intent or if we are further along.
+	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_PAUSING || state.Actor.GetStatus() == ateapipb.Actor_STATUS_PAUSED, nil
+}
+func (s *MarkPausingStep) Execute(ctx context.Context, input *PauseInput, state *PauseState) error {
+	if state.Actor.GetStatus() != ateapipb.Actor_STATUS_RUNNING {
+		return nil
```
BenTheElder, line 83: This seems like an error state?

AI:
> Pausing a non-RUNNING actor degrades to a confusing error — [workflow_pause.go:78-86](vscode-webview://0fkcl6ums9vtip3ovadfrq4hr9jta3v4q09ji1t04h4c4mh5cgi3/cmd/ateapi/internal/controlapi/workflow_pause.go#L78)
MarkPausingStep.Execute returns nil without changing state when status isn't RUNNING (e.g. an already-SUSPENDED actor). The workflow then proceeds to CallAteletPauseStep, which fails with "actor is in PAUSING state but has no active worker" — misleading, since it was never marked PAUSING. Consider rejecting non-RUNNING actors up front with an InvalidArgument/FailedPrecondition.

cmd/atelet/main.go
```diff
@@ -495,6 +526,70 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 	return &ateletpb.RestoreResponse{}, nil
 }
 
+func (s *AteomHerder) copyLocalCheckpoint(_ context.Context, snapshtoPrefix string, srcDir, dstDir string) error {
+	for _, fileName := range []string{"checkpoint.img", "pages.img", "pages_meta.img"} {
+		src := filepath.Join(srcDir, snapshtoPrefix, fileName)
```
BenTheElder: nit, typo:
```suggestion
func (s *AteomHerder) copyLocalCheckpoint(_ context.Context, snapshotPrefix string, srcDir, dstDir string) error {
	for _, fileName := range []string{"checkpoint.img", "pages.img", "pages_meta.img"} {
		src := filepath.Join(srcDir, snapshotPrefix, fileName)
```

cmd/atelet/main.go
```diff
@@ -495,6 +526,70 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 	return &ateletpb.RestoreResponse{}, nil
 }
 
+func (s *AteomHerder) copyLocalCheckpoint(_ context.Context, snapshtoPrefix string, srcDir, dstDir string) error {
```
BenTheElder: We should probably respect the context, in case we are copying a large snapshot? //TODO?

cmd/ateapi/internal/controlapi/functional_test.go
```diff
@@ -1123,14 +1124,112 @@ func TestSuspendActor(t *testing.T) {
...
+// 1. Creates a mock ActorTemplate.
+// 2. Creates a mock Atelet Pod on 'node1'.
+// 3. Creates a mock worker Pod on 'node1'.
+// 4. Waits for the WorkerPoolSyncer to mirror the worker to Redis.
+// 5. Creates an actor.
+// 6. Calls ResumeActor to transition it to RUNNING.
+// 7. Calls PauseActor RPC.
+// 8. Verifies that the fake Atelet received the Pause call.
+func TestPauseActor(t *testing.T) {
+	ns := namespaceForTest("ns-suspend")
```
BenTheElder: Should we use say, `ns-pause` ?

cmd/ateapi/internal/controlapi/functional_test.go
```diff
@@ -1123,14 +1124,112 @@ func TestSuspendActor(t *testing.T) {
...
+
+	// Resume first to make it running
+	_, err = tc.client.ResumeActor(context.Background(), &ateapipb.ResumeActorRequest{
+		ActorId: id,
+	})
+	if err != nil {
+		t.Fatalf("ResumeActor failed: %v", err)
+	}
+
+	// Suspend
```
BenTheElder: nit, copy-paste?

```suggestion
	// Pause
```

BenTheElder (commented):
Mostly nits inline.
 
- Can we add coverage for this to the e2e tests? I think this should work on gvisor+kind and it would be nice to cover in CI. We could extend the demo counter e2e.
- Worker.node_name is only populated when the syncer re-mirrors a pod, so workers created before this rollout won't satisfy the node restriction until re-synced.
- I know the storage situation is temporary, but we have no GC for snapshots and these may fill local disk quickly. Maybe a TODO somewhere or a doc warning?

cmd/ateapi/internal/controlapi/functional_test.go
dberkov: fixed. Thanks

cmd/ateapi/internal/controlapi/functional_test.go
dberkov: fixed.

cmd/atelet/main.go
dberkov: I added a validation for ctx.Err() in between steps. In any case this code is temporary and going to be replaced with upcoming change that @msau42 is working on in next 2 weeks.

cmd/atelet/main.go
dberkov: fixed. Thanks

internal/ateompath/ateompath.go
dberkov: removed

cmd/ateapi/internal/controlapi/workflow_pause.go
dberkov: I will fix it in one of the following PRs when I will introduce a CRASHED state.

cmd/ateapi/internal/controlapi/workflow_pause.go
dberkov: It is definitely an issue and error thrown is wrong. Same behavior exists today in SUSPEND and RESUME operations. They do not handle transition correctly.  

I created https://github.com/agent-substrate/substrate/issues/244 to fix it.

dberkov:
>Can we add coverage for this to the e2e tests? I think this should work on gvisor+kind and it would be nice to cover in CI. We could extend the demo counter e2e.

I have extended existing e2e demo test to test the PAUSE API.


>Worker.node_name is only populated when the syncer re-mirrors a pod, so workers created before this rollout won't satisfy the node restriction until re-synced.

Added a disclaimer in the PR description.

>I know the storage situation is temporary, but we have no GC for snapshots and these may fill local disk quickly. Maybe a TODO somewhere or a doc warning?

Added TODO in the code.

internal/e2e/suites/demo/demo_test.go
```diff
@@ -26,18 +26,242 @@ import (
...
+	if _, err := clients.SubstrateAPI.GetActor(ctx, &ateapipb.GetActorRequest{
+		ActorId: actorID,
+	}); err == nil {
+		t.Fatalf("expected actor %q to be deleted, but it still exists", actorID)
+	}
+
+	return nil
+}
+
+func suspendActor(ctx context.Context, t *testing.T, clients *e2e.Clients, nsObj *e2e.Namespace, at *v1alpha1.ActorTemplate) error {
```
BenTheElder, line 216: this appears to be the same as `pauseActor`?

cmd/ateapi/internal/controlapi/workflow.go
```diff
@@ -191,6 +191,35 @@ func (w *ActorWorkflow) SuspendActor(ctx context.Context, id string) (*ateapipb.
...
 
+// PauseActor executes the workflow to pause a running actor. Idempotent.
+func (w *ActorWorkflow) PauseActor(ctx context.Context, id string) (*ateapipb.Actor, error) {
+	input := &PauseInput{
+		ActorID: id,
+	}
+	state := &PauseState{}
+
+	// Acquire lock and get the timeout context for the workflow
+	// Lock TTL is 7 seconds, with 2 seconds padding for workflow timeout
```
BenTheElder:
```suggestion
	// Lock TTL is 30 seconds, with 2 seconds padding for workflow timeout
```

internal/e2e/suites/demo/demo_test.go
```diff
@@ -26,18 +26,242 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func TestDemo3(t *testing.T) {
-	env, err := e2e.CheckEnv("BUCKET_NAME", "KO_DOCKER_REPO")
-	if err != nil {
-		t.Fatalf("CheckEnv failed: %v", err)
-	}
-
+func TestActorLifecycle(t *testing.T) {
```
BenTheElder, line 37: TODO for follow-up ... we are not vetting that we actually preserved memory.

internal/e2e/suites/demo/demo_test.go
dberkov: oops. Fixed

internal/e2e/suites/demo/demo_test.go
dberkov: I extend the test cases to actually call to actor and check the counter's value been preserved after suspend and pause operations.

cmd/ateapi/internal/controlapi/workflow.go
dberkov: @juli4n I copied the comment and content of 30 sec from suspend workflow. The comment says "7" sec. Is it a comment incorrect or the actual timeout set to 30 sec instead of 7?

juli4n: It looks like the comment is outdated. It used to be 7 seconds, then it was bumped to 30. See https://github.com/agent-substrate/substrate/blame/808806db7f7b19304a44a491730004239110ef04/cmd/ateapi/internal/controlapi/workflow.go#L145.

dberkov: comments are updated.

BenTheElder:
needs a rebase

dberkov:
> needs a rebase

Done

BenTheElder (approved):

PR #234, opened 2026-06-12
fix: Allow large image pulls to succeed by decoupling workflow deadline from lock TTL via heartbeat

### LINKED ISSUE #233: Allow large image pulls to succeed by decoupling workflow deadline from lock TTL
## Expected Behavior

User Facing: I should be able to pull a large image and schedule it to a worker.
(such as one containing a large number of executables).

Technical: `ResumeActor` and `SuspendActor` should succeed for any actor whose image pull + extract + gVisor restore can reasonably complete, including cold pulls on slow networks, large container images, and first-time pulls on freshly scheduled worker pods.
The Redis lock TTL that serializes `ResumeActor`/`SuspendActor` on a single actor (so a crashed `ateapi` replica doesn't pin the actor on a stale lock) should be tunable independently of how long an individual workflow is allowed to run.

## Actual Behavior
User Facing: Prior to #231, large images (specifically our test image) would eventually finish pulling in  ~20 min, at the cost of RSS increasing in ateleti.e. worker successfully has actor scheduled to it at the  cost of overall system stability. This wasn't an acceptable tradeoff. Now the control plane is no longer at risk of  OOM, but still needs to successfully pull larger images.

Technical:
`ActorWorkflow.ResumeActor` and `SuspendActor` derive the workflow context from the Redis lock TTL via:

```go
ctx, releaseLock, err := w.acquireActorLock(ctx, id, 30*time.Second, 2*time.Second)
```

at `cmd/ateapi/internal/controlapi/workflow.go:145` (and `:174`). `acquireActorLock` makes the workflow ctx expire at `lockTTL − padding = 28s`, conflating two unrelated concerns into a single knob:

1. **How long a single Resume/Suspend workflow is allowed to run.** Bounded by what the slowest legitimate path (image pull + decompress + extract + gVisor restore) actually needs.
2. **How long a peer must wait to retry an actor after a crashed `ateapi` mid-workflow.** Bounded by tolerable failover latency.

Raising the knob to fix (1) makes (2) unacceptably slow — a crashed `ateapi` would pin an individual actor for minutes. Lowering it to keep (2) fast breaks (1) — image pulls that can't complete inside 28s death-loop forever. There is no operator-facing knob to raise either.

After #231 (the Bug B / RSS amplification fix on parent #230), the loop is no longer an OOM risk to `atelet`, but the user-visible symptom is unchanged:

- `kubectl ate get actor <golden-actor-id>` stays `STATUS_RESUMING`.
- `kubectl get actortemplate ... -o jsonpath='{.status.phase}'` stays `PhaseResumeGoldenActor`.
- `ate-controller` logs spray `DeadlineExceeded` every ~30s.
- The workflow makes no progress — and after #231 it is *less* likely to escape via the previously-observed "incidental success" mechanism (bandwidth variance + warmed HTTP keep-alives), because cancelled requests now drop their underlying TCP connections cleanly instead of leaving them warm for the next retry.

Most commonly hit when:

- Resuming an actor for the first time on a freshly scheduled worker pod.
- Resuming on a fresh cluster (no upstream registry caching, `atelet` in-process pull cache empty).
- ActorTemplate uses a large image (north of ~200MB compressed on non-LAN bandwidth).
- Self-hosted or rate-limited image registries.
- After a worker reboot or `atelet` restart (in-process pull cache lost).

Identical underlying reproduction to #230; the death loop happens whether or not #231 is in place — what changes is whether `atelet` RSS amplifies during it.

1. Set up kind + install substrate:

   ```bash
   ./hack/create-kind-cluster.sh
   ./hack/install-ate-kind.sh --deploy-ate-system
   ```

1. Build the `ateom-gvisor` image into the kind-local registry and capture the resolved reference:

   ```bash
   export KO_DOCKER_REPO=localhost:5001
   export KO_DEFAULTPLATFORMS=linux/$(go env GOARCH)
   ATEOM_IMAGE=$(./hack/run-tool.sh ko build -B ./cmd/ateom-gvisor)
   echo "ateom image: $ATEOM_IMAGE"   # localhost:5001/ateom-gvisor@sha256:...
   ```

1. Install metrics-server (needed for `kubectl top` on a single-node kind cluster):

   ```bash
   kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/high-availability-1.21+.yaml
   kubectl scale -n kube-system deploy/metrics-server --replicas=1
   kubectl patch -n kube-system deploy metrics-server --type=json \
     -p='[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--kubelet-insecure-tls"}]'
   kubectl rollout status -n kube-system deploy/metrics-server
   ```

1. Apply the WorkerPool + ActorTemplate. The container image is a ~360MB compressed / ~1.2GB rootfs sandbox image observed in the field; the container's `command` is irrelevant because the death loop happens during pull/extract, before the container ever starts:

   ```bash
   kubectl apply -f - <<EOF
   apiVersion: v1
   kind: Namespace
   metadata:
     name: ate-repro
   ---
   apiVersion: ate.dev/v1alpha1
   kind: WorkerPool
   metadata:
     name: repro-pool
     namespace: ate-repro
   spec:
     replicas: 1
     ateomImage: ${ATEOM_IMAGE}
   ---
   apiVersion: ate.dev/v1alpha1
   kind: ActorTemplate
   metadata:
     name: repro-fat-image
     namespace: ate-repro
   spec:
     workerPoolRef:
       name: repro-pool
       namespace: ate-repro
     runsc:
       amd64:
         url: "gs://gvisor/releases/nightly/2026-05-19/x86_64/runsc"
         sha256Hash: "a397be1abc2420d26bce6c70e6e2ff96c73aaaab929756c56f5e2089ea842b63"
       arm64:
         url: "gs://gvisor/releases/nightly/2026-05-19/aarch64/runsc"
         sha256Hash: "1ba2366ae2efceba166046f51a4104f9261c9cb72c6db8f5b3fe2dc57dea86b9"
     pauseImage: "registry.k8s.io/pause:3.10.2@sha256:f548e0e8e3dc1896ca956272154dde3314e8cc4fde0a57577ee9fa1c63f5baf4"
     containers:
       - name: fat
         image: ghcr.io/kagent-dev/nemoclaw/sandbox-base@sha256:d52bee415dc4c0dba7164f9eabe727574c056d4f211781f20af249707883a3b4
         command: ["/bin/sh", "-c", "sleep 3600"]
     snapshotsConfig:
       location: gs://ate-snapshots/repro/
   EOF
   ```

   No manual `kubectl ate resume actor` is needed — `ate-controller` calls `ResumeActor` against the golden-actor ID as part of the ActorTemplate reconcile, which is enough to exercise the bug.

1. Confirm the failure mode:

   ```bash
   # ate-controller retries every ~30s with DeadlineExceeded indefinitely
   kubectl logs -n ate-system deploy/ate-controller --tail=0 -f | grep -i "deadline\|golden"

   # AT never leaves ResumeGoldenActor
   kubectl get actortemplate -n ate-repro repro-fat-image -o jsonpath='{.status.phase}{"\n"}'

   # Golden actor never leaves STATUS_RESUMING
   GOLDEN_ID=$(kubectl get actortemplate -n ate-repro repro-fat-image -o jsonpath='{.status.goldenActorID}')
   kubectl ate get actor "$GOLDEN_ID"
   ```

> **If the bug doesn't trigger on your bandwidth:** the cold-pull path completes inside 28s on fast corp LAN / warmed registry mirrors. Force it by throttling the kind node (`docker exec kind-control-plane tc qdisc add dev eth0 root tbf rate 10mbit burst 32kbit latency 400ms` — `apt-get install -y iproute2` first if `tc` is missing) or by swapping the container `image:` for something much larger (e.g. `nvcr.io/nvidia/pytorch:24.05-py3` at ~12GB compressed). Watch for at least 5 minutes before concluding it didn't reproduce — the loop occasionally escapes via bandwidth variance.

## Specifications

- Version: HEAD of `main` (#231 merged or unmerged — Bug A behaves identically either way).
- Platform: any cluster where the chosen image's cold-pull path exceeds ~28s. Reproduced on kind (single-node, control-plane only) on darwin/arm64.

## Proposed Fix

Decouple the Redis lock TTL from the workflow deadline via a heartbeat. Sketch:

1. **Add `RefreshLock(ctx, key, value, ttl) (bool, error)` to `store.Interface`.** Redis implementation is a CAS Lua script — `GET key == value ? PEXPIRE key ttl : 0` — mirroring the existing `ReleaseLock` script's shape at `cmd/ateapi/internal/store/ateredis/ateredis.go:577`. Returns `false` if we no longer own the lock so the caller can abort.

2. **Keep the Redis lock TTL short** — internal constant (~30s). This is the only thing that bounds how long a peer waits to retry an actor after a crashed `ateapi`.

3. **Make the workflow deadline a separate, operator-configurable knob.** New `--actor-workflow-deadline` pflag on `ateapi` with a 5-minute default — long enough for cold cluster pulls on typical bandwidth, short enough that hung workflows stay bounded for ops. This is what bounds a single Resume/Suspend.

4. **Spawn a heartbeat goroutine on lock acquire.** Refreshes the lock every `lockTTL/3` (~10s) for the full workflow duration. On `RefreshLock=false` or any Redis error (peer stole the lock, Redis blip), cancel the workflow ctx with a distinguishable cause (`errLostActorLock`) so in-flight workflow steps see `ctx.Err()` and tear down cleanly. Preserves the original mutual-exclusion invariant — only one workflow runs per actor at any instant.

5. **`releaseLock()` stops the heartbeat first** (waits for goroutine exit), then best-effort `ReleaseLock`.

### Customer-visible impact after fix

- Resume on a cold node completes in one call (taking however long the pull legitimately takes), instead of looping forever.
- Crash-recovery latency unchanged — peer failover stays bounded by the ~30s Redis lock TTL.
- A workflow torn down because its peer stole the lock returns a specific `errLostActorLock` cause, useful for triaging flapping `ateapi` replicas.

### Why this shape (not alternatives)

- *Just raise the 28s knob.* Rejected — operators do not want multi-minute failover delays after a crashed `ateapi`.
- *Heartbeat only, no server-side workflow deadline (workflow ctx inherits from caller).* Considered: `ate-controller` and other gRPC callers already carry their own ctx, so in the happy path the heartbeat alone is enough for large pulls to succeed. Rejected on defense-in-depth grounds — substrate does not configure gRPC keepalive, so a network partition or a hung-but-not-killed caller would leave the workflow running indefinitely on `ateapi`; and a workflow step that forgets to propagate ctx would have no upstream timeout to fall back on. The 5-minute server-side cap covers both cases at the cost of one config knob.
- *Move image pulls off the workflow path entirely* (background pull worker + status-polling RPC). The "right" long-term fix, but a much bigger redesign; should follow the on-disk/shared layer cache (`internal/memorypullcache/memorypullcache.go:47-58` TODO and `ategcs.ObjectStorage`). The heartbeat fix is small enough to ship now without prejudicing that work.

jjamroga:
Fixes #233.

Customer Impact: Prior to #230, large images (specifically our test image) would eventually finish pulling in ~20 min, at the cost of RSS increasing in atelet i.e. worker successfully has actor scheduled to it at the cost of overall system stability. This wasn't an acceptable tradeoff. Orphaned image pulls resulted in atelet consuming almost 10 GiB of memory before being garbage collected.

Now, long running image pulls have a process of reacquiring a lock, preventing it from expiring before the image pull has completed. Introduced a high level, user configurable timeout ensuring that there is a window in which the lock can be lost and be lost in the event of image pull hanging (instead of a similar bug where a process is left hanging despite the caller having abandoned it.)

## Summary

Decouples the per-actor workflow deadline from the Redis lock TTL via a heartbeat, so large image pulls can finish without raising the lock TTL (which also bounds peer failover after a crashed `ateapi`).

**Lock TTL** (~30s) bounds *failover latency* — how long a peer `ateapi` waits to retry an actor if the original instance crashes mid-workflow. **Workflow deadline** (~5min) bounds *legitimate work* — how long a single Resume/Suspend/Pause is allowed to run end-to-end before we give up. The **lock refresh** (heartbeat goroutine, PEXPIREs every TTL/3) is the bridge: it keeps the lock alive across the long workflow window without raising the TTL itself, so we get both fast failover AND room for slow legitimate operations.

**Lifecycle RPCs** (Resume/Suspend/Pause) all go through `acquireActorLock` on entry. That call (a) grabs the Redis lock with the short TTL, (b) returns a workflow-bounded ctx via `WithTimeout(workflowDeadline)`, and (c) spawns the heartbeat. While the RPC runs, the lock stays held; if the heartbeat ever fails or returns "you no longer own this lock," it cancels the workflow ctx so steps tear down promptly. On normal return, release stops the heartbeat and frees the lock. Mutual exclusion per actor holds throughout — exactly one workflow per actor at any instant, just like before, but no longer pinned to a 28-second ceiling.

## Why a server-side deadline at all (not just heartbeat)

Considered keeping only the heartbeat and letting the workflow ctx inherit from the caller. Rejected on defense-in-depth grounds:

- Substrate does not configure gRPC keepalive, so a network partition or a hung-but-not-killed caller would leave the workflow running indefinitely on `ateapi`.
- A workflow step that forgets to propagate ctx would have no upstream timeout to fall back on.

The 5-minute server-side cap covers both cases at the cost of one config knob. Operators can raise it for slow registries; the lock TTL is independent so peer failover stays bounded at ≤30s regardless.

## Test plan

Automated:

- [x] `go test ./cmd/ateapi/internal/store/ateredis/... -run 'TestRefreshLock|TestAcquireLock'` — 7 lock tests pass (3 new + 4 existing).
- [x] `go test ./cmd/ateapi/internal/controlapi/ -run TestAcquireActorLock` — 4 new heartbeat tests pass:
  - Heartbeat keeps lock alive past TTL.
  - Lost lock cancels workflow ctx with `errLostActorLock` cause.
  - Release callback removes the lock.
  - Second concurrent acquire returns `codes.Aborted`.
- [x] `go test ./... -count=1` — full repo suite green (419 tests / 65 packages).
- [x] `go vet ./...` — clean.

Manual:

```aiignore
# Setup cluster
./hack/create-kind-cluster.sh && ./hack/install-ate-kind.sh --deploy-ate-system
export KO_DOCKER_REPO=localhost:5001
export KO_DEFAULTPLATFORMS=linux/$(go env GOARCH)
./hack/run-tool.sh ko build -B ./cmd/ateom-gvisor

# Install metrics server to keep tabs on memory
kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/high-availability-1.21+.yaml
# Tweak values for cluster
kubectl patch -n kube-system deploy metrics-server --type=json -p='[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--kubelet-insecure-tls"}]'
kubectl scale -n kube-system deploy/metrics-server --replicas=1
kubectl rollout status deploy/metrics-server -n kube-system

# Watch memory consumption
while true; do date; kubectl top pod -n ate-system -l app=atelet --containers >> bin/output-fix.log; sleep 5; done

# Check atelet, image pull starts and eventually completes in a reasonable amount of time
kubectl logs daemonset/atelet -n ate-system --follow
{"time":"2026-06-12T17:33:16.609666504Z","level":"INFO","msg":"Using S3 storage backend"}
{"time":"2026-06-12T17:33:16.61079017Z","level":"INFO","msg":"Starting Prometheus metrics server on :9090"}
{"time":"2026-06-12T17:33:16.626096171Z","level":"INFO","msg":"WorkersManagerService listening","address":{"IP":"::","Port":8085,"Zone":""}}
{"time":"2026-06-12T17:36:40.508440418Z","level":"INFO","msg":"Ref includes digest, checking for cache hit","ref":"ghcr.io/kagent-dev/nemoclaw/sandbox-base@sha256:d52bee415dc4c0dba7164f9eabe727574c056d4f211781f20af249707883a3b4","digest":"sha256:d52bee415dc4c0dba7164f9eabe727574c056d4f211781f20af249707883a3b4","ate.dev/trace-id":"f2bd930bd82771821f8d5fcabcb895a0"}
{"time":"2026-06-12T17:36:40.508567918Z","level":"INFO","msg":"Ref includes digest, checking for cache hit","ref":"registry.k8s.io/pause:3.10.2@sha256:f548e0e8e3dc1896ca956272154dde3314e8cc4fde0a57577ee9fa1c63f5baf4","digest":"sha256:f548e0e8e3dc1896ca956272154dde3314e8cc4fde0a57577ee9fa1c63f5baf4","ate.dev/trace-id":"f2bd930bd82771821f8d5fcabcb895a0"}
{"time":"2026-06-12T17:36:40.508644543Z","level":"INFO","msg":"Cache miss","ref":"registry.k8s.io/pause:3.10.2@sha256:f548e0e8e3dc1896ca956272154dde3314e8cc4fde0a57577ee9fa1c63f5baf4","ate.dev/trace-id":"f2bd930bd82771821f8d5fcabcb895a0"}
{"time":"2026-06-12T17:36:40.508595126Z","level":"INFO","msg":"Cache miss","ref":"ghcr.io/kagent-dev/nemoclaw/sandbox-base@sha256:d52bee415dc4c0dba7164f9eabe727574c056d4f211781f20af249707883a3b4","ate.dev/trace-id":"f2bd930bd82771821f8d5fcabcb895a0"}
{"time":"2026-06-12T17:38:35.082124179Z","level":"INFO","msg":"Handle RPC","method":"/atelet.AteomHerder/Run","req":{"target_ateom_uid":"739d5e60-905f-4681-bac4-171190ab7cba","actor_template_namespace":"ate-repro","actor_template_name":"repro-fat-image","actor_id":"69bcf578-73af-48da-8cb8-c73f4b09622b","runsc":{"amd64":{"sha256_hash":"a397be1abc2420d26bce6c70e6e2ff96c73aaaab929756c56f5e2089ea842b63","url":"gs://gvisor/releases/nightly/2026-05-19/x86_64/runsc"},"arm64":{"sha256_hash":"1ba2366ae2efceba166046f51a4104f9261c9cb72c6db8f5b3fe2dc57dea86b9","url":"gs://gvisor/releases/nightly/2026-05-19/aarch64/runsc"}},"spec":{"containers":[{"name":"fat","image":"ghcr.io/kagent-dev/nemoclaw/sandbox-base@sha256:d52bee415dc4c0dba7164f9eabe727574c056d4f211781f20af249707883a3b4","command":["/bin/sh","-c","sleep 3600"]}],"pause_image":"registry.k8s.io/pause:3.10.2@sha256:f548e0e8e3dc1896ca956272154dde3314e8cc4fde0a57577ee9fa1c63f5baf4"}},"resp":{},"err":null,"elapsed-time":"2m0.322655054s","ate.dev/trace-id":"f2bd930bd82771821f8d5fcabcb895a0"}
{"time":"2026-06-12T17:38:55.689723883Z","level":"INFO","msg":"Handle RPC","method":"/atelet.AteomHerder/Checkpoint","req":{"target_ateom_uid":"739d5e60-905f-4681-bac4-171190ab7cba","actor_template_namespace":"ate-repro","actor_template_name":"repro-fat-image","actor_id":"69bcf578-73af-48da-8cb8-c73f4b09622b","runsc":{"amd64":{"sha256_hash":"a397be1abc2420d26bce6c70e6e2ff96c73aaaab929756c56f5e2089ea842b63","url":"gs://gvisor/releases/nightly/2026-05-19/x86_64/runsc"},"arm64":{"sha256_hash":"1ba2366ae2efceba166046f51a4104f9261c9cb72c6db8f5b3fe2dc57dea86b9","url":"gs://gvisor/releases/nightly/2026-05-19/aarch64/runsc"}},"spec":{"containers":[{"name":"fat","image":"ghcr.io/kagent-dev/nemoclaw/sandbox-base@sha256:d52bee415dc4c0dba7164f9eabe727574c056d4f211781f20af249707883a3b4","command":["/bin/sh","-c","sleep 3600"]}],"pause_image":"registry.k8s.io/pause:3.10.2@sha256:f548e0e8e3dc1896ca956272154dde3314e8cc4fde0a57577ee9fa1c63f5baf4"},"snapshot_uri_prefix":"gs://ate-snapshots/repro/69bcf578-73af-48da-8cb8-c73f4b09622b/2026-06-12T17:38:55Z-XHXXXIU65H5YZCKRHLGC3XH6F4"},"resp":{},"err":null,"elapsed-time":"685.691709ms","ate.dev/trace-id":"41774e0b1b60502ad62d6351bafd4a1b"}
```

## Out of scope / follow-ups

- Moving image pulls off the workflow path entirely (background pull worker + status-polling RPC) is the long-term fix per `internal/memorypullcache/memorypullcache.go:47-58`. Should follow the on-disk / shared layer cache work via `ategcs.ObjectStorage`. The heartbeat shape introduced here does not preclude that redesign.
- gRPC keepalive on `ateapi` server + client sides would tighten the partition-detection story. Left as a separate change.

### DIFF
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -310,7 +310,7 @@ func setupTest(t *testing.T, ns string) *testContext {
 	}
 
 	dialer := NewAteletDialer(workerInformer.GetIndexer(), ateletInformer.GetIndexer())
-	service := NewService(persistence, wc, actorTemplateLister, workerPoolLister, sandboxConfigLister, dialer, k8sClient)
+	service := NewService(persistence, wc, actorTemplateLister, workerPoolLister, sandboxConfigLister, dialer, k8sClient, 30*time.Second)
 
 	// 5. Start REAL gRPC Server for ATE API
 	grpcServer := grpc.NewServer(grpc.UnaryInterceptor(ateinterceptors.ServerUnaryInterceptor))
--- a/cmd/ateapi/internal/controlapi/service.go
+++ b/cmd/ateapi/internal/controlapi/service.go
@@ -15,6 +15,8 @@
 package controlapi
 
 import (
+	"time"
+
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/workercache"
 	listersv1alpha1 "github.com/agent-substrate/substrate/pkg/client/listers/api/v1alpha1"
@@ -34,7 +36,8 @@ type Service struct {
 
 var _ ateapipb.ControlServer = (*Service)(nil)
 
-// NewService creates a service.
+// NewService creates a service. actorWorkflowDeadline bounds how long a single
+// Resume/Suspend workflow can run end-to-end.
 func NewService(
 	persistence store.Interface,
 	workerCache *workercache.Cache,
@@ -43,13 +46,14 @@ func NewService(
 	sandboxConfigLister listersv1alpha1.SandboxConfigLister,
 	dialer *AteletDialer,
 	kubeClient kubernetes.Interface,
+	actorWorkflowDeadline time.Duration,
 ) *Service {
 	s := &Service{
 		persistence:         persistence,
 		actorTemplateLister: actorTemplateLister,
 		workerPoolLister:    workerPoolLister,
 		dialer:              dialer,
-		actorWorkflow:       NewActorWorkflow(persistence, workerCache, dialer, actorTemplateLister, workerPoolLister, sandboxConfigLister, kubeClient),
+		actorWorkflow:       NewActorWorkflow(persistence, workerCache, dialer, actorTemplateLister, workerPoolLister, sandboxConfigLister, kubeClient, actorWorkflowDeadline),
 	}
 
 	return s
--- a/cmd/ateapi/internal/controlapi/workflow.go
+++ b/cmd/ateapi/internal/controlapi/workflow.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"time"
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
@@ -32,6 +33,18 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 )
+
+// actorLockTTL is the Redis TTL on the per-actor workflow lock. It bounds how
+// long a peer must wait to retry an actor after this process crashes mid-workflow.
+const actorLockTTL = 30 * time.Second
+
+// actorLockHeartbeatInterval is how often the heartbeat refreshes the lock.
+// Chosen so we get ~3 attempts before the TTL would otherwise lapse.
+const actorLockHeartbeatInterval = actorLockTTL / 3
+
+// errLostActorLock is the context cause set when the heartbeat can no longer
+// keep the actor lock alive (peer stole it, or Redis returned an error).
+var errLostActorLock = errors.New("lost actor lock during workflow")
 
 // WorkflowStep represents a single, idempotent operation in a workflow graph.
 // Params is the immutable parameters used to start the workflow.
@@ -123,9 +136,14 @@ type ActorWorkflow struct {
 	sandboxConfigLister listersv1alpha1.SandboxConfigLister
 	kubeClient          kubernetes.Interface
 	secretCache         *envSecretCache
+	// workflowDeadline is the maximum duration of a single Resume/Suspend
+	// workflow. The lock is kept alive across this duration by a heartbeat,
+	// independent of actorLockTTL.
+	workflowDeadline time.Duration
 }
 
-// NewActorWorkflow creates a new ActorWorkflow.
+// NewActorWorkflow creates a new ActorWorkflow. workflowDeadline bounds how
+// long a single Resume/Suspend can run end-to-end.
 func NewActorWorkflow(
 	store store.Interface,
 	workerCache *workercache.Cache,
@@ -134,6 +152,7 @@ func NewActorWorkflow(
 	workerPoolLister listersv1alpha1.WorkerPoolLister,
 	sandboxConfigLister listersv1alpha1.SandboxConfigLister,
 	kubeClient kubernetes.Interface,
+	workflowDeadline time.Duration,
 ) *ActorWorkflow {
 	return &ActorWorkflow{
 		store:               store,
@@ -144,6 +163,7 @@ func NewActorWorkflow(
 		sandboxConfigLister: sandboxConfigLister,
 		kubeClient:          kubeClient,
 		secretCache:         newEnvSecretCache(envSecretCacheTTL),
+		workflowDeadline:    workflowDeadline,
 	}
 }
 
@@ -156,9 +176,7 @@ func (w *ActorWorkflow) ResumeActor(ctx context.Context, atespace, name string, 
 	}
 	state := &ResumeState{}
 
-	// Acquire lock and get the timeout context for the workflow
-	// Lock TTL is 30 seconds, with 2 seconds padding for workflow timeout
-	ctx, releaseLock, err := w.acquireActorLock(ctx, atespace, name, 30*time.Second, 2*time.Second)
+	ctx, releaseLock, err := w.acquireActorLock(ctx, atespace, name, actorLockTTL, actorLockHeartbeatInterval)
 	if err != nil {
 		return nil, err
 	}
@@ -186,9 +204,7 @@ func (w *ActorWorkflow) SuspendActor(ctx context.Context, atespace, name string)
 	}
 	state := &SuspendState{}
 
-	// Acquire lock and get the timeout context for the workflow
-	// Lock TTL is 30 seconds, with 2 seconds padding for workflow timeout
-	ctx, releaseLock, err := w.acquireActorLock(ctx, atespace, name, 30*time.Second, 2*time.Second)
+	ctx, releaseLock, err := w.acquireActorLock(ctx, atespace, name, actorLockTTL, actorLockHeartbeatInterval)
 	if err != nil {
 		return nil, err
 	}
@@ -216,9 +232,7 @@ func (w *ActorWorkflow) PauseActor(ctx context.Context, atespace, name string) (
 	}
 	state := &PauseState{}
 
-	// Acquire lock and get the timeout context for the workflow
-	// Lock TTL is 30 seconds, with 2 seconds padding for workflow timeout
-	ctx, releaseLock, err := w.acquireActorLock(ctx, atespace, name, 30*time.Second, 2*time.Second)
+	ctx, releaseLock, err := w.acquireActorLock(ctx, atespace, name, actorLockTTL, actorLockHeartbeatInterval)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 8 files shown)


jjamroga:
@bowei Would you potentially be able to take a look at this? This is followup work to that memory leak issue I put a PR up for.

Issue #170, opened 2026-06-03
Secure mTLS authentication and authorization between all ate-system components.


zoez7:

ahmedtd:
@zoez7 Looks like this is a dupe (ish) of #164

mtaufen:
For this, a few basic things for alpha would be good:
- ensure mTLS certs are there for all ate system components
- ensure the trust bundles exposed to those components are limiting trust to that system set (could be expanded later to support more clients w/ mTLS, but let's start with this)
- hardcode some basic authz policy, like "this is what's reasonable for atelet to call in the API", in the system components; this can be expanded into a more configurable rules system later

mtaufen:
Agree this goes together with #164

zoez7:
https://github.com/agent-substrate/substrate/pull/237 has merged, ate apiserver now requires client to authenticate either with a client cert, or a k8s JWT. No authorization yet.

EItanya:
I think we can close this as done

PR #237, opened 2026-06-12
Added mTLS between ate system components

zoez7:
Part of [#170](https://github.com/agent-substrate/substrate/issues/170)
Establishes mutual TLS between all ate system components, and updates the certificate plumbing it depends on.

Main changes:
1. The atenet router now verifies ate apiserver' serving certificate, and presents its client cert to ate apiserver. Previously the connection used `InsecureSkipVerify`.
2. AteApi server verifies atelet's serving cert.

Minor bug fixes:
1. Prevent `servicednssigner` from signing a cert with no DNS SANs. Also updated valkey cluster's cert configuration, because it was relying on the cert with empty DNS.

- [x] Tests pass


ahmedtd:
Can you talk a little bit about the background of splitting out separate CAs / signers for the workerpools and ate-system components?

If at all possible, I would prefer to restrict us to two signers:
* One that issues certs with DNS SANs based on the K8s services that a pod is part of (though opt-in with an annotation or label is fine).
* One that issues SPIFFE certs that encode the k8s service account identity of the pod (and potentially, the pod ID, node ID etc in an additional extension).

The reason is that we can be reasonably confident that signers like these will land in upstream K8s in the near-ish future.  The certificates here aren't special or unique to substrate, and they aren't part of the value proposition of the project.  So I think we should plan to rebase on native K8s capabilities as they become available.  That's easiest if we don't get very custom with the signers.

zoez7:
I think the main motivations for splitting out separate CAs / signers was, if I mount the "podidentity" CA bundle as trusted client CAs in ate-system services, workerpool pods (therefore actors) would also be able to connect to ate-system servers like the Valkey cluster and atelet server.

But after typing it out loud - it seems this needs to be solved anyway (probably in a later milestone) by authorization inside servers such as valkey, instead of using separate CAs. Does that make sense to you?

ahmedtd:
> I think the main motivations for splitting out separate CAs / signers was, if I mount the "podidentity" CA bundle as trusted client CAs in ate-system services, workerpool pods (therefore actors) would also be able to connect to ate-system servers like the Valkey cluster and atelet server.
> 
> But after typing it out loud - it seems this needs to be solved anyway (probably in a later milestone) by authorization inside servers such as valkey, instead of using separate CAs. Does that make sense to you?

Yeah --- we should always draw a bright line between authentication and authorization.  Assuming that a given client is authorized just because they were able to get a certificate from a particular CA has caused a lot of security problems in the past.

Unfortunately, for the specific case of Valkey we might be limited by its capabilities, in which case it might require a separate signer.  Let me check the docs.

ahmedtd:
It looks like valkey can do it: https://valkey.io/topics/tls/#:~:text=Client%20certificate%20authentication

cmd/atenet/internal/router/router.go
```diff
@@ -62,22 +58,24 @@ func init() {
...
-	HealthInterval time.Duration
-	HttpsPort      int
-	EnvoyCertPath  string
-	LogLevel       string
-	MetricsAddr    string
+	Standalone           bool
+	Namespace            string
+	Kubeconfig           string
+	AteapiAddr           string
+	AteapiClientCertPath string
```
bowei: Can we group the cert and auth related config into its own struct? Makes it easier to read:

type AuthConfig struct {...}

cmd/atenet/internal/router/router.go
```diff
@@ -149,6 +152,53 @@ func NewRouterServer(cfg RouterConfig) (*RouterServer, error) {
...
+	}
+
+	return &tls.Config{
+		RootCAs:              rootCAs,
+		GetClientCertificate: credbundle.ClientLoader(cfg.AteapiClientCertPath),
+	}, nil
+}
+
+func fileExists(path string) bool {
+	if path == "" {
```
bowei: Do we need this case actually? Empty file will get a does not exist error anyway.

cmd/atenet/internal/router/router.go
```diff
@@ -149,6 +152,53 @@ func NewRouterServer(cfg RouterConfig) (*RouterServer, error) {
...
+		return false
+	}
+	_, err := os.Stat(path)
+	return err == nil
```
bowei: not clear we should eat the error if it's not a ENOENT error

cmd/atenet/internal/router/router.go
```diff
@@ -62,22 +58,24 @@ func init() {
 
 // RouterConfig holds deployment setup and endpoint options for the router node instance.
 type RouterConfig struct {
```
bowei: suggest:

* Move RouterConfig to its own file (config.go)
* Make it not exported (rename to routerConfig)
* make ateAPITransportCreds() a member function
* rename ateAPITransportCreds to apiTransportCredentials

cmd/atenet/internal/router/router.go
```diff
@@ -149,6 +152,53 @@ func NewRouterServer(cfg RouterConfig) (*RouterServer, error) {
...
+func ateapiTransportCreds(cfg RouterConfig) (credentials.TransportCredentials, error) {
+	tlsCfg, err := ateapiTLSConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return credentials.NewTLS(tlsCfg), nil
+}
+
+func ateapiTLSConfig(cfg RouterConfig) (*tls.Config, error) {
+	haveCA := fileExists(cfg.AteapiCACertsPath)
```
bowei: suggest: just call stat and return the error. If the file exists, it should stat() with no error anyway.

zoez7: Done. Returned error from os.Stat().

cmd/atenet/internal/router/router.go
zoez7: Done.

cmd/atenet/internal/router/router.go
zoez7: Done.

cmd/atenet/internal/router/router.go
zoez7: Deleted this entirely. Surfaced error from os.Stat.

cmd/atenet/internal/router/router.go
zoez7: Done.

cmd/atenet/internal/router/router.go
```diff
@@ -149,6 +129,46 @@ func NewRouterServer(cfg RouterConfig) (*RouterServer, error) {
...
 }
 
+// ateapiTransportCreds builds the TLS credentials the router uses to dial
+// ateapi. When both the servicedns trust bundle and the podidentity client
+// credential bundle are present (the in-cluster case, mounted via projected
+// pod-certificate volumes), it performs mTLS: it verifies ateapi's serving cert
+// against the servicedns trust bundle and presents its own podidentity SPIFFE
+// client cert. When that material is absent, it returns an error rather than
+// falling back to an insecure connection.
+func (cfg routerConfig) apiTransportCredentials() (credentials.TransportCredentials, error) {
```
bowei: use pointer receiver: cfg *routerConfig

put this func in config.go

cmd/atenet/internal/router/router.go
```diff
@@ -149,6 +129,46 @@ func NewRouterServer(cfg RouterConfig) (*RouterServer, error) {
...
+	tlsCfg, err := apiTLSConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return credentials.NewTLS(tlsCfg), nil
+}
+
+func apiTLSConfig(cfg routerConfig) (*tls.Config, error) {
```
bowei: put this in config.go 

make this a method?

bowei (commented):
some nits

cmd/atenet/internal/router/router.go
zoez7: Done.

cmd/atenet/internal/router/router.go
zoez7: Done.

cmd/atenet/internal/router/config.go
```diff
@@ -12,25 +12,85 @@
...
+			cfg.Auth.AteapiClientCertPath, err)
+	}
+
+	return &tls.Config{
+		RootCAs:              rootCAs,
+		GetClientCertificate: credbundle.ClientLoader(cfg.Auth.AteapiClientCertPath),
+	}, nil
+}
+
 func NewRouterCmd() *cobra.Command {
```
bowei: move this to cmd.go

zoez7: Done.

bowei (commented):
Router changes look ok.

bowei:
needs rebase

bowei:
Ping on this?

zoez7:
Sorry for the delay! Rebased and force-pushed this branch, also added the mTLS between other ate system components as well.

cmd/ateapi/internal/controlapi/dialer.go
```diff
@@ -97,3 +127,68 @@ func (d *AteletDialer) DialForWorker(workerPodNamespace, workerPodName string) (
...
+	if err != nil {
+		return nil, err
+	}
+
+	tlsConfig := tls.Config{
+		MinVersion:           tls.VersionTLS12,
+		GetClientCertificate: credbundle.ClientLoader(clientBundlePath),
+		// Skip the default verification because the peer is dialed by IP and its
+		// certificate has no DNS/IP SAN.
+		InsecureSkipVerify: true,
```
mtaufen, line 158: Should we include DNS or IP SANs in the certs so we don't need to skip this? https://spiffe.io/docs/latest/spiffe-specs/x509-svid/#2-spiffe-id says:

> An X.509 SVID MAY contain any number of other SAN field types, including DNS SANs.

cmd/ateapi/internal/controlapi/dialer.go
```diff
@@ -97,3 +127,68 @@ func (d *AteletDialer) DialForWorker(workerPodNamespace, workerPodName string) (
...
+
+		var foundPodUID customExtensionVal
+		for _, ext := range leaf.Extensions {
+			if ext.Id.Equal(oidPodUID) {
+				if _, err := asn1.Unmarshal(ext.Value, &foundPodUID); err != nil {
+					return fmt.Errorf("failed to parse oidPodUID extension: %w", err)
+				}
+			}
+		}
+		if foundPodUID.S != expectedPodUID {
```
mtaufen: Is there any scenario where these could both end up as the empty string (e.g. due to zero-value initialization) and incorrectly pass this check? Maybe we should also confirm non-empty?

cmd/atelet/main.go
```diff
@@ -1060,3 +1076,23 @@ func resetActorDirs(atespace, actorName string) error {
...
+	caBytes, err := os.ReadFile(clientCAPath)
+	if err != nil {
+		return nil, fmt.Errorf("read CA bundle %s: %w", clientCAPath, err)
+	}
+	clientCAs := x509.NewCertPool()
+	if !clientCAs.AppendCertsFromPEM(caBytes) {
+		return nil, fmt.Errorf("parse CA bundle from %s", clientCAPath)
+	}
+	return &tls.Config{
+		MinVersion:     tls.VersionTLS12,
```
mtaufen: Why not TLS 1.3?

internal/ateapiauth/client_test.go
```diff
@@ -29,3 +51,179 @@ func TestDialOptionsRequiresCAFile(t *testing.T) {
...
+			Mode:             ModeMTLS,
+			CAFile:           caFile,
+			ClientCredBundle: clientBundle,
+		})
+		if err != nil {
+			t.Fatalf("DialOptions() error = %v", err)
+		}
+		// Unimplemented proves the TLS handshake and HTTP/2 setup succeeded
+		// and the RPC reached the server (which has no services registered).
+		if code := invokeCode(ctx, t, lis.Addr().String(), opts); code != codes.Unimplemented {
```
mtaufen: Is this something we'll have to update later? (whenever whatever is unimplemented, is actually implemented?)

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
```diff
@@ -121,16 +130,34 @@ func (h *Impl) MakeCert(ctx context.Context, pcr *certsv1beta1.PodCertificateReq
...
+		return fmt.Errorf("failed to marshal pod UID %s: %w", pod.UID, err)
+	}
 	template := &x509.Certificate{
 		BasicConstraintsValid: true,
 		NotBefore:             notBefore,
 		NotAfter:              notAfter,
 		URIs:                  []*url.URL{spiffeURI},
 		KeyUsage:              x509.KeyUsageDigitalSignature,
-		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
```
mtaufen: Do we want to give ALL the certs we issuer ServerAuth, or just atelet for now? Maybe this should be configurable depending on the workload?

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
```diff
@@ -38,6 +40,9 @@ import (
 const Name = "podidentity.podcert.ate.dev/identity"
 const CTBPrefix = "podidentity.podcert.ate.dev:identity:"
 
+// oidPodUID represents the Pod UID within a certificate extension.
+var oidPodUID = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 6, 1, 3}
```
mtaufen: How did we determine the object identifier? Is this a standard set of numbers?

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
```diff
@@ -121,16 +130,34 @@ func (h *Impl) MakeCert(ctx context.Context, pcr *certsv1beta1.PodCertificateReq
...
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		// Link the leaf to its issuing CA by key id so verifiers can disambiguate
+		// a multi-CA trust bundle (e.g. valkey trusts both the servicedns and
+		// podidentity CAs).
+		// https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.1
+		AuthorityKeyId: parent.SubjectKeyId,
+		ExtraExtensions: []pkix.Extension{
+			{
+				Id:       oidPodUID,
+				Critical: false,
```
mtaufen: If we always reject certs that don't have this extension, should we also mark it as critical so that any systems that don't recognize it also reject it, or is that going too far (e.g. would that break identity federation)?

internal/credbundle/credbundle.go
```diff
@@ -38,6 +38,17 @@ func Loader(path string) func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
 	}
 }
 
+// ClientLoader is the client-side counterpart to Loader. It returns a function
+// suitable for use as GetClientCertificate in a tls.Config, re-reading the
+// bundle on each handshake so that in-place pod-certificate rotations are
+// picked up.
+func ClientLoader(path string) func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+	// TODO: Introduce caching.
```
mtaufen, line 46: Would it be possible to fix this in this PR?

cmd/ateapi/main.go
```diff
@@ -70,6 +70,9 @@ var (
 	sessionIDCAPoolFile = pflag.String("session-id-ca-pool", "", "The file that contains the CA pool for signing session JWTs")
 	workerpoolCACerts   = pflag.String("workerpool-ca-certs", "", "The file that contains the CA for verifying workerpool client certificates.")
 
+	ateletClientCredBundle = pflag.String("atelet-client-cred-bundle", "/run/podidentity.podcert.ate.dev/credential-bundle.pem", "Credential bundle presented as the client certificate when dialing atelet.")
+	ateletCACerts          = pflag.String("atelet-ca-certs", "/run/podidentity.podcert.ate.dev/trust-bundle.pem", "CA bundle used to verify atelet serving certificates.")
```
mtaufen: I see worker pool certs above, does that also currently cover other system components like the router, controllers, etc that talk to ateapi?

internal/ateapiauth/server.go
```diff
@@ -129,26 +134,66 @@ func serverAuthenticatorFor(cfg ServerConfig) serverAuthenticator {
 type mtlsServerAuthenticator struct{}
 
 func (mtlsServerAuthenticator) authenticate(ctx context.Context) (context.Context, error) {
-	// TODO: Extract the transport-authenticated client identity and attach it
-	// to ctx once ateapi has an authorization layer.
-	return ctx, nil
+	pInfo := principal.PrincipalInfo{
+		ID: "anonymous",
```
mtaufen: Should we actually bother to support "anonymous" or should we just flat out reject anything that lacks credentials? We had a lot of issues with anonymous auth in K8s and have done our best to disable it wherever possible at this point. Personally I'd lean towards always requiring authn.

cmd/ateapi/internal/controlapi/dialer.go
zoez7: The problem is that kubelet first waits for the pod certificate volume mount https://github.com/kubernetes/kubernetes/blob/v1.34.0/pkg/kubelet/kubelet.go#L2066 before creating a pod, if we try to add IPAddress in cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go, it will be empty.

cmd/ateapi/main.go
zoez7: Yes. It does. Once we add authorization to the ate api, the workerpool pods should not be able to talk to the ateapi.

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
zoez7: Yes, this is a standard set up numbers I registerered in cl/949739997. The implementation now mirrors the future upstream pod identity signer https://github.com/kubernetes/kubernetes/pull/127251/changes#diff-985739ccf0a8431134e42720fd9a285e18e9fe49a5586062ff405d205e648e06. All changes related to the extension are in 0003f76abd2d45cbef660ead4303643f200570ac.

internal/credbundle/credbundle.go
zoez7: I'll create an open issue for this, there are a couple places in internal/credbundle/credbundle.go that we need to do this for and I didn't want to make this PR even bigger.

cmd/atelet/main.go
zoez7: Done in e39db9fa2f8ffe036a9c4913dc0e4ae3358cb983. Replaced all occurrences of VersionTLS12 with VersionTLS13

cmd/ateapi/internal/controlapi/dialer.go
zoez7: Done in e39db9fa2f8ffe036a9c4913dc0e4ae3358cb983. Added check to confirm non-empty.

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
zoez7: Done in e39db9fa2f8ffe036a9c4913dc0e4ae3358cb983. Limited to atelet only.

internal/ateapiauth/client_test.go
zoez7: No, this was a hack to test mtls connection was successful by calling a unimplemented service. I updated the test to register a healthcheck service in mock grpc instead. Updated tests in e39db9fa2f8ffe036a9c4913dc0e4ae3358cb983.

internal/ateapiauth/server.go
zoez7: Done in 620a215ddc051d74f8f81e88401d256244de7047. Now we use a chained authenticator, if a client does not present either bear token or peer cert, we fail the authentication.

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
zoez7: Reading https://datatracker.ietf.org/doc/html/rfc5280#section-4.2 I kept it as non-critical in case the cert will be used to authenticate to other system that is not the ate-apiserver.

cmd/ateapi/main.go
mtaufen: Should we name it something more generic, then, to indicate the CA covers more than just the worker pool? Maybe just --ate-trust-bundle?

zoez7: Done in d32f7a7a8a890ab9362107ecb146931a538f3ddc. Updated the trust bundle name to `pod-identity-ca-certs`

internal/credbundle/credbundle.go
zoez7: Created https://github.com/agent-substrate/substrate/issues/459

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
mtaufen: Ok, sounds good.

cmd/ateapi/internal/controlapi/dialer.go
mtaufen: What about DNS? The Actor DNS name should always be the same, right?

zoez7: oh, this part of the logic is unrelated to actor's DNS names. This is the mTLS config is for ateapi -> atelet requests, and the atelet serving cert is signed by cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go

cmd/ateapi/internal/controlapi/dialer.go
mtaufen: With PodCertificates and short lifetimes, should be be reloading this dynamically at some fraction of expiry time?

mtaufen: More broadly (maybe a question for @ahmedtd ), how is CA rotation expected to work with Pod Certificates?

cmd/ateapi/internal/controlapi/dialer.go
```diff
@@ -165,16 +158,18 @@ func verifyAteletServerCert(roots *x509.CertPool, expectedPodUID string) func(tl
...
-				}
-			}
+		identity, err := substratex509.PodIdentityFromCertificate(leaf)
+		if err != nil {
+			return fmt.Errorf("failed to parse PodIdentity extension: %w", err)
+		}
+		if identity == nil {
+			return fmt.Errorf("server certificate has no PodIdentity extension, expected pod UID %q", expectedPodUID)
+		}
+		if identity.PodUID == "" || expectedPodUID == "" {
```
mtaufen: We can validate `expectedPodUID` in `verifyAteletServerCert` before we construct the `func(tls.ConnectionState)`

cmd/ateapi/internal/controlapi/dialer.go
zoez7: For CA rotation in the future, we can copy what DCD does today and use a sidecar that reloads the CA trust bundle

cmd/ateapi/internal/controlapi/dialer.go
zoez7: Done in 11b4d880806caaf49d41f699fd0fa08c94eaef9f. Moved expectedPodUID == "" to the beginning of verifyAteletServerCert.

internal/substratex509/substratex509.go
```diff
@@ -0,0 +1,131 @@
...
+
+func PodIdentityFromCertificate(cert *x509.Certificate) (*PodIdentity, error) {
+	podIdentityCount := 0
+
+	pod := &PodIdentity{}
+	for _, ext := range cert.Extensions {
+		if ext.Id.Equal(oidPodIdentity) {
+			podIdentityCount++
+
+			if err := json.Unmarshal(ext.Value, pod); err != nil {
```
mtaufen: Should we wait to unmarshal until we've confirmed there's exactly one?

cmd/ateapi/internal/controlapi/dialer.go
mtaufen: Would we then terminate tls in the sidecar?

cmd/ateapi/internal/controlapi/dialer.go
mtaufen: hmm, maybe atelets should have DNS names?

internal/substratex509/substratex509.go
```diff
@@ -0,0 +1,131 @@
...
+func AddPodIdentityToCertificate(pod *PodIdentity, template *x509.Certificate) error {
+	if err := validatePodIdentity(pod); err != nil {
+		return fmt.Errorf("while validating PodIdentity input: %w", err)
+	}
+	podIdentityBytes, err := json.Marshal(pod)
+	if err != nil {
+		return fmt.Errorf("while json-marshaling PodIdentity extension: %w", err)
+	}
+
+	template.ExtraExtensions = append(template.ExtraExtensions, pkix.Extension{
```
mtaufen, line 68: How compatible do we expect the JSON blob in this extension to be with existing proxies and policy engines? Will anyone who cares need to add custom support to extract this?

internal/substratex509/substratex509.go
zoez7: Done in d7b934a1cd572bcc062e863d5c54fdac1b947724, moved the unmarshalling logic after checking podIdentityCount.

cmd/ateapi/internal/controlapi/dialer.go
zoez7: The sidecar can only fetch the updated CA bundle, and write to a file path. The binary itself still needs to periodically reload the CA bundle from that filepath.

internal/substratex509/substratex509.go
zoez7: If other proxies wants to support this new extension, they should extract it from the cert using the new podIdentity extension ID above, and get a bag of json bytes and then unmarshal.

cmd/ateapi/internal/controlapi/dialer.go
zoez7: Does it need to? AteAPI has custom verification logic which checks the atelet pod UID that we are connecting to, I think that is a more strict check than IP or DNS name?

internal/substratex509/substratex509.go
mtaufen: Yeah but is there a format that would already automatically result in a k/v map being readable in the context of a proxy's policy engine? If not, then I think this is fine, but wanted to check.

cmd/ateapi/internal/controlapi/dialer.go
mtaufen: That's what I'm asking, should we have the binary auto-reload, or require it to be restarted for rotation?

cmd/ateapi/internal/controlapi/dialer.go
mtaufen: In theory it's fine... but every time I see custom authn logic, there's usually some subtle mistake that results in a vuln down the line...

mtaufen: Ideally we'd be able to do the vanilla TLS verification first, then add our own custom logic on top. Or at least use an official spiffe library to verify the URI SAN part, and then layer our custom extension logic on top of that? If that also doesn't exist/doesn't work, then ok, this is the way :).

internal/substratex509/substratex509.go
mtaufen: Gemini keeps telling me we should stuff it all in the URI SAN, but I don't think we want to pollute that either... particularly because these attributes can change across suspend/resume, and we don't want that to change the actor identity.

zoez7: +1 to not putting it in URI SAN for the exact reason you said.

Also this will be replaced by the upstream k8s podidentity built-in signer (ETA 1.37), so the shape should conform to what Taahir has planned to do in 
https://github.com/kubernetes/kubernetes/pull/127251/changes#diff-985739ccf0a8431134e42720fd9a285e18e9fe49a5586062ff405d205e648e06

cmd/ateapi/internal/controlapi/dialer.go
zoez7: I think we should have the binary auto-reload to avoid restarting just for cert rotation. (The CA rotation operation today in GKE is very complex and requires multiple rounds of restarting kube-apiserver, definitely not fun to debug those)

cmd/ateapi/internal/controlapi/dialer.go
zoez7: The go-spiffe package https://github.com/spiffe/go-spiffe/blob/v2.8.1/spiffetls/tlsconfig/config.go#L187 also skips the vanilla TLS verification. I imported the go-spiffe package here to use their `x509svid.Verify` instead of custom logic.

EItanya: I think to get a DNS name here we'd have to create a custom service which did the mapping that this dialer does, or otherwise mess with DNS config in a way that seems less than ideal. I could definitely be missing something though

internal/principal/principal.go
EItanya, line 1: This will balloon over time as we add RBAC, but i think it's a good starting pointj

internal/substratex509/substratex509.go
```diff
@@ -0,0 +1,133 @@
...
+	"encoding/asn1"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+var (
+	// GoogleSubstratePEN is the ASN.1 Private Enterprise Number arc used to
+	// name X.509 extensions that communicate Substrate-specific concepts.
+	GoogleSubstratePEN = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 12}
```
EItanya, line 34: What does this have to do with google?

zoez7: The OID prefix 1.3.6.1.4.1.11129 is registered under the Google's enterprise number 11129  https://pki.goog/oids/index.html, after we donate this to CNCF, we'll replace with a new different OID under CNCF.

internal/substratex509/substratex509.go
mtaufen: ack

cmd/ateapi/internal/controlapi/dialer.go
mtaufen: Using go-spiffe sgtm

cmd/ateapi/internal/controlapi/dialer.go
```diff
@@ -97,3 +130,70 @@ func (d *AteletDialer) DialForWorker(workerPodNamespace, workerPodName string) (
...
+		// certificate has no DNS/IP SAN.
+		InsecureSkipVerify: true,
+		VerifyConnection:   verify,
+	}
+
+	return &tlsConfig, nil
+}
+
+func verifyAteletServerCert(bundle *x509bundle.Bundle, expectedID spiffeid.ID, expectedPodUID string) (func(tls.ConnectionState) error, error) {
+	if expectedPodUID == "" {
```
mtaufen, line 166: should we also be checking that expectedID is nonempty?

zoez7: Done in bbe2c7b506edef1f085eecb83280d4bb2e40481b.

mtaufen:
Thanks @zoez7 for incorporating all the feedback! I think we're good to merge and we can follow up with the dynamic cert rotation and update authn docs as follow-ups.

PR #238, opened 2026-06-12
Flatten the deep dir structure in atenet

bowei:
I just noticed that the directory structure is pretty deep. I want to make this less deep before we get too far into the project:

* cmd/atenet/internal/app << this directory level is not needed. get rid of it.
* cmd/atenet/internal/router/* move everything but the cobra flags to internal/router
* cmd/atenet/internal/* just has the cobra flags definitions

This is moving files -- no logic change.

- [x] Tests pass


BenTheElder (approved):

bowei:
Looks like this moved a bit too far: https://github.com/agent-substrate/substrate/issues/49
Sorry about the churn, I will update to the right place.

PR #239, opened 2026-06-12
WIP: introduce ateom-cloud-hypervisor

### LINKED ISSUE #123: [Feature] MicroVM support
For Alpha, we need to support running actors using a MicroVM hypervisor, preferably without regressing on gVisor support.

Open Questions:
* Should we dispatch to Kata, or directly interface with Cloud Hypervisor?  Kata apparently does not currently implement the checkpoint/restore parts of the containerd shim API.
* Should we revive NoVM, if we think we will need to be patching the hypervisor?
* MicroVM checkpoints will be larger, since the hypervisor doesn't have a granular understanding of which pages are in use.
  * One alternative is to launch MicroVMs without checkpoint/restore.  The actor always boots from scratch, and we just restore a data directory (or mount it via NFS or similar protocol).

BenTheElder:
NOTE: This is currently a rough draft for visibility. It does not quite have everything how I'd want, even for MVP. 

High Level:
- New peer binary to ateom-gvisor
- Cloud Hypervisor + Kata + virtiofsd
   -  LF Projects, richer functionality than firecracker, same underlying VMM crate from crosVM.
- Support for development in `kind` IF the host has `/dev/kvm`. 
   - Devs on Linux just need to have KVM enabled + docker, on macOS M2+ hardware you can `limactl start --nested-virt` https://lima-vm.io/
- Actor Template supports "runtime" (IE ateom / sandbox selection) and more generic asset specification (more on that below)

There are some pretty obvious TODOs (non-exhaustive):
- Documentation [Will update before merge].
- guest VM memory sizing (major perf implications ... but perhaps punted to iterate)
- gvisor config is unnecessarily disjoint, while a generic config is introduced, I'm not totally sold on it, and gvisor + demos haven't been moved yet
- gvisor and uVM assets should probably be pulled out of actor template to a shared resource admins can manage with separate RBAC? These feel awkward. The ActorTemplate needs to be coupled to a version of the sandbox, but maybe not all the finer details inline. (punted to iterate)
- It depends on a recent virtiofsd build and upstream builds lack arm64 so we have to build ourselves, that's annoying. It does cache well though. Eventually we can use kata's builds.
- update cmd/setup-gcp to optionally provision a nested virt pool (though also I think we should move this to another repo or something ... but as long as it exists ...)
- e2e tests and CI: ... this is problematic, free GHA runners don't support nested virt AFAICT. We can probably at least get the "integration" tests running ...

----

Fixes #123 

- [x] Tests pass

This PR is AI assisted. Everything will be completely vetted before converting from draft mode. (It is tested and I have taken a review pass, but sharing a bit early)


EItanya:
Overall I really like the direction here, I just have a few overarching questions/comments:
1. This PR makes most API splits between `gvisor` and `uVM` into `oneofs`. Do we want to go down that road or the more generic extension road. I'm not arguing either just want to bring it up.
2. There is a quite a bit of code to "default" to `gvisor` in the code. Also in terms of backwards compat on snapshots. Do we want to have a default or require a runtime?
3. Cloud hypervisor added [on demand restore](https://www.cloudhypervisor.org/blog/cloud-hypervisor-v52.0-released/#userfaultfd-demand-paged-snapshot-restore). We did an internal POC where we got this working, do we want to/plan to add it here as well?

I have some other more specific questions, but honestly they all come down to the answer of 1, so I will hold off for now

BenTheElder:
> This PR makes most API splits between gvisor and uVM into oneofs. Do we want to go down that road or the more generic extension road. I'm not arguing either just want to bring it up.

> There is a quite a bit of code to "default" to gvisor in the code. Also in terms of backwards compat on snapshots. Do we want to have a default or require a runtime?


Yeah ... I've been thinking about those over the past weekend, drafting a proposal today (not implemented yet), these two are my least favorite part of the MVP, followed by the hardcoded label / taint handling for KVM (should be possible to replace with something like #249).

I think we can split it out entirely.

Rough Proposal: https://docs.google.com/document/d/1U1Q7Njy-XDKwnhhU4t6KJGfWKtBL-bZvmpLIB4MA3RI/edit?tab=t.0

(shared with the ate-dev mailinglist for access, will capture whatever we wind up doing back to github in some form).


> Cloud hypervisor added [on demand restore](https://www.cloudhypervisor.org/blog/cloud-hypervisor-v52.0-released/#userfaultfd-demand-paged-snapshot-restore). We did an internal POC where we got this working, do we want to/plan to add it here as well?

So I'm expecting [tiered local => remote blob storage](https://github.com/agent-substrate/substrate/pull/227#:~:text=The%20local%20file%20management%20solution%20is%20temporary%20and%20will%20be%20replaced%20once), as a result I don't think we want to jump to complicating the snapshot fetch part too much yet, which leaves "should we trivially enable ondemand mode".

We did a decent bit of testing during https://groups.google.com/g/ate-dev/c/fkij70u31s8/m/DHraMym5BAAJ

It's sometimes slower to use chv ondemand than copy, vs sparse snapshots are pretty consistently a big win.

That only surfaces if you look for e2e response time of an actor, since it may be faster to "start" but then pay huge penalties demand-paging in necessary memory before FTTR.

The biggest problem is it's workload dependent on the wake side memory access patterns after a restore ... so it will be hard to even have a reliable heuristic without some awkward layering violations.

I think we can iterate on that part later anyhow.

BenTheElder:
I stashed some cosmetic changes here: https://github.com/agent-substrate/substrate/compare/main...BenTheElder:ateom-microvm?expand=1

Pulled out the refactoring for sandbox binaries to https://github.com/agent-substrate/substrate/issues/253 and will send a pre-req PR so we can go ahead and address that for gvisor instead of making the microVM MVP even larger ...

BenTheElder:
https://github.com/agent-substrate/substrate/pull/261 to implement https://github.com/agent-substrate/substrate/issues/253, will follow-up with a refactor on top of that + other latest changes in main (e.g. pause vs suspend) ...

BenTheElder:
I'm going to leave this for reference and fire off a clean PR with some changes atop #261, the pause/suspend split, and the workerpool scheduling changes at HEAD.

Issue #98, opened 2026-05-28
Easy Stress Testing


maxsmythe:
Substrate's goal is to enable more-efficient serving of large workloads. Ideally we should have a way to validate that the changes we're making are moving towards this goal.

I suggest we create a specialized workload that makes it easy to test system behavior at varying levels of scale:

* Arbitrarily large numbers of actors
* The ability to consume network/RAM/disk/kernel memory at customizable levels
* Varying rates of suspend/resume activity
* Combinations of the above functionality.

Ideally, we should be able to customize the workload without instantiating new actors each time. It should be customizable enough to test specific functionality under development and scale as project functionality/infrastructure capabilities scale.

I'm working on a low-overhead binary that can be configured via RPC calls to exhibit the desired behavior (disk consumption, network requests, etc.). This should allow for easy scaling while leaning on Locust to provide the behavioral guidance, enabling live-customization of load characteristics and non-homogeneous load profiles.

maxsmythe:
PR to add "glutton" image: #243

Still need to add customizable behavior and user traffic should go through atenet

EItanya:
Given that #243 has merged, i vote we close this and create follow-up issues for more work that needs to be done

maxsmythe:
SGTM

PR #243, opened 2026-06-13
Add baseline glutton

maxsmythe:
Adds the glutton image to support easier load testing contemplated in #98

This is a rough implementation to help bootstrap more repeatable testing and is expected to change over time as we learn more about performance/testing requirements.


EItanya:
> Adds the glutton image

What does this mean?

Can you please add a longer PR description so it's clearer to potential reviewers what's happening here?

maxsmythe:
Sorry, linked the issue to the PR, but not the PR to the issue. Updated description.

EItanya:
> Sorry, linked the issue to the PR, but not the PR to the issue. Updated description.

Thanks :)

cmd/testing/glutton/main.go
```diff
@@ -0,0 +1,488 @@
...
+	"io"
+	"log/slog"
+	"net"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sync"
+	"time"
+
+	"github.com/agent-substrate/substrate/internal/proto/glutton"
```
bowei: nit: put Substrate imports on their own block at the end

cmd/testing/glutton/main.go
```diff
@@ -0,0 +1,488 @@
...
+	peers := s.peers
+	s.peers = make(map[string]*peerGossip)
+	s.mu.Unlock()
+	for _, p := range peers {
+		p.cancel()
+		<-p.done
+	}
+}
+
+func (s *gluttonService) WriteRAM(ctx context.Context, req *glutton.WriteRAMRequest) (*glutton.WriteRAMResponse, error) {
```
bowei, line 254: I don't think this makes a huge difference, but this seems to do more than what I think of for writeRAM as it allocates 2x the RAM (if the argument Size is like 32 Gb or something).

Overwrite should just touch bytes but do not extra allocation?

cmd/testing/glutton/main.go
```diff
@@ -0,0 +1,488 @@
...
+		s.gossipLatency.Record(ctx, latency, attrs)
+		if err != nil {
+			if errors.Is(ctx.Err(), context.Canceled) {
+				return
+			}
+			slog.WarnContext(ctx, "Gossip ping failed", slog.String("host", pg.host), slog.Any("err", err))
+			continue
+		}
+		if resp.GetMessage() != msg {
+			slog.WarnContext(ctx, "Gossip ping returned unexpected message",
```
bowei, line 466: do we have a metric that we bump if glutton encounters errors?

bowei (commented):
Generally lgtm. You might want to document in the RPC defs what each glutton action intends to do.

cmd/testing/glutton/main.go
maxsmythe: ack, thanks. Fixed (once commit is pushed)

cmd/testing/glutton/main.go
maxsmythe: This is a good point. Writing out a larger array will still require swapping the buffer, but we can at least deal with writing to subsets of the buffer.

My main hope with this one is to give us the ability to more easily see the effect of different snapshotting systems.

cmd/testing/glutton/main.go
maxsmythe: Added a "outcome" tag to align with extproc.go

maxsmythe:
Thanks for the review! addressed comments and rebased.

maxsmythe:
Missed the documentation in the protobuf comment, will integrate into automation PR

benchmarking/locust/tests/glutton.py
```diff
@@ -0,0 +1,193 @@
...
+    # separate attribute so it's not clobbered when host points elsewhere
+    # (e.g. when running with other user classes via --class-picker).
+    host = "api.ate-system.svc.cluster.local:443"
+    api_host = "api.ate-system.svc.cluster.local:443"
+    template_name = "glutton"
+
+    def on_start(self):
+        update_user_count(1, self.__class__.__name__)
+
+        target = self.api_host.replace("http://", "").replace("https://", "")
```
bowei, line 62: this probably deserves a comment on what this is for

benchmarking/locust/tests/glutton.py
```diff
@@ -0,0 +1,193 @@
...
+        if self.glutton_channel is not None:
+            try:
+                self.glutton_channel.close()
+            except Exception as e:
+                logger.warning(f"Failed to close glutton channel: {e}")
+        self._teardown_actor()
+        self.api_channel.close()
+
+    def _teardown_actor(self):
+        try:
```
bowei, line 129: do we need to track any of these teardown failures?

benchmarking/locust/tests/glutton.py
```diff
@@ -0,0 +1,193 @@
...
+        with tracer.start_as_current_span("GluttonPing") as span:
+            headers = {}
+            inject(headers)
+            metadata = list(headers.items())
+            try:
+                resp = self.glutton_stub.Ping(
+                    glutton_pb2.PingRequest(message=msg),
+                    metadata=metadata,
+                )
+                duration = (time.time() - start_time) * 1000
```
bowei: durationMs?

benchmarking/workloads/manifests/full_workloads.yaml.tmpl
```diff
@@ -113,3 +113,35 @@ spec:
...
+
+apiVersion: ate.dev/v1alpha1
+kind: ActorTemplate
+metadata:
+  name: glutton
+  namespace: benchmark-workloads
+spec:
+  runsc:
+    amd64:
+      url: "gs://gvisor/releases/nightly/2026-05-19/x86_64/runsc"
```
bowei, line 108: might want a comment here as to where this reference comes from?

cmd/testing/glutton/main.go
```diff
@@ -0,0 +1,503 @@
...
+	defer serverboot.ShutdownProvider("TracerProvider", tp.Shutdown)
+
+	mp, err := serverboot.InitMetrics(ctx, "glutton")
+	if err != nil {
+		serverboot.Fatal(ctx, "Failed to initialize metrics", err)
+	}
+	defer serverboot.ShutdownProvider("MeterProvider", mp.Shutdown)
+
+	if err := os.MkdirAll(*dataDir, 0o700); err != nil {
+		serverboot.Fatal(ctx, "Failed to create data directory", err)
```
bowei: add the dataDir to the error

cmd/testing/glutton/main.go
```diff
@@ -0,0 +1,503 @@
...
+
+	svc, err := newGluttonService(*dataDir)
+	if err != nil {
+		serverboot.Fatal(ctx, "Failed to construct glutton service", err)
+	}
+	defer svc.Close()
+
+	lis, err := net.Listen("tcp", *listenAddr)
+	if err != nil {
+		serverboot.Fatal(ctx, "Failed to start listener", err)
```
bowei: Is the listenAddr in the error (for debugging)

cmd/testing/glutton/main.go
```diff
@@ -0,0 +1,503 @@
...
+	dataDir string
+
+	// TODO: split this into per-resource locks (ram, fds, peers). A single
+	// global mutex serializes unrelated operations across all three.
+	mu    sync.Mutex
+	ram   map[string][]byte
+	fds   []*os.File
+	peers map[string]*peerGossip
+
+	ramWriteBytes  metric.Int64Counter
```
bowei, line 143: Do you think it would be easier to maintain if we should split off each of these kind of operations into different structs with a abstract interface? then we wouldn't have to throw all of the operation code into this one struct.

something like 

```
type action interface {
  register()
  do()
  stop()
}
```

bowei: personal preference to also split this into different files to avoid this become a mega file.

the other thing you will want to do is to keep the `package main` to a minimum and put the real logic into a testable package. You can't add unit tests to the `main` package.

Move the code to `cmd/testing/glutton/internal`...

cmd/testing/glutton/main.go
bowei: would be helpful to summarize behavior in a 1-2 sentences for the WriteRAM docstring.

bowei (commented):
Do we want to add `typing` to the Python code? I know Python types are optional, but it does quickly become a maintenance issue.

benchmarking/workloads/manifests/full_workloads.yaml.tmpl
maxsmythe: This goes away upon rebase (that parameter has been removed)

benchmarking/locust/tests/glutton.py
maxsmythe: in later PRs this gets traced, also this code moves to a boomer worker (golang implementation of Locust protocol), probably not worth fixing.

cmd/testing/glutton/main.go
maxsmythe: I think some refactoring makes sense, but I don't want to do it in this specific PR because other PRs are already built on top of it and splitting files starts to get hairy WRT rebasing. Happy to tackle this in a follow-on PR.

cmd/testing/glutton/main.go
maxsmythe: added

cmd/testing/glutton/main.go
maxsmythe: added

benchmarking/locust/tests/glutton.py
maxsmythe: changed

benchmarking/locust/tests/glutton.py
maxsmythe: added

maxsmythe:
Added doc strings.

Addressed all comments except ones that will be mooted by follow-on PRs and deferring glutton refactor to avoid rebase difficulty.

bowei (approved):

PR #245, opened 2026-06-14
docs: add architecture diagrams (resource model, activation, lifecycle)

fkautz:
Adds three Mermaid diagrams to `docs/architecture.md`, each placed in the section whose prose already describes it:

- **Resource model** (class diagram) in **API Resource Models**, showing the CRDs and control-plane records (ActorTemplate, WorkerPool, Actor, Worker) with their relationships and multiplicities.
- **Activation flow** (sequence diagram) and the **lifecycle state machine** in **Actor Lifecycle**, giving a visual overview alongside the phase-by-phase walkthrough.

These diagrams previously lived in the glossary. This PR is the companion to #200, which slims the glossary back to a plain list of terms per review feedback; the diagrams move here, where people look when modifying the system, rather than being dropped.

Docs only; no code changes.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -262,6 +262,48 @@ and Performance:
       and Templates) allows platform teams to apply familiar RBAC, auditing,
       and policy enforcement to the underlying infrastructure.
 
+### Resource Model
+
+The CRDs and control-plane records described above, with their relationships and
+multiplicities (UML class diagram):
+
+```mermaid
+classDiagram
+    namespace KubernetesObjects {
+        class ActorTemplate {
+            <<CRD>>
+        }
+        class WorkerPool {
+            <<CRD>>
+        }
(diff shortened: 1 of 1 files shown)


docs/architecture.md
```diff
@@ -262,6 +262,48 @@ and Performance:
...
+            <<CRD>>
+        }
+        class Deployment
+        class WorkerPod {
+            ateom
+            runsc
+        }
+    }
+
+    namespace ControlPlaneRecords {
```
juli4n: Maybe ate-apiserver instead of ControlPlaneRecords and kube-apiserver instead of KubernetesObjects?

fkautz: Good idea, done.

juli4n (approved):

bowei:
lgtm

Issue #212, opened 2026-06-10
feat: Support pod resources, scheduling, and priority in the worker pool CRD


han2ni3bal-pixel:
**Problem**
WorkerPool only exposes spec.replicas and spec.ateomImage. The reconciler hard-codes the worker Pod template with no resources, priorityClassName, nodeSelector, affinity, or tolerations.

Worker pods therefore run as BestEffort / priority 0, are easily evicted under node pressure, and cannot be declaratively pinned to dedicated or larger nodes. This is awkward in mixed-workload or heterogeneous clusters, and it conflicts with docs/architecture.md, which describes WorkerPool as defining hardware shape.

Resume also picks a free worker at random (workflow_resume.go), with no awareness of node capacity.

**Impact**

1. Worker pods are among the first evicted during resource pressure
2. A terminated worker can leave actors stranded in RUNNING (related: #23)
3. No supported way to isolate workers or prefer certain nodes via the WorkerPool API
4. replicas means “warm slots”, not CPU capacity

**Proposal**
Add a template (or equivalent) field to WorkerPoolSpec so operators can configure at least:

- resources (requests/limits)
- priorityClassName
- nodeSelector, tolerations, affinity
- extra labels/annotations
Propagate these into the managed Deployment Pod template and document an example in docs/api-guide.md.

**Workaround today**
Manually patch the generated Deployment, or rely on cluster-level node pools/taints. Neither is reflected in the WorkerPool CR.

juli4n:
Let's dissect the proposal, because I think it touches on different areas. Some of them overlap with other issues that have been reported already.

1.  Worker pods always have QoS BestEffort.

We don't have a way to specify resource requirements in the WorkerPod spec today. This means that all worker pods have QoS BestEffort. This makes them first in row for eviction, but it also seems like a feature gap to me. I feel it's a reasonable feature request to add resource requirements to the WorkerPool spec.

2. We don't handle Pod deletion gracefully.

As you pointed out, this is already reported in #23. I would leave it out of scope from this issue.

3. No supported way to isolate workers or prefer certain nodes via the WorkerPool API

Agree. This is a gap today. We have #47 which has a proposal to address it. See the proposal in https://github.com/agent-substrate/substrate/issues/47#issuecomment-4587973653. Is that expressive enough to accommodate your use cases?

4. No way to propagate labels and annotations to worker pods.

This looks similar to point 1. Sounds like a reasonable feature request to add this to the WorkerPool spec.


[...]

Issue #47, opened 2026-05-21
Need a way for actors to select workers, and vice-versa


thockin:
Consider a service which has a free-tier and a paid-tier.  They might reasonably provision "good" machines for their paid-tier users and "cheap" machines for their free-tier users.  When a free-tier user does something which causes an actor to be instantiated, we need to offer a way to designate that this actor must use only workers on the "cheap" machines.

This could be done as a k8s-style selector (the actor has a selector, the workers have labels like "tier=free"), but that's not the only way to do it, so before we lock it down, we should at least consider other approaches.

Perhaps it is a per-user concept rather than per-actor?  Or per-namespaces (see #21)?

Is it something that the person who creates the actor is allowed to set?  k8s pods having a node selector gives them a LOT of control over themselves, so much so that people impose policies to restrict those.  If we are trying to keep this as simple as possible, we should seek to avoid things like webhook policy implementations.  :)

Maybe it's better expressed on the worker: "this worker can handle actors from namespaces where tier=free".

We need to consider how the control plane will track and query for matches.  K8s labels started simple (exactly equal) but added in, not-in, etc semantics.  Ideally we can offload some of the matching to the optimized storage layer(s)?

maxsmythe:
Are different classes of workers just entirely different worker pools?

thockin:
I assumed that different pools could have different metadata, but within a single pool they would all be identical

EItanya:
I agree this seems like exactly the use-case for different worker pools. I wonder if this is actually a limitation of actor templates mapping onto worker pools. Rather the ability to map onto multiple depending on some conditions. Does that track with what you're thinking?

maxsmythe:
That was my thought, that this was less about the worker and more about the pool.

You could have a 1:1 mapping of actor:pool (like we have now), which can support multiple tiers via multiple actors. Problems there include redundant config and no native way of handling the upgrade use-case (or a downgrade).

Tying it to the user could make sense, but that would make upgrades all-or-nothing.

Is it only a binary (free/paid), or is it something like resource requests where different users can run actors on larger or smaller workers? Something akin to autoscaling. Do we expect an actor to be permanently bound to a worker pool once created?

What if we allowed a QoS signal for each resume?

That starts to get into set logic (workerpool allowed for ActorTemplate AND user)

thockin:
I think the overly-abstract way of expressing this is that any actor can express requirements about workers it wants to use, and any worker can express requirements about which workers are allowed to use it.

[...]

PR #247, opened 2026-06-14
Add WorkerPool scheduling fields

han2ni3bal-pixel:
**Updated Original PR Description**
## Summary

This PR adds a small, controlled set of scheduling fields to WorkerPool so WorkerPool-managed Pods can be placed onto appropriate Kubernetes nodes.

Added fields:

- `nodeSelector`
- `tolerations`
- `priorityClassName`
- `nodeAffinity`

## Scope

This intentionally does not expose a full `PodTemplateSpec`, and does not add support for full `affinity`, `podAffinity`, or `podAntiAffinity`.

Resource requests/limits are intentionally left out of scope for this CL while the WorkerPool resource model is still being discussed separately.

The goal is to keep the first version small while addressing the WorkerPool-to-Kubernetes-Node scheduling gap discussed in [#212](https://github.com/agent-substrate/substrate/issues/212).

`nodeAffinity` is included to support heterogeneous node pools where `nodeSelector` is too limited, such as requiring or preferring one of several equivalent accelerator/local-SSD/cache node pools, or expressing soft preferences during node pool migration.

`priorityClassName` is included to help preserve warm WorkerPool capacity under cluster pressure. Since WorkerPool pods are the slots available for actor resume, higher-priority pools can protect latency-sensitive or interactive actor workloads from displacement by lower-priority batch workloads, and allow different WorkerPools to represent different service classes, e.g. interactive vs. batch.

## Related issue

Part of #212.
Related to #47.


pkg/api/v1alpha1/workerpool_types.go
```diff
@@ -15,9 +15,58 @@
...
+	// Requests describes the minimum amount of compute resources required.
+	//
+	// +optional
+	Requests corev1.ResourceList `json:"requests,omitempty"`
+}
+
+// WorkerPoolPodTemplate defines optional scheduling and resource settings for
+// worker pods. Resources apply to the fixed ateom container; nodeAffinity is
+// mapped to spec.affinity.nodeAffinity on the pod.
+type WorkerPoolPodTemplate struct {
```
juli4n, line 24: Can we scope this down this CL to NodeSelector and Tolerations? Do you have use cases for node affinity and priorityClass? 

I would leave resources out of scope as that is being discussed in https://docs.google.com/document/d/1_i__JEdmRexKOGP4VJ7bOWrgdS9vPd-BQSR0sxg7GQ4/edit?resourcekey=0-jtYHhxfAGZCNcoknXn2GbA&tab=t.0

han2ni3bal-pixel: @juli4n Thanks for the feedback!

That makes sense to me regarding resources. We can leave resource requests/limits out of scope for this CL while the WorkerPool resource model is still being discussed separately.

For nodeAffinity, the use case I had in mind is supporting WorkerPools across heterogeneous node pools where nodeSelector is too limited. For example, a WorkerPool may want to require or prefer one of several equivalent accelerator/local-SSD/cache node pools, or express a soft preference during a node pool migration. In those cases, nodeSelector only supports exact key/value matching, while nodeAffinity gives us required/preferred terms and set-based matching.

For priorityClassName, the use case is preserving warm WorkerPool capacity under cluster pressure. Since WorkerPool pods are the slots available for actor resume, giving certain pools a higher priority can help protect latency-sensitive or interactive actor workloads from being displaced by lower-priority batch workloads. It also allows different WorkerPools to represent different service classes, e.g. interactive vs. batch.

That said, I’m happy to scope this PR down if you still prefer to keep the initial API surface smaller. Would you prefer that I keep nodeAffinity and priorityClassName in this CL with the above use cases, or should I remove them and follow up separately after nodeSelector/tolerations land?

Thanks again for taking a look.

han2ni3bal-pixel: I’ve removed the `resources` field from this CL, btw.

cmd/atecontroller/internal/controllers/workerpool_controller.go
```diff
@@ -120,6 +120,40 @@ func (r *WorkerPoolReconciler) syncStatus(ctx context.Context, wp *atev1alpha1.W
...
+		WithSecurityContext(corev1ac.PodSecurityContext().
+			WithRunAsUser(0).
+			WithRunAsGroup(0)).
+		WithVolumes(corev1ac.Volume().
+			WithName("run-ateom").
+			WithHostPath(corev1ac.HostPathVolumeSource().
+				WithPath(ateompath.BasePath).
+				WithType(corev1.HostPathDirectoryOrCreate)))
+
+	applyWorkerPoolPodTemplate(podSpecAC, containerAC, wp.Spec.Template)
```
juli4n: While I understand that with the introduction of the template now creating the DeploymentApplyConfiguration is more complex / dynamic (it cannot be represented with a single struct literal), I find the approach of creating the static part here and then calling a method (in another file) to add the dynamic part a bit convoluted.

How about we move buildDeploymentApplyConfig to a new file (workerpool_apply.go?) and implement the  whole generation of *appsv1ac.DeploymentApplyConfiguration in that method? Then we can add some table driven tests that exercise the buildDeploymentApplyConfig method e2e that take WorkerPool as input and assert expected *appsv1ac.DeploymentApplyConfiguration as output.

pkg/api/v1alpha1/workerpool_types.go
juli4n: Thanks. Added some comments. Could you also update the PR description to reflect the current scope?

cmd/atecontroller/internal/controllers/workerpool_controller.go
han2ni3bal-pixel: - Moved `buildDeploymentApplyConfig` out of `workerpool_controller.go` into `workerpool_apply.go`.
- Kept WorkerPool template field mapping as an internal detail of the apply configuration builder.
- Replaced lower-level helper / JSON shape tests with table-driven tests for:
  - default WorkerPool
  - `nodeSelector`
  - `tolerations`
  - `nodeAffinity`
  - `priorityClassName`
  - combined scheduling fields
- Preserved SSA clearing behavior for controller-owned pod template fields.

pkg/api/v1alpha1/workerpool_types.go
han2ni3bal-pixel: @juli4n, Hi, description has been updated and the modified code has been uploaded. If you have more comments, feel free to add, thanks!

juli4n (approved):

PR #248, opened 2026-06-15
feat: add jwt auth for ateapi

### LINKED ISSUE #222: Support portable JWT auth for system components
## Summary

Support JWT-based authentication as an alternative to mTLS for Substrate system components.

The current alpha auth direction appears to rely on mTLS between `ate-system` components, which in turn depends on certificate provisioning paths such as the Kubernetes PodCertificate API / pod certificate controller. That creates a deployment blocker for clusters where PodCertificate is not available or not enabled.

## Problem

mTLS and JWT authentication are both reasonable options for in-cluster component identity, but requiring mTLS as the only supported auth path makes Substrate harder to install and evaluate broadly.

In particular:

- PodCertificate is not generally available across Kubernetes distributions today.
- Some managed clusters may not expose or support the required API/features.
- Local/dev/Kind installs may need extra bootstrap machinery just to get component auth working.
- Operators may already have a working projected service account token / OIDC JWT path available.

This means auth for core Substrate components can become coupled to a certificate delivery mechanism that is not portable enough yet.

## Proposal

Add support for JWT-based authentication and authorization for Substrate system components, either as:

1. A first-class alternative to mTLS, or
2. A portable default/bootstrap mode, with mTLS also available when certificate provisioning is available.

The JWT path could use Kubernetes projected service account tokens, with appropriate audiences and short lifetimes, and validate them at the receiving component boundary.

For example:

- `atelet` authenticates to `ate-api-server` with a projected service account JWT.
- `ate-api-server` validates issuer, audience, expiry, subject, namespace, and service account identity.
- Authorization maps trusted service accounts to allowed component roles/actions.
- mTLS remains supported as another authentication mode.

## Goals

- Do not require PodCertificate availability for a basic secure install.
- Provide a portable auth path that works on standard Kubernetes clusters.
- Support mTLS as a parallel authentication option where available.
- Keep authorization explicit and component-scoped.
- Make install docs clear about the supported auth modes and tradeoffs.

## Non-Goals

- Deprecating or removing mTLS.
- Solving actor identity credential delivery in this issue.
- Designing a full dynamic policy language for all authz decisions.

## Open Questions

- Should JWT auth be alpha's default path?
- Should mTLS and JWT be mutually exclusive modes, or can both be accepted at once during migration?
- What JWT audience should each component require?
- Should authz be hardcoded initially by service account identity, matching the current alpha direction?
- Which components need this path first: `atelet` -> `ate-api-server`, `ateom` -> `atelet`, router -> API server, others?
- How should this relate to #164, #170, and #124?

## Related Issues

- #164
- #170
- #124

EItanya:
## Description

  Closes agent-substrate/substrate#222.

  Adds a portable JWT authentication path for ateapi clients while keeping the
  existing mTLS mode as the default.

  ### What changed

  - Added `internal/ateapiauth` with:
    - `mtls` and `jwt` auth modes
    - server-side JWT gRPC interceptors
    - client-side dial options for projected ServiceAccount tokens
  - Wired JWT auth into:
    - `ate-api-server`
    - `ate-controller`
    - `atenet router`
    - `kubectl-ate` port-forward client path
  - Updated Kubernetes JWT verification to support custom HTTP clients for
  OIDC/JWKS discovery.
  - Added JWT install overlays:
    - `manifests/ate-install/jwt`
    - `manifests/ate-install/kind-jwt`
  - Updated `hack/install-ate.sh` to opt into JWT mode with:
    - `ATE_API_AUTH_MODE=jwt`
    - `--auth-mode=jwt`

  ### Notes

  - Default install behavior remains `mtls`. I think this deserves a second look as JWT will support more clusters.
  - The JWT overlay projects short-lived ServiceAccount tokens with audience
  `api.ate-system.svc` and mounts the service-DNS trust bundle for ateapi TLS
  verification.
- The current discovery mechanism for JWT mode involves reading the deployment which I really don't like, but we don't have a "config file" concept so that bit is a massive TODO.

  ### Validation

```bash
KIND_CLUSTER_NAME=substrate-jwt ./hack/create-kind-cluster.sh

KO_DOCKER_REPO=localhost:5001 \
ATE_INSTALL_KIND=true \
./hack/install-ate.sh --auth-mode=jwt --deploy-ate-system

Validation results:

kubectl config current-context
# kind-substrate-jwt

kubectl get pods -n ate-system
# all runtime pods Running; init jobs Completed

go run ./cmd/kubectl-ate --context kind-substrate-jwt get actors
# succeeded, empty actor list
```


cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -169,7 +171,7 @@ func Verify(ctx context.Context, jwt string, expectedIssuer, expectedAudience st
 	}
 
 	// TODO: Cache keys, and only fetch new keys if the JWT's key ID is not in the cache.
-	keys, err := discoverKeysForIssuer(ctx, rawClaims.Issuer)
+	keys, err := discoverKeysForIssuer(ctx, httpClient, rawClaims.Issuer)
```
mtaufen, line 178: Do we ever plan to support arbitrary issuers or follow issuer URLs in tokens? Asking since it looks like this client includes a KSA JWT, and that could be a future path to token leakage.

internal/ateapiauth/client.go
```diff
@@ -0,0 +1,115 @@
...
+	// as a Bearer credential. Required for ModeJWT.
+	TokenFile string
+}
+
+// DialOptions returns the grpc.DialOption set described by cfg, suitable to
+// pass to grpc.NewClient.
+func DialOptions(cfg ClientConfig) ([]grpc.DialOption, error) {
+	switch cfg.Mode {
+	case "", ModeMTLS:
+		tlsCfg := &tls.Config{InsecureSkipVerify: true} //nolint:gosec // explicit opt-in
```
mtaufen, line 65: it's not really mTLS if we're using InsecureSkipVerify

cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -424,10 +426,12 @@ func discoverKeysForIssuer(ctx context.Context, issuer string) ([]*KeyAndID, err
 	return ret, nil
 }
 
-func fetchJSON[T any](url string) (T, error) {
+func fetchJSON[T any](httpClient *http.Client, url string) (T, error) {
 	var parsedBody T
-
-	resp, err := http.Get(url)
+	if httpClient == nil {
+		httpClient = http.DefaultClient
```
mtaufen: IIRC the Go default client is missing timeouts and other production-grade settings. Can we harden this config? cc @ahmedtd who has some past experience improving the default client config in production.

internal/ateapiauth/client.go
EItanya: That’s what it’s doing today already which I assumed was just placeholder for the mTLS mode

cmd/ateapi/internal/k8sjwt/k8sjwt.go
EItanya: > Asking since it looks like this client includes a KSA JWT, and that could be a future path to token leakage

Can you explain what you mean by that, not 100% sure how the k8s JWTs would cause that

internal/ateapiauth/client.go
mtaufen: ack

cmd/ateapi/internal/k8sjwt/k8sjwt.go
mtaufen: If the K8s JWT is always bound to this client, and the client in the future gets reused to query arbitrary issuer URLs from untrusted tokens (not saying we should do that!), then it could be a leak risk. If we'll never do that then it's not really a concern, though I wonder if we should wrap the client type to make it obvious (like call it `K8sServiceAccountIssuerDiscoveryClient`).

zoez7: > Do we ever plan to support arbitrary issuers or follow issuer URLs in tokens?

I don't think we should support arbitrary issuers, but we can follow what [k8s AuthenticationConfig](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#:~:text=AuthenticationConfiguration) does, and allow user to allowlist OIDC endpoints

cmd/ateapi/internal/k8sjwt/k8sjwt.go
mtaufen: The more I look into all the options, the more complicated it gets :P. I think we can start by just setting http.Client.Timeout, which covers the whole request: `&http.Client{Timeout: 10 * time.Second}` should be the same as the default client, but with the timeout set.

cmd/ateapi/internal/k8sjwt/k8sjwt.go
EItanya: Good call. I tightened this up in ba351a23b507d85150d97ac781b8b34357ec5b51.

The discovery client is now explicitly named/scoped as a Kubernetes ServiceAccount issuer-discovery client, and its transport only injects the pod SA token for URLs under the configured issuer scope. Out-of-scope discovery/JWKS URLs are rejected before the base transport is called, so a future arbitrary issuer URL cannot receive the token. I also added focused tests for the scoping behavior and token injection path.

cmd/ateapi/internal/k8sjwt/k8sjwt.go
EItanya: Done in 2e847dd33aad5ca641126037a90e70e595bb6ecf. The nil-client fallback now uses `&http.Client{Timeout: 10 * time.Second}` and the `Verify` doc comment reflects that.

manifests/ate-install/kind-jwt/kustomization.yaml
zoez7, line 1: Doesn't have to be in the same PR, but to make the ate-system fully deployable without the k8s podcertificate feature, we'd also need to update the [valkey cluster configuration](https://github.com/agent-substrate/substrate/blob/bd33e076069b86dfa0b17d3c7de6baa0c1020c3a/manifests/ate-install/valkey.yaml#L130)

internal/ateapiauth/client.go
```diff
@@ -0,0 +1,115 @@
...
+		return []grpc.DialOption{
+			grpc.WithTransportCredentials(credentials.NewTLS(tlsCfg)),
+		}, nil
+
+	case ModeJWT:
+		if cfg.CAFile == "" {
```
zoez7: Nit: we can check `cfg.CAFile == ""` in both JWT and mTLS modes, and move the `--ateapi-ca-file=/run/servicedns-ca/trust-bundle.pem` from Kustomization to the base manifest.

cmd/ateapi/main.go
```diff
@@ -327,3 +359,79 @@ func buildServerCreds(ctx context.Context) (credentials.TransportCredentials, er
 		ClientCAs:      clientCAs,
 	}), nil
 }
+
+// buildK8sServiceAccountIssuerDiscoveryClient returns an *http.Client scoped to
+// Kubernetes ServiceAccount issuer discovery. It trusts caFile for TLS
+// verification and injects the pod's ServiceAccount Bearer token only for URLs
+// under issuer. Returns nil (use http.DefaultClient) if issuer or caFile is
+// empty or unreadable.
+func buildK8sServiceAccountIssuerDiscoveryClient(ctx context.Context, caFile, issuer string) *http.Client {
```
zoez7, line 379: currently this caFile is used to get the OIDC discovery doc, and is always `DefaultServiceAccountCAFile = /var/run/secrets/kubernetes.io/serviceaccount/ca.crt`, this will break people who are using GKE clusters (which is most googlers here), because the OIDC discovery doc is at `https://container.googleapis.com/...` https://github.com/agent-substrate/substrate/blob/bd33e076069b86dfa0b17d3c7de6baa0c1020c3a/hack/install-ate.sh#L185

internal/ateapiauth/server.go
```diff
@@ -0,0 +1,182 @@
...
+type ServerConfig struct {
+	Mode Mode
+
+	// VerifyBearerToken verifies a Bearer token presented by a client. Required
+	// for ModeJWT and ignored for ModeMTLS.
+	VerifyBearerToken func(context.Context, string) error
+
+	// HTTPClient is kept for source compatibility with older callers. Token
+	// verification belongs in VerifyBearerToken.
+	HTTPClient *http.Client
```
zoez7: remove unused field?

internal/ateclient/builder.go
```diff
@@ -230,6 +238,74 @@ func dialPortForward(ctx context.Context, kubeconfigPath, k8sContext string, tra
...
+	}
+
+	expirationSeconds := int64(3600)
+	tokenRequest := &authv1.TokenRequest{
+		Spec: authv1.TokenRequestSpec{
+			Audiences:         []string{"api.ate-system.svc"},
+			ExpirationSeconds: &expirationSeconds,
+		},
+	}
+	token, err := clientset.CoreV1().ServiceAccounts("ate-system").CreateToken(ctx, "ate-controller", tokenRequest, metav1.CreateOptions{})
```
zoez7: Nit: Create a new SA and token dedicated to the client instead of relying on the ate-controller one?

internal/ateapiauth/client.go
EItanya: Done in 49d3bf7. `DialOptions` now validates `CAFile` before mode-specific handling, and the manifests put `--ateapi-ca-file=/run/servicedns-ca/trust-bundle.pem` in the base ate-controller / atenet-router args. I also adjusted the JWT overlay to add auth/token args via JSON patches so the base CA arg is preserved in rendered output.

internal/ateapiauth/server.go
EItanya: Done in 49d3bf7. Removed the unused `HTTPClient` field from `ServerConfig`.

cmd/ateapi/main.go
EItanya: Fixed in 49d3bf7. External issuers such as `https://container.googleapis.com/...` now use system roots with a 10s timeout and do not use the ServiceAccount-token transport. The in-cluster `https://kubernetes.default.svc` issuer still uses the serviceaccount CA plus scoped ServiceAccount token transport.

internal/ateclient/builder.go
EItanya: Done in 49d3bf7. Added a dedicated `ate-client` ServiceAccount to the install manifests and changed the client TokenRequest path to mint tokens for that ServiceAccount instead of `ate-controller`.

manifests/ate-install/kind-jwt/kustomization.yaml
EItanya: Agreed. I did not fold the Valkey podcertificate-free changes into this PR because that needs a separate pass over the Valkey server cert/key material, the init Job `valkey-cli --cert/--key` paths, and the ateapi Redis client cert settings. This PR now avoids the GKE OIDC CA issue and separates client JWT identity, but Valkey still depends on the podcertificate path today.

EItanya:
@zoez7 addressed your comments. Also Just FYI I changed the e2e test workflow to use a matrix and install substrate in both modes to make sure they work. Let me know if you think that's overkill

internal/ateclient/builder.go
```diff
@@ -52,6 +52,8 @@ type Client struct {
 	tracerProvider *sdktrace.TracerProvider
 }
 
+const ateAPIClientServiceAccount = "ate-client"
```
zoez7: Nit: inline this since it's only used once?

cmd/ateapi/main.go
```diff
@@ -135,18 +147,37 @@ func main() {
...
 	}
 
+	authCfg := ateapiauth.ServerConfig{
+		Mode: authModeParsed,
+		VerifyBearerToken: func(ctx context.Context, bearer string) error {
+			_, err := k8sjwt.Verify(ctx, jwtIssuerDiscoveryClient, bearer, *clientJWTIssuer, *clientJWTAudience, time.Now())
+			return err
+		},
+	}
+	if err := ateapiauth.ValidateServerConfig(authCfg); err != nil {
```
zoez7, line 176: Let's also check if `jwtIssuerDiscoveryClient` is nil when authMode==JWT?

cmd/ateapi/main.go
```diff
@@ -64,7 +69,9 @@ var (
 	sessionIDCAPoolFile = pflag.String("session-id-ca-pool", "", "The file that contains the CA pool for signing session JWTs")
 	workerpoolCACerts   = pflag.String("workerpool-ca-certs", "", "The file that contains the CA for verifying workerpool client certificates.")
 
-	showVersion = pflag.Bool("version", false, "Print version and exit.")
+	showVersion     = pflag.Bool("version", false, "Print version and exit.")
+	authMode        = pflag.String("auth-mode", "mtls", "Auth mode for incoming gRPC: mtls|jwt. 'mtls' (default) relies on transport-level mTLS for client identity. 'jwt' additionally requires a Kubernetes ServiceAccount Bearer token on every RPC.")
```
zoez7: Can we add  "... substrate will drop support for JWT auth mode once the Pod Certificates feature is enabled by default in the minimum supported Kubernetes version."

zoez7 (dismissed):
Thanks, let's rebase and merge.

internal/ateclient/builder.go
EItanya: Done in b79b5a5. Removed the one-use constant and inlined `"ate-client"` at the TokenRequest call site.

cmd/ateapi/main.go
EItanya: Done in b79b5a5. `ateapi` now fails fast in JWT mode if the Kubernetes ServiceAccount issuer discovery client could not be built, instead of passing nil into JWT verification.

cmd/ateapi/main.go
EItanya: Done in b79b5a5. Added the Pod Certificates deprecation note to the `--auth-mode` flag help text.

zoez7:
Hmm the e2e-test check is stuck even though the test cases passed, because the it's matching the exact name "e2e-test"

image

Let's add a aggregate gate job with the name `e2e-test` that waits for the test matrix?

EItanya:
Done in 8422f51. I renamed the matrix job id to `e2e-test-matrix` and added an aggregate `e2e-test` job that always runs after the matrix and fails unless the matrix result is `success`. That should restore the exact required check name while keeping `e2e-test (mtls)` and `e2e-test (jwt)` visible.

EItanya:
@zoez7 a change went in to the pr workflow that caused a conflict, should be good to go now.

cmd/ateapi/internal/credbundle/credbundle.go
```diff
@@ -59,8 +61,9 @@ func Parse(bundlePath string) (*tls.Certificate, error) {
 		switch block.Type {
 		case "CERTIFICATE":
 			chainBytes = append(chainBytes, block.Bytes)
-		case "PRIVATE KEY":
+		case "PRIVATE KEY", "RSA PRIVATE KEY", "EC PRIVATE KEY":
```
ahmedtd: Why are these needed?  Pod Certificates will always write a PKCS#8 private key in a PRIVATE KEY block.

.github/workflows/pr-workflow.yaml
```diff
@@ -38,8 +38,13 @@ jobs:
   # One kind cluster exercises BOTH runtimes. Free x86-64 ubuntu-latest runners
   # expose /dev/kvm (with a udev rule), so create-kind-cluster.sh mounts it and the
   # micro-VM (kata + cloud-hypervisor) sandbox class works alongside gVisor.
-  e2e-test:
+  e2e-test-matrix:
```
ahmedtd, line 41: @BenTheElder I think we were getting throttled on runners --- will a test matrix double the problem?

cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -424,10 +427,12 @@ func discoverKeysForIssuer(ctx context.Context, issuer string) ([]*KeyAndID, err
 	return ret, nil
 }
 
-func fetchJSON[T any](url string) (T, error) {
+func fetchJSON[T any](httpClient *http.Client, url string) (T, error) {
 	var parsedBody T
-
-	resp, err := http.Get(url)
+	if httpClient == nil {
+		httpClient = &http.Client{Timeout: 10 * time.Second}
```
ahmedtd: We should not allocate a new client on every verification --- use http.DefaultClient

internal/ateapiauth/server.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (mtlsServerAuthenticator) authenticate(ctx context.Context) (context.Context, error) {
+	return ctx, nil
+}
+
+type jwtServerAuthenticator struct {
+	verifyBearerToken func(context.Context, string) error
+}
+
+func (a jwtServerAuthenticator) authenticate(ctx context.Context) (context.Context, error) {
```
ahmedtd, line 141: Can you leave some TODOs on the authenticators --- we still need to actually place the parsed identity into the inbound context.  But I think we want to do that once we add authorization.

ahmedtd (dismissed):

cmd/ateapi/internal/credbundle/credbundle.go
EItanya: Done in 1c663aa. I removed support for `RSA PRIVATE KEY` / `EC PRIVATE KEY` blocks and now only accept the Pod Certificates PKCS#8 `PRIVATE KEY` block.

.github/workflows/pr-workflow.yaml
EItanya: Agreed that this doubles the expensive path. For this PR I kept the matrix because it was the shortest path to make sure the new auth mode did not break either install/test path while we were actively changing the ateapi connectivity mode. If runner pressure becomes a problem, I think the next optimization is to keep the full gVisor+micro-VM coverage in `mtls` and make the `jwt` leg a narrower auth/connectivity smoke test.

cmd/ateapi/internal/k8sjwt/k8sjwt.go
EItanya: Done in 1c663aa. The nil-client path now uses a package-level shared HTTP client instead of allocating on every verification. I kept the 10s timeout from the earlier review rather than using raw `http.DefaultClient` with no timeout.

internal/ateapiauth/server.go
EItanya: Done in 1c663aa. Added TODOs on both authenticators for attaching the authenticated identity to the inbound context once authorization consumes it.

zoez7 (dismissed):

EItanya:
Hey there @zoez7, the `atespaces` PR ended up causing my jobs to fail. fixed now

zoez7 (dismissed):

zoez7:
Looks like we need another rebase.

EItanya:
> Looks like we need another rebase.

Done. Can we please try and merge this ASAP. It has been approved for a bit now but I've had to resolve merge conflicts a few times already.

Issue #255, opened 2026-06-16
use runsc shim at ateom-gvisor


milantracy:
ateom-gvisor at cmd/ateom-gvisor currently starts sandbox using runsc (an OCI compatible runtime) directly, I would like to move substrate to use runsc shim.

we can potentially get benefits on lifecycle management by doing the migration,  The shim daemon manages runsc child processes (sandbox, gofer) and reaps them automatically. Right now ateom does this itself with go-reap and reapLock coordination around every exec.Command.

And we can introduce a runtime-agnostic interface that we can introduce to manage other runtime as well, for example, https://github.com/agent-substrate/substrate/issues/123

milantracy:
There will be 2 phases for this migration:

phase 1: Upstream gVisor changes
* we will define service proto that expose checkpoint and restore ttrpc service via runsc shim
* the new ttrpc service will be registered at runsc shim along with existing task.RegisterTaskService (for containerd)
    - containerd will call ttrpc at `containerd.task.v2.Task`
    - ateom-gvisor will call ttrpc at its own name, e.g. `gvisor.shim.v1.TaskServiceExt`

phase 2: substrate changes
* extra a runtime agnostic interface that manages sandbox from ateom-gvisor
* implement the interface by update the existing methods in ateom-gvisor
* replacing runsc implementation with runsc-shim implementation by invoking ttrpc requests

BenTheElder:
I think it's not clear that migrating to this is better, and it will increase the resource overhead.

PR #261, opened 2026-06-17
Decouple Sandbox Binaries / Config from ActorTemplate

### LINKED ISSUE #253: Decouple Sandbox Binaries From Actor Template
Broken off from:
- https://github.com/agent-substrate/substrate/pull/239
- https://github.com/agent-substrate/substrate/issues/121
- #123 

---

# Summary

## Context

**Goal**: Enable multiple ateom implementations (specifically gVisor + microVM) for substrate worker pools by decoupling sandbox binary configuration from ActorTemplate.

**The problem today:** ActorTemplate must specify the runsc (gVisor) binary for every architecture (amd64 + arm64), including paths and hashes. This is required because gVisor snapshots aren't compatible across versions. It has two downsides:

**Usability** — every ActorTemplate author must know about and pin a gVisor version.
**Security** — anyone who can create an ActorTemplate can point to a potentially compromised sandbox binary.
Proposed changes:

## Split runsc config out of ActorTemplate.

Add `sandboxClass` to ActorTemplate to select a sandbox implementation.
Support per-ateom config formats.

**Store binary version with the snapshot:** Since the sandbox binary version is really only needed at GoldenSnapshot creation, move binary configuration to a CRD/ConfigMap on the worker pool. When a golden snapshot is created, the worker records everything needed to use it (versions/paths/config) in the snapshot itself; on restore, that config is fetched.

### Benefits:

Reduces trust needed for ActorTemplate creation (binary selection moves to a separate role/config).
Casual users no longer need to know about sandbox binaries.
Per-implementation config without polluting the shared ActorTemplate API.

**Downside**: A small config artifact must be fetched from the snapshot before the sandbox binaries — but binaries are usually cached in atelet and rarely change, and can be prefetched on startup. Analogous to how container images have a top-level manifest. Could later enable broader snapshot versioning.

## SandboxClass

SandboxClass (deferred): Today the default sandbox (gVisor) is non-configurable. Proposes a configurable default (like storageClass), possibly per-subspace rather than per-cluster (ties into the "subspaces" proposal). Deferrable — keep gVisor as default for now.

---

Full doc: https://docs.google.com/document/d/1U1Q7Njy-XDKwnhhU4t6KJGfWKtBL-bZvmpLIB4MA3RI/edit?tab=t.0
(shared with the ate-dev mailinglist documented in the README, join the list for access)

BenTheElder:
Fixes https://github.com/agent-substrate/substrate/issues/253

- [x] Tests pass

Should make something like https://github.com/agent-substrate/substrate/pull/239 less awkward, and also https://github.com/agent-substrate/substrate/issues/255 (we don't have to cram more fields into actortemplate for the shim binary and update _all_ of the demo actortemplates to include them).


pkg/api/v1alpha1/sandboxconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+// SandboxClass selects the sandbox runtime family. It is shared by WorkerPool
+// (which family a pool runs) and SandboxConfig (which family a config is for).
+type SandboxClass string
+
+const (
+	// SandboxClassGvisor is the gVisor/runsc runtime (cmd/ateom-gvisor). Default.
+	SandboxClassGvisor SandboxClass = "gvisor"
+	// SandboxClassMicroVM is the micro-VM runtime (cmd/ateom-microvm); needs
+	// /dev/kvm and vhost devices.
+	SandboxClassMicroVM SandboxClass = "microvm"
```
ahmedtd, line 30: We probably shouldn't squat on the "microvm" name for one particular microvm implementation?

pkg/api/v1alpha1/sandboxconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// +optional
+	Assets map[string]map[string]AssetFile `json:"assets,omitempty"`
+}
+
+// SandboxConfig is cluster-scoped configuration describing the sandbox binaries
+// for a sandbox runtime family. It is referenced (or defaulted) by WorkerPools
+// and decouples sandbox binary selection from ActorTemplate.
+//
+// +genclient
+// +genclient:nonNamespaced
```
ahmedtd, line 86: Why non-namespaced?  Wouldn't namespaced, and in the same namespace as the workerpool make sense?

A cluster scoped resource might be restrictive if you're not the admin of the Kubernetes cluster.

ahmedtd (dismissed):

pkg/api/v1alpha1/sandboxconfig_types.go
BenTheElder: I've been on the fence about this. 

We could call it "cloud-hypervisor" (ignores kata, verbose) "kata" (ignores specific vm implementation).

I think it's OK to squat on this, most users that want a microvm probably don't need a more specific name, it gives us some freedom to iterate on the implementation details.

gvisor on the other hand seems to be in a class of it's own with a different model and no comparable alternatives.

If there's a really compelling additional microvm-like backend ... we'll come up with a name for it ("firecracker" ?), so far the other folks looking at this have settled on chv as well I think.

pkg/api/v1alpha1/sandboxconfig_types.go
BenTheElder: > Why non-namespaced? Wouldn't namespaced, and in the same namespace as the workerpool make sense?

I was patterning it after StorageClass which is cluster scoped and has one tagged default. 

So you could avoid configuring this repeatedly if you just want all worker pools of the same sandboxClass to use the same binaries.

It's not _quite_ the same since you have one default SandboxConfig *per sandboxClass*.

> A cluster scoped resource might be restrictive if you're not the admin of the Kubernetes cluster.

Yeah, at first blush I thought that _might_ be a feature, it's pretty privileged to configure this, since we'll run the referenced binaries with high permissions.

I imagine similar to StorageClass a platform / vendor might want to configure these across all namespaces?

I don't feel super strongly about this part though.

I was mostly thinking about how to factor this out towards something semi-extensible that the casual user could ignore. I'm genuinely not sure if nonNamespaced makes sense.

BenTheElder:
[trivial rebase to pickup CI fix]

pkg/api/v1alpha1/sandboxconfig_types.go
BenTheElder: Slightly orthogonal, but I've also been mulling putting defaults into the ateom binaries, so if you really don't care you don't have to configure this at all.

BenTheElder:
> @BenTheElder [Benjamin Elder (BenTheElder)](https://github.com/BenTheElder) dismissed [Taahir Ahmed (ahmedtd)](https://github.com/ahmedtd)’s [stale review](https://github.com/agent-substrate/substrate/pull/261#pullrequestreview-4511416554) via [10f24ce](https://github.com/agent-substrate/substrate/commit/10f24ce4ea2bdb514c90029b37068983a94ebfe8)

This is a weird github-ism. AFAICT what I did was ... push with an additional commit? But it's phrased as if I went out of my way to "dismiss" the review ...

pkg/api/v1alpha1/sandboxconfig_types.go
ahmedtd: > So you could avoid configuring this repeatedly if you just want all worker pools of the same sandboxClass to use the same binaries.

I also don't feel strongly.  However, I think that actually having it namespaced does not preclude referencing it from a different namespace.  So a cluster administrator could provide some default ones in a special namespace, and give everyone permission to refer to them (via some K8s RBAC, I guess)

EItanya: Not that it’s the same, but the GatewayClass is also cluster scoped. I agree that namespaced can be shared, but given that other similar resources are cluster scoped, it seems easier to start there and then fan out later. What do yall think?

cmd/ateapi/internal/controlapi/functional_test.go
```diff
@@ -413,6 +421,62 @@ func createTemplateWithContainers(t *testing.T, tc *testContext, ns string, cont
...
+	t.Helper()
+	const name = "default-gvisor"
+	sc := &atev1alpha1.SandboxConfig{
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+		Spec: atev1alpha1.SandboxConfigSpec{
+			SandboxClass: atev1alpha1.SandboxClassGvisor,
+			Default:      true,
+			Assets: map[string]map[string]atev1alpha1.AssetFile{
+				"amd64": {"runsc": {
+					URL:    "gs://gvisor/releases/nightly/2026-05-19/x86_64/runsc",
```
dims, line 436: when do these get pruned? (from the gcs bucket)

cmd/ateapi/internal/controlapi/sandbox_assets.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, fmt.Errorf("while getting WorkerPool %s/%s: %w", ref.Namespace, ref.Name, err)
+	}
+
+	class := wp.Spec.SandboxClass
+	if class == "" {
+		class = atev1alpha1.SandboxClassGvisor
+	}
+
+	var sc *atev1alpha1.SandboxConfig
+	if name := wp.Spec.SandboxConfigName; name != "" {
```
dims, line 48: what happens if they specify `"default-gvisor"`?

cmd/atelet/main.go
```diff
@@ -466,6 +420,16 @@ func (s *AteomHerder) uploadExternalCheckpoint(ctx context.Context, req *ateletp
...
 		return err
 	}
+
+	// Pin the sandbox binaries into a manifest beside the images so a Restore on
+	// any node is self-describing.
+	manifest, err := json.Marshal(rec)
+	if err != nil {
+		return fmt.Errorf("while marshaling snapshot manifest: %w", err)
+	}
+	if err := ategcs.SendBytesToGCS(ctx, s.gcsClient, prefix+"/"+sandboxManifestName, manifest); err != nil {
```
dims, line 430: do we read/use this back at some point?

pkg/api/v1alpha1/sandboxconfig_types.go
dims: +1 to call it `"cloud-hypervisor"` or `"chv"`

dims:
@BenTheElder there is consensus that anyone can create their own microvm implementation without that code living in our repo? (just making sure we are all in violent agreement).

cmd/ateapi/internal/controlapi/functional_test.go
BenTheElder: No idea, but this value isn't new to the PR. This is what we ship in clusters. We can update both when a release has the fixes we need.

BenTheElder:
> [Benjamin Elder (@BenTheElder)](https://github.com/BenTheElder) there is consensus that anyone can create their own microvm implementation without that code living in our repo? (just making sure we are all in violent agreement).

It sounds like a nice future but landing one that works first is the top priority for POC 2 ("alpha", end of June). This is going way out of the way to make the layering more prepared for that sort of thing over just landing it but it doesn't solve all of the blockers.

EDIT: IOW: yes ... eventually, but it's not clear yet if that's an ateom, an atelet or what. It's not totally clear if the ateom/atelet split will even survive, or if we should merge these two ateoms later (and let the same worker fungibly run different sandbox types, I intentionally didn't do that for now so we can figure out different ateoms / worker pools). This PR gets us closer to extensible.

pkg/api/v1alpha1/sandboxconfig_types.go
BenTheElder: I'd prefer to stick with microvm for now. If this is fully extensible later then a different implementation can use the same sandbox class name and be more "portable" for users. 

"chv" is too opaque to an end-user and kata + chv are currently implementation details that don't even exist yet.

cmd/atelet/main.go
BenTheElder: Yes, line 456 we fetch the file, then later we read from it to get the binaries we need from that snapshot.

cmd/ateapi/internal/controlapi/sandbox_assets.go
BenTheElder: It's like StorageClass, we will only consult the default if you don't specify. If you specify it should match.

We put "default" in the name of our default, but you don't have to.

That said, we used inconsistent names for the test vs the one we ship OOTB, so I'm aligning them.

pkg/api/v1alpha1/sandboxconfig_types.go
BenTheElder: Later we might want to move to an open string, but I don't think we are there yet and I didn't want to stall uVM further. When we have 2 decent ateoms we will have a better idea what extensible looks like. Modal for example offers "sandbox" (gvisor) and "vm sandbox" (chv?).

cmd/ateapi/internal/controlapi/sandbox_assets.go
BenTheElder: Done.

pkg/api/v1alpha1/sandboxconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// SHA256 is the lower-case hex SHA256 of the asset. It both names the cached
+	// file (preventing collisions) and verifies the download's integrity.
+	//
+	// +required
+	// +kubebuilder:validation:Pattern=`^[a-f0-9]{64}$`
+	SHA256 string `json:"sha256"`
+}
+
+// SandboxConfigSpec is the desired state of a SandboxConfig.
```
EItanya, line 52: When I originally read the issue I think I was envisioning something more akin to `SandboxClass`. Wherein the Sandbox class would define the assets, and there could be  any number of them with arbitrary names. This one seems to instead have hardcoded class names. If we're going to use hardcoded class names I have a couple of follow-up questions:
1. why not make the config blocks for `gvisor` vs `ch` explicit rather than a generic map?
2. Why would a user need multiple `SandboxConfig`?

pkg/api/v1alpha1/sandboxconfig_types.go
BenTheElder: I think we should merge and iterate on this point. It's easy enough to refactor and we have a lot to add on top.

pkg/api/v1alpha1/sandboxconfig_types.go
BenTheElder: > [...] This one seems to instead have hardcoded class names. [...]

To start, because it's not really possible to bring a custom one _yet_. But you can see how that would update pretty easily (here, the other parts are trickier)

> why not make the config blocks for gvisor vs ch explicit rather than a generic map?

It's not really necessary so far. They both just need a set of assets which should be recorded alongside the snapshot. A per-class VAP (see last commit) can ensure you have the right list of assets while the generic config shape can make sure you have set url+sha etc.

I think other properties of the sandbox, like say, memory capacity, volume mounts, will be on the actortemplate side.

> Why would a user need multiple SandboxConfig?

You can set different SandboxConfig on different workerpools. For example to A/B a new gvisor release.

EDIT: Which to an ActorTemplate is still the same "sandbox class", it's more akin to a different release of that class. Everything in SandboxConfig is a versioned asset for a sandboxClass. It might need a different name ...

ahmedtd (approved):

PR #262, opened 2026-06-17
ateom/atelet: Remove actor template name/namespace from directories

ahmedtd:
Earlier in development, the actor ID was namespaced under its corresponding ActorTemplate.  This is no longer the case --- Actor IDs are now unique within the substrate database, and we want to support migration of actors between ActorTemplates (for upgrade scenarios).

This means that some of the the things atelet/ateom do no longer make sense:
* The actor state directory no longer needs to be qualified with ActorTemplate information.
* The actor logs should not be qualified with the ActorTemplate information
* The ActorTemplate information no longer needs to be passed through the atelet / ateom layers.

### DIFF
--- a/internal/proto/ateompb/ateom.proto
+++ b/internal/proto/ateompb/ateom.proto
@@ -48,9 +48,7 @@ service Ateom {
 }
 
 message RunWorkloadRequest {
-  string actor_template_namespace = 1;
-  string actor_template_name      = 2;
-  string actor_id                 = 3;
+  string actor_id = 3;
 
   string runsc_path = 4;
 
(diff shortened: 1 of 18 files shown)


BenTheElder (commented):
The tests are flaky, if you rebase on main it will pickup the test fix.

We're going to have to lock in these paths at some point (to deal with atelet / ateom skew), but I don't think we're anywhere near that yet.

From AI review (discarding irrelevant bits):

`tools/setup-gcp/dashboards/ate-snapshot-dashboard.json:17` and `:86` do sum by (... `actor_template_name` ...) — these queries will collapse to a single empty-label series after deploy (broken per-template breakdown).

`docs/observability.md:99` still documents `actor_template_name` as a label of `atelet.snapshot.size`.

LGTM overall.

ahmedtd:
@juli4n I have rebased, and slightly cut down the PR

> docs/observability.md:99 still documents actor_template_name as a label of atelet.snapshot.size.

This is now true --- atelet will continue writing the template name on this metric.

internal/proto/ateompb/ateom.proto
```diff
@@ -48,9 +48,7 @@ service Ateom {
 }
 
 message RunWorkloadRequest {
-  string actor_template_namespace = 1;
-  string actor_template_name      = 2;
-  string actor_id                 = 3;
+  string actor_id = 3;
```
juli4n, line 51: Now that we added atespaces this should include both atespace and actor id. 

FYI @HavenXia and @thockin as you are also touching some of this code.

PR #267, opened 2026-06-17
Drop replicas for kind setup for valkey

### LINKED ISSUE #225: Valkey statefulset needs to be restarted daily
## Actual Behavior
Seems like valkey is not reloading the tls-cert-file after process start up?

1.
1.
1.

## Specifications

- Version:
- Platform:

npolshakova:
Fixes https://github.com/agent-substrate/substrate/issues/225
(Or at least reduce the cost to spin it up in kind)

- [x] Tests pass

### DIFF
--- a/manifests/ate-install/valkey.yaml
+++ b/manifests/ate-install/valkey.yaml
@@ -75,7 +75,7 @@ metadata:
   namespace: ate-system
 spec:
   serviceName: valkey-cluster-service
-  replicas: 6
+  replicas: 3
   podManagementPolicy: Parallel
   selector:
     matchLabels:
@@ -159,7 +159,7 @@ spec:
         - |
           set -e
           echo "Waiting for all Valkey pods to resolve..."
-          for i in 0 1 2 3 4 5; do
+          for i in 0 1 2; do
             until getent hosts valkey-cluster-${i}.valkey-cluster-service.ate-system.svc >/dev/null 2>&1; do
               echo "Waiting for valkey-cluster-${i} DNS..."
               sleep 2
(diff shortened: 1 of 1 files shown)


npolshakova:
@BenTheElder you mentioned this might affect test coverage, so if 3 replicas isn't enough I can maybe just make it configurable in the script?

BenTheElder:
Yeah -=- I think we were concerned that shards would change behavior (need to be careful not to attempt to query across shards) but probably not replication. We may want to only disable replication in the local testing, and I'm not sure about shard count or more generally what is desirable. 

I do want to avoid a lot of skew between local and "real" deployments where possible but I'm somewhat skeptical that we need redundant copies in memory for local dev ..

cc @MushuEE @juli4n

BenTheElder:
Needs a rebase after the other PR, thanks! Raised in the community meeting today for input.

MushuEE:
There will still be crosslot concerns without replicas. But is this making 3 replicas with a single shard or 3 shards with no replicas? We are handling cross-shard queries in things like ListActors but those calls have no consistency promises. I think we could allow for "local" vs "real" deployments as the abstraction layers that need scale and "real" testing should enable that. 

Can we build an abstraction over the .yaml that allows users to set the replication with a standard default?

BenTheElder:
> Can we build an abstraction over the .yaml that allows users to set the replication with a standard default?

We apply kustomizations to install on kind already, so we can definitely do that. I think we should.

But we should also be careful to make sure it doesn't impact the quality of testing. My rough read is that valkey replication won't impact that much but if we did say, only one cluster shard, that might be too unrealistic.

PR #270, opened 2026-06-17
Decouple ActorTemplates from WorkerPools.

juli4n:
This change removes the hard-link from templates to worker pools, allowing pools to be selected using k8s labels.

Both actors and actor templates are augmented with a label selector field.

When an actor is resumed, we resolve the set of *eligible* worker pools by evaluating the intersection of both actor and template selectors. Then we find a free worker from one of them. The overall design is described in [1].

[1]: https://github.com/agent-substrate/substrate/issues/47#issuecomment-4712334620


cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -248,7 +304,7 @@ func (s *CallAteletRestoreStep) Execute(ctx context.Context, input *ResumeInput,
 		// Booting from scratch: resolve the sandbox binaries from the pool's
 		// SandboxConfig and send them so atelet can fetch and record them.
 		// (Restores above are self-describing via the snapshot manifest.)
-		sandboxAssets, err := resolveSandboxAssets(s.workerPoolLister, s.sandboxConfigLister, state.ActorTemplate)
+		sandboxAssets, err := resolveSandboxAssets(s.workerPoolLister, s.sandboxConfigLister, state.Actor.GetAteomPodNamespace(), state.Actor.GetWorkerPoolName())
```
BenTheElder, line 307: At some point we may have to consider this namespace coupling.

manifests/ate-install/generated/ate.dev_actortemplates.yaml
```diff
@@ -177,54 +177,60 @@ spec:
                 required:
                 - location
                 type: object
-              workerPoolRef:
-                description: |
-                  Name of the worker pool to use for the actor.
+              workerSelector:
```
BenTheElder, line 180: should we validate max items etc?

BenTheElder (commented):
I want to revisit if we should select just on labels or if we should also filter by say, sandboxClass.

With the direct workerpool ref that felt redundant, but now it might make sense to have something less abstract than arbitrary labels to say "this actortemplate requires `sandboxClass: gvisor`". I originally had this in that PR, but again, redundant to the workerpool ref previously. 

I suspect particularly the nil selector case will be a problem otherwise (we will try to resume across pools that are different sandboxes?)

I don't think that's a blocker though (especially as we haven't landed another sandboxClass yet), and I think it's important to stop directly coupling actortemplate <> workerpool ...

BenTheElder (approved):

juli4n:
> I want to revisit if we should select just on labels or if we should also filter by say, sandboxClass.
> 
> With the direct workerpool ref that felt redundant, but now it might make sense to have something less abstract than arbitrary labels to say "this actortemplate requires `sandboxClass: gvisor`". I originally had this in that PR, but again, redundant to the workerpool ref previously.
> 
> I suspect particularly the nil selector case will be a problem otherwise (we will try to resume across pools that are different sandboxes?)
> 
> I don't think that's a blocker though (especially as we haven't landed another sandboxClass yet), and I think it's important to stop directly coupling actortemplate <> workerpool ...

I think we will probably end up adding first class support for sandbox class (as a field in the actor template). While you can use labels for this, it sounds a bit of a stretch. Something similar is proposed for "size classes" (see the https://docs.google.com/document/d/1_i__JEdmRexKOGP4VJ7bOWrgdS9vPd-BQSR0sxg7GQ4/edit?resourcekey=0-jtYHhxfAGZCNcoknXn2GbA&tab=t.6xfy48ndh26m#heading=h.u4tld13tpire). You can approach this with labels, but it's not a great UX.

manifests/ate-install/generated/ate.dev_actortemplates.yaml
juli4n: Yes. I would propose I do that in a follow up PR though (I would like to do another pass on validation in general). We are having a lot of changes on these files and would prefer having not to rebase again (the last one was painful :) WDYT?

BenTheElder:
>  think we will probably end up adding first class support for sandbox class (as a field in the actor template). While you can use labels for this, it sounds a bit of a stretch. Something similar is proposed for "size classes" (see the https://docs.google.com/document/d/1_i__JEdmRexKOGP4VJ7bOWrgdS9vPd-BQSR0sxg7GQ4/edit?resourcekey=0-jtYHhxfAGZCNcoknXn2GbA&tab=t.6xfy48ndh26m#heading=h.u4tld13tpire). You can approach this with labels, but it's not a great UX.

💯 , exactly my thoughts. Let's merge and iterate :-)

PR #278, opened 2026-06-18
fix(atenet): retry transient upstream resets when routing to resumed actors

### LINKED ISSUE #218: Intermittent `503 connection termination` errors when resuming suspended actors
## Expected Behavior
Expected to get actual response from the server running in the actor.

## Actual Behavior
```
HTTP 503 | upstream connect error or disconnect/reset before headers. reset reason: connection termination
```

1. Create actortemplates and actors. I was using 
```
./hack/install-ate.sh --deploy-demo-multi-template
kubectl ate create actor f1 --template ate-demo-multi-template-fspersist/fspersist
kubectl port-forward -n ate-system svc/atenet-router 8000:80   # leave running
```
1. Suspend and then resume actors right away
```
for round in $(seq 1 20); do
  kubectl ate suspend actor f1
  curl -s -H "Host: f1.actors.resources.substrate.ate.dev" http://localhost:8000
done
```
1. Observe intermittent error messages `upstream connect error or disconnect/reset before headers. reset reason: connection termination`

mccormickt:
Adds a retry policy to the actor route so transient failures are retried until an actor is ready. This manifests commonly as 503 errors during the brief period between the request hitting envoy and the actor actually resuming and being able to serve requests. This change sets a circuit breaker to envoy's default of 20% of load, and a concurrency check of 20 as an arbitrary floor above the default of 3. Open to suggestions on these values from experience, or we can simply observe and iterate from here.

Reproduced with Claude on a kind cluster with the multi-template demo to inform the changes.

Fixes #218.

- [X] Tests pass

### DIFF
--- a/cmd/atenet/internal/router/xds.go
+++ b/cmd/atenet/internal/router/xds.go
@@ -50,6 +50,7 @@ import (
 	endpointgrpc "github.com/envoyproxy/go-control-plane/envoy/service/endpoint/v3"
 	listenergrpc "github.com/envoyproxy/go-control-plane/envoy/service/listener/v3"
 	routegrpc "github.com/envoyproxy/go-control-plane/envoy/service/route/v3"
+	typev3 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
 	cachev3 "github.com/envoyproxy/go-control-plane/pkg/cache/v3"
 	resourcev3 "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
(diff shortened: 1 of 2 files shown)


bowei:
I am wondering if we should be doing the retries here or add a ready notion to an Actor so we can tell if it can signal to the system that it's done resuming. One concern I have is that there are not going to be good ways to set the retry policy across a diverse set of Actors.

mccormickt:
> I am wondering if we should be doing the retries here or add a ready notion to an Actor so we can tell if it can signal to the system that it's done resuming. One concern I have is that there are not going to be good ways to set the retry policy across a diverse set of Actors.

Fair point, this was really just a coarse default data-plane policy. So perhaps a more complete solution is readiness probes for ActorTemplates.

mccormickt:
#330 implemented readiness probes and I don't think we need the retries in envoy.

Issue #21, opened 2026-05-20
Design: Should Agent Substrate have its own concept of “namespaces”?


thockin:
xref #18 

## Problem statement

Agent Substrate has the concept of an "Actor" (aka "Session", aka "Agent"). We expect to have LOTS of Actors.  We know that Actors can have relationships with each other, e.g. parents and children forming a sort of tree. We know that a given human, team, or system might own multiple trees of Actors, and probably we will want access control or policies on those groups. 

If an entire substrate instance is running in a small set of Kubernetes namespaces, how do we delineate things like actor ownership, access control, policies, etc.? 

## By analogy to k8s

The earliest versions of Kubernetes didn't have namespaces!  All of the resources existed in one big flat space, and the only way to sort through them was via labels. This was clunky for a lot of reasons, including the fact that it was basically unacceptable for all users to be able to see metadata about all other users (and WATCH makes it difficult to do simple filtering). So we added namespaces.

Once we had namespaces, we found many other ways to use them, and overall the UX of namespaces is pretty good.  

In Agent Substrate today, things like actor names exist in a flat, global space (and we don't even have metadata or selectors yet).

## Proposal

I think Substrate needs a namespace concept for similar reasons, and in fact we should take it further.  We should not have non-namespaced concepts at all.  We don't have to call them “ namespaces”, of course. They could be "subspaces" or "namestrates" or "users" or "realms" or "fiefs" or "forests" or "districts" or ...  For clarity in this doc, I will say "subspace".  These "subspaces" would NOT be mapped to Kubernetes namespaces; they would exist only in Agent substrate.

Actors would exist under subspaces.  Any single substrate-namespace could hold actors from any number of ActorTemplates, regardless of the k8s namespace those came from.  Access control would be governed first at the subspace layer. A single subspace could be a single user, or a team, whatever makes sense for the user. Most operations would be scoped to a subspace, just like Kubernetes. Perhaps we need nested subspaces (e.g. tenant->team->project->tree->session), but that would need to be explored and justified independently.

Unlike Kubernetes, things like Workers would ALSO exist in subspaces.  These could either be "special" subspaces or just regular subspaces with policies governing access control.

@HavenXia @juli4n

brandonrjacobs:
[...]

PR #280, opened 2026-06-19
Atespace: per-tenant scoping for the actor lifecycle

HavenXia:
This is part of solution for #21.

First incremental slice of the [atespace design](http://docs/document/d/1zTfgZ8UrYfjLysxdWavmIHfAeCkMD1NkpH1SvOJTYEc?tab=t.0) for actors — part of #21. An atespace is a mandatory tenant boundary that every actor belongs to. It's folded into the actor's identity and storage key (`actor:<atespace>:<id>`), so list/get/delete within a tenant is a cheap key-prefix operation, and actors in different atespaces can reuse the same id without colliding.

This PR adds atespace through the actor lifecycle end-to-end (proto → store → control API → `kubectl-ate`). 

It doesn't touch DNS, snapshots, scheduling, or auth. Landing it on its own so the future changes are additive, and everything that isn't actor-CRUD is explicitly out of scope below.

## This PR changes: 

- Proto — every actor RPC request carries an atespace; Actor carries it as part of its identity.
- Store — actors are `keyed actor:<atespace>:<id>` - `GetActor/DeleteActor/ListActors` take an atespace. Listing is a scoped SCAN `actor:<atespace>:*`, or SCAN `actor:*` for all atespaces.
- Control API — create / get / delete / suspend / pause / resume / update are all atespace-scoped. atespace is required and validated as a `DNS-1123` label. The syncer's dead-worker recovery is atespace-aware by adding `Worker.actor_atespace`.
- Atespace object + CRUD — atespace with `CreateAtespace / GetAtespace / ListAtespaces / DeleteAtespace`. `CreateActor` now requires the atespace to exist first (`FailedPrecondition` otherwise); `DeleteAtespace` only removes an empty atespace.
- changes on `kubectl-ate`:
  - `--atespace` / `-a` on every actor subcommand (create/get/delete/resume/suspend/pause/logs).
  - `-A` / `--all-atespaces` to list across all tenants.
  - Add an `ATESPACE` column in the actor table, the existing namespace column is renamed `TEMPLATE NS` to disambiguate it from the atespace.
  - Create / get / delete atespace subcommands.

## Examples
Scope a listing to one tenant (-a is shorthand for --atespace):

### Atespace Creation
```
$ kubectl ate create atespace team-a
NAME
team-a
$ kubectl ate create atespace team-b
NAME
team-b
```

### Actor Creation
```
$ kubectl ate create actor test -t ate-demo-counter/counter --atespace team-a
ATESPACE   TEMPLATE NS        TEMPLATE   ID     STATUS             ATEOM POD   ATEOM IP   VERSION
team-a     ate-demo-counter   counter    test   STATUS_SUSPENDED   <none>                 1

$ kubectl ate create actor test2 -t ate-demo-counter/counter --atespace team-a
ATESPACE   TEMPLATE NS        TEMPLATE   ID      STATUS             ATEOM POD   ATEOM IP   VERSION
team-a     ate-demo-counter   counter    test2   STATUS_SUSPENDED   <none>                 1

$ kubectl ate create actor test -t ate-demo-counter/counter --atespace team-b
ATESPACE   TEMPLATE NS        TEMPLATE   ID     STATUS             ATEOM POD   ATEOM IP   VERSION
team-b     ate-demo-counter   counter    test   STATUS_SUSPENDED   <none>                 1

# Fail when atespace not exist
$ kubectl ate create actor test -t ate-demo-counter/counter --atespace team-c
Error: failed to create actor: rpc error: code = FailedPrecondition desc = Atespace team-c not found
```

### Get by atespaces
```
$ kubectl ate get actors -A
ATESPACE   TEMPLATE NS        TEMPLATE   ID      STATUS             ATEOM POD   ATEOM IP   VERSION
team-a     ate-demo-counter   counter    test    STATUS_SUSPENDED   <none>                 1
team-a     ate-demo-counter   counter    test2   STATUS_SUSPENDED   <none>                 1
team-b     ate-demo-counter   counter    test    STATUS_SUSPENDED   <none>                 1

$ kubectl ate get actors -a team-a
ATESPACE   TEMPLATE NS        TEMPLATE   ID      STATUS             ATEOM POD   ATEOM IP   VERSION
team-a     ate-demo-counter   counter    test    STATUS_SUSPENDED   <none>                 1
team-a     ate-demo-counter   counter    test2   STATUS_SUSPENDED   <none>                 1

$ kubectl ate get actors -a team-b
ATESPACE   TEMPLATE NS        TEMPLATE   ID     STATUS             ATEOM POD   ATEOM IP   VERSION
team-b     ate-demo-counter   counter    test   STATUS_SUSPENDED   <none>                 1
```

###  Resume & Suspend
```
$ kubectl ate resume actor test -a team-a
ATESPACE   TEMPLATE NS        TEMPLATE   ID     STATUS           ATEOM POD                                              ATEOM IP     VERSION
team-a     ate-demo-counter   counter    test   STATUS_RUNNING   ate-demo-counter/counter-deployment-5d9b77d6fd-r846n   10.52.3.86   3

$ kubectl ate get actors -a team-a
ATESPACE   TEMPLATE NS        TEMPLATE   ID      STATUS             ATEOM POD                                              ATEOM IP     VERSION
team-a     ate-demo-counter   counter    test    STATUS_RUNNING     ate-demo-counter/counter-deployment-5d9b77d6fd-r846n   10.52.3.86   3
team-a     ate-demo-counter   counter    test2   STATUS_SUSPENDED   <none>                                                              1

$ kubectl ate suspend actor test -a team-a
ATESPACE   TEMPLATE NS        TEMPLATE   ID     STATUS             ATEOM POD   ATEOM IP   VERSION
team-a     ate-demo-counter   counter    test   STATUS_SUSPENDED   <none>                 5
```

### Deletion
```
$kubectl ate delete actor test2 -a team-a
actor "test2" deleted

# Fail if deleting a non-empty atespace
$ kubectl ate delete atespace team-a
Error: failed to delete atespace: rpc error: code = FailedPrecondition desc = Atespace team-a is not empty

$ kubectl ate delete actor test -a team-a
actor "test" deleted

$ kubectl ate delete atespace team-a
atespace "team-a" deleted
```

Scope / non-goals

Deferred to later atespace increments (intentionally not in this PR): cascade delete of a non-empty atespace (soft-delete + finalizers; for now DeleteAtespace only removes an empty atespace), DNS names, snapshot paths, template grants, the worker's own (system) atespace, and quota.

- [x] Tests pass
- [x] Appropriate documentation

### DIFF
--- a/cmd/ateapi/internal/controlapi/create_actor.go
+++ b/cmd/ateapi/internal/controlapi/create_actor.go
@@ -49,6 +49,7 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
 		ActorTemplateNamespace: req.GetActorTemplateNamespace(),
 		ActorTemplateName:      req.GetActorTemplateName(),
 		WorkerSelector:         req.GetWorkerSelector(),
+		Atespace:               req.GetAtespace(),
 	}
 	err = s.persistence.CreateActor(ctx, actor)
 	if err != nil {
@@ -58,7 +59,7 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
 		return nil, fmt.Errorf("while recording actor: %w", err)
 	}
 
-	storedActor, err := s.persistence.GetActor(ctx, id)
+	storedActor, err := s.persistence.GetActor(ctx, req.GetAtespace(), id)
 	if err != nil {
 		return nil, fmt.Errorf("while fetching recorded actor from DB: %w", err)
 	}
@@ -79,6 +80,12 @@ func validateCreateActorRequest(req *ateapipb.CreateActorRequest) error {
 		return status.Error(codes.InvalidArgument, "actor_id is required")
 	}
 	if err := resources.ValidateActorID(req.GetActorId()); err != nil {
+		return status.Error(codes.InvalidArgument, err.Error())
+	}
+	if req.GetAtespace() == "" {
+		return status.Error(codes.InvalidArgument, "atespace is required")
+	}
+	if err := resources.ValidateAtespace(req.GetAtespace()); err != nil {
 		return status.Error(codes.InvalidArgument, err.Error())
 	}
 	if err := validateSelector(req.GetWorkerSelector()); err != nil {
--- a/cmd/ateapi/internal/controlapi/create_actor.go
+++ b/cmd/ateapi/internal/controlapi/create_actor.go
@@ -41,6 +41,15 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
 		return nil, fmt.Errorf("while getting ActorTemplate: %w", err)
 	}
 
+	// The atespace must already exist.
+	exists, err := s.persistence.AtespaceExists(ctx, req.GetAtespace())
+	if err != nil {
+		return nil, fmt.Errorf("while checking atespace: %w", err)
+	}
+	if !exists {
+		return nil, status.Errorf(codes.FailedPrecondition, "Atespace %s not found", req.GetAtespace())
+	}
+
 	id := req.GetActorId()
 	actor := &ateapipb.Actor{
 		ActorId:                id,
--- a/cmd/ateapi/internal/controlapi/create_actor.go
+++ b/cmd/ateapi/internal/controlapi/create_actor.go
@@ -42,15 +42,15 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
 	}
 
 	// The atespace must already exist.
-	exists, err := s.persistence.AtespaceExists(ctx, req.GetAtespace())
+	exists, err := s.persistence.AtespaceExists(ctx, req.GetActorRef().GetAtespace())
 	if err != nil {
 		return nil, fmt.Errorf("while checking atespace: %w", err)
 	}
 	if !exists {
-		return nil, status.Errorf(codes.FailedPrecondition, "Atespace %s not found", req.GetAtespace())
+		return nil, status.Errorf(codes.FailedPrecondition, "Atespace %s not found", req.GetActorRef().GetAtespace())
 	}
 
-	id := req.GetActorId()
+	id := req.GetActorRef().GetName()
 	actor := &ateapipb.Actor{
 		ActorId:                id,
 		Version:                1,
@@ -58,7 +58,7 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
 		ActorTemplateNamespace: req.GetActorTemplateNamespace(),
 		ActorTemplateName:      req.GetActorTemplateName(),
 		WorkerSelector:         req.GetWorkerSelector(),
-		Atespace:               req.GetAtespace(),
+		Atespace:               req.GetActorRef().GetAtespace(),
 	}
 	err = s.persistence.CreateActor(ctx, actor)
 	if err != nil {
@@ -68,7 +68,7 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
 		return nil, fmt.Errorf("while recording actor: %w", err)
 	}
 
-	storedActor, err := s.persistence.GetActor(ctx, req.GetAtespace(), id)
+	storedActor, err := s.persistence.GetActor(ctx, req.GetActorRef().GetAtespace(), id)
 	if err != nil {
 		return nil, fmt.Errorf("while fetching recorded actor from DB: %w", err)
 	}
@@ -85,16 +85,16 @@ func validateCreateActorRequest(req *ateapipb.CreateActorRequest) error {
 	if req.GetActorTemplateName() == "" {
 		return status.Error(codes.InvalidArgument, "actor_template_name is required")
 	}
-	if req.GetActorId() == "" {
+	if req.GetActorRef().GetName() == "" {
 		return status.Error(codes.InvalidArgument, "actor_id is required")
 	}
-	if err := resources.ValidateActorID(req.GetActorId()); err != nil {
+	if err := resources.ValidateActorID(req.GetActorRef().GetName()); err != nil {
 		return status.Error(codes.InvalidArgument, err.Error())
 	}
-	if req.GetAtespace() == "" {
+	if req.GetActorRef().GetAtespace() == "" {
 		return status.Error(codes.InvalidArgument, "atespace is required")
 	}
-	if err := resources.ValidateAtespace(req.GetAtespace()); err != nil {
+	if err := resources.ValidateAtespace(req.GetActorRef().GetAtespace()); err != nil {
 		return status.Error(codes.InvalidArgument, err.Error())
 	}
 	if err := validateSelector(req.GetWorkerSelector()); err != nil {
--- a/cmd/ateapi/internal/controlapi/pause_actor.go
+++ b/cmd/ateapi/internal/controlapi/pause_actor.go
@@ -29,7 +29,7 @@ func (s *Service) PauseActor(ctx context.Context, req *ateapipb.PauseActorReques
 		return nil, err
 	}
 
-	actor, err := s.actorWorkflow.PauseActor(ctx, req.GetActorId())
+	actor, err := s.actorWorkflow.PauseActor(ctx, req.GetAtespace(), req.GetActorId())
 	if err != nil {
 		if errors.Is(err, store.ErrPersistenceRetry) {
 			return nil, status.Error(codes.Aborted, "concurrent update conflict, please retry")
@@ -46,6 +46,9 @@ func (s *Service) PauseActor(ctx context.Context, req *ateapipb.PauseActorReques
 func validatePauseActorRequest(req *ateapipb.PauseActorRequest) error {
 	if req.GetActorId() == "" {
 		return status.Error(codes.InvalidArgument, "id is required")
+	}
+	if req.GetAtespace() == "" {
+		return status.Error(codes.InvalidArgument, "atespace is required")
 	}
 	return nil
 }
--- a/cmd/ateapi/internal/controlapi/pause_actor.go
+++ b/cmd/ateapi/internal/controlapi/pause_actor.go
@@ -29,13 +29,13 @@ func (s *Service) PauseActor(ctx context.Context, req *ateapipb.PauseActorReques
 		return nil, err
 	}
 
-	actor, err := s.actorWorkflow.PauseActor(ctx, req.GetAtespace(), req.GetActorId())
+	actor, err := s.actorWorkflow.PauseActor(ctx, req.GetActorRef().GetAtespace(), req.GetActorRef().GetName())
 	if err != nil {
 		if errors.Is(err, store.ErrPersistenceRetry) {
 			return nil, status.Error(codes.Aborted, "concurrent update conflict, please retry")
 		}
 		if errors.Is(err, store.ErrNotFound) {
-			return nil, status.Errorf(codes.NotFound, "Actor %s not found", req.GetActorId())
+			return nil, status.Errorf(codes.NotFound, "Actor %s not found", req.GetActorRef().GetName())
 		}
 		return nil, err
 	}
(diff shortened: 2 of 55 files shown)


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -125,10 +125,21 @@ message Actor {
...
+  string atespace = 15;
+}
+
+// Atespace is the tenant boundary an Actor is created into. Placeholder for now
+// (name only).
+message Atespace {
+  string name = 1;
 }
 
 message GetActorRequest {
```
thockin, line 181: Should we just join `name` and `atespace` into a message and use it everywhere?  It seems like there's never a context where we want one and not the other?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -125,10 +125,21 @@ message Actor {
...
   string worker_pool_name = 14;
+
+  // The atespace (tenant boundary) this actor belongs to. Part of the actor's
+  // resource identity; folded into the Redis key as actor:<atespace>:<actor_id>.
+  string atespace = 15;
+}
+
+// Atespace is the tenant boundary an Actor is created into. Placeholder for now
+// (name only).
+message Atespace {
```
thockin, line 147: I assumed I would find CRUD APIs for Atespaces, but I don't see them?

cmd/ateapi/internal/store/store.go
```diff
@@ -49,10 +49,10 @@ type Interface interface {
 	UpdateActor(ctx context.Context, actor *ateapipb.Actor, expectedVersion int64) error
 
 	// Removes an actor. Returns ErrNotFound if missing, or ErrFailedPrecondition if not suspended.
-	DeleteActor(ctx context.Context, id string) error
+	DeleteActor(ctx context.Context, atespace, id string) error
 
-	// Lists all known actors. Returns a page of actors and a next page token.
-	ListActors(ctx context.Context, pageSize int32, pageToken string) ([]*ateapipb.Actor, string, error)
+	// Lists actors in the given atespace (scoped scan). Returns a page of actors and a next page token.
+	ListActors(ctx context.Context, atespace string, pageSize int32, pageToken string) ([]*ateapipb.Actor, string, error)
```
thockin, line 56: Does "" mean "all atespaces" ?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -149,6 +160,9 @@ message CreateActorRequest {
   // worker_selector sets the actor's placement constraint at creation time.
   // If empty, the actor matches any pool admitted by the template's selector.
   Selector worker_selector = 4;
+
+  // The atespace to create the actor into.
+  string atespace = 5;
```
thockin: I appreciate the caution of no renumbering proto tags, but we should strive to make this as readable as possible, so I would put it next to ID

cmd/kubectl-ate/internal/cmd/create_actor.go
```diff
@@ -60,5 +62,7 @@ var createActorCmd = &cobra.Command{
 func init() {
 	createActorCmd.Flags().StringVarP(&templateFlag, "template", "t", "", "Template to derive the actor from in <namespace>/<name> format (required)")
 	_ = createActorCmd.MarkFlagRequired("template")
+	createActorCmd.Flags().StringVar(&atespaceFlag, "atespace", "", "Atespace (tenant) to create the actor in (required)")
```
thockin: Let's avoid the word "tenant"

cmd/atecontroller/internal/controllers/actortemplate_controller.go
```diff
@@ -73,6 +73,7 @@ func (r *ActorTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 		createReq := &ateapipb.CreateActorRequest{
 			ActorId:                actorID,
+			Atespace:               at.ObjectMeta.Namespace,
```
thockin: This is not obvious to me -- we should never assume k8s namespaces == atespaces -- they are not the same.

We are taking a snapshot, so we should probably use an atespace that is specifically put aside for this.  If we move to CRUD of atespaces as a first-class resource, we should "reserve" any atespace whose name being with "ate-", so this can be something like "ate-golden"?

cmd/atenet/internal/dns/corefile.go
```diff
@@ -39,11 +39,12 @@ func buildTemplate() string {
...
 	directives = append(directives, "reload")
 
-	// Construct match pattern for <ActorID>.<dnsDomain>.
+	// Construct match pattern for <ActorID>.<atespace>.<dnsDomain>. Both the
+	// actor id and the atespace are DNS-1123 labels (same regex).
 	directives = append(directives, fmt.Sprintf("template IN A %s {", resources.ActorDNSSuffix))
 	dnsDomainParts := strings.Split("."+resources.ActorDNSSuffix+".", ".")
 	dnsDomainRef := strings.Join(dnsDomainParts, `\.`)
-	directives = append(directives, fmt.Sprintf(`  match "^%s%s$"`, resources.ActorIDRegexPattern, dnsDomainRef))
+	directives = append(directives, fmt.Sprintf(`  match "^%s\.%s%s$"`, resources.ActorIDRegexPattern, resources.ActorIDRegexPattern, dnsDomainRef))
```
thockin: This is fishy -- why do we need an explicit `\.` in one place but not the other?

Debugger says: the input is ".actors.resources.substrate.ate.dev." (leading and trailing dots).  That makes `dnsDomainParts` have empty first and last entries.  That means `dnsDomainRef` has leading and trailing `\.`.  That makes this correct.

It's WEIRD but correct.  I'll send a followup to document it

internal/resources/actor.go
```diff
@@ -57,3 +58,32 @@ func ValidateAtespace(atespace string) error {
...
+// part of the name because an actor id is only unique within its atespace.
+func ActorDNSName(atespace, actorID string) string {
+	return actorID + "." + atespace + "." + ActorDNSSuffix
+}
+
+// ParseActorDNSName parses a mesh DNS name of the form
+// "<actor_id>.<atespace>.actors.resources.substrate.ate.dev" (a trailing dot is
+// tolerated) into its atespace and actor id, validating both. It does not accept
+// a host:port; callers must strip the port first.
+func ParseActorDNSName(name string) (atespace, actorID string, err error) {
```
thockin, line 79: This seems like a perfect target for a small unit test?

thockin (commented):
Currently an atespace is created merely by using it in an actor.  I think we want to move to Atespaces being explicitly created and managed, right?  I'm fine with that as a followup, or as more commits in here.

pkg/proto/ateapipb/ateapi.proto
dberkov: Agree, we are still in a phase, we can break API. Just put a comment in PR that this is a breaking change and all actors needs to be deleted.

In any case, all existing actors will not work, since they will miss namespace.

cmd/ateapi/internal/controlapi/create_actor.go
```diff
@@ -81,6 +82,12 @@ func validateCreateActorRequest(req *ateapipb.CreateActorRequest) error {
 	if err := resources.ValidateActorID(req.GetActorId()); err != nil {
 		return status.Error(codes.InvalidArgument, err.Error())
 	}
+	if req.GetAtespace() == "" {
+		return status.Error(codes.InvalidArgument, "atespace is required")
+	}
+	if err := resources.ValidateAtespace(req.GetAtespace()); err != nil {
```
dberkov: I am curious, why "resources.ValidateAtespace" does not validate for emptry string. 
I see similar behavior is defined for resources.ValidateActorID too.
Might be we need to fix all the "resource validator" to check for empty string.

@juli4n - WDYT?

cmd/ateapi/internal/controlapi/pause_actor.go
```diff
@@ -47,5 +47,8 @@ func validatePauseActorRequest(req *ateapipb.PauseActorRequest) error {
 	if req.GetActorId() == "" {
 		return status.Error(codes.InvalidArgument, "id is required")
 	}
+	if req.GetAtespace() == "" {
```
dberkov: have you considered to have full valudation? resources.ValidateNamespace.

It might worth to add similar full validation for actorID itself.

P.S - please make similar fix for all APIs, if you decide to accept the comment.

cmd/ateapi/internal/controlapi/create_actor.go
thockin: We need to bring some consistency to validation early.  K8s left it all open-coded for far too long and it got very painful.

cmd/ateapi/internal/controlapi/pause_actor.go
thockin: While I 200% agree that we need rigorous validation done at the right moment in the stack, I don't think this PR should block on it.  We can TODO it and come back

pkg/proto/ateapipb/ateapi.proto
HavenXia: Discussed with Julian and added in new commits(not pushed yet)

pkg/proto/ateapipb/ateapi.proto
HavenXia: That sounds neat, I will make them merged into an `ActorReference`.

cmd/ateapi/internal/store/store.go
HavenXia: Yes — for ListActors only, an empty atespace means "all atespaces" (it backs `kubectl ate get actors -A`). Every other operation requires a non-empty atespace. I documented it at https://github.com/agent-substrate/substrate/pull/280/changes#diff-101ee4ffedf6cf67cec6ea55a251c663205e7333f629d1e3edbb397b2c6e3017R49 but not the interface here, added comments to make it clear to readers.

pkg/proto/ateapipb/ateapi.proto
HavenXia: Resolved by `ActorReference` — actor_id and atespace are now a single `ActorReference ref = 1` at the top of each request, so there's no separate atespace field left to reposition.

cmd/kubectl-ate/internal/cmd/create_actor.go
HavenXia: removed

cmd/atecontroller/internal/controllers/actortemplate_controller.go
HavenXia: Sounds good, will update to use "ate-golden".

internal/resources/actor.go
HavenXia: Thanks, added.

cmd/ateapi/internal/controlapi/create_actor.go
HavenXia: Marked that as a TODO, will create an issue to track the unification of actor/atespace validation across the control API RPCs.

cmd/ateapi/internal/controlapi/pause_actor.go
HavenXia: Marked that as a TODO, will create an issue to track the unification of actor/atespace validation across the control API RPCs.

cmd/atenet/internal/dns/corefile.go
HavenXia: I rewrote it to be explicit — escape the suffix's dots and spell out every \. in the pattern to improve readability.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -125,10 +137,49 @@ message Actor {
   // suspend/pause since eligibility is no longer a single fixed pool
   // reference on the ActorTemplate.
   string worker_pool_name = 14;
+
+  // The atespace (tenant boundary) this actor belongs to. Part of the actor's
+  // resource identity; folded into the Redis key as actor:<atespace>:<actor_id>.
+  string atespace = 15;
```
thockin, line 143: Should this + actor_id be replaced by ActorRef ?  OK to do as a followup, but asking if there's a reason not to?

cmd/kubectl-ate/internal/cmd/create_atespace.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/agent-substrate/substrate/cmd/kubectl-ate/internal/printer"
+	"github.com/agent-substrate/substrate/internal/ateclient"
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+	"github.com/spf13/cobra"
+)
+
+var createAtespaceCmd = &cobra.Command{
+	Use:   "atespace [name]",
+	Short: "Create an atespace (tenant boundary)",
```
thockin, line 28: I think we should get rid of this sort of parenthetical.  It's useful TODAY but we don't need it long term

thockin:
Fire in the hole!

PR #281, opened 2026-06-19
fix(atelet): stream sandbox asset downloads instead of buffering in memory

dims:
Previously, fetchAsset retrieved each asset through FetchFromGCS, which reads the entire object into a byte slice before hashing and writing it. Peak memory therefore scaled with the asset size. This was acceptable for the roughly 50 MB runsc binary, but when we go to micro-VM(s) the runtime's kernel and root filesystem images range from hundreds of megabytes to several gigabytes and would exhaust the memory of atelet, which is shared across all actors on a node.

The download now streams directly to the temporary file, computing the SHA-256 digest in the same pass via io.MultiWriter and bounding the transfer with an io.LimitReader. Peak memory is now the size of the io.Copy buffer, independent of the asset size. The LimitReader also provides a secondary safeguard for disk usage against a misconfigured or malicious URL that returns an unbounded stream. The limit is 8 GiB and is declared as a variable so that tests may lower it.

The digest is verified only after the copy completes. A size or hash failure therefore leaves the data at the temporary path, which is subsequently removed; it is never renamed to the content-addressed cache path, so a failed download cannot corrupt the cache.

This change also introduces ategcs.Open, a streaming reader, along with the accompanying TestFetchAssetStreaming.

- [ ] Tests pass

### DIFF
--- a/cmd/atelet/internal/ategcs/objects.go
+++ b/cmd/atelet/internal/ategcs/objects.go
@@ -58,6 +58,20 @@ func FetchFromGCS(ctx context.Context, client ObjectStorage, gsURL string) ([]by
 	return content, nil
 }
 
+// Open streams the object at gsURL; the caller must Close the returned reader.
+// Unlike FetchFromGCS it does not buffer the whole object in memory.
+func Open(ctx context.Context, client ObjectStorage, gsURL string) (io.ReadCloser, error) {
+	bucket, object, err := parseGCSURL(gsURL)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


dims:
cc @thockin @BenTheElder

EItanya (approved):
Great idea

BenTheElder (approved):
Thanks! 

I have some other related optimizations to GCS in the uVM branch, let me merge this and rebase and send those :-)

Issue #220, opened 2026-06-11
[Feature] Distinct lifecycle for “rootfs” and memory snapshots vs. “working” space. Needs API surface of where to mount.


chenggui53:
i've see this in [roadmap](https://github.com/agent-substrate/substrate/blob/main/docs/roadmap.md#storage), when will this function be ready? we realy need mount workspace volume when actor startup

BenTheElder:
It's an alpha requirement, so we're aiming for end of month.

https://github.com/agent-substrate/substrate/issues/119 covers it (search issues for "homedir").

I've been looking at this, it will also have overlap with #123 .

> we realy need mount workspace volume when actor startup

Do you need to mount a volume, or do you need persistence? (which may or may not involve a volume)

chenggui53:
@BenTheElde mount a volume, in our use case, we need mount a given volume when actor resume, this volume will not define in actortemplate, because we need resume many actor from a single actor template, each actor has its own workspace (from a volume mount)

dberkov:
@chenggui53 

>@BenTheElde mount a volume, in our use case, we need mount a given volume when actor resume, this volume will not define in actortemplate, because we need resume many actor from a single actor template, each actor has its own workspace (from a volume mount)

Are you referring to external volume , like NFS / Block storage or a different folder on the rootfs?

chenggui53:
> [chenggui53 (@chenggui53)](https://github.com/chenggui53)
> 
> > @BenTheElde mount a volume, in our use case, we need mount a given volume when actor resume, this volume will not define in actortemplate, because we need resume many actor from a single actor template, each actor has its own workspace (from a volume mount)
> 
> Are you referring to external volume , like NFS / Block storage or a different folder on the rootfs?

@dberkov external volume mount when resume an actor. for each actor persist its data in different volume

thockin:
https://github.com/agent-substrate/substrate/pull/295

EItanya:
I'm going to close this as #295 has merged. External Volumes are already covered by #232 as well

Issue #172, opened 2026-06-03
Snapshot local storage management


HavenXia:
Something to manage the local state of snapshots.  How are snapshots stored / protected? How are they secured across tenants? What's its security model? 

We need to secure the snapshot access for local snapshots stored on nodes.

HavenXia:
stale issue

PR #282, opened 2026-06-22
feat: add network storage volume mount support for actors

### LINKED ISSUE #232: [Feature] External volume mounts
Continuing the discussion in https://github.com/agent-substrate/substrate/issues/119#issuecomment-4687231897 and https://github.com/agent-substrate/substrate/issues/220, let's start discussing the use cases and requirements for external volume mounts.

## Use Cases
One use case I see discussed in the above issues is the need for each actor to have their own workspace to store data. Rootfs or working dir is insufficient because:
- At larger capacities, there is a performance penalty of checkpointing local directories every time an actor is paused/suspended
- Actor requires higher persistence guarantees with every IO rather than at the pause/suspend timeframe.

### Open questions
- Do volumes need to be sharable across multiple actors? If so, as read-only or writeable?

## Proposal
Here's my initial thoughts on external volume semantics we can add to substrate.

### Lifecycle
External volumes are an independent filesystem mount that follows the Actor lifecycle. A unique volume will be created per actor. Once a volume is created, it is persisted across the lifecycle of the actor. It is deleted when the actor is deleted.

If we support actor updates in the future, the volume will persist across updates. A volume can be added/removed across updates, but an existing volume cannot be mutated (for now). When a volume is deleted from the template, the underlying volume is also deleted after the rollout has completed.

### Data Persistence Semantics
External volume mounts are expected to provide filesystem consistency semantics. Writes are persisted when flushed to the backing store. This is different than rootfs semantics where data is only guaranteed to be persisted when the actor is paused/suspended.

When an actor is paused/suspended, we do not take a snapshot of the volume mount (at least in the initial phase). We just unmount the volume and remount it when the actor is resumed. This implies that on resumption, the actor will get the latest volume state, regardless of if the actor gracefully completed its previous task or it crashed prematurely. The actor has to be able to handle the potential difference between its memory state and filesystem state.

In a future milestone, we can explore adding external volume snapshot support on actor suspension, and the ability to create actor volumes from a given volume snapshot.

### Storage Backends
The storage backend has to support the following functionality at the required performance / scale:
- Create/delete volumes
- Attach/detach volumes to nodes
- Mount/unmount volumes to nodes
- Security/isolation of volume mounts
- (Future) snapshot of volumes and clone from snapshot

NFS appears to be a commonly requested storage integration for this use case, however, while the file protocol is standard, the control plane APIs are typically vendor-specific. Rather than add a hard-dependency on NFS in substrate, I propose that we focus on supporting Container Storage Interface (CSI), which is well-defined and supported in the Kubernetes ecosystem.

### API
```
kind: ActorTemplate
metadata:
  name: my-agent-template
spec: 
  ...
  containers:
  - image: my-agent-image:v1.2.3
    volumeMounts:
    - name: vol1
      mountPath: /my-data
  volumes:
  - name: vol1
    externalVolumeTemplate:
      capacity: 5Gi
      storageClassName: storage-pool1
```

chenggui53:
## Summary

Add the ability to mount network storage (NFS, PVC, HostPath) into actor sandboxes when starting actors. This enables actors to access shared persistent storage for workspace, model cache, and other use cases.

Closes #232
Related: #220, #172

### CRD Types
- **WorkerPool**: Add `StorageVolume` type and `storageVolumes` field to declare available storage infrastructure (NFS, PVC, HostPath sources)
- **ActorTemplate**: Add `ActorVolume` type and `volumes` field to declare mount requirements (name, mountPath, subPath, readOnly)

### Proto
- `atelet.proto`: Add `VolumeMount` message and `volume_mounts` field to `WorkloadSpec`

### Controller
- WorkerPool controller injects storageVolumes as K8s volumes into worker pods and mounts them at `/var/lib/ateom-storage/<name>`

### API Server
- `workloadSpecFromActorTemplate` converts ActorTemplate volumes to proto VolumeMounts
- `CallAteletRestoreStep` validates volume refs against WorkerPool storageVolumes

### Atelet (OCI)
- `resolveVolumeMounts` converts proto mounts to host-path configs
- `buildActorOCISpec` creates OCI bind mounts for storage volumes
- Supports `${ACTOR_ID}` placeholder in SubPath for per-actor isolation

### Validation
- `ValidateVolumeMountPath` ensures mount paths are absolute and safe

## Usage

```yaml
# WorkerPool (infrastructure)
apiVersion: ate.dev/v1alpha1
kind: WorkerPool
metadata:
  name: gpu-pool
spec:
  replicas: 5
  ateomImage: gcr.io/project/ateom:latest
  storageVolumes:
  - name: shared-data
    nfs:
      server: nfs.example.com
      path: /exports/actor-data

# ActorTemplate (workload)
apiVersion: ate.dev/v1alpha1
kind: ActorTemplate
metadata:
  name: code-agent
spec:
  pauseImage: registry.k8s.io/pause:3.10@sha256:...
  containers:
  - name: agent
    image: gcr.io/project/agent@sha256:...
  volumes:
  - name: shared-data
    mountPath: /workspace
    subPath: ${ACTOR_ID}/workspace
```

## Testing
- ✅ Unit tests for OCI spec generation with volume mounts (7 new tests)
- ✅ CRD validation verified on kind cluster
- ✅ WorkerPod volume injection verified
- ✅ Full test suite passes (`go test ./...`)


thockin:
#295 has more general support for volumes, so I think this will fit better once that is in.  In truth, I am still unsure exactly how this is to fit in, but there's a lot in common, so we can re-consider this once #295 lands

chenggui53:
> #295 has more general support for volumes, so I think this will fit better once that is in. In truth, I am still unsure exactly how this is to fit in, but there's a lot in common, so we can re-consider this once #295 lands

Thanks @thockin — now that #295 has merged, I rebased on top of it and took a first pass at fitting network storage into that volume framework rather than adding a parallel one. Wanted to run the shape past you before going further.

The idea is to make network storage just another source in the same VolumeSource oneof #295 introduced for durableDir:
- WorkerPool.storageVolumes declares the infra (NFS / PVC / HostPath) — #295 added nothing at the pool level, so this part is additive.
- ActorTemplate volumes get a workerPoolStorage source next to durableDir, referenced by name; mounting reuses #295's per-container VolumeMount so both kinds share one path.
- The pool gates what an actor can request (can't mount storage the pool doesn't provide), mirroring durableDir. Being external shared storage, it's excluded from snapshots.

Does this match how you were picturing it fitting in?

chenggui53:
@msau42 ready for another look

msau42:
Hi apologies for the delayed response. I think the main concern I have is tying the volumes to the worker pool. This means every actor template can get access to the volume if it selects that worker pool. Also with the introduction of atespaces we should also consider the interaction there.

I have a POC where the volume lifecycle is tied to actor lifecycle instead of worker pool. Right now it only supports volume per actor instead of sharing an existing volume across multiple actors, but I think it can be extended: https://github.com/msau42/substrate/tree/mock-mount-dev

I plan to rebase my POC and get it into a mergable state next week.

Issue #228, opened 2026-06-12
Use a node-local overlayfs rootfs cache to eliminate per-restore untar during actor resume


chenggui53:
# Summary

During actor resume, nearly all wall-clock time is spent rebuilding the OCI rootfs from scratch (prepareOCIDirectory → untar), not restoring the checkpoint. This proposal builds on #166 and recommends overlayfs as the concrete first implementation: cache one extracted, read-only rootfs per immutable image digest on each node, and materialize each actor's bundle as a thin overlay mount instead of re-untarring the whole image.

# Background / Problem

Today, every Restore call in atelet does the following for each container (pause + application), in cmd/atelet/oci.go (prepareOCIDirectory):

1. os.RemoveAll(rootfs) — wipe the actor rootfs
2. os.MkdirAll(rootfs)
3. pullCache.Fetch(ref) — fetch the flattened image tar (the memorypullcache avoids re-pulling, but not re-extracting)
4. untar(...) — extract the entire rootfs into the actor bundle

This repeats the full rootfs materialization on every resume, even when the same image digest has already been extracted on the same node many times before.

Observed cost (coding-agent-style image, consistent with #166):
> prepareOCIDirectory (untar rootfs)     ~15–20 s 
> Actual runsc restore (checkpoint restore)     ~268 ms  

So resume latency is dominated ~99% by rootfs extraction, not by the checkpoint restore itself. This is especially painful for larger agent images (e.g. Node-based images) that bundle a lot of files.

There is currently no overlayfs, copy-on-write, reflink, or layer-dedup mechanism: the rootfs is a flat directory that is torn down and rebuilt on every resume.

# Proposed Approach: node-local digest-keyed overlay

Change the scaling behavior from "every restore pays extraction cost" to "first restore of a digest on a node pays extraction; later restores pay only an overlay mount."

On first use of an image digest on a node
- Pull + extract the flattened rootfs once into a node-local, read-only cache directory keyed by the image digest (not tag), e.g. BasePath/rootfs-cache/<sha256>/lower.

On every restore using the same digest
- Instead of RemoveAll + untar, set up an overlay for the actor bundle's rootfs:
  - lowerdir = the cached, read-only extracted rootfs (shared, never mutated)
  - upperdir + workdir = per-actor, actor-private writable layers
- Then proceed directly to runsc create + runsc restore.

This keeps the cached lower layer immutable and gives each actor an isolated writable layer, so container writes never pollute the shared cache.

[...]

PR #283, opened 2026-06-22
feat: use node-local overlayfs rootfs cache to eliminate per-restore untar (#228)

chenggui53:
## Summary

Implements [#228](https://github.com/agent-substrate/substrate/issues/228): cache one extracted, read-only rootfs per immutable image digest on each node, and materialize each actor's bundle as a thin overlayfs mount instead of re-untarring the whole image on every restore.

## Problem

Every `Restore` call in atelet fully reconstructs the rootfs by pulling and untarring the OCI image — even when the same image digest has already been extracted on the same node many times before. Observed cost:

- `prepareOCIDirectory` (untar rootfs): **~15–20s**
- `runsc restore` (checkpoint restore): **~268ms**

Rootfs extraction dominates resume latency by ~99%.

## Solution

Change the scaling behavior from "every restore pays extraction cost" to "first restore of a digest on a node pays extraction; later restores pay only an overlay mount."

**On first use of an image digest on a node:**
- Pull + extract the flattened rootfs once into a node-local, read-only cache directory keyed by the image digest: `/var/lib/ateom-gvisor/rootfs-cache/<sha256>/lower/`

**On every restore using the same digest:**
- Instead of `RemoveAll` + `untar`, set up an overlayfs mount for the actor bundle's rootfs:
  - `lowerdir` = the cached, read-only extracted rootfs (shared, never mutated)
  - `upperdir` + `workdir` = per-actor, actor-private writable layers

## Changes

| File | Operation | Description |
|------|-----------|-------------|
| `internal/ateompath/ateompath.go` | Modified | Added `RootfsCacheDir` and `RootfsCacheLowerDir()` |
| `cmd/atelet/internal/rootfscache/rootfscache.go` | **New** | Core cache module: `EnsureRootfs`, `Untar`, `ValidateTarName`, LRU eviction, concurrent dedup |
| `cmd/atelet/internal/rootfscache/rootfscache_test.go` | **New** | Unit tests: cache miss/hit, concurrent safety, partial cleanup, eviction, digest validation |
| `cmd/atelet/overlay.go` | **New** | overlayfs mount/unmount helpers + `isOverlayfsAvailable()` |
| `cmd/atelet/oci.go` | Modified | `prepareOCIDirectory` integrates overlayfs path with untar fallback; `extractDigestFromRef`; `unmountActorRootfs` |
| `cmd/atelet/main.go` | Modified | Creates rootfs cache, wires into `AteomHerder`, `resetActorDirs` adds unmount before cleanup |

## Key Design Decisions

1. **Fallback safety**: tag-only refs (no digest) automatically fall back to the existing untar path
2. **Concurrent safety**: per-digest `inflightEntry` dedup — N goroutines requesting the same digest only trigger 1 untar
3. **Crash safety**: `.ready` sentinel file; `loadIndex` auto-cleans partial entries from previous crashes
4. **Eviction**: LRU by `.last_access` timestamp, async trigger, 20GB default cap
5. **Unmount cleanup**: `resetActorDirs` does `MNT_DETACH` unmount on overlayfs rootfs before `RemoveAll`

## Expected Impact

| Scenario | Before | After |
|----------|--------|-------|
| First restore (same node + digest) | ~15-20s | ~15-20s (populates cache) |
| Subsequent restore (cache hit) | ~15-20s | **<1s** (overlayfs mount) |
| Checkpoint restore | ~268ms | ~268ms (unchanged) |
| Total resume latency (cache hit) | ~15-20s | **<1.3s** |

## Testing

- ✅ All unit tests pass (`go test ./cmd/atelet/...`)
- ✅ Rootfscache tests: cache miss, cache hit, concurrent misses, partial entry cleanup, LRU eviction, digest validation
- ✅ Existing oci_test.go tests continue to pass (untar, path traversal, symlink escape, hardlink escape)
- ✅ Built and deployed to kind cluster — atelet running, rootfs-cache directory initialized, overlayfs kernel module available

## Open Questions (for follow-up)

- Eviction policy tuning (size cap, reference counting)
- Tag-based images without digest — should we resolve digest via HEAD request?
- Observability: cache hit/miss rate metrics (counters already added, need Prometheus dashboard)


BenTheElder:
I'd like to hold this one momentarily while we land #123 (one of our POC2 / alpha milestone requirements), as we need to make it work there too and there's already a lot in flight.

I agree that we should do something roughly like this though, and appreciate the PR.

cmd/atelet/internal/rootfscache/rootfscache.go
```diff
@@ -0,0 +1,624 @@
...
+	"sync"
+	"time"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// DefaultMaxCacheBytes is the default disk budget for the rootfs cache (20 GiB).
+const DefaultMaxCacheBytes int64 = 20 * 1024 * 1024 * 1024
```
rogeroger-yu, line 41: Need to allow configurable size

cmd/atelet/internal/rootfscache/rootfscache.go
```diff
@@ -0,0 +1,624 @@
...
+
+// EvictLRU removes the least-recently-used cache entry and returns its digest
+// and size, or ("", 0) if the cache is empty.  This is exported for tests.
+func (c *Cache) EvictLRU() (string, int64) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	var oldest *entryState
+	for _, e := range c.entries {
+		if oldest == nil || e.lastAccess.Before(oldest.lastAccess) {
```
rogeroger-yu, line 471: There is no reference count between the cache entry and how many overlay mounts are currently using this lowerDir. lastAccess records "the most recent restore", not "whether it is still mounted and used".
As a result, the rootfs "in use" will be deleted, causing damage to the running container.

cmd/atelet/overlay.go
```diff
@@ -0,0 +1,87 @@
...
+func teardownOverlayfs(target string) error {
+	if err := unix.Unmount(target, unix.MNT_DETACH); err != nil {
+		return fmt.Errorf("unmounting overlayfs at %s: %w", target, err)
+	}
+	return nil
+}
+
+// isOverlayfsAvailable checks whether the overlayfs kernel module is available
+// by attempting a mount on a temporary directory.
+func isOverlayfsAvailable() bool {
```
rogeroger-yu: This function is unused

cmd/atelet/oci.go
```diff
@@ -61,22 +64,56 @@ func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryP
...
-		return fmt.Errorf("while clearing rootfs %q: %w", rootPath, err)
-	}
+	// Try the overlayfs cache path first.  This succeeds when:
+	//   1. rootfsCache is non-nil, AND
+	//   2. the image ref includes a digest (@sha256:…).
+	// On a cache hit, tarData is NOT consumed, so we can skip the untar
+	// entirely.  On a miss, the cache extracts and caches for next time.
+	digest := extractDigestFromRef(ref)
+	if rootfsCache != nil && digest != "" {
+		tarData, err := pullCache.Fetch(ctx, ref)
```
rogeroger-yu: Cache hits still pullCache.Fetch the entire tar ？

cmd/atelet/oci.go
```diff
@@ -61,22 +64,56 @@ func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryP
...
-	defer tarData.Close()
+		// Create the overlay mount target.
+		if err := os.MkdirAll(rootPath, 0o700); err != nil {
+			return fmt.Errorf("in os.MkdirAll for rootfs mount target: %w", err)
+		}
+
+		upperDir := path.Join(bundlePath, "upper")
+		workDir := path.Join(bundlePath, "work")
+		if err := setupOverlayfs(rootPath, lowerDir, upperDir, workDir); err != nil {
+			return fmt.Errorf("setting up overlayfs (lower=%s, target=%s): %w", lowerDir, rootPath, err)
```
rogeroger-yu: If setupOverlayfs (i.e. unix.Mount("overlay", ...)) fails — kernel without overlayfs, no CAP_SYS_ADMIN/EPERM, nested gVisor/runsc, mount-count limits, etc. — the whole restore returns an error instead of falling back to untar；
The PR description advertises a "digest-only fallback to the old untar path," but that fallback only triggers when there's no digest or no cache — it does nothing for a mount failure.

cmd/atelet/oci.go
```diff
@@ -61,22 +64,56 @@ func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryP
...
+		}
+
+		span.SetAttributes(attribute.String("rootfs_method", "overlay"))
+	} else {
+		// Fallback: no digest or no cache — extract directly (original path).
+		if err := os.RemoveAll(rootPath); err != nil {
```
rogeroger-yu, line 130: The normal flow relies on resetActorDirs deleting the whole bundleDir between uses — but if prepareOCIDirectory is ever called without a preceding reset, a leftover non-empty upper leaks the previous container instance's writes into the new container (isolation bug), or stacks a second overlay on an already-mounted rootPath.

BenTheElder:
Note: if you rebase on main we have uVM now (including e2e tests), also FYI #346.

uVM is already mounting the image rootfs ro with a tmpfs upper to handle writes, similar to gVisor. I'm not sure we need the overlay on the atelet side vs just caching the unpacked images. We would need overlay if we started handling image layer sharing but that has other complexity.

PR #285, opened 2026-06-22
feat: add SandboxClass support to ActorTemplate

BenTheElder:
Big UX improvement for iterating on #123 

Follow-up to https://github.com/agent-substrate/substrate/pull/270 (previously worker pools were directly referenced by actors so you just made sure to pick an appropriate workerpool, we discussed there)

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -1131,7 +1131,7 @@ func TestResumeActor_NoEligiblePool(t *testing.T) {
 	_, err = tc.client.ResumeActor(context.Background(), &ateapipb.ResumeActorRequest{
 		ActorId: createResp.GetActor().GetActorId(),
 	})
-	assertGrpcError(t, err, codes.FailedPrecondition, "no worker pool matches the template and actor selectors")
+	assertGrpcError(t, err, codes.FailedPrecondition, "no worker pool matches the template's sandboxClass and the template/actor selectors")
 }
 
 // TestResumeActor_MultiPoolSelector exercises the AND-of-two-selectors path
--- a/cmd/ateapi/internal/controlapi/workflow_resume.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume.go
@@ -80,7 +80,7 @@ func (s *LoadActorForResumeStep) Execute(ctx context.Context, input *ResumeInput
 
 func (s *LoadActorForResumeStep) RetryBackoff() *wait.Backoff { return nil }
 
-func eligibleWorkerPools(pools []*atev1alpha1.WorkerPool, templateSelector *metav1.LabelSelector, actorSelector *ateapipb.Selector) (map[types.NamespacedName]struct{}, error) {
+func eligibleWorkerPools(pools []*atev1alpha1.WorkerPool, templateClass atev1alpha1.SandboxClass, templateSelector *metav1.LabelSelector, actorSelector *ateapipb.Selector) (map[types.NamespacedName]struct{}, error) {
 	templateSel := labels.Everything()
 	if templateSelector != nil {
 		sel, err := metav1.LabelSelectorAsSelector(templateSelector)
@@ -94,6 +94,13 @@ func eligibleWorkerPools(pools []*atev1alpha1.WorkerPool, templateSelector *meta
 
 	eligible := make(map[types.NamespacedName]struct{})
 	for _, pool := range pools {
+		// Snapshots are not portable across sandbox classes, so the pool's class
+		// must match the template's. This is a hard gate AND'd with the label
+		// selectors below. Both classes are populated by the CRD default (gvisor),
+		// so we compare them directly.
+		if pool.Spec.SandboxClass != templateClass {
+			continue
+		}
 		set := labels.Set(pool.GetLabels())
 		if templateSel.Matches(set) && actorSel.Matches(set) {
 			eligible[types.NamespacedName{Namespace: pool.GetNamespace(), Name: pool.GetName()}] = struct{}{}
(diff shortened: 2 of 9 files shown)


pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -157,6 +157,16 @@ type ActorTemplateSpec struct {
...
 	SnapshotsConfig SnapshotsConfig `json:"snapshotsConfig"`
 
+	// SandboxClass selects the sandbox runtime family this template's actors run
+	// on. Only worker pools whose SandboxClass matches are eligible. Snapshots are
+	// not portable across classes, so this is a hard gate, AND'd with WorkerSelector
+	// and the actor's worker_selector. Defaults to gvisor.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=gvisor;microvm
+	// +kubebuilder:default=gvisor
```
BenTheElder, line 177: Note: This is both for compatibility, and because ActorTemplate is immutable, so we can't cleanly pin it to whatever sandboxClass it lands on if unset ...

We might want to rethink the default and extend sandboxClass to an arbitrary string later ... but that seems like a later problem while we're still in alpha / POC-2.

Right now this unblocks having both worker pool types in one cluster with reasonable behavior.

docs/api-guide.md
```diff
@@ -81,12 +81,15 @@ The `ActorTemplate` defines the code, environment, and state-management policies
 | Field | Type | Description |
 | :--- | :--- | :--- |
 | `containers` | `[]Container` | **Required.** The workload definition (image, command, env, ports). |
+| `sandboxClass` | `string` | Optional. The sandbox runtime family this template's actors require: `gvisor` (default) or `microvm`. Only `WorkerPool`s whose `sandboxClass` matches are eligible. |
```
dims, line 84: we don't yet support `microvm` in the current code base, so hold off on announcing this in the api?

docs/api-guide.md
```diff
@@ -81,12 +81,15 @@ The `ActorTemplate` defines the code, environment, and state-management policies
...
 | `snapshotsConfig` | `SnapshotsConfig` | **Required.** GCS bucket and folder where memory snapshots are stored. |
 | `pauseImage` | `string` | **Required.** The image used for the sandbox root (e.g. `gcr.io/gke-release/pause`). |
 
 The sandbox binaries (e.g. the gVisor `runsc` binary) are **no longer configured on the `ActorTemplate`**. They are resolved from the referenced `WorkerPool`'s [`SandboxConfig`](#3-sandboxconfig-sandbox-binaries) — by name (`workerPool.spec.sandboxConfigName`) or, by default, the cluster default `SandboxConfig` for the pool's `sandboxClass`.
 
+Because a snapshot is not restorable across sandbox runtimes, `sandboxClass` is a **hard scheduling gate**: an actor is only ever placed on a `WorkerPool` of the matching class. It is AND'd with `workerSelector` (and the actor's `worker_selector`) — those can only narrow the eligible pools further, never cross class boundaries. It defaults to `gvisor` and, like the rest of the spec, is immutable, so each template's class is fixed at creation.
```
dims: `, never cross class boundaries` seem like extra words?

docs/api-guide.md
BenTheElder: This is a direct match to the sandboxconfig changes in #253

BenTheElder: That PR is next anyhow, and we don't support this in prod yet :-)

dims: ✅

docs/api-guide.md
BenTheElder: Agreed. Pushing ...

cmd/ateapi/internal/controlapi/functional_test.go
```diff
@@ -1131,7 +1131,7 @@ func TestResumeActor_NoEligiblePool(t *testing.T) {
 	_, err = tc.client.ResumeActor(context.Background(), &ateapipb.ResumeActorRequest{
 		ActorId: createResp.GetActor().GetActorId(),
 	})
-	assertGrpcError(t, err, codes.FailedPrecondition, "no worker pool matches the template and actor selectors")
+	assertGrpcError(t, err, codes.FailedPrecondition, "no worker pool matches the template's sandboxClass and the template and actor selectors")
```
dims: too many `and`'s

dims (approved):

pkg/api/v1alpha1/actortemplate_types.go
thockin: Can you leave a fat TODO here?  Along the lines of:

```
TODO: This is almost certainly insufficient.  We have to decide a number of things:
1) How does someone discover what classes are available, or what they mean?
2) How does someone define a new sandbox class?
3) Does a class mean the specific type of sandbox tech or does it include some aspect of config (e.g. can we have 2 different classes which both use gVisor with different config, or 2 classes which use different microvms)
4) How does the default get set and who sets it?

File: cmd/ateapi/internal/controlapi/workflow_resume.go [hunk 5]
```diff
@@ -80,7 +80,18 @@ func (s *LoadActorForResumeStep) Execute(ctx context.Context, input *ResumeInput
 
 func (s *LoadActorForResumeStep) RetryBackoff() *wait.Backoff { return nil }
 
-func eligibleWorkerPools(pools []*atev1alpha1.WorkerPool, templateSelector *metav1.LabelSelector, actorSelector *ateapipb.Selector) (map[types.NamespacedName]struct{}, error) {
+// defaultedSandboxClass treats an empty SandboxClass as gvisor, matching the
+// CRD defaulting on WorkerPool and ActorTemplate. This keeps the class match
+// correct even for objects constructed without server-side defaulting (e.g. in
+// tests).
+func defaultedSandboxClass(c atev1alpha1.SandboxClass) atev1alpha1.SandboxClass {
+	if c == "" {
```
thockin: given that we have `+kubebuilder:default` do we need this check?  I'd rather panic() or something - it should never not be set

thockin (commented):
Otherwise LGTM

cmd/ateapi/internal/controlapi/workflow_resume.go
BenTheElder: Persisted data? I'm OK dropping this.

pkg/api/v1alpha1/actortemplate_types.go
BenTheElder: Adding ...

BenTheElder: Something to keep in mind: This mirrors the workerpool portion, so this TODO applies to both.

BenTheElder: (Though we have _partially_ addressed 3))

cmd/ateapi/internal/controlapi/workflow_resume.go
BenTheElder: done

pkg/api/v1alpha1/actortemplate_types.go
BenTheElder: Added above

thockin:
LGTM

PR #287, opened 2026-06-23
microvm MVP

### LINKED ISSUE #123: [Feature] MicroVM support

BenTheElder:
Fixes #123 

cloud-hypervisor + kata (guest only).

- [x] Tests pass

TODO:
- CI coverage ... not sure if we can pull this off in GHA.

There are some other obvious TODOs that should be resolvable but are not resolved yet:
- We require a custom base image (for mkfs), and we have no where to host them currently, so it has to be built on demand.
- Only supports one container. (Should be a quick fix ... EDIT: [it is](https://github.com/BenTheElder/substrate/commits/microvm-multicontainer/) but it could stay a follow-up given the PR size ...)
- We have to hardcode the kvm mount etc in the workerpool controller. Punting generalizing that as pluggability is a long-term goal vs the short term POC-2 goals. We can think about how much config we want to expose there.
- cmd/setup-gcp isn't provisioning with nested virt yet, I added a node pool manually
- We probably need docs for setting up local dev. I'm using limactl on an M4 mac, with nested virt enabled (non-default config). The kind script will handle this IF the container env has KVM available.
- No `/run/ate` projected mount. Easy enough to fix ™️

I've started on these but I'm not blocking on them.


dims:
@BenTheElder I was able to test this with my openshell integration with some changes, please see:
https://github.com/agent-substrate/substrate/commit/c3aa08965d76e785118ab7aa2b24bc884bcd8c74

It's mostly timeout bumps but there's a suggestion on making building Ext4 images a bit more robust.

```
Verified on bigbox (CPU kind cluster, /dev/kvm + nested virt):
- #287's own counter-microvm demo passes (boot → suspend → resume, in-RAM count survives the VM snapshot).
- OpenShell helpdesk golden snapshot reaches Ready — the supervisor boots inside a kata/cloud-hypervisor VM and is checkpointed.
- A resumed user actor's agent serves traffic: GET /status → {"history_turns":0,"uptime_seconds":79.7,"model":"gpt-oss:20b-cloud"}.

What was broken + the fix

#287's MVP builds each actor's virtio-blk rootfs synchronously with mkfs.ext4 -d, bound by the resume RPC deadline. Sub-second for a tiny image; for OpenShell's 3.2 GB helpdesk image it takes minutes → the context cancels → mkfs … signal: killed, at four spots. The branch addresses all:
1. ateom-microvm golden-boot build → deadline-detached + idempotent (temp→atomic-rename).
2. ateom-microvm restore-path rebuild → deadline-detached.
3. ateapi ResumeActor lock TTL 30s→5m.
4. atenet router implicit-resume 15s→5m.
```

BenTheElder:
> It's mostly timeout bumps but there's a suggestion on making building Ext4 images a bit more robust.

ACK on ext4 more robust, I'm also working on moving the ext4 provisioning into Go (without exec) so we can drop the custom base image and make them more deterministic, but it's a complex change. I think it works fine now but I haven't battle tested it yet.

Our mock workloads have pretty small disks, will have to reconsider this aspect. Those timeouts seem abysmal!

I think I know how we want to fix this ... let me test some things.

hack/update/licenses.sh
```diff
@@ -74,5 +74,23 @@ for target in "${targets[@]}"; do
   rm -rf "${tmp_out}"
 done
 
+# Collect upstream licenses for forked / copied-in third-party SOURCE under in-tree
```
bowei, line 77: Can you pull this out into its own PR? Slim down the number of changes in your uVM patch set.

docs/roadmap.md
```diff
@@ -68,7 +68,7 @@ Below is a collection of finer-grained efforts which we believe align with the a
 * Audit logging on API and lifecycle operations.
 * Sandbox integrations for threat detection telemetry.
 * Harden actor networking to further isolate from the surrounding node (e.g. with current networking
-* Support for additional sandboxing technologies beyond gVisor, including at least one flavor of microVM.
+* Support for additional sandboxing technologies beyond gVisor. An initial microVM sandbox class (based on Cloud Hypervisor) is now available.
```
bowei: isn't this just [x] done?

hack/update/licenses.sh
BenTheElder: Sure, I've just been spending a lot of time rebasing the branch, and we didn't have a use case for this until this PR. I can split out more PRs.

hack/ateom-base/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM debian:stable-slim
+# e2fsprogs provides mkfs.ext4, which the ateom-microvm worker uses to build the
+# actor's writable /dev/vdb ext4 rootfs from the OCI bundle. debian-slim also
+# provides coreutils (cp --sparse) and glibc for the fetched cloud-hypervisor binary.
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends e2fsprogs \
+ && rm -rf /var/lib/apt/lists/*
```
bowei: Go the full way?

 && apt-get purge -y --auto-remove \
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*

hack/update/licenses.sh
BenTheElder: Done https://github.com/agent-substrate/substrate/pull/291

BenTheElder: I think [28722b4](https://github.com/agent-substrate/substrate/pull/287/commits/28722b45c53942ee145ddf35a37606afffe382a8) could also be split out.

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+	"golang.org/x/sys/unix"
+
+	"github.com/agent-substrate/substrate/internal/serverboot"
+)
+
+const (
+	hostVethName      = "ateom0"
+	actorVethName     = "eth0"
+	actorVethTempName = "ateom1"
+	hostVethCIDR      = "169.254.17.1/30"
```
bowei, line 58: do we want to pull these into a /internal so we don't have to repeat the constants?

BenTheElder:
[just trivial rebases for now]

cmd/atelet/main.go
```diff
@@ -326,16 +328,24 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
...
-		RunscPath:              runscPath,
+		RunscPath:              runscPathFor(assetPaths),
+		RuntimeAssetPaths:      assetPaths,
 		Spec:                   buildAteomWorkloadSpec(req.GetSpec()),
-	}); err != nil {
+	})
+	if err != nil {
 		return nil, fmt.Errorf("while calling ateom.CheckpointWorkload: %w", err)
 	}
+	sandboxRec.SnapshotFiles = resp.GetSnapshotFiles()
```
msau42, line 345: Are there more files stored in the checkpoint directory besides the snapshot files? Wondering if we could require that the entire directory only contain the necessary files.

BenTheElder: It should only have the snapshot files, but the specific list is currently the restore contract / recorded to the manifest. If we want to pivot to e.g. tarring them up, that should be fine.

go.mod
```diff
@@ -23,6 +24,7 @@ require (
 	github.com/hashicorp/go-reap v0.0.0-20260220095743-4e27870b4f51
 	github.com/klauspost/compress v1.18.5
 	github.com/opencontainers/runtime-spec v1.3.0
+	github.com/pelletier/go-toml/v2 v2.4.0
```
dims, line 27: looks like we need this for reading kata's configuration.toml

BenTheElder: yes. and ttrpc is for talking to kata-agent (should be in the commit message)

cmd/ateom-microvm/internal/third_party/kata/PROVENANCE.md
```diff
@@ -0,0 +1,52 @@
+# third_party/kata — vendored kata-containers sources
+
+Source copied from [kata-containers](https://github.com/kata-containers/kata-containers)
```
dims, line 3: Anything we can ask kata folks to do here to make it easier for us consume?

cmd/ateom-microvm/internal/third_party/kata/PROVENANCE.md
```diff
@@ -0,0 +1,52 @@
...
+name via `ttrpc.Client.Call(ctx, "grpc.AgentService", "<Method>", req, resp)`, so only
+the message types are needed.
+
+### RPCs actually used
+
+ateom drives a small subset of `AgentService`:
+`CreateSandbox`, `CreateContainer`, `StartContainer`, `UpdateInterface`, `UpdateRoutes`,
+`AddARPNeighbors`, `ReadStdout`, `ReadStderr`.
+
+### Regenerating
```
dims, line 46: if we touch this again, we'll need to add a make file or script!

cmd/atecontroller/internal/controllers/workerpool_apply.go
```diff
@@ -60,6 +60,7 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool) *appsv1ac.Deployment
 				WithType(corev1.HostPathDirectoryOrCreate)))
 
 	applyWorkerPoolPodTemplate(podSpecAC, containerAC, wp.Spec.Template)
+	applyMicroVMPodShape(podSpecAC, containerAC, wp)
```
dims: `applyMicroVMPodShape` just uses `wp.Spec.SandboxClass`, do we want to pass in just that? (similar to how we just pass in `wp.Spec.Template` only as the third param to `applyWorkerPoolPodTemplate` one line above

cmd/ateom-microvm/internal/third_party/kata/PROVENANCE.md
BenTheElder: ~agents :-)

cmd/ateom-microvm/internal/third_party/kata/PROVENANCE.md
BenTheElder: I suppose we could ask for it to be split to another module?

I'm not sure if this is even supported, the kata-agent is something of an implementation detail currently, we could pretty easily substitute our own. 

At the moment the most annoying part would be having another binary to plumb, especially without hosted builds.

BenTheElder: It seemed premature to ask for that now. We can consider it for a follow-up, but it's functionally pretty similar to vendoring. If we need to upgrade this often, that's a signal that we should move to a more stable integration point.

I was originally using the kata shim, but ... the gap to do snapshot/restore is pretty big.

pkg/api/v1alpha1/sandboxconfig_validation_test.go
```diff
@@ -103,9 +117,18 @@ func TestSandboxConfigValidation(t *testing.T) {
 		sc:      sandboxConfig("ok-gvisor", SandboxClassGvisor, map[string]map[string]AssetFile{"amd64": {"runsc": runscAsset()}, "arm64": {"runsc": runscAsset()}}),
 		wantErr: false,
 	}, {
-		name:    "microvm is unconstrained by the gvisor rule",
-		sc:      sandboxConfig("ok-microvm", "microvm", map[string]map[string]AssetFile{"amd64": {"cloud-hypervisor": runscAsset()}}),
+		name:    "valid microvm with full asset set",
+		sc:      sandboxConfig("ok-microvm", "microvm", map[string]map[string]AssetFile{"amd64": microVMAssets()}),
```
dims, line 121: guess we could do one more with `"arm64"`?

cmd/ateom-gvisor/main.go
```diff
@@ -273,9 +274,33 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 
 	s.cleanupActorNetworkOrExit(ctx, "Failed to clean up actor network after checkpoint")
 
+	// Report exactly the files runsc wrote so atelet ships precisely this set
+	// (checkpoint.img plus any pages images), rather than a hardcoded list.
+	snapshotFiles, err := listSnapshotFiles(checkpointPath)
```
dims, line 279: does this belong to another commit? (does not match the commit message)

hack/microvm-assets/stage-to-rustfs.sh
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Stage the assembled micro-VM asset set into the kind cluster's rustfs S3 bucket
+# under kata-assets/, where atelet fetches it (per demos/counter/counter-microvm.yaml.tmpl).
+# Run after the cluster is up (hack/install-ate-kind.sh) and assemble.sh has produced $OUT.
+#
+# Requires the `aws` CLI. Env: OUT (asset dir, default ./bin/microvm-assets/arm64),
```
dims, line 21: let's fail fast when `aws` is not present?

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -113,46 +116,144 @@ func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL
...
+//     to sign + set Content-Length (a non-seekable pipe hangs there), so we compress
+//     to a SEEKABLE temp file first.
 func sendToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, content io.Reader) (err error) {
 	bucket, object, err := parseGCSURL(gsURL)
 	if err != nil {
 		return fmt.Errorf("while parsing URL: %w", err)
 	}
+	tStart := time.Now()
+
+	if sp, ok := client.(streamingPutter); ok && sp.SupportsStreamingPut() {
```
dims: isn't the presence of the interface enough? :)

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -113,46 +116,144 @@ func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL
...
 		return fmt.Errorf("while creating temp compress file: %w", err)
 	}
 	defer os.Remove(tmpFile.Name())
 	defer tmpFile.Close()
 
-	zwc, err := zstd.NewWriter(tmpFile)
-	if err != nil {
-		return fmt.Errorf("while creating zstd writer: %w", err)
+	t0 := time.Now()
+	var logical, dataBytes int64
```
dims: why are we calling this `logical`? this is a count right?

also, when i saw `dataBytes`, it felt like that would be the actual bytes of data .. not a count?

cmd/ateom-microvm/internal/ch/api.go
```diff
@@ -0,0 +1,130 @@
...
+	"net/http"
+)
+
+// apiBase is a placeholder host; the real transport always dials the unix
+// api-socket, so the host portion of the URL is ignored.
+const apiBase = "http://localhost"
+
+// apiClient speaks the cloud-hypervisor REST API over its unix api-socket.
+//
+// cloud-hypervisor serves an HTTP/1.1 REST API on the api-socket, and we drive snapshot/restore
```
dims, line 33: Wish there was something we could reuse instead of rolling our own :(

cmd/ateom-microvm/internal/kata/agentclient.go
```diff
@@ -0,0 +1,273 @@
...
+			out.WriteString(line)
+		}
+		if err != nil {
+			break
+		}
+	}
+	return out.String()
+}
+
+// AgentClient is a thin ttrpc client for the kata-agent RPCs ateom drives
```
dims, line 87: Wish there was something we could reuse instead of rolling our own :(

internal/proto/ateompb/ateom.pb.go
```diff
@@ -42,8 +42,13 @@ type RunWorkloadRequest struct {
 	ActorId                string                 `protobuf:"bytes,3,opt,name=actor_id,json=actorId,proto3" json:"actor_id,omitempty"`
 	RunscPath              string                 `protobuf:"bytes,4,opt,name=runsc_path,json=runscPath,proto3" json:"runsc_path,omitempty"`
 	Spec                   *WorkloadSpec          `protobuf:"bytes,5,opt,name=spec,proto3" json:"spec,omitempty"`
-	unknownFields          protoimpl.UnknownFields
-	sizeCache              protoimpl.SizeCache
+	// runtime_asset_paths maps a runtime asset name (e.g. "kata-shim",
```
dims: should we limit the examples to what we validate elsewhere?

cloud-hypervisor, kata-kernel, kata-image, kata-config

cmd/atecontroller/internal/controllers/workerpool_apply.go
```diff
@@ -81,6 +82,53 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool) *appsv1ac.Deployment
...
+	podSpecAC *corev1ac.PodSpecApplyConfiguration,
+	containerAC *corev1ac.ContainerApplyConfiguration,
+	wp *atev1alpha1.WorkerPool,
+) {
+	if wp.Spec.SandboxClass != atev1alpha1.SandboxClassMicroVM {
+		return
+	}
+
+	// The micro-VM runtime needs /dev/kvm. The container is already privileged
+	// (so it can also reach vhost devices), but we mount /dev/kvm explicitly.
```
dims, line 104: you say `explicitly`... defensive move?

cmd/atelet/internal/ategcs/sparsezstd.go
```diff
@@ -0,0 +1,190 @@
...
+	exts := make([]extent, numExt)
+	for i := range exts {
+		if err := binary.Read(src, binary.LittleEndian, &exts[i].off); err != nil {
+			return 0, err
+		}
+		if err := binary.Read(src, binary.LittleEndian, &exts[i].length); err != nil {
+			return 0, err
+		}
+	}
+	if err := dst.Truncate(size); err != nil {
```
dims, line 163: `size` comes from the wire/disk? do we want to validate it?

cmd/ateom-microvm/internal/kata/specconv.go
```diff
@@ -0,0 +1,174 @@
...
+		out.Mounts = append(out.Mounts, &agentpb.Mount{
+			Destination: m.Destination,
+			Source:      m.Source,
+			Type:        m.Type,
+			Options:     m.Options,
+		})
+	}
+
+	if s.Linux != nil {
+		l := &agentpb.Linux{
```
dims, line 131: we don't want to forward the rest?

Seccomp, ApparmorProfile/SelinuxLabel, and Sysctl ...

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+	}
+	return nil
+}
+
+func installActorNftablesRules(podIP net.IP) error {
+	// Dedicated ateom-owned IPv4 table (cheap cleanup, no CNI chain mutation):
+	//   * postrouting: masquerade actor egress (169.254.17.2) behind the pod IP.
+	//   * prerouting: DNAT pod-IP:80/tcp to the actor veth IP.
+	//   * forward: accept forwarded packets between the actor veth and pod eth0.
+	// Mirrors cmd/ateom-gvisor (same compatibility-bridge caveats and TODOs).
```
dims, line 341: https://github.com/agent-substrate/substrate/blob/f5df01bd82620fb1c5c2a1d1884e21533ff6b184/cmd/ateom-gvisor/main.go#L600-L618

cmd/atecontroller/internal/controllers/workerpool_apply.go
```diff
@@ -81,6 +82,53 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool) *appsv1ac.Deployment
...
+// applyMicroVMPodShape adds the /dev/kvm device and node placement a micro-VM
+// (kata + cloud-hypervisor) worker pool needs, on top of any pod-template
+// settings. No-op for other sandbox classes.
+//
+// TODO: this hardcodes one sandbox class's pod requirements in the controller.
+// Consider making it generic so a sandbox class can declare its own pod shape
+// (e.g. required devices/mounts + node placement on the SandboxConfig spec)
+// instead of branching on SandboxClass here, so new classes don't need a
+// controller change.
+func applyMicroVMPodShape(
```
bowei: [minor] maybeApplyUVMPodShape

name to indicate that this does not do anything if it's not a uVM

cmd/atelet/internal/ategcs/objects.go
bowei: +1 to this -- do we need to both check for the interface AND have the method return a bool?

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -113,46 +116,144 @@ func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL
...
+//     to a SEEKABLE temp file first.
 func sendToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, content io.Reader) (err error) {
 	bucket, object, err := parseGCSURL(gsURL)
 	if err != nil {
 		return fmt.Errorf("while parsing URL: %w", err)
 	}
+	tStart := time.Now()
+
+	if sp, ok := client.(streamingPutter); ok && sp.SupportsStreamingPut() {
+		return sendToGCSStreaming(ctx, client, bucket, object, content, tStart)
```
bowei: it might be better for symmetry to have 

```
sendtoGCSWIthZStd
  -> sendStreaming
  -> sendNonStreaming
```

  just delegate to helper method in both paths instead of having a forking off just for the GCS case.

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -113,46 +116,144 @@ func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL
...
+// snapshots are unaffected.
+//
+// Upload strategy depends on the backend:
+//   - Streaming backends (GCS) accept a non-seekable body, so we pipe the compressor
+//     straight into PutObject: the compress overlaps the network PUT and we never
+//     stage the ~100MiB compressed payload to a temp file.
+//   - S3/rustfs PutObject hands the body to the AWS SDK, which needs a seekable body
+//     to sign + set Content-Length (a non-seekable pipe hangs there), so we compress
+//     to a SEEKABLE temp file first.
 func sendToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, content io.Reader) (err error) {
```
bowei: [naming] sendZStd

You are in a package that is ategcs already, so putting gcs in the name is kind of redundant...

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -113,46 +116,144 @@ func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL
...
 
-	// Seek back to the beginning of the temp file
-	if _, err := tmpFile.Seek(0, 0); err != nil {
-		return fmt.Errorf("while seeking temp file: %w", err)
+// sendToGCSStreaming compresses content and uploads it in one overlapped pass: a
+// goroutine writes the (sparse-extent or plain) zstd stream into an io.Pipe while
+// PutObject streams the read end to the object store. No seekable temp file, and
+// the compress runs concurrently with the network PUT. Used only for streaming
+// backends (GCS); see sendToGCSWithZstd.
+func sendToGCSStreaming(ctx context.Context, client ObjectStorage, bucket, object string, content io.Reader, tStart time.Time) error {
```
bowei: suggest: sendStreamingZtd

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -113,46 +116,144 @@ func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL
...
+		return fmt.Errorf("while seeking temp file: %w", err)
 	}
-	if err := zwc.Close(); err != nil {
-		return fmt.Errorf("while closing zstd writer: %w", err)
+	if err := client.PutObject(ctx, bucket, object, tmpFile); err != nil {
+		return fmt.Errorf("while putting object %q: %w", object, err)
 	}
+	slog.InfoContext(ctx, "Compressed zstd upload",
+		slog.String("object", object), slog.Bool("sparse", sparse),
+		slog.Int64("logical", logical), slog.Int64("data", dataBytes),
```
bowei: minor: this terminology is kind of confusing. Is this just:

size and compressed_size?

BenTheElder:
Thanks, will respond to the rest of these in the morning.

Also, aside from large images, https://github.com/agent-substrate/substrate/pull/295/changes#r3464778367 / https://github.com/agent-substrate/substrate/issues/119#issuecomment-4786279186 will be relevant for the final shape of things. I think I've been treating the [gvisor-ateom] full-rootfs-writes-in-memory approach as a bug (versus just process memory and reset disk to golden), but if we want that ... easy enough to swap out rootfs approaches.

The worst part is once we re-introduce virtiofsd, we will require a HEAD build for a while until we can get a packaged build of the upcoming release ... otherwise restores are broken (I landed a commit for this). Kata just cut their monthly release but virtiofsd is pending the next release. We can handle that in the asset staging script but it's clunky / slow.

I have some local branches with things like:
- multiple container support
- ext4 in go (no exec)
- virtiofsd rootfs + overlay

But I'm not sure yet if it makes sense to fold any of that back into this PR vs follow-ups given the size. Depending on the snapshot type discussion (and large actor image performance ...) it's probably at least worth swapping the rootfs approach.

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -113,46 +116,144 @@ func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL
 	return nil
 }
 
+// streamingPutter is an ObjectStorage whose PutObject accepts a non-seekable
+// streaming body without buffering (GCS). See gcsClient.SupportsStreamingPut.
+type streamingPutter interface{ SupportsStreamingPut() bool }
+
+// sendToGCSWithZstd zstd-compresses content and uploads it to gsURL.
```
bowei: potential simplification for both of the paths?

```
type writeContentResult struct {
  logical int64
  sparse bool
 ...
}

func writeContent(out io.Writer, content io.Reader) (writeContentResult, error) {
	if f, ok := content.(*os.File); ok {
		// Sparse-extent format: compress ONLY the populated extents (skip the holes).
    ...
		logical, dataBytes, err = writeSparseZstd(tmpFile, f)
		if err != nil {
			return fmt.Errorf("while sparse-compressing %q: %w", object, err)
		}
		return writeContentResult{
       sparse: true,
    }, nil
	}

  // Non-file reader (no holes to exploit): plain zstd stream.
  logical, err = plainZstd(tmpFile, content)
	if err != nil {
		return fmt.Errorf("while compressing data to temp file: %w", err)
	}
	...
}
```

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -202,20 +303,110 @@ func fetchFromGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL strin
 		}
 	}()
 
-	zrc, err := zstd.NewReader(rc, zstd.WithDecoderConcurrency(1))
+	// Peek the first bytes to pick the decode path: the sparse-extent format starts
```
bowei: recommend factoring the segments of code that only needs to know about io.Reader/io.Writer into separate funcs so they can be unit tested.

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -202,20 +303,110 @@ func fetchFromGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL strin
...
 
 	return nil
 }
 
+// copyZstdSparse copies src into dst skipping all-zero blocks, so dst becomes a
+// sparse file (the skipped regions are holes). Returns the logical size (total bytes
+// consumed from src) and the bytes actually written (non-zero). dst is truncated to
+// the logical size at the end so trailing zero regions become a hole and the file
+// size is exact. dst must be a fresh/truncated regular file opened for writing.
+func copyZstdSparse(dst *os.File, src io.Reader) (size int64, written int64, err error) {
```
bowei, line 391: You might want to be defensive and force an ftruncate(0) to guarantee that the file is empty. otherwise the naming is a bit of a misnomer this is less a "copy" than just writing out only the non-zero chunks.

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -202,20 +303,110 @@ func fetchFromGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL strin
...
+	// 64KiB blocks: a multiple of the 4KiB fs block (so skipped runs align to whole
```
bowei, line 397: This value might be kind of conservatively big vs the syscall cost.

cmd/atelet/internal/ategcs/sparsezstd.go
```diff
@@ -0,0 +1,190 @@
...
+// extent is a populated (non-hole) byte range of a sparse file.
+type extent struct {
+	off    int64
+	length int64
+}
+
+// sparseExtents returns the populated regions of f (via SEEK_DATA/SEEK_HOLE) and
+// f's logical size. Holes are the gaps between extents. It leaves f's offset
+// undefined (callers seek explicitly before reading).
+func sparseExtents(f *os.File) ([]extent, int64, error) {
```
bowei: alternative is to create an iterator over the sparse regions and have the zstd feed off of the iterator.

marginally more efficient -- if there are lots of tiny holes

cmd/atelet/internal/ategcs/sparsezstd.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		exts = append(exts, extent{off: ds, length: de - ds})
+		off = de
+	}
+	return exts, size, nil
+}
+
+// writeSparseZstd encodes a sparse file src to dst in the sparse-extent format:
+//
+//	magic[8] | version:u32 | totalSize:i64 | numExtents:i64 | (off:i64,len:i64)*numExtents | zstd(data...)
```
bowei: it's probably better to have the format be streamable with an end frame rather than write the numExtents up front.

cmd/atelet/internal/ategcs/sparsezstd.go
```diff
@@ -0,0 +1,190 @@
...
+// version-neutral; the format version follows it (see sparseVersion).
+const sparseMagic = "ATESPRSE"
+
+// sparseVersion is the sparse-extent format version, written as a little-endian
+// uint32 immediately after sparseMagic. Bump it on any incompatible layout change
+// so readers reject snapshots they don't understand instead of misparsing them.
+const sparseVersion uint32 = 1
+
+// extent is a populated (non-hole) byte range of a sparse file.
+type extent struct {
```
bowei: I think zstd is super efficient at zipping large extents of 0s... do we want to create our own format or just do the sparse file format while writing out to disk? You can still store the file compressed as a solid archive, because zstd take care of the zero blocks anyway.

cmd/atelet/internal/ategcs/sparsezstd.go
bowei: These functions should get a lot of unit testing.

bowei (commented):
Partial review, making my way through the commits.

cmd/ateom-microvm/internal/ch/ch.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Ping returns nil if the VMM api-socket answers vmm.ping.
+func (c *Client) Ping(ctx context.Context) error {
+	return c.api.get(ctx, "/api/v1/vmm.ping")
+}
+
+// WaitReady blocks until the api-socket answers vmm.ping or the deadline passes.
+func (c *Client) WaitReady(ctx context.Context, deadline time.Duration) error {
+	end := time.Now().Add(deadline)
```
bowei, line 53: why not use context deadline?

cmd/ateom-microvm/internal/ch/merge.go
```diff
@@ -0,0 +1,219 @@
...
+	if _, err := overlayDataRegions(d, m); err != nil {
+		return err
+	}
+	// No fsync: atelet reads the merged image back from page cache on this same node
+	// and ships it to GCS — that upload is the durability point, so an ~150MiB fsync
+	// on the suspend critical path buys nothing.
+	if err := m.Close(); err != nil {
+		return err
+	}
+	// Put the merged image at delta's name. We UNLINK CH's old delta FIRST, then
```
bowei: these comments are somewhat distracting detail-wise and could be summarized by something like "partially written back data will be discarded so there is no need to worry about sync'ing pages fully to disk".

Same with comment above.

cmd/ateom-microvm/internal/ch/merge.go
```diff
@@ -0,0 +1,219 @@
...
+	// A partial local file after a node crash is just discarded + the suspend retried,
+	// so paying an ~150MiB fsync on the suspend critical path buys nothing.
+	if err := o.Close(); err != nil {
+		return err
+	}
+	return os.Rename(tmp, out)
+}
+
+// MergeDeltaIntoBase produces the same COMPLETE merged snapshot as
+// MergeSparseOverlay(base, delta, delta) — base with delta's populated pages
```
bowei: I found this comment confusing -- especially since it describes what MergeSparseOverlay does. Can we have the comment just describe what MergeDeltaIntoBase does directly?

Also, some wierdness: MergeSparseOverlay(base, delta, ?delta?)

cmd/ateom-microvm/internal/ch/merge.go
```diff
@@ -0,0 +1,219 @@
...
+// CH we are about to tear down, and discarded afterward — so instead of copying it
+// we rename it next to delta, overlay delta's (small) faulted pages onto it, and
+// swap it into delta's place. That turns an O(working-set) copy into an O(delta)
+// write plus two metadata renames.
+//
+// base and delta are siblings under the actor dir (restore-state/ and
+// checkpoint-state/), so the renames are same-filesystem (metadata-only). On the
+// off chance they straddle a mount boundary (EXDEV), it falls back to the copying
+// MergeSparseOverlay (base is untouched until the first rename succeeds).
+func MergeDeltaIntoBase(ctx context.Context, base, delta string) error {
```
bowei: suggest: base-> baseFile, deltaFile

cmd/ateom-microvm/internal/ch/merge.go
```diff
@@ -0,0 +1,219 @@
...
+	// pages from the download, so a replace-rename costs ~0.5-0.8s (measured on GKE).
+	// Renaming to a name that does NOT exist skips that flush; the dirty pages stay in
+	// page cache for atelet to read + ship. This is what takes the merge ~840ms→~5ms.
+	if err := os.Remove(delta); err != nil {
+		return fmt.Errorf("remove old delta: %w", err)
+	}
+	return os.Rename(merged, delta)
+}
+
+// overlayDataRegions copies every populated (non-hole) region of src onto dst at
```
bowei: suggest just calling this copySparseRegions. There is no overlay being produced, you are overwriting the dst.

cmd/ateom-microvm/internal/ch/merge.go
```diff
@@ -0,0 +1,219 @@
...
+	for off < size {
+		// Next populated region [ds, de) in src.
+		ds, err := unix.Seek(sfd, off, unix.SEEK_DATA)
+		if err != nil {
+			if errors.Is(err, unix.ENXIO) {
+				break // no more data
+			}
+			return copied, fmt.Errorf("SEEK_DATA: %w", err)
+		}
+		de, err := unix.Seek(sfd, ds, unix.SEEK_HOLE)
```
bowei, line 183: minor: there is an iterator-based impl over sparse that probably would be easier to understand.

cmd/ateom-microvm/internal/ch/restorefds.go
```diff
@@ -0,0 +1,241 @@
...
+	if dl, ok := ctx.Deadline(); ok {
+		_ = conn.SetDeadline(dl)
+	} else {
+		_ = conn.SetDeadline(time.Now().Add(60 * time.Second))
+	}
+
+	// Raw HTTP/1.1 over the unix socket: net/http cannot attach SCM_RIGHTS, and
+	// CH's micro_http collects fds from the recvmsg ancillary data of the
+	// request that carries them.
+	req := fmt.Sprintf("PUT /api/v1/vm.restore HTTP/1.1\r\nHost: localhost\r\nAccept: */*\r\nContent-Type: application/json\r\nContent-Length: %d\r\n\r\n%s", len(body), body)
```
bowei, line 136: this is kind of gross -- can we not just use the standard http transport here?

cmd/ateom-microvm/internal/ch/restorefds.go
```diff
@@ -0,0 +1,241 @@
...
+	if err != nil {
+		return fmt.Errorf("dialing api-socket: %w", err)
+	}
+	defer conn.Close()
+	if dl, ok := ctx.Deadline(); ok {
+		_ = conn.SetDeadline(dl)
+	} else {
+		_ = conn.SetDeadline(time.Now().Add(30 * time.Second))
+	}
+	req := fmt.Sprintf("PUT /api/v1/vm.add-net HTTP/1.1\r\nHost: localhost\r\nAccept: */*\r\nContent-Type: application/json\r\nContent-Length: %d\r\n\r\n%s", len(body), body)
```
bowei, line 185: same here... is there a good reason why we have to do this completely raw?

cmd/ateom-microvm/internal/kata/agentclient.go
```diff
@@ -0,0 +1,273 @@
...
+	defer conn.Close()
+	_ = conn.SetDeadline(time.Now().Add(8 * time.Second))
+	if _, err := fmt.Fprintf(conn, "CONNECT %d\n", debugConsoleVsockPort); err != nil {
+		return "debug-console CONNECT: " + err.Error()
+	}
+	br := bufio.NewReader(conn)
+	if _, err := br.ReadString('\n'); err != nil { // the "OK <n>" line
+		return "debug-console CONNECT reply: " + err.Error()
+	}
+	// The kata debug console is an INTERACTIVE shell on a PTY (console.rs spawns
```
bowei, line 61: or you could look for the sentinel appearing exactly twice?

cmd/ateom-microvm/internal/kata/agentclient.go
```diff
@@ -0,0 +1,273 @@
...
+// hybrid-vsock unix socket — the same channel the kata shim would use.
+type AgentClient struct {
+	conn   net.Conn
+	client *ttrpc.Client
+}
+
+// DialAgent connects to the kata-agent through the hybrid-vsock socket at
+// vsockPath (VsockSocketPath(id)): plain-text "CONNECT <port>" handshake with
+// the VMM, then ttrpc over the stream.
+func DialAgent(ctx context.Context, vsockPath string) (*AgentClient, error) {
```
bowei, line 102: Some of this stuff is in agent/protocols/client/client.go in the Kata repo? Any reason we replicate it vs import from Kata?

cmd/ateom-microvm/internal/kata/cleanup_linux.go
```diff
@@ -0,0 +1,91 @@
...
+
+// CleanupSandboxState removes kata's leftover host-side state for a sandbox id
+// (the shared sandbox dir, the per-VM runtime dir, and the sbs dir), lazily
+// unmounting anything still mounted underneath them first, and kills orphaned
+// per-sandbox processes. Because ateom drives the shim directly (no
+// containerd), a failed Create does not fully self-clean; the deterministic
+// sandbox id (= actor id) then collides on the next attempt: "listen unix
+// .../virtiofsd.sock: bind: address already in use", "Could not bind mount
+// .../shared/sandboxes/<id>/mounts", "directory not empty". Calling this
+// before each run gives kata a clean slate. Safe when nothing exists.
```
bowei: [minor] delete "Safe when nothing exists" somewhat weird text

cmd/ateom-microvm/internal/kata/cleanup_linux.go
```diff
@@ -0,0 +1,91 @@
...
+		if perr != nil || pid == os.Getpid() {
+			continue
+		}
+		cmdline, rerr := os.ReadFile(filepath.Join("/proc", e.Name(), "cmdline"))
+		if rerr != nil || !strings.Contains(string(cmdline), id) {
+			continue
+		}
+		argv0 := strings.SplitN(string(cmdline), "\x00", 2)[0]
+		if strings.Contains(argv0, "cloud-hypervisor") || strings.Contains(argv0, "virtiofsd") || strings.Contains(argv0, "containerd-shim-kata") {
+			_ = unix.Kill(pid, unix.SIGKILL)
```
bowei: we should log on error instead of eating it

cmd/ateom-microvm/internal/kata/cleanup_linux.go
```diff
@@ -0,0 +1,91 @@
...
+				if mp == d || strings.HasPrefix(mp, d+"/") {
+					mounts = append(mounts, mp)
+					break
+				}
+			}
+		}
+		// Deepest paths first so child mounts unmount before their parents.
+		sort.Slice(mounts, func(i, j int) bool { return len(mounts[i]) > len(mounts[j]) })
+		for _, mp := range mounts {
+			_ = unix.Unmount(mp, unix.MNT_DETACH)
```
bowei: we should log on error, if we eat it and there is something wrong, then we would not know what happened?

cmd/ateom-microvm/internal/kata/cleanup_linux.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}
+	for _, d := range dirs {
+		_ = os.RemoveAll(d)
```
bowei: same here

cmd/ateom-microvm/internal/kata/agentclient.go
BenTheElder: I mean, with gRPC / proto it's pretty normal to generate your own, it's unusual how k8s uses proto :-)

internal/proto/ateompb/ateom.pb.go
BenTheElder: This is shared between the ateoms <> atelet? So that layer doesn't care which implementation. It just passes through what was recorded in the snapshot.

cmd/ateom-microvm/internal/kata/config.go
```diff
@@ -0,0 +1,103 @@
...
+			KernelParams  string `toml:"kernel_params"`
+		} `toml:"clh"`
+	} `toml:"hypervisor"`
+}
+
+// ParseConfig reads the guest sizing and kernel_params from a kata
+// configuration.toml. memDefault/vcpuDefault are substituted when the key is
+// absent or non-positive (kata also accepts default_vcpus = -1 meaning "all host
+// CPUs", which the owned boot does not support).
+func ParseConfig(base []byte, memDefault, vcpuDefault int) (KataConfig, error) {
```
bowei, line 56: *KataConfig? avoid copying by default

hack/ateom-base/Dockerfile
BenTheElder: It's probably negligible but sure.

Also I am going to get rid of this hopefully. 
Even if we still need ext4, I have some fun doing that in Go :-)

cmd/ateom-microvm/main.go
```diff
@@ -0,0 +1,232 @@
...
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"golang.org/x/sys/unix"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+var (
+	podUID = flag.String("pod-uid", "", "The UID of the current pod")
+
```
bowei: [minor] any reason to space this out like this?

cmd/ateom-microvm/main.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+}
+
+func do(ctx context.Context) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Share one synchronized writer between the runtime logger and the actor-log
+	// forwarder (created below) so the two log streams to the pod's stdout don't
+	// interleave-corrupt each other's lines (mirrors ateom-gvisor).
```
bowei: [minor] probably don't want to leave a bunch of "mirrors ateom-gvisor" in the code.

cmd/ateom-microvm/service_integration_test.go
```diff
@@ -0,0 +1,487 @@
...
+	}
+	got := kata.DebugConsoleDump(ctx, vsock, "cat "+rootfsDir+"/disk-sentinel 2>&1; echo END")
+	if strings.Contains(got, diskSentinel) {
+		t.Fatalf("reset-to-golden FAILED: disk sentinel persisted after restore#2: %q", got)
+	}
+	t.Logf("reset-to-golden OK: discarded the rootfs write (disk sentinel gone) while RAM continuity held: %q", strings.TrimSpace(got))
+}
+
+func lastLines(s string, n int) string {
+	lines := []string{}
```
bowei: var lines []string

cmd/ateom-microvm/service_integration_test.go
```diff
@@ -0,0 +1,487 @@
...
+		t.Fatalf("restore#2 RAM continuity failed: %q", got)
+	}
+	got := kata.DebugConsoleDump(ctx, vsock, "cat "+rootfsDir+"/disk-sentinel 2>&1; echo END")
+	if strings.Contains(got, diskSentinel) {
+		t.Fatalf("reset-to-golden FAILED: disk sentinel persisted after restore#2: %q", got)
+	}
+	t.Logf("reset-to-golden OK: discarded the rootfs write (disk sentinel gone) while RAM continuity held: %q", strings.TrimSpace(got))
+}
+
+func lastLines(s string, n int) string {
```
bowei, line 421: any reason why not

```
lines := strings.Split(s, "\n")
if len(lines) < n { return strings.Join(lines, "\n") }
etc
```

cmd/ateom-microvm/net.go
bowei: saw comment about later dedupe.

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	hostAddr, err := parseAddr(hostVethCIDR)
+	if err != nil {
+		return err
+	}
+	hostMAC, err := net.ParseMAC(hostVethMAC)
```
bowei: can we just do this in init()

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+			if err := netlink.LinkDel(l); err != nil {
+				slog.WarnContext(ctx, "Failed to delete leftover interior link", slog.String("link", l.Attrs().Name), slog.Any("err", err))
+			}
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	hostAddr, err := parseAddr(hostVethCIDR)
```
bowei: do this in init()?

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+	}
+	if err := netlink.LinkSetName(actorLink, actorVethName); err != nil {
+		return fmt.Errorf("while renaming actor veth to %q: %w", actorVethName, err)
+	}
+	actorLink, err = netlink.LinkByName(actorVethName)
+	if err != nil {
+		return fmt.Errorf("while reacquiring actor veth in interior netns: %w", err)
+	}
+
+	actorAddr, err := parseAddr(actorVethCIDR)
```
bowei: init()

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+		return err
+	}
+	if err := netlink.AddrReplace(actorLink, actorAddr); err != nil {
+		return fmt.Errorf("while assigning actor veth address: %w", err)
+	}
+	if err := netlink.LinkSetUp(actorLink); err != nil {
+		return fmt.Errorf("while bringing up actor veth: %w", err)
+	}
+
+	gw := net.ParseIP(actorVethGateway).To4()
```
bowei: init()

cmd/ateom-microvm/net.go
bowei, line 58: suggest throwing all these into a struct, parsing the values at init() and then passing the struct to the various funcs below already cooked.

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+	}
+
+	return nil
+}
+
+// cleanupActorNetwork removes all per-activation network state owned by ateom.
+// Intentionally idempotent: runs before setup, after checkpoint, and from
+// setup-failure cleanup.
+func (s *AteomService) cleanupActorNetwork(ctx context.Context) error {
+	if err := removeActorNftablesRules(); err != nil {
```
bowei, line 248: Should we gather errors but keep going if the intent is to run even on partial success?

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("while parsing address %q: %w", cidr, err)
+	}
+	return addr, nil
+}
+
+func enableIPv4Forwarding() error {
+	// Actor packets enter the worker pod via the host-side veth and leave
+	// through the pod's eth0; the kernel will not route between them otherwise.
+	if err := os.WriteFile("/proc/sys/net/ipv4/ip_forward", []byte("1\n"), 0o644); err != nil {
```
bowei: don't you want open, write instead of WriteFile, which also creates a file if it doesn't exist.

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+		return fmt.Errorf("while enabling IPv4 forwarding in worker pod netns: %w", err)
+	}
+	return nil
+}
+
+func installActorNftablesRules(podIP net.IP) error {
```
bowei, line 336: For future: put the nftables stuff into its own source file

cmd/ateom-microvm/net.go
```diff
@@ -0,0 +1,589 @@
...
+}
+
+// actorVethMTU reads the MTU of the actor veth (eth0 in the interior netns) so the
+// owned-boot path can configure the guest eth0 with a matching MTU via the agent
+// (UpdateInterface). Defaults to 1500 if the link can't be read.
+func (s *AteomService) actorVethMTU(ctx context.Context) int {
+	mtu := 1500
+	_ = netNSDo(ctx, s.interiorNetNS, func(ctx context.Context) error {
+		if l, err := netlink.LinkByName(actorVethName); err == nil {
+			mtu = l.Attrs().MTU
```
bowei, line 609: I would log an warning or info here

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	return os.IsNotExist(err)
+}
+
+// copyDiskFile copies a (sparse) disk image verbatim, preserving holes so the
+// (mostly-empty) ext4 image doesn't materialize its scratch blocks. Used to
+// save/restore the golden rootfs disk template.
+func copyDiskFile(ctx context.Context, src, dst string) error {
+	tmp := dst + ".tmp"
+	_ = os.Remove(tmp)
+	if out, err := exec.CommandContext(ctx, "cp", "--sparse=always", src, tmp).CombinedOutput(); err != nil {
```
bowei, line 118: don't we actually have go code that does this directly instead of shelling out to `cp`?

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	if err := os.Rename(tmp, dst); err != nil {
+		return fmt.Errorf("rename %s -> %s: %w", tmp, dst, err)
+	}
+	return nil
+}
+
+// resolvedRuntime holds the concrete binary/config paths for a request, taken
+// from fetched runtime assets when present, else the process flags.
+type resolvedRuntime struct {
+	ch         string
```
bowei: please document `ch` or expand the var name

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	configFile string
+}
+
+func firstNonEmpty(a, b string) string {
```
bowei: what if all empty?

also 

func firstNonEmpty(l ... string) string { ?

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	// Forward the actor container's stdout/stderr into the pod logs (parity with
+	// ateom-gvisor). StartBlkWorkload uses containerID==execID==id, so the agent
+	// keys the streams by id. The goroutines read over ac for the actor's lifetime
+	// and exit (io.EOF) when teardownActor closes ac.
+	s.startActorLogForwarding(ac, id, name, ns, containerName)
+
+	s.actorLogger.EmitLifecycleLog("Actor started", id, name, ns)
+	slog.InfoContext(ctx, "Actor started (owned-boot, virtio-blk rootfs)", slog.String("id", id))
+	return &ateompb.RunWorkloadResponse{}, nil
+}
```
bowei, line 421: may want to split this func up (it's 400 lines long)

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+
+// dialAgentRetry polls DialAgent until the kata-agent answers the hybrid-vsock
+// CONNECT (the socket file exists at boot, but the agent only listens once the
+// guest reaches kata-containers.target) or timeout elapses.
+func dialAgentRetry(ctx context.Context, vsockPath string, timeout time.Duration) (*kata.AgentClient, error) {
+	deadline := time.Now().Add(timeout)
+	var lastErr error
+	for {
+		dctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
bowei, line 451: the loop goes every 500 ms but the timeout is 5 s? Should timeout be < loop time?

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+			dev("c", 1, 9, "rwm"),    // /dev/urandom
+			dev("c", 5, 1, "rwm"),    // /dev/console
+			dev("c", 136, -1, "rwm"), // pts
+			dev("c", 5, 2, "rwm"),    // /dev/ptmx
+		},
+		CPU: &specs.LinuxCPU{Shares: &shares},
+	}
+}
+
+// CheckpointWorkload suspends the actor and writes a portable CH snapshot.
```
bowei: suggest breaking this file up `checkpoint.go`, `fs (mounts).go`, etc

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+//
+// Contract with atelet (mirrors ateom-gvisor): after we return, atelet uploads
+// the checkpoint dir to object storage, then tears down bundles and resets the
+// actor dir.
+//
+// ateom drives the ateom-owned CH's REST api-socket: pause -> snapshot
+// file://<CheckpointStateDir> (config.json + state.json + sparse memory-ranges) ->
+// tear the VMM down. The actor's rootfs writes are on the host-backed /dev/vdb
+// (NOT guest RAM), so the snapshot is naturally memory-only and small — no overlay
```
bowei: ? we have no page cache in the guest? what does this statement mean. I see this text repeated in multiple places in the code BTW, might be confusing if these details ever change...

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+// tmpfs upper to wipe and no balloon to inflate before snapshot.
+func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.CheckpointWorkloadRequest) (*ateompb.CheckpointWorkloadResponse, error) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	ns := req.GetActorTemplateNamespace()
+	name := req.GetActorTemplateName()
+	id := req.GetActorId()
+
+	s.actorLogger.EmitLifecycleLog("Actor checkpointing", id, name, ns)
```
bowei: [minor] argument order is a bit weird -- ns before name

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	tPause := time.Now()
+	if err := client.Pause(ctx); err != nil {
+		return nil, fmt.Errorf("while pausing guest: %w", err)
+	}
+	dPause := time.Since(tPause)
+
+	checkpointDir := ateompath.CheckpointStateDir(ns, name, id)
+	// Start from a clean dir so CH's snapshot files are the only contents.
+	if err := os.RemoveAll(checkpointDir); err != nil {
+		return nil, fmt.Errorf("while clearing checkpoint dir: %w", err)
```
bowei: add checkpoint dir path to the error?

fmt.Errorf("RemoveAll(%q): %w", checkpointDir, err)?

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	}
+	if err := os.MkdirAll(checkpointDir, 0o700); err != nil {
+		return nil, fmt.Errorf("while creating checkpoint dir: %w", err)
```
bowei: Add checkpointDir to err

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	// restore, losing the invariant golden id).
+	baseID := id
+	if ra != nil && ra.baseID != "" {
+		baseID = ra.baseID
+	}
+	if err := os.WriteFile(filepath.Join(checkpointDir, baseIDFile), []byte(baseID), 0o600); err != nil {
+		return nil, fmt.Errorf("while writing %s: %w", baseIDFile, err)
+	}
+
+	// NB: the snapshot is MEMORY-ONLY (config/state/memory-ranges + base-id). The
```
bowei: do we need this comment text repeated in multiple places?

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	// ateom (rootfs from the image at restore).
+
+	slog.InfoContext(ctx, "Snapshotting guest", slog.String("id", id), slog.String("dir", checkpointDir))
+	tSnapshot := time.Now()
+	if err := client.Snapshot(ctx, checkpointDir); err != nil {
+		return nil, fmt.Errorf("while snapshotting guest: %w", err)
+	}
+	dSnapshot := time.Since(tSnapshot)
+
+	// Diff-snapshot completion for an OnDemand-restored actor: CH's snapshot here is
```
bowei: same here -- are you reminding yourself of this?

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	// Tear down: the actor returns to "available". Best-effort; the snapshot is
+	// already on disk for atelet to ship.
+	tTeardown := time.Now()
+	s.teardownActor(ctx, id, ra, client)
+	dTeardown := time.Since(tTeardown)
+	delete(s.running, id)
+
+	// Tear down the per-activation actor network (mirrors gVisor).
+	if err := s.cleanupActorNetwork(ctx); err != nil {
+		slog.WarnContext(ctx, "Failed to clean up actor network after checkpoint", slog.Any("err", err))
```
bowei: I'm wondering if we keep going or just panic if the environment seems to have failed to be cleaned up

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	if err := os.MkdirAll(kata.VMDir(id), 0o700); err != nil {
+		return nil, fmt.Errorf("while creating VM dir: %w", err)
+	}
+
+	// Recreate the /dev/vdb backing file the snapshot references (the actor dir),
+	// reset-to-golden. Two ways, both byte-consistent with the golden snapshot's
+	// guest ext4 cache:
+	//   - same-node: a verbatim golden template (copyDiskFile) — guaranteed identical.
+	//   - cross-node: rebuild from the OCI image atelet unpacked to the bundle at
+	//     restore (mkfs.ext4 -d is LAYOUT-deterministic for identical inputs, so the
```
bowei: is this assumption always true for mkfs.ext4? probably ok in practice, although it would be kind of hard to diagnose if there were subtle shifts

cmd/ateom-microvm/run.go
```diff
@@ -0,0 +1,1059 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("while launching VMM for restore: %w", err)
+	}
+	defer func() {
+		if retErr != nil && chCmd.Process != nil {
+			_ = chCmd.Process.Kill()
+		}
+	}()
+	// OnDemand (userfaultfd) memory restore: ~75ms vs ~1.8s eager, and it keeps the
```
bowei: is this comment needed here?

bowei:
I took a pass through everything... 

Given the current MVP state where we still have to hash out a bunch of things, I didn't find anything show stopping in the review. I'm sure we need to morph quite a few things with the merging of the gvisor and CHV shared code. One thing that would be good to fix is to break up some of the really long func and files into more self contained (understandable) pieces that are unit tested.

cmd/ateom-microvm/main.go
BenTheElder: We can pull this out to high level docs. I want the behaviors to align so switching sandboxClass is ~easy.

cmd/ateom-microvm/service_integration_test.go
BenTheElder: Any particular reason to prefer that format?

bowei: mostly style -- it also doesn't alloc memory by default

docs/roadmap.md
BenTheElder: Yeah that seems fair. We technically have a follow-up milestone planned for making it robust but it's not dramatically less robust than the rest :^)

cmd/atelet/internal/ategcs/objects.go
BenTheElder: Agreed.

cmd/ateom-microvm/main.go
BenTheElder: no, some weird artifact, I AI split the commits (aside from generating much of the code). cleaning up along with most of these comments.

cmd/ateom-microvm/run.go
BenTheElder: Yeah, this is an explicit TODO. I have a stab at moving to go, but really we will just switch to virtiofsd + tmpfs overlay upper in the short term (perhaps before this merges even ...)

cmd/atecontroller/internal/controllers/workerpool_apply.go
BenTheElder: Yeah, I'd also like to trim down permissions in future PRs, but not here, I don't want the host cgroups for example.

cmd/ateom-microvm/run.go
BenTheElder: After discussion about https://github.com/agent-substrate/substrate/pull/295 it's pretty clear that we're not going to support this mode anyhow. Will rework.

cmd/ateom-microvm/internal/kata/agentclient.go
BenTheElder: Sure. Didn't seem super important. This approach works fine.

cmd/ateom-microvm/internal/kata/agentclient.go
BenTheElder: We could third_party it as well, the dependency tree is huge and we just want a small subset for the client. This file is basically all crud on top of the protos + TTRPC.

BenTheElder:
I'm working on addressing the comments tonight.

> I'm sure we need to morph quite a few things with the merging of the gvisor and CHV shared code

Yeah. We could do more code dedupe here.

I'm slightly concerned about lots of low level gvisor changes (e.g. networking stuff) landing into microvm in the short term, since we will have no CI coverage at least for now. But I suppose either way it has to be dealt with ASAP.

"Production Grade uVM" is a Beta target, FWIW. 

I already have a bunch of local changes waiting to rebase and stack on this.

Leaning towards folding in some of the smaller and more critical changes given the timing.

dims:
@BenTheElder yes let's land this and iterate!

cmd/ateom-microvm/internal/ch/restorefds.go
BenTheElder: I don't think so, because we need to do the SCM_RIGHTS stuff ...

cmd/ateom-microvm/run.go
BenTheElder: We want to do a `cp` locally but make sure it preserves sparseness. Go doesn't offer this, and our DIY format is for archival to object stores. We could DIY a loop with SEEK_DATA/SEEK_HOLE but I'd defer that to a follow-up and check the performance, these can be pretty big.

cmd/ateom-microvm/run.go
BenTheElder: 500ms is the delay before retry if we didn't succeed, 5s is the dial timeout per attempt in the derivative context. They're ~arbitrary but not really related.

cmd/ateom-microvm/run.go
BenTheElder: I think we want to best effort tear down everything. Maybe *after* attempting to do all cleanup it panics.

cmd/ateom-microvm/internal/kata/specconv.go
BenTheElder: TODO? I'm not sure what might break.

cmd/ateom-microvm/run.go
BenTheElder: Yeah it's weird but this is the existing API?

BenTheElder: The order here is correct vs the existing method, though I'd argue the method should swap these.

cmd/ateom-microvm/run.go
BenTheElder: just AI nonsense. will clean up.

cmd/ateom-microvm/run.go
BenTheElder: We need these details somewhere and we're setting the mode here, if there's temptation to switch to `"Copy"` mode it will be problematic for repeated roundtrips.

cmd/atelet/internal/ategcs/sparsezstd.go
BenTheElder: Yes, I thought about that, but it's still worth doing our own thing ... feeding zstd a ton of zeros at compress time is expensive even if they get compressed to not take much space. When the VM memory is over-provisioned we have 2GB memory with 150MB resident ...  this lets us just seek past the memory holes.

BenTheElder: Maybe we can get zstd to handle that without actually feeding it a ton of zeros? Leaving for a follow-up in any case. We're not stuck with this part.

cmd/ateom-microvm/internal/kata/config.go
BenTheElder: 2 ints and a string? probably not worth it?

cmd/atelet/internal/ategcs/objects.go
BenTheElder: This seems to be performing fine, but we can bench and optimize later?

cmd/atelet/internal/ategcs/objects.go
BenTheElder: ack

cmd/atelet/internal/ategcs/objects.go
BenTheElder: ack

cmd/atelet/internal/ategcs/objects.go
BenTheElder: ack

BenTheElder:
I'm pushing commits on top to make them more reviewable ... in theory. But I think it's probably going to be worth folding back into a logical stream again before merge ...

This is a lot for a single commit, so I don't think we should squash merge it, but I also don't necessarily think we want a ton of fixup commits at the end either.

cmd/atecontroller/internal/controllers/workerpool_apply.go
BenTheElder: done

cmd/atecontroller/internal/controllers/workerpool_apply.go
BenTheElder: done

cmd/atelet/internal/ategcs/sparsezstd.go
BenTheElder: added

cmd/ateom-microvm/internal/third_party/kata/PROVENANCE.md
bowei: Let's take a TODO -- agents sort of do it but it's hard to repeatedly verify exactly what was done to go from the Kata code to this codebase. It's easy to leave things out of a readme file because that is not executed deterministically.

cmd/atelet/internal/ategcs/objects.go
bowei: take a minor todo -- we may want to track how well fragmented the memory will get, but this is P2

cmd/atelet/internal/ategcs/sparsezstd.go
bowei: I actually think this is important wrt to the format, especially if the files get big, but I'm happy if we take a TODO to resolve in a future PR.

cmd/atelet/internal/ategcs/sparsezstd.go
bowei: Let's leave this alone for now. There is always storage v2.

cmd/ateom-microvm/internal/ch/restorefds.go
bowei: You might be able to hack it with a custom round tripper, but probably save that for later. For right now, can we make it a bit more structured:

func formatHTTPRequest(method, path string, headers map[string]string, body string) string

and 

func parseHTTPResponse(...) ...

so we can unit test this?

cmd/ateom-microvm/internal/kata/agentclient.go
bowei: It relies on bash string pasting semantics with the single quote, not the most obvious thing in the world.

cmd/ateom-microvm/internal/kata/agentclient.go
bowei: Would be good to document that very obviously...

BenTheElder:
https://github.com/agent-substrate/substrate/compare/main...BenTheElder:substrate:microvm-blk-rootfs-review-address-snapshot has the commits as reviewed + the review addressing commits ... I'll leave that as is for reference and then clean up the history here.

cmd/ateom-microvm/internal/kata/specconv.go
EItanya: Agreed, let's iterate

cmd/ateom-microvm/run.go
bowei: k, let's take a TODO on this. Prefer never shelling out

bowei (approved):

bowei:
mega pr for micro vm

BenTheElder:
... I meant to squash back in all the "Address review" comments before merge, got too tied up with oncall ... oh well.

cleaning up some follow-up in smaller PRs ...

PR #289, opened 2026-06-23
Update ValKey to v9.1; ValKey reload TLS certs every 12h

bowei:
Update to latest version of ValKey that supports cert reload from the conf file.

Without reload, ValKey would continue serving the old cert and cause the API server to stop working.

- [x] Tests pass

### DIFF
--- a/manifests/ate-install/valkey.yaml
+++ b/manifests/ate-install/valkey.yaml
@@ -30,6 +30,8 @@ data:
     tls-key-file /run/servicedns.podcert.ate.dev/credential-bundle.pem
     tls-ca-cert-file /etc/valkey-ca/ca.crt
     tls-auth-clients yes
+    # Reload the cert every 12h.
+    tls-auto-reload-interval 43200
 
     # Enable cluster mode
     cluster-enabled yes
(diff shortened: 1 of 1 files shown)


manifests/ate-install/valkey.yaml
```diff
@@ -87,7 +89,7 @@ spec:
     spec:
       containers:
       - name: valkey
-        image: valkey/valkey:8.0
+        image: valkey/valkey:9.1
```
BenTheElder: let's digest pin these while we're at it.

manifests/ate-install/valkey.yaml
```diff
@@ -30,6 +30,8 @@ data:
     tls-key-file /run/servicedns.podcert.ate.dev/credential-bundle.pem
     tls-ca-cert-file /etc/valkey-ca/ca.crt
     tls-auth-clients yes
+    # Reload the cert every 12h.
```
BenTheElder, line 33: how often are we refreshing the underlying cert currently?

bowei: never

manifests/ate-install/valkey.yaml
bowei: done

manifests/ate-install/valkey.yaml
BenTheElder: er, I meant in the cert controller

bowei: @zoez7

PR #293, opened 2026-06-23
Make SandboxConfig SHA256 optional to support moving runsc version.

### LINKED ISSUE #284: use gVisor golden weekly release
per https://github.com/agent-substrate/substrate/blob/bbafda0d3729465a2faedde2fcd83cceb4553ed0/manifests/ate-install/sandboxconfig-gvisor.yaml#L30 which is the default config for ateom gvisor

substrate now uses nightly release from a specific date of gvisor runsc binary.

using gvisor golden weekly release is a better way to consume upstreamed gvsior

milantracy:
The default gvisor SandboxConfig now points at the "latest" release URL (gs://gvisor/releases/release/latest/{arch}/runsc) instead of a pinned nightly build. Because the binary behind "latest" changes with each gVisor release, the SHA256 field can no longer be hardcoded in the manifest.

When SHA256 is omitted from a SandboxConfig asset, atelet downloads the binary and computes the hash on the fly. An in-memory URL->hash cache on AteomHerder prevents redundant downloads within the same atelet lifetime; restarting atelet picks up whatever "latest" currently resolves to. The computed hash is written back into the sandboxAssetsRecord before persisting, so checkpoint/restore manifests remain pinned to the exact binary that created the snapshot — the moving-target semantics apply only to new Run requests, never to restores.

Fixes #284

- [x] Tests pass

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 
 	"cloud.google.com/go/storage"
 	"github.com/agent-substrate/substrate/cmd/atelet/internal/ategcs"
@@ -187,6 +188,9 @@ type AteomHerder struct {
 	pullCache     *memorypullcache.MemoryPullCache
 	anonGCSClient ategcs.ObjectStorage
 	gcsClient     ategcs.ObjectStorage
+
+	urlHashMu    sync.Mutex
+	urlHashCache map[string]string
 }
 
 var _ ateletpb.AteomHerderServer = (*AteomHerder)(nil)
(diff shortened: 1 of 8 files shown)


BenTheElder:
I'm not sure we're OK with this. Imagine for example inexplicable CI failures due to a compatibility change.

Can we not just switch to a recent weekly release instead of moving to a floating version?

cc @ahmedtd

milantracy:
I can update the default template in a separate PR.

can you elaborate "Imagine for example inexplicable CI failures due to a compatibility change"

BenTheElder:
If we don't pin a version, then the runsc version changes without any commits, it won't be obvious when upgrading runsc has some regression. Test runs are not recording the version used.

Ditto users with local workloads

dberkov:
>If we don't pin a version, then the runsc version changes without any commits, it won't be obvious when upgrading runsc has some regression. Test runs are not recording the version used.

+1 to what @BenTheElder said, we cannot point to latest, since what if latest will be broken. We always supposed to pointed out to the well tested version.

PR #295, opened 2026-06-24
DurableDir support

dberkov:
Implement a layer allowing to take durableDir snapshot only, without taking snapshot of entire memory. This is a part of the   #119 feature.

The "durableDir" support  is introduced via adding Volumes support to ActorTemplate.

## Knowns issues / limitations

1. No MicroVM support for homedir yet. Will be implemented in following PR.

2. Resume from SNAPSHOT_SCOPE_DATA temporary behavior
Current behavior  of resume from durableStorage returns  immediately after gVisor process is started, it means container's bootstrap is not completed by the time resume API returned to the caller. It causes the issue that first curl request for suspended actor throws an HTTP exception, since webserver listening on the port inside container is not running yet.
PR #330 fixes the issue, for a meantime the workaround is to call ` kubectl ate resume actor` and after it call to the curl itself.

- [X ] Unit tests  pass
- [X ] E2E tests are updated and pass


internal/proto/ateletpb/atelet.proto
```diff
@@ -123,6 +148,15 @@ enum CheckpointType {
   CHECKPOINT_TYPE_EXTERNAL = 2;
 }
 
+enum SnapshotScope {
+  // Not valid option; should never happen.
+  SNAPSHOT_SCOPE_UNSPECIFIED = 0;
+  // Snapshot memory and the full rootfs (including homedir content).
```
BenTheElder: Do we actually want the full rootfs persistence, or just process memory?

This is something I've been toying with in #123 & various implemenations ...

AIUI we're implementing this at the cost of all writes having to live in memory on a tmpfs (which is bounded and easily exhausted, and can significantly bloat the snapshot if you only wanted to preserve process memory)

If we want full rootfs + process, I think we should probably come up with a clearer name for this mode, and keep "SNAPSHOT_SCOPE_PROCESS" for just the process memory later :thinking:

internal/ateompath/ateompath.go
```diff
@@ -23,6 +23,10 @@ const (
 	// The base path.  This is both the path of the root shared folder on the
 	// host filesystem, and when it is mounted into ateom and atelet containers.
 	BasePath = "/var/lib/ateom-gvisor"
+
+	// Homedir snapshots are temporarily stored in subfolder relative to process checkpoint path.
+	// This is because gVisor missing capability to separete homedir content from rest of rootfs upon checkpointing.
+	HomedirSnapshotsSubfoldderName = "homedir"
```
BenTheElder: Why does anything other than ateom-gvisor need to know this path? Shouldn't the result for atelet just be another snapshot artifact to upload? Can't ateom create the homedir internally without atelet?

https://github.com/agent-substrate/substrate/pull/287/commits/964d2eed8a5c05d92902d50a579030332730cd1e

BenTheElder:
> The "homedir" support is introduced via adding Volumes support to ActorTemplate.

I was expecting we might just have the actorTemplate specify what path is homedir and if specified we would internally provision the appropriate backing storage technology for the sandbox implementation and handle preserving it. We might not even need to mount a volume vs just tarring up that path from a shared filesystem after suspending the actor.

If we're supporting multiple homedirs per container, we might want to think about a more appropriate name?

internal/proto/ateletpb/atelet.proto
dberkov: this is not exactly "full" rootfs. It supposed to be a diff between base oci rootfs and changes on top of it.

I have replied on the #119 I did not find any good use-case for taking memory snapshot only. Lets discuss it on the #119 if you believe there is a use case for it.

dberkov: About the name, do you have better name in the mind other than PROCESS?

internal/ateompath/ateompath.go
dberkov: this supposed to be a temporary solution for gVisor. Today it can take either memory+rootfs or just homedir.

The gVisor team is working on a more generic snapshot file, that at time of memory+rootfs checkpoint, it will separate files to chunks, and homedir content will be stored separately from the rest of rootfs.  It will allow to take one "snapshot" operation and later to upload homedir to bucket without rest of rootfs. 

We actually need design how this feature will be supported in microVM.

dberkov:
@BenTheElder 

> was expecting we might just have the actorTemplate specify what path is homedir and if specified we would internally provision the appropriate backing storage technology for the sandbox implementation and handle preserving it. We might not even need to mount a volume vs just tarring up that path from a shared filesystem after suspending the actor.


1. In the future, we might want to support more than one dir
2. It is designed as a volume to support a sidecar feature and mount same volume to both containers under different path.

Ax team explicitly asked for both 1&2.


>If we're supporting multiple homedirs per container, we might want to think about a more appropriate name?

This is a good point. Do you have any other name in a ming you would like to suggest?

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -29,6 +29,48 @@ const (
...
+}
+
+// Represents the source of a volume to mount.
+// Exactly one of its members must be specified.
+//
+// When adding a new source type, list it in the ExactlyOneOf marker below.
+//
+// +kubebuilder:validation:ExactlyOneOf={homeDir}
+type VolumeSource struct {
+	// homeDir represents a directory on rootfs that will participate in snapshots.
```
msau42: It'll be good to explain in more detail the difference between rootfs and homedir.

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -125,15 +173,47 @@ type SecretKeySelector struct {
 	Optional *bool `json:"optional,omitempty"`
 }
 
+// SnapshotScope defines what components to include in a snapshot.
+// +kubebuilder:validation:Enum=process;homedir
+type SnapshotScope string
+
+const (
+	// Process memory plus the full rootfs (homedir included).
+	SnapshotScopeProcess SnapshotScope = "process"
```
msau42: Instead of the scope being inclusive of other scopes, I wonder if it would be better to make them mutually exclusive and require that the user specify multiple.

For example, in the future we probably also want to support external volume mount snapshots.

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -29,6 +29,48 @@ const (
...
+	// This must match the Name of a Volume.
+	//
+	// +required
+	// +kubebuilder:validation:MaxLength=63
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+	// Path within the actor at which the volume should be mounted. Must
+	// not contain ':'.
+	//
+	// +required
+	// +kubebuilder:validation:MaxLength=4096
```
thockin, line 75: Can you leave a TODO to validate this more carefully

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -29,6 +29,48 @@ const (
...
+	// volumeSource represents the location and type of the mounted volume.
+	VolumeSource `json:",inline" protobuf:"bytes,2,opt,name=volumeSource"`
+}
+
+// VolumeMount describes a mounting of a Volume within a actor.
+type VolumeMount struct {
+	// This must match the Name of a Volume.
+	//
+	// +required
+	// +kubebuilder:validation:MaxLength=63
```
thockin, line 67: // +kubebuilder:validation:XValidation:rule="!format.dns1123Label().validate(self).hasValue()",message="Name must be a valid DNS label"

?

pkg/api/v1alpha1/actortemplate_types.go
thockin: Agree.  We should add them to the glossary

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -185,6 +265,12 @@ type ActorTemplateSpec struct {
 	//
 	// +optional
 	WorkerSelector *metav1.LabelSelector `json:"workerSelector,omitempty"`
+
+	// Volumes defines the volumes to mount into all containers in the actor.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=32
+	Volumes []Volume `json:"volumes,omitempty"`
```
thockin, line 323: Since I can define multiple Volumes, I can define multiple homedirs.

a) Do we want to enable that (if so, we may want a better name)?
b) Does this mean "mount the same homedir at multiple places" or "mount multiple homedirs"?

The answers to this will form how I think about the API schema here.

pkg/api/v1alpha1/actortemplate_types.go
thockin: I agree that it MIGHT be useful to allow homedir+rootfs without process at some point in the future.  I don't know if rootfs without homedir is useful.  So the combinations are not arbitrary, nor are they successively inclusive.

We can either convert it to 3 bools in a struct or enumerate Everything vs HomedirOnly and later RootfsAndHomedir and maybe RootfsOnly.  Writing it out, I am now leaning back to 3 bools?

BenTheElder:
Let's add validation to reject homedir mode when `sandboxClass == microvm` until we've landed microvm _with_ homdir mode. The POC has the equivalent of process mode (will be updated to include rootfs before merge). 

I'm pretty sure I know how we'll do it but I'm already getting complaints about the size of the uVM POC which is not merged yet :-)

pkg/api/v1alpha1/actortemplate_types.go
dberkov: >For example, in the future we probably also want to support external volume mount snapshots.

All snapshots are supposed to be in sync; otherwise, it will be a mess, and some data will be out of sync with other data. 
From a layers point of view, the dependencies are supposed to be (from min to max):
* None
* External only
* Homedir(s)
* Rootfs (including homedir) (TBD???)
* Memory


----
* None - no snapshot. It means the actor always starts fresh. This is like deleting an actor and creating a new one. Therefore, "None" does not need to be included.
* External Storage - this becomes the minimal one and always needs to be taken. This is the difference between Pause and Commit. On Pause, we will most likely just detach it from the actor and attach it on resume. On Commit, we will need to detach it and take a snapshot of the latest state. I did not add it to the list since we don't have support for external storage yet.
* Homedir - these are local files. Taking a snapshot of a local file and not taking a snapshot of external storage logically does not make sense (see the statement above).
* Rootfs - this is an interesting case; we don't have a clear requirement for this capability yet. However, if we add it in the future, the homedir will be included in it in any case.
* Memory - this is a full snapshot of everything. Taking a memory snapshot with no rootfs or external storage does not work, since memory is cached based on either the rootfs, external storage, or both.

>I agree that it MIGHT be useful to allow homedir+rootfs without process at some point in the future.

See the comments above; we might add it in the future if it is required.

>We can either convert it to 3 bools in a struct or enumerate Everything vs HomedirOnly and later RootfsAndHomedir and maybe RootfsOnly. Writing it out, I am now leaning back to 3 bools?

> I wonder if it would be better to make them mutually exclusive and require that the user specify multiple.

Based on the explanation above, this does not look like a mutually exclusive pattern; it is supposed to be one of them.

pkg/api/v1alpha1/actortemplate_types.go
BenTheElder: I think a) yes and yes. but naming is hard and I'm not sure what to call it.
b) multiple homedirs, but LATER sharing that homedir across containers ? I think we could allow mounting to multiple paths in the same container but it's probably not very useful and may limit our implementation options.

pkg/api/v1alpha1/actortemplate_types.go
BenTheElder: I think it should stay an enum because bools make kinda bad APIs and we probably won't accept all combinations which just makes it an enum in need of good names?

We should support:
- RootFS + Process (Also "process" isn't quite right if we also snapshot the uVM kernel ... it's more like FS + Memory)
- FS
- Homedir

As a follow-up might also want to support some kind of "temp" volume in FS and FS+Memory modes that is explicitly not persisted. Maybe "emptyDir" again from k8s? (which can also explicitly be memory or disk backed?)

dberkov: >As a follow-up might also want to support some kind of "temp" volume in FS and FS+Memory modes that is explicitly not persisted. Maybe "emptyDir" again from k8s? (which can also explicitly be memory or disk backed?)

Taking memory snapshot with not "temp dir" is logically make snapshot in-consistent. We are coming to the same discussion again, what if memory has opened file description form "temp" folder. 

Having full file system snapshot minus "temp" dir - what is a use case for it? "Homedir" is exactly designed for those cases - snapshot will include folders with application data - everything else is disposed.

pkg/api/v1alpha1/actortemplate_types.go
dberkov: The feature is  intentionally designed in the following way:
* single actor might contain multiple "homedir" volumes.
* mount path for each volume can be different between sidecars.  
* same homedir volume can be shared between sidecars.
* more than one "homedir" volume can be mounted to a single container.
* mount same volume twice to same container under different mount path - I think it is overkill and we can add restriction validation that prevents it.


From all the comments above, it clearly can been seen - the "homedir" name is confusing and it needs to be renamed.

pkg/api/v1alpha1/actortemplate_types.go
dberkov: The validation existed on the ActorTemplate level and was restricted to homeDir only and this was a mistake, since the mountVolume validation needs to be applied for any kind of volumes.

Moved the validation it to the right location.

pkg/api/v1alpha1/actortemplate_types.go
dberkov: Done for both Volume and VolumeMount. I also a added unit tests for it.

dberkov:
>Let's add validation to reject homedir mode when sandboxClass == microvm until we've landed microvm with homdir mode. The POC has the equivalent of process mode (will be updated to include rootfs before merge).

I'm pretty sure I know how we'll do it but I'm already getting complaints about the size of the uVM POC which is not merged yet :-)

Added a "+kubebuilder:validation:XValidation" validation and adjusted unit tests.

internal/proto/ateletpb/atelet.proto
```diff
@@ -123,6 +148,15 @@ enum CheckpointType {
   CHECKPOINT_TYPE_EXTERNAL = 2;
 }
 
+enum SnapshotScope {
+  // Not valid option; should never happen.
+  SNAPSHOT_SCOPE_UNSPECIFIED = 0;
```
EItanya, line 168: Why even do this and not just have a default?

internal/proto/ateletpb/atelet.proto
EItanya: I tend to agree with @BenTheElder here. The full rootfs, even diffs, could pick up A LOT of things if the user is not careful. Egress controls might save them from downloading too many things I suppose, but definitely a potential footgun

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -125,15 +177,47 @@ type SecretKeySelector struct {
...
+	// OnCommit specifies what to include in the snapshot when a commit is requested.
+	// If not provided, the "process" behavior is used by default.
+	// The OnCommit must be a subset of the OnPause content.
+	//
+	// For example:
+	//   - if OnPause is "process", then OnCommit can be "process" or "homedir".
+	//   - if OnPause is "homedir", then OnCommit must be "homedir".
+	//
+	// +optional
+	OnCommit SnapshotScope `json:"onCommit,omitempty"`
```
EItanya, line 259: Pause vs commit are slightly confusing to me as verbs. I had to dig through the code to figure out the difference. I wonder if we should think through this for 2 reasons:
1. This is a very low level concept being surfaced in the API, node local snapshotting vs remote. I wonder how relevant this difference is for most users, and whether we should actually surface it somewhere else?
2. Pause is very similar to Suspend, and commit is very generic. I think we should try and come up with words that are a bit more clear to purpose.

EItanya (commented):
Just some high level comments, this is looking great

pkg/api/v1alpha1/actortemplate_types.go
BenTheElder: > Taking memory snapshot with not "temp dir" is logically make snapshot in-consistent. We are coming to the same discussion again, what if memory has opened file description form "temp" folder.

Yeah I don't think we should do only memory. I am suggested we offer modes for "memory + rootfs writes", "rootfs writes", and whatever we rename "homedir"

> Having full file system snapshot minus "temp" dir - what is a use case for it? "Homedir" is exactly designed for those cases - snapshot will include folders with application data - everything else is disposed.

Full rootfs means you don't have to know _where_ writes land up front. That might be convenient. It's functionally an overlay homedir on top of the container root (vs homedir may be a whiteout mount).

internal/proto/ateletpb/atelet.proto
dberkov: The default value  in PROTO is "0". We will not be able to change the default later. Therefore it is a safe pattern in proto make enum first value as 0 and later decide in the code, either do logic to accept 0 and make default or throw an error.  In this case, this is an internal API and defaulting will be done by API server, therefore "0" becomes non valid at all.

pkg/api/v1alpha1/actortemplate_types.go
dberkov: The pattern I imagine is PAUSE if you want to do a short suspension. The benefit of it, memory and rootfs can be preserved and resume will be fast. The "Commit" is somethings that been persistence and can be always back to it.  For those cases, process snapshot might be too expensive and it collect a lot of data that might be non relevant for execution. 

When we did PoC for mega-atelet, idea was in PAUSE not "kill process" at all, just keep it running and give to process 0 CPU. It allowed linux kernel to offload memory pages to SSD when it was no available.  The #119 has more info what is a difference between pause and suspend.

I kind of imagine, OnPause developers will use  "process" as a default behavior  and OncCommit is "homedir".

pkg/api/v1alpha1/actortemplate_types.go
thockin: Multi-response:

> None - no snapshot. It means the actor always starts fresh. This is like deleting an actor and creating a new one. Therefore, "None" does not need to be included.

I am not convince they are the same.  There's always going to be a small cost to "create + resume" vs. "resume".  Plus it shifts responsibility onto the consumer of the substrate when it doesn't need to.  I think "none" makes sense (we don't have to implement it now, but we should not rule it out).

>> For example, in the future we probably also want to support external volume mount snapshots.
>
> All snapshots are supposed to be in sync; 

Devil's advocate -- do we really need snapshots of external storage?  I do not have a good handle on the real requirements here.   

> Homedir - these are local files. Taking a snapshot of a local file and not taking a snapshot of external storage logically does not make sense (see the statement above).

Do we think that substrate-managed snapshots are the ONLY way to implement homedir?  Or could we. hypothetically, use some REALLY fast network-attached device?  Does that change anything here?

 > Rootfs - this is an interesting case; we don't have a clear requirement for this capability yet. However, if we add it in the future, the homedir will be included in it in any case.

In theory I could make an argument for rootfs without homedir.  In practice maybe not interesting.

> bools make kinda bad APIs

While I wrote that kube convention, there are places where bools can be good, and it's exactly this pattern (a set of N non-stacking, independent decisions.  But ultimately it's about ergonomics.

> As a follow-up might also want to support some kind of "temp" volume in FS and FS+Memory modes that is explicitly not persisted.

Can I say YAGN I for this?  I'll be content to be wrong, but let's not get too hypothetical.

As I wrote this out I had a thought.

Somewhere (ActorTemplate -> Actor) we have a "map" of which volumes we want mounted on which dirs.  If you don't want homedir semantics, don't mount a "homedir" volume.  If you don't want external volumes, don't mount external volumes.  We don't need MORE config for them  We only need 3 options:
1) Snapshot Nothing
2) Snapshot Everything (memory + rootfs + anything you have mounted)
3) Snapshot Volumes (anything you have mounted: homedir, external, maybe something else in the future)

pkg/api/v1alpha1/actortemplate_types.go
thockin: >> Since I can define multiple Volumes, I can define multiple homedirs.
>>
>> a) Do we want to enable that
>
> I think a) yes

Do we have a use-case for this?  Or do we at least have reasonable belief that allowing this doesn't cause some burden on implementations?

> b) multiple homedirs, but LATER sharing that homedir across containers

I am skeptical that disallowing multi-mount would be smart.  Like in kube, the abstraction is that a volume can be mounted at multiple places, across containers.  Deviating from that would be surprising.

> From all the comments above, it clearly can been seen - the "homedir" name is confusing 

Sure, I coined that when I assumed it would be singular and mounted at a fixed place :)

pkg/api/v1alpha1/actortemplate_types.go
dberkov: > ..If you don't want homedir semantics, don't mount a "homedir" volume. If you don't want external volumes, don't mount external volumes
* Snapshot Nothing
* Snapshot Everything (memory + rootfs + anything you have mounted)
* Snapshot Volumes (anything you have mounted: homedir, external, maybe something else in the future)


I actually like the framing (Snapshot Everything / Snapshot Volumes).  This is very accurate summary.


"Snapshot Nothing" - I still not convinced about it if we really need it. However, with the proposed concept above, we alway can add it later if we find a right use case for it.


>bools make kinda bad APIs....

The bool discussion is relevant either OnCommit/OnPause is a single value or multiple values. This kind of critical for design, since later on, API will be broken to make it array. 
However, if we agree on Concept of
* Snapshot Nothing (potential future extension)
* Snapshot Everything 
* Snapshot Volumes

the discussion becomes irrelevant, because Nothing+Everything or Everything+Volumes does not make sense.

pkg/api/v1alpha1/actortemplate_types.go
thockin: The difference between "pause" and "commit" (happy to rename) is meaningful.  Commit means "push my changes to perma-storage", which has a cost (possibly charged-back?).  Pause means "I'm idle now but I am not at a place where I want to push my changes yet".  A committed snapshot is eligible for a fork, but a pause is not.

We chose "commit" because it felt vaguely git-ish.  We chose "pause" because we kept mixing up "snapshot", "checkpoint", "suspend" in all our discussions :)  "Pause" feels more temporary, which is the intended semantic.

Again, happy to rename.

thockin: https://docs.google.com/drawings/d/13ZFdJhmOfX1RaNInjMfs6xjzmPQlmY5BiBEPgV0lFJY/edit?pli=1

internal/proto/ateletpb/atelet.proto
dberkov: I have updated the glossary file and renamed confusing names like HomeDir/Process. Please take a look into commits 0c74700b4a9d49fe73813c89e79563c347aaba3b and  88a7925eb3182bfc21b150e5f4593d54c9bfedde.

internal/ateompath/ateompath.go
dberkov: I have updated the glossary file and renamed confusing names like HomeDir/Process. Please take a look into commits 0c74700b4a9d49fe73813c89e79563c347aaba3b and  88a7925eb3182bfc21b150e5f4593d54c9bfedde.

pkg/api/v1alpha1/actortemplate_types.go
dberkov: I have updated the glossary file and renamed confusing names like HomeDir/Process. Please take a look into commits 0c74700b4a9d49fe73813c89e79563c347aaba3b and  88a7925eb3182bfc21b150e5f4593d54c9bfedde.

pkg/api/v1alpha1/actortemplate_types.go
dberkov: I have updated the glossary file and renamed confusing names like HomeDir/Process. Please take a look into commits 0c74700b4a9d49fe73813c89e79563c347aaba3b and  88a7925eb3182bfc21b150e5f4593d54c9bfedde.

pkg/api/v1alpha1/actortemplate_types.go
dberkov: I have updated the glossary file and renamed confusing names like HomeDir/Process. Please take a look into commits 0c74700b4a9d49fe73813c89e79563c347aaba3b and  88a7925eb3182bfc21b150e5f4593d54c9bfedde.

dberkov:
@BenTheElder I have rebased code. Following last  3 commits are changes in names & fixes of the code after rebase:
* 924b914c418cf1158c40a360a136af3d7e42b7ad
* a1cf7f83a5f7afe9a2c954edc31933f427b65462
* 411cddb33414927cef4f6d1e3c9c6afe44edcf29

PTAL

EItanya (approved):
DurDir :)

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -125,15 +179,50 @@ type SecretKeySelector struct {
...
 }
 
+// SnapshotScope defines what components to include in a snapshot.
+// +kubebuilder:validation:Enum=full;data
+type SnapshotScope string
+
+const (
+	// Full captures process memory plus the entire filesystem delta on top of
+	// the OCI image (including any attached DurableDir volumes).
+	SnapshotScopeFull SnapshotScope = "full"
```
thockin: Kube convention for string constants is "Full" and "Data" (leading capital letter)

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -125,15 +179,50 @@ type SecretKeySelector struct {
...
 	//
 	// +required
 	// +kubebuilder:validation:MinLength=1
 	Location string `json:"location"`
+
+	// OnPause specifies what to include in the snapshot when the actor is paused.
+	// If not provided, the "full" behavior is used by default.
+	//
+	// +optional
+	OnPause SnapshotScope `json:"onPause,omitempty"`
```
thockin, line 247: `+kubebuilder:validation:Enum=Full,Data`
`+kubebuilder:default=Data`

Or should the default be Full?

AFAICT there's not an easy way to do conditional defaulting here, so we can't easily set the default for OnCommit to "whatever OnPause is", unless we use an MAP for this.  How important is it?

What we generally want to avoid is having consumers of this API do things like "if obj.X == nil { /* assume "full" */ } -- that default is then baked into a million places and if you ever iterate on the API can break.

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -125,15 +179,50 @@ type SecretKeySelector struct {
...
+	// Full captures process memory plus the entire filesystem delta on top of
+	// the OCI image (including any attached DurableDir volumes).
+	SnapshotScopeFull SnapshotScope = "full"
+	// Data captures only the contents of attached volumes that support
+	// snapshots (currently DurableDir-typed volumes). Process memory and
+	// the rest of rootfs are excluded.
+	SnapshotScopeData SnapshotScope = "data"
+)
+
+// +kubebuilder:validation:XValidation:rule="(has(self.onPause) ? self.onPause : 'full') == 'full' || (has(self.onCommit) ? self.onCommit : 'full') == (has(self.onPause) ? self.onPause : 'full')",message="OnCommit must be a subset of OnPause"
```
thockin: It seems that defaults apply AFTER validations which is...unfortunate.  You are embedding the default value all over

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -125,15 +179,50 @@ type SecretKeySelector struct {
...
 	Location string `json:"location"`
+
+	// OnPause specifies what to include in the snapshot when the actor is paused.
+	// If not provided, the "full" behavior is used by default.
+	//
+	// +optional
+	OnPause SnapshotScope `json:"onPause,omitempty"`
+
+	// OnCommit specifies what to include in the snapshot when a commit is requested.
+	// If not provided, the "full" behavior is used by default.
```
thockin: We should be more precise with default behaviors.  If a user specifies `onPause: Data` then this can't default to `Full` unless you intend for that to be an error?

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -125,15 +179,50 @@ type SecretKeySelector struct {
...
+	//   - if OnPause is "full", then OnCommit can be "full" or "data".
+	//   - if OnPause is "data", then OnCommit must be "data".
+	//
+	// +optional
+	OnCommit SnapshotScope `json:"onCommit,omitempty"`
 }
 
 // ActorTemplateSpec defined desired spec of an actor.
+//
+// +kubebuilder:validation:XValidation:rule="!has(self.containers) || self.containers.all(c, !has(c.volumeMounts) || c.volumeMounts.filter(vm, has(self.volumes) && self.volumes.exists(v, v.name == vm.name && has(v.durableDir))).size() <= 1)",message="A container may mount at most one DurableDir-typed volume"
```
thockin, line 265: I thought you wanted to allow multiple DurDir?

pkg/api/v1alpha1/actortemplate_types.go
thockin: error strings should refer to fields in their jsonName style - `onCommit` and `onPause`

pkg/api/v1alpha1/actortemplate_validation_test.go
```diff
@@ -377,6 +377,376 @@ func TestActorTemplateValidation(t *testing.T) {
...
+		wantErr: true,
+		errMsg:  "MountPath must be a clean absolute Unix path",
+	}, {
+		name: "Volumes: DurableDir MountPath containing dotfile is valid (only bare '.' / '..' components are rejected)",
+		mutate: func(at *ActorTemplate) {
+			at.Spec.Volumes = []Volume{
+				{Name: "vol1", VolumeSource: VolumeSource{DurableDir: &DurableDirVolumeSource{}}},
+			}
+			at.Spec.Containers[0].VolumeMounts = []VolumeMount{
+				{Name: "vol1", MountPath: "/home/.config"},
```
thockin, line 776: We should also try something like `..config` or `x..y`

pkg/api/v1alpha1/actortemplate_validation_test.go
```diff
@@ -377,6 +377,376 @@ func TestActorTemplateValidation(t *testing.T) {
...
+		wantErr: true,
+		errMsg:  "MountPath must be a clean absolute Unix path",
+	}, {
+		name: "Volumes: DurableDir MountPath with control character is invalid",
+		mutate: func(at *ActorTemplate) {
+			at.Spec.Volumes = []Volume{
+				{Name: "vol1", VolumeSource: VolumeSource{DurableDir: &DurableDirVolumeSource{}}},
+			}
+			at.Spec.Containers[0].VolumeMounts = []VolumeMount{
+				{Name: "vol1", MountPath: "/home\t/user"},
```
thockin, line 832: Strictly speaking tabs, spaces, etc are valid in pathnames.  Is there a good reason to prevent them?

```
$ ls -l "a       b"
ls: cannot access 'a'$'\t''b': No such file or directory

$ touch "a       b"

$ ls -l "a       b"
-rw-r--r-- 1 thockin primarygroup 0 Jun 26 08:56 'a'$'\t''b'
```

thockin, line 832: How about a testcase for `"/my home directory"` (spaces) which should be allowed

pkg/api/v1alpha1/actortemplate_types.go
thockin: I will accumulate some smaller changes as I flag them for a followup

thockin (commented):
Overall LGTM  Some of my comments are minor

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -91,7 +92,9 @@ func (s *MarkPausingStep) Execute(ctx context.Context, input *PauseInput, state
 func (s *MarkPausingStep) RetryBackoff() *wait.Backoff { return nil }
 
 type CallAteletPauseStep struct {
-	dialer *AteletDialer
+	dialer      *AteletDialer
+	kubeClient  kubernetes.Interface
```
BenTheElder: 🔴 **Blocking:** these new fields are never populated. `workflow.go:227` constructs `CallAteletPauseStep{dialer: w.dialer}` only, so at runtime `s.kubeClient`/`s.secretCache` are nil. `Execute` then calls `workloadSpecFromActorTemplate` (line 120), and any ActorTemplate with a `secretKeyRef` env var fails with `FailedPrecondition: …Kubernetes client is unavailable` → the actor can no longer be paused (regression vs. the old inline spec-building, which only copied literal env).

The resume step (`workflow.go:169`) already wires these — mirror it:

```go
&CallAteletPauseStep{dialer: w.dialer, kubeClient: w.kubeClient, secretCache: w.secretCache},
```

Worth adding a pause/suspend unit test that exercises a `secretKeyRef` template so this can't silently regress (the counter e2e uses only literal env, so it won't catch it).

cmd/ateapi/internal/controlapi/workflow_suspend.go
```diff
@@ -93,7 +94,9 @@ func (s *MarkSuspendingStep) Execute(ctx context.Context, input *SuspendInput, s
 func (s *MarkSuspendingStep) RetryBackoff() *wait.Backoff { return nil }
 
 type CallAteletSuspendStep struct {
-	dialer *AteletDialer
+	dialer      *AteletDialer
+	kubeClient  kubernetes.Interface
```
BenTheElder: 🔴 **Blocking (same root cause as the pause step):** `workflow.go:198` builds `CallAteletSuspendStep{dialer: w.dialer}` without these fields, so suspend fails for any `secretKeyRef` template. Mirror the resume wiring:

```go
&CallAteletSuspendStep{dialer: w.dialer, kubeClient: w.kubeClient, secretCache: w.secretCache},
```

cmd/atelet/main.go
```diff
@@ -637,10 +649,36 @@ func (s *AteomHerder) prepareOCIBundles(
...
 	g.Go(func() error {
+		annotations := map[string]string{
+			"io.kubernetes.cri.container-type": "sandbox",
+			"io.kubernetes.cri.container-name": "pause",
+		}
+		// add annotation for every durable-dir volume
+		// TODO(dberkov) needs to revisit this logic once gVisor supports multiple durable-dir volumes.
+		for _, vol := range spec.GetVolumes() {
+			if vol.GetType() == ateletpb.VolumeType_VOLUME_TYPE_DURABLE_DIR {
+				annotations["dev.gvisor.spec.mount.durabledir.type"] = "bind"
```
BenTheElder, line 676: 🟡 The annotation key is hardcoded to `durabledir`, so with two DurableDir volumes only the **last** one's source survives — the others won't participate in `fscheckpoint`. The `TODO(dberkov)` acknowledges this, but the API validation only restricts **one DurableDir mount per container**, not one per template — so a valid template (vol1 in container A, vol2 in container B) reaches this broken path.

Either gate the TODO with validation (reject >1 DurableDir volume per template until supported), or key the annotation by volume name so it's correct now. (Related: the `data`-scope restore only passes `--fs-restore-image-path` to the pause container, and the e2e covers single-container only — multi-container `data` restore is untested.)

cmd/atelet/main.go
```diff
@@ -367,6 +368,16 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	return &ateletpb.CheckpointResponse{}, nil
 }
 
+func toAteomSnapshotScope(scope ateletpb.SnapshotScope) ateompb.SnapshotScope {
+	// assumption the request already been valdated and scope is in the valid values set
```
BenTheElder: nit: typo `valdated` → `validated`.

demos/counter/counter.go
```diff
@@ -27,13 +27,38 @@ import (
...
+	defer fileMutex.Unlock()
+	counter := 0
+	data, err := os.ReadFile(fileCounterPath)
+	if err == nil {
+		if i, err := strconv.Atoi(string(data)); err == nil {
+			counter = i
+		}
+	}
+	counter++
+	err = os.WriteFile(fileCounterPath, []byte(strconv.Itoa(counter)), 0644)
```
BenTheElder: nit: the repo uses `0o` octal literals elsewhere (`0o700`, `0o755`); prefer `0o644`. Demo-only.

internal/e2e/suites/demo/demo_test.go
```diff
@@ -72,7 +72,171 @@ func TestActorLifecycle(t *testing.T) {
...
+			}); err != nil {
+				t.Fatalf("failed to resume Actor again: %v", err)
+			}
+			waitForActorStatus(ctx, t, clients, actorID, ateapipb.Actor_STATUS_RUNNING)
+
+			resp, err = callActor(t, actorID)
+			if err != nil {
+				t.Fatalf("failed to call actor again: %v", err)
+			}
+			validateCounterResponse(t, resp, "after paise", tc.wantMemoryAfterPause, tc.wantFileAfterPause)
```
BenTheElder: nit: typo `after paise` → `after pause`. Also: the same `validateCounterResponse(..., "after pause", 2, 2)` is reused inside `suspendActor`, where the stage label should read "after suspend".

BenTheElder:
[Note: Review is an experiment, I spoke to Dima about it, not dropping 🤖 on random PRs]

cmd/ateapi/internal/controlapi/workflow_pause.go
dberkov: It supposed to be nil, since at suspend/pause we dont need to pre-load security variables, they are required for resume only. I have removed it from the CallAteletPauseStep/CallAteletSuspendStep.

cmd/ateapi/internal/controlapi/workflow_suspend.go
dberkov: It supposed to be nil, since at suspend/pause we dont need to pre-load security variables, they are required for resume only. I have removed it from the CallAteletPauseStep/CallAteletSuspendStep.

cmd/atelet/main.go
dberkov: I have added validation that at most one DurDir is allowed per actor, until gVisor and microVM will support more.

cmd/atelet/main.go
dberkov: fixed

demos/counter/counter.go
dberkov: fixed

internal/e2e/suites/demo/demo_test.go
dberkov: fixed

pkg/api/v1alpha1/actortemplate_types.go
dberkov: Fixed

pkg/api/v1alpha1/actortemplate_types.go
dberkov: I made default Full for both and created #339 to decide the new defaults.

dberkov: >+kubebuilder:validation:Enum=Full,Data

```
// +kubebuilder:validation:Enum=Full;Data
type SnapshotScope string
```

already have it.  The `SnapshotsConfig: OnPause invalid enum value` test validates it

pkg/api/v1alpha1/actortemplate_types.go
dberkov: The `"SnapshotsConfig: OnPause unset (defaults to Full), OnCommit=Data",` test validates it and ensured that this is invalid

pkg/api/v1alpha1/actortemplate_types.go
dberkov: with " +kubebuilder:default=Full" it is enforced now.

pkg/api/v1alpha1/actortemplate_types.go
dberkov: Yes, we want it, unfortunately  gVisor does not support it today. The gVisor team works on it. So we will remove the restriction once they will add support for it.

pkg/api/v1alpha1/actortemplate_types.go
dberkov: fixed

pkg/api/v1alpha1/actortemplate_validation_test.go
dberkov: I have added unit tests for it

pkg/api/v1alpha1/actortemplate_validation_test.go
dberkov: I have added unit test for it

dberkov: new test added

dberkov:
I have addressed all the comments

BenTheElder (dismissed):
LGTM

We can extend to microvm in a follow-up, and further bikeshed the names etc. Let's merge and iterate this one.

Issue #265, opened 2026-06-17
Support actors listening on multiple ports


juli4n:
Today we only support actors listening on port 80. This is hard-coded in atenet (https://github.com/agent-substrate/substrate/blob/00a90a8eda76800c9af107398c5f6b65616e7e5c/cmd/atenet/internal/router/extproc.go#L162). 

This task is to allow actor containers to listen on multiple ports. 

I guess one open question is how the router deals with? Will it add listeners dynamically as actor templates are created? Will the DNS names be extended to include the port in the name (some sandboxing tools do this).

BenTheElder:
See also: https://github.com/agent-substrate/substrate/issues/240 (it's also hardcoded in ateom at the moment, but I think we can/should lift that part)

rogeroger-yu:
One concrete design that fits this issue is to encode the target actor port in the left-most DNS label, while preserving the existing host format as the default-port path.

Proposed routing format:

```text
<actor-id>.actors.resources.substrate.ate.dev            -> <worker-ip>:80
<port>-<actor-id>.actors.resources.substrate.ate.dev     -> <worker-ip>:<port>
```

Examples:

```text
my-actor.actors.resources.substrate.ate.dev        -> my-actor on port 80
8080-my-actor.actors.resources.substrate.ate.dev   -> my-actor on port 8080
49983-sbx1.actors.resources.substrate.ate.dev      -> sbx1 on port 49983
```

Why this shape:

- It does not require dynamic listener creation in Envoy or `atenet`.
- It should work with the existing wildcard DNS shape because the host still ends in `actors.resources.substrate.ate.dev`.
- It is backwards compatible: hosts without a numeric prefix keep routing to port 80.
- It matches the convention used by several sandbox systems where a sandbox service port is encoded in the hostname.
- It enables generic multi-service actors, not just one specific SDK or compatibility layer.

Implementation sketch:

1. Extend the actor-host parsing helper in `cmd/atenet/internal/router/extproc_in.go` so it returns both `actorID` and `port`.
2. If the left-most label matches `^(\d+)-(.+)$`, parse the leading digits as the port and the remainder as the actor ID.
3. If there is no numeric prefix, keep the existing behavior and default the port to `80`.
4. Validate the parsed port is in `1..65535`; invalid ports should fail routing rather than falling back silently.
[...]

PR #296, opened 2026-06-24
Route actor hosts by optional "<port>-" prefix

rogeroger-yu:
## What

Encode an optional target service port in the left-most DNS label of an actor host, so a single actor can expose multiple in-actor services without dynamic Envoy listeners or per-actor reverse proxies.

```
<actor-id>.<suffix>            -> <worker-ip>:80   (unchanged default)
<port>-<actor-id>.<suffix>     -> <worker-ip>:<port>
```

Examples:

```
my-actor.actors.resources.substrate.ate.dev        -> my-actor on port 80
8080-my-actor.actors.resources.substrate.ate.dev   -> my-actor on port 8080
49983-sbx1.actors.resources.substrate.ate.dev      -> sbx1 on port 49983
```

This is the design proposed in #265.

## Why this shape

- No dynamic listener creation in Envoy or `atenet`.
- Works with the existing wildcard DNS shape: the host still ends in `actors.resources.substrate.ate.dev`.
- Backwards compatible: hosts without a numeric prefix keep routing to port 80.
- Enables generic multi-service actors (command/filesystem API, code execution, VNC, user HTTP servers, debugging endpoints, ...) rather than one specific SDK/compat layer.

## Changes

Two layers had to change for this to work end to end:

1. **`atenet` routing** (`cmd/atenet/internal/router`): `parseActorID` now returns `(actorID, port)`. A left-most label matching `^([0-9]+)-(.+)$` is parsed as `<port>-<actorID>`; the port is validated to be in `1..65535` and otherwise fails routing with a 404 rather than silently falling back. `handleRequestHeaders` forwards to `<workerIP>:<port>` instead of a hard-coded `:80`.

2. **`ateom-gvisor` actor DNAT** (`cmd/ateom-gvisor/main.go`): the actor prerouting DNAT previously matched only TCP/80 and rewrote both address and port to the actor veth on 80, which made port-prefixed routing a no-op end to end. It now matches all inbound TCP to the worker pod IP and DNATs only the destination address to the actor veth, preserving the original port. (`ateom` serves its own RPCs over a unix socket, not a pod-IP TCP port, so this catch-all does not shadow the control path.)

## Ambiguity

Actor IDs that themselves start with `<digits>-` are interpreted as port-prefixed. DNS-1123 permits such labels, so this is a documented, deterministic rule: tools that mint actor IDs for multi-port access must avoid IDs beginning with `<digits>-`. IDs that start with digits but have no `<digits>-` prefix (e.g. `123abc`) keep routing to the default port.

## Tests

- Unit tests in `extproc_in_test.go` cover default-port routing, prefixed-port routing, min/max ports, port 0, out-of-range ports, leading-zero normalization, and the existing host/connection-port parsing.
- Manually verified end to end on a local kind cluster with a counter actor listening on both 80 and 8080:
  - `my-actor.<suffix>` -> 200, served from port 80
  - `8080-my-actor.<suffix>` -> 200, served from port 8080 (distinct response body)
  - `0-my-actor.<suffix>` / `99999-my-actor.<suffix>` -> 404 (invalid port)
  - `atenet` logs confirm `targetAddr` of `<ip>:8080` vs `<ip>:80`.

## Notes

- UDP DNAT for actors exposing QUIC/UDP is left as a follow-up (existing TODO retained).

### DIFF
--- a/cmd/atenet/internal/router/extproc.go
+++ b/cmd/atenet/internal/router/extproc.go
@@ -129,7 +129,7 @@ func (s *ExtProcServer) handleRequestHeaders(
 	metadata := newRequestMetadata(reqHeaders.Headers.GetHeaders())
 	slog.InfoContext(ctx, "Request", slog.String("metadata", metadata.String()))
 
-	actorID, err := parseActorID(metadata.host)
+	actorID, port, err := parseActorID(metadata.host)
 	if err != nil {
 		// Host is invalid, respond with 404.
 		return nil, metadata, "", "", "", invalidHostErr(metadata.host, err)
@@ -158,8 +158,7 @@ func (s *ExtProcServer) handleRequestHeaders(
 			"actor %q routing failed", actorID)
 	}
 
-	// TODO(bowei) -- handle more than port 80 on the actor.
-	targetAddr := net.JoinHostPort(workerIP, "80")
+	targetAddr := net.JoinHostPort(workerIP, port)
 
 	slog.InfoContext(ctx, "Route ok", slog.String("actorID", actorID), slog.String("targetAddr", targetAddr))
 
--- a/cmd/atenet/internal/router/extproc_in.go
+++ b/cmd/atenet/internal/router/extproc_in.go
@@ -17,6 +17,8 @@ package router
 import (
 	"fmt"
 	"net"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/agent-substrate/substrate/internal/resources"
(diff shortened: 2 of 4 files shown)


bowei:
Would be good to resolve the conversations on the linked issue before we get too far along in the implementation.

bowei:
Hi -- any updates? Looks like there are also some merge conflicts.

LiorLieberman:
>Hi -- any updates? Looks like there are also some merge conflicts.


Given the feedback we have on https://github.com/agent-substrate/substrate/issues/265, https://github.com/agent-substrate/substrate/issues/445, https://github.com/agent-substrate/substrate/issues/484 -- I think that this solution would likely not be ideal for now. 

The approach here would collide with currently valid actor names, and if we split it with a dot it will have challenges with TLS.

I would love us to coordinate this multi port support soultion with the the non-http protocols solutions.

We should unify all the discussion into one, I am looking into doing this

EItanya:
> > Hi -- any updates? Looks like there are also some merge conflicts.
> 
> Given the feedback we have on #265, #445, #484 -- I think that this solution would likely not be ideal for now.
> 
> The approach here would collide with currently valid actor names, and if we split it with a dot it will have challenges with TLS.
> 
> I would love us to coordinate this multi port support soultion with the the non-http protocols solutions.
> 
> We should unify all the discussion into one, I am looking into doing this

100% agree, I'm not in favor of this approach

bowei:
Hi, given the CONNECT protocol for accessing additional ports, I am going to close this PR for the time being. If you are going to pick this back up, please reopen.

PR #298, opened 2026-06-24
Prefer workers with cached actor images during scheduling

### LINKED ISSUE #276: Prefer workers with cached actor images during scheduling
## Description

Atelet maintains a node-local image cache, but the scheduler does not know which images are cached on each node. As a result, an actor may be assigned to a cold node and pull its images again, even when another eligible node already has them cached.

Image preparation happens synchronously during `ResumeActor`, so unnecessary pulls increase actor startup latency and registry traffic.

## Proposal

- Let atelet report its cached image digests to the control plane.
- Store cache information by node.
- After applying existing constraints, prefer available workers whose node has all required actor images cached.
- Fall back to the current random selection when no cache hit is available.

Local snapshot availability remains a hard constraint. Image cache affinity is applied only among nodes that can restore the selected snapshot.

A general scoring framework is not required: use cache-hit preference with fallback.

## Expected Outcome

Reduce unnecessary image pulls and improve Actor Resume latency and consistency.

## Related Issues

- #166 and #228 optimize node-local image/rootfs reuse.
- #135 discusses locality-aware scheduling for cached snapshots.
- #233 tracks cold image pulls exceeding the `ResumeActor` deadline.

This proposal complements those efforts by making worker selection aware of node-local image cache availability.

han2ni3bal-pixel:
Fixes #276

## Changes

- atelet periodically reports image digests held in its in-memory cache to ate-api
- ate-api stores cache information in Valkey, keyed by node
- During ResumeActor worker selection, prefer nodes where all required images are already cached; fall back to the existing random strategy when there is no cache hit
- Local snapshot node restrictions remain a hard constraint; image cache affinity is applied only after that

## Testing

- [x] Unit tests
- [x] Manual verification on Multipass + k3s: resume counter actor; Cache miss/hit behaves as expected; Redis contains `imageDigests`


juli4n:
There have been discussions around soft-constraints for scheduling based on node properties (like this), this not only applies to OCI images but other artifacts like volume snapshots. I would recommend we discuss this in #276 and agree on a path forward before turning it into an implementation.

FYI @msau42

PR #299, opened 2026-06-24
Add an in-memory workers cache.

juli4n:
Remove the need for scanning all workers from ValKey during `ResumeActor`. Extend store.Interface with a "watcher" interface so that clients can keep track of workers in memory and do cache invalidation.

For ValKey, the watcher is implemented by publishing events to a ValKey channel. Clients subscribe to changes using SUBSCRIBE and receive them via the WorkerWatch channel.

### DIFF
--- a/cmd/ateapi/internal/store/ateredis/ateredis.go
+++ b/cmd/ateapi/internal/store/ateredis/ateredis.go
@@ -47,6 +47,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"sort"
 	"strings"
 	"sync"
@@ -59,10 +60,16 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+type workerPubSubMsg struct {
+	Type   int    `json:"t"`
+	Worker string `json:"w"` // protojson-encoded Worker
+}
+
 type redisClient interface {
 	redis.Cmdable
 	ForEachMaster(ctx context.Context, fn func(ctx context.Context, client *redis.Client) error) error
 	Watch(ctx context.Context, fn func(*redis.Tx) error, keys ...string) error
+	Subscribe(ctx context.Context, channels ...string) *redis.PubSub
 }
 
 // Persistence is a service that stores information about applications in Redis.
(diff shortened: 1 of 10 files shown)


thockin:
This LGTM - who else wants to review this?

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -87,6 +94,77 @@ func workerDBKey(namespace, poolName, podName string) string {
...
+const workerPubSubChannel = "worker-changes"
+
+func (s *Persistence) publishWorkerEvent(ctx context.Context, eventType store.WorkerEventType, worker *ateapipb.Worker) {
+	payload, err := marshalWorkerEvent(eventType, worker)
+	if err != nil {
+		slog.ErrorContext(ctx, "worker event marshal failed", slog.Any("err", err))
+		return
+	}
+	if err := s.rdb.Publish(ctx, workerPubSubChannel, payload).Err(); err != nil {
+		slog.ErrorContext(ctx, "worker event publish failed", slog.Any("err", err))
```
EItanya, line 130: Do we need to do anything else here, this could lead to other watchers getting into a bad state

cmd/ateapi/internal/store/store.go
```diff
@@ -84,3 +91,39 @@ type Interface interface {
 	// DebugClearAll drop all data from the database. Useful for debugging / local testing/
 	DebugClearAll(ctx context.Context) error
 }
+
+// WorkerEventType indicates the type of change to a Worker.
```
EItanya, line 95: nit: do we want this API to be proto so we can compress and marshal/unmarshal faster?

EItanya (commented):
This code makes sense to me overall. One thing I'll note is that these redis watches can have dragons hiding in them, and we need to be extra careful with recovering for dropped events, and always ensuring that this channel is best effort.

@thockin definitely not a blocking review, but some notes I think are worth taking a look at.

cmd/ateapi/internal/store/ateredis/ateredis.go
thockin: Good question - from an error-handling POV we probably should make a big deal when this happens - initially so we can figure out how to handle common things (if those exist) and later to know "OMG something bad happened".

I don't know what errors might actually crop up here.

cmd/ateapi/internal/store/store.go
thockin: IIUC this is only the in-process API (client), so it's not something that will be marshalled ?  Did I miss a jump somewhere?

cmd/ateapi/internal/store/ateredis/ateredis.go
thockin: Looking at this more, should it retry some errors internally or propagate those up to the operation that triggered this?

cmd/ateapi/internal/store/store.go
juli4n: The `WorkerEvent` is the payload of the pubsub message. 

> do we want this API to be proto so we can compress and marshal/unmarshal faster?

I wouldn't optimize early. Sounds like an easy change if we find it too slow.

juli4n:
> This code makes sense to me overall. One thing I'll note is that these redis watches can have dragons hiding in them, and we need to be extra careful with recovering for dropped events, and always ensuring that this channel is best effort.
> 

Agree. I think we eventually will want to add a resync period that does a re-list to recover from drifts. I thought about adding that, but would rather start small and learn more about the failure modes.

> Tim Hockin (@thockin) definitely not a blocking review, but some notes I think are worth taking a look at.

cmd/ateapi/internal/store/ateredis/ateredis.go
juli4n: To me the pubsub system is fire and forget. If the system degrades in some way and this fails, it will result in stale entries in other replicas. This is not a correctness issue, because we do optimistic locking during scheduling, it will just make those replicas slower to schedule. 

I think eventually we will want to add a ticker that does a re-list, similar to k8s reflectors.

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -87,6 +94,77 @@ func workerDBKey(namespace, poolName, podName string) string {
...
+	}
+	return store.WorkerEvent{Type: store.WorkerEventType(msg.Type), Worker: worker}, nil
+}
+
+const workerPubSubChannel = "worker-changes"
+
+func (s *Persistence) publishWorkerEvent(ctx context.Context, eventType store.WorkerEventType, worker *ateapipb.Worker) {
+	payload, err := marshalWorkerEvent(eventType, worker)
+	if err != nil {
+		slog.ErrorContext(ctx, "worker event marshal failed", slog.Any("err", err))
```
thockin, line 126: We should add a metric which counts the number of times that a "publish" operation has failed.

I will accept a TODO

cmd/ateapi/internal/workercache/workercache.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// Cache maintains an in-memory snapshot of all workers.
+//
+// TODO: add metrics — at minimum a gauge for worker count, a counter for
+// resync events, and a ready/not-ready gauge for alerting on extended outages.
+type Cache struct {
```
thockin, line 38: IIUC there's nothing here that couldn't be templated away.  E.g. `Cache[ateapipb.Worker]` could easily then cover `ateapipb.Actor` or anything else.

thockin:
Appreciate the TODO - 1 more and I am LGTM

cmd/ateapi/internal/store/ateredis/ateredis.go
juli4n: Done.

cmd/ateapi/internal/workercache/workercache.go
juli4n: That would require abstracting away store's Interface.ListWorkers, WatchWorker, etc. I would worry about a generic cache if we ever need it.

PR #301, opened 2026-06-24
fix: ensure image oci config is included when building actor oci spec

### LINKED ISSUE #300: Image OCI config is dropped when building actor runtime spec
## Expected Behavior

When an actor is created from an ActorTemplate, the workload container should inherit the
environment variables baked into its image's OCI config (`Config.Env`), the same way `docker run`
and containerd/CRI do.

## Actual Behavior

The image's OCI config is discarded when building the actor's runtime spec. The process only ever
sees a hardcoded `PATH` plus whatever env the ActorTemplate sets explicitly; the image's `ENV`
directives never reach the process.

1. Build and push an image with `ENV` vars set:
   ```dockerfile
   FROM busybox:1.36
   ENV DEMO_VAR=example
   ```
   ```
   docker buildx build --platform linux/arm64 -t <registry>/env-demo:v1 --push .
   ```
2. Create an ActorTemplate that references the image:
   ```yaml
   apiVersion: ate.dev/v1alpha1
   kind: ActorTemplate
   metadata:
     name: env-demo
     namespace: <ns>
   spec:
     workerPoolRef: { name: <pool>, namespace: <ns> }
     pauseImage: <pause-image@sha256:...>
     snapshotsConfig: { location: <gs://.../env-demo> }
     runsc: { arm64: { url: <...>, sha256Hash: <...> }, amd64: { url: <...>, sha256Hash: <...> } }
     containers:
       - name: app
         image: <registry>/env-demo@sha256:...   # the image from step 1, digest-pinned
         command: ["/bin/sh","-c","echo DEMO_VAR=$DEMO_VAR; env; sleep 3600"]
   ```
3. Inspect the actor process's environment which should be forwarded to stdout. Note that it `DEMO_VAR` is empty.

jmhbh:
**Summary**
- Fixes a bug where image oci config is not included when building the actor oci spec by ensuring that we extract the image supplied env from the config manifest and propagate it to the `buildActorOCISpec` function.

**Testing**
- Followed the steps listed in issue https://github.com/agent-substrate/substrate/issues/300 and confirmed that the image env is now captured correctly in the logs and that they are correctly overridden by the actor template env for duplicate keys.

Fixes #300

- [ ] Tests pass


cmd/atelet/oci.go
```diff
@@ -101,15 +101,37 @@ func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryP
...
+			}
+			if _, ok := seen[key]; ok {
+				continue
+			}
+			seen[key] = struct{}{}
+			out = append(out, e)
+		}
+	}
+
+	add("PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin")
```
juli4n: Should we drop this now?

cmd/atelet/oci.go
```diff
@@ -101,15 +101,37 @@ func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryP
...
+	add(templateEnv...)
```
juli4n, line 127: The precedence here seems flipped, templates envs should override image envs.

cmd/atelet/internal/memorypullcache/memorypullcache.go
```diff
@@ -142,36 +148,48 @@ func (c *MemoryPullCache) Fetch(ctx context.Context, ref string) (io.ReadCloser,
 
 	img, err := remote.Image(parsedRef, remoteOptions...)
 	if err != nil {
-		return nil, fmt.Errorf("in remote.Image: %w", err)
+		return nil, nil, fmt.Errorf("in remote.Image: %w", err)
+	}
+
+	var imageEnv []string
+	if cfg, cfgErr := img.ConfigFile(); cfgErr != nil {
```
juli4n: Isn't the config manifest mandatory? Why continue and not returning and error and hard-failing?

jmhbh: its currently only consumed on this code path which is why I added a warn log and opted not to hard fail. However, I think there is merit in your suggestion since if the resulting image needs the env vars it would be broken anyways particularly given that I've made the other change to remove the default path.

cmd/atelet/oci.go
jmhbh: yeah realistically speaking the path will be supplied in the image env.

cmd/atelet/oci.go
jmhbh: fixed.

jmhbh:
@juli4n thanks for your initial review, do you mind taking another pass?

jmhbh:
@juli4n sorry for pinging again, but could you please take another pass when you have a chance? Thanks!

jmhbh:
@juli4n bumping this for another pass. Thanks!

cmd/atelet/oci.go
juli4n: Looking at containerd, it looks like it does have a default (https://github.com/containerd/containerd/blob/52f81245d3aa434123dfa5d2a7f0db6644a70882/pkg/oci/spec.go#L42) so we should probably keep it?

cmd/atelet/internal/memorypullcache/memorypullcache.go
```diff
@@ -37,11 +38,15 @@ type MemoryPullCache struct {
...
 
-	// Map from hexadecimal sha256 hash of image to byte contents of composed
-	// tarball
+	// Map from hexadecimal sha256 hash of image to the cached image (tarball + config env).
 	cache *lru.Cache
 }
 
+type cachedImage struct {
+	tar []byte
+	env []string
```
juli4n: How about storing v1.Config here?

```
type cachedImage struct {
  tar []byte
  cfg v1.Config
}
```

The two main reasons are that it's more self-documented, and that it includes data we very likely will need soon as we start honoring more base execution parameters from the image.

jmhbh: good callout, I updated this as you suggested

cmd/atelet/oci.go
jmhbh: I added the default path back only if the imageEnv is empty

cmd/atelet/internal/memorypullcache/memorypullcache.go
```diff
@@ -142,36 +148,44 @@ func (c *MemoryPullCache) Fetch(ctx context.Context, ref string) (io.ReadCloser,
...
 	if err != nil {
-		return nil, fmt.Errorf("in remote.Image: %w", err)
+		return nil, nil, fmt.Errorf("in remote.Image: %w", err)
+	}
+
+	var imageCfg v1.Config
+	if cfg, cfgErr := img.ConfigFile(); cfgErr != nil {
+		return nil, nil, fmt.Errorf("while reading image config: %w", cfgErr)
+	} else if cfg != nil {
+		imageCfg = cfg.Config
```
juli4n: Is this really possible? Can `ConfigFile()` return no error but a nil `cfg`? If not, I would just remove the `else if cfg != nil` statement. Otherwise, we need to worry about the downstream implications of returning a pointer to a zero-initialized `imageCfg` from here on

cmd/atelet/oci_test.go
```diff
@@ -116,9 +117,67 @@ func TestBuildActorOCISpec_IdentityMount(t *testing.T) {
 	}
 }
 
+// mergeActorEnv test template env overrides image env.
+func TestMergeActorEnv(t *testing.T) {
```
juli4n, line 120: Nit: I think `mergeActorEnv` is actually a great candidate for a table-driven test (pure, side-effect free function, simple parameters). Can we restructure it that way? I think the main benefit is that the assertion logic becomes the same across all the test cases. Additionally, if we assert on the exact returned slice, we don't really need the `countKey` utility function.

I'm thinking something like this:

```
   func TestMergeActorEnv(t *testing.T) {
        defaultPath := "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
  
        tests := []struct {
            name        string
            imageEnv    []string
            templateEnv []string
            want        []string
        }{
            {
                name:        "template overrides image by key",
                imageEnv:    []string{"FOO=image"},
                templateEnv: []string{"FOO=template"},
                want:        []string{"FOO=template", defaultPath},
            },

             // Other test cases.
        }
  
        for _, tc := range tests {
            t.Run(tc.name, func(t *testing.T) {
                got := mergeActorEnv(tc.imageEnv, tc.templateEnv)
                if !slices.Equal(got, tc.want) {
                    t.Errorf("mergeActorEnv(%v, %v) =\n  %v\nwant:\n  %v", tc.imageEnv, tc.templateEnv, got, tc.want)
                }
            })
        }
    }
```

jmhbh: sure, done

juli4n (approved):

PR #302, opened 2026-06-25
fix: update DNS suffix in agent-secret demo

mesutoezdil:
Fixes incorrect DNS suffix in demo curl commands.

- `actors.resources.substrate.k8s.io` -> `actors.resources.substrate.ate.dev` (3 occurrences)


bowei:
Need to sign CLA.

bowei (approved):

Issue #126, opened 2026-05-29
Add transparent actor egress policy and credential injection


EItanya:
## Description

Substrate should support transparent egress policy for actors. This should let operators define which external destinations an actor can reach without requiring actor code changes or proxy environment variables.

The policy model should be actor-aware rather than only pod-aware. Actors are dynamically started, restored, suspended, and reused across workers, so egress policy needs to follow actor and template identity through that lifecycle.

Credential handling is a large motivation. Today, if an actor needs to call an external service, credentials generally need to be made available inside the actor runtime. That increases the blast radius if the actor is compromised or prompt-injected, because the actor can potentially exfiltrate credentials or use them against unintended destinations.

Substrate should instead support destination-scoped credential injection. The runtime should be able to attach a configured credential only when the actor is calling an explicitly allowed destination. The actor should not need direct access to the raw credential by default.

## Goals

* Default-deny actor egress.
* Allow egress by host and port.
* Apply policy without actor application changes.
* Support destination-scoped credential injection for allowed destinations.
* Keep raw credentials out of the actor environment when possible.
* Support plaintext HTTP policy.
* Support HTTPS passthrough and HTTPS inspection modes.
* Clean up policy state when actors suspend, fail, or move workers.
* Expose effective policy status for debugging.
* Emit audit-relevant allow and deny decisions.

## Non-Goals

* Do not require actor applications to set proxy environment variables.
* Do not require actor application code changes.

## HTTPS Inspection and MITM

HTTPS policy has two useful modes:

* Passthrough: allow or deny TLS connections based on connection metadata such as hostname and port, without decrypting application traffic.
* Inspection: terminate actor-side TLS, inspect or modify the HTTP request, then establish upstream TLS to the destination.

Credential injection for HTTPS generally requires inspection. If the runtime needs to add an authorization header or similar request credential, it must be able to see and modify the HTTP request before it is sent upstream.

Inspection requires the actor to trust a Substrate-managed certificate authority. That trust can be prepared before golden snapshot creation or provided through a runtime trust bundle. This should be explicit in the policy mode because some workloads will not support inspection:

* Certificate-pinned clients may fail.
* Workloads with private trust stores may need extra integration.
* Some users may prefer passthrough mode for destinations where inspection is not required.

## Example Capability Sketch

This is intentionally illustrative rather than a final API.

```yaml
egress:
  defaultAction: Deny
  rules:
  - to:
      host: api.example.com
      ports: [443]
    mode: Inspect
[...]

PR #303, opened 2026-06-25
Add initial threat model for substrate

mtaufen:
This adds an initial threat model for Substrate. The threat model was originally authored in this Google Doc and shared on the mailing list.

- https://docs.google.com/document/d/1UAuJzfgkXe6Qa9TTp4wwG0zuHFIGIxEXqafveGq8f9s/edit?resourcekey=0-J-c0L_AJFNjXBX_8TjMbMA&tab=t.0#heading=h.j4pfitmou99h
- https://groups.google.com/g/ate-dev/c/xEPg4RKv-Uk/m/hMoatPvEAAAJ

Note that access to the doc is gated by mailing list membership, you must join the list to view it and and individual access requests can't be granted due to a policy restriction. But please comment on this PR at this point if you have feedback :).


mesutoezdil:
Thanks for landing this in the repo. For some reason, I couldn't access the Google link that was shared.

While reading it, for me part of these threats come from the architecture itself. I cannot share my excalidraw link, but you will find a screenshot.

If we move a few things the list gets shorter. 

I put an alternative in the attached image, the reasoning behind it:

1. ate-api-server keeps api, db and identity broker in one box. This is the worst point for me, one compromise and all 3 go at the same time. Identity should not live in the same component as the state db. Better a separate SPIFFE/SPIRE issuer, short lived SVIDs, and no long lived secret kept in the db.

2. If Router is on the A2A path, a compromised Router reads and rewrites every agent to agent message, this is a large MITM surface. I would keep Router only for control plane (discovery, policy) and do the real A2A over mesh mTLS between the SVIDs, with the token bound to the channel, so even the mesh can not impersonate an agent.

3. In the worker pod, ateom and the sandbox look like the same trust domain. If the sandbox runs agent code and escapes, it is sitting next to the keys. The component holding the agent private key should not be in the same namespace as the sandbox. Sandbox in a microVM (Kata or Firecracker), identity terminated on the node side (Atelet), short ttl on the SVID.

4. Pod Snapshots Object Storage is the part I worry about most. A memory snapshot carries secrets, live tokens, also in flight messages and data from other agents. 2 problems from this. First, a token taken from a snapshot, if it is not channel bound, can be replayed after restore on another connection. Second, if you can snapshot or fork, you get 2 live copies with the same identity, single writer is broken and the same side effect can run twice. So checkpoints should be encrypted and credential scrubbed, creds re-minted on restore instead of restored from the dump, plus generation/epoch fencing and idempotency keys so a fork can not act twice. The Actor Directory could be the place enforcing single activation. And it is a fair question if a full memory snapshot is needed at all, or if explicit session state is enough for warm start and safer.

5. I could not find a fail closed point for egress. (I have share some insisght here as well: https://docs.google.com/document/d/1KmpIFu2gnqy9gp95wASgIo_vkJ_dA1DZckV8upET6bs/edit?tab=t.0#heading=h.3wbvdrfz1pyw) Tool calls and A2A go to external systems. If the policy plane is down, what is the behaviour? If it fails open, a partition becomes free egress for a compromised agent. Egress should be per node, default deny, SNI/ECH aware (no SNI spoof bypass, and ECH should not blind the enforcement), and fail closed.

For me the hardest one is 4, the checkpoint and fork and identity part, I do not have a fully clean answer there yet.

Screenshot 2026-06-26 at 23 36 55

docs/threat-model.md
```diff
@@ -0,0 +1,137 @@
...
+![Substrate Architecture Diagram](./assets/threat-model-diagram.svg)
+
+# Key Components
+
+* **ate-api-server:** Actor creation and scheduling and credential issuance.  
+* **atelet:** Per-node daemon, performs snapshot/resume.  
+* **ateom:** Per-worker Pod sidecar, running inside the worker Pod. Ateom sets up "interior" sandboxes in the worker Pod and manages sandbox lifecycle, including image pulls. It currently uses gvisor but Substrate will support multiple microvm solutions.  
+* **Worker:** Preprovisioned Pods that actors get scheduled to.  
+* **Actor:** The core compute primitive, gets scheduled to/from worker via Run for cold start and Resume for snapshot resume.  
+* **Actor IP:** Actor networking is based on Pod networking. Each actor gets the IP of the worker it's currently scheduled to. The ateom has the opportunity to set up additional rules when it sets up interior sandboxes.
```
thockin, line 44: I'm not sure it matters here, but I am not sure that an actor having "real" pod networking is going to mean anything.  All traffic will be captured, so the touchpoint is near zero.

docs/threat-model.md
```diff
@@ -0,0 +1,137 @@
...
+* **Threats:** Expected risks in a naïve implementation of Substrate.  
+* **Mitigating Invariants:** High level properties which, if true, mitigate the threat.  
+* **Suggested Concrete Mitigations:** Specific options for implementing the mitigating invariants, based on current understanding of Substrate.  
+* **Notes:** Additional relevant information.
+
+## Attacks from External Networks
+
+| GitHub Issue | Priority | Threat | Mitigating Invariants | Suggested Concrete Mitigations | Notes |
+| :---- | :---- | :---- | :---- | :---- | :---- |
+|  | Critical | External attacker can access actors over the internet | Access to actors over the external internet is blocked. | Recommend use of infrastructure firewall to limit external ingress/egress in documentation (cloud specific). Use Kubernetes NetworkPolicy to limit external ingress/egress by default. Use additional network policy features depending on CNI. |  |
```
thockin: I think this can be split up - I see the "internal" section below but no equivalent of this line.

External: Actors, workers, and the atenet routers are not directly exposed to the internet in most clusters.  Doing so would require explicit configuration.

Internal: Many kube clusters are managed as first-class citizens of their local network, so peers on the internal network COULD  reach then at an IP level.  This is not different than any other workload running in Kubernetes.

docs/threat-model.md
```diff
@@ -0,0 +1,137 @@
...
+|  | Critical | External attacker can access nodes over the internet | Access to nodes over the external internet is blocked. | Recommend use of infrastructure firewall to limit external ingress/egress in documentation (cloud specific). Use Kubernetes NetworkPolicy to limit external ingress/egress by default. Use additional network policy features depending on CNI. |  |
```
thockin: Saying that access is "blocked" in all these cases feels wrong. It feels like it is implying that we did something when in actuality it's the "normal" disposition of Kubernetes.  This sort of document is outside my core expertise -- is that distinction important?

docs/threat-model.md
```diff
@@ -0,0 +1,137 @@
...
+|  | High | Privilege escalation via access to sensitive labels. | If Substrate offers its own resource labeling mechanism, it must also offer a way to authorize label updates on a per-label basis. | Substrate authorization system requires explicit authorization to update metadata, separate from updating resource body. Substrate authorization system supports per-label authorization rules. | Plenty of attacks in K8s were possible because labels had semantic meaning, but the permission model could implicitly granted access to modify labels, even if it was inappropriate. For example, /status subresource allows label updates. Substrate shouldn't repeat this mistake.  |
```
thockin, line 77: This is sort of presuming what labels are for, but we don't really even have labels yet.  I think the point about being more careful is important, though.

docs/threat-model.md
```diff
@@ -0,0 +1,137 @@
...
+|  | High | Attacker gains control of Substrate API server, router, or other ingress/egress proxy. | Isolate the control plane from the data plane, and isolate data plane ingress from sandboxes. | Don't co-locate ate-apiserver or other control-plane components on the same machines as the untrusted sandboxes. Consider running any gateway/router that enables direct interaction with sandboxes on a separate VM from the sandboxes, or using a zero-trust architecture where traffic is encrypted and authenticated end-to-end. |  |
```
thockin: > Don't co-locate ate-apiserver or other control-plane components on the same machines as the untrusted sandboxes.

I think the point here should be clarified.  While we trust the sandboxing tech, there's some ADDITIONAL security to be had by segregating.  For large users, or users who run truly 3rd part code in sandboxes, the exposure risk of not segregating likely outweighs any potential benefits.

docs/threat-model.md
```diff
@@ -0,0 +1,137 @@
...
+|  | High | Attacker who can create ActorTemplates specifies malicious runtime. | Ensure available runtime can only be configured by administrators. | Consider a mechanism like RuntimeClass to decouple configuration of available runtimes from consumption of available runtimes. |  |
+|  | High | Attacker who can create ActorTemplates can read or write any storage buckets atelet has access to. | Ensure that bucket access is least-privilege. | Use credentials derived from actor identity to read snapshots. Configure permissions to prevent atelet or nodes from having access to sensitive buckets. | For example: Attacker creates an ActorTemplate with the runsc URL or golden snapshot URL pointing to an arbitrary bucket in the same project/resource scope as the cluster. If atelet has project-wide access to buckets, this could cause the state to be pulled into the worker pod or malicious actor. Similarly, an attacker could set the snapshots URL to point to an internal infrastructure bucket, causing data to be written to that bucket. |
```
thockin: This suggests to me that having a snapshot URL is overly-powerful.  Why do we need more than an indentifier which atelet can combine with other configuration to produce that URL?

docs/threat-model.md
thockin: This overlaps a bit with "Access to the internal network allows arbitrary actions to be performed on ate-apiserver" ?

docs/threat-model.md
```diff
@@ -0,0 +1,137 @@
...
+|  | Medium | Flat namespace of actors encourages broad permission grants or complex graph-oriented policy. | Support a grouping mechanism that can be used in policy controls. | Add namespaces to Substrate, similar to Kubernetes. |  |
+|  | Medium | DNS misconfiguration | Access to DNS configuration should be limited to authoritative controllers. Routing should use stable configurations and query the API for the current IP before routing each request.  | Don't co-locate controllers with access to sensitive system state on the same nodes as actors. Limit permissions to update DNS configuration. Actively query Substrate API to ensure IPs are as up-to-date as possible. Potentially use mTLS based on actor DNS name between ate-router and actors. | As noted above, a flood of requests could create high read load on ate-apiserver. Caching could be considered, but cache invalidation during actor rescheduling would be important to avoid misrouting traffic.  Establishing a backend mTLS tunnel between ate-router and each actor based on a serving cert signed for the actor's DNS name could be another approach to avoiding misrouting. |
```
thockin, line 91: Do we consider DNS info-leaks as a threat?  E.g. I can brute-force discover the names of other namespaces and actors by probing DNS

thockin (commented):
It took me a while to get the swing of reading this.  I left my early comments, but other than internal/external I don't think the rest matter.

mtaufen:
Thanks @mesutoezdil, agree with many of those points, especially risks associated with snapshots.

Some other thoughts below. Overall we're trying to keep this threat model slightly above current architectural choices as the architecture is changing too rapidly to build a more detailed threat model on. That said, the intent is to help guide these debates so architectural feedback is still valid:

> ate-api-server keeps api, db and identity broker in one box. This is the worst point for me, one compromise and all 3 go at the same time. Identity should not live in the same component as the state db. Better a separate SPIFFE/SPIRE issuer, short lived SVIDs, and no long lived secret kept in the db.

The issuer needs to consult Substrate's record of where actors are scheduled to determine if it's valid to issue a credential to an actor in a particular worker. We could probably make improvements, but that query is still necessary, and it's likely that a compromise of the Substrate API would still grant enough control to schedule a workload and get a credential from the Issuer. One thing we might want to do is support KMS APIs for credential signing. Right now (AIUI) we are not storing any secret _directly_ in the DB. cc @ahmedtd for thoughts on these suggestions. 

> If Router is on the A2A path, a compromised Router reads and rewrites every agent to agent message, this is a large MITM surface. I would keep Router only for control plane (discovery, policy) and do the real A2A over mesh mTLS between the SVIDs, with the token bound to the channel, so even the mesh can not impersonate an agent.

This is a good point. We do currently have a suggestion to switch to a zero trust model in the suggested mitigations.

> In the worker pod, ateom and the sandbox look like the same trust domain. If the sandbox runs agent code and escapes, it is sitting next to the keys. The component holding the agent private key should not be in the same namespace as the sandbox. Sandbox in a microVM (Kata or Firecracker), identity terminated on the node side (Atelet), short ttl on the SVID.

I'd like to see support for proxies hosting credentials on behalf of sandboxes, as well. There are a few cases where the thing in the sandbox is not an LLM and may need direct access to a credential, Substrate wants to support that, too. 

> Pod Snapshots Object Storage is the part I worry about most. A memory snapshot carries secrets, live tokens, also in flight messages and data from other agents. 2 problems from this. First, a token taken from a snapshot, if it is not channel bound, can be replayed after restore on another connection. Second, if you can snapshot or fork, you get 2 live copies with the same identity, single writer is broken and the same side effect can run twice. So checkpoints should be encrypted and credential scrubbed, creds re-minted on restore instead of restored from the dump, plus generation/epoch fencing and idempotency keys so a fork can not act twice. The Actor Directory could be the place enforcing single activation. And it is a fair question if a full memory snapshot is needed at all, or if explicit session state is enough for warm start and safer.

Big +1 to overall snapshot risk. We need some new strategies to deal with the snapshotting risk, not least of which is lateral movement potential via rescheduling. I believe Substrate plans to leave "once" semantics to whichever framework users bolt on top. Would be great to see broader support for channel-bound tokens across the industry :). I'm not sure there's a great way to scrub credentials from snapshots (I think we can avoid them during snapshotting if they are in predictable filesystem locations, but harder and likely breaking to scrub them from memory). On TTLs/fencing: @ahmedtd what are we currently setting the lifetimes to?

> I could not find a fail closed point for egress. (I have share some insisght here as well: https://docs.google.com/document/d/1KmpIFu2gnqy9gp95wASgIo_vkJ_dA1DZckV8upET6bs/edit?tab=t.0#heading=h.3wbvdrfz1pyw) Tool calls and A2A go to external systems. If the policy plane is down, what is the behaviour? If it fails open, a partition becomes free egress for a compromised agent. Egress should be per node, default deny, SNI/ECH aware (no SNI spoof bypass, and ECH should not blind the enforcement), and fail closed.

Some of the egress stuff is currently under discussion in #126 and https://docs.google.com/document/d/1KmpIFu2gnqy9gp95wASgIo_vkJ_dA1DZckV8upET6bs/edit?tab=t.0 and https://cloud-native.slack.com/archives/C0B6M3E2J3D/p1781216438948749.

docs/threat-model.md
mtaufen: Possibly. Actor IDs are pretty opaque, but namespace names could be sensitive if human readable. But do we need to include namespaces in DNS names, given flat routing namespace (or could we just use namespace IDs instead of human names)? cc @bowei

docs/threat-model.md
mtaufen: It's different, "Access to the internal network allows arbitrary actions to be performed on ate-apiserver" is "there's no authz so anyone can do whatever they want via the API" whereas this one is "they figured out how to own the API, now they're on the inside."

docs/threat-model.md
mtaufen: Agree. I think we should redesign the API here to be canonical (or use a plugin?) about how to find golden snapshots, runsc, etc, and we should make that access subject to a scheduling-aware authz check. cc @juli4n

docs/threat-model.md
mtaufen: SG, I'll add some clarification

docs/threat-model.md
mtaufen: Yeah it's a bit forward-looking. Someone always eventually wants some arbitrary k/v map embedded in metadata, and then someone eventually uses it for sensitive configuration :).

docs/threat-model.md
mtaufen: To be fair, calling out access to Nodes from the internet (instead of just access to Workers/Actors/API) is probably going a bit beyond what Substrate is responsible for. I added it for completeness, but I'll add a note on that.

docs/threat-model.md
mtaufen: Can you expand more on this?

docs/threat-model.md
mtaufen: I'll add a clarification on external being less likely, and a recommendation to also use firewalls as part of "Access to the internal network allows arbitrary actions...", but I'm not sure exactly what you're asking here?

mesutoezdil:
> Thanks Mesut Oezdil (@mesutoezdil), agree with many of those points, especially risks associated with snapshots.
> 
> Some other thoughts below. Overall we're trying to keep this threat model slightly above current architectural choices as the architecture is changing too rapidly to build a more detailed threat model on. That said, the intent is to help guide these debates so architectural feedback is still valid:
> 
> > ate-api-server keeps api, db and identity broker in one box. This is the worst point for me, one compromise and all 3 go at the same time. Identity should not live in the same component as the state db. Better a separate SPIFFE/SPIRE issuer, short lived SVIDs, and no long lived secret kept in the db.
> 
> The issuer needs to consult Substrate's record of where actors are scheduled to determine if it's valid to issue a credential to an actor in a particular worker. We could probably make improvements, but that query is still necessary, and it's likely that a compromise of the Substrate API would still grant enough control to schedule a workload and get a credential from the Issuer. One thing we might want to do is support KMS APIs for credential signing. Right now (AIUI) we are not storing any secret _directly_ in the DB. cc Taahir Ahmed (@ahmedtd) for thoughts on these suggestions.
> 
> > If Router is on the A2A path, a compromised Router reads and rewrites every agent to agent message, this is a large MITM surface. I would keep Router only for control plane (discovery, policy) and do the real A2A over mesh mTLS between the SVIDs, with the token bound to the channel, so even the mesh can not impersonate an agent.
> 
> This is a good point. We do currently have a suggestion to switch to a zero trust model in the suggested mitigations.
> 
> > In the worker pod, ateom and the sandbox look like the same trust domain. If the sandbox runs agent code and escapes, it is sitting next to the keys. The component holding the agent private key should not be in the same namespace as the sandbox. Sandbox in a microVM (Kata or Firecracker), identity terminated on the node side (Atelet), short ttl on the SVID.
> 
> I'd like to see support for proxies hosting credentials on behalf of sandboxes, as well. There are a few cases where the thing in the sandbox is not an LLM and may need direct access to a credential, Substrate wants to support that, too.
> 
> > Pod Snapshots Object Storage is the part I worry about most. A memory snapshot carries secrets, live tokens, also in flight messages and data from other agents. 2 problems from this. First, a token taken from a snapshot, if it is not channel bound, can be replayed after restore on another connection. Second, if you can snapshot or fork, you get 2 live copies with the same identity, single writer is broken and the same side effect can run twice. So checkpoints should be encrypted and credential scrubbed, creds re-minted on restore instead of restored from the dump, plus generation/epoch fencing and idempotency keys so a fork can not act twice. The Actor Directory could be the place enforcing single activation. And it is a fair question if a full memory snapshot is needed at all, or if explicit session state is enough for warm start and safer.
> 
> Big +1 to overall snapshot risk. We need some new strategies to deal with the snapshotting risk, not least of which is lateral movement potential via rescheduling. I believe Substrate plans to leave "once" semantics to whichever framework users bolt on top. Would be great to see broader support for channel-bound tokens across the industry :). I'm not sure there's a great way to scrub credentials from snapshots (I think we can avoid them during snapshotting if they are in predictable filesystem locations, but harder and likely breaking to scrub them from memory). On TTLs/fencing: Taahir Ahmed (@ahmedtd) what are we currently setting the lifetimes to?
> 
> > I could not find a fail closed point for egress. (I have share some insisght here as well: https://docs.google.com/document/d/1KmpIFu2gnqy9gp95wASgIo_vkJ_dA1DZckV8upET6bs/edit?tab=t.0#heading=h.3wbvdrfz1pyw) Tool calls and A2A go to external systems. If the policy plane is down, what is the behaviour? If it fails open, a partition becomes free egress for a compromised agent. Egress should be per node, default deny, SNI/ECH aware (no SNI spoof bypass, and ECH should not blind the enforcement), and fail closed.
> 
> Some of the egress stuff is currently under discussion in #126 and https://docs.google.com/document/d/1KmpIFu2gnqy9gp95wASgIo_vkJ_dA1DZckV8upET6bs/edit?tab=t.0 and https://cloud-native.slack.com/archives/C0B6M3E2J3D/p1781216438948749.

Thx for the detailed answer.
on point 3 // proxy hosting credentials for non-LLM sandboxes: makes sense, not everything in the sandbox is an LLM. 1 thing to flag, this proxy becomes a new high-value target by itself, so it probably needs the same isolation thinking we want for ateom today.

on point 4 // about "once" semantics, yu said Substrate plans to leave this to the framework layer. I m not sure the framework can really enforce single-writer on its own, if the platform itself can produce 2 live copies with the same identity after a fork, the framework only sees one connection at a time, it has no way 2 know the other copy exists unless the platform exposes sth like a generation or epoch id it can check against.  

Also, on credential scrubbing, my idea was not about scrubbing memory at snapshot time, it was about never trusting whatever creds come back after a restore at all, mint new ones on every resume bound to the new generation/channel, so it doesnt matter what is still sitting in memory, those old creds are useless on restore. That avoids the memory scrubbing problem instead of solving it.                                
                                                                                                         
On point 5 // I’m not sure if taken note of them, but I’d left a few comments on the doc earlier.

mtaufen:
Thanks :).

>on point 3 // proxy hosting credentials for non-LLM sandboxes: makes sense, not everything in the sandbox is an LLM. 1 thing to flag, this proxy becomes a new high-value target by itself, so it probably needs the same isolation thinking we want for ateom today.

Definitely :)

> on point 4 // about "once" semantics, yu said Substrate plans to leave this to the framework layer. I m not sure the framework can really enforce single-writer on its own, if the platform itself can produce 2 live copies with the same identity after a fork, the framework only sees one connection at a time, it has no way 2 know the other copy exists unless the platform exposes sth like a generation or epoch id it can check against.

AIUI, there shouldn't be 2 live copies with the same actor_id. @juli4n @MushuEE  is this currently possible? If we have more to do here, we should probably discuss this in a separate issue as names/ids/coordination/idempotence/etc are quite fundamental properties that should be supported in the APIs. Perhaps https://github.com/agent-substrate/substrate/issues/12?

> Also, on credential scrubbing, my idea was not about scrubbing memory at snapshot time, it was about never trusting whatever creds come back after a restore at all, mint new ones on every resume bound to the new generation/channel, so it doesnt matter what is still sitting in memory, those old creds are useless on restore. That avoids the memory scrubbing problem instead of solving it.

We can probably do this for actor identity... we'll need to think about what to bind to and how many workloads will actually honor it.

mtaufen:
Discussed in standup today, we'll merge this in the current state and then start filing issues. Folks please feel free to send follow up PRs if you have additions to make.

mesutoezdil:
> Discussed in standup today, we'll merge this in the current state and then start filing issues. Folks please feel free to send follow up PRs if you have additions to make.

was there a meeting today? didnt know it

PR #309, opened 2026-06-25
fix: delete actors before templates in demo cleanup

### LINKED ISSUE #215: -delete-demo...only deletes the templates, doesn't delete actors
Example: https://github.com/agent-substrate/substrate/blob/main/hack/install-demo-counter.sh

Ideally before deleting the "counter" actortemplate, it would
1. finds all actor instances of the template with `kubectl ate get actors`...
2. suspend and delete the actors
3. delete the actortemplate itself.

supreme-gg-gg:
Fixes #215 

## Summary

This PR adds a shared `delete_demo_actors` helper and calls it from each demo’s `--delete-demo-*` hook. Before demo manifests are removed, it finds actors created from that demo’s `ActorTemplates`, gets them to a deletable state, and deletes them so demo teardown doesn’t leave dangling actors in the control plane.

`delete_demo_actors` uses `kubectl-ate` to list actors, filters by the demo’s namespace and template pairs, and for each match ensures the actor is in `STATUS_SUSPENDED` before deleting it.

## Testing

I tested this with several demo setups as follows:

```bash
hack/install-ate-kind.sh --deploy-demo-counter

kubectl ate resume actor my-counter-1 # STATUS_RUNNING

hack/install-ate-kind.sh --delete-demo-counter

kubectl ate get actors # empty
```

- [x] Tests pass


hack/install-ate.sh
```diff
@@ -102,6 +102,8 @@ run_kubectl_ate() {
     "$@"
 }
 
+source "${ROOT}"/hack/lib-demo-delete.sh
```
thockin: Rather than sourcing (which usally ends up in a tangle of deps) why not just include the cleanup functions in this file?

supreme-gg-gg: Makes sense, I've moved them to `install-ate.sh`

PR #290, opened 2026-06-23
ateom-gvisor: tag forwarded actor logs with ate.dev/container_name

BenTheElder:
While working on multi-container support for uVM in https://github.com/agent-substrate/substrate/pull/287 I noticed how difficult it is to debug multi-container workloads.

This makes it a lot easier. TODO: We could support filtering on this in kubectl ate (cc @HavenXia)

PR #311, opened 2026-06-25
feat: kubectl ate logs filter container supervisor

### LINKED ISSUE #294: Allow kubectl ate logs to filter by container and supervisor
We should add kubectl ate logs a way to specify whether you want all logs (default), logs for a specific container, or the ateom supervisor logs.

After #290 everything is actually in place, this should be just changes to the kubectl-ate client.

@HavenXia @BenTheElder

MayorFaj:
Fixes #294 

- [x] Tests pass

## Summary
Implements `kubectl ate logs actors <id>` can now filter by source:

- **default** — all of the actor's logs (unchanged)
- **`-c, --container <name>`** — only that container's logs
- **`--supervisor`** — only ateom lifecycle logs (`Actor restoring`/`restored`/…)

`--container` and `--supervisor` are mutually exclusive. After #290 every
forwarded container line carries an `ate.dev/container_name` label and
supervisor lifecycle lines carry none, so this is implemented client-side.

## What changed
**CLI (`cmd/kubectl-ate`)**
- New `--container/-c` and `--supervisor` flags.
- `logLineFilter` + `matchesSource()` encode the selection rules.
- `filterAndDisplayLogLine` reads `actor_id` and `container_name` atomically
  from the first recognized label key, and returns the zero time for any
  non-displayed line so follow mode's resume cursor only advances on lines the
  user actually sees (a reconnect can't skip wanted lines).
- `validateLogFilterFlags` enforces mutual exclusion, called from the cobra
  command and from `LogsActorRunner.Run` (fail-fast for direct construction).

**ateom log wrapper (`cmd/ateom-gvisor`) — security hardening**
- `WrapContainerLogs` strips application-provided `ate.dev/*` labels from *both*
  recognized label keys before injecting Substrate metadata, so an app cannot
  spoof another actor's `actor_id`/`container_name` under the key a consumer
  treats as authoritative.

**Docs**
- `cmd/kubectl-ate/README.md` and `docs/observability.md` document the flags and
  mutual exclusivity; examples use the required `kubectl ate logs actors` form.

## Design notes
- Mutual exclusion uses a value-based check rather than cobra's
  `MarkFlagsMutuallyExclusive`, which keys off `flag.Changed` and would wrongly
  reject `--container x --supervisor=false`.

## Testing
- Unit: filter table cases (container/supervisor/default, dual-key precedence +
  atomicity, foreign-actor), `validateLogFilterFlags`, `Run` fail-fast,
  follow-mode resume-cursor regression, and `TestWrapContainerLogs_CrossKeyLabelSpoofing`.
- `make test` green; `gofmt` / `go vet` clean.
- Verified on a kind cluster (counter demo): default, `-c counter`, `--supervisor`,
  and the mutual-exclusion error all behave as expected.

### DIFF
--- a/cmd/kubectl-ate/internal/cmd/logs_actors.go
+++ b/cmd/kubectl-ate/internal/cmd/logs_actors.go
@@ -38,8 +38,12 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-var followLogs bool
-var logsAtespaceFlag string
+var (
+	followLogs       bool
+	containerFlag    string
+	supervisorFlag   bool
+	logsAtespaceFlag string
+)
 
 var logsActorsCmd = &cobra.Command{
 	Use:     "actors <actor-id>",
@@ -53,6 +57,8 @@ func init() {
 	logsActorsCmd.Flags().BoolVarP(&followLogs, "follow", "f", false, "Specify if the logs should be streamed.")
 	logsActorsCmd.Flags().StringVarP(&logsAtespaceFlag, "atespace", "a", "", "Atespace the actor lives in")
 	_ = logsActorsCmd.MarkFlagRequired("atespace")
+	logsActorsCmd.Flags().StringVarP(&containerFlag, "container", "c", "", "Show logs only from the named container within the actor. Mutually exclusive with --supervisor.")
+	logsActorsCmd.Flags().BoolVar(&supervisorFlag, "supervisor", false, "Show only the ateom supervisor (lifecycle) logs. Mutually exclusive with --container.")
 	logsCmd.AddCommand(logsActorsCmd)
 }
 
@@ -84,6 +90,8 @@ type LogsActorRunner struct {
 	stdout            io.Writer
 	stderr            io.Writer
 	follow            bool
+	container         string
+	supervisor        bool
 	pollInterval      time.Duration
 	reconnectInterval time.Duration
 	tickerInterval    time.Duration
@@ -102,6 +110,9 @@ func (r *LogsActorRunner) Run(ctx context.Context, actorID string) error {
 	}
 
 	defer r.apiClient.Close()
+	if err := validateLogFilterFlags(r.container, r.supervisor); err != nil {
+		return err
+	}
 	if r.follow {
 		return r.runFollow(ctx, actorID)
 	}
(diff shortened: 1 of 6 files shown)


cmd/kubectl-ate/internal/cmd/logs_actors.go
```diff
@@ -320,30 +384,41 @@ func filterAndDisplayLogLine(line, targetActorID string, w io.Writer) (time.Time
...
-			if labels, ok := labelsAny.(map[string]any); ok {
-				if id, ok := labels["ate.dev/actor_id"].(string); ok && id != "" {
-					actorID = id
-					break
-				}
-			}
+	// Read actor_id and container_name from the same (first actor-identifying)
+	// label map so a line's source is never split across the two keys.
+	var actorID, containerName string
+	for _, labelKey := range logActorLabelKeys {
```
juli4n: Instead of iterating over all label keys, how about having a method that takes the log entry map (i.e. `m`) and returns `labels`. It iterates on the possible keys and probes them, and return the first one that matches. That simplifies the control flow at the caller.

MayorFaj: Done, extracted it into a helper so the caller no longer iterates the keys inline. Based on follow-up review it ended up as actorSource(m) returning - actorID, containerName - (reads both from the first map carrying `ate.dev/actor_id`), which keeps the same-map guarantee and drops the loop at the call site. Thanks for the nudge!

cmd/kubectl-ate/internal/cmd/logs_actors.go
```diff
@@ -50,6 +54,8 @@ var logsActorsCmd = &cobra.Command{
 
 func init() {
 	logsActorsCmd.Flags().BoolVarP(&followLogs, "follow", "f", false, "Specify if the logs should be streamed.")
+	logsActorsCmd.Flags().StringVarP(&containerFlag, "container", "c", "", "Show logs only from the named container within the actor. Mutually exclusive with --supervisor.")
+	logsActorsCmd.Flags().BoolVar(&supervisorFlag, "supervisor", false, "Show only the ateom supervisor (lifecycle) logs. Mutually exclusive with --container.")
```
juli4n: Why making `--container` and `supervisor` mutually exclusive? They seem orthogonal to me.

I was expecting:
* `--supervisor` (boolean, defaults to false): whether to include supervisor logs or not
* `--container` (string, default to ""): if set, restricts to the logs of this container.

WDYT?

MayorFaj: You're right that they read as orthogonal, the exclusivity is an artifact of how I defined `--supervisor`, not something fundamental.

I'll switch to a union model: each flag *includes* a source, and a bare invocation (no flags) still shows everything

The one thing I'd want to agree on is the default semantics: bare --supervisor=false wouldn't *hide* lifecycle lines, it just wouldn't restrict to them, so the no-flags default stays "show all.

MayorFaj: @juli4n PTAL when you get the chance

PR #313, opened 2026-06-26
microvm MVP cleanup: align rootfs behavior to gvisor, ...

BenTheElder:
Follow-up to https://github.com/agent-substrate/substrate/pull/287 / #123 

- align snapshot behavior: use a tmpfs (for writes) on top of read-only viritio-fs mount for the container image rootfs instead of ext4 images
- enable multiple container support
- cleanup stale comments

- [x] Tests pass


BenTheElder:
@dims this should significantly unblock large images. Still needs follow-ups to optimize image pull / unpack. I tested with the 3GB openshell image.

dims:
@BenTheElder does this need me to build/use `virtiofsd` from source?

cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -0,0 +1,241 @@
...
+			Driver:        "overlayfs",
+			Source:        "overlay",
+			Fstype:        "overlay",
+			MountPoint:    ovlRoot,
+			DriverOptions: []string{createDir + "=" + upper, createDir + "=" + work},
+			Options:       []string{"lowerdir=" + lower, "upperdir=" + upper, "workdir=" + work},
+		},
+	}
+	pbSpec := SpecToAgentPB(spec)
+	pbSpec.Root = &agentpb.Root{Path: ovlRoot, Readonly: false}
```
dims, line 235: let's give each workload container its own cgroup? (similar to carrier's per-id path above)

```suggestion
	pbSpec.Root = &agentpb.Root{Path: ovlRoot, Readonly: false}

     // Per-workload cgroup: the shaped spec carries the actor-wide /ateomchv/<actorID>
     // (spec.go), which collides across containers — mirror the carrier's per-id path.
     if pbSpec.Linux != nil {
         pbSpec.Linux.CgroupsPath = "/ateomchv/" + workloadID	
     }
```

BenTheElder:
> [Benjamin Elder (@BenTheElder)](https://github.com/BenTheElder) does this need me to build/use virtiofsd from source?

Unfortunately yes but the scripts will handle it.

cmd/ateom-microvm/internal/kata/overlay_linux.go
BenTheElder: Nice catch! Fixed ...

cmd/ateom-microvm/run.go
```diff
@@ -431,16 +487,38 @@ func (s *AteomService) startActorContainer(ctx context.Context, ac *kata.AgentCl
...
+	return nil
+}
+
+// startOverlayContainer brings up one container's rootfs as overlay(virtio-fs RO
+// lower + guest-tmpfs upper): a carrier container (id == name) eager-binds the RO base
+// to /run/kata-containers/<name>/rootfs, then the workload (id == <name>-ovl) overlays
+// it with a tmpfs upper. On failure it dumps the guest overlay state.
+func startOverlayContainer(ctx context.Context, ac *kata.AgentClient, vsockPath string, c actorContainer) error {
+	carrierCtx, carrierCancel := context.WithTimeout(ctx, 30*time.Second)
+	err := ac.CreateCarrier(carrierCtx, c.name, c.spec)
```
dims, line 516: Will there be a problem if the user container itself has `-ovl` as a suffix? Do we need to make sure there is no collision between carrier and workload ids?

BenTheElder: Yes, we can switch to `_ovl` since k8s DNS-label restrictions shouldn't apply here but do apply to the user specified containers. Testing / fixing.

BenTheElder: Addressed this, I think that fix works well. It seems unlikely we'll want to allow invalid k8s container names ... later we might consider something like the dynamic containers KEP.

BenTheElder:
I need to rework the branch atop the CI changes now that we have those, so we get coverage on this PR :-)

Will do that before we merge [IE even if this LGTM to reviewers, let's wait until that is done, pointing an agent at it ...]

BenTheElder:
OK, should be synced with the e2e changes now

cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -0,0 +1,247 @@
...
+func StartVirtiofsd(ctx context.Context, o VirtiofsdOptions) (*exec.Cmd, error) {
+	bin := o.Binary
+	if bin == "" {
+		bin = "virtiofsd"
+	}
+	_ = os.Remove(o.SocketPath)
+	cmd := exec.Command(bin,
+		"--socket-path="+o.SocketPath,
+		"--shared-dir="+o.SharedDir,
+		"--cache=auto",
```
BenTheElder: Deferring: currently we only mount this ro, so we could use `cache=always`, but it's not clear yet how we will handle projected volumes cc @ahmedtd 

I think we might want to do a special writer for those and leverage the ability to exec into the sandboxes, in which case we could leave the virtiofsd mount fully read-only with no host <> guest churn and aggressively cache in the guest to improve performance. 

But until we decide I'm just going to leave this, using it is already a massive startup improvement versus writing to an ext4 for multi-gigabyte disks ... may be worse for read heavy workloads but we can iterate.

BenTheElder:
We have to inject the virtiofsd digest at build time, build isn't reproducible and it's probably too fragile to fix, plus we plan to exit compiling ourselves ASAP anyhow.

BenTheElder:
fixed the virtiofsd hash issue, and this is now running and passing in CI

cmd/ateom-microvm/internal/kata/overlay_linux.go
BenTheElder: Actually, I think we should use cache=always until we have proof we need otherwise. I have some hope we can avoid writes through virtiofsd and strictly use it to ~stream in the bulk rootfs.

BenTheElder: Done, with a TODO note.

dberkov (approved):
LGTM

PR #315, opened 2026-06-26
ateom-microvm: give the micro-VM guest DNS so name-based egress works

dims:
## Problem
A micro-VM (`ateom-microvm`) actor gets a guest interface, default route, and host NAT, but no `/etc/resolv.conf`: ateom drops atelet's resolv.conf bind (a host bind is meaningless inside a VM) and never sends `CreateSandbox.Dns`. The guest can reach IPs but not resolve names, so workloads doing name-based egress — e.g. an OpenShell helpdesk agent calling Ollama Cloud (`/chat`) — fail; only no-egress paths work.

## Fix
Write `/etc/resolv.conf` into the bundle rootfs before `mkfs.ext4` at golden boot (`RunWorkload`), copied verbatim from the worker pod's own resolv.conf (cluster DNS) — the same source the kata shim's `getDNS` uses. Errors out if the host has no usable resolv.conf rather than guessing. The restore path reuses the golden disk, so it inherits the file. One helper + one call; no behavior change for workloads that don't resolve names.

## Validation (kind on a KVM host)
- A helpdesk micro-VM actor `/chat` returns real model completions and survives suspend/resume.
- counter-microvm unaffected.

### DIFF
--- a/cmd/ateom-microvm/restore.go
+++ b/cmd/ateom-microvm/restore.go
@@ -99,6 +99,12 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
 		slog.InfoContext(ctx, "Reset actor rootfs disk to golden (template)", slog.String("id", id))
 	} else {
 		bundleRootfs := filepath.Join(ateompath.OCIBundlePath(ns, name, id, containers[0].GetName()), "rootfs")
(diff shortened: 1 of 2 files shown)


BenTheElder (approved):
I guess at some point we'll need to revisit passing it through like this, but this should get us rough parity with gVisor.

PR #317, opened 2026-06-26
fix: remove debug logs and update stale DNS comments in agent-secret

mesutoezdil:
- Remove two DEBUG log lines left in production handler
- Remove dead code block (actorID is always non-empty after the unknown fallback)


mesutoezdil:
@HavenXia would you mind taking a look when you get a chance, since you touched this file recently?

HavenXia (approved):
nit: The 2nd point in description says "Update stale substrate.k8s.io references to substrate.ate.dev in comments" but seems it has been cleared, please remove.

mesutoezdil:
Thanks for the review!

PR #323, opened 2026-06-26
fix: correct terminology and flag inconsistencies in docs

mesutoezdil:
- `README.md`: `--template` used with a space; `cmd/kubectl-ate/README.md` uses `=`. Aligned to `=`.
- `README.md`: removed stale note about navigating to root before running `hack/install-ate.sh`; the script anchors itself with `git rev-parse --show-toplevel`.
- `docs/architecture.md`: `WarmPools` is not a project term; changed to `WorkerPools`.
- `docs/architecture.md`: missing `to` in "it's important clarify".
- `COLLABORATING.md`: `(GMT-7)` is only correct during PDT; changed to `(PT)`.
- `docs/api-guide.md`: `pauseImage` example uses a GKE-specific image with no note for other clusters; added a comment with the generic alternative.


thockin:
It looks like checks are stuck and I don't know why

mesutoezdil:
> It looks like checks are stuck and I don't know why

A similar situation sometimes occurs in other repos as well

mesutoezdil:
@HavenXia Doc terminology fixes here, would you mind reviewing when you have time?

PR #324, opened 2026-06-26
fix: correct stale lock TTL value in comment

mesutoezdil:
Comment said 7 seconds but the code passes 30 seconds to acquireActorLock.


mesutoezdil:
Reopening with a clean branch from main.

PR #327, opened 2026-06-26
ci: run the micro-VM e2e in the kind e2e job

BenTheElder:
Free x86-64 ubuntu-latest runners expose /dev/kvm, so the existing kind e2e job can exercise the micro-VM (kata + cloud-hypervisor) runtime alongside gVisor on one shared control plane:
  - enable /dev/kvm (udev rule) before create-kind-cluster.sh, which then mounts it and labels the node for the microvm sandbox class;
  - deploy the counter-microvm demo (run-microvm-demo-kind.sh) next to the gVisor counter and run the demo lifecycle suite against both;
  - cache the assembled micro-VM assets (keyed on assemble.sh) so the expensive virtiofsd-from-source build only runs when the pins change; push-to-main + a weekly schedule keep the cache warm for PRs.

The demo lifecycle suite (create -> suspend -> resume -> in-RAM continuity) is reused for the micro-VM by parameterizing the source ActorTemplate via E2E_TEMPLATE_NAMESPACE/NAME (default the gVisor counter) and the golden-ready wait via E2E_TEMPLATE_READY_TIMEOUT.

PR AI assisted.

### DIFF
--- a/.github/workflows/pr-workflow.yaml
+++ b/.github/workflows/pr-workflow.yaml
@@ -17,6 +17,11 @@ on:
   pull_request:
   push:
     branches: [main]
+  schedule:
+    # Weekly run on the default branch keeps the micro-VM asset cache warm (GitHub
+    # evicts caches idle for 7 days). The push-to-main run populates the cache that
+    # PRs inherit; this refreshes it so PRs keep hitting it (no virtiofsd rebuild).
+    - cron: '37 4 * * 1'
 jobs:
   run-tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


BenTheElder:
I plan to follow-up with integration tests in CI, we have some in the repo, also https://github.com/agent-substrate/substrate/pull/313/changes will need to rebase on this and pick up the script cleanups, but I just want to land some basic regression coverage that includes atelet + ateom with uVM ...

dberkov (approved):
LGTM.

BenTheElder:
BTW:
https://docs.github.com/en/actions/concepts/runners/github-hosted-runners#supported-runners-and-hardware-resources:~:text=While%20nested%20virtualization%20is%20technically%20possible%20while%20using%20runners%2C%20it%20is%20not%20officially%20supported.%20Any%20use%20of%20nested%20VMs%20is%20experimental%20and%20done%20at%20your%20own%20risk%2C%20we%20offer%20no%20guarantees%20regarding%20stability%2C%20performance%2C%20or%20compatibility.

https://github.blog/changelog/2024-04-02-github-actions-hardware-accelerated-android-virtualization-now-available/

So technically support is best effort. If we see flakiness with the KVM infra we can split it to another job, but for now one job to improve efficiency.

PR #329, opened 2026-06-26
fix: skip empty node name in local snapshot info after pause finalization

mesutoezdil:
When the worker pod is gone from the DB during pause finalization, nodeName stays empty. 
The code still wrote NodeVmsWithLocalSnapshots: []string{""}, so on the next ResumeActor the scheduler looked for a worker with node name "", found none, and returned "no free workers available" permanently.
Fix: only set NodeVmsWithLocalSnapshots when nodeName is non-empty.

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_pause.go
+++ b/cmd/ateapi/internal/controlapi/workflow_pause.go
@@ -157,8 +157,11 @@ type FinalizePausedStep struct {
 
 func (s *FinalizePausedStep) Name() string { return "FinalizePaused" }
 func (s *FinalizePausedStep) IsComplete(ctx context.Context, input *PauseInput, state *PauseState) (bool, error) {
-	// The workflow is completely done ONLY if the status is PAUSED *and* we've successfully freed the worker.
-	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_PAUSED && state.Actor.GetAteomPodNamespace() == "", nil
+	// The workflow is done once the worker is freed and the actor reached PAUSED,
+	// or CRASHED (node name was lost, so it can never be safely resumed).
+	status := state.Actor.GetStatus()
+	terminal := status == ateapipb.Actor_STATUS_PAUSED || status == ateapipb.Actor_STATUS_CRASHED
+	return terminal && state.Actor.GetAteomPodNamespace() == "", nil
 }
 func (s *FinalizePausedStep) Execute(ctx context.Context, input *PauseInput, state *PauseState) error {
 	latestActor, err := s.store.GetActor(ctx, input.Atespace, input.ActorName)
@@ -202,9 +205,13 @@ func (s *FinalizePausedStep) Execute(ctx context.Context, input *PauseInput, sta
 			return err
 		}
 		latestActor.Status = ateapipb.Actor_STATUS_PAUSED
-		// TODO(dberkov) - what if we still don't know the node name? Maybe move to CRASHED status?
 		if nodeName == "" {
-			slog.Warn("Node name not found during finalize pause", "actor", input.ActorName)
+			// Without a node name we cannot record where the local snapshot lives,
+			// so the actor can never be resumed (findFreeWorker would search for a
+			// worker on an unknown node forever). Crash it instead of leaving it
+			// stuck in PAUSED.
+			slog.Warn("Node name not found during finalize pause, crashing actor", "actor", input.ActorName)
+			latestActor.Status = ateapipb.Actor_STATUS_CRASHED
 		}
 		// TODO(dberkov) - what if InProgressSnapshot is empty? That shouldn't be possible.
 		if latestActor.InProgressSnapshot != "" {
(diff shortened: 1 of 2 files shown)


cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -218,14 +218,15 @@ func (s *FinalizePausedStep) Execute(ctx context.Context, input *PauseInput, sta
 		}
 		// TODO(dberkov) - what if InProgressSnapshot is empty? That shouldn't be possible.
 		if latestActor.InProgressSnapshot != "" {
+			localInfo := &ateapipb.LocalSnapshotInfo{
+				SnapshotPrefix: latestActor.InProgressSnapshot,
+			}
+			if nodeName != "" {
+				localInfo.NodeVmsWithLocalSnapshots = []string{nodeName}
```
dberkov, line 222: resume from "SnapshotType_SNAPSHOT_TYPE_LOCAL" supposed to land on machine where files are stored locally, otherwise resume will fail. 

we need to find a root cause of the issue that node is empty. This is a bug/exception

dberkov:
@mesutoezdil - could you please explain how you was able to reproduce the issue, since the proposed fix will not work. The root cause of the issue is somewhere else.

mesutoezdil:
> Mesut Oezdil (@mesutoezdil) - could you please explain how you was able to reproduce the issue, since the proposed fix will not work. The root cause of the issue is somewhere else.

here are the reproduction steps, i wrote 2 unit tests:

T1 // (TestFinalizePausedStep_WorkerGone): worker pod is gone from db before FinalizePausedStep runs. GetWorker returns ErrNotFound. 
and
T2 // (TestFindFreeWorker_EmptyNodeRestriction): shows why it gets stuck, and old code wrote []string{""} into NodeVmsWithLocalSnapshots. findFreeWorker then looks for a worker with NodeName == "", finds none, returns nil forever. 

Screenshot 2026-06-27 at 07 07 43

see your point about root cause. Even with this fix, resume could pick a wrong node and snapshot restore fails.
and I saw your TODO on line 215, should we move actor to CRASHED when nodeName is empty instead?

dberkov:
@zoez7 is working on a PR introducing the CRASHED state. Could you please wait for her PR to be submitted and after it make a change to move the actor to this stage in case of this condition.

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -202,19 +205,24 @@ func (s *FinalizePausedStep) Execute(ctx context.Context, input *PauseInput, sta
...
 		}
 		latestActor.Status = ateapipb.Actor_STATUS_PAUSED
-		// TODO(dberkov) - what if we still don't know the node name? Maybe move to CRASHED status?
 		if nodeName == "" {
-			slog.Warn("Node name not found during finalize pause", "actor", input.ActorName)
+			// Without a node name we cannot record where the local snapshot lives,
+			// so the actor can never be resumed (findFreeWorker would search for a
+			// worker on an unknown node forever). Crash it instead of leaving it
+			// stuck in PAUSED.
+			slog.Warn("Node name not found during finalize pause, crashing actor", "actor", input.ActorName)
```
zoez7: Let's use `slog.ErrorCtx` instead?

zoez7 (dismissed):

cmd/ateapi/internal/controlapi/workflow_pause.go
mesutoezdil: thx zoe, done

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -202,19 +205,24 @@ func (s *FinalizePausedStep) Execute(ctx context.Context, input *PauseInput, sta
...
+			// stuck in PAUSED.
+			slog.ErrorContext(ctx, "Node name not found during finalize pause, crashing actor", "actor", input.ActorName)
+			latestActor.Status = ateapipb.Actor_STATUS_CRASHED
 		}
 		// TODO(dberkov) - what if InProgressSnapshot is empty? That shouldn't be possible.
 		if latestActor.InProgressSnapshot != "" {
+			localInfo := &ateapipb.LocalSnapshotInfo{
+				SnapshotPrefix: latestActor.InProgressSnapshot,
+			}
+			if nodeName != "" {
```
dberkov: the `nodeName != ""` no more relevant, since it is already validated above

mesutoezdil: done

cmd/ateapi/internal/controlapi/workflow_pause.go
mesutoezdil: it should be a separete issue i think,i dont solve it in this pr

zoez7: +1 @dberkov I think you described a different bug than the one this PR is solving

I think this PR solves a idempotency problem that we don't handle today.
1. CallAteletPauseStep succeeded, but we havn't called FinalizePausedStep yet
2. ate-apiserver restarts.
3. WorkerPod gets deleted.
4. Retry PauseActor.
Now we don't know which node the snapshot is on, we can only crash the actor. 

I think we should check this PR in as a temporary fix, but we also need to add node name as a field in the actor API, so Pause can be idempotent.

mesutoezdil:
@zoez7 could you take another look when you have a moment, since the last commit addressed the outstanding comments?

PR #332, opened 2026-06-26
api: open the sandboxClass enum for out-of-tree ateom-* backends

dims:
Replace the closed `+kubebuilder:validation:Enum=gvisor;microvm` on `SandboxConfig`/`WorkerPool`/`ActorTemplate` `.spec.sandboxClass` with a DNS-label `Pattern` + `MaxLength=63`. `gvisor`/`microvm` stay the built-in defaults, but a custom `ateom-*` backend can now define its own `sandboxClass` (and ship a matching `SandboxConfig` + `WorkerPool`) without forking substrate — admission no longer rejects the class. Regenerate CRDs; update the ActorTemplate validation test (custom `foobar` accepted, pattern-violating `Bad_Class` rejected).

- [ ] Tests pass

### DIFF
--- a/manifests/ate-install/generated/ate.dev_actortemplates.yaml
+++ b/manifests/ate-install/generated/ate.dev_actortemplates.yaml
@@ -186,9 +186,8 @@ spec:
                   4) How does the default get set and who sets it?
 
                   See Also: WorkerPool SandboxClass
-                enum:
-                - gvisor
-                - microvm
+                maxLength: 63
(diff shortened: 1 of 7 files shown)


BenTheElder:
Intentionally didn't do this while they won't work / aren't supported ... cc @thockin

dims:
@BenTheElder i have a branch where i copied ateom-microvm, renamed everything to ateom-foobar and got it working, so we can't say "they won't work" ... we can definitely say "aren't supported" until we have tests :)

PR #336, opened 2026-06-26
contract: neutralize asset-cache names, drop gVisor runsc_path (in-tree)

dims:
Two in-tree cleanups to the atelet to ateom contract. There is no package move in this PR.

Companion to #332 (open the `sandboxClass` enum) and #333 (declarative WorkerPool pod shape).

## What changed
1. Name the asset cache for what it holds, not for gVisor. `RunSCBinaryPath` becomes `CachedAssetPath` and `ValidateRunscHash` becomes `ValidateAssetHash`. The cache holds any backend's content addressed asset, not just gVisor's runsc. The on disk `runsc-<sha>` filename is left unchanged on purpose, because existing snapshots embed that absolute path. No behavior change.
2. Drop the gVisor specific `runsc_path` proto field (field 4, now `reserved`) from RunWorkload, Checkpoint, and Restore. gVisor reads its binary from `runtime_asset_paths["runsc"]` like every other backend. The wire contract now has zero backend specific fields.

## What is no longer here
An earlier revision of this PR also relocated the contract to `pkg/` (`ateompb` and `ateompath`) so an out of tree backend could `go get` it. That commit is dropped. The ateom API is a private, first party, in repo seam, so a public out of tree contract is not wanted. In repo backends import the contract from `internal/` directly. The public plugin contract for Agent Substrate is the atelet API, not the ateom API.

Verified locally: `go build ./...` and `go test` on the affected packages pass.

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -249,13 +249,13 @@ func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (*atele
 		return nil, err
 	}
 
-	// Tell ateom to start the workload. gVisor uses RunscPath; the micro-VM
-	// runtime uses the full RuntimeAssetPaths set.
+	// Tell ateom to start the workload. Every backend reads the assets it needs
+	// from RuntimeAssetPaths (gVisor reads "runsc"; the micro-VM runtime reads its
+	// cloud-hypervisor/kata-* set).
 	if _, err := client.RunWorkload(ctx, &ateompb.RunWorkloadRequest{
 		ActorTemplateNamespace: ns,
 		ActorTemplateName:      tmpl,
 		ActorId:                actorID,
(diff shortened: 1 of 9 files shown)


BenTheElder:
BTW I was chatting with @thockin about this and one key suggestion was that we might wind up with atelet as the extension mechanism. Undecided I think ...

dims:
> BTW I was chatting with [Tim Hockin (@thockin)](https://github.com/thockin) about this and one key suggestion was that we might wind up with atelet as the extension mechanism. Undecided I think ...

ACK @BenTheElder some of what i am doing is still useful to get rid of gvisor-isms i think. I am ok with going to `atelet` and have that conversation and the shape of it. I want to make sure what we have now is pluggable as is as well. it won't stop us from switching over when we decide to.

BenTheElder:
Ack ... I'm not sure we should publish to pkg/, given the guidelines for that ...

Issue #124, opened 2026-05-29
[Feature] Actor Identity


ahmedtd:
For Alpha, we need to give Actors access to OIDC JWTs and SPIFFE certificates issued by substrate.  We already have the broker RPCs implemented, but we have the following open items:

* How do we expose the credentials into the Actor (filesystem mounts, metadata server, vsock)?
* Formalize the claims in each credential.
* Make atelet/ateom use the actor credentials for snapshot storage / retrieval.
* (For GCP) Prove out federation of the substrate IDP back to GCP, and demonstrate how it can be used against GCS.

## Basic design

The substrate control plane includes a broker API, where a caller can exchange a K8s-layer credential (service account JWT or service account certificate) for a substrate JWT / certificate.

The substrate credential will carry at least the following claims:
* Actor Template Namespace
* Actor Template Name
* Actor ID

These credentials will be federatable, and so should be able to be used against GCS, S3, and other cloud provider services.  They could also be used for actor-to-actor authentication, and general actor-to-service authentication.

## (GCP-specific) Federation

Both OIDC JWTs and SPIFFE certificates can be federated into GCP IAM using Workload Identity Federation.

While it will not be scalable to create and delete per-actor IAM policies for each actor, it should be possible to use IAM conditions to write a single policy that authorizes every actor to access their own data.  For example, "all callers from substrate issuer X have storage/object.admin on GCS bucket Y, as long as the object path starts with `actors/${ate.dev/actor-id}`".

EItanya:
This design makes a ton of sense to me. I have a quick question about the `broker API` you mentioned above.

Do you envision the actors being able to reach the control-plane directly, or the ateom having some sort of metadata service that the actor can access from within it's isolated space.

ahmedtd:
When we originally wrote the broker, we were still using an "exterior gVisor" model, where we had the actors running directly as pods under a gVisor runtimeclass.  In that model, we had the actor workloads directly call the broker and pass their K8s credential.

Now that we are on interior gVisor, I'm leaning towards the default solution being something similar to k8s projected token and podCertificate volumes, so that actor workloads read their credentials from the filesystem.  Ateom/atelet would handle fetching the actor tokens/certs, putting them into the actor filesystem, and keeping them refreshed.

However, I know we are also going to need to support a model where an egress proxy injects the credentials.

EItanya:
[...]

PR #338, opened 2026-06-26
Egress tunnel capture + agent identity

npolshakova:
POC for https://github.com/agent-substrate/substrate/issues/126

Based on design discussed in https://docs.google.com/document/d/1KmpIFu2gnqy9gp95wASgIo_vkJ_dA1DZckV8upET6bs/edit?usp=sharing

- [x] Tests pass

Summary:
This is a proof of concept egress capture path for actors. It introduces a reusable `internal/egress` package that:
- starts local capture listeners
- derives the CONNECT authority from HTTP `Host` or TLS SNI, and opens a CONNECT-style tunnel to a resolved PEP address.

The gVisor and microvm runtimes wire this into actor network setup by redirecting actor HTTP/HTTPS egress traffic to local capture ports. The Linux-only glue shared by both runtimes (SO_ORIGINAL_DST lookup, nftables redirect rules, identity construction) lives in `internal/ateomegress` so the two runtimes stay in sync. Agentgateway is used as the receiving proxy to prove that captured actor traffic reaches the tunnel endpoint.

Capture is enabled per actor rather than by a global switch: ate-api watches Gateways labeled `ate.dev/egress-pep`, resolves the best-matching PEP for each actor at resume time (actor-specific > atespace-scoped > global), sends the PEP address down through atelet to ateom, and records the resolved address on the Actor status(`egress_pep_address`) so the binding is observable to the user. 
NOTE: Currently, an actor with no matching PEP resolves to an empty address and runs with capture off.

Notes:
- Tested the gvisor setup and microvm setups in kind. Can split these up into separate chunks to make it easier to review, but this shows the shared code.
- Agent identity is currently passed as unsigned metadata headers:
   - `x-ate-actor-id`
   - `x-ate-atespace`
   - `x-ate-actor-template`
   - `x-ate-actor-template-namespace`
   - `x-ate-original-destination`
   - `x-ate-connect-authority`
- Signed agent/actor identity is out of scope for this PR and should replace the current metadata headers approach once agent identity goes in https://github.com/agent-substrate/substrate/issues/124
- PEP selection is label-based on Gateways and trusts every labeled Gateway in the cluster, so Gateway RBAC is the security boundary (documented in `docs/egress-capture.md`). A first-class PEP binding/control plane is out of scope of this PR — agentgateway is just used to prove the proxy gets traffic from the tunnel.
- The PEP binding is a snapshot taken at resume: a running actor keeps its PEP until it is suspended/paused and resumed. ate-api resolves it from the Gateway's `status.addresses` (falling back to the `<gateway>.<namespace>.svc` convention) and only considers `Programmed=True` Gateways.
- Enforcement is best-effort/observability-grade for now: capture is not yet acknowledged back from ateom, so a worker running an older ateom image can ignore the PEP while the Actor still reports a binding (TODO left in `ateom.proto`). TLS stays end-to-end; agentgateway only routes the encrypted bytes after CONNECT.


cmd/ateom-gvisor/egress_capture.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/agent-substrate/substrate/internal/egresscapture"
+	"github.com/google/nftables"
+	"github.com/google/nftables/binaryutil"
+	"github.com/google/nftables/expr"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	egressCaptureHTTPPort   = uint16(15001)
+	egressCaptureHTTPSPort  = uint16(15002)
```
shashankram: Why do you need to capture HTTP vs HTTPS separately? A single listener could handle both and since SO_ORIGINAL_DST is used anyway to lookup the original port in `deriveConnectAuthority`.

npolshakova: Yeah I think that's a good clean up- originally I was testing http first and had one redirect per protocol.

cmd/ateom-microvm/egress_proxy.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/google/nftables"
+	"github.com/google/nftables/binaryutil"
+	"github.com/google/nftables/expr"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	egressCapturePort       = uint16(15001)
+	egressOriginalHTTPPort  = uint16(80)
+	egressOriginalHTTPSPort = uint16(443)
```
EItanya: Is there any specific reason we're only capturing these ports, and not just redirecting everything?

cmd/ateom-gvisor/egress_proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/google/nftables"
+	"github.com/google/nftables/binaryutil"
+	"github.com/google/nftables/expr"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	egressCapturePort       = uint16(15001)
+	egressOriginalHTTPPort  = uint16(80)
+	egressOriginalHTTPSPort = uint16(443)
```
EItanya: Same question about only supporting these 2 ports

cmd/ateom-microvm/net.go
```diff
@@ -249,6 +253,12 @@ func (s *AteomService) cleanupActorNetwork(ctx context.Context) error {
 		cleanupErr = errors.Join(cleanupErr, fmt.Errorf("while removing actor nftables rules: %w", err))
 		slog.WarnContext(ctx, "Failed to remove actor nftables rules; continuing actor netns cleanup", slog.Any("err", err))
 	}
+	if s.egressCapture != nil {
```
EItanya, line 256: Is there a time we would want no egress capture?

demos/counter/counter.go
```diff
@@ -61,6 +62,8 @@ func incrementFileCounter() int {
 	return counter
 }
 
+const defaultEgressURL = "https://httpbin.org/get"
```
EItanya: Why was this added to the counter, seems like we should have a different demo potentially

internal/egress/capture.go
EItanya, line 1: Should we name this package `egress`, or maybe `tunnel`?

internal/egresscapture/capture.go
```diff
@@ -0,0 +1,621 @@
...
+	Open(ctx context.Context, identity ActorIdentity, originalDst net.Addr, authority string) (io.ReadWriteCloser, error)
+}
+
+type TunnelTransportFactory func(Config) (TunnelTransport, error)
+
+var tunnelTransportFactories = map[string]TunnelTransportFactory{}
+
+func init() {
+	// Keep tunnel protocol support behind factories so additional transports
+	// such as HBONE can plug in without changing capture/listener logic.
```
EItanya: I know we discussed keeping the protocol generic, but do we need to start there, it makes the first merge more confusing to read, and I'm not clear if it's actually necessary

cmd/ateom-microvm/egress_proxy.go
npolshakova: So right now ateom derives the CONNECT authority from HTTP Host header for port 80 OR TLS SNI for port 443. For other ports, the current capture code just falls back to raw original destination IP:port. 

We could support trying to figure out the authority with other ports, but it would be a little more complicated because we'd need some sort of classifier: 
- try TLS SNI on any port
- try HTTP Host on any port
- fall back to recent per-actor DNS correlation
- if still can't figure it out, use IP:port

demos/counter/counter.go
npolshakova: It seemed more lightweight to just have a new endpoint under the counter demo for future testing, but we can make it a separate demo!

cmd/ateom-microvm/net.go
npolshakova: That's a good question- I think in general, egress capture should always be enabled but you can imagine scenarios where that doesn't make sense:
1. No PEP egress is setup. In this case I think traffic should still be captured by the ateom proxy, but not forwarded to any PEP
2. Flexibility for workloads using traffic we do not capture yet (UDP/QUIC, non-80/443 TCP if we don't want to do the [classifier](https://github.com/agent-substrate/substrate/pull/338#discussion_r3500749244), etc.)

internal/egresscapture/capture.go
npolshakova: Yeah I was hoping to show how it could be extended in the future, but let me clean it up.

cmd/ateom-microvm/net.go
npolshakova: @mikemorris This is kind of related to the discussion we had in slack. Like you brought up, if we're expecting _all_ traffic to always egress through a PEP, then the PEP registration can be a first-class API. Are there cases though where we don't want the traffic to get captured/egress through a PEP?

yanavlasov:
I agree with this approach overall. However there are problems that will need to be addressed:

1. there is no support for a fleet of gateways. I think right now the first server will always be picked.
2. Using CONNECT is quite limiting. Multiplexing TCP connections into a single H/2 connections can lead to issues at scale. Inner connections will not be loadbalanced well across multiple gateways. Also it complicates gateway's business logic if it need to enforce policies other than coarse destination based.

To improve the story with gateway pluggability, I propose to add support for PROXY protocol. This will make it easier to plug other gateways, like KAN or Envoy AI Gateway. This can be a substrate egress configuration option, which will make config more complex, and can be added as a followup (I volunteer myself for this if needed).

Eventually and option for terminating TLS at the proxy and re-encrypting using mTLS with agent identity could be added as well. This will make it even easier to plug other gateways as PEPs.

yanavlasov:
Eventually even the thin proxy should be pluggable. Operators will want telemetry, tracing, access logging from it. It will need to be able to load balance across a dynamically scaled fleet of PEPs, implement retries, health checking, etc. We do not want to recreate all this stuff in a custom go proxy. We can start with just a DNS name and go forwarder, but eventually it will need evolve and replicating functionality that already exists.

Issue #175, opened 2026-06-04
Harden interior net filtering rules


bowei:
We need to hardness the net filtering installed by default with Ateom to isolate Actors from the Node. For example, prevents access to instance metadata, Kubernetes node ports and the host network namespace.

EItanya:
This will be easier with #110

haiyanmeng:
Here are the things we can do to isolate an actor from the node:

1.   **Anti-Spoofing**: Drop any packet where the source IP is NOT `169.254.17.2` (the interior IP address assigned to the actor's sandbox).  
2.   **Drop Localhost**: Drop egress traffic destined for `127.0.0.0/8`.  
3.   **Drop Multicast/Broadcast**: Drop egress traffic destined for `224.0.0.0/4` and `255.255.255.255/32`.  
4.   **Drop Instance Metadata**: Drop traffic destined to `169.254.169.254`.  
5.   **Drop Kubernetes Node Ports**: Drop TCP and UDP traffic with destination ports falling within the configurable `nodePortStart-nodePortEnd` range.  
6.   **Drop Pod IP**: Drop all egress traffic destined directly for the worker pod's IP (`podIP`) to isolate the actor from `ateom`'s internal services.  
7.   **Drop Cluster CIDR**: If `clusterCIDR` is provided, drop all egress traffic destined for that network block to prevent lateral movement across the cluster.

@bowei , what do you think?

Even after Tunneling Gateway is implemented, we can still keep these `nftables` rules as a Defense in Depth safeguard. Right?

haiyanmeng:
https://github.com/agent-substrate/substrate/pull/488 added basic network policy for Ingress to Workers

bowei:
Dropping to P1 now that we have a path for Egress and atunnel.

EItanya:
Should we close this issues as a lot of work has been done since to change this whole system, including the recent rework from @BenTheElder

PR #340, opened 2026-06-26
Add basic network policy for Ingress to Workers

bowei:
Restrict Ingress to Workers and hence Actors to be only from the Router component.

This makes the atecontroller responsible for K8s level network policy

Part of hardening network connectivity for:

https://github.com/agent-substrate/substrate/issues/175

### DIFF
--- a/cmd/atecontroller/internal/controllers/networkpolicy_controller.go
+++ b/cmd/atecontroller/internal/controllers/networkpolicy_controller.go
@@ -1,0 +1,140 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controllers
+
+import (
+	"context"
+	"crypto/sha256"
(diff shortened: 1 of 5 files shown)


cmd/atecontroller/internal/controllers/networkpolicy_controller.go
```diff
@@ -0,0 +1,140 @@
...
+								WithMatchLabels(map[string]string{"kubernetes.io/metadata.name": "ate-system"})).
+							WithPodSelector(metav1ac.LabelSelector().
+								WithMatchLabels(map[string]string{"app": "atenet-router"})),
+					),
+			).
+			WithEgress(
+				networkingv1ac.NetworkPolicyEgressRule(),
+			))
+
+	// TODO: don't implement any Egress policy yet.
```
npolshakova, line 116: So for egress, the NetworkPolicy would  apply to ateom -> PEP and prevents worker pods from bypassing the PEP at the Kubernetes network layer?

bowei: Yes, we can't express the policy directly in Kubernetes NetworkPolicy ... nor do we want to, at least in the K8s's current incarnation. The way to understand this PR is that it is very targeted.

bowei: I avoided doing anything here right now as to wait for more progress on what the Egress APIs might look and also some of the rules may not be expressible at the k8s layer.

haiyanmeng:
@bowei , the failed e2e-test does not seem to relate to this PR. `/retest` does not seem to trigger a new e2e test run. Any suggestions?

haiyanmeng:
@bowei , let us close this PR and follow up on https://github.com/agent-substrate/substrate/pull/488

bowei:
Looks like HaiYan has taken this one, closing.

PR #347, opened 2026-06-29
Instrument both workercache and valkey store with metrics.

juli4n:
Add metrics to track the caching layer. Emit metrics to count pub/sub `PUBLISH` events (and failures), number of cached workers, amount of time the cache is disconnected, total resyncs (caused by a watch disconnection) and total relists.

### DIFF
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -253,7 +253,11 @@ func setupTest(t *testing.T, ns string) *testContext {
 	rdb := redis.NewClusterClient(&redis.ClusterOptions{
 		Addrs: []string{mr.Addr()},
 	})
-	persistence := ateredis.NewPersistence(rdb)
+	persistence, err := ateredis.NewPersistence(rdb)
+	if err != nil {
+		mr.Close()
+		t.Fatalf("ateredis.NewPersistence: %v", err)
+	}
 
 	// 2. Initialize Clientsets using global cfg
 	k8sClient, err := kubernetes.NewForConfig(cfg)
@@ -291,7 +295,12 @@ func setupTest(t *testing.T, ns string) *testContext {
 	substrateInformerFactory.WaitForCacheSync(ctx.Done())
 
 	// 4. Initialize Service
-	wc := workercache.New(persistence, 5*time.Minute)
+	wc, err := workercache.New(persistence, 5*time.Minute)
+	if err != nil {
+		cancel()
+		mr.Close()
+		t.Fatalf("workercache.New: %v", err)
+	}
 	if err := wc.Start(ctx); err != nil {
 		cancel()
 		mr.Close()
--- a/cmd/ateapi/internal/store/ateredis/ateredis.go
+++ b/cmd/ateapi/internal/store/ateredis/ateredis.go
@@ -56,6 +56,9 @@ import (
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
 	"github.com/redis/go-redis/v9"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 )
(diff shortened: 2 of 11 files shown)


cmd/ateapi/internal/controlapi/functional_test.go
```diff
@@ -253,7 +253,11 @@ func setupTest(t *testing.T, ns string) *testContext {
 	rdb := redis.NewClusterClient(&redis.ClusterOptions{
 		Addrs: []string{mr.Addr()},
 	})
-	persistence := ateredis.NewPersistence(rdb)
+	persistence, err := ateredis.NewPersistence(rdb)
+	if err != nil {
+		mr.Close()
```
EItanya, line 258: Is there a way to do `testing.Defer` or something similar so we don't need to remember to do this every time?

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -128,6 +140,9 @@ func (s *Persistence) publishWorkerEvent(ctx context.Context, eventType store.Wo
 	}
 	if err := s.rdb.Publish(ctx, workerPubSubChannel, payload).Err(); err != nil {
 		slog.ErrorContext(ctx, "worker event publish failed", slog.Any("err", err))
+		s.metricWorkerPubsubMessages.Add(ctx, 1, metric.WithAttributes(attribute.String("error.type", "_OTHER")))
```
EItanya, line 143: I'm a little confused on this. Is it a typical pattern to increment the same metric with an error attribute, or have a separate metric for failures?

juli4n: Not an otel expert, but I think this is the recommended way of reporting errors (https://opentelemetry.io/docs/specs/semconv/general/recording-errors/#recording-errors-on-metrics).

cmd/ateapi/internal/controlapi/functional_test.go
juli4n: That sounds like a good idea. We are already using this pattern throughout this file so I would probably tackle that as a separate PR?

cmd/ateapi/internal/workercache/workercache.go
```diff
@@ -43,17 +43,53 @@ type Cache struct {
...
-func New(store store.Interface, relistInterval time.Duration) *Cache {
-	return &Cache{
-		store:          store,
+func New(s store.Interface, relistInterval time.Duration) (*Cache, error) {
+	c := &Cache{
+		store:          s,
 		relistInterval: relistInterval,
 		workers:        make(map[string]*ateapipb.Worker),
 	}
+	m := otel.Meter("workercache")
```
botengyao, line 62: nit: it could be better to create a helper method for metric, and should we move this to `Start` method?

cmd/ateapi/internal/workercache/workercache.go
```diff
@@ -132,8 +173,9 @@ func (c *Cache) watchEvents(ctx context.Context, watch *store.WorkerWatch) {
 					return // context cancelled
 				}
 				c.ready.Store(true)
+				c.metricNotReadyDuration.Add(ctx, time.Since(notReadySince).Seconds())
```
krisztianfekete, line 176: This seems to fire only after resync succeeds, so you can't alert on "cache been not ready for 2+ minutes" while it's actually happening, which is kinda the whole point, right? Also, if the pod gets oomkilled, we never record the duration.

Could we also expose readiness as a gauge?

cmd/ateapi/internal/store/ateredis/ateredis.go
krisztianfekete: Putting `error.type` only on the failure series is 👌

krisztianfekete: nit: always `_OTHER` means we can't differentiate timeouts/etc from other outages. Could we add a classifier here (and in the relist one as well) to make this more actionable?

cmd/ateapi/internal/store/ateredis/ateredis_test.go
krisztianfekete, line 1: What do you think about also covering `persistence.worker.pubsub.messages` (success + `error.type`) and `cache.relists`?

cmd/ateapi/internal/workercache/workercache.go
```diff
@@ -43,17 +43,53 @@ type Cache struct {
...
-		store:          store,
+func New(s store.Interface, relistInterval time.Duration) (*Cache, error) {
+	c := &Cache{
+		store:          s,
 		relistInterval: relistInterval,
 		workers:        make(map[string]*ateapipb.Worker),
 	}
+	m := otel.Meter("workercache")
+	var err error
+	c.metricWorkerCount, err = m.Int64Gauge(
```
krisztianfekete, line 64: Can we make this be an `Int64ObservableGauge` instead? Right now, we record on every add/update/delete for a value the SDK just keeps as last value anyway. It'd also let us drop the ctx we added to `applyEvent`.

PR #349, opened 2026-06-29
fix: correct setup-gcp command in AGENTS.md

mesutoezdil:
`--all` flag does not exist in `setup-gcp`. The correct subcommand is `bootstrap`.

Verified with `go run ./tools/setup-gcp --help`.


mesutoezdil:
@HavenXia One-line fix in AGENTS.md, could you take a look when you get a chance?

PR #350, opened 2026-06-29
Make substrate's namespace, Service names and ServiceAccount names configurable

jjamroga:
# Description

**Substrate assumes the canonical install layout, and every deviation fails closed.** The namespace `ate-system`, the Services `api` / `atenet-router`, and the ServiceAccounts `atelet` / `atenet-router` are compiled-in constants. An install in a per-developer namespace, or under a deployment that prefixes resource names, breaks — and no failure points at the naming.

This series makes the namespace, the Service names, and the ServiceAccount names configurable. Every option defaults to the canonical value. A canonical install is byte-for-byte unaffected.

| Hardcoded assumption | Failure | What it looks like instead |
|---|---|---|
| atelet's namespace in ateapi's SPIFFE check | ateapi rejects every atelet | an mTLS handshake failure, not a naming error |
| atelet's namespace in the worker's broker check | no actor obtains a certificate | `credential broker is not atelet` |
| NetworkPolicy's ingress namespace | the CNI drops every request to the pool | a silent network fault |
| ateapi Service name and namespace in the client | the client cannot authenticate | `services "api" not found`, then `invalid bearer token` |
| Resource names in `hack/install-ate.sh` and the e2e harness | authorities land in the wrong namespace | suites die in preflight |

**A SPIFFE ID breaks on two axes.** It names a namespace and a ServiceAccount. Both were constants, and the failure surfaces as a rejected peer, not a missing object.

## Design notes

- **ate-controller passes the worker-side identities.** ateom already took `--atunnel-client-identity` as a flag but relied on its default. A new `--atunnel-broker-identity` flag alone would be inert: correct only where the constant was already correct. The controller knows the control plane's namespace, so it supplies both.
- **ateapi takes the whole expected identity, not a namespace.** The dialer and `ateletauth` used the namespace only to build one string. One place decides how atelet's identity is spelled.
- **`ateletauth` deduplicates without taking the dependency it was avoiding.** Its constants are duplicated rather than imported so the package does not depend on `controlapi` for three strings. `ateletdial` would make a third copy, so the strings move to `internal/installdefaults` instead — a leaf package of constants that imports only the standard library, so consuming it does not reintroduce the coupling the duplication was there to prevent.
- **The client reads environment variables, not flags.** It runs outside the cluster. It has no downward API and nothing to discover from.
- **Namespaces come from the downward API, not flags.** Every supported topology co-locates these components. Only Service and ServiceAccount names, which a deployment may legitimately rename, get flags.
- **`hack/install-ate.sh` and the e2e harness are in scope.** A relocated install cannot be bootstrapped or exercised without them. The script refuses `ATE_NAMESPACE` or `ATE_API_SERVICE_NAME` overrides on the manifest-applying subcommands, which would half-install: `manifests/ate-install/` names `ate-system` and `api` literally.

## Why this is one PR

The series is stacked, not parallel. Commit 1 creates `internal/installdefaults` and seeds seven imports; commit 2 adds fourteen more; every later commit builds on those constants. Split into separate PRs, each blocks on the previous merging and none reviews independently.

The guard test cannot land first. Applied to `main` it fails on eleven hardcoded literals across `ateletauth`, `controlapi/informer.go`, `networkpolicy_controller.go`, both `ateom` mains, `ateclient`, `ateletdial` and three e2e files. It is green only because the preceding commits removed them.

A partial series still fails closed. Each commit's audit turned up more places assuming `ate-system`, so landing commits 1-3 without 4-5 leaves a relocated install broken later and less legibly than before.

`CONTRIBUTING.md` covers this case: when the intermediate steps are not useful on their own, keep the change as one PR split into commits at logical break points. Review commit by commit, and preserve the commits on merge. If you would still rather split, the one defensible cut is by axis — relocation (commits 1-3) then renaming (commits 4-5) with the guard test rebased on top.

## Rebase notes (2026-09-15)

The series is rebased onto `main` at `2b255001`, forty-nine commits on from the previous base:

- `main` moved actor JWT and certificate minting out of `cmd/ateapi/internal/actoridentity` into `controlapi`, deleting the package and dropping its inline atelet authorization check in favor of the unified authorizer. This series previously threaded the expected atelet identity into that check; the check no longer exists, so that adaptation is gone. `ateletauth` survives with one caller, `workerservice.SetWorkerCapacity`, which still takes the configured identity.
- The same refactor replaced `BrokerConfig.ExpectedActorUID` with an `ActorAtespace` / `ActorName` / `ActorUID` triple. `AteletSPIFFEID` is additive and still required by `ateletdial.TLSConfig`.
- `ateletauth` (ateapi's side) and `ateletdial` (the worker's side) each re-declared the hardcoded SPIFFE ID. Both take the expected identity as a parameter.
- `main` deleted the atenet DNS subsystem. The series no longer touches it, and `installdefaults` carries no DNS Service name.
- The controller passes `--atunnel-broker-identity` only when it differs from the canonical default. An ateom old enough to predate the flag exits on it, and `docs/upgrade.md` keeps such a pool serving during a rolling upgrade. A relocated install gets the flag and necessarily runs an image that accepts it.
- `main` added `internal/e2e/collector_metrics.go` (agentgateway CI support), which addresses the router through the package-level `routerNamespace` and `routerService` that this series replaces. Its two call sites become `SystemNamespace()` and `ResourceName("atenet-router")`, matching `router_client.go` and `statusz.go`. A reviewer diffing against the older base sees those call sites move; nothing else in that file changes.
- The guard test's allowlist entry for the two inert `ate-system` literals follows the code from `actoridentity.go` to `controlapi/actor.go`. The refactor re-introduced exactly the class of constant this series removes, so the guard earns its place.

# Testing

- `go test -race ./...` passes. `make verify` passes boilerplate, codegen, go-modules, gofmt, golangci-lint, kube-api-linter, licenses, metrics and postgresql-migrations. `proto-fmt` needs `clang-format`, which is absent locally; this series touches no `.proto` files.
- Every commit builds and vets individually, via `git rebase -x 'go build ./... && go vet ./...'`.
- On a fresh kind cluster, all twelve e2e suites pass: `capabilities`, `combinedvolumes`, `demo`, `egressauthz`, `egressmitm`, `example`, `identity`, `metrics`, `networking`, `networkpolicy`, `parking`, `sizing`. The `demo`, `metrics`, `networkpolicy`, `parking` and `networking` suites need the `--deploy-demo-counter` and `--deploy-demo-egress` fixtures, and the MITM path needs `--deploy-demo-egress-mitm`; without them they fail on `actor template not found`, which reads as a control-plane fault rather than a missing fixture.
- With the sdsmint egress gateway (`--deploy-atenet --experimental-use-sdsmint`, `E2E_EGRESS_MITM=1`), `TestActorEgressMITMTrust` and `TestActorEgressHTTPSByHostnameMITM` both pass, and the `networking` suite is green at 25 passed. The two modes are mutually exclusive by design: `TestActorEgressHTTPSByHostnamePassthrough` covers the plain gateway and skips under MITM, and the two MITM tests skip without it. Both modes were run, so every case executed in one of them.
- The new unit tests use a relocated namespace and renamed ServiceAccounts. Reintroducing each hardcoding makes them fail.

One e2e caveat that predates this series and misleads: the `identity` suite calls `ReplaceEgressTrustPool`, which takes over the shared `egress-mitm-ca-pool` Secret, overwrites it and registers no cleanup. Any MITM test running afterwards fails with `certificate signed by unknown authority`, which reads as a broken interception path rather than a mutated fixture. Reinstall the gateway before running `egressmitm`.

# Additional Notes

**Credential contents stay canonical.** `controlapi/actor.go` still mints credentials naming `api.ate-system.svc`: the JWT issuer and the certificate's Issuer CN. Relying parties validate these, so changing them is a compatibility decision, not a lookup fix. The code carries a TODO to make the issuer a globally unique, OIDC-compliant name.

### DIFF
--- a/cmd/ateapi/internal/controlapi/dialer.go
+++ b/cmd/ateapi/internal/controlapi/dialer.go
@@ -47,7 +47,7 @@ var ErrNoAteletOnNode = errors.New("no atelet pod found on node")
 
 // The SPIFFE identity that atelet serving certs carry, as minted by the
 // podidentity signer (cmd/podcertcontroller/internal/podidentitysigner).
-// The namespace part is ateletNamespace, declared in informer.go.
+// The namespace part is the dialer's ateletNamespace.
 const (
 	trustDomainName = "cluster.local"
 	ateletSA        = "atelet"
@@ -75,14 +75,16 @@ func WithDialCredentials(build func(expectedPodUID string) (credentials.Transpor
 }
 
 // NewAteletDialer creates a new AteletDialer. clientBundlePath and serverCAPath
-// are used to build the per-atelet mTLS credentials used for every atelet connection.
-func NewAteletDialer(workerIndexer cache.Indexer, ateletIndexer cache.Indexer, clientBundlePath, serverCAPath string, opts ...DialerOption) *AteletDialer {
+// are used to build the per-atelet mTLS credentials used for every atelet
+// connection, and ateletNamespace is the namespace segment of the SPIFFE ID
+// those credentials expect on the atelet serving cert.
+func NewAteletDialer(workerIndexer cache.Indexer, ateletIndexer cache.Indexer, ateletNamespace, clientBundlePath, serverCAPath string, opts ...DialerOption) *AteletDialer {
 	d := &AteletDialer{
 		workerIndexer: workerIndexer,
 		ateletIndexer: ateletIndexer,
 		ateletConns:   newAteletConnCache(1024),
 		dialCredentials: func(expectedPodUID string) (credentials.TransportCredentials, error) {
-			tlsConfig, err := buildTLSConfig(clientBundlePath, serverCAPath, expectedPodUID)
+			tlsConfig, err := buildTLSConfig(ateletNamespace, clientBundlePath, serverCAPath, expectedPodUID)
 			if err != nil {
 				return nil, err
 			}
@@ -189,7 +191,7 @@ func (d *AteletDialer) DialForAteletOnNode(nodeName string) (*grpc.ClientConn, e
 	return ateletConn, nil
 }
 
-func buildTLSConfig(clientBundlePath, serverCAPath, expectedPodUID string) (*tls.Config, error) {
+func buildTLSConfig(ateletNamespace, clientBundlePath, serverCAPath, expectedPodUID string) (*tls.Config, error) {
 	trustDomain, err := spiffeid.TrustDomainFromString(trustDomainName)
 	if err != nil {
 		return nil, fmt.Errorf("while parsing trust domain %q: %w", trustDomainName, err)
--- a/cmd/ateapi/internal/controlapi/dialer.go
+++ b/cmd/ateapi/internal/controlapi/dialer.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/agent-substrate/substrate/internal/atelet"
 	"github.com/agent-substrate/substrate/internal/credbundle"
+	"github.com/agent-substrate/substrate/internal/installdefaults"
 	"github.com/agent-substrate/substrate/internal/substratex509"
 	"github.com/spiffe/go-spiffe/v2/bundle/x509bundle"
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
@@ -44,14 +45,6 @@ var ErrWorkerPodNotFound = errors.New("worker pod not found")
 // Distinct from ErrWorkerPodNotFound, which callers treat as crash-worthy;
 // this one is retryable.
 var ErrNoAteletOnNode = errors.New("no atelet pod found on node")
-
-// The SPIFFE identity that atelet serving certs carry, as minted by the
-// podidentity signer (cmd/podcertcontroller/internal/podidentitysigner).
-// The namespace part is the dialer's ateletNamespace.
-const (
-	trustDomainName = "cluster.local"
-	ateletSA        = "atelet"
-)
 
 // AteletDialer handles gRPC connections to Atelet pods.
 type AteletDialer struct {
@@ -192,15 +185,15 @@ func (d *AteletDialer) DialForAteletOnNode(nodeName string) (*grpc.ClientConn, e
 }
 
 func buildTLSConfig(ateletNamespace, clientBundlePath, serverCAPath, expectedPodUID string) (*tls.Config, error) {
-	trustDomain, err := spiffeid.TrustDomainFromString(trustDomainName)
+	trustDomain, err := spiffeid.TrustDomainFromString(installdefaults.AteletTrustDomain)
 	if err != nil {
-		return nil, fmt.Errorf("while parsing trust domain %q: %w", trustDomainName, err)
+		return nil, fmt.Errorf("while parsing trust domain %q: %w", installdefaults.AteletTrustDomain, err)
 	}
 	bundle, err := x509bundle.Load(trustDomain, serverCAPath)
 	if err != nil {
 		return nil, fmt.Errorf("while loading CA bundle from %s: %w", serverCAPath, err)
 	}
-	expectedID, err := spiffeid.FromSegments(trustDomain, "ns", ateletNamespace, "sa", ateletSA)
+	expectedID, err := spiffeid.FromSegments(trustDomain, "ns", ateletNamespace, "sa", installdefaults.AteletServiceAccount)
 	if err != nil {
 		return nil, fmt.Errorf("while building expected atelet SPIFFE ID: %w", err)
 	}
--- a/cmd/ateapi/internal/controlapi/dialer.go
+++ b/cmd/ateapi/internal/controlapi/dialer.go
@@ -69,15 +69,15 @@ func WithDialCredentials(build func(expectedPodUID string) (credentials.Transpor
 
 // NewAteletDialer creates a new AteletDialer. clientBundlePath and serverCAPath
 // are used to build the per-atelet mTLS credentials used for every atelet
-// connection, and ateletNamespace is the namespace segment of the SPIFFE ID
-// those credentials expect on the atelet serving cert.
-func NewAteletDialer(workerIndexer cache.Indexer, ateletIndexer cache.Indexer, ateletNamespace, clientBundlePath, serverCAPath string, opts ...DialerOption) *AteletDialer {
+// connection, and ateletSPIFFEID is the identity those credentials expect on
+// the atelet serving cert.
+func NewAteletDialer(workerIndexer cache.Indexer, ateletIndexer cache.Indexer, ateletSPIFFEID, clientBundlePath, serverCAPath string, opts ...DialerOption) *AteletDialer {
 	d := &AteletDialer{
 		workerIndexer: workerIndexer,
 		ateletIndexer: ateletIndexer,
 		ateletConns:   newAteletConnCache(1024),
 		dialCredentials: func(expectedPodUID string) (credentials.TransportCredentials, error) {
-			tlsConfig, err := buildTLSConfig(ateletNamespace, clientBundlePath, serverCAPath, expectedPodUID)
+			tlsConfig, err := buildTLSConfig(ateletSPIFFEID, clientBundlePath, serverCAPath, expectedPodUID)
 			if err != nil {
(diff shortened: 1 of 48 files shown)


cmd/ateapi/main.go
```diff
@@ -66,6 +66,8 @@ var (
 	workerpoolCACerts   = pflag.String("workerpool-ca-certs", "", "The file that contains the CA for verifying workerpool client certificates.")
 
 	showVersion = pflag.Bool("version", false, "Print version and exit.")
+
+	ateletNamespace = pflag.String("atelet-namespace", controlapi.DefaultAteletNamespace, "Namespace where atelet pods run. Override when the deployment runs atelet in a namespace other than the default.")
```
EItanya: Wouldn't a good default behavior here be to inject the Namespace via the downward API and use that?

cmd/atenet/internal/dns/dns.go
```diff
@@ -33,10 +33,15 @@ import (
...
+// Default resource names match the canonical install manifests in
+// manifests/ate-install/ (atenet-router and dns Services in the
+// ate-system namespace). Deployments that use a different namespace or
+// rename these Services must pass the actual values via the matching
+// flags on the dns command.
 const (
-	// serviceName is the name of the CoreDNS service.
-	serviceName     = "dns"
-	systemNamespace = "ate-system"
+	DefaultSystemNamespace   = "ate-system"
```
EItanya: This is defined here and above in `main.go` What do you think about a constant file or package?

cmd/atenet/internal/router/router.go
```diff
@@ -63,12 +63,22 @@ func init() {
 	utilruntime.Must(v1alpha1.AddToScheme(scheme))
 }
 
+// DefaultRouterServiceName matches the atenet-router Service name in the
+// canonical install manifests under manifests/ate-install/. Deployments
+// that rename the Service must pass the actual name via
+// --router-service-name.
+const DefaultRouterServiceName = "atenet-router"
```
EItanya: This is also defined twice

EItanya (commented):
Functionality makes sense. There are some duplicated constants and some env behavior I think we could take advantage of

cmd/ateapi/main.go
jjamroga: Makes sense, removing this flag and we're going to use POD_NAMESPACE

BenTheElder:
Please don't put links to github issues / PRs into git commits, this is specified in our AGENTS.md.

It makes the github event history for the page absurd (see above), you can't really read the comment stream because there's pages of github pingbacks.

jjamroga:
> Please don't put links to github issues / PRs into git commits, this is specified in our AGENTS.md.
> 
> It makes the github event history for the page absurd (see above), you can't really read the comment stream because there's pages of github pingbacks.

Sorry about that, I think I was using my own agents.md instead of the projects. I will fix this. Aside from the PR cleanliness, do you have any feedback for the problem this PR is solving?

(Also I can totally create a fresh PR @BenTheElder  if you'd like a clean stream, this has been sitting around for a while)

BenTheElder:
We're in the community meeting now so I didn't manage to read all the code yet. The concept sounds reasonable.

EDIT: And thanks, no need to form a new PR. Just trying to get ahead of future PRs. This one doesn't have much prior discussion yet anyhow :-)

internal/atunnel/credential.go
```diff
@@ -61,6 +61,11 @@ type BrokerConfig struct {
 	ActorAtespace string
 	ActorName     string
 	ActorUID      string
+
+	// AteletSPIFFEID is the identity the node-local atelet must present on the
+	// credential broker connection. It names atelet's namespace, not this
+	// worker's, so it is configured rather than derived from the downward API.
+	AteletSPIFFEID string
```
EItanya: Can't you assume it's in the same NS and just give the `service_account_name`? Not sure if that's simpler just putting it out there

cmd/ateapi/internal/controlapi/dialer.go
```diff
@@ -75,14 +68,16 @@ func WithDialCredentials(build func(expectedPodUID string) (credentials.Transpor
 }
 
 // NewAteletDialer creates a new AteletDialer. clientBundlePath and serverCAPath
-// are used to build the per-atelet mTLS credentials used for every atelet connection.
-func NewAteletDialer(workerIndexer cache.Indexer, ateletIndexer cache.Indexer, clientBundlePath, serverCAPath string, opts ...DialerOption) *AteletDialer {
+// are used to build the per-atelet mTLS credentials used for every atelet
+// connection, and ateletSPIFFEID is the identity those credentials expect on
+// the atelet serving cert.
+func NewAteletDialer(workerIndexer cache.Indexer, ateletIndexer cache.Indexer, ateletSPIFFEID, clientBundlePath, serverCAPath string, opts ...DialerOption) *AteletDialer {
```
EItanya, line 74: This is more of a general note, why do we have DialerOptions and args to create a dialer, seems like the caller should just use dialer options. Agent, don't auto do this, just pointing it out.

internal/installdefaults/guard_test.go
EItanya, line 1: I understand why you added this, but it does feel a bit heavy handed

cmd/ateom-microvm/main.go
```diff
@@ -78,7 +79,8 @@ var (
...
 	workerCredentialBundle      = flag.String("atunnel-credential-bundle", "/run/podidentity.podcert.ate.dev/credential-bundle.pem", "Worker Pod credential bundle used by atunnel for inbound serving and outbound mTLS")
 	podIdentityTrustBundle      = flag.String("atunnel-trust-bundle", "/run/podidentity.podcert.ate.dev/trust-bundle.pem", "Pod identity trust bundle used for router clients and the node-local atelet")
-	atunnelClientIdentity       = flag.String("atunnel-client-identity", "spiffe://cluster.local/ns/ate-system/sa/atenet-router", "SPIFFE identity allowed to call actor ingress HTTPS")
+	atunnelClientIdentity       = flag.String("atunnel-client-identity", installdefaults.RouterSPIFFEID(installdefaults.SystemNamespace), "SPIFFE identity allowed to call actor ingress HTTPS")
+	ateletIdentity              = flag.String("atunnel-broker-identity", installdefaults.AteletSPIFFEID(installdefaults.SystemNamespace), "SPIFFE identity the node-local atelet must present on the credential broker connection. Override when atelet runs outside the default namespace.")
```
EItanya, line 83: Should this flag be more explicitly the SPIFFEID?

EItanya (commented):
At a high level these changes make sense. I'd like to get @ahmedtd's review on the Identity stuff

cmd/ateom-microvm/main.go
jjamroga: I agree with you here, I think that `atunnel-client-spiffe-id` would be better since it explains the type, but this argument already existed and I'm afraid of changing the convention since that would be breaking.

internal/installdefaults/guard_test.go
jjamroga: I can remove this

jjamroga: Definitely heavy handed, but in updating this branch with main and resolve merge conflicts (twice), it caught a few new instances of hardcoded namespace + spiffe ID

```
Rebase 1 — onto 26c38616 (Sep 15, 42 commits)

Guard failed, two literals:

cmd/ateapi/internal/controlapi/actor.go:585:11: "https://api.ate-system.svc"
cmd/ateapi/internal/controlapi/actor.go:685:17: "api.ate-system.svc.cluster.local"
---
Rebase 2 — onto 2b255001 (Sep 18, 7 commits)

Guard passed. No catch.
---
Rebase 3 — onto d0d85c39 (today, 43 commits)

Guard failed, one literal:

cmd/credential-provider/kubernetes-secrets/main.go:50:26:
  "spiffe://cluster.local/ns/ate-system/sa/atenet-egress"
```

If you want I will remove this, but until it's a habit for reviewers to avoid hardcoded namespaces + spiffe ids, I think this test has a purpose.

ahmedtd (dismissed):

EItanya (approved):
Giving the LGTM post merge which @ahmedtd already LGTM'd

PR #351, opened 2026-06-29
Add initial version of a style guide for substrate control plane.

juli4n:
Add an initial version of a style guide for the substrate public API. It uses Google's AIP as a starting point, and clarified when we intentionally deviate from it.

### DIFF
--- a/docs/api-style-guide.md
+++ b/docs/api-style-guide.md
@@ -1,0 +1,457 @@
+# Substrate gRPC API Style Guide
+
+This document is the authoritative style guide for Substrate public APIs. Today this only includes `ate-apiserver` API.
+
+This guide is derived from [Google's AIP (API Improvement Proposals)](https://google.aip.dev/) and adopts the large majority of its conventions. Divergences are called out explicitly with rationale.
+
+---
+
+## 1. Resource-Oriented Design
+
+*Follows [AIP-121](https://google.aip.dev/121).*
+
+APIs are structured around **resources** (nouns) and a small set of standard **methods** (verbs). Standard methods are Get, List, Create, Update, and Delete. **Custom methods** handle operations that don't fit these patterns (e.g., Suspend, Resume, Pause for Actors).
+
+Rules:
+- Every primary noun the API exposes is a resource.
+- Standard methods are strongly preferred. Custom methods are the exception, not the norm.
+- The resource schema must be identical across all standard methods that reference it (i.e., Get, Create, and Update all return the same `Actor` message).
+
+---
+
+## 2. Resource Naming and Identity
+
+**Diverges from [AIP-122](https://google.aip.dev/122).**
+
+AIP-122 identifies resources by a single opaque path string (e.g., `publishers/123/books/les-miserables`). Substrate uses a **two-field identity** instead: an `atespace` (namespace) and a `name`. This is analogous to how Kubernetes identifies objects and avoids the ambiguity of parsing hierarchical path strings.
+
+Resources are either **atespace-scoped** or **global-scoped**. Scope is a fixed property of the resource type, not of individual instances.
+
+### 2.1 Atespace-scoped resources
+
+Atespace-scoped resources belong to an atespace. Their identity is `(atespace, name)`, unique within the resource type. The first two fields **must** be:
+
+```proto
+message Actor {
+  // The atespace this actor belongs to.
+  string atespace = 1;
+  // The name of this actor, unique within its atespace.
+  string name = 2;
+
+  // ... other fields
+}
+```
+
+- `atespace` is the Substrate namespace for the resource.
+- `name` is the resource name within the atespace. It is not a path; it is a short identifier.
+
+### 2.2 Global-scoped resources
+
+Some resources are global across the entire deployment and do not belong to any atespace. For these, the identity is `name` alone. The first field **must** be:
+
+```proto
+message Worker {
+  // The name of this worker, globally unique (e.g. the node name).
+  string name = 1;
+
+  // ... other fields
+}
+```
+
+- There is no `atespace` field as the resource is global-scoped.
+- `name` must be globally unique within the resource type, across all atespaces.
+
+### 2.3 Character constraints
+
+Both `atespace` and `name` must conform to [DNS-1123 label](https://tools.ietf.org/html/rfc1123) syntax:
+
+- Lowercase alphanumeric characters and hyphens only.
+- Must start with a lowercase alphanumeric character.
+- Must end with a lowercase alphanumeric character.
+- Maximum 63 characters.
+- Regex: `^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$`
+
+### 2.4 `{Resource}Ref` — compound identity type
+
+Every atespace-scoped resource type **must** have a corresponding `{Resource}Ref` message that bundles `atespace` + `name` as a single, typed unit:
+
+```proto
+message ActorRef {
+  string atespace = 1;
+  string name     = 2;
+}
+```
+
+Use `{Resource}Ref` in places where you need to reference a resource. For example:
+
+**1. Request messages** — to identify which specific resource to act on:
+
+```proto
+message GetActorRequest {
+  ActorRef actor = 1;
+}
+
+message DeleteActorRequest {
+  ActorRef actor   = 1;
+  int64     version = 2;
+}
+```
+
+**2. Cross-resource references** — when one resource's fields refer to another atespace-scoped resource:
+
+```proto
+message Actor {
+  string atespace = 1;
+  string name     = 2;
+
+  // The ActorTemplate this actor was derived from.
+  ActorTemplateRef actor_template = 3;
+
+  // ... other fields
+}
+```
+
+The field name is the logical name of the reference (e.g., `actor_template`), not `actor_template_name` or `actor_template_ref`.
+Note that this assumes that ActorTemplates are also resources in the substrate gRPC API (not in KRM).
+
+**Global-scoped resources** have no atespace, so no wrapper type is needed — use a plain `string {resource}_name` field:
+
+```proto
+message Actor {
+  // worker_name references the global-scoped Worker assigned to this actor.
+  string worker_name = 10;
+}
+```
+
+* Do not embed the full resource message as a reference field.
+- Do not use a single combined string like `"atespace/name"`. Callers would have to parse it.
+
+---
+
+## 3. Standard Methods
+
+The following sections cover each standard method. The primary adaptation from AIP-13x is how resources are identified in requests: two fields (`atespace` + `name`) instead of a single `name` path string.
+
+### 3.1 Get
+
+*Follows [AIP-131](https://google.aip.dev/131).*
+
+```proto
+rpc GetActor(GetActorRequest) returns (Actor) {}
+
+message GetActorRequest {
+  ActorRef actor = 1;
+}
+```
+
+Rules:
+- RPC name **must** begin with `Get` followed by the singular resource name.
+- Request message name **must** match the RPC name with a `Request` suffix.
+- Response **must** be the resource itself — not a `GetActorResponse` wrapper.
+- Request **must** identify the resource with a single `{Resource}Ref` field (atespace-scoped) or `string name` (global-scoped).
+- If the resource does not exist: return `NOT_FOUND`.
+- If the caller lacks permission: return `PERMISSION_DENIED` (checked before existence).
+
+### 3.2 List
+
+*Follows [AIP-132](https://google.aip.dev/132).*
+
+```proto
+rpc ListActors(ListActorsRequest) returns (ListActorsResponse) {}
+
+message ListActorsRequest {
+  // The atespace to list actors from.
+  string atespace = 1;
+
+  // Maximum number of actors to return. The server may return fewer.
+  // If unspecified, defaults to a server-chosen value.
+  // The maximum value is 1000; values above 1000 are coerced to 1000.
+  int32 page_size = 2;
+
+  // Pagination token from a previous ListActors response.
+  // Omit or leave empty for the first request.
+  string page_token = 3;
+}
+
+message ListActorsResponse {
+  repeated Actor actors = 1;
+
+  // Pagination token for the next page.
+  // Empty if this is the last page.
+  string next_page_token = 2;
+}
+```
+
+Rules:
+- RPC name **must** begin with `List` followed by the **plural** resource name.
+- Both the request and response message names **must** match the RPC name with `Request`/`Response` suffixes. (Unlike Get/Create/Update, List responses are not the resource itself.)
+- The `page_size` and `page_token` fields **must be defined** on every List request message (even though the token will be empty on the first call).
+- `next_page_token` **must** be present on every List response message. It **must** be empty when there are no further pages.
+- The repeated resource field **must** use the plural form of the resource name (e.g., `actors`, not `actor`).
+- If a user provides a `page_size` above the maximum, coerce it silently. If a user provides a negative value, return `INVALID_ARGUMENT`.
+- List **may** accept an empty `atespace` to list across all atespaces, if the caller has sufficient permission. Document this clearly if supported.
+
+### 3.3 Create
+
+*Adapted from [AIP-133](https://google.aip.dev/133).*
+
+```proto
+rpc CreateActor(CreateActorRequest) returns (Actor) {}
+
+message CreateActorRequest {
+  // The actor to create.
+  // actor.atespace and actor.name together specify the resource's identity
+  // and must both be set by the caller.
+  Actor actor = 1;
+}
+```
+
+Rules:
+- RPC name **must** begin with `Create` followed by the singular resource name.
+- Response **must** be the resource itself — not a `CreateActorResponse` wrapper.
+- The resource body is the only field in the request. There are no separate top-level `atespace` or `{resource}_id` fields.
+- `actor.atespace` and `actor.name` are **required** and caller-specified. The server does not generate them.
+- If a resource already exists with the same `(atespace, name)`: return `ALREADY_EXISTS`. If the caller lacks permission to observe the conflicting resource: return `PERMISSION_DENIED`.
+
+**Divergence from AIP-133:** AIP-133 separates `parent` + `{resource}_id` from the resource body because AIP-122 makes the resource `name` field output-only (constructed by the server from the parent path). In Substrate's model, `atespace` and `name` are directly caller-specified identity fields on the resource, so duplicating them at the top level of the request adds no information and creates ambiguity about which one wins. The embedded resource is the single source of truth for identity on create.
+
+### 3.4 Update
+
+*Follows [AIP-134](https://google.aip.dev/134). Diverges on `update_mask` requirement and `*` support.*
+
+Updates use a **partial-update model** (equivalent to HTTP PATCH). The mask is always required.
+
+```proto
+rpc UpdateActor(UpdateActorRequest) returns (Actor) {}
+
+message UpdateActorRequest {
+  // The actor to update.
+  // The actor's atespace and name fields identify which resource to update.
+  Actor actor = 1;
+
+  // The set of fields to update. Required.
+  //
+  // Field paths are relative to the Actor message (e.g., "worker_selector").
+  google.protobuf.FieldMask update_mask = 2;
+}
+```
+
+Rules:
+- RPC name **must** begin with `Update` followed by the singular resource name.
+- Response **must** be the resource itself — not an `UpdateActorResponse` wrapper.
+- `update_mask` **must** be of type `google.protobuf.FieldMask` and **must** be named `update_mask`.
+- `update_mask` is **required**. An absent or empty mask **must** return `INVALID_ARGUMENT`.
+- The special value `*` is **not supported**. Clients must enumerate the exact fields to update.
+- The resource's `atespace` and `name` identify the resource to update; they are not themselves updatable.
+- If the resource does not exist: return `NOT_FOUND`.
+
+**Divergence from AIP-134:** AIP-134 makes `update_mask` optional (omission implies updating all populated fields) and requires support for `*`. Substrate requires an explicit mask.
+
+### 3.5 Delete
+
+*Follows [AIP-135](https://google.aip.dev/135).*
+
+```proto
+rpc DeleteActor(DeleteActorRequest) returns (google.protobuf.Empty) {}
+
+message DeleteActorRequest {
+  ActorRef actor   = 1;
+  int64     version = 2; // optional freshness guard; 0 = skip check
+}
+```
+
+Rules:
+- RPC name **must** begin with `Delete` followed by the singular resource name.
+- Response **must** be `google.protobuf.Empty` (no `DeleteActorResponse` wrapper).
+- Request **must** identify the resource with a `{Resource}Ref` field (atespace-scoped) or `string name` (global-scoped).
+- If the resource does not exist: return `NOT_FOUND`.
+- If the caller lacks permission: return `PERMISSION_DENIED` (checked before existence).
+
+---
+
+## 4. Custom Methods
+
+*Follows [AIP-136](https://google.aip.dev/136).*
+
+Custom methods are for operations that don't map cleanly to CRUD: lifecycle transitions (Suspend, Resume, Pause), long-running actions, or commands with side effects that standard Update semantics would misrepresent.
+
+```proto
+rpc SuspendActor(SuspendActorRequest) returns (Actor) {}
+
+message SuspendActorRequest {
+  ActorRef actor = 1;
+}
+```
+
+Rules:
+- RPC name **must** be a verb phrase: `{Verb}{Resource}` (e.g., `SuspendActor`, `ResumeActor`).
+- Request message name **must** match the RPC name with a `Request` suffix.
+- The request **must** identify the target resource using a `{Resource}Ref` field (atespace-scoped) or `string name` (global-scoped).
+- The response **should** return the updated resource when the operation mutates it (e.g., all Actor lifecycle methods return the updated `Actor`).
+
+---
+
+## 5. Field Naming
+
+*Follows [AIP-140](https://google.aip.dev/140) and [AIP-149](https://google.aip.dev/149).*
+
+- Field definitions in proto files **must** use `lower_snake_case`.
+- Boolean fields **must** omit the `is_` prefix: use `disabled`, not `is_disabled`. Exception: use `is_` when the bare word would be a reserved keyword in common languages.
+- Repeated fields **must** use the plural noun form: `containers`, not `container`.
+- Non-repeated fields **must** use the singular form: `container`, not `containers`.
+- Field names **must** be nouns, not verbs: `worker_selector`, not `select_workers`.
+- Use standard abbreviations where well-established: `config`, `spec`, `id`, `info`, `stats`.
+- Adjectives come before the noun: `suspended_actors`, not `actors_suspended`.
+- Avoid prepositions in field names: `error_reason`, not `reason_for_error`.
+- Do not use `name` for any purpose other than a resource's own name field (see section 2.1). Use specific names: `display_name`, `actor_template_name`, etc.
+
+### 5.1 Field presence (`optional`)
+
+Use the `optional` keyword on a scalar field **only** when null and the zero value (`false`, `0`, `""`) are semantically distinct for that field's meaning. Do not use `optional` universally or as a workaround for update semantics — the required `update_mask` (§3.4) handles that.
+
+```proto
+// Only if "unrated" is meaningfully different from "rated 0":
+optional int32 priority = 5;
+
+// No optional needed — false and unset mean the same thing here:
+bool cordoned = 6;
+```
+
+Because `update_mask` is required, the server always knows which fields the client intends to change. `optional` is reserved for cases where the resource itself has a three-state semantic (set-to-zero, set-to-nonzero, not-set).
+
+---
+
+## 6. Standard Fields
+
+*Follows [AIP-148](https://google.aip.dev/148) and [AIP-142](https://google.aip.dev/142).*
+
+Certain fields appear on nearly every resource and must use the exact names and types below. All standard fields are **output-only**: the server sets them and the client must not send them on create or update (the server ignores any value provided).
+
+The recommended field ordering within a resource message is: identity fields (`atespace`, `name`), then `uid`, then timestamps, then `version`, then resource-specific fields.
+
+```proto
+message Actor {
+  string atespace = 1;
+  string name     = 2;
+
+  // uid is a server-assigned, globally unique identifier for this resource.
+  // It is stable across renames and updates. Distinct from name.
+  string uid = 3;
+
+  // create_time is the time the resource was created.
+  google.protobuf.Timestamp create_time = 4;
+
+  // update_time is the time the resource was last updated by a user action.
+  google.protobuf.Timestamp update_time = 5;
+
+  // delete_time is set when the resource is soft-deleted.
+  // Only present on resources that support soft delete.
+  google.protobuf.Timestamp delete_time = 6;
+
+  // version is incremented on every mutation. See §7.
+  int64 version = 7;
+
+  // ... resource-specific fields
+}
+```
+
+### 6.1 `uid`
+
+- Type: `string`.
+- A server-assigned [UUID4](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)).
+- Stable across all mutations: updating or "renaming" a resource (changing its `name` in a future API version) does not change its `uid`.
+- Useful for correlation across logs, events, and audit trails where the resource `name` may not be available.
+- All public resources (`ate-apiserver`) **must** include `uid`. Internal resources **should** include it.
+
+### 6.2 `create_time`
+
+- Type: `google.protobuf.Timestamp`.
+- Records when the resource was created.
+- Set once at creation; never updated.
+- All public resources **must** include `create_time`.
+
+### 6.3 `update_time`
+
+- Type: `google.protobuf.Timestamp`.
+- Records when the resource was last modified by a user action (Create, Update, or a custom mutating method).
+- Updated on every mutation. Internal state changes made by the system (e.g., a scheduler assigning a worker) **may** also update this field, but are not required to.
+- All public resources **must** include `update_time`.
+
+### 6.4 `delete_time`
+
+- Type: `google.protobuf.Timestamp`.
+- Only relevant for resources that support **soft delete** (marking as deleted without immediately purging).
+- Set when the resource is soft-deleted; absent (zero value) when the resource is live.
+- Resources that do not support soft delete **must not** include this field.
+- Substrate's current resources use hard delete; add `delete_time` only if soft delete is introduced for a specific resource type.
+
+---
+
+## 7. Resource Freshness and Optimistic Concurrency
+
+*Inspired by [AIP-154](https://google.aip.dev/154). Diverges in field name and type.*
+
+When two clients update the same resource concurrently, the second write may silently overwrite the first. Freshness validation lets a client prove it is operating on the state it thinks it is, so the server can reject stale writes.
+
+Substrate uses a field named **`version`** of type `int64` for this. AIP-154 uses an opaque `etag` string; we diverge for a concrete reason: Substrate maintains an in-memory worker cache that guards against applying stale watch events over newer cached state using a numeric `>=` comparison. An opaque string cannot serve this role — the ordering guarantee IS the implementation contract, so the field type should reflect it.
+
+The name `version` is intentional: it increments on every write (like Kubernetes's `resourceVersion`) and is a transparent, comparable integer (like Kubernetes's `generation`). It serves both roles in one field, so neither Kubernetes name fits cleanly.
+
+### 7.1 The `version` field
+
+`version` is a standard output-only field on every resource:
+
+```proto
+message Actor {
+  string atespace = 1;
+  string name     = 2;
+  string uid      = 3;
+  google.protobuf.Timestamp create_time = 4;
+  google.protobuf.Timestamp update_time = 5;
+  int64 version = 6;
+
+  // ... resource-specific fields
+}
+```
+
+- Type: `int64`.
+- Output-only: the server sets it. Starts at `1` on creation, incremented by `1` on every mutation.
+- **Monotonically increasing:** safe to compare numerically. A higher value is always newer.
+- Updated on every mutation — both user-visible changes and system-internal ones (e.g. the scheduler binding a worker).
+- The field **must** be named `version`.
+- All public resources **must** include `version`.
+
+### 7.2 Using `version` to guard writes
+
+A client that wants to guard against concurrent modification echoes back the `version` it last observed. The server rejects the request if the value no longer matches.
+
+**Update:** `version` rides in the embedded resource body:
+
+```proto
+// Client read actor at version 5, now updating:
+UpdateActorRequest {
+  actor: Actor {
+    atespace: "my-space"
+    name:     "my-actor"
+    version:  5            // guard: fail if server is not at 5
+    worker_selector: ...
+  }
+  update_mask: "worker_selector"
+}
+```
+
+**Delete and custom methods:** add an optional `int64 version` field to the request:
+
+```proto
+message DeleteActorRequest {
+  ActorRef actor  = 1;
+  // Optional. If non-zero, the deletion is rejected with ABORTED if the
+  // server's current version does not match.
+  int64 version   = 2;
+}
+```
+
+**Server behavior:**
+- If the client provides a non-zero `version` that does not match the server's current value: return `ABORTED`.
+- If the client omits `version` (zero value, the proto3 default): skip the freshness check and proceed.
+- The freshness check is always optional from the client's perspective.
--- a/docs/api-style-guide.md
+++ b/docs/api-style-guide.md
@@ -25,41 +25,36 @@ Rules:
 
 AIP-122 identifies resources by a single opaque path string (e.g., `publishers/123/books/les-miserables`). Substrate uses a **two-field identity** instead: an `atespace` (namespace) and a `name`. This is analogous to how Kubernetes identifies objects and avoids the ambiguity of parsing hierarchical path strings.
 
-Resources are either **atespace-scoped** or **global-scoped**. Scope is a fixed property of the resource type, not of individual instances.
+Resources are either **atespace-scoped** or **global-scoped**. Scope is a fixed property of the resource type, not of individual instances. For example, `Actor` resources
+are **atespace-scoped**, whereas `Atespace` resources are naturally **global-scoped**.
 
 ### 2.1 Atespace-scoped resources
 
-Atespace-scoped resources belong to an atespace. Their identity is `(atespace, name)`, unique within the resource type. The first two fields **must** be:
+* Atespace-scoped resources belong to an atespace. Their identity is `(atespace, name)`, unique within the resource type.
+* Global-scoped resources are global across the entire deployment and do not belong to any atespace. For these, the identity is `name` alone.
+
+In both cases, a `meta` field contains both `atespace` and `name`. For global resources, the `atespace` must always be empty.
 
 ```proto
 message Actor {
   // The atespace this actor belongs to.
+  ResourceMetadata meta = 1;
+
+  // ... other fields
+}
+
+
+message ResourceMetadata {
+  // The atespace this resource belongs to.
   string atespace = 1;
-  // The name of this actor, unique within its atespace.
+  // The name of this resource, unique within its atespace.
   string name = 2;
 
-  // ... other fields
+  // ... other common resource fields
 }
 ```
 
-- `atespace` is the Substrate namespace for the resource.
-- `name` is the resource name within the atespace. It is not a path; it is a short identifier.
-
-### 2.2 Global-scoped resources
-
-Some resources are global across the entire deployment and do not belong to any atespace. For these, the identity is `name` alone. The first field **must** be:
-
-```proto
-message Worker {
-  // The name of this worker, globally unique (e.g. the node name).
-  string name = 1;
-
-  // ... other fields
-}
-```
-
-- There is no `atespace` field as the resource is global-scoped.
-- `name` must be globally unique within the resource type, across all atespaces.
+- All resources in Substrate must have a `ResourceMetadata meta = 1` field to hold common fields, which includes both `atespace` and `name`.
 
 ### 2.3 Character constraints
 
@@ -74,28 +69,28 @@ A valid resource name must comply the following rules:
 
 Resource names are valid RFC-1123 DNS labels.
 
-### 2.4 `{Resource}Ref` — compound identity type
+### 2.4 `ObjectRef` — reference type
 
-Every atespace-scoped resource type **must** have a corresponding `{Resource}Ref` message that bundles `atespace` + `name` as a single, typed unit:
+The `ObjectRef` message represents a *pointer* to a Substrate resource.
 
 ```proto
-message ActorRef {
+message ObjectRef {
   string atespace = 1;
   string name     = 2;
 }
 ```
 
-Use `{Resource}Ref` in places where you need to reference a resource. For example:
+Use `ObjectRef` in places where you need to reference a resource. For example:
 
 **1. Request messages** — to identify which specific resource to act on:
 
 ```proto
 message GetActorRequest {
-  ActorRef actor = 1;
+  ObjectRef actor = 1;
 }
 
 message DeleteActorRequest {
-  ActorRef actor   = 1;
+  ObjectRef actor   = 1;
   int64     version = 2;
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+
+### 2.3 Character constraints
+
+Both `atespace` and `name` must conform to [DNS-1123 label](https://tools.ietf.org/html/rfc1123) syntax:
+
+- Lowercase alphanumeric characters and hyphens only.
+- Must start with a lowercase letter.
+- Must end with a lowercase alphanumeric character.
+- Maximum 63 characters.
+- Regex: `^[a-z]([a-z0-9-]{0,61}[a-z0-9])?$`
```
HavenXia: Should this be `[a-z0-9]` according to https://github.com/agent-substrate/substrate/blob/0ec7884218fd7d9a683998a2cd99b14ac9549c11/internal/resources/actor.go#L24

HavenXia (dismissed):

docs/api-style-guide.md
juli4n: Done

EItanya (commented):
What do you think about linking to this from the `agents.md` or a similar file so that agents are always aware to keep these rules in mind?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+AIP-122 identifies resources by a single opaque path string (e.g., `publishers/123/books/les-miserables`). Substrate uses a **two-field identity** instead: an `atespace` (namespace) and a `name`. This is analogous to how Kubernetes identifies objects and avoids the ambiguity of parsing hierarchical path strings.
+
+Resources are either **atespace-scoped** or **global-scoped**. Scope is a fixed property of the resource type, not of individual instances.
+
+### 2.1 Atespace-scoped resources
+
+Atespace-scoped resources belong to an atespace. Their identity is `(atespace, name)`, unique within the resource type. The first two fields **must** be:
+
+```proto
+message Actor {
```
thockin, line 39: should we do something like:

```
message AtespaceScopedMetadata {
    string atespace = 1;
    string name = 2;
    // future stuff
}
message GlobalScopedMetadata {
    // NOTE: tag 1 is reserved
    string name = 2;
    // future stuff
}
```

I am generally against schema sharing but in this case I think we want common metadata across types?  E.g. all the fields in sections 6 and 7...

thockin: Also, I'm getting "ate" fatigue - perhaps we should actually reconsider putting it in front of every word.  "Atespace" - "Space" ?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+  // ... other fields
+}
+```
+
+- There is no `atespace` field as the resource is global-scoped.
+- `name` must be globally unique within the resource type, across all atespaces.
+
+### 2.3 Character constraints
+
+Both `atespace` and `name` must conform to [DNS-1123 label](https://tools.ietf.org/html/rfc1123) syntax:
```
thockin: The referencing of DNS has become somewhat confusing in k8s, since DNS technically is case-insensitive.  I think we should instead define this as a "resource name" and simply say that all resources in agent substrate are named as per the following rules, which makes them compatible with RFC-1123 DNS labels.

In k8s we were TOO flexible.  I would rather add flexibility only when needed.

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+
+- Lowercase alphanumeric characters and hyphens only.
+- Must start with a lowercase alphanumeric character.
+- Must end with a lowercase alphanumeric character.
+- Maximum 63 characters.
+- Regex: `^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$`
+
+### 2.4 `{Resource}Ref` — compound identity type
+
+Every atespace-scoped resource type **must** have a corresponding `{Resource}Ref` message that bundles `atespace` + `name` as a single, typed unit:
```
thockin: Why is this better than a single common struct for those fields?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+
+The field name is the logical name of the reference (e.g., `actor_template`), not `actor_template_name` or `actor_template_ref`.
+Note that this assumes that ActorTemplates are also resources in the substrate gRPC API (not in KRM).
+
+**Global-scoped resources** have no atespace, so no wrapper type is needed — use a plain `string {resource}_name` field:
+
+```proto
+message Actor {
+  // worker_name references the global-scoped Worker assigned to this actor.
+  string worker_name = 10;
```
thockin: Why no, for symmetry, define a message anyway?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+}
+```
+
+The field name is the logical name of the reference (e.g., `actor_template`), not `actor_template_name` or `actor_template_ref`.
+Note that this assumes that ActorTemplates are also resources in the substrate gRPC API (not in KRM).
+
+**Global-scoped resources** have no atespace, so no wrapper type is needed — use a plain `string {resource}_name` field:
+
+```proto
+message Actor {
```
thockin: Worker

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+## 3. Standard Methods
+
+The following sections cover each standard method. The primary adaptation from AIP-13x is how resources are identified in requests: two fields (`atespace` + `name`) instead of a single `name` path string.
+
+### 3.1 Get
+
+*Follows [AIP-131](https://google.aip.dev/131).*
+
+```proto
+rpc GetActor(GetActorRequest) returns (Actor) {}
```
thockin, line 133: This is not what we have today (`rpc GetActor(GetActorRequest) returns (GetActorResponse) {}`) - is this proposing to change that?  Is there ever a need to return additional metadata?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+  // actor.atespace and actor.name together specify the resource's identity
+  // and must both be set by the caller.
+  Actor actor = 1;
+}
+```
+
+Rules:
+- RPC name **must** begin with `Create` followed by the singular resource name.
+- Response **must** be the resource itself — not a `CreateActorResponse` wrapper.
+- The resource body is the only field in the request. There are no separate top-level `atespace` or `{resource}_id` fields.
```
thockin: In k8s we have at least 1 interesting field in CreateOptions - `DryRun bool` .  If we wanted to implement a dry-run, how would we do it?

thockin: AIP 163 says it would go in the request

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+- `actor.atespace` and `actor.name` are **required** and caller-specified. The server does not generate them.
+- If a resource already exists with the same `(atespace, name)`: return `ALREADY_EXISTS`. If the caller lacks permission to observe the conflicting resource: return `PERMISSION_DENIED`.
```
thockin: s/observe/create ?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+```
+
+Rules:
+- RPC name **must** begin with `Update` followed by the singular resource name.
+- Response **must** be the resource itself — not an `UpdateActorResponse` wrapper.
+- `update_mask` **must** be of type `google.protobuf.FieldMask` and **must** be named `update_mask`.
+- `update_mask` is **required**. An absent or empty mask **must** return `INVALID_ARGUMENT`.
+- The special value `*` is **not supported**. Clients must enumerate the exact fields to update.
+- The resource's `atespace` and `name` identify the resource to update; they are not themselves updatable.
+- If the resource does not exist: return `NOT_FOUND`.
```
thockin, line 243: permission denied case needed?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+
+**Divergence from AIP-134:** AIP-134 makes `update_mask` optional (omission implies updating all populated fields) and requires support for `*`. Substrate requires an explicit mask.
+
+### 3.5 Delete
+
+*Follows [AIP-135](https://google.aip.dev/135).*
+
+```proto
+rpc DeleteActor(DeleteActorRequest) returns (google.protobuf.Empty) {}
```
thockin: Is there a reason not to return the last value when successful?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+
+- Field definitions in proto files **must** use `lower_snake_case`.
+- Boolean fields **must** omit the `is_` prefix: use `disabled`, not `is_disabled`. Exception: use `is_` when the bare word would be a reserved keyword in common languages.
+- Repeated fields **must** use the plural noun form: `containers`, not `container`.
+- Non-repeated fields **must** use the singular form: `container`, not `containers`.
+- Field names **must** be nouns, not verbs: `worker_selector`, not `select_workers`.
+- Use standard abbreviations where well-established: `config`, `spec`, `id`, `info`, `stats`.
+- Adjectives come before the noun: `suspended_actors`, not `actors_suspended`.
+- Avoid prepositions in field names: `error_reason`, not `reason_for_error`.
+- Do not use `name` for any purpose other than a resource's own name field (see section 2.1). Use specific names: `display_name`, `actor_template_name`, etc.
```
thockin: how about things that are nested but named - e.g. volume names?  Would you make me say `at.volumes[i].volume_name` ?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+  google.protobuf.Timestamp create_time = 4;
+
+  // update_time is the time the resource was last updated by a user action.
+  google.protobuf.Timestamp update_time = 5;
+
+  // delete_time is set when the resource is soft-deleted.
+  // Only present on resources that support soft delete.
+  google.protobuf.Timestamp delete_time = 6;
+
+  // version is incremented on every mutation. See §7.
```
thockin: Is this guaranteed to increment by 1 or to increase by an arbitrary amount?  The latter would have made some things in k8s easier.

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+
+  // ... resource-specific fields
+}
+```
+
+### 6.1 `uid`
+
+- Type: `string`.
+- A server-assigned [UUID4](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)).
+- Stable across all mutations: updating or "renaming" a resource (changing its `name` in a future API version) does not change its `uid`.
```
thockin: This implies that rename is something we intend to support - I don't think it is, I would just remove that bit.  IOW `space`+`name` is unique across space (at a point in time) and `uid` is unique across space and time.

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+- RPC name **must** begin with `Get` followed by the singular resource name.
+- Request message name **must** match the RPC name with a `Request` suffix.
+- Response **must** be the resource itself — not a `GetActorResponse` wrapper.
+- Request **must** identify the resource with a single `{Resource}Ref` field (atespace-scoped) or `string name` (global-scoped).
+- If the resource does not exist: return `NOT_FOUND`.
+- If the caller lacks permission: return `PERMISSION_DENIED` (checked before existence).
+
+### 3.2 List
+
+*Follows [AIP-132](https://google.aip.dev/132).*
```
jpbetz, line 149: Does this mean ordering and filtering (https://google.aip.dev/160) are supported?

docs/api-style-guide.md
jpbetz: In k8s, we had to graft on a delete response type because it turns out to be super useful for our lifecycles. In k8s a resource deletion is not immediate and so tracking the post-delete lifecycle becomes interesting/complicated

I don't know what the plans are for deletion/deletion-hooks in this system, but something you could consider future-proofing against?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+  string uid = 3;
+
+  // create_time is the time the resource was created.
+  google.protobuf.Timestamp create_time = 4;
+
+  // update_time is the time the resource was last updated by a user action.
+  google.protobuf.Timestamp update_time = 5;
+
+  // delete_time is set when the resource is soft-deleted.
+  // Only present on resources that support soft delete.
```
jpbetz: More detail on soft delete in the section about delete would have been enlightening for me, I didn't realize it was possible/supported until I read this part..

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+- Resources that do not support soft delete **must not** include this field.
+- Substrate's current resources use hard delete; add `delete_time` only if soft delete is introduced for a specific resource type.
+
+---
+
+## 7. Resource Freshness and Optimistic Concurrency
+
+*Inspired by [AIP-154](https://google.aip.dev/154). Diverges in field name and type.*
+
+When two clients update the same resource concurrently, the second write may silently overwrite the first. Freshness validation lets a client prove it is operating on the state it thinks it is, so the server can reject stale writes.
```
jpbetz, line 437: In k8s, we initially set up resources so clients could opt-in to optimistic concurrency (for k8s it was an RV, here it would be "version").  We ended up regretting that and in newer APIs we require optimistic-concurrency.  It's not clear to me what choice you've made here, but I wanted to raise that as something we believe we got wrong initially.

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+
+Substrate uses a field named **`version`** of type `int64` for this. AIP-154 uses an opaque `etag` string; we diverge for a concrete reason: Substrate maintains an in-memory worker cache that guards against applying stale watch events over newer cached state using a numeric `>=` comparison. An opaque string cannot serve this role — the ordering guarantee IS the implementation contract, so the field type should reflect it.
+
+The name `version` is intentional: it increments on every write (like Kubernetes's `resourceVersion`) and is a transparent, comparable integer (like Kubernetes's `generation`). It serves both roles in one field, so neither Kubernetes name fits cleanly.
```
jpbetz: Sorry to nit.. k8s generation on incremented only when a kubernetes spec changes, where-as resourceVersion is just etcd's revision number bubbled up the stack, so if version is incremented only when a specific object changes, it's slightly different than both those concepts.

Not a criticism of the core idea here, though. If you don't have a spec/status split, then version sounds like a nice way to handle optimistic concurrency.

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+  // The atespace this actor belongs to.
+  string atespace = 1;
+  // The name of this actor, unique within its atespace.
+  string name = 2;
+
+  // ... other fields
+}
+```
+
+- `atespace` is the Substrate namespace for the resource.
```
jpbetz: Sounds like you're taking roughly the same approach k8s did to namespaces. Is this done because you need to map to k8s concepts?  Would a list of things we regret with the global + flat namespaces approach be useful to consider?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+  google.protobuf.FieldMask update_mask = 2;
+}
+```
+
+Rules:
+- RPC name **must** begin with `Update` followed by the singular resource name.
+- Response **must** be the resource itself — not an `UpdateActorResponse` wrapper.
+- `update_mask` **must** be of type `google.protobuf.FieldMask` and **must** be named `update_mask`.
+- `update_mask` is **required**. An absent or empty mask **must** return `INVALID_ARGUMENT`.
+- The special value `*` is **not supported**. Clients must enumerate the exact fields to update.
```
jpbetz, line 241: Good call. I don't understand why '*' is required by the AIP guidelines but it looks like a bad idea to me.

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+
+```proto
+rpc CreateActor(CreateActorRequest) returns (Actor) {}
+
+message CreateActorRequest {
+  // The actor to create.
+  // actor.atespace and actor.name together specify the resource's identity
+  // and must both be set by the caller.
+  Actor actor = 1;
+}
```
jpbetz, line 198: FWIW, In k8s, we ended up regretting not defining more validation in the API definitions and have spent quite a lot of time fixing that.  If you have mechanisms to define core validation (requiredness, defaults, bounds/limits, ...), I highly recommend using them so that you have a standard way of defining validation

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
+# Substrate gRPC API Style Guide
```
jpbetz, line 1: Do you have (or want to have) the idea of stability levels and/or feature gating/enablement, either at the resource level or field level?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+
+*Follows [AIP-136](https://google.aip.dev/136).*
+
+Custom methods are for operations that don't map cleanly to CRUD: lifecycle transitions (Suspend, Resume, Pause), long-running actions, or commands with side effects that standard Update semantics would misrepresent.
+
+```proto
+rpc SuspendActor(SuspendActorRequest) returns (Actor) {}
+
+message SuspendActorRequest {
+  ActorRef actor = 1;
```
jpbetz: Given that you have refs, how do you manage deletion such that resources don't become orphaned?

docs/api-style-guide.md
thockin: IT's mostly because the problem-spaces are similar.  But yes,more lessons.  I know a lot about the issues around namespaces, but I bet you know more :)

docs/api-style-guide.md
juli4n: Not yet, but it's something I want to define soon!

docs/api-style-guide.md
jpbetz: I'll give it a shot! Not all of these have consensus. So feel free to disagree on any/all of them. I consider them worth discussion, at the very least.

Global/Namespace scope split issues:

  - `ClusterRoleBinding` and `RoleBinding` are basically the same resource, but one is global scoped and the other is namespace scoped. This exemplifies that SOME resources are useful in both scopes and having a hard split between the scopes limits composition.  Other examples: `SubjectAccessReview` vs `LocalSubjectAccessReview`. Note how the naming is inconsistent :(   (This has proliferated into the ecosystem. E.g. `ClusterIssuer` vs `Issuer` in cert-manager).
  - Resources such as `ValidatingAdmissionPolicy` are global scope but have a field that limits the scope of use to a namespace (`matchResources.namespaceSelector` in the case of `ValidatingAdmissionPolicy`). This exemplifies resources that are  resources lack a namespace scoped version (`LocalValidatingAdmissionPolicy` !?), that would allow the resource to be defined within the scope it influences.
  - We lack global `ResourceQuota` and `LimitRange` for no good reason other than the inconvenience of having to define them separately
  - `IngressClass` (global) relates to an `Ingress` (namespaced) and has an explicit `scope` field (set to namespace or cluster) plus a namespace field..  So even though it's global scope, it can be assigned a namespace via fields...
 
Reference issues:

  - `PersistentVolume` (global) is bound 1:1 with a `PVC`) (namespaced) via a reference from global scope *to* namespace scope. This exmplifies the need to have references *from* global scope *to* namespace scope (`PV.spec.claimRef`.. (APIService -> Service does this too)

  - `Secret` and `ConfigMap` are namespace scoped but often are needed across namespaces, but can't be referenced cleanly so we end up with a lot of copying and odd plumbing. Both are arguably needed at the cluster level.

Organizational issues:

- Multiple "Namespace as a tenant" project have more or less failed due to the inability to configure a namespace from within the namespace. Attempts to graft on hierarchical namespaces have hit similar issues (as well as some issues specific to hierarchical namespaces).

My personal takeaways:

- Deciding "should this be global or namespaced" a priori is hard to get right.  Some (all ?!) resources should be both.
- If we had hierarchical resources, then "global" would presumably just be the root namespace.
- While a hierarchical namespace system would have been a lot more complicated to implement (and may very well have been intractable early on during k8s development), the presence of the hierarchy would have served as a forcing function to address a lot of the problems that have emerged as a result of the global/namespace split...

docs/api-style-guide.md
jpbetz: If you have namespaces, the ability to list by namespace is a particularly important "filter"..

docs/api-style-guide.md
jpbetz: Re `DryRun`.. The list of [options](https://github.com/kubernetes/kubernetes/blob/b6b907fda0d598ea8c2c92bd886f4e22ae10844a/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go) for the k8s verbs are the result of a lot of hard earned API lessons. 

For example, if you allow resources to be deleted and then later created with the same name, how do you prevent a caller from updating/deleting a new lifecycle of the same-named resource on accident?  `version` isn't enough.

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+- `name` is the resource name within the atespace. It is not a path; it is a short identifier.
+
+### 2.2 Global-scoped resources
+
+Some resources are global across the entire deployment and do not belong to any atespace. For these, the identity is `name` alone. The first field **must** be:
+
+```proto
+message Worker {
+  // The name of this worker, globally unique (e.g. the node name).
+  string name = 1;
```
jpbetz: After a resource with a given name is deleted, can it be created again? How are you planning on dealing with the per-lifecycle identity?

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+  // Omit or leave empty for the first request.
+  string page_token = 3;
+}
+
+message ListActorsResponse {
+  repeated Actor actors = 1;
+
+  // Pagination token for the next page.
+  // Empty if this is the last page.
+  string next_page_token = 2;
```
jpbetz, line 173: The absence of a field with the "total count" or "remaining item count" is notable.  Is this deliberate?  This can be expensive to compute.  If you're deliberately choosing not to provide this, I recommend making the decision explicit here and including the rationale so that you fully lock in the decision.

docs/api-style-guide.md
thockin: Thanks for the notes!  Nothing here is exactly "news" to me, though some were reminders of how messy it can get :)

> Deciding "should this be global or namespaced" a priori is hard to get right. Some (all ?!) resources should be both. 

What prevents us from using the same Kind on different URL paths in k8s?  I know the answer is partly related to how Kind and Resource are mapped, but that seems like a problem of our own making.

> If we had hierarchical resources, then "global" would presumably just be the root namespace.

Hierarchy is a singular concept.  It allows you to define a single "relationship" but makes other relationships harder.  E.g. "who can access a resource" is not necessarily the same as "who pays for the resource" nor "what quota is applied to this resource".  I say this as someone who STILL thinks that kube should have at least one extra level of hierarchy, so ...

We're looking at a different sort of API here - RPC oriented (even if we have resources) and narrower (for now!).  It's hard to make assertions about what we will or won't need in the future and keep a straight face - our ability to see the future has a poor track record.  I wonder how we can best leave room for expansion here?  For example, we could define atespaces as hierarchical but only allow a single layer for now.  E.g. "" is a the root space ("global"), while  "foo" is just the first layer in hierarchy and MAYBE EVENTUALLY "foo/bar/baz/qux" is nested.  That has a lot of ripple on things like auth and quota and networking.  If they do not implement as hierarchical, retrofitting it will be hard (or we might pick semantics that are impossible).

Is this something worth investing in now?

docs/api-style-guide.md
juli4n: Agree.

docs/api-style-guide.md
thockin: For clarity:

ListOptions:
* LabelSelector
* FieldSelector
* Watch
* AllowWatchBookmarks
* ResourceVersion
* TimeoutSeconds
* Limit
* Continue
* SendInitialEvents
* ShardSelector

Many of those are related to watch semantics.  We will need some selector behavior, but our notion of watch and RV is pretty different (for better or worse) for now.

GetOptions:
* ResourceVersion

We don't have a global or per-resource RV, so this doesn't (yet) map.

DeleteOptions:
* GracePeriodSeconds
* Preconditions
* OrphanDependents
* DryRun
* IgnoreStoreReadErrorWithClusterBreakingPotential

We haven really defined general "soft delete" yet, but DryRun seems apt.

CreateOptions:
* DryRun
* FieldManager
* FieldValidation

DryRun again

PatchOptions, ApplyOptions, UpdateOptions:
* DryRun
* FieldManager
* FieldValidation

DryRun

docs/api-style-guide.md
thockin: Maybe I can repurpose validation-gen :)

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+message UpdateActorRequest {
+  // The actor to update.
+  // The actor's atespace and name fields identify which resource to update.
+  Actor actor = 1;
+
+  // The set of fields to update. Required.
+  //
+  // Field paths are relative to the Actor message (e.g., "worker_selector").
+  google.protobuf.FieldMask update_mask = 2;
+}
```
thockin, line 230: Another case for dry-run (in the future)

docs/api-style-guide.md
```diff
@@ -0,0 +1,457 @@
...
+### 3.5 Delete
+
+*Follows [AIP-135](https://google.aip.dev/135).*
+
+```proto
+rpc DeleteActor(DeleteActorRequest) returns (google.protobuf.Empty) {}
+
+message DeleteActorRequest {
+  ActorRef actor   = 1;
+  int64     version = 2; // optional freshness guard; 0 = skip check
```
thockin: probably should add optional UID precodition here

```
message Preconditions {
    int64 version = 1; 0 = do not compare
    string uid = 2; "" = do not compare
}
```

docs/api-style-guide.md
juli4n: I personally like having dedicated messages (and generated Go types) for references. It's very common in our codebase to pass refs through the call stack, sometimes methods take refs, even of different types. Having a dedicated type for each type of reference (e.g. ActorRef, ActorTemplateRef) makes the code less error-prone. 

Following this reasoning, I guess the logical followup will be to also have ref messages for global resources (so they have a dedicated type instead of a raw string), which is what you asked for in line 122

thockin: That's OK when we have 5 resources, cumbersome when we have 15, and a PITA when we have 25.  I'm OK with it for now, but if it becomes a problem, we revisit?

docs/api-style-guide.md
juli4n: This is talking about cross-resource references, and using Actor -> Assigned Worker as an example. The section about identity of global resources is line 48.

docs/api-style-guide.md
juli4n: See my comment above. I would be in favor of that, not so much because of symmetry, but because it allows us to pass dedicated ref types through the call stack.

docs/api-style-guide.md
juli4n: This was following AIP basically. I would probably say that returning something *additional* to the resource sounds like a design smell? Having the style guide preventing from that type of scope creeping sounds like a good idea?

docs/api-style-guide.md
juli4n: Clarified that ordering and filtering are not in scope. This would be fairly challenging unless we move out of ValKey I think.

docs/api-style-guide.md
juli4n: Do you mean declarative validations that are defined as metadata in the proto?

docs/api-style-guide.md
juli4n: Yes, that would go in the request. I realize the wording was confusing so I removed it. It was initially trying to say that no *resource* fields should be top level.

There are use cases for fields that are not resource field, dry-run (as you mentioned), idempotency token.

docs/api-style-guide.md
juli4n: I would rather delete everything related to permissions. I think we can extend this guide once we figure out our authz story. WDYT?

docs/api-style-guide.md
juli4n: How about skipping authz until we have a design for it?

docs/api-style-guide.md
juli4n: Argh, this was not supposed to be here :) Removed.

docs/api-style-guide.md
juli4n: The idea is that all resources will have a `version` field. If a client does a naive read-modify-write, the optimistic concurrency check will trigger, because the update endpoint will honor it.

docs/api-style-guide.md
juli4n: Agree. Done.

docs/api-style-guide.md
juli4n: I'm a bit on the fence on this one. It sounds tempting to DRY the fields, but I kind of like having dedicated ref types (see my comment below).

docs/api-style-guide.md
thockin: I'm fine with it, I guess.  I can't actually think of WHAT we would return that isn't part of the resource itself.  I saw later that you confirmed this IS a change from today's code

docs/api-style-guide.md
thockin: What about selectors - we have to have some basic selector-logic for scheduling, are we just going to leave that as client-side filtering for now?

docs/api-style-guide.md
thockin: That's how I read his comment.  It's a bit premature right now, but also it's easy for handwritten validation to become a snowball which is hard to stop.

docs/api-style-guide.md
thockin: Should we define `<op>Options` messages and embed those in each such `<op><type>Request` ?  Or do you think there's value in type-specific messages for this, too?

docs/api-style-guide.md
thockin: It seems largely orthogonal, so OK.  Maybe leave a section that says, more or less:

Every method must check permissions before anything else. Example: A GetActor() call for "a" must check that the caller is allowed to read actors in the context of "a" (e.g. that enclosing atespace) and if not, return PERMISSION_DENIED whether "a" exists or not (as opposed to returning NOT_FOUND).

docs/api-style-guide.md
thockin: That works for RMW (but does the field mask allow the user to ignore the version field?)

For things like "patch" - should we require version and/or UID preconditions?

For things like Delete() - should we require version and/or UID preconditions?

docs/api-style-guide.md
jpbetz: > What prevents us from using the same Kind on different URL paths in k8s? I know the answer is partly related to how Kind and Resource are mapped, but that seems like a problem of our own making.

It's 100% of our own making.

> Is this something worth investing in now?

I don't know enough to take a strong stance on the hierarchy-or-not decision, but if you don't provide hierarchy, I'd encourage you to still consider a mental model of "root namespace and one level of namespaces under that" instead of "global and namespaced". If you squint it's the same thing, but if you consider the compositional requirements that trickle down from this, you could maybe avoid some of the mistakes k8s made.

docs/api-style-guide.md
thockin: I can buy the strong-typing argument for Ref types (though it's annoying to have so many, they are exceedingly unlikely to change).  In Go this would be a perfect use of a type-def.  Alas.

What I am asking is whether we should have one common message which is ALL the metadata, name, space, labels, uid, version, etc.  All the stuff we want every "resource message" to carry.

thockin:
Not exactly related to style/convention, but something we might consider:

a problem in k8s is "history".  I sometimes want to delete something but later find it is useful to cross-reference.  A long time back we talked about having a "hardlink" model where you could read by UID or by name.  When you deleted an object it could still be found by its uid for X time.

Is that something we want to think about?

docs/api-style-guide.md
juli4n: Yes, you can. Each resource will have a UID field which is server side generated, which is the per-lifecycle identity.

docs/api-style-guide.md
juli4n: We increment by one today. My intuition would be to *not* make it part of the contract. Today the only usage is to establish a total order on "snapshots" of a resource (e.g. in worker cache). That only requires being monotonically increasing, not sequential.

docs/api-style-guide.md
juli4n: It was deliberate, yes, as you said, depending on how the data model is designed it could be hard to compute. Added.

docs/api-style-guide.md
juli4n: Got it. I brought this up because I think you were also proposing to use ref types in the resource itself [1]. In that case, we cannot have a generic metadata message. Right?

[1]: https://github.com/agent-substrate/substrate/pull/280/changes#r3494492912

thockin: If we go with ActorRef and WorkerRef and WhateverRef, then you are right.

docs/api-style-guide.md
juli4n: Done

docs/api-style-guide.md
thockin: Discussed - let's leave it out for now, even though we suspect it will come back.

docs/api-style-guide.md
juli4n: Discussed. Updated style guide to reflect the decision. We will have an ObjectRef which will be reused across all resources.

docs/api-style-guide.md
juli4n: IMO we should be able to use `name` in nested fields if it makes sense. Remove this item.

docs/api-style-guide.md
juli4n: I clarified a bit the delete story. Changed to return the last version of the resource. I'm sure we will iterate on this, for now delete is synchronous but I suspect we will switch to an async (ala k8s) model soon.

docs/api-style-guide.md
juli4n: Added some more details.

docs/api-style-guide.md
juli4n: The style guide talks about version preconditions for Update/Delete. It doesn't mention UID preconditions, but it mentions how we can extend it to support it.

docs/api-style-guide.md
juli4n: Thanks, yes the paragraph is a bit light in details. You are right in that the proposed schema is kind of a hybrid of resourceVersion and generation.

docs/api-style-guide.md
juli4n: I know @thockin has been looking into how to introduce some sort of validation framework via metadata. I'll close this comment for now, once we have it in place we can augment it with guidelines on validation metadata.

docs/api-style-guide.md
juli4n: Added a mention to the dry run use case.

docs/api-style-guide.md
juli4n: Not sure how strong you feel about this. In the current version I'm proposing top level fields for this.

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+
+* Atespace-scoped resources belong to an atespace. Their identity is `(atespace, name)`, unique within the resource type.
+* Global-scoped resources are global across the entire deployment and do not belong to any atespace. For these, the identity is `name` alone.
+
+In both cases, a `meta` field contains both `atespace` and `name`. For global resources, the `atespace` must always be empty.
+
+```proto
+message Actor {
+  // The atespace this actor belongs to.
+  ResourceMetadata meta = 1;
```
thockin: super nitpick - "metadata" to be like k8s?

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+
+  // ... other fields
+}
+
+
+message ResourceMetadata {
+  // The atespace this resource belongs to.
```
thockin: Comment that this is "" for global resources?

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+  // ... other common resource fields
+}
+```
+
+- All resources in Substrate must have a `ResourceMetadata meta = 1` field to hold common fields, which includes both `atespace` and `name`.
+
+### 2.3 Character constraints
+
+Both `atespace` and `name` must be valid resource names.
+A valid resource name must comply the following rules:
```
thockin: "comply with"

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+**1. Request messages** — to identify which specific resource to act on:
+
+```proto
+message GetActorRequest {
+  ObjectRef actor = 1;
+}
+
+message DeleteActorRequest {
+  ObjectRef actor   = 1;
+  int64     version = 2;
```
thockin: I would leave the version out here and just add `// ...other fields`, otherwise it is confusing

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+  // ... other fields
+}
+```
+
+The field name is the logical name of the reference (e.g., `actor_template`), not `actor_template_name` or `actor_template_ref`.
+Note that this assumes that ActorTemplates are also resources in the substrate gRPC API (not in KRM).
+
+- Do not embed the full resource message as a reference field.
+- Do not use a single combined string like `"atespace/name"`. Callers would have to parse it.
+- Do not use a plain `string {resource}_name` field for global-scoped references — use `ObjectRef` for consistency and type safety.
```
thockin, line 118: Question - often refs will be within the same atespace.  Are we expected to fill in the `atespace` field or leave it empty and validate that it is not specified?

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+```
+
+Rules:
+- RPC name **must** begin with `List` followed by the **plural** resource name.
+- Both the request and response message names **must** match the RPC name with `Request`/`Response` suffixes. (Unlike Get/Create/Update, List responses are not the resource itself.)
+- `next_page_token` **must** be present on every List response message. It **must** be empty when there are no further pages.
+- The repeated resource field **must** use the plural form of the resource name (e.g., `actors`, not `actor`).
+- If a user provides a `page_size` above the maximum, coerce it silently. If a user provides a negative value, return `INVALID_ARGUMENT`.
+- Sorting and filtering as specified in AIP-132 are not supported.
+- Clients must iterate over all pages until an empty `next_page_token` is returned. Clients should assume that an empty result means the end of the page stream.
```
thockin: Clarify "empty result" here?  It's ambiguous to me when the first part is clearer.

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+  // and must both be set by the caller.
+  Actor actor = 1;
+}
+```
+
+Rules:
+- RPC name **must** begin with `Create` followed by the singular resource name.
+- Response **must** be the resource itself — not a `CreateActorResponse` wrapper.
+- `actor.meta.atespace` and `actor.meta.name` are **required** and caller-specified. The server does not generate them.
+- Other meta fields such as `uid`, timestamps, `version`, etc, are server side generated.
```
thockin: ", and are ignored when specified as input" or is it an error to specify them?

docs/api-style-guide.md
thockin: I think we want the same preconditions available for all resources, right?  So we will either dedup them in a struct or embed them in every DeleteFooRequest -- struct seems better (same for dry run an other CreateOptions, etc)?

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+  ObjectRef actor   = 1;
+  int64     version = 2; // optional freshness guard; 0 = skip check
+}
+```
+
+Rules:
+- RPC name **must** begin with `Delete` followed by the singular resource name.
+- Response **must** be the deleted resource.
+- Request **must** identify the resource with an `ObjectRef` field (for both atespace-scoped and global-scoped resources).
+- If the resource does not exist: return `NOT_FOUND`.
```
thockin, line 279: This makes Delete non-idempotent.  Do we think this matters?

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+  // Field paths are relative to the Actor message (e.g., "worker_selector").
+  google.protobuf.FieldMask update_mask = 2;
+}
+```
+
+Rules:
+- RPC name **must** begin with `Update` followed by the singular resource name.
+- Response **must** be the resource itself — not an `UpdateActorResponse` wrapper.
+- `update_mask` **must** be of type `google.protobuf.FieldMask` and **must** be named `update_mask`.
+- `update_mask` is **required**. An absent or empty mask **must** return `INVALID_ARGUMENT`.
```
thockin, line 237: What happens if output-only fields are included in the mask?  Ignored or error?

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+```proto
+message ResourceMetadata {
+  string atespace = 1;
+  string name     = 2;
+
+  // uid is a server-assigned, globally unique identifier for this resource.
+  // Immutable throughout the lifecycle of the resource.
+  string uid = 3;
+
+  // version is incremented on every mutation.
```
thockin: Can we say "increased" rather than incremented?  That leaves more room

docs/api-style-guide.md
```diff
@@ -0,0 +1,435 @@
...
+  int64 version = 4;
+
+  // create_time is the time the resource was created.
+  google.protobuf.Timestamp create_time = 5;
+
+  // update_time is the time the resource was last updated by a user action.
+  google.protobuf.Timestamp update_time = 6;
```
thockin, line 386: I assume delete_time comes when we have soft delete?

docs/api-style-guide.md
thockin: a) Should we add UID preconditions now, later, or never?

b) Should we require any preconditions for any ops or is it totally up to users?

docs/api-style-guide.md
juli4n: Yes.

docs/api-style-guide.md
juli4n: Done.

docs/api-style-guide.md
juli4n: Done.

docs/api-style-guide.md
juli4n: Done.

docs/api-style-guide.md
juli4n: Done.

docs/api-style-guide.md
juli4n: I don't think it matters much? In practice, this is just an if condition on the caller side. FWIW, AIP has https://google.aip.dev/135#delete-if-existing but I don't think it adds much value.

docs/api-style-guide.md
juli4n: Good call. Clarified that it should fail. This is clearly a bug in the client as the intent is explicit (e.g. bad understanding of the resource model). Ignoring would only hide client side bugs.

docs/api-style-guide.md
juli4n: Done.

docs/api-style-guide.md
juli4n: I'm leaning towards fill it in, the server would validate the invariant that the referenced resource is within the same namespace. k8s has dedicated types for local object refs, that would be an alternative, but I would say let's iterate as we don't have that yet.

thockin: The counter-argument is that (with redis/valkey) all the data is in-memory.  Less is more.

docs/api-style-guide.md
thockin: Kube is not idempotent here, either.  It's always sort of bothered me that an errant proxy-level retry might get an error which bubble up to the user, but it has never been an issue in practice, and it feels weird to make all delete ops return success when objects don't exist (likewise create ops would have to compare exact state).  It seems like a lot of work for low ROI

docs/api-style-guide.md
thockin: For precondition fields, are those included in the mask or not?

docs/api-style-guide.md
thockin: Pinging this one to clarify.  I read "Clients must iterate over all pages until an empty `next_page_token` is returned" combined with the assertion elsewhere that a list result may return 0 items + valid next_page_token as saying the only way to know you are done is when `next_page_token` is "".

But then "assume that an empty result means the end of the page stream" is vague - does "empty result" mean that you received 0 items or that `next_page_token` is ""?

docs/api-style-guide.md
thockin: ping

juli4n: Done.

docs/api-style-guide.md
juli4n: Done.

docs/api-style-guide.md
juli4n: Updated.

docs/api-style-guide.md
juli4n: Clarified this in the section about optimistic locking. They should not be included in the mask.

docs/api-style-guide.md
thockin: We can decide this later, but it won't bee too long before we come upon it.

Add a TODO?

docs/api-style-guide.md
thockin: I think you should mention it here too, otherwise the two sections are in conflict.  This says "setting an output_only field == INVALID" but later it says you can set "version".

docs/api-style-guide.md
```diff
@@ -0,0 +1,442 @@
...
+  google.protobuf.FieldMask update_mask = 2;
+}
+```
+
+Rules:
+- RPC name **must** begin with `Update` followed by the singular resource name.
+- Response **must** be the resource itself — not an `UpdateActorResponse` wrapper.
+- `update_mask` **must** be of type `google.protobuf.FieldMask` and **must** be named `update_mask`.
+- `update_mask` is **required**. An absent or empty mask **must** return `INVALID_ARGUMENT`.
+- If `update_mask` includes an immutable field (e.g. output only), **must** return `INVALID_ARGUMENT`.
```
thockin: nit: "output-only" and "immutable" are not the same thing.  This is ambiguous WRT things like version and mtime which DO change (are not immutable) but are output-only.

docs/api-style-guide.md
```diff
@@ -0,0 +1,442 @@
...
+*Follows [AIP-140](https://google.aip.dev/140) and [AIP-149](https://google.aip.dev/149).*
+
+- Field definitions in proto files **must** use `lower_snake_case`.
+- Boolean fields **must** omit the `is_` prefix: use `disabled`, not `is_disabled`. Exception: use `is_` when the bare word would be a reserved keyword in common languages.
+- Repeated fields **must** use the plural noun form: `containers`, not `container`.
+- Non-repeated fields **must** use the singular form: `container`, not `containers`.
+- Field names **must** be nouns, not verbs: `worker_selector`, not `select_workers`.
+- Use standard abbreviations where well-established: `config`, `spec`, `id`, `info`, `stats`.
+- Adjectives come before the noun: `suspended_actors`, not `actors_suspended`.
+- Avoid prepositions in field names: `error_reason`, not `reason_for_error`.
```
thockin, line 322: worth mentioning const names?

docs/api-style-guide.md
```diff
@@ -0,0 +1,442 @@
...
+
+### 5.1 Field presence (`optional`)
+
+Use the `optional` keyword on a scalar field **only** when null and the zero value (`false`, `0`, `""`) are semantically distinct for that field's meaning. Do not use `optional` universally or as a workaround for update semantics — the required `update_mask` handles that.
+
+```proto
+// Only if "unrated" is meaningfully different from "rated 0":
```
thockin: nit: "rated" is a word word when the field in example is "priority"

docs/api-style-guide.md
juli4n: Added details.

docs/api-style-guide.md
juli4n: Added TODO.

docs/api-style-guide.md
juli4n: Clarified.

docs/api-style-guide.md
juli4n: Reworded.

docs/api-style-guide.md
juli4n: Added.

docs/api-style-guide.md
juli4n: Done.

docs/api-style-guide.md
juli4n: a) Added UID preconditions now (discussed over chat).
b) Opt-in for now. We can revisit this later if needed.

docs/api-style-guide.md
juli4n: Reusing <op>Options sounds reasonable. I added this for delete (as we need it for preconditions).

PR #352, opened 2026-06-29
Update default GKE cluster version and enable Managed OTel feature

### LINKED ISSUE #341: Default cluster version is no longer supported by GKE in us-central1-c
When I'm running `go run ./tools/setup-gcp bootstrap` on a new GCP project, it is failing as of today.

The bootstrapping step is failing with an error message saying the default cluster version 1.35.0-gke.2398000 is no longer supported by GKE in us-central1-c.

Updating the script to use a newer version like `1.35.6-gke.1049000` by default fixes the problem.

zoez7:
1) Always create a cluster with GKE Managed OpenTelemetry feature and
2) Update the default cluster version to 1.35.5-gke.1163012 which is the current [regular channel default version](https://docs.cloud.google.com/kubernetes-engine/docs/release-notes#current_versions) to fix https://github.com/agent-substrate/substrate/issues/341

Tested:
* Successfully recreated GKE cluster, redeployed ate-system and ran counter demo.

### DIFF
--- a/cmd/kubectl-ate/README.md
+++ b/cmd/kubectl-ate/README.md
@@ -36,7 +36,9 @@ The CLI supports on-demand tracing using the `--trace` flag. When enabled, the C
 gcloud services enable cloudtrace.googleapis.com --project=PROJECT_ID
 ```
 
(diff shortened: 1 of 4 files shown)


maxsmythe (approved):
LGTM

Issue #292, opened 2026-06-23
Add the CRASHED state to Actor State Machine


zoez7:
Description: Discover all scenarios where actors are in a state where manual recovery is required. 
https://github.com/agent-substrate/substrate/issues/119 has listed some examples. Implement the transition to the CRASHED state.

 `CRASHED` means the actor can not be automatically recovered, and requires manual intervention to revert to a known good state.

zoez7:
Out of the transitions to CRASHED state mentioned in https://github.com/agent-substrate/substrate/issues/119, I think we have 2 categories
1. Ate system detects a not-automatically recoverable error during Chekpoint/Restore operations. For example, we detect missing or corrupted snapshots during checkpoint or restore. - We have entry points in the codebase that we can add the `curr_state -> CRASHED` transition.
2. Atelet or ateom crashes or becomes unhealthy. This will impact any actors that are in the RUNNING or PAUSED state, but not actors in other states. - Currently we don't watch atelet or ateom health, so this could go undetected.

In the second category, if atelet itself is unreachable it's possible that the actor and ateom is still running, only atelet is unhealthy.

zoez7:
Related to `snapshotConfig` - I think the snapshotConfig is what the user intended be snapshotted (none | homedir | process), but when the Actor is crashed, it's possible we can partially snapshot something (for example, if the actor OOM'd, we can still take snapshot of homedir) so we could save an additional `actualSnapshotted` for recovery?

zoez7:
Next step: Create a failure matrix with the following columns:

- **Actor starting state**: Running|Paused etc.
- **Failure reason**: eg. Actor OOM? Ateom crash?? 
- **End state**: CRASHED  
- **Detected by**: Ateom  ?
- **Recoverable layers:**
     - memory: ?
     -  homedir: ?
     -  rootfs: ?
- **Possible recovery steps:**
  1. \[Substrate\] snapshot filesystem state (excluding memory) to a new tag, do not delete the previous snapshot.  
  2. \[AgentDev\] run `ate actor commit actor_id` to revert to revert to a chosen snapshot.

bowei:
I think CRASHED could be detected by Ateom, but you might as well have Ateom crash itself on error and have the detection in Atelet. Atelet will have to handle Ateom crashing anyway?

zoez7:
Thanks @bowei ! This will simplify the Ateom logic. So when an Actor crashed, we can have Ateom:
[...]

PR #353, opened 2026-06-29
Detect `CRASHED` Actors during Checkpoint and Restore

zoez7:
Related to #292 and #119

Changes:

1. ResumeActor:  
   1. During `ResumeActor` we'll put the actor in CRASHED state if the Resume failed due to any errors related to the snapshot itself, which includes:  
      1. Missing or corrupted snapshot when downloading from GCS  
      2. Runsc command fails.  
   2. All other types of errors will not put the Actor in CRASHED state.  
2. SuspendActor or PauseActor  
   1. During `SuspendActor` or `PauseActor` we'll put the actor in CRASHED state if we cannot produce a valid snapshot, which includes:  
      1. cannot connect to Ateom after retries,  
      2. runsc command fails.  
      3. cannot upload the snapshot to GCS (during suspend) ; or cannot save snapshot locally (during pause).
   2. All other types of errors will not put the Actor in CRASHED state.


internal/ateerrors/ateerrors.go
```diff
@@ -0,0 +1,71 @@
...
+// to classify the failure.
+func NewGRPCError(grpcCode codes.Code, sentinel, err error) error {
+	st, derr := status.New(grpcCode, err.Error()).WithDetails(
+		&epb.ErrorInfo{
+			Domain: errorDomain,
+			Reason: sentinel.Error(),
+		},
+	)
+	if derr != nil {
+		return status.Error(grpcCode, err.Error())
```
dberkov: it silently downgrades to a plain status without ErrorInfo. WithDetails on *epb.ErrorInfo essentially never fails in practice, but if it ever will, the reason would silently disappear and a real crash would be misclassified as a transient error. An slog.Error on the fallback would catch any regression that breaks the round-trip.

cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -0,0 +1,63 @@
...
+
+// crashActor moves the actor to CRASHED and removes its worker pointer.
+func crashActor(ctx context.Context, st store.Interface, atespace, actorID string) error {
+	actor, err := st.GetActor(ctx, atespace, actorID)
+	if err != nil {
+		return fmt.Errorf("while loading actor to crash: %w", err)
+	}
+	actor.Status = ateapipb.Actor_STATUS_CRASHED
+	// TODO(zoezhao): Mark the worker as unhealthy if needed.
+	actor.AteomPodNamespace = ""
```
dberkov: could you please add TODO, that we need to keep the machine where the actor was running if CRASH came from the resume of PAUSED state. The #119 has a future concept of `ate actor dump` that might  can recover some data of actor when the image was stored locally.

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -57,6 +59,12 @@ func (s *LoadActorForPauseStep) Execute(ctx context.Context, input *PauseInput,
 	if err != nil {
 		return err
 	}
+
+	if actor.GetStatus() == ateapipb.Actor_STATUS_CRASHED {
```
dberkov: @juli4n is this is a pattern you want to keep  to check the state before transition? We discussed similar behavior for other APIs, like pause already suspended actor.

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -101,6 +110,10 @@ func (s *CallAteletPauseStep) IsComplete(ctx context.Context, input *PauseInput,
 	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_PAUSED, nil
 }
 func (s *CallAteletPauseStep) Execute(ctx context.Context, input *PauseInput, state *PauseState) error {
+	if state.Actor.GetStatus() != ateapipb.Actor_STATUS_PAUSING {
```
dberkov: @juli4n same question here.

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -101,6 +110,10 @@ func (s *CallAteletPauseStep) IsComplete(ctx context.Context, input *PauseInput,
...
+		return fmt.Errorf("expected actor in PAUSING state, got: %v", state.Actor.GetStatus())
+	}
+
 	if state.Actor.GetAteomPodNamespace() == "" {
 		return fmt.Errorf("actor is in PAUSING state but has no active worker")
```
dberkov: @juli4n - I think this is something that I discussed with you last time when I developed this code. Is it error or we need to move to CRASH state here too.

cmd/atelet/main.go
```diff
@@ -341,21 +343,22 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 		Scope:                  toAteomSnapshotScope(req.GetScope()),
 	})
 	if err != nil {
-		return nil, fmt.Errorf("while calling ateom.CheckpointWorkload: %w", err)
+		return nil, ateerrors.NewGRPCError(codes.DataLoss, ateerrors.ErrReasonCrashActor, err)
```
dberkov: is it always supposed to be crashActor?

cmd/atelet/main.go
```diff
@@ -341,21 +343,22 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
...
 	if len(sandboxRec.SnapshotFiles) == 0 {
-		return nil, fmt.Errorf("ateom reported no snapshot files for checkpoint")
+		return nil, ateerrors.NewGRPCError(codes.DataLoss, ateerrors.ErrReasonCrashActor, fmt.Errorf("ateom reported no snapshot files for checkpoint"))
 	}
 
 	switch req.GetType() {
 	case ateletpb.CheckpointType_CHECKPOINT_TYPE_EXTERNAL:
 		if err := s.uploadExternalCheckpoint(ctx, req, checkpointDir, sandboxRec); err != nil {
-			return nil, err
+			// TODO: If we can cache the snapshot locally when it fails to upload, we won't have to crash the Actor right away.
```
dberkov: This is an interesting case. Could you please create an issue describing this case, otherwise we will never back to it.

internal/ateompath/ateompath.go
```diff
@@ -19,13 +19,11 @@ import (
 	"path/filepath"
 )
 
-const (
+var (
```
dberkov: can BasePath stay under "const"?

dberkov (changes requested):
1. minor comments.
2. could you please update microvm ateom to support same behavior as for gvisor.
3. E2E is failing, i scheduled retry and it still fails. Could you please run e2e on your local environment.

cmd/ateapi/internal/controlapi/workflow_pause.go
zoez7: This was just a adhoc change for the CRASHED state, but I think a better pattern would be to add a CheckPrerequisite() step here: https://github.com/agent-substrate/substrate/blame/b755ae7388814f05418169f986cda0921cd48b03/cmd/ateapi/internal/controlapi/workflow.go#L42 to add validations like this, wdyt?

cmd/ateapi/internal/controlapi/workflow_pause.go
zoez7: IMO this should also result in a CRASHED actor, because if we are trying to Pause an actor, and found out we don't know which Ateom it is running on, either
1. The Actor is actually running on some Ateom, but we are not able to snapshot, in which case we have lost Actor's dirty state, and the only thing user can do is to revert to a previous snapshot.
2. The Actor was not running on any Ateom, but somehow is in Running or Suspending state. Running `actor commit` will restore it to SUSPENDED.

cmd/atelet/main.go
zoez7: Created https://github.com/agent-substrate/substrate/issues/362

internal/ateompath/ateompath.go
zoez7: Done.

internal/ateerrors/ateerrors.go
zoez7: Done in 1183ed318f8c549e9d93578311666d231a04b7f8 . Added slog.Err and appended the error reason to the message in case WithDetails fail.

cmd/ateapi/internal/controlapi/crash.go
zoez7: Done in 1183ed318f8c549e9d93578311666d231a04b7f8 . Added TODOs to handle bad worker state as well as preserving node info.

cmd/ateapi/internal/controlapi/workflow_pause.go
zoez7: Done in https://github.com/agent-substrate/substrate/commit/1183ed318f8c549e9d93578311666d231a04b7f8 . Crashed the Actor if we are trying to suspend or pause it, but couldn't find the worker assignment.

dberkov (approved):

PR #356, opened 2026-06-30
ateredis: encode Actor/Worker records as binary protobuf (#307)

### LINKED ISSUE #307: Adopt binary protobuf encoding for control-plane records (Actor/Worker)
## Summary

Switch the on-the-wire encoding of `Actor` and `Worker` records in the Valkey state store from **protojson** to **binary protobuf**. Pre-alpha, no production data, so this is a clean cutover with no migration path needed. This is the cheapest, highest-leverage step toward the **1 billion record** north star: it roughly halves per-record memory and is ~4x faster to encode, with no schema changes and no change to the locking path.

## Motivation

The store keeps every Actor/Worker record resident in RAM with no TTL, so per-record size directly sets the capacity bill. At the 1B target, encoding alone moves the provisioned fleet from **~3.8 TB to ~1.4 TB (~2.2 TB saved)**.

Measurements (full methodology and all five encodings in [`docs/record-encoding-benchmarks.md`](https://github.com/fkautz/substrate/blob/explore/record-encoding-and-state-store/docs/record-encoding-benchmarks.md)):

| | protojson (today) | binary protobuf | binary + field trims |
|---|---|---|---|
| Value size | 609 B | ~321 B | 170 B |
| In-Valkey (w/ overhead) | 731 B | n/a | 283 B |
| Encode CPU | 3124 ns | **721 ns (4x)** | ~same |

Binary protobuf is the necessary-but-not-sufficient first step: a constant-factor win that extends runway. It does **not** change the cost class of holding 1B mostly-idle records in RAM. That is the job of hot/cold tiering, tracked separately (see #12).

## Proposed change

All in `cmd/ateapi/internal/store/ateredis/ateredis.go`:

- **Write path:** `protojson.Marshal` to `proto.Marshal` for Actor and Worker records (Create/Update).
- **Read path:** `protojson.Unmarshal` to `proto.Unmarshal`.

Clean cutover. Since there's no production data, any existing dev keyspace is flushed (records are a re-derivable cache anyway). The version-check logic stays in Go, so the `WATCH`/`MULTI` optimistic-concurrency paths are unaffected. The lock-release Lua (`ateredis.go:578`) is also unaffected: it compares an opaque, ephemeral lock token by byte equality and never touches the record encoding, so the change is invisible to it (verified against a real Valkey cluster, including a token carrying an embedded NUL byte).

## Non-goals / out of scope

- **No change to the lock release Lua** (`ateredis.go:578`). The compare-and-delete is opaque byte equality and is binary-safe (verified against a real Valkey cluster, including an embedded NUL byte in the token).
- **Field trims** (170 B / 283 B) are a separate phase. They involve semantic decisions about droppable/derivable fields and carry regression risk.
- **zstd + dictionary** (112 B) is a different tradeoff (~1 ms/record encode, ~1400x), only justified if RAM becomes the hard binding constraint.
- **Hot/cold tiering** and any **transactional-store** direction are tracked elsewhere.

## Acceptance criteria

- [ ] Actor and Worker records are written and read as binary protobuf.
- [ ] Round-trip equivalence test against a **real Valkey cluster** (not just `miniredis`) for both record types.
- [ ] No change to `store.Interface`; existing store tests pass unmodified.
- [ ] Lock acquire/release behavior unchanged.

## Risks

- **Hard cutover:** any pre-existing dev data won't decode, so flush keyspaces on deploy of this change. Acceptable pre-alpha; must land before any production use.
- **Debuggability:** values in `valkey-cli` are no longer human-readable.
- **Test coverage gap:** `miniredis` can't run cluster commands, so add at least one real-cluster round-trip test.

## References

- Benchmarks: [`docs/record-encoding-benchmarks.md`](https://github.com/fkautz/substrate/blob/explore/record-encoding-and-state-store/docs/record-encoding-benchmarks.md)
- State store design + capacity analysis: [`docs/state-store.md`](https://github.com/fkautz/substrate/blob/explore/record-encoding-and-state-store/docs/state-store.md)
- Tiering direction: #12

fkautz:
## Summary

Switch the on-the-wire encoding of `Actor` and `Worker` records in the Valkey
state store from protojson to binary protobuf. This roughly halves per-record
RAM and is about 4x faster to encode, which extends runway toward the
1-billion-record target. The change is scoped to the resident Actor/Worker
records, as described in #307.

This is a clean cutover with no migration path. Since this is pre-alpha with no
production data, the keyspace is flushed on deploy. protojson and binary
protobuf cannot decode each other, so it is not safe as a rolling update across
the cutover. Deploy by stopping the replicas, flushing the keyspace, then
starting the new version.

## Changes

- Write path: `proto.Marshal` for Actor and Worker (Create and Update).
- Read path: `proto.Unmarshal` (Get, Update, Delete, List).
- The optimistic-concurrency version check stays in Go and compares the decoded
  `version` field rather than the raw bytes, so the `WATCH`/`MULTI` paths are
  unaffected and the encoding is never inspected from Redis Lua.
- Added key-identity guards to the list paths (`fetchActors` and `ListWorkers`).
  Unlike protojson, `proto.Unmarshal` accepts an empty byte slice and returns a
  zero-valued message without an error. `GetActor` and `GetWorker` already
  rejected that case via their identity checks, and the list paths now do too.

## Out of scope (intentionally still protojson)

- Atespace records, which are low cardinality and use negligible RAM at scale.
- The worker-change pub/sub payload, which is ephemeral and never persisted.

## Testing

- Unit (miniredis): a binary-on-the-wire assertion for both record types, which
  catches an accidental revert to protojson, plus rejection of empty, corrupt,
  and identity-mismatched values across both the Get and List paths.
- Integration (`-tags integration`): real-cluster round-trip for both types,
  an optimistic-concurrency conflict for both types, and a binary-safe lock test
  using a token with a NUL byte. These skip unless `ATE_TEST_REDIS_ADDR` points
  at a dedicated, empty cluster, and they refuse to run against a non-empty or
  non-cluster server.

Verified locally against a single-node Valkey 9.1.0 cluster:

    ATE_TEST_REDIS_ADDR=127.0.0.1:7300 go test -tags integration \
        ./cmd/ateapi/internal/store/ateredis/

All unit and integration tests pass, and `go vet` and `gofmt` are clean. There
is no change to `store.Interface`, and existing store tests pass unmodified.

Closes #307.

### DIFF
--- a/cmd/ateapi/internal/store/ateredis/ateredis.go
+++ b/cmd/ateapi/internal/store/ateredis/ateredis.go
@@ -15,12 +15,19 @@
 // Package ateredis is an ate storage backend built on Redis.
 //
 // Actors are stored in keys of the form
-// `actor:<atespace>:<actor-id>`.  They are
-// stored as DBActor JSON-serialized objects, which lets us manipulate them from
-// Redis lua.
+// `actor:<atespace>:<actor-id>`, holding a binary protobuf-serialized Actor
+// record. Binary protobuf roughly halves per-record memory versus protojson and
+// is ~4x faster to encode; the optimistic-concurrency version check stays in Go
+// (comparing the decoded version field, never the raw bytes), so the encoding is
+// never inspected from Redis lua.
 //
 // Workers are stored in keys of the form
-// `worker:<namespace>:<pool-name>:<pod-name>`, holding a DBWorker JSON object.
+// `worker:<namespace>:<pool-name>:<pod-name>`, holding a binary
+// protobuf-serialized Worker record.
+//
+// Only the resident Actor/Worker records use binary protobuf. Atespace records
+// and the worker-change pub/sub payloads intentionally stay on protojson: they
+// are low-cardinality or ephemeral, so they do not drive RAM at scale.
 //
 // Note that redis lua scripting has a restriction that informed the data design
 // here -- a lua script must predeclare all keys it is going to access.  It
@@ -311,7 +318,7 @@ func (s *Persistence) GetActor(ctx context.Context, atespace, id string) (*ateap
 	}
 
 	actor := &ateapipb.Actor{}
-	if err := protojson.Unmarshal(dbActorBytes, actor); err != nil {
+	if err := proto.Unmarshal(dbActorBytes, actor); err != nil {
 		return nil, fmt.Errorf("while unmarshaling actor: %w", err)
 	}
 
@@ -330,9 +337,9 @@ func (s *Persistence) CreateActor(ctx context.Context, actor *ateapipb.Actor) er
 	dbActor := proto.Clone(actor).(*ateapipb.Actor)
 	dbActor.Version = 1
 
-	dbActorBytes, err := protojson.Marshal(dbActor)
+	dbActorBytes, err := proto.Marshal(dbActor)
 	if err != nil {
-		return fmt.Errorf("in protojson.Marshal: %w", err)
+		return fmt.Errorf("in proto.Marshal: %w", err)
 	}
 
 	ok, err := s.rdb.SetNX(ctx, dbKey, dbActorBytes, 0).Result()
@@ -354,9 +361,9 @@ func (s *Persistence) CreateWorker(ctx context.Context, worker *ateapipb.Worker)
 	dbWorker := proto.Clone(worker).(*ateapipb.Worker)
 	dbWorker.Version = 1
 
-	dbWorkerBytes, err := protojson.Marshal(dbWorker)
+	dbWorkerBytes, err := proto.Marshal(dbWorker)
 	if err != nil {
-		return fmt.Errorf("in protojson.Marshal: %w", err)
+		return fmt.Errorf("in proto.Marshal: %w", err)
 	}
 
 	ok, err := s.rdb.SetNX(ctx, dbKey, dbWorkerBytes, 0).Result()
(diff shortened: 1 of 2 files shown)


cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -15,12 +15,19 @@
 // Package ateredis is an ate storage backend built on Redis.
 //
 // Actors are stored in keys of the form
-// `actor:<atespace>:<actor-id>`.  They are
-// stored as DBActor JSON-serialized objects, which lets us manipulate them from
-// Redis lua.
+// `actor:<atespace>:<actor-id>`, holding a binary protobuf-serialized Actor
+// record. Binary protobuf roughly halves per-record memory versus protojson and
+// is ~4x faster to encode; the optimistic-concurrency version check stays in Go
+// (comparing the decoded version field, never the raw bytes), so the encoding is
```
juli4n, line 21: I don't know how relevant this comment is:

"Binary protobuf roughly halves per-record memory versus protojson and
// is ~4x faster to encode; the optimistic-concurrency version check stays in Go
Comment view// (comparing the decoded version field, never the raw bytes), so the encoding is
// never inspected from Redis lua."

Looks more like a point in time commit message than package documentation (e.g. 4x faster than what?).

I would just remove it or turn it into a message that justifies *why* we use binary proto encoding instead of JSON / textproto.

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -15,12 +15,19 @@
...
+// never inspected from Redis lua.
 //
 // Workers are stored in keys of the form
-// `worker:<namespace>:<pool-name>:<pod-name>`, holding a DBWorker JSON object.
+// `worker:<namespace>:<pool-name>:<pod-name>`, holding a binary
+// protobuf-serialized Worker record.
+//
+// Only the resident Actor/Worker records use binary protobuf. Atespace records
```
juli4n, line 28: I think we should be consistent and use the same encoding across all resources. Over time we will add more resources, and having to think about which ones are in which encoding will add significant cognitive burden.

fkautz: Not a problem, I'll move the others over too.

EItanya:
@fkautz would you mind resolving the conflicts, i'd still like to see this land

HavenXia:
+1 One follow-up: with binaryproto, field numbers become permanent identity — renumbering or reusing one silently misreads old records. We should start reserved-ing removed numbers as team practice once we merged this and I will update style guide for it.

HavenXia:
Given we have changed to use Postgres for substrate  and delete Redis, and binary proto is default there, I will close this PR

PR #357, opened 2026-06-30
Add Valkey operations handbook

MushuEE:
Operational reference for the Valkey persistence tier and the in-process worker cache that sits in front of it. Written for operators, on-call engineers, and contributors who need to reason about the storage tier without first reading the code. This will be used as ref for further investigation on if Valkey will really stand-up.

NOTE: We chose Valkey due to it's speed for the critical scheduling requirements. With WorkerCache now being in memory, we get the speed for scheduling and still rely on eventual consistency. Do we now want more durability? Also still need to answer: what data can we 100% NOT lose in these failure situations.

  README.md       — entry point, scope, conventions
  topology.md     — what's deployed, sizing math, configuration knobs
  lifecycle.md    — actor state machine, four workflows, worker cache
                    and eligibility model
  operations.md   — failure modes with inline recovery, common admin
                    operations, open risks

#12 

- [x] Tests pass

### DIFF
--- a/docs/valkey/README.md
+++ b/docs/valkey/README.md
@@ -12,16 +12,22 @@ own documentation.
 
 ## Pages
 
-- [`topology.md`](./topology.md) — what's deployed, how the pieces fit
-  together, sizing math for MVP and target scale, identity / TLS wiring,
-  and the configuration knobs we have not set yet.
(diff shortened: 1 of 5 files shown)


juli4n:
ValKey support was dropped in #940 . Closing this PR as it doesn't apply any longer.

PR #359, opened 2026-06-30
fix: correct field name in api-guide ResumeActor docs

mesutoezdil:
The Actor proto has no worker_ip field. The actual field for the worker IP is ateom_pod_ip (pkg/proto/ateapipb/ateapi.proto:119).
Verified ateom_pod_ip is the field name in ateapi.proto, no worker_ip field exists.


mesutoezdil:
@juli4n could you take a look when you have a moment, small doc fix?

mesutoezdil:
@HavenXia Small doc fix here too, would appreciate a quick look if you have a moment.

PR #363, opened 2026-06-30
Put Worker.actor fields into an "Assignment" message

thockin:
This is a cleanup I noticed as I was reviewing other PRs.  I tried to keep the commits clean, so you can see them one by one.


cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -167,10 +167,14 @@ func (s *WorkerPoolSyncer) releaseActorOnDeadWorker(ctx context.Context, namespa
 		}
 		return err
 	}
-	if worker.GetActorId() == "" {
+	if worker.Assignment == nil {
 		return nil
 	}
-	actor, err := s.persistence.GetActor(ctx, worker.GetActorAtespace(), worker.GetActorId())
+	if worker.Assignment.Actor == nil {
+		slog.ErrorContext(ctx, "BUG: worker has assignment but no actor", slog.Any("worker", worker.String()))
```
juli4n: I'm curious about why not just let it crash under impossible conditions like this. If we add defensive checks, how deep we go? What's the framework to think where we should check?

thockin: Yeah, I waffled on this point and I expected to discuss.

It does seem "impossible" but we don't currently validate Workers very strongly.  We can add that and then let it fail as "impossible".  Let me look at how hard that would be.

thockin: Oh, actually this goes to our other conversation.  There's no CreateWorker() method exposed because the k8s->ate syncer is in-process.  So it is safe now because there is literally 1 caller.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -300,9 +300,13 @@ message Worker {
...
 
 message Assignment {
-  string actor_template_namespace = 1;
-  string actor_template_name  = 2;
-  ActorRef actor = 3;
+  KubeNamespacedObjectRef actor_template = 1;
+  ActorRef actor = 2;
+}
+
+message KubeNamespacedObjectRef {
```
HavenXia, line 307: Why are we using such generic naming, is it going to used by other message other than Assignment?

cmd/ateapi/internal/controlapi/syncer.go
thockin: I'll do some prefactoring for validation logic in this PR, hang tight

pkg/proto/ateapipb/ateapi.proto
thockin: I see a few other places, though really we might just make these all go away.  This data is only used in one demo

cmd/ateapi/internal/controlapi/syncer.go
juli4n: > Oh, actually this goes to our other conversation. There's no CreateWorker() method exposed because the k8s->ate syncer is in-process. So it is safe now because there is literally 1 caller.

Whether we sync workers via an in-process goroutine or via a gRPC API (as we will do eventually) is orthogonal to this point.

The invariant which turns this into an impossible condition is that when we assign a worker to an actor, we never partially assign a Assignment proto, this enforced by the ResumeActor workflow.

thockin: Sure, I was just looking for where CreateActor() gets validated, and I didn't find it because it's entirely internal :)

HavenXia (dismissed):

cmd/ateapi/internal/controlapi/syncer.go
thockin: OK, this update expanded this PR a bit, but the commits are still individually complete (I hope).

I overhauled how actor validation works.  There's more to do but it's more complete now.  And added tests.

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -114,14 +117,28 @@ func (s *WorkerPoolSyncer) syncWorkerToStore(ctx context.Context, pod *corev1.Po
...
+			// move there.
+			if errs := resources.ValidateWorker(worker, nil); len(errs) > 0 {
+				var errstr string
+				if len(errs) == 1 {
+					errstr = errs[0].Error()
+				} else {
+					errstr = errs.ToAggregate().Error()
+				}
+				err := status.Error(codes.InvalidArgument, errstr)
+				slog.ErrorContext(ctx, "Invalid worker", slog.Any("err", err))
```
juli4n, line 133: Is this intentional? I was expecting you would return from here instead of storing an invalid worker.

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -114,14 +117,28 @@ func (s *WorkerPoolSyncer) syncWorkerToStore(ctx context.Context, pod *corev1.Po
...
 				Ip:              pod.Status.PodIP,
 				WorkerPodUid:    string(pod.UID),
 				NodeName:        pod.Spec.NodeName,
-			})
+			}
+			// TODO(thockin): for now this is the only place Workers are
+			// created.  If/when this becomes a regular API, validation should
+			// move there.
+			if errs := resources.ValidateWorker(worker, nil); len(errs) > 0 {
+				var errstr string
```
juli4n: Why handling `len(errs)==1` ? Looking at `ToAggregate()` it handles this case just fine.

Just do this?

```
err := status.Error(codes.InvalidArgument, errs.ToAggregate().Error())
```

internal/resources/validate.go
```diff
@@ -136,3 +170,149 @@ func ValidateSnapshotURIPrefix(prefix string) error {
...
+	//TODO(thockin): prevent localhost and link-local addresses which might confuse callers?
+
+	return errs
+}
+
+// ValidateUUID verifies that the specified value is a valid UUID (RFC 4122).
+//   - must be 36 characters long
+//   - must be in the normalized form `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`
+//   - must use only lowercase hexadecimal characters
+func ValidateUUID(uuid string, fldPath *field.Path) field.ErrorList {
```
juli4n, line 296: Is there a reason to write our own UUID validation? We already depend on github.com/google/uuid, can we just use it?

cmd/ateapi/internal/controlapi/syncer.go
thockin: right, I forgot that is handled deep within

cmd/ateapi/internal/controlapi/syncer.go
thockin: not intentional!

internal/resources/validate.go
thockin: To get consistent error messages and behavior.

thockin:
Now with tests that should pass.  Unsurprisingly, adding validation caused sketchy tests to break.

juli4n (approved):

thockin:
Reviewer merges

PR #364, opened 2026-06-30
Activate python virtual environments for scripts that operate within them

maxsmythe:
This makes sure the environment is more insulated from idiosyncratic local machine configurations.

### DIFF
--- a/hack/verify/python-licenses.sh
+++ b/hack/verify/python-licenses.sh
@@ -52,36 +52,27 @@ mapfile -t PROJECTS < <(
 # vs. PEP-639 expression).
 DISALLOWED="AGPL;GPLv2;GPLv3;GPL-2.0;GPL-3.0;GNU General Public License;GNU Affero General Public License;Server Side Public License;SSPL;Commons Clause"
 
-ensure_venv() {
-  local dir="$1"
-  local venv="${dir}/venv"
-  if [[ ! -d "${venv}" ]]; then
-    echo "Creating venv at ${venv}..."
-    python3 -m venv "${venv}" || return 1
-    "${venv}/bin/pip" install --quiet --upgrade pip || return 1
-    "${venv}/bin/pip" install --quiet -r "${dir}/requirements.txt" || {
-      echo "ERROR: failed to install ${dir}/requirements.txt into ${venv}" >&2
-      return 1
-    }
-  fi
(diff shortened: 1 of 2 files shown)


hack/verify/python-licenses.sh
```diff
@@ -52,36 +52,27 @@ mapfile -t PROJECTS < <(
...
 check_project() {
   local dir="$1"
   local venv="${dir}/venv"
-  ensure_venv "${dir}" || return 1
   echo "==> ${dir}"
-  # --ignore-packages skips the tools themselves so the venv doesn't fail
-  # itself; pip / setuptools / pip-licenses are infrastructure.
-  "${venv}/bin/pip-licenses" \
-    --fail-on="${DISALLOWED}" \
-    --ignore-packages pip setuptools pip-licenses prettytable wcwidth tomli
```
zoez7, line 84: Did we mean to drop the `    --ignore-packages pip setuptools pip-licenses prettytable wcwidth tomli` argument?

hack/verify/python-licenses.sh
```diff
@@ -52,36 +52,27 @@ mapfile -t PROJECTS < <(
...
-    python3 -m venv "${venv}" || return 1
-    "${venv}/bin/pip" install --quiet --upgrade pip || return 1
-    "${venv}/bin/pip" install --quiet -r "${dir}/requirements.txt" || {
-      echo "ERROR: failed to install ${dir}/requirements.txt into ${venv}" >&2
-      return 1
-    }
-  fi
-  if ! "${venv}/bin/pip" show pip-licenses >/dev/null 2>&1; then
-    "${venv}/bin/pip" install --quiet pip-licenses || {
-      echo "ERROR: failed to install pip-licenses into ${venv}" >&2
```
zoez7, line 69: Can we preserve these detailed error messages as well?

zoez7 (dismissed):

hack/verify/python-licenses.sh
maxsmythe: Yep, figured better safe than sorry and those packages' licenses are currently compatible with the check.

hack/verify/python-licenses.sh
maxsmythe: Good point, added.

maxsmythe:
Thanks @zoez7. Should be good for another pass.

PR #370, opened 2026-07-01
refactor: remove redundant SnapshotType enum and use protobuf oneof d…

dberkov:
refactor: remove redundant SnapshotType enum and use protobuf oneof d…

Code cleaning. 

- [X] Tests pass

### DIFF
--- a/pkg/proto/ateapipb/ateapi.proto
+++ b/pkg/proto/ateapipb/ateapi.proto
@@ -77,11 +77,13 @@ message LocalSnapshotInfo {
 }
 
 message SnapshotInfo {
-  reserved 1;
   oneof data {
     ExternalSnapshotInfo external = 2;
     LocalSnapshotInfo    local    = 3;
   }
+
+  // type
+  reserved 1;
 }
 
 // Selector matches worker pools by label.
--- a/pkg/proto/ateapipb/ateapi.proto
+++ b/pkg/proto/ateapipb/ateapi.proto
@@ -66,12 +66,6 @@ service Control {
   rpc DebugClear(DebugClearRequest) returns (DebugClearResponse) {}
 }
 
-enum SnapshotType {
-  SNAPSHOT_TYPE_UNSPECIFIED = 0;
-  SNAPSHOT_TYPE_LOCAL       = 1;
-  SNAPSHOT_TYPE_EXTERNAL      = 2;
-}
-
 message ExternalSnapshotInfo {
   string snapshot_uri_prefix = 1;
 }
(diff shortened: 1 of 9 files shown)


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -83,7 +77,7 @@ message LocalSnapshotInfo {
 }
 
 message SnapshotInfo {
-  SnapshotType type = 1;
+  reserved 1;
```
ahmedtd: Also reserve the name: `reserved "type"`;

And conventionally the reserved blocks go at the end.

cmd/atelet/main.go
```diff
@@ -905,12 +905,35 @@ func validateActorRequest(namespace, template, actorID, targetAteomUID string, s
...
 }
 
-// writeFileAtomic writes data to path by writing a temp file in the same
-// directory, syncing, and renaming it over the target, then syncing the
+// writeFileAtomic writes data to path atomically vs concurrent readers via
+// temp file + rename. It does NOT fsync: the identity directory is regenerated
+// on every actor resume, so durability across power loss isn't required and
+// the fsync latency (hundreds of ms on the backing volume) isn't worth paying.
+// perm is applied at open time and is subject to the process umask.
+func writeFileAtomic(path string, data []byte, perm os.FileMode) error {
```
ahmedtd: What's going on with this?  This seems unrelated, and pretty dubious

dberkov: oops, I pushed 33000542f50363f50c8364e987b7ed6c5fbd5799 to wrong branch. Removed it from this branch.

pkg/proto/ateapipb/ateapi.proto
dberkov: fixed by adding "type" in the comment and move block to the end.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -83,11 +77,13 @@ message LocalSnapshotInfo {
 }
 
 message SnapshotInfo {
-  SnapshotType type = 1;
   oneof data {
     ExternalSnapshotInfo external = 2;
     LocalSnapshotInfo    local    = 3;
   }
+
+  // type
```
ahmedtd: Reserved the name as well: `reserved "type";`

ahmedtd (dismissed):

pkg/proto/ateapipb/ateapi.proto
ahmedtd: Oh, I commented this again.

Don't just put the name in the comment --- protobuf can reserve the name, so that no one ever accidentally adds a field with the same name again (which will break cleartext serializations, like JSON).

pkg/proto/ateapipb/ateapi.proto
dberkov: nice :)  was not aware about it.

PR #373, opened 2026-07-01
Propagate (atespace/actor_id) to atelet.

juli4n:
This fixes an issue where atelet/ateom will clash the state of actors with same ID within different atespaces.

Once substrate resources get UIDs, we should revisit the way atelet / ateom does the bookkeeping and check in which cases it should use UID instead of (atespace,actor_id).

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -8,6 +8,7 @@ This guide explains how Agent Substrate achieves observability across these susp
 
 To make underlying infrastructure transitions transparent, Agent Substrate establishes a standardized metadata model to identify actors across worker pods:
 * `ate.dev/actor_id`: The unique identifier of the actor (e.g., `my-counter-1` or `test`).
+* `ate.dev/actor_atespace`: The atespace the actor lives in (e.g., `ate-demo-counter`).
 * `ate.dev/actor_template_name`: The name of the actor's ActorTemplate (e.g., `counter`).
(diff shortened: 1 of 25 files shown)


internal/actorlog/logger.go
HavenXia, line 1: Adding `actor_atespace` to log labels LGTM — logs are the right layer for per-atespace attribution.

But could we keep `actor_template_name` (`actor_template_namespace` can be removed) on the log labels in addition, rather than replacing them? 

"Show my logs from all actors of template X" is a real debugging workflow IMO, and since one atespace can run actors from many templates, atespace itself can't answer it.

docs/observability.md
```diff
@@ -75,11 +74,11 @@ To track the unified, continuous lifecycle of a single actor regardless of how m
 labels.actor_id="test"
 ```
 
-#### 2. Template-Centric View
-To monitor or debug all actor instances created from a specific template (e.g., analyzing the collective behavior or error rates of all counter actors):
+#### 2. Atespace-Centric View
```
HavenXia, line 79: Same to my another comment.

internal/actorlog/logger.go
juli4n: Done.

docs/observability.md
juli4n: Done.

HavenXia (approved):

PR #374, opened 2026-07-02
Add CheckPrerequisite to WorkflowStep to validate state machine edges

### LINKED ISSUE #369: [Actor state machine] Add CheckPrerequisite steps to WorkflowStep definition
See background discussion thread: https://github.com/agent-substrate/substrate/pull/353/changes#r3496338513

Add a CheckPrerequisite step to WorkflowStep definition, and validate the prerequisite of each edge in the state machine: https://github.com/agent-substrate/substrate/issues/119#issuecomment-4584936058

zoez7:
Fixes #369

Add a CheckPrerequisite method to the WorkflowStep interface, called by RunWorkflow after IsComplete returns false and before Execute, so that each workflow validates its actor state-machine edge up front while retried (reentrant) workflows still fast-forward past completed steps.


cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -82,40 +86,73 @@ func (s *LoadActorForResumeStep) Execute(ctx context.Context, input *ResumeInput
...
 
 func (s *AssignWorkerStep) IsComplete(ctx context.Context, input *ResumeInput, state *ResumeState) (bool, error) {
+	// Only RUNNING is past this step. RESUMING intentionally re-runs because
+	// a retry must be able to release a stale worker whose pool became
+	// ineligible and pick a fresh one.
 	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_RUNNING, nil
 }
+func (s *AssignWorkerStep) CheckPrerequisite(ctx context.Context, input *ResumeInput, state *ResumeState) error {
+	// The resume edge exists from SUSPENDED and PAUSED.
+	// RESUMING is allowed for retrying this step.
```
zoez7: This is because of line 168 below.

cmd/ateapi/internal/controlapi/workflow_suspend.go
```diff
@@ -154,8 +167,13 @@ type FinalizeSuspendedStep struct {
 
 func (s *FinalizeSuspendedStep) Name() string { return "FinalizeSuspended" }
 func (s *FinalizeSuspendedStep) IsComplete(ctx context.Context, input *SuspendInput, state *SuspendState) (bool, error) {
-	// The workflow is completely done ONLY if the status is SUSPENDED *and* we've successfully freed the worker.
-	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_SUSPENDED && state.Actor.GetAteomPodNamespace() == "", nil
```
zoez7, line 167: I removed this check, because we actually update the Actor's status and worker assignment  in the same UpdateActor request in Execute() below, so we only need to check if status==suspended.

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -106,6 +114,12 @@ func (s *CallAteletPauseStep) IsComplete(ctx context.Context, input *PauseInput,
...
 	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_PAUSED, nil
 }
+func (s *CallAteletPauseStep) CheckPrerequisite(ctx context.Context, input *PauseInput, state *PauseState) error {
+	if state.Actor.GetStatus() != ateapipb.Actor_STATUS_PAUSING {
+		return status.Errorf(codes.FailedPrecondition, "CallAteletPauseStep prerequisite not met for Actor: %s (got: %v, want %s)", input.ActorName, state.Actor.GetStatus(), ateapipb.Actor_STATUS_PAUSING)
+	}
+	return nil
+}
 func (s *CallAteletPauseStep) Execute(ctx context.Context, input *PauseInput, state *PauseState) error {
 	if state.Actor.GetAteomPodNamespace() == "" || state.Actor.GetAteomPodName() == "" {
```
dberkov, line 121: this is a pre-requisite  too, however function in current implementation does not support transition to crashActor.

Might be rename the "CheckPrerequisite" to different name? might be allow transition to crash? or might be from the beginning we were not supposed to be in this state, that actor does not have pod or namespace?

cmd/ateapi/internal/controlapi/workflow_pause_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// TestPauseActorWorkflow exercises the pause workflow end-to-end against
+// seeded actor statuses, covering both the rejected and the idempotent-success
+// paths. The atelet dialer is nil, so any step that unexpectedly reaches it
+// panics.
+func TestPauseActorWorkflow(t *testing.T) {
```
dberkov: TestPauseSteps_CheckPrerequisite validates Prerequisite that matches name of the test. 
In this case, the name is generic for entire Step, however it validates  CheckPrerequisite  transition only.

cmd/ateapi/internal/controlapi/workflow_resume.go
dberkov: could you please clarify what is special at line 168? Is it for cases it was an error in resuming process and we are OK to try resuming one more time?

Same case might happening with Pausing and the CheckPrerequisite for CallAteletPauseStep does not accept pausing state.

cmd/ateapi/internal/controlapi/workflow_pause.go
zoez7: We should crash the actor here. We can only get here if MarkPausing has succeeded, not having the ateom pod or namespace should crash the actor. I changed the implementation.

cmd/ateapi/internal/controlapi/workflow_pause_test.go
zoez7: I renamed the test to more specific name.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: I updated the logic here and stopped allowing resuming state to reenter the AssignWorkerStep. Because setting the state to Resuming is the last step of AssignWorkerStep, if the actor is already in resuming, we know it has completed all the steps and don't need to retry.

On line 168, we would check which worker is assigned and if the worker is no longer eligible, we will pick another one, but this is for cleaning up and doesn't have to be retried during this operation.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -80,6 +82,36 @@ func (s *LoadActorForResumeStep) Execute(ctx context.Context, input *ResumeInput
 	}
 	state.ActorTemplate = actorTemplate
 
+	// If the Actor is in Resuming state, it means a previous attempt crashed after AssignWorkerStep.
+	// We don't need to repeat the AssignWorkerStep, load the Worker now.
```
zoez7, line 85: I added this step so if we retry resumeActor for Actor already in RESUMING state, it will still load worker to the ResumeState

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7, line 306: I added more checks in CheckPrerequisite to make sure that, if we retry ResumeActor, the Worker assignment and eligibility are still correct for the actor.

PR #377, opened 2026-07-02
Docs: Add atespaces to local quickstart readme

### LINKED ISSUE #376: Docs: Quickstart guide broken due to missing atespaces
When following the [quickstart guide in the root README](https://github.com/agent-substrate/substrate/blob/main/README.md?plain=1#L97), step 5 fails to create an actor:

```
Error: required flag(s) "atespace" not set
```

The guide should be updated to account for:
- Creating the atespace
- Creating the actor IN that atespace
- Updating the curl to use the atespace in the request

I'm happy to open a PR to fix this.

Sodman:
Existing instructions fail due to lack of atespaces, this PR adds a step to create on (default) and then updates the actor creation command as well as the following curl command to account for it.

Fixes #376

~- [ ] Tests pass~ N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,8 +114,11 @@ hack/install-ate-kind.sh --deploy-demo-counter
 # install kubectl-ate
 go install ./cmd/kubectl-ate
 
+# create an atespace to deploy an actor into
+kubectl ate create atespace default
+
 # create a counter actor and demo it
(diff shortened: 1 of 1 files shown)


HavenXia (approved):
Thank you for the quick fix!

PR #380, opened 2026-07-03
fix(atelet): size cache guard by image size, not manifest size

chenggui53:
memorypullcache.Fetch guarded its in-memory cache with img.Size() > 100MiB to avoid holding large images in RAM. But go-containerregistry img.Size() returns the manifest size (a few KB), not the image size, so the guard never triggered for any real image — every image, including multi-GB ones, was fully read into memory via io.ReadAll, risking atelet OOM.

Sum the compressed layer sizes from the manifest and compare that against the threshold instead. Compressed size is used (rather than the extracted size actually held in memory) because the extracted size is not recorded in any image metadata and could only be learned by downloading and decompressing every layer — the very work this guard avoids for large images; the compressed sum is a conservative lower bound available from the manifest alone. The threshold is extracted into a named constant.

### DIFF
--- a/cmd/atelet/internal/memorypullcache/memorypullcache.go
+++ b/cmd/atelet/internal/memorypullcache/memorypullcache.go
@@ -32,6 +32,20 @@ import (
 	"k8s.io/utils/lru"
 )
 
+// maxCacheableImageBytes bounds which images are held in the in-memory pull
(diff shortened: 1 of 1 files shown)


chenggui53:
@thockin please help review this fix, thanks ~

PR #355, opened 2026-06-30
Update README.md and scripts to adopt atespaces for existing demos.

HavenXia:
Update README.md and scripts to adopt atespaces for existing demos.

It's intentional to let user choose atespace in demos since actors are also created by them.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

PR #382, opened 2026-07-03
Fix demo actor cleanup for atespaces

haoqixu:
Pass the required `atespace` argument to `kubectl ate` in demo actor cleanup.


haoqixu:
Close this PR because #355 already fixes this script.

PR #389, opened 2026-07-03
POC: use k8s validation-gen

thockin:
This is a little hacky and it depends on some uncommitted changes to k8s tools (https://github.com/kubernetes/kubernetes/pull/140204) , but it works! 

@juli4n given the unreleased changes, this is too early to merge now, but what do you think directionally?

@maxsmythe can you speak to internal/ateinterceptors/ateinterceptors.go - we're disassembling an error and then reassembling it, losing the details.  This copying works but is kind of yuck...

@jpbetz FYI


pkg/proto/ateapipb/validate.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+
+	codes "google.golang.org/grpc/codes"
+	status "google.golang.org/grpc/status"
+	operation "k8s.io/apimachinery/pkg/api/operation"
+)
+
+func (x *CreateActorRequest) Validate(ctx context.Context) error {
+	op := operation.Operation{Type: operation.Create}
```
juli4n, line 26: I guess for updates we will apply the mutation to the resource (using the field mask) and then validate as a an Update operation?

We don't have an update method with field mask yet, but it sounds like something that would be good to prototype here as well. I'm sure things will come up (e.g. merge / validation semantics for non-scalar values, basically lists and maps).

cmd/ateapi/internal/controlapi/functional_test.go
```diff
@@ -1891,6 +1891,7 @@ func TestValidation(t *testing.T) {
 	tc := setupTest(t, ns)
 	defer tc.cleanup()
 
+	// FIXME: this should move next to the proto files and the validate method
```
juli4n, line 1894: I actually like having a test here for validation as it not only asserts that the validator works, but also that the gRPC server (as we whole) is propagating the Go error to the right gRPC status response.

pkg/proto/ateapipb/validate.go
thockin: Yeah, that's how kube works - update-oriented validators get the "old" and "new" state of the resource, so they can consider the change.  There's also an automatic rule that we never re-validate saved data.  If it was allowed into the database, it is, by definition, valid.  This means a) we save cycles re-validating known-good data and b) a user should never get a validation error for something they didn't change.

pkg/proto/ateapipb/validate.go
```diff
@@ -0,0 +1,47 @@
...
+	errs := Validate_CreateActorRequest(ctx, op, nil, x, nil)
+
```
juli4n, line 28: qq, how do you envision more complex validations (e.g. cross-field) being implemented? Would that be a custom validation referenced from a tag in the field comment or just an extra call from here to the custom validator function?

cmd/ateapi/internal/controlapi/functional_test.go
thockin: I ACK the value of complete testing, but also point out that doing rigorous-but-not-brittle testing requires things like something like "origin" which really shouldn't be part of the user-facing API but is fine for internal API.

pkg/proto/ateapipb/validate.go
thockin: Declarative validation supports some cross-field things like one-of or contingent evaluation.  It's incomplete, but they need to solve that anyway, so we're in the same boat.  Finally there is an escape hatch for "this is weird, so we wrote some code by hand", and there MAY eventually be a "this is weird, here's some CEL"

cmd/ateapi/internal/controlapi/functional_test.go
juli4n: How about we move all the exhaustive test cases next to the proto / validate methods, and make it assert on the full (type, field, detail, origin) tuple, but then scope down `ValidationError` (or the AIP equivalent `FieldViolation`[1]) to what we actually want to put on the write (I assume origin, for example, is not meant to be part of the API contract). Then we have a thin layer of tests here just to make sure we properly map validation errors to gRPC status.

[1]: https://github.com/googleapis/googleapis/blob/ba4573adabf42e4e0bb9877f11de8c8f1005aa0a/google/rpc/error_details.proto#L238

thockin: That seems plausible. I added ValidationError because FieldValidation could not carry origin.

internal/ateinterceptors/ateinterceptors.go
```diff
@@ -61,7 +62,23 @@ func ServerUnaryInterceptor(ctx context.Context, req any, info *grpc.UnaryServer
 
 		if errors.As(err, &statusErr) {
 			st := statusErr.GRPCStatus()
-			return nil, status.Error(st.Code(), st.Message())
+			//FIXME: investigate why we are doing this error deconstruction
```
maxsmythe, line 65: Not 100% sure why this is here. It looks like it has the effect of sanitizing internal details from the response by unwrapping the error? But then it could just return the unwrapped error directly and avoid copying details.

Also, if that's the goal, then returning the raw error at the bottom still can leak internals.

Git blame shows @ahmedtd as committer, adding him in case there is a Chesterton's Fence issue.

cmd/ateapi/internal/controlapi/create_actor.go
```diff
@@ -30,7 +29,7 @@ import (
 )
 
 func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequest) (*ateapipb.CreateActorResponse, error) {
-	if err := validateCreateActorRequest(req); err != nil {
+	if err := req.Validate(ctx); err != nil {
```
jpbetz, line 32: Looks like a good place to try out DV.

Later would we hoist this call into `ServerUnaryInterceptor` via an interface assertion (`if v, ok := req.(interface{ Validate(context.Context) error }); ok { ... }`) so every RPC gets validated uniformly and individual handlers can't forget to call it?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -59,17 +58,27 @@ service Control {
...
 
   // Debugging: drop all data from the ate database.
   rpc DebugClear(DebugClearRequest) returns (DebugClearResponse) {}
 }
 
+// FIXME: comments
+message ValidationError {
+  string Type = 1;
+  string Field = 2;
+  string Detail = 3;
```
jpbetz, line 73: nit: proto style wants lower_snake_case field names (`type`, `field`, `detail`, `origin`).  This would also leak Type/Field/... into the JSON names.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -86,42 +95,46 @@ message SnapshotInfo {
...
 }
 
 // Selector matches worker pools by label.
 // Only equality-based matching is supported.
 message Selector {
+  // +k8s:optional
+  // +k8s:maxProperties=10
+  // +k8s:alpha(since: "0.0")=+k8s:eachKey=+k8s:format=k8s-label-key
+  // +k8s:alpha(since: "0.0")=+k8s:eachVal=+k8s:format=k8s-label-value
   map<string, string> match_labels = 1;
```
jpbetz, line 109: Looks like `validate.MaxProperties` is public on k/k master now, which should fix the validation-gen for `k8s:maxProperties`.  We'll need to pull that in before merging this.

I think we can drop `+k8s:alpha(since: "0.0")` here now too.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -59,17 +58,27 @@ service Control {
...
   // Debugging: drop all data from the ate database.
   rpc DebugClear(DebugClearRequest) returns (DebugClearResponse) {}
 }
 
+// FIXME: comments
+message ValidationError {
+  string Type = 1;
+  string Field = 2;
+  string Detail = 3;
+  string Origin = 4; // FIXME: passing this back up to users is ugly, but we
```
jpbetz, line 74: Looks like we can avoid exposing origin to users here we could just use `google.rpc.BadRequest.FieldViolation`.

internal/ateinterceptors/ateinterceptors.go
jpbetz: +1 to simplify to `return nil, st.Err()` unless we `status.Error(st.Code(), st.Message())` detail stripping is needed.

Nudge @ahmedtd

thockin: This logic was touched by Taahir but not created by him.  It actually predates the original OSS drop, in a commit by @juli4n with very vague message.  I will poke a bit more.

thockin:
superceded

PR #391, opened 2026-07-04
perf: cache OIDC verification keys in k8sjwt

botengyao:
Previously every `k8sjwt.Verify` call performed a full OIDC discovery (two HTTP round trips: the discovery document plus the JWKS) against the issuer. In JWT auth mode that cost was paid on every authenticated RPC, and MintJWT paid it again for the client token.

Introduce `k8sjwt.Verifier`, which caches each issuer's verification keys in memory. Keys are only refetched when a JWT presents an unknown key ID (i.e. on key rotation). Since the triggering key ID comes from an unverified token, refetches are rate-limited to one per issuer per 10s so bogus key IDs cannot force a fetch storm, and concurrent misses are coalesced with `singleflight`. The auth interceptor and the session-identity service share one Verifier.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/k8sjwt/k8sjwt.go
+++ b/cmd/ateapi/internal/k8sjwt/k8sjwt.go
@@ -31,7 +31,10 @@ import (
 	"net/http"
 	"slices"
 	"strings"
+	"sync"
 	"time"
+
+	"golang.org/x/sync/singleflight"
 )
 
 // KeyAndID wraps a crypto.PublicKey along with the key ID that will identify it during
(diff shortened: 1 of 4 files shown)


cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -115,17 +118,103 @@ var (
...
+		httpClient: httpClient,
+		issuers:    make(map[string]*issuerKeys),
+	}
+}
+
+// keyForIssuer returns the issuer's verification key with the given key ID,
+// (re)fetching the issuer's keys if the key ID is not already cached.
+func (v *Verifier) keyForIssuer(ctx context.Context, issuer, keyID string, now time.Time) (crypto.PublicKey, error) {
+	v.mu.Lock()
+	state, ok := v.issuers[issuer]
```
bowei, line 180: add a comment 

// Can't defer Unlock() as we need to manually manage it for throttling.

cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -115,17 +118,103 @@ var (
...
+// issuer. The key ID that triggers a refetch comes from an unverified token,
+// so without this floor a client could force an OIDC discovery round-trip per
+// request just by minting tokens with bogus key IDs.
+const keyRefetchInterval = 10 * time.Second
+
+// Verifier verifies Kubernetes JWTs, caching each issuer's verification keys
+// in memory. Keys are fetched on first use and refetched only when a JWT
+// presents an unknown key ID (i.e. on key rotation), rate-limited to one
+// fetch per issuer per keyRefetchInterval.
+type Verifier struct {
```
bowei: maybe call this CachedVerifier? 

Verifier would then be an interface type if we need it at some later point.

cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -115,17 +118,103 @@ var (
...
+	httpClient *http.Client
+
+	// flight coalesces concurrent key fetches for the same issuer.
+	flight singleflight.Group
+
+	mu      sync.Mutex
+	issuers map[string]*issuerKeys
```
bowei, line 153: Is it a concern that we could end up with a lot of entries cached over time? Do we need to proactively clear out old entries at all?

cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -115,17 +118,103 @@ var (
...
+		v.issuers[issuer] = state
+	}
+	if key := findKey(state.keys, keyID); key != nil {
+		v.mu.Unlock()
+		return key.PublicKey, nil
+	}
+	throttled := !state.lastFetch.IsZero() && now.Sub(state.lastFetch) < keyRefetchInterval
+	v.mu.Unlock()
+	if throttled {
+		return nil, fmt.Errorf("unknown key ID %q", keyID)
```
bowei: nit: key ID -> keyID (use exact same string when possible, make it easier to search for)

bowei: I think the error message should include the fact that this came from throttling:

fmt.Errorf("keyID %q not found in cache, request throttled, last fetch was at %v

cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -115,17 +118,103 @@ var (
...
+		// failing issuer isn't hammered either.
+		state.lastFetch = now
+		if err != nil {
+			return nil, err
+		}
+		state.keys = keys
+		return nil, nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("while discovering keys from issuer: %w", err)
```
bowei: include the keyID?

cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -115,17 +118,103 @@ var (
...
+	}
+
+	v.mu.Lock()
+	key := findKey(state.keys, keyID)
+	v.mu.Unlock()
+	if key == nil {
+		return nil, fmt.Errorf("unknown key ID %q", keyID)
```
bowei: unknown keyID

cmd/ateapi/internal/k8sjwt/k8sjwt.go
botengyao: good call, done.

cmd/ateapi/internal/k8sjwt/k8sjwt.go
botengyao: done.

cmd/ateapi/internal/k8sjwt/k8sjwt.go
botengyao: done

cmd/ateapi/internal/k8sjwt/k8sjwt.go
botengyao: done

cmd/ateapi/internal/k8sjwt/k8sjwt.go
botengyao: nice call, done

cmd/ateapi/internal/k8sjwt/k8sjwt.go
botengyao: good point, the eviction for issuer should be bounded by the config now (the expected issuer), but there could be a real gap - revoked keys never aged out and added a 5-minute max-age so revoked keys stop being trusted within one interval, PTAL.

Issue #326, opened 2026-06-26
Pluggable Ingress Support


EItanya:
An initial implementation of this was added in #85, but that PR has not moved, so I wanted to revisit that in issue form to try and gather consensus.

## Background

Today Substrate ingress is tightly coupled to the bundled `atenet-router` deployment:

- `manifests/ate-install/atenet-router.yaml` deploys one Substrate router container plus one proxy sidecar.
- `cmd/atenet/internal/router` owns proxy configuration, deployment reconciliation, external request processing, actor ID extraction, `ResumeActor`, and upstream rewrite behavior.
- The current HTTP path assumes the actor ID is encoded in `Host` as `<actor-id>.actors.resources.substrate.ate.dev`.
- The request is routed by resuming the actor through `ate-api-server`, taking the returned worker pod IP, and rewriting the request target to `<worker-ip>:80`.

This works for the bundled demo and development path, but it makes it hard to integrate Substrate with other gateway stacks, managed load balancers, service meshes, platform gateways, or custom infrastructure.

## Problem

Substrate's actor activation and routing decision logic is valuable independently from the current bundled proxy deployment. Operators should be able to choose how traffic enters while reusing the same Substrate semantics:

1. Extract actor identity from a request.
2. Resume or locate the actor.
3. Return a routable backend endpoint.
4. Preserve observability, status, errors, and security controls.
5. Avoid putting Kubernetes control-plane convergence on the request path.

Right now these concerns are mixed together in `cmd/atenet/internal/router`, so replacing the ingress layer means replacing or forking core routing logic.

## Proposal

Introduce a pluggable ingress/router boundary for Substrate.

At a high level, split the current router into:

1. **Substrate routing core**
   - request metadata -> actor ID
   - actor ID -> `ResumeActor`
   - actor status/backend validation
   - route result / error mapping
   - metrics and tracing

2. **Ingress adapters**
   - current bundled proxy adapter
   - adapters for any gateway/proxy that can delegate request processing to Substrate
   - ExtProc-compatible integrations
   - possible future generic gRPC/HTTP route-resolution APIs for integrations that do not use ExtProc

The current implementation should remain the default install path for now, but it should be treated as one adapter, not as the architecture. The design should not assume any specific gateway implementation. In particular, it should avoid baking any specific proxy's xDS, filter-chain, deployment, or configuration concepts into the Substrate routing API.

Substrate's stable contract should be about actor-aware routing:

1. Extract actor identity from request metadata.
[...]

PR #393, opened 2026-07-06
POC: pluggable ingress for substrate

supreme-gg-gg:
POC for #326, based on the design in https://docs.google.com/document/d/1djurA9d7f7UIW2DhSqyX85Ad8JgkpPPyCbpDy7G7HOs/edit?tab=t.r12pfwjkq9tf#heading=h.68j4yb47rvbt

## Key Changes

 - Removed bundled Envoy/xDS proxy from atenet-router, which now only runs actor-aware routing core + ExtProc server
  - `manifests/ate-install/atenet-router.yaml` is the single (extproc-only) router manifest
  - Examples and docs for deploying Envoy and Agentgatweay using standalone or Gateway API configuration
  - `hack/install-ate.sh` deploys the basic reference Envoy gateway alongside the router by default, tests are updated
  - DNS controller's default ingress target flipped from atenet-router to atenet-gateway (router no longer serves HTTP)


cmd/atenet/internal/router/route_errors.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2026 Google LLC
```
bowei, line 1: Is there a reason why some of the files were renamed? I'm not against refactoring, but if you can keep the meaningful changes to a minimal set. If we rename files, ideally you would contain that in its own commit.

supreme-gg-gg: I don't think they were renamed, at least not literally. The motivation here is to move from Envoy specific type (which was in `errors.go`) to a protocol neutral interface in `route.go, route_resolve.go, route_errors.go`, and that's why some old files are replaced by new ones. If there's any ways to simplify the refactor that you would suggest I'm glad to do that

bowei: Right, but the bodies of the functions remained almost the same:

Example: the big switch statement. If you keep the same name as the previous file (errors.go), then the diff will be obvious. Otherwise to review, I will have to manually do a diff between files.

If you do this as separate commits, then it will be much clearer what is going on. Makes reviews faster...

EItanya:
Closing in favor of #461

PR #400, opened 2026-07-07
Set version via ldflags in ko

### LINKED ISSUE #271: Builds done with `make` are inconsistent with `hack/install-ate.sh`
The builds done with make have LD-based version stamping.
hack/install-ate.sh does not currently support this. We should make everything consistent somehow for least surprise. We could:

* Use Go lang version stamping instead of LDFLAGS based one
* Funnel everything through one way of doing things and remove the alternatives.

https://github.com/agent-substrate/substrate/pull/162#pullrequestreview-4490191193

thockin:
This depends on a bump of ko.

Fixes #271


botengyao (approved):

EItanya (approved):

thockin:
self-merge to clear it from the queue

PR #401, opened 2026-07-07
fix: explain locality conflicts instead of 'no free workers available'

### LINKED ISSUE #398: Locality-restricted resume failures report a misleading "no free workers available" error
### Summary

When an actor's latest snapshot is a **local** checkpoint, `AssignWorkerStep.findFreeWorker` hard-restricts candidate workers to the node(s) listed in `LatestSnapshotInfo.Local.NodeVmsWithLocalSnapshots` (correctly — a local checkpoint only exists on the node that wrote it). But when that restriction is what eliminates all candidates, the returned error is the generic:

```
FailedPrecondition: no free workers available
```

This is misleading in both real failure modes:

- **The snapshot's node has no free workers** — free workers exist elsewhere in eligible pools, but they cannot restore the local checkpoint. The error reads as cluster-wide capacity exhaustion when it is actually a locality conflict.
- **The snapshot's node is gone** (scaled down, failed) — the actor's checkpointed state is unreachable, which is much more serious than "no free workers", and the operator gets no signal that this is the situation.

### Proposed fix

When no worker is picked **and** a node restriction was in effect, report a diagnostic error that distinguishes the cases, e.g.:

- `actor's local snapshot is on node(s) [gke-node-abc] but all 4 eligible workers on those nodes are busy; 12 free workers on other nodes cannot restore the local snapshot`
- `actor's local snapshot is on node(s) [gke-node-abc] but no eligible workers exist on those nodes (the node(s) may have been removed); 12 free workers on other nodes cannot restore the local snapshot`

The status code stays `FailedPrecondition`; only the message improves.

### Non-goals

This does not make the actor schedulable in either case. Real remediation (falling back to an external snapshot when one exists, snapshot placement tracking with liveness, load-shedding via snapshot replication) requires the snapshot data model to record external and local snapshot state simultaneously — today's `SnapshotInfo` oneof erases the external URI on pause. That redesign is a separate proposal; this issue only stops the error from pointing operators at the wrong problem.

MushuEE:
Fixes #398

> [!NOTE]
> Based on #399 (both change `findFreeWorker`); the first commit here is that PR. Review only the second commit, or merge #399 first and this rebases cleanly.

When an actor's latest snapshot is a local checkpoint, `findFreeWorker` hard-restricts candidates to the node(s) holding it — correctly, since the checkpoint only exists there. But when that restriction eliminates every candidate, the error was the generic `no free workers available`, which reads as cluster-wide capacity exhaustion and hides the far more serious case where the snapshot's node no longer exists and the actor's checkpointed state is unreachable.

`findFreeWorker` now returns a `workerScan` carrying the picked worker plus one flag gathered in its existing single pass (no extra iteration): whether any eligible worker exists on the restricted node(s). The failure message distinguishes:

- `actor's local snapshot is on node(s) [X] but all eligible workers on those nodes are busy`
- `actor's local snapshot is on node(s) [X] but no eligible workers exist on those nodes (the node(s) may have been removed)`

The status code stays `FailedPrecondition`, and the generic message is unchanged when no locality restriction is in effect (existing functional tests pass unmodified).

This intentionally does not make the actor schedulable in either case — that requires the snapshot model to retain the external URI alongside local locality, which is part of a separate architecture proposal.

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_pause.go
+++ b/cmd/ateapi/internal/controlapi/workflow_pause.go
@@ -203,12 +203,21 @@ func (s *FinalizePausedStep) Execute(ctx context.Context, input *PauseInput, sta
 		}
 		// TODO(dberkov) - what if InProgressSnapshot is empty? That shouldn't be possible.
 		if latestActor.InProgressSnapshot != "" {
+			// If the worker record was already gone we don't know which node
+			// holds the checkpoint. Record no locality rather than an empty
+			// node name: findFreeWorker treats any non-empty restriction list
+			// as a hard filter, and no worker has an empty node name, so [""]
+			// would make the actor permanently unschedulable.
+			var nodesWithSnapshot []string
+			if nodeName != "" {
+				nodesWithSnapshot = []string{nodeName}
+			}
 			latestActor.LatestSnapshotInfo = &ateapipb.SnapshotInfo{
 				Type: ateapipb.SnapshotType_SNAPSHOT_TYPE_LOCAL,
 				Data: &ateapipb.SnapshotInfo_Local{
 					Local: &ateapipb.LocalSnapshotInfo{
 						SnapshotPrefix:            latestActor.InProgressSnapshot,
-						NodeVmsWithLocalSnapshots: []string{nodeName},
+						NodeVmsWithLocalSnapshots: nodesWithSnapshot,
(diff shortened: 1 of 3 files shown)


thockin:
needs rebase

thockin (approved):
Ping me after you rebase

PR #403, opened 2026-07-07
controlapi: Add worker pool metadata to workers

sfunkenhauser:
Decouple the actor resumption scheduler from Kubernetes by avoiding direct queries to WorkerPool CRDs. Instead, cache the worker pool's labels and sandbox class on the Worker object in Redis during pod synchronization, and update the scheduler to evaluate worker eligibility using these cached fields.

### DIFF
--- a/pkg/proto/ateapipb/ateapi.proto
+++ b/pkg/proto/ateapipb/ateapi.proto
@@ -294,6 +294,8 @@ message Worker {
   int64  version         = 6;
   string worker_pod_uid  = 7;
   string node_name       = 8;
+  string sandbox_class   = 9;
+  map<string, string> labels = 10;
 }
 
 message Assignment {
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -283,7 +283,7 @@ func setupTest(t *testing.T, ns string) *testContext {
 
 	ctx, cancel := context.WithCancel(context.Background())
 
-	syncer := NewWorkerPoolSyncer(persistence, workerInformer)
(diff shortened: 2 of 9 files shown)


sfunkenhauser:
@juli4n PTAL

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -297,6 +297,8 @@ message Worker {
   int64  version         = 6;
   string worker_pod_uid  = 7;
   string node_name       = 8;
+  string sandbox_class   = 9;
```
juli4n, line 297: I wonder if we should make this a repeated field to accommodate workers that implement multiple classes.

sfunkenhauser: Currently WorkerPools only support a single class, and Workers are mapped to WorkerPools 1:1, so I don't believe it's possible for a Worker to belong to multiple sandbox classes? Are we thinking of changing this in the future?

https://github.com/agent-substrate/substrate/blob/59d89fea60e762bfa48e42484ab314ce653929e0/manifests/ate-install/generated/ate.dev_workerpools.yaml#L78

juli4n: I guess once we start doing binary encoding we can always safely switch from string to repeated so no need to worry about that now.

juli4n (dismissed):

pkg/proto/ateapipb/ateapi.proto
EItanya: FWIW I had the same question. I'm also wondering if this isn't just another label

juli4n (approved):

Issue #149, opened 2026-06-02
Define and follow a style guide for Substrate gRPC APIs


juli4n:
We need a style guide for Substrate gRPC APIs. Today this encompasses [ate-apiserver](https://github.com/agent-substrate/substrate/blob/main/pkg/proto/ateapipb/ateapi.proto), [atelet](https://github.com/agent-substrate/substrate/blob/main/internal/proto/ateletpb/atelet.proto) and [ateom](https://github.com/agent-substrate/substrate/blob/main/internal/proto/ateompb/ateom.proto) APIs.

Ate-apiserver is exposed to downstream consumers while atelet and ateom are private and only meant to be consumed by substrate core components. They should all follow the same guidelines though.

Note that we also have k8s resources (e.g. ActorTemplate, WorkerPool), but those are subject to [k8s conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md).

[Google's AIP](https://google.aip.dev/general) is an option to consider, although there might be others.

This issue is to define which style guide we are going to stick with, and make sure our APIs follow them before going alpha.

juli4n:
One important part of this is the versioning model. How will we evolve substrate APIs over time? Are we going to have alpha/beta/stable flavors of the API? What backward compatibility guarantees would we offer? 

If we go with AIP, https://google.aip.dev/185 is topical.

juli4n:
PR #351 defines an initial version of the style guide. Next step is to restructure the server implementation to align with it.

zlammerts-svg:
@juli4n can we close this one now? Thanks!

juli4n:
Not yet, I still need to land some more PRs to complete the API overhaul. I'm working on it.

juli4n:
I'll mark this as done as all the remaining gaps have tracking issues already.

PR #404, opened 2026-07-07
[HOLD] Stop storing actor identity in ate Redis's value

HavenXia:
The key `actor:<atespace>:<id>` already carries both fields, so storing them in the value duplicates `~20 bytes plus both name lengths` per actor.

The only added CPU is a string parse on the list function.  `GetActor/UpdateActor/DeleteActor` just assign the two fields from the function args. list function needs one `strings.Split` per key — nanoseconds(based on local benchmark by agents), next to the per-object `protojson.Unmarshal` which is O(microseconds). So not a big addition of latency.

Ram saved: Values can shrink by `~20 bytes + the two name lengths`, given both atespace and actor_id are DNS labels so we can save 20+2x1=22 to 20+2x63=146 bytes per object. For 100M actors that's 4~8 GB.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/store/ateredis/ateredis.go
+++ b/cmd/ateapi/internal/store/ateredis/ateredis.go
@@ -307,6 +307,27 @@ func (s *Persistence) DebugClearAll(ctx context.Context) error {
 	return err
 }
 
+// marshalStoredActor serializes dbActor for storage, clearing the identity
+// fields as they are stored in the key instead of the value. It mutates
+// dbActor: callers pass a clone they own.
+func marshalStoredActor(dbActor *ateapipb.Actor) ([]byte, error) {
+	dbActor.Atespace = ""
+	dbActor.ActorId = ""
+	return protojson.Marshal(dbActor)
+}
+
+// unmarshalStoredActor parses a stored actor value and backfills the identity
+// from the key components.
+func unmarshalStoredActor(data []byte, atespace, id string) (*ateapipb.Actor, error) {
+	actor := &ateapipb.Actor{}
+	if err := protojson.Unmarshal(data, actor); err != nil {
+		return nil, err
+	}
+	actor.Atespace = atespace
+	actor.ActorId = id
+	return actor, nil
+}
+
 func (s *Persistence) GetActor(ctx context.Context, atespace, id string) (*ateapipb.Actor, error) {
 	dbKey := actorDBKey(atespace, id)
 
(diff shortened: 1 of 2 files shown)


cmd/ateapi/internal/store/ateredis/ateredis.go
juli4n: Should we extract the logic that does marshalling (and cleans up the redundant fields) and the one that does unmarshalling (and backfills redundant fields)?

EItanya (commented):
One thing I'll say in advance is that this sets the precedent that there's no "schema" for the data in redis which may lead to confusion later on. Is it worth considering having stored data types?

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -307,6 +307,27 @@ func (s *Persistence) DebugClearAll(ctx context.Context) error {
...
+// dbActor: callers pass a clone they own.
+func marshalStoredActor(dbActor *ateapipb.Actor) ([]byte, error) {
+	dbActor.Atespace = ""
+	dbActor.ActorId = ""
+	return protojson.Marshal(dbActor)
+}
+
+// unmarshalStoredActor parses a stored actor value and backfills the identity
+// from the key components.
+func unmarshalStoredActor(data []byte, atespace, id string) (*ateapipb.Actor, error) {
```
thockin, line 321: If we intersect this with the new API style guide, we will need to replicate this function for every resource, right?  I don't see a way to do it generically.

thockin (commented):
@EItanya We *do* have a schema for stored values - it's the resource proto.  It's incomplete (in the sense that what is literally in valkey would not pass validation) but if we replaced valkey with a SQL DB that might be just as true, right?

I was going to make a comment about whether this is at the right layer or not, but I think it is -- the API assumes the storage layer produces fully correct values, regardless of how iut stores them.

thockin:
@juli4n is this optimization worthwhile?  Or is it premature?

juli4n:
> Julian Gutierrez Oschmann (@juli4n) is this optimization worthwhile? Or is it premature?

Considering all the rework we are doing (right now) in #149 I think it's probably too early (I would also expect patterns to emerge once we add common metadata, where we should be able to reuse some logic across resources).

thockin:
I initially thought it might be clever, but I think I would like to see the API overhaul land and then consider the impact of this.

@HavenXia are you OK to hold on this one for a while?

HavenXia:
> I initially thought it might be clever, but I think I would like to see the API overhaul land and then consider the impact of this.
> 
> Haven Xia (@HavenXia) are you OK to hold on this one for a while?

Yes. I will also check Julian's rework PR.

EItanya:
> Eitan Yarmush (@EItanya) We _do_ have a schema for stored values - it's the resource proto. It's incomplete (in the sense that what is literally in valkey would not pass validation) but if we replaced valkey with a SQL DB that might be just as true, right?
> 
> I was going to make a comment about whether this is at the right layer or not, but I think it is -- the API assumes the storage layer produces fully correct values, regardless of how iut stores them.

Very fair answer, just wanted to point this out. In the past I have gotten tripped up with these sorts of read/write time operations, but they definitely have their place.

HavenXia:
After the api rework, identity now sits in `ResourceMetadata` next to uid/version/timestamps, which must stay in the value anyway. 
So this can no longer drop the whole ResourceMetadata — only carve out two identity subfields, with the same **absolute** savings but a smaller share of a current object.
I think it's not worth the complexity at current scale, We can revisit pre-GA if memory becomes a real constraint.

WDYT @juli4n @thockin @EItanya

PR #405, opened 2026-07-08
Add basic per-actor external volume flow

msau42:
Initial part of #232.

Extends the ActorTemplate API to add support for per-actor external volumes and adds control plane and atelet hooks following the Actor lifecycle.

Callouts to volume operations are abstracted with a Volume interface. Right now, only a mock volume plugin for testing has been implemented, but we will add CSI support next. 

- [x] Tests pass
- [ ] Appropriate changes to documentation are included in the PR - Not going to update documentation until we add CSI support.


pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -39,12 +40,27 @@ type DurableDirVolumeSource struct {
...
+	ExternalVolumeTemplate *ExternalVolumeTemplate `json:"externalVolumeTemplate,omitempty" protobuf:"bytes,3,opt,name=externalVolumeTemplate"`
+}
+
+type ExternalVolumeTemplate struct {
+	// capacity specifies the size of the volume to create.
+	// +required
+	Capacity resource.Quantity `json:"capacity" protobuf:"bytes,1,opt,name=capacity"`
+	// storageClassName refers to the StorageClass to create the volume from.
+	// +required
+	StorageClassName string `json:"storageClassName" protobuf:"bytes,2,opt,name=storageClassName"`
```
msau42: open for discussion: can we reference k8s objects or should we create our own api?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -96,6 +96,24 @@ message Selector {
   map<string, string> match_labels = 1;
 }
 
+message ExternalVolume {
```
msau42, line 117: discuss: should ExternalVolume be its own standalone object rather than embedded fields into the Actor?

Right now ExternalVolume is tied to an Actor but we are also thinking about the scenario where you want to share a volume across multiple actors.

cmd/ateapi/internal/controlapi/create_actor.go
```diff
@@ -59,9 +64,12 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
 		ActorTemplateName:      req.GetActorTemplateName(),
 		WorkerSelector:         req.GetWorkerSelector(),
 		Atespace:               req.GetActorRef().GetAtespace(),
+		ActorVolumes:           volumes,
 	}
 	err = s.persistence.CreateActor(ctx, actor)
 	if err != nil {
+		// Cleanup created volumes if DB write fails
```
msau42, line 85: Discuss: In general, how should we approach errors when an operation only partially succeeds? Should we rollback all the partially completed operations? What happens if the rollback fails?

Or we leave the system in the partial state and require users to cleanup by calling Delete? Even then, if deletion fails and Create is called again, we could still have partially leftover resources from a previous invocation.

msau42: My current thoughts:

If there is a pair of "Create" and "Delete" operations, then we can leave the partial state and clean it up on the Delete. This does imply we will have to add some "Creating" and possibly "Deleting" state to Actor before we start creating resources for it. I already have marked that as a todo. 

For operations like "Resume", there isn't a corresponding delete operation if the resume failed. It's possible resume could be called on a different node. That means we could have partial state left on the first node. In those cases we can try to rollback the state as much as possible but that would still be a best effort operation. This has a few implications:
- We probably need to have some process on the node that monitors oprhaned resources and tries to clean them up and/or alert
- We need to make sure these orphaned resources don't get reused if a new actor with the same name gets created. This potentially implies that we should have some sort of actor uuid and be using that in our actor directories.

cmd/ateapi/internal/controlapi/volumes.go
```diff
@@ -0,0 +1,77 @@
...
+
+// createActorVolumes provisions external volumes specified in the actor template.
+// It returns the list of created external volumes, or an error if any creation fails.
+// If any volume creation fails, it cleans up any volumes created in this call on a best-effort basis.
+func (s *Service) createActorVolumes(ctx context.Context, actorID string, template *atev1alpha1.ActorTemplate) ([]*ateapipb.ExternalVolume, error) {
+	var volumes []*ateapipb.ExternalVolume
+	for _, vol := range template.Spec.Volumes {
+		if vol.ExternalVolumeTemplate != nil {
+			// Use a unique name for the volume to ensure idempotency
+			uniqueVolName := fmt.Sprintf("%s-%s", actorID, vol.Name)
```
HavenXia: This name isn't unique cluster-wide: actor names are only unique within an atespace. We already have atespace in all ateom path, I think we should also make the external volume atespaced, e.g. ` <atespace>-<actorID>-<volName>`, WDYT?

pkg/api/v1alpha1/actortemplate_types.go
thockin: I don't think there's a cleaner way. In theory we could have an ateapi level api which just acts as an indirect, but I don't see a reason at this moment.

internal/volume/plugin.go
```diff
@@ -0,0 +1,29 @@
...
+
+package volume
+
+import (
+	"context"
+)
+
+// VolumePlugin abstracts storage operations.
+type VolumePlugin interface {
+	CreateVolume(ctx context.Context, name string, capacity string, storageClass string) (volumeID string, err error)
```
thockin, line 23: Do you see this as "kubernetes PVC is the plugin", or "CSI is the plugin", or "NFS is the plugin" ?

What I really want to understand is what credentials are granted to the API server to create and destroy volumes, and what additional layers are in between.  E.g. if I need to create a million k8s PVCs and PVs, that's trouble, but if I need to give ateapi binaries the ability to delete EBS volumes in the same context as k8s, that's also trouble.

thockin (commented):
I didn't get too far into the reading of this yet, but wanted to send a couple questions before EOD

internal/volume/plugin.go
msau42: My thought is that substrate will directly call CSI grpc APIs (so ateapi needs to be authorized to communicate with CSI drivers). We will not be creating PVC objects, although we will need to have a substrate API reference to the volume IDs (which I have added in this PR).

cmd/atelet/oci.go
```diff
@@ -220,12 +228,27 @@ func buildActorOCISpec(atespace string, actorID string, args []string, env []str
...
+			srcPath = ateompath.VolumeHostPath(atespace, actorID, vm.GetName())
+		default:
+			continue
+		}
 		spec.Mounts = append(spec.Mounts, specs.Mount{
 			Destination: vm.GetMountPath(),
 			Type:        "bind",
-			Source:      ateompath.DurableDirVolumeMountPoint(atespace, actorID, vm.GetName()),
+			Source:      srcPath,
+			Options:     []string{"bind", "rw"},
```
msau42, line 316: double check this. durabledir didn't set any options.

cmd/ateapi/internal/controlapi/volumes.go
msau42: good catch, updated

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -39,12 +40,27 @@ type DurableDirVolumeSource struct {
...
 	// durableDir represents a durable directory on rootfs that persists across
 	// resumes and participates in snapshots.
 	// +optional
 	DurableDir *DurableDirVolumeSource `json:"durableDir,omitempty" protobuf:"bytes,2,opt,name=durableDir"`
+
+	// externalVolumeTemplate represents an external volume dynamically provisioned
+	// for each actor. The volume only lives as long as the actor and is deleted
+	// when the actor is deleted.
+	// +optional
+	ExternalVolumeTemplate *ExternalVolumeTemplate `json:"externalVolumeTemplate,omitempty" protobuf:"bytes,3,opt,name=externalVolumeTemplate"`
```
thockin: Do we need the word "Template" in here?  All of these are templates

juli4n: I don't think the `protobuf` directives control anything in CRDs. We should just remove them from this and other fields.

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -151,6 +154,51 @@ func (s *CallAteletPauseStep) Execute(ctx context.Context, input *PauseInput, st
 
 func (s *CallAteletPauseStep) RetryBackoff() *wait.Backoff { return nil }
 
+type DetachVolumesForPauseStep struct {
```
thockin, line 174: Could use comments on why this is different than suspend

cmd/ateapi/internal/controlapi/workflow_suspend.go
```diff
@@ -153,6 +156,51 @@ func (s *CallAteletSuspendStep) Execute(ctx context.Context, input *SuspendInput
 
 func (s *CallAteletSuspendStep) RetryBackoff() *wait.Backoff { return nil }
 
+type DetachVolumesStep struct {
```
thockin, line 175: Could use comments on how this is different from pause

thockin (commented):
I read this.  It seems broadly correct.  I am still worried about permissions, but not enough to block this.

I swear I had another comment, and it was important, but I forgot what it was.  Oh well, Friday. :)

pkg/api/v1alpha1/actortemplate_types.go
EItanya: What exactly does `External` mean here, External to `substrate`? If we are just wrapping a StorageClass anyway should we name it `PersistentVolumeSource` to be closer to the source material?

pkg/proto/ateapipb/ateapi.proto
EItanya: I vote 👍

internal/volume/plugin.go
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+package volume
+
+import (
+	"context"
+)
+
+// VolumePlugin abstracts storage operations.
+type VolumePlugin interface {
```
EItanya: Should we split this plugin by implementor? Specifically the `Ateom` currently performs:
```
MountVolume(ctx context.Context, volumeID string, targetPath string) error
UnmountVolume(ctx context.Context, volumeID string, targetPath string) error
```
And `atelet` does the rest.

Do we envision those boundaries being crossed?

pkg/api/v1alpha1/actortemplate_types.go
thockin: It's not REALLY a PV in k8s, it's just reusing k8s storage classes

pkg/proto/ateapipb/ateapi.proto
thockin: We all know how tricky volumes can be in k8s, with RWO, RWX, RO, mounts on the same node being virtually RWX but not ACTUALLY RWX, etc.  If we make this a first-class resource, do we need to track the same sorts of things (e.g. NFS can be RWX, blockdev can't)?

msau42: Yes, we will need to add the notion of access modes and enforce that shared volumes can only be RWX or ROX.

pkg/api/v1alpha1/actortemplate_types.go
msau42: Open to naming, I am trying to distinguish between volumes that will trigger dynamic provisioning, and volumes that won't. So for shared volumes, for instance, I am thinking we could have something like:

```
volumes:
  externalVolumeTemplate:
  - name: my-own-volume
    storageClassName: class1
  externalVolume:
  - name: shared-volume
    volumeName: someExistingVolume (in the same atespace) # This volume is already created and populated with data through other means
 ```

Comment (msau42) [reply pkg/api/v1alpha1/actortemplate_types.go]: Open to naming :)

But agree with Tim that this doesn't have quite the same semantics as PVs in k8s (and even the "persistent" part in k8s is misleading sometimes). At least with the per-actor volumes, the volume is definitely not persistent in the traditional sense (it only lives as long as the actor).

By "External", I'm trying to distinguish with the sandbox-backed filesystems, like rootfs and durabledir. The volumes in this case are backed by storage systems external to substrate (ie nfs, remote block, pfs, distributed fs, etc).

File: internal/volume/plugin.go (hunk 10 above)
Comment (msau42) [reply internal/volume/plugin.go]: Done, I've split it up into VolumePluginControlPlane and VolumePluginWorkerPlane. Although it's actually atelet that calls the worker plane, not ateom. And I don't expect boundaries to be crossed at this point

File: cmd/ateapi/internal/controlapi/workflow_pause.go (hunk 8 above)
Comment (msau42) [reply cmd/ateapi/internal/controlapi/workflow_pause.go:174]: I refactored so pause and suspend use the same detach flow. I left a TODO for a future optimization where we could not detach during pause.

File: pkg/api/v1alpha1/actortemplate_types.go (hunk 7 above)
Comment (msau42) [reply pkg/api/v1alpha1/actortemplate_types.go]: done

File: cmd/ateapi/internal/controlapi/volumes.go [hunk 11]
```diff
@@ -0,0 +1,146 @@
...
+		if !isVolumeMounted(matchedTemplateVol.Name, template) {
+			slog.InfoContext(ctx, "Volume not mounted in template, skipping", slog.String("volume_id", vol.GetStorageVolumeId()))
+			continue
+		}
+		mounted = append(mounted, vol)
+	}
+	return mounted
+}
+
+func actorVolumeID(ref *ateapipb.ObjectRef, volumeName string) string {
```
BenTheElder, line 113: 🟡 🤖 The hyphen join is ambiguous: atespace, actor name, and volume name may all contain hyphens, so distinct actors can produce the same ID (actor `a` in atespace `x-y` vs actor `y-a` in atespace `x`). This string is the idempotency name passed to `CreateVolume`, and CSI dedupes by name — colliding actors would silently share one storage volume. The existing TODO about actor UUID would fix it; an unambiguous encoding of the three parts would too.

cmd/ateapi/internal/controlapi/delete_actor.go
```diff
@@ -33,7 +33,21 @@ func (s *Service) DeleteActor(ctx context.Context, req *ateapipb.DeleteActorRequ
...
+	actor, err := s.persistence.GetActor(ctx, atespace, name)
+	if err != nil {
+		if errors.Is(err, store.ErrNotFound) {
+			return nil, status.Errorf(codes.NotFound, "Actor %s not found", name)
+		}
+		return nil, fmt.Errorf("while fetching actor: %w", err)
+	}
+
+	// Delete associated volumes (TODO: best effort?)
+	s.deleteActorVolumes(ctx, req.GetActor(), actor.GetActorVolumes())
```
BenTheElder: 🟡 🤖 Two issues with volume deletion here. Volumes are never detached first — for an actor deleted while running or paused, CSI `DeleteVolume` on an attached volume fails. And since failures are only logged before the record is deleted below, the volume leaks with no record pointing at it. Consider detaching first (reuse `detachActorVolumes`) and deleting the record only after volume deletion succeeds, or persisting a deletion intent for background retry.

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -39,12 +40,27 @@ type DurableDirVolumeSource struct {
...
 	// resumes and participates in snapshots.
 	// +optional
-	DurableDir *DurableDirVolumeSource `json:"durableDir,omitempty" protobuf:"bytes,2,opt,name=durableDir"`
+	DurableDir *DurableDirVolumeSource `json:"durableDir,omitempty"`
+
+	// externalVolumeTemplate represents an external volume dynamically provisioned
+	// for each actor. The volume only lives as long as the actor and is deleted
+	// when the actor is deleted.
+	// +optional
+	ExternalVolumeTemplate *ExternalVolumeTemplate `json:"externalVolumeTemplate,omitempty"`
```
BenTheElder, line 64: 🟡 🤖 `externalVolumeTemplate` needs the same microvm gate DurableDir has (the CEL rule below only checks `durableDir`). Nothing in the microvm path handles `VOLUME_TYPE_EXTERNAL`: atelet writes a bind mount whose source is a host path that doesn't exist inside the guest (only the bundle rootfs is shared over virtio-fs), so container creation fails at runtime instead of admission.

BenTheElder (commented):
[AI First Pass Experiment]

cmd/ateapi/internal/controlapi/delete_actor.go
msau42: This is an interesting question. While the state machine here (https://github.com/agent-substrate/substrate/issues/119) does allow delete from SUSPENDED, PAUSED and CRASHED states, I believe for pause/crash, there's probably still some cleanup that has to happen from atelet, and that workflow is currently missing today. @dberkov do we plan to add this workflow?

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -39,12 +40,27 @@ type DurableDirVolumeSource struct {
...
+}
+
+type ExternalVolumeTemplate struct {
```
dberkov: nit: could you please move it above the VolumeSource, so it will be close to DurDirSource

dberkov: nit:have you considered to call it "ExternalVolumeSource" rather then template?

pkg/proto/ateapipb/ateapi.proto
dberkov: >should ExternalVolume be its own standalone object rather than embedded fields into the Actor?

It can be standalone message,  as we have other standalone messages. 
About standalone object? Do you mean customer will do "kubectl ate create external_volume"?  We dont have requirements to have it today. So lets open it for discussion when it will be required.

cmd/ateapi/internal/controlapi/create_actor.go
dberkov: The current implementation does not handle error in deleteActorVolume.

Might be we need introduce "FAILED" state. So the delete volume will be cleaned in this state.

We dont have this problem today, since during create we just create record in suspended state.

cmd/ateapi/internal/controlapi/delete_actor.go
dberkov: Agree with @BenTheElder , something is broken here.  From moment we started deletion -> we cannot just ignore the error or keep it in suspended state. 
I think we need split DELETE in 2 phases:
* move to DELETING first
* do actual delete all sub-resources before deleting the record from redis.

cmd/ateapi/internal/controlapi/volumes.go
```diff
@@ -0,0 +1,146 @@
...
+	"google.golang.org/grpc/status"
+)
+
+var (
+	globalVolumePlugin = volume.NewMockVolumePlugin()
+)
+
+// TODO: Replace with actual volume plugin search
+func getVolumePlugin() volume.VolumePluginControlPlane {
+	return globalVolumePlugin
```
dberkov, line 37: this is what actually was I was missing when I reviewed pkg/api/v1alpha1/actortemplate_types.go.

How the storage plugins will be registered in the substrate instance?

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -260,6 +260,49 @@ func (s *AssignWorkerStep) RetryBackoff() *wait.Backoff {
...
+
+	node := worker.GetNodeName()
+	if node == "" {
+		return fmt.Errorf("assigned worker has no node name")
+	}
+
+	ref := &ateapipb.ObjectRef{Atespace: state.Actor.GetMetadata().GetAtespace(), Name: state.Actor.GetMetadata().GetName()}
+	for _, vol := range getMountedActorVolumes(ctx, ref, state.Actor.GetActorVolumes(), state.ActorTemplate) {
+		slog.InfoContext(ctx, "Attaching volume to node", slog.String("volume_id", vol.GetStorageVolumeId()), slog.String("node", node))
+		err := getVolumePlugin().AttachVolume(ctx, vol.GetStorageVolumeId(), node)
```
dberkov, line 296: what does attachment supposed to do? Attach volume to node VM? Is this code supposed to be executed on the atelet?

internal/volume/plugin.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"context"
+)
+
+// VolumePluginControlPlane abstracts storage operations performed on the control plane.
+type VolumePluginControlPlane interface {
+	CreateVolume(ctx context.Context, name string, capacity string, storageClass string) (volumeID string, err error)
+	DeleteVolume(ctx context.Context, volumeID string) error
+	AttachVolume(ctx context.Context, volumeID string, node string) error
```
dberkov, line 25: what information about node  do you need for attach? Is it just node name? Do you need more information, like IP, zone, etc?

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -260,6 +260,49 @@ func (s *AssignWorkerStep) RetryBackoff() *wait.Backoff {
...
+	// TODO replace with a proper check on the volumes.
+	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_RUNNING, nil
+}
+
+func (s *AttachVolumesStep) CheckPrerequisite(ctx context.Context, input *ResumeInput, state *ResumeState) error {
+	return nil
+}
+
+func (s *AttachVolumesStep) Execute(ctx context.Context, input *ResumeInput, state *ResumeState) error {
+	if state.Actor.GetAteomPodNamespace() == "" {
```
dberkov, line 279: could you please take a look to https://github.com/agent-substrate/substrate/blob/9e2a669e33dbed6b926f3ee1ac2c1af711500930/cmd/ateapi/internal/controlapi/crash.go#L50. Zoe recently added CRASH state.

demos/counter/counter.yaml.tmpl
```diff
@@ -59,3 +62,7 @@ spec:
   volumes:
   - name: data
     durableDir: {}
+  - name: external-data
+    externalVolumeTemplate:
+      capacity: 1Gi
+      storageClassName: standard
```
dberkov: out of curiosity what exactly it will do today, from what storage class data will be exeucted?

hack/run-demo-counter-kind.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
```
dberkov: out of curiosity, why do we need this shell script? I thought any demo can run on kind, it just depends how substrate was installed

pkg/api/v1alpha1/actortemplate_types.go
msau42: As mentioned in https://github.com/agent-substrate/substrate/pull/405#discussion_r3625789490, I am trying to think about how to add the "import existing volume" scenario for the shared volume use case in the future. So in the API I want to distinguish between "this will be a new volume created per actor" vs "we're accessing an existing volume that will be shared".

demos/counter/counter.yaml.tmpl
```diff
@@ -42,13 +42,16 @@ spec:
   containers:
   - name: counter
     image: ko://github.com/agent-substrate/substrate/demos/counter
+    command: ["/ko-app/counter", "--validate-existing-file-path=/external-data/test.txt"]
```
dberkov: is the external storage validation needs to be always enabled? have you considered to create a separate template instead?

cmd/ateapi/internal/controlapi/volumes.go
msau42: I plan to address this as a folowup (I already have a todo) as it will require adding a "CREATING" status to Actor and persisting that before calling CreateVolume so that the Actor will have a UID by the time we create the volumes.

cmd/ateapi/internal/controlapi/delete_actor.go
msau42: I am making a change to make volume deletion a hard failure for Actor Deletion so that we do not delete the Actor record until all volumes are successfully deleted.

However my question is about transitioning from PAUSE or CRASH directly to DELETE. Right now this just directly tries to Delete without telling the workers. I think we are missing some workflow to atelet to cleanup the nodes.

cmd/ateapi/internal/controlapi/volumes.go
msau42: The CSI integration will include plugin discovery. We're still POC-ing CSI support right now but the current thought is to add a new CRD to register the CSI endpoints to use for substrate.

cmd/ateapi/internal/controlapi/workflow_resume.go
msau42: This depends on the actual volume driver implementation but it's essentially "anything you need to do to make the volume accessible on that node". For block storage, that would be attaching the device to the node. For file storage, that could be configuring network policies to allow the node access to the volume. This is run from the CP (ateapi)

internal/volume/plugin.go
msau42: The CSI spec call only takes node id as input. The driver is responsible for interpreting that and getting any additional metadata about that node that they need.

demos/counter/counter.yaml.tmpl
msau42: It's not used right now, but once we add CSI driver support, we will extract the driverName and the opaque driver-specific properties to pass to volume creation (things like tier, performance parameters, etc)

hack/run-demo-counter-kind.sh
msau42: I was just using this as an easy way to reset my test cluster and deploy every time I want to test my changes. I can leave it out.

demos/counter/counter.yaml.tmpl
msau42: I can create a second template so that it's optional but I mainly didn't want it to diverage if people make changes to the main counter demo and forget about the volume version.

cmd/ateapi/internal/controlapi/create_actor.go
msau42: Yeah I am thinking of 2 main changes to CreateActor:
1) Add a new "CREATING" state and persist that to DB before starting to create volumes.
2) Then if CreateVolume fails, we can introduce  a FAILED state.
3) User has to call DeleteActor in order to trigger deletion of volumes and Actor DB entry

demos/counter/counter.yaml.tmpl
msau42: i found a way. I'm making the argument a template option

pkg/api/v1alpha1/actortemplate_types.go
msau42: done

pkg/proto/ateapipb/ateapi.proto
msau42: Yeah I mean standalone object. It's primarily for the shared volume use case when you want to import an existing volume that already exists.

pkg/api/v1alpha1/actortemplate_types.go
msau42: done

cmd/ateapi/internal/controlapi/workflow_resume.go
msau42: I took a look, right now it just updates DB Actor state. I don't see anything to add here wrt volumes at the moment. Like I mentioned in https://github.com/agent-substrate/substrate/pull/405#discussion_r3635304770, I think we are missing some workflow to cleanup the worker/node when a PAUSED or CRASHED actor gets deleted.

msau42:
Here's the items I would like to followup in separate PRs to limit the size of this one, and to help unblock the CSI integration work;
- [ ] naming of ExternalVolume API also taking into account potential future enhancement to support imported shares volumes (https://github.com/agent-substrate/substrate/pull/405#discussion_r3625826602, https://github.com/agent-substrate/substrate/pull/405#discussion_r3625789490)
- [ ] Handling of partial volume creation failure during CreateActor (https://github.com/agent-substrate/substrate/pull/405#discussion_r3635375062)
- [ ] Using actor uid in the volume name (depends on the above) (https://github.com/agent-substrate/substrate/pull/405#discussion_r3635296986)
- [ ] Volume cleanup coordination/handling between node and CP (https://github.com/agent-substrate/substrate/pull/405#discussion_r3635304770)
- [ ] Optimization to make attach non-blocking so that certain atelet steps can be done in parallel: https://github.com/agent-substrate/substrate/pull/405#discussion_r3642540155

hack/run-demo-counter-kind.sh
dberkov: please remove it

cmd/ateapi/internal/controlapi/workflow_resume.go
dberkov: The disadvantage to run this code in control plane, it makes resume operation slower. By running this code in atelet, this code can be executed in parallel with code that downloads images from GCS or download OCI image from artifcat registry.

msau42: The main issue is security. In the case of block storage, we don't want to give nodes permissions to attach any volume, and in the case of filers, we don't want to give the node permissions to set ACLs for itself.

In the future, we can explore optimizations to make this a non-blocking operation.

hack/run-demo-counter-kind.sh
msau42: it's already been done

msau42:
I have added tracking for the followup items in https://github.com/agent-substrate/substrate/issues/232#issuecomment-5072613033

PR #406, opened 2026-07-08
chore(otel): bump semconv v1.21.0 -> v1.40.0 and pin schema_url

krisztianfekete:
This PR bumps the imported OpenTelemetry semantic conventions from `v1.21.0` to `v1.40.0`, and explicitly pins `schema_url`.

### Rationale

The current semconv (v1.21.0) predates every GenAI conventions (and a lot more beside that), so this is a required change to set the foundation for any additional GenAI telemetry work coming next.

### No functional changes

The only semconv attributes in use are `ServiceName`, `ServiceInstanceID`, and `UserAgentOriginal` all stable and unchanged between the two versions and since v1.40.0 was already vendored, no go.mod changes.

- [x] Tests pass


EItanya (approved):
LGTM

PR #409, opened 2026-07-08
Add `ResourceMetadata` to both actor and atespace resources.

juli4n:
Replace actor_id and atespace name with `metadata` field which holds common fields.

This commit *only* changes the ate-apiserver to support this field, but intentionally skips upstream / downstream systems (such as kubectl-ate, atenet, atelet, etc) to keep the scope manageable. This will come in subsequent PRs.

Subsequent PRs will rework Create/Update/Delete/List methods to align with the style guide (validations, preconditions, etc).

#149

### DIFF
--- a/cmd/ateapi/internal/controlapi/create_atespace.go
+++ b/cmd/ateapi/internal/controlapi/create_atespace.go
@@ -31,15 +31,18 @@ func (s *Service) CreateAtespace(ctx context.Context, req *ateapipb.CreateAtespa
 		return nil, err
 	}
 
-	atespace := &ateapipb.Atespace{Name: req.GetName()}
-	if err := s.persistence.CreateAtespace(ctx, atespace); err != nil {
+	atespace := &ateapipb.Atespace{
+		Metadata: &ateapipb.ResourceMetadata{Name: req.GetName()},
+	}
+	stored, err := s.persistence.CreateAtespace(ctx, atespace)
+	if err != nil {
 		if errors.Is(err, store.ErrAlreadyExists) {
 			return nil, status.Errorf(codes.AlreadyExists, "Atespace %s already exists", req.GetName())
 		}
 		return nil, fmt.Errorf("while recording atespace: %w", err)
 	}
 
-	return &ateapipb.CreateAtespaceResponse{Atespace: atespace}, nil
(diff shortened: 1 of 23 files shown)


cmd/ateapi/internal/controlapi/create_atespace.go
```diff
@@ -31,15 +31,18 @@ func (s *Service) CreateAtespace(ctx context.Context, req *ateapipb.CreateAtespa
 		return nil, err
 	}
 
-	atespace := &ateapipb.Atespace{Name: req.GetName()}
-	if err := s.persistence.CreateAtespace(ctx, atespace); err != nil {
+	atespace := &ateapipb.Atespace{
+		Metadata: &ateapipb.ResourceMetadata{Name: req.GetName()},
+	}
+	stored, err := s.persistence.CreateAtespace(ctx, atespace)
```
HavenXia, line 37: nit: `storedAtespace`, like `storedActor` in create_actor.go.

HavenXia: Or both use `stored` as I saw in `update_actor.go` there's just `updated`.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -109,40 +135,36 @@ message Actor {
...
-  string atespace = 15;
+  string worker_pool_name = 12;
 }
 
-// Atespace is the isolation boundary an Actor is created into.
+// Atespace is the isolation boundary an Actor is created into. Global-scoped:
+// metadata.atespace is always empty; the atespace's identity is metadata.name.
 message Atespace {
-  string name = 1;
+  // Common resource metadata: uid, version, timestamps.
```
HavenXia: nit: missing `name` before uid?

cmd/ateapi/internal/controlapi/create_atespace.go
juli4n: Done

pkg/proto/ateapipb/ateapi.proto
juli4n: Done.

HavenXia (approved):

Issue #174, opened 2026-06-03
Observability At-Scale


maxsmythe:
If we are load-testing a system, we should also be able to validate that the system is functioning appropriately. Observing metrics from the load-source is well-understood, observing the behaviors of Substrate are not:

- Actors can be suspended which makes prometheus-style scraping a problem (push-based may help here if we flush metrics before sleep, but that interferes with batching/sampling)
- how do we attribute resource consumption to an individual actor? Do we need to?
- Can most benchmarking metrics be satisfied via scraping atelets/API server to validate things like # of successful suspend/resume?

Logs are already persistent, are there vital metrics we should push out to logs? Do we want an easy way of re-aggregating these metrics.

Example uses:
- Validate total # of bytes written/overwritten (or FDs opened or network requests, etc.)
- Aggregate non-fatal error counts (e.g. a long-running process attempts to poll an endpoint but times out)
- Observe latency of requests as seen by actor
- Aggregate request statistics when agent is acting as a server

alexvanboxel:
I'm willing to help here. I'm a contributor to OpenTelemetry, and I think this could be up my alley: making sure this project is OTel first, and I see in the long-term bringing some of the discoveries we're making back into the OpenTelemetry conventions.

MushuEE:
I don't know if we want to be scraping apiserver/atelet as leaving bandwidth for critical path might be needed in load spikes. An advantage of using the DB layer to track via Debezium or PubSub. Sudo CDC. Though that would tie the observability layer to the underlying DB solution and if that is pluggable, the observation layer would need to be equally pluggable.

maxsmythe:
@alexvanboxel Would love to work with people on figuring out a solution for this.

I definitely think there is something to scraping a local FD rather than using a live endpoint -- seems like it might make observing suspended actors easier. That seems roughly analogous to Debezium?

@MushuEE, I'm sensitive to not wanting to hog bandwidth, but there will always be some kind of resource drain associated with monitoring. Are you thinking that push-based models would be a better fit under load spikes?

zoez7:
I think If we look at the dimension "where is the metric emitted from" there are 2 types.
1.  Metrics captured by the agent workload (for example, latency of requests as observed by the actor)
2. Actor metrics captured by ate system (for example, CPU and memory usage, number of successful suspend and resumes)

If we look at the dimension "what is the granularity", there are several possibilities:
[...]

PR #412, opened 2026-07-09
feat(otel): set ate.* actor telemetry identity on ateapi spans

krisztianfekete:
This PR sets consistent, namespaced actor-identity set on the spans we already emit:

- `ate.atespace`, `ate.actor.id`, `ate.actor.template.{name,namespace}`, `ate.actor.version`
- ateapi: the RPC server span for create/resume/suspend/pause/delete

We do this, so platform traces are queryable by actor, atespace, and template. Cross-hop correlation and per-tenant/per-template filtering, all on traces (not TSDB labels, as discussed in #174). This is a general, workload-neutral telemetry identity plumbing. We could add gen_ai specific stuff on top of this later.

This is a resume:
image

**Notes:**
- Left as it was on purpose: metric attribute names and the stdout log labels (`ate.dev/*`), but we should probably do this as well. Maybe in this PR?
- Bikeshed is welcomed on the `ate.*` spelling before merge, renaming span attrs later is painful.

- [x] Tests pass


internal/ateattr/ateattr_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func toMap(kvs []attribute.KeyValue) map[attribute.Key]attribute.Value {
+	m := make(map[attribute.Key]attribute.Value, len(kvs))
+	for _, kv := range kvs {
+		m[kv.Key] = kv.Value
+	}
+	return m
+}
+
+func TestActorIdentity(t *testing.T) {
```
dberkov: could you please rewrite test based on data driven pattern?

tests := []struct {....}
for _, tt := range tests {
  t.Run(tt.name, func(t *testing.T) {
....

cmd/ateapi/internal/controlapi/span_identity_test.go
```diff
@@ -0,0 +1,149 @@
...
+	if !ok {
+		t.Errorf("missing %s", key)
+		return
+	}
+	if v.AsString() != want {
+		t.Errorf("%s = %q, want %q", key, v.AsString(), want)
+	}
+}
+
+func TestCreateActor_StampsFullSpanIdentity(t *testing.T) {
```
dberkov: the "cmd/ateapi/internal/controlapi/span_identity_test.go" supposed to concentrate on methods exposed in the cmd/ateapi/internal/controlapi/span_identity.go:
* setSpanActorIdentity
* setSpanActorRefIdentity

The "TestCreateActor_StampsFullSpanIdentity" method needs to be tested in create_actor_test.go. that does not exist today :)

cmd/ateapi/internal/controlapi/pause_actor.go
```diff
@@ -30,6 +30,7 @@ func (s *Service) PauseActor(ctx context.Context, req *ateapipb.PauseActorReques
 	if err := validatePauseActorRequest(req); err != nil {
 		return nil, err
 	}
+	setSpanActorRefIdentity(ctx, req.GetActor().GetAtespace(), req.GetActor().GetName())
```
dberkov: out of curiosity, what will be happened if instead of setSpanActorRefIdentity, the "setSpanActorIdentity" will be called. 
The only value that is changing between line 33 to 46 is version. 
I guess the intention is to keep the latest version right? Is it possible to call to setSpanActorIdentity twice. before s.actorWorkflow.PauseActor and after? The latest version will be submitted in successful case and previous version will be submitted in case of error.

dberkov:
>* Left as it was on purpose: metric attribute names and the stdout log labels (ate.dev/*), but we should probably do this as well. Maybe in this PR?
>* Bikeshed is welcomed on the ate.* spelling before merge, renaming span attrs later is painful.

What do you mean by it?

internal/ateattr/ateattr_test.go
krisztianfekete: Sure, done!

cmd/ateapi/internal/controlapi/span_identity_test.go
krisztianfekete: Makes sense, I split it up.

cmd/ateapi/internal/controlapi/pause_actor.go
krisztianfekete: We can't call the full one early. At line 33 we only have the `ObjectRef` from the request and `setSpanActorIdentity` needs the resolved `*Actor` that we don't have until the workflow returns it.

Attributes overwrite by key, so the later wins and we get the  latest version on success. On error we return early with only the ref, so no version.

I guess we could also add the version on mid-workflow failures if we move it inside the workflow. Maybe that's better as a follow-up if it's worth it. What do you think?

krisztianfekete:
> > * Left as it was on purpose: metric attribute names and the stdout log labels (ate.dev/*), but we should probably do this as well. Maybe in this PR?
> > * Bikeshed is welcomed on the ate.* spelling before merge, renaming span attrs later is painful.
> 
> What do you mean by it?

- This PR only covers span attributes. I left the stdout log labels and metric attribute names as-is to keep the scope small. Would you prefer to do the reconsiliation in this PR, or as a follow-up?
- The `ate.*` bikeshed was just about deciding on the namespace before merge. I haven't seen an explicit decision on `ate.*` formally (might have just missed it), so wanted to bring it up.

Thanks for the review!

cmd/ateapi/internal/controlapi/span_identity.go
```diff
@@ -0,0 +1,36 @@
...
+
+// setSpanActorIdentity annotates the RPC's server span (from ctx) with the
+// actor's full identity. A no-op when ctx carries no recording span.
+func setSpanActorIdentity(ctx context.Context, a *ateapipb.Actor) {
+	trace.SpanFromContext(ctx).SetAttributes(ateattr.ActorIdentity(a)...)
+}
+
+// setSpanActorRefIdentity is setSpanActorIdentity for the identity subset known
+// before the Actor record resolves, so a failed lookup still carries who/where.
+func setSpanActorRefIdentity(ctx context.Context, atespace, actorID string) {
```
zoez7: the parameter name is `actorID` but the callsites are passing in actor name, can we change it to `actorName`?

cmd/ateapi/internal/controlapi/span_identity.go
```diff
@@ -0,0 +1,36 @@
...
+	trace.SpanFromContext(ctx).SetAttributes(ateattr.ActorRefIdentity(atespace, actorID)...)
```
zoez7: Similarly here, can we rename `ActorRefIdentity` to ActorObjectReference to be consistent with [pkg/proto/ateapipb/ateapi.proto](https://github.com/agent-substrate/substrate/blob/c1ab0958f85202faf9f87ea66cd98903a9de763b/pkg/proto/ateapipb/ateapi.proto#L170)

internal/ateattr/ateattr.go
```diff
@@ -0,0 +1,54 @@
...
+// ActorRefIdentity returns the subset knowable before the Actor record resolves.
+func ActorRefIdentity(atespace, actorID string) []attribute.KeyValue {
+	return []attribute.KeyValue{
+		AtespaceKey.String(atespace),
+		ActorIDKey.String(actorID),
+	}
+}
+
+// ActorIdentity is nil-safe; a nil Actor yields zero-valued attributes.
+func ActorIdentity(a *ateapipb.Actor) []attribute.KeyValue {
```
zoez7: Nit: I think this should be `ActorAttributes` since it logs several fields that I would not consider as the "identity" of the actor.

cmd/ateapi/internal/controlapi/create_actor_test.go
```diff
@@ -0,0 +1,55 @@
...
+				ActorTemplateNamespace: ns,
+				ActorTemplateName:      "tmpl1",
+			},
+		}); err != nil {
+			t.Fatalf("CreateActor: %v", err)
+		}
+	})
+
+	assertSpanStr(t, attrs, ateattr.AtespaceKey, testAtespace)
+	assertSpanStr(t, attrs, ateattr.ActorIDKey, "id1")
```
zoez7: nit: also create constants for "id" similar to testAtespace above?

cmd/ateapi/internal/controlapi/pause_actor.go
zoez7: +1 without looking at the code, it's not obvious which "version" is in the attribute (is it the version before or after the pause?) we can update the key to "new_version" or "updated_version" for clarification.

Image

internal/ateattr/ateattr.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"go.opentelemetry.io/otel/attribute"
+
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+)
+
+// Dotted ate.* matches the metric-instrument naming (atenet.*, atelet.*), not the
+// ate.dev/ slash form used for k8s labels and stdout log fields.
+const (
+	AtespaceKey               = attribute.Key("ate.atespace")
```
zoez7: ate.actor.atespace?

cmd/ateapi/internal/controlapi/pause_actor.go
zoez7, line 1: Add a test in pause_actor_test.go?

cmd/ateapi/internal/controlapi/span_identity_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+
+	"github.com/agent-substrate/substrate/internal/ateattr"
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+)
+
+// installSpanRecorder swaps in a recording global TracerProvider. Span-producing
```
zoez7: Let's have each test create its own trace provider to avoid this global provider shared by all tests? We will want to run tests in parallel later to speed up testing.

zoez7:
While reviewing this I realized today the atenet router uses different keys as trace attribute
https://github.com/agent-substrate/substrate/blob/c1ab0958f85202faf9f87ea66cd98903a9de763b/cmd/atenet/internal/router/resumer.go#L49-L50C22

and the traces are not connected with the ateapi one, for the same ResumeActor request, we have one trace which traces the request through AteAPI-Atelet -Ateom 
Image

but another, detached one with just the atenet router:
Image

We don't need to address it in this PR. Could you help us create a tracking issue for this?

internal/ateattr/ateattr.go
krisztianfekete: This is now done.

cmd/ateapi/internal/controlapi/span_identity.go
krisztianfekete: Fair, but this turned out to be a bit deeper when started looking into it. `ate.actor.id` actually had the name while we also have a real uid. Went full [k8s-style as per OTel semconv](https://opentelemetry.io/docs/specs/semconv/resource/k8s/): the attr is now `ate.actor.name` and I added `ate.actor.uid`.

cmd/ateapi/internal/controlapi/create_actor_test.go
krisztianfekete: This is done, added `testActorID`.

cmd/ateapi/internal/controlapi/pause_actor.go
krisztianfekete: I'd keep it as `ate.actor.version`. It means the same thing on every op, which is the version the op resulted in. Forking it to new_version on `pause` but version on `create` breaks querying for spans that have actor A at version X. 

Happy to add a line to the attr doc spelling out "resulting version" or similar if that makes it better?

cmd/ateapi/internal/controlapi/pause_actor.go
krisztianfekete: Makes sense, added this!

cmd/ateapi/internal/controlapi/span_identity_test.go
krisztianfekete: Great idea, added this, thanks!

cmd/ateapi/internal/controlapi/span_identity.go
krisztianfekete: I went with `ActorRefAttributes` instead of `ActorObjectReference`, as it returns attributes, not a reference object, and it matches `ActorAttributes`. What do you think?

internal/ateattr/ateattr.go
krisztianfekete: I think I prefer `ate.atespace`. Atespace is the tenant/isolation boundary, where its own resource, templates is too), so if it's a top level attr you can filter spans by tenants regardless of resource type. It's the same reason OTel uses `k8s.namespace.name`, not `k8s.pod.namespace`. What do you think?

krisztianfekete:
> While reviewing this I realized today the atenet router uses different keys as trace attribute https://github.com/agent-substrate/substrate/blob/c1ab0958f85202faf9f87ea66cd98903a9de763b/cmd/atenet/internal/router/resumer.go#L49-L50C22
> 
> and the traces are not connected with the ateapi one, for the same ResumeActor request, we have one trace which traces the request through AteAPI-Atelet -Ateom Image
> 
> but another, detached one with just the atenet router: Image
> 
> We don't need to address it in this PR. Could you help us create a tracking issue for this?

Yep, I am not addressing the router keys here, opened this one to track it (happy to pick it up as well):  https://github.com/agent-substrate/substrate/issues/427

internal/ateattr/ateattr.go
zoez7: the ActorTemplateNameKey on line 37 should be `ate.template.name` instead?

cmd/ateapi/internal/controlapi/create_actor.go
```diff
@@ -76,6 +77,7 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
 		return nil, fmt.Errorf("while recording actor: %w", err)
 	}
 
+	setSpanActorIdentity(ctx, stored)
```
zoez7: nit: rename this to`setSpanActorAttributes` to distinguish from `setSpanActorRefIdentity`?

cmd/ateapi/internal/controlapi/span_identity.go
```diff
@@ -0,0 +1,36 @@
...
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/agent-substrate/substrate/internal/ateattr"
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+)
+
+// setSpanActorIdentity annotates the RPC's server span (from ctx) with the
+// actor's full identity. A no-op when ctx carries no recording span.
+func setSpanActorIdentity(ctx context.Context, a *ateapipb.Actor) {
+	trace.SpanFromContext(ctx).SetAttributes(ateattr.ActorAttributes(a)...)
```
zoez7, line 29: check if the span is nil first before setting the attributes?

zoez7 (dismissed):

cmd/ateapi/internal/controlapi/span_identity.go
krisztianfekete: `trace.SpanFromContext` never returns nil, it's a no-op span when the context has none. `TestSetSpanActorIdentity_NoRecordingSpanIsNoop` covers exactly this situation.

krisztianfekete:
Thanks @zoez7, I addressed your comments. Can you please take another look as I need an approve (and eventually merge) again?

zoez7 (approved):

Issue #207, opened 2026-06-09
Integrate with  github.com/kubernetes-sigs/kube-api-linter


juli4n:
This is a follow up of https://github.com/agent-substrate/substrate/pull/188#pullrequestreview-4438620801 where @BenTheElder suggested integrating with the `kube-api-linter` to lint our k8s types.

mesutoezdil:
ran kube-api-linter against pkg/api/v1alpha1/ (3 files: actortemplate_types.go, workerpool_types.go, sandboxconfig_types.go). and found 50 issues across these categories:

(rule // count // impact)

commentstart // 38 // doc-only, no api change 
optionalfields // 6 // api change: string/bool fields become pointers, callers need updates 
defaults // 2 // annotation-only, no api change  
nophase // 1 // api change: phase field rename, needs discussion
nomaps // 1 // api change: map[string]map[string]assetfile -> list type, significant refactor
defaultorrequired // 1 // annotation-only             
conditions  // 1  // annotation-only            

proposed approach from me:

// pr1: commentstart + defaults + defaultorrequired + conditions fixes (doc/annotation only, no api change)

// pr2: optionalfields, nophase, nomaps; these touch the api surface and likely need separate discussion before implementation

happy to proceed with pr 1 if this looks right. 

does the team want to tackle the api-breaking rules separately or skip some of them (e.g. nomaps is a significant refactor)?

kannon92:
As someone who has added kube-api-linter to existing projects, I recommend getting the plumbing (CI and all) and fix the non invasive problems first.

Its worth excluding the existing failures (pr2) so that new APis don't regress.

mesutoezdil:
PR is up: #416

It follows the split above plus @kannon92's suggestion: CI plumbing, the non invasive fixes (commentstart, conditions, defaultorrequired, optionalorrequired), and the invasive rules (nomaps, nonpointerstructs, nophase, optionalfields, requiredfields, ssatags) excluded in the config so new APIs do not regress. Fixing those on the existing types would be the follow up here.

One correction to my earlier count: the 50 was capped by the default golangci-lint issue limit, the real total is 86. The hidden ones are requiredfields (20) and ssatags (4), both in the excluded set.

PR #416, opened 2026-07-09
ci: add kube-api-linter and exclude current findings

mesutoezdil:
Part of #207, following the plan discussed there with @juli4n and @kannon92: plumbing plus the non invasive fixes first, with the existing invasive findings excluded so new APIs do not regress (as @kannon92 suggested in https://github.com/agent-substrate/substrate/issues/207#issuecomment-4921051580). The linter itself was originally suggested by @BenTheElder in #188.

Plumbing:
- hack/tools/kube-api-linter module with the golangci-lint-kube-api-linter tool, same pattern as the other tools.
- .golangci-kal.yaml config. The pre-existing findings from rules that need Go API changes (nomaps, nonpointerstructs, nophase, optionalfields, requiredfields, ssatags) are excluded rather than disabled, and each exclusion rule names the fields it excuses, so the rules still apply to new API files and to new fields on the existing types. Fixing them on the existing types is the follow up in #207.
- hack/verify/kube-api-linter.sh runs it against ./pkg/api/... and is picked up by hack/verify-all.sh in CI.

Fixes in pkg/api/v1alpha1 (doc and marker only, no Go API change):
- commentstart: field godocs now start with the serialized field name.
- conditions: Conditions moved to the first position in ActorTemplateStatus and got the listType, listMapKey and patch markers.
- defaultorrequired: SandboxConfigSpec.SandboxClass had both a default and required; now optional with omitempty, matching the same field on WorkerPoolSpec and ActorTemplateSpec. The ValidatingAdmissionPolicy in sandboxconfig-validation.yaml is unaffected since CRD defaulting runs before admission, so spec.sandboxClass is always set by then.
- optionalorrequired: missing +optional markers added on ActorTemplateStatus fields.
- defaults: configured preferredDefaultMarker to kubebuilder:default since CRDs are generated with controller-gen.

Generated CRDs regenerated. Structural schema changes are only the conditions listType/listMapKey and sandboxClass no longer in the required list (it is defaulted, so behavior is the same); the rest is description text.

Note: my earlier count of 50 issues in #207 was capped by the default golangci-lint issue limit. With the cap removed the real total is 86; the extra ones are requiredfields (20) and ssatags (4), both in the excluded set above.

Test plan
- hack/verify/kube-api-linter.sh passes (0 issues).
- Exclusions verified both ways: dropping them brings the pre-existing findings back, and a scratch field added to WorkerPoolSpec is still reported (optionalfields), which the earlier per-file exclusion swallowed.
- TestSandboxConfigValidation passes against the regenerated CRD.
- go build ./... and go test ./pkg/api/... ./cmd/atecontroller/... pass.
- gofmt, shellcheck and the regular golangci-lint on pkg/api are clean.


kannon92:
And sorry if my comment sends you on a wild goose chase. I am not really involved in this project but I follow it for interest. 

cc @BenTheElder @thockin 

I think having kube-api-linter is useful btu not sure what maintainers think.

.golangci-kal.yaml
```diff
@@ -0,0 +1,55 @@
...
+# Config for kube-api-linter (github.com/kubernetes-sigs/kube-api-linter),
+# run by hack/verify/kube-api-linter.sh against the API types in pkg/api/.
+# It is a separate file from .golangci.yaml because the kubeapilinter module
+# plugin only exists in the golangci-lint-kube-api-linter binary, not in the
+# regular golangci-lint one.
+
+version: "2"
+
+linters:
+  default: none
```
kannon92, line 24: Honestly I'm having a hard time understanding what rules you are actually enforcing.

For Kueue/JobSet, we have a config like https://github.com/kubernetes-sigs/kueue/blob/main/.golangci-kal.yaml.

Could you add something similar to make it clear what rules you want this project to enforce?

mesutoezdil: good 2 know, i didnt see this file before. updated the config to the same shape as the kueue one, the enforced rules are now listed explicitly in the file, lets see what the maintainers-reviewers will say, thx!

kannon92:
@BenTheElder 
seems like you mentioned you'd like to see this linter. Do you want to take a look?

BenTheElder (commented):
In the future I'd recommend:
1. split introducing the linter into one commit, and fixing the failure types into a commit per grouped type
2. prevent including merge commits

this time I think it's small enough to just squash to one commit on merge.

@thockin @juli4n PTAL

mesutoezdil:
> In the future I'd recommend:
> 
> 1. split introducing the linter into one commit, and fixing the failure types into a commit per grouped type
> 2. prevent including merge commits
> 
> this time I think it's small enough to just squash to one commit on merge.
> 
> Tim Hockin (@thockin) Julian Gutierrez Oschmann (@juli4n) PTAL

thx for feedback

mesutoezdil:
@HavenXia Would you mind taking a look at this one as well when you have some time?

HavenXia:
> Haven Xia (@HavenXia) Would you mind taking a look at this one as well when you have some time?

I don't have expertise in this area, the linter looks good to me. @BenTheElder can you review this and give a formal sign off?

.golangci-kal.yaml
```diff
@@ -0,0 +1,83 @@
...
+              - "requiredfields" # Required fields must not be pointers or have omitempty.
+              - "ssatags" # Array fields need listType markers for Server-Side Apply.
+              - "uniquemarkers" # Single-definition markers must not repeat.
+            # Everything not listed above stays off until evaluated.
+            disable:
+              - "*"
+          lintersConfig:
+            conditions:
+              # CRD types are not protobuf-serialized in this project.
+              useProtobuf: Ignore
```
jpbetz, line 67:
```suggestion
              useProtobuf: Ignore
              usePatchStrategy: Ignore
```

 Since KAL's docs at the pinned commit say it directly: "When linting CRD based types, set the `useProtobuf` and `usePatchStrategy` config option to `Ignore` or `Forbid`."

.golangci-kal.yaml
```diff
@@ -0,0 +1,83 @@
...
+              preferredDefaultMarker: kubebuilder:default
+
+  exclusions:
+    rules:
+      # Pre-existing findings from rules that require Go API changes
+      # (pointer fields, omitempty/omitzero tags, map-to-list refactors)
+      # on the existing v1alpha1 types. Excluded here per file rather than
+      # disabling the rules, so new API files are still checked. Fixing
+      # these on the existing types is tracked in issue #207.
+      - path: 'pkg/api/v1alpha1/(actortemplate|sandboxconfig|workerpool)_types\.go'
```
jpbetz: This will match*new* field added to these three files and exclude the rules for those fields as well.

Consider tightening this to message-level exclusions to plug this hole.

.golangci-kal.yaml
```diff
@@ -0,0 +1,83 @@
...
+  settings:
+    custom:
+      kubeapilinter:
+        type: module
+        description: Kubernetes API conventions linter.
+        settings:
+          # The enabled set below is the default set of the pinned linter
+          # version, spelled out so the enforced rules are visible here.
+          linters:
+            enable:
```
jpbetz, line 37: This looks good to me. I reviewed the list of linters available and this looks right.

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -359,17 +359,27 @@ type ActorTemplateSpec struct {
...
 // TODO: add validation
 type ActorTemplateStatus struct {
-	// Phase of the actor template.
+	// conditions defines the status conditions array
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	// +patchStrategy=merge
+	// +patchMergeKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
jpbetz: I can confirm this is the standard/recommented set of tags for `metav1.Condition`. This really does matter. For example, with `x-kubernetes-list-type: map` on the CRD this ends up being atomic from an SSA perspective.

manifests/ate-install/generated/ate.dev_actortemplates.yaml
```diff
@@ -496,14 +498,22 @@ spec:
                   - type
                   type: object
                 type: array
+                x-kubernetes-list-map-keys:
+                - type
+                x-kubernetes-list-type: map
```
jpbetz: For other reviewers' benefit: I went through the regenerated manifests and the only structural changes are these list-map extensions on status.conditions and the required-list removal for sandboxClass in the sandboxconfigs CRD. This looks right.

mesutoezdil:
@jpbetz @BenTheElder @juli4n PTAL

BenTheElder:
cc @thockin @juli4n

BenTheElder:
Can we squash these merge commits? As-is review is unwieldy.

mesutoezdil:
@BenTheElder thx, squashed into one commit
there are many changes for weeks :)

BenTheElder (commented):
I think the better way to do this is:
1. introduce linter with all current failures as exceptions
2. fix exceptions by class or type

as-is this is a rebase magnet and yet also needs pretty careful review for the API changes

mesutoezdil:
Done, split into two
#1314

BenTheElder (approved):
Thanks!

PR #419, opened 2026-07-09
Rename actor_id to actor_name in atelet/ateom protos.

HavenXia:
Aligns the internal ateletpb/ateompb services with the API style guide's actor naming. Note that internal APIs keep dedicated identity fields (not the public ObjectRef).

- [x] Tests pass


internal/proto/ateompb/ateom.proto
```diff
@@ -48,8 +48,10 @@ service Ateom {
 }
 
 message RunWorkloadRequest {
+  // TODO(haoyuxia): internal APIs keep dedicated identity fields (not the
```
juli4n: I don't think we want to reuse messages across API boundaries. I would just remove this comment.

cmd/ateom-microvm/restore.go
```diff
@@ -54,7 +54,7 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
 	defer s.lock.Unlock()
 
 	atespace := req.GetAtespace()
-	id := req.GetActorId()
+	id := req.GetActorName()
```
juli4n: name := ?

cmd/ateom-microvm/run.go
```diff
@@ -189,7 +189,7 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
 	defer s.lock.Unlock()
 
 	atespace := req.GetAtespace()
-	id := req.GetActorId()
+	id := req.GetActorName()
```
juli4n: name := ?

internal/proto/ateompb/ateom.proto
HavenXia: done

cmd/ateom-microvm/restore.go
HavenXia: updated

cmd/ateom-microvm/run.go
HavenXia: updated

juli4n (approved):

PR #421, opened 2026-07-10
Update `atenet` component to use actor *name* instead of *id*.

juli4n:
This is part of #149. Change `atenet` to not mention *actor id* any longer.

### DIFF
--- a/internal/resources/validate.go
+++ b/internal/resources/validate.go
@@ -38,6 +38,15 @@ func ValidateResourceName(name string, fldPath *field.Path) field.ErrorList {
 	return errs
 }
 
+// IsValidResourceName reports whether name is a valid Substrate resource name
+// (a DNS-1123 label; see ValidateResourceName for the rules). Use this for
+// internal, non-proto checks where a plain predicate is wanted; to validate a
+// proto request field with structured field-path errors, use
+// ValidateResourceName. Empty is not a valid name.
+func IsValidResourceName(name string) bool {
+	return len(content.IsDNS1123Label(name)) == 0
+}
+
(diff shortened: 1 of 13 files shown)


internal/resources/validate.go
```diff
@@ -38,6 +38,15 @@ func ValidateResourceName(name string, fldPath *field.Path) field.ErrorList {
 	return errs
 }
 
+// IsValidResourceName reports whether name is a valid Substrate resource name
+// (a DNS-1123 label; see ValidateResourceName for the rules). Use this for
+// internal, non-proto checks where a plain predicate is wanted; to validate a
+// proto request field with structured field-path errors, use
+// ValidateResourceName. Empty is not a valid name.
+func IsValidResourceName(name string) bool {
+	return len(content.IsDNS1123Label(name)) == 0
```
HavenXia, line 47: Should we use `ResourceNameRegexPattern.MatchString(name)` that you refactored in actors.go?

HavenXia (changes requested):

bowei:
atenet changes lgtm

internal/resources/validate.go
juli4n: We used to use the same regexp for validating *and* for building the DNS configuration, but that changed some time ago when we started using `IsDNS1123Label`. Considering that `ValidateResourceName` uses `IsDNS1123Label` I think we should be consistent.

That being said, it's a bit unfortunate that the validation logic and the regexp don't have a single SoT, but it looks like that's already the case.

HavenXia (approved):

PR #423, opened 2026-07-10
Rework List endpoints to align with API style guide.

HavenXia:
Rework List endpoints to align with API style guide.

- [x] Tests pass

### DIFF
--- a/pkg/proto/ateapipb/ateapi.proto
+++ b/pkg/proto/ateapipb/ateapi.proto
@@ -45,10 +45,10 @@ service Control {
   // Delete an actor. Only suspended actors can be deleted.
   rpc DeleteActor(DeleteActorRequest) returns (Actor) {}
 
-  // List all workers currently reflected in redis.
+  // List Workers.
   rpc ListWorkers(ListWorkersRequest) returns (ListWorkersResponse) {}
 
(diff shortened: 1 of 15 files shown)


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -249,35 +263,46 @@ message DeleteActorRequest {
...
-  // The maximum number of actors to return. The server may enforce a hard limit
-  // (e.g., 1000) regardless of the requested size.
-  int32 page_size = 1;
+  // The atespace to list actors from. Empty lists across all atespaces.
+  string atespace = 1;
 
-  // An opaque pagination token obtained from a previous ListActorsResponse.
-  // Empty for the first request.
-  string page_token = 2;
+  // Maximum number of actors to return. The server may return fewer.
```
troychiu: Since the storage layer implements pagination using Redis SCAN (where COUNT is only a hint), the server can actually return more items than the requested `page_size`.  Should we update the comments to clarify that the server may also return more results than requested?

HavenXia: Thanks, updated all list request comments to clarify that

troychiu (approved):

juli4n (approved):

Issue #433, opened 2026-07-13
Initial set of platform metrics around lifecycle, scheduling, and fleet-saturation


krisztianfekete:
Opening this issue to track the foundational work around performance/scaling-oriented platform metrics that came up in #174 and in the load-testing threads (#168, #176, #98, #44). I proposing a high-level plan here first, but if we get too deep in the details (exact attributes, buckets, etc.), happy to open a Google Doc so we can discuss it there as well.

## Scope
When we load test substrate today, we cannot really answer core operational questions with today's telemetry, e.g. is the fleet healthy under load, and where is the bottleneck?
We have three metrics total (`rpc.server.call.duration`, `atenet.router.route.duration`, `atelet.snapshot.size`), but nothing for lifecycle operations, pool saturation, or scheduling outcomes, and some components (ateom, atecontroller) emit no metrics at all. I think we have a good foundation in #174, so I am cleaning that up and using this to track the implementation.

1. otlp push instead of pull
2. Cardinality-wise actor, atespace, and session atrs go on traces/logs/events only, never on metric labels. Template, worker pool, sandbox class, outcome/phase, and cluster are the metric dimensions. We can add exemplars on the duration histograms, so a latency spike can be tracked back to a sampled actor trace. This can be a follow-up.
3. Actor level analytics will be covered by  events (named LogRecords in OTel). This came up in #174's "are there vital metrics we should push out to logs". It needs the cgroup work first (#161, blocked on #288), so just mentioning this here for context for now

## Initial set of metrics

ateapi emits the lifecycle/scheduler/fleet signals, atelet emits the restore/checkpoint phase information.

| Metric | Type | Use-case |
|---|---|---|
| `ate.actor.lifecycle.operation.duration` (`operation.name`, `error.type`, template, pool, sandbox class, snapshot kind) | histogram | activation-latency, suspend/resume rates and error rates |
| `ate.actors` (`status`, template) | updowncounter | fleet density, actor-to-worker multiplexing ratio |
| `ate.workerpool.workers` (pool, `state`, sandbox class) | observable updowncounter | pool saturation (see #198) |
| `ate.scheduler.assignment.duration` (pool, `outcome`, `error.type`) | histogram | "no free workers" rate (see #27, #398) |
| `ate.actor.restore.duration` / `ate.actor.checkpoint.duration` (`phase`, template, sandbox class, snapshot kind) | histograms | cold-start bottleneck, see related #166/#228 |

[...]

PR #434, opened 2026-07-13
[DONOTMERGE](feat): Add initial platform metrics

krisztianfekete:
This PR adds the initial platform metrics from #433:

-  lifecycle/scheduler/worker-pool instruments on ateapi, 
- restore/checkpoint timing on atelet, and
-  a shared internal/ateattr registry so every ate.* key means the same thing across spans, logs, and metrics.

Also onboards ateom and atecontroller onto the OTLP push stack and adds an e2e read-back that scrapes the kind collector to catch silent metric regressions.

**This will be superseded by smaller PRs attached to the original issue. This PR is still here if anyone wants to review/test the changes holistically.**

- [ ] Tests pass


cmd/ateapi/internal/controlapi/metrics.go
```diff
@@ -0,0 +1,171 @@
...
+
+	_, err = meter.RegisterCallback(func(_ context.Context, o metric.Observer) error {
+		ws, err := workers()
+		if err != nil {
+			// Cache not ready (warmup/reconnect): skip so no zero-valued points.
+			return nil
+		}
+		type key struct{ pool, state, class string }
+		tally := make(map[key]int64, len(ws))
+		for _, w := range ws {
```
dhaifley, line 113: I think this tally only creates a series for states present in the current worker slice. If a pool's last idle worker becomes assigned, `{state="idle"}` will disappear, rather than becoming `0`. As such, an `idle == 0` type saturation query or alert would only see an absent/stale vector and may not return results or fire.

A pool with zero registered workers disappears entirely, which also makes “all replicas are down” indistinguishable from “pool does not exist.”

Is it possible to drive the callback from the known WorkerPools/classes and emit both bounded states (`idle` and `assigned`) for each pool, including zero values, only skipping the collection when the cache itself is unavailable?

internal/proto/ateletpb/atelet.proto
```diff
@@ -228,6 +228,11 @@ message RestoreRequest {
 
   // What content to restore from the checkpoint.
   SnapshotScope scope = 10;
+
+  // Which snapshot the restore reads from ("golden" or "latest"), decided by
+  // ateapi. Carried only so atelet can label its restore-duration metric; boots
+  // use Run, not Restore, so "boot" never appears here.
+  string snapshot_kind = 11;
```
dhaifley: This free-form string is copied directly to the `ate.snapshot.kind` metric label, but `validateRestoreRequest` never bounds it. Can this be a bounded proto enum or validate/map it to `golden|latest` before recording?

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -108,14 +112,24 @@ func isWorkerEligibleForActor(worker *ateapipb.Worker, templateClass atev1alpha1
...
 	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_RUNNING, nil
 }
-func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, state *ResumeState) error {
+func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, state *ResumeState) (err error) {
+	start := time.Now()
+	// Default to the error outcome; the success and capacity paths override it
+	// before returning. pool is stamped only once a worker is actually assigned.
+	outcome := ateattr.SchedulerOutcomeError
+	pool := ""
+	defer func() {
```
dhaifley, line 180: `Execute` runs inside `runStep`'s `ExponentialBackoff`, so does this defer record every retry attempt? Would a handled `store.ErrPersistenceRetry` emit `outcome="error"`, then the eventual success emits a second `assigned` observation? Could this inflate the error rate?

cmd/ateapi/internal/controlapi/metrics.go
krisztianfekete: Good catch, agreed. Reworked this and added a test!

internal/proto/ateletpb/atelet.proto
krisztianfekete: Yes, I've been tracking this as well on my side, just wanted to get a draft PR up for discussions. I made this bounded.

cmd/ateapi/internal/controlapi/workflow_resume.go
krisztianfekete: Yep, it would, so I now skip recording those attempts and only count the terminal outcome, thanks!

krisztianfekete:
Superseded by https://github.com/agent-substrate/substrate/pull/492, #514, #776, #562, and #754

PR #438, opened 2026-07-14
fix: make atelet / ateom use actor UID for its internal bookkeeping

### LINKED ISSUE #422: Make atelet / ateom use actor UID for its internal bookkeeping
Today we use the `(atespace,name)` tuple, but now that we have a server-assigned UID per actor, we should consider switching to using it. This will prevent issues when an actor is deleted and recreated with same name, within the same atespace.

ericdbishop:
Fixes #422 

- [X] Tests pass

Gen AI disclosure: Used GPT-5.6 sol to learn about the codebase / issue context, enumerate where the actor name/atespace were being used for bookkeeping, and for adversarial review of my code.

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -560,6 +560,7 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 		RuntimeAssetPaths:      assetPaths,
 		Spec:                   buildAteomWorkloadSpec(req.GetSpec()),
 		Scope:                  toAteomSnapshotScope(req.GetScope()),
+		ActorUid: req.GetActorUid(),
 	}); err != nil {
 		// TODO: classify the errors returned by Ateom and crash the actor if needed.
 		return nil, fmt.Errorf("while calling ateom.RestoreWorkload: %w", err)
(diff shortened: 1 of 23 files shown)


cmd/atelet/main.go
```diff
@@ -566,7 +569,7 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 
 	// Record the (manifest-pinned) sandbox binaries on-node so a subsequent
 	// Checkpoint of this restored actor can re-pin the same version.
-	if err := writeSandboxRecord(atespace, actorName, sandboxRec); err != nil {
+	if err := writeSandboxRecord(req.ActorUid, sandboxRec); err != nil {
```
HavenXia: nit: can just use `actorUID`

internal/proto/ateletpb/atelet.proto
```diff
@@ -47,6 +47,8 @@ message RunRequest {
   // fetches the relevant assets and records them with the actor's on-node state
   // so a later Checkpoint can pin the same version into the snapshot manifest.
   SandboxAssets sandbox_assets = 7;
+
+  string actor_uid = 8;
```
HavenXia: I think in this stage, proto renumbering is fine, can you move actor_uid below actor_name to keep the readability?
Same for all protos you modified in this PR.

HavenXia (changes requested):

HavenXia (dismissed):

HavenXia (approved):

HavenXia (approved):

HavenXia (approved):
need rebase

PR #441, opened 2026-07-15
Avoid redundant object naming suffixes

howardjohn:
Best practices is not to put a -deployment deployment, -role role, etc; the kind already declares the kind we don't need it in the name as well. Additionally we did not do it consistently anyways.

- [x] Tests pass


EItanya (commented):
Makes sense to me, looks like CI is failing

juli4n:
Looks like a flaky test.

juli4n (approved):

PR #442, opened 2026-07-15
fix: pass atespace on self-suspend in agent-secret demo

mesutoezdil:
Host parsing kept only the actor id and dropped the atespace label,
so the self-suspend call always sent an empty atespace. The API
rejects that with atespace required, so self-suspend never worked.

Now atespace is parsed and passed through to SuspendActor.

### DIFF
--- a/demos/agent-secret/main.go
+++ b/demos/agent-secret/main.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/agent-substrate/substrate/internal/resources"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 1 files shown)


demos/agent-secret/main.go
```diff
@@ -64,16 +65,13 @@ func handleRequest(w http.ResponseWriter, r *http.Request) {
...
 		host := r.Host
 		if host == "" {
 			host = r.Header.Get("Host")
 		}
-		// Extract the actor id (first label) from <id>.<atespace>.actors.resources.substrate.ate.dev
-		parts := strings.Split(host, ".")
-		if len(parts) > 1 {
-			actorID = parts[0]
-		}
+		atespace, actorID, _ = resources.ParseActorDNSName(host)
```
juli4n: actorID -> actorName.

While you are here, can you replace this throughout the whole file?

mesutoezdil: > actorID -> actorName.
> 
> While you are here, can you replace this throughout the whole file?

Sure! Could you have a look at my other PRs as well? I’m getting a bit tired of updating them every day :) 
https://github.com/agent-substrate/substrate/pulls/mesutoezdil

HavenXia (approved):

PR #446, opened 2026-07-16
Remove `DebugClear` method from ate-apiserver

### LINKED ISSUE #443: Remove `DebugClear` method from ate-apiserver
The `DebugClear` method in `ate-apiserver` deletes all data from ValKey. This has been useful in very early stages of the project where we change the data schema almost on a daily basis. As the schema gets more stable, we should remove this method, at least from the `Control` service.

https://github.com/agent-substrate/substrate/blob/c1ab0958f85202faf9f87ea66cd98903a9de763b/pkg/proto/ateapipb/ateapi.proto#L67

troychiu:
Fixes #443

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/debugapi/debug_clear.go
+++ b/cmd/ateapi/internal/debugapi/debug_clear.go
@@ -1,0 +1,36 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package debugapi
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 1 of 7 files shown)


zoez7 (dismissed):

benchmarking/locust/common/ateapi_pb2.py
zoez7, line 1: https://github.com/agent-substrate/substrate/blob/a814761b6a50029b51ca7c3724dc88a75a693dc8/benchmarking/locust/tests/ate_api.py#L65 still references the old ActorRef, could you replace all references to the old API in benchmarking/locust/ as well?

juli4n:
Note that we are still making breaking changes to the API, so we need *some* way to delete all data. The spirit of #443 was to remove the endpoint from the `Control` service, but might need the ability to wipe out everything for a bit longer. Maybe we can separate the method to a separate `Debug` gRPC service?

troychiu:
> Note that we are still making breaking changes to the API, so we need _some_ way to delete all data. The spirit of #443 was to remove the endpoint from the `Control` service, but might need the ability to wipe out everything for a bit longer. Maybe we can separate the method to a separate `Debug` gRPC service?

That makes sense. Let me update the PR accordingly. Thank you

benchmarking/locust/common/ateapi_pb2.py
troychiu: I have discarded the changes to those files as they have been outdated. I believe it would be better to update them in a separate PR.

juli4n (approved):

PR #448, opened 2026-07-16
cli: add auto-completion for atespace and actors

howardjohn:
This mirrors how kubectl works and makes it much easier to select an actor.

- [x] Tests pass

### DIFF
--- a/cmd/kubectl-ate/internal/cmd/resource_completions.go
+++ b/cmd/kubectl-ate/internal/cmd/resource_completions.go
@@ -1,0 +1,143 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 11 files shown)


cmd/kubectl-ate/internal/cmd/resource_completions.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveNoFileComp
+	}
+	defer client.Close()
+
+	resp, err := client.ListAtespaces(ctx, &ateapipb.ListAtespacesRequest{})
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveNoFileComp
+	}
+
```
HavenXia, line 68: Can we also add the loop for pages, in #423 we are adding that support for `ListAtespace` API.

HavenXia (changes requested):
There's no docs — could you add a "auto completion" section to cmd/kubectl-ate/README.md

cmd/kubectl-ate/internal/cmd/resource_completions.go
```diff
@@ -0,0 +1,143 @@
...
+	registerAtespaceFlagCompletion(cmd)
+	cmd.ValidArgsFunction = completeActors
+}
+
+func registerAtespaceArgCompletion(cmd *cobra.Command) {
+	cmd.ValidArgsFunction = completeAtespaceArg
+}
+
+func completeAtespaces(cmd *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	ctx, cancel := resourceCompletionContext(cmd)
```
juli4n, line 55: How does this fail when there is a large number of resources? Does it freeze for `resourceCompletionTimeout` seconds?

PR #449, opened 2026-07-16
ci: optimize e2e tests a bit

howardjohn:
This should save about 30-60s off the CI runtime.

| change | Before | After | Saving |
|---|---:|---:|---:|
| Disk cleanup/check | 22.4s | 0.2s | **22.2s** |
| Identity E2E | 73.9s | 40.5s | **33.3s** |

- [x] Tests pass

### DIFF
--- a/hack/run-e2e-kind.sh
+++ b/hack/run-e2e-kind.sh
@@ -19,6 +19,11 @@ set -o errexit -o nounset -o pipefail
 ROOT="$(git rev-parse --show-toplevel)"
 cd "${ROOT}"
 
+# Build only for the Kind node's host architecture, not every .ko.yaml platform.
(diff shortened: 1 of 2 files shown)


hack/run-e2e-kind.sh
```diff
@@ -19,6 +19,11 @@ set -o errexit -o nounset -o pipefail
 ROOT="$(git rev-parse --show-toplevel)"
 cd "${ROOT}"
 
+# Build only for the Kind node's host architecture, not every .ko.yaml platform.
+# shellcheck disable=SC2155 # safe initialization
```
BenTheElder, line 23: nit: I don't think you need this line anymore, since we're capturing goarch seperately. it doesn't fail in shellcheck.net without it.

PR #453, opened 2026-07-17
fix: diagnose locality-restricted resume failures instead of generic error

### LINKED ISSUE #398: Locality-restricted resume failures report a misleading "no free workers available" error

mesutoezdil:
Locality-restricted resume failures now say why instead of the generic "no free workers available": either the restricted node's eligible workers are all busy, or the node has no eligible workers at all (may have been removed). Status code unchanged, message only.

Test plan: `go build ./...`, `go test ./cmd/ateapi/...`, added `TestNoFreeWorkerError`.

Fixes #398

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_resume.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume.go
@@ -21,6 +21,7 @@ import (
 	"log/slog"
 	"math/rand"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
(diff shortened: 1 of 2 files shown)


mesutoezdil:
@dberkov would you mind taking a look when you get a chance?

HavenXia:
#401 seems already fixed it but need a rebase

mesutoezdil:
Closing in favor of #401, which covers the same diagnostic and does it more efficiently (single pass via workerScan vs. this PR's two passes), and was opened first (2026-07-06). Thanks @HavenXia for flagging.

Issue #445, opened 2026-07-15
Consider removing custom DNS for actors


howardjohn:
Currently we rely on a fairly wonky controller that rewrites CoreDNS or other providers (https://github.com/agent-substrate/substrate/issues/348) to add the '.substrate.ate.dev' single wildcard -> IP domain. This works but it doesn't feel great.

* This targets one/two DNS provider, but there are many we will need to add
* This makes subtrate responsible for reconciling a kubernetes system resource, that will quite plausibly get reconciled back by the core cluster infrastructure.
* Modifying only the cluster-local DNS means I cannot resolve it outside of the cluster
* 'my-counter-1.demo.actors.resources.substrate.ate.dev' style naming means we have two wildcards. DNS allows a single wildcard to cover multiple labels, but TLS doesn't. This means I cannot serve a `*.actors.resources.substrate.ate.dev` TLS cert.

One option briefly discussed was using paths identify the actor. For example `curl atenet.svc.cluster.local/actors/demo/my-counter/some/path` would forward to 'my-counter' in 'demo' atespace and send to /some/path. One possible concern with this is it looks us into HTTP (unless we start doing CONNECT or something). Today, we are only implementing HTTP but the design would at least facilitate TLS as we could do SNI routing; moving to path based would remove that possibility.

bowei:
These are good questions.

The use of hostname is to offer the easiest compatibility with existing client code.

If you rely only on HTTP attributes like path, it may be impossible for users to get their clients (SDKs) to do the modifications to route their requests to a specific actor, whereas hostname should be more universal.

The possibility of SNI routing could be a benefit, but may not be fundamental.

I do think for usability we should offer the `path`-munging approach, but it feels somewhat orthogonal to this proposal

EItanya:
I never thought I'd say this, but `DNS` is quire hard, even in k8s there are a million ways to do it. I wonder if it would make sense to separate advertised hostnames from core substrate and make them optional. 

Maybe a higher level `Service` type concept or something like that.

edit: I understand `Service` is a loaded word so please don't leap on that 😆

bowei:
Yes, I think we may need to provide something like vanity domains for specific actors. This will not configure DNS necessarily but tells the ATE router ingress to map `my-agent.my-domain.com` to `ate-space-foo:actor-foo`.

Couple of challenges from past experience:

[...]

Issue #430, opened 2026-07-13
Pluggable Egress Support


bowei:
We want to make it easy for users to integrate their agent policy enforcement frameworks into the substrate Actor system. Proposal for network Egress [link](https://docs.google.com/document/d/1KmpIFu2gnqy9gp95wASgIo_vkJ_dA1DZckV8upET6bs/edit), lists the following pieces:

1. Sandbox (uVM, GVisor, etc) traffic capture to a Worker or Node local proxy.
2. Local proxy for identity, metadata injection, secure redirection to PEP
3. PEP, can be outside of Substrate.

Each of these /could/ be a point of pluggability, but this is a tradeoff -- we may not need so much flexibility at this point and we do want a default install of Substrate to come with some basic functionality for real usage. 

* For (1), there are precedents on how this can be done for in Pod redirection (see mesh ecosystem). One thought is that it is less about the interception itself and more about coordination with the lifecycle of the Actor, Worker and intercept, as the latency of setup is a big factor. Given this, it probably makes sense to make a "good enough" decision and move ahead.
* There is a coupling between (2) + (3), in terms of forwarding protocol. Can we assume that (2) + (3) are coupled together in most cases so the protocol is an implementation detail OR try to standardize on a protocol to decouple (2) from (3).
  * One proposal is to use HTTP CONNECT as the metadata wrapper. This does not handle UDP easily, but can take care of basic HTTP.

xref: 

* https://github.com/agent-substrate/substrate/issues/126.
* https://github.com/agent-substrate/substrate/pull/338/
* https://docs.google.com/document/d/1KmpIFu2gnqy9gp95wASgIo_vkJ_dA1DZckV8upET6bs/edit

stephnangue:
Warden (https://github.com/stephnangue/warden) is an identity-aware PEP build for AI agents that supports CEL policies, accepts JWT/KSA/SPIFFE/TLS cert identities, and mints and injects short-lived credentials into outbound requests.
I am interested in working on the plugable PEP.

EItanya:
> Warden (https://github.com/stephnangue/warden) is an identity-aware PEP build for AI agents that supports CEL policies, accepts JWT/KSA/SPIFFE/TLS cert identities, and mints and injects short-lived credentials into outbound requests. I am interested in working on the plugable PEP.

Hey there, @howardjohn is already working on an implementation of this.

bowei:
[...]

PR #460, opened 2026-07-18
Implement Milestone2 Networking (Ingress and Egress)

howardjohn:
This implements most of the [approved M2 networking milestone](https://docs.google.com/document/d/14QECgLl9GCGWH7hetF4gHwFbAXClJmnFiMgg0eWEUV4/edit?tab=t.xo2jqgm56l5p#bookmark=id.cikuruh587wc) scope.

## Changes

* A new `atunnel` component is added within `ateom`. This is embedded in the same binary so the `atunnel` name is mostly internal. `atunnel` has 2 responsibilities: terminating incoming traffic from the ingress, and capturing and forwarding egress traffic.
  * Actors no longer accept direct traffic. Traffic must flow through the ingress. Note the DNS already goes through the ingress, so existing usages _should_ already be doing this.
    * This is currently enforce by lack of DNAT rules on the worker to direct to the actor (which we used to have; now removed). We _could_ have NetworkPolicy additionally locking down to only 443 (where `atunnel` listens) or can leave that to users.
  * `atunnel` will listener on port 443, terminate TLS, and forward to the actor's port 80. Additional ports are a TODO pending an API for this.
  * `atunnel` additionally listens on port 15001 and accepts egress traffic from the application. A `REDIRECT` rule _conditionally_ captures all traffic leaving the actor and forwards to the `atunnel`. The `atunnel` will establish a CONNECT+mTLS request to the configured egress proxy.
  * Egress proxying is configured only if egress policies are set for the actor. This is currently impossible to set in this PR (in an effort to decouple API decisions from this PR; more on this below), so not actually triggered.
* The Ext proc code is removed from `atenet`. This leaves only the DNS stuff in `atenet`; depending on the outcome of https://github.com/agent-substrate/substrate/issues/445 we can remove `atenet` entirely (if not, we should probably rename it or just merge it into `ate-controller` probably).
* `atenet-router` is renamed to `ateway-ingress` and has a new implementation. 
  * Only the `standalone` mode is supported now. This matches the existing default. This means the ingress is its own standalone deployment.
  * The high level flow remains the same: the ingress introspects the request to find the actor from the host, makes a ResumeActor call to the ate-api-server, and then dynamically forwards to that IP address. New in this PR: 
     * mTLS from ingress to ateom/atunnel
     * atunnel verifies the actor the ingress send to is actually still the active actor
     * Actor IP assignment is cached in the ingress. Because atunnel can verify the actor, this allows us to actively evict+retry the cache if its stale.
   * Ingress is substrate aware and adds the `ate.*` attributes to logs/traces/etc as defined in https://github.com/agent-substrate/substrate/pull/412
* `ateway-egress` is added. This terminates the CONNECT, and can enforce policies before dynamically forwarding the request.
  * Currently, there is no policy, as there is not yet and API and this warrants a small PR that we can focus on the API. I had originally built out a simple stub API to verify this end to end; you can see the `[Revert egress policy](https://github.com/agent-substrate/substrate/commit/21f854ef6a1d6d9cdc8cae9f506125c4432f9199)` removing it. In parallel I will likely open a PR adding the API, but it doesn't need to block this
  * The M2 scope policies are ACLs, static credential injection, and MITM. These have all been proved out as mentioned, just not included in this PR.
  * Egress works similar to ingress; on each request a `GetActor` call is made to load the actor policies. This is cached. `atunnel` includes the actor id, atespace, and _actor version_. This ensures that if an actor policy is updated, we do not incorrectly apply stale policies from the cache.
  * atunnel is configured to send to the egress IFF egress policy exists (note; this also means we should have an egress policy that says "allow anything", as a distinct thing from "no egress policy", to allowing sending to the egress. This is useful for custom egress, or just to get observablity. This is an API thing though).
  * Redirection applies to TCP currently. All UDP is passed through for DNS; future PRs will lock down non-DNS UDP and/or proxy the DNS.
* A new `egress` demo is added to showcase a worker that sends outbound requests.

* Blocking merge: this PR currently uses an unofficial image for the ateways. I will swap it once we get high level buy in for this (just to avoid needing to do a bunch of releases if we need to iterate on this).

## Testing

Some basic testing using the counter demo shows substantially reduced tail latencies on the ingress flow:

```
DEST            P50      P90      P99
ThisPR          0.171ms  0.254ms  0.517ms
Current         3.667ms  9.816ms  17.348ms
```

Example showing end to end flow with tracing:

Aside from unit testing this PR also adds some basic end to end tests of the ingress and egress flow

## Extensibility

The M2 design is to be batteries included but swappable. Both ingress and egress can be swapped by replacing the ateway-ingress/ateway-egress deployments with their own. These components are identified by their hostname/service by the relevant components (atenet dns, ateom, etc) and can be overriden (or simply re-using the same name).

The ingress component is expected to identify an actor (through implementation-specific means), and forward to the appropriate worker on port 443 with mTLS. While the ingress _could_ support custom hostnames on the incoming request flow it must send them in the standard `my-counter.demo.actors.resources.substrate.ate.dev` style.

The egress component is expected to accept CONNECT requests over mTLS and server the appropriate configuration for the actor identified by the X-Ate-Atespace, X-Ate-Actor, and X-Ate-Actor-Version headers. The `:authority` header contains the original destination of the traffic.

Note: I do not care if we keep the 'cute' names of `ateway` and `atunnel`. I am also fine to keep `router` though `ingress` aligns with `egress` which is why I shifted it.

For https://github.com/agent-substrate/substrate/issues/430
For https://github.com/agent-substrate/substrate/issues/126
For https://github.com/agent-substrate/substrate/issues/326
Replaces https://github.com/agent-substrate/substrate/pull/338
Replaces https://github.com/agent-substrate/substrate/pull/393
Related to https://github.com/agent-substrate/substrate/pull/340

- [x] Tests pass
  * Note: egress policy tests are written but disabled as they are not yet included in this PR


cmd/atenet/internal/router/atstore.go
bowei, line 1: Did these files get moved? Where did this all go?

howardjohn: This was the old router logic for serving XDS/ext proc which is no longer there.

Looking at atstore specifically its used in two places:
* For some reason we call it and ignore the result in the router controller
* statusz page on the router.

I can add fetching the tempaltes and exposing them in atenet but it seems weird for atenet to expose these in its statusz given it doesn't care about them. I also think we can remove atenet (folding it into the controller) now that it does so little anyways

bowei:
Probably want to split this up into:

* egress
* ingress
* tunnel

in the real PRs

cmd/atelet/main.go
```diff
@@ -840,6 +844,9 @@ func (d *AteomDialer) DialAteomPod(ctx context.Context, podUID string) (*grpc.Cl
 // internal/resources so other components can apply them at their boundaries.
 func validateRunRequest(req *ateletpb.RunRequest) error {
 	var errs field.ErrorList
+	if req.GetActorVersion() < 1 {
+		errs = append(errs, field.Invalid(field.NewPath("actor_version"), req.GetActorVersion(), "must be positive"))
+	}
```
LiorLieberman: nit: question: whats the reason for that? why would it be less than 1?

cmd/ateom-gvisor/main.go
```diff
@@ -53,6 +55,13 @@ import (
 var (
 	podUID = pflag.String("pod-uid", "", "The UID of the current pod")
 
+	atunnelListenAddress       = pflag.String("atunnel-listen-address", "0.0.0.0:443", "Address for actor ingress HTTPS")
+	atunnelCredentialBundle    = pflag.String("atunnel-credential-bundle", "/run/podidentity.podcert.ate.dev/credential-bundle.pem", "PEM credential bundle for actor ingress HTTPS")
+	atunnelTrustBundle         = pflag.String("atunnel-trust-bundle", "/run/podidentity.podcert.ate.dev/trust-bundle.pem", "PEM trust bundle for actor ingress clients")
+	atunnelClientIdentity      = pflag.String("atunnel-client-identity", "spiffe://cluster.local/ns/ate-system/sa/ateway-ingress", "SPIFFE identity allowed to call actor ingress HTTPS")
+	atunnelEgressListenAddress = pflag.String("atunnel-egress-listen-address", "0.0.0.0:15001", "Address for transparently intercepted actor egress TCP")
```
LiorLieberman: nit: should 15001 be configured as a shared const somewhere to ensure consistency with workload_apply.go?

cmd/ateom-gvisor/main.go
```diff
@@ -164,22 +227,36 @@ type AteomService struct {
 
 	interiorNetNS netns.NsHandle
 	actorLogger   *actorlog.ActorLogger
+	atunnel       *atunnel.Server
```
LiorLieberman: nit: atunnelIngress ?

cmd/ateom-gvisor/main.go
```diff
@@ -430,24 +516,74 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+	}
+	if s.atunnel != nil {
+		if err := s.atunnel.Activate(atespace, actorName); err != nil {
+			return fmt.Errorf("while activating actor ingress: %w", err)
+		}
+	}
+	if egressClient != nil {
+		if err := s.atunnelEgress.Activate(egressClient, atespace, actorName, actorVersion, ""); err != nil {
+			if s.atunnel != nil {
+				_ = s.atunnel.Deactivate(context.Background())
```
LiorLieberman: why deactivating ingress if egress had a problem?

cmd/ateom-gvisor/main.go
```diff
@@ -669,26 +781,24 @@ func enableIPv4Forwarding() error {
...
+	//     listener. REDIRECT preserves SO_ORIGINAL_DST for the CONNECT authority.
+	//   * postrouting: masquerade traffic not handled by the TCP tunnel, notably
+	//     DNS over UDP, so hostname resolution continues to work.
 	//   * forward: accept forwarded packets between the actor veth and pod eth0.
 	//
-	// This is not the final egress policy path. The later AgentGateway phase
-	// should replace the broad masquerade path with transparent TCP capture and
-	// default-deny rules.
+	// TODO: Restrict the compatibility masquerade to DNS traffic sent to the
+	// configured cluster resolver and drop all other non-tunneled actor egress.
```
LiorLieberman: ~whats cluster resolver?  you mean dns resolver? So that only DNS traffic can flow without a tunnel?~

added a more relevant comment on the same lines in ateom-microvm, moving the discussion there -- https://github.com/agent-substrate/substrate/pull/460/changes#r3610697561

cmd/ateom-gvisor/main.go
```diff
@@ -816,26 +899,30 @@ func ipPayloadEqual(offset uint32, ip string) []expr.Any {
 	}
 }
 
-func tcpDestinationPortEqual(port uint16) []expr.Any {
+func tcpProtocol() []expr.Any {
```
LiorLieberman: function name sounds weird. maybe tcpProtocolMatcher()?

cmd/ateom-microvm/main.go
```diff
@@ -138,12 +148,55 @@ func do(ctx context.Context) error {
 	// logWriter with the runtime logger so the two streams to os.Stdout are
 	// serialized through one SyncedWriter and never interleave-corrupt lines.
 	actorLogger := actorlog.NewActorLogger(logWriter, metadata.OnGCE())
+	upstream, err := url.Parse("http://169.254.17.2:80")
```
LiorLieberman: nit: you defined actorHTTPUpstream somewhere, would be better to use that constant

cmd/ateom-microvm/main.go
LiorLieberman, line 1: majority of changes and content of this file is the same as ateom-gvisor, why not using shared funcs? is it by purpose?

bowei:
Taking a quick scan through the changes, I do think we have to discuss what it means for us to delete functionality out of the Substrate project and move it entirely into the agent-gateway repo, which is an independently evolving repo. This is not necessarily a good precedent to set for this project.

Specifically, I see that in order for this change to work, a changeset here (https://github.com/agentgateway/agentgateway/pull/2586) also needs to go into Agent Gateway.

I would much prefer keeping as much of the core logic associated with Substrate including the networking as part of this project in a self-contained way. At the very least, one requirement in my mind is that code and implementation should be hosted here. It can be vendored outside to create plugable implementations, but at this point, we do need to keep things more contained than have a deep dependency on outside...

LiorLieberman (commented):
Thanks @howardjohn, half way through it, still havent reviewed atunnel itself. Meanwhile, some nit comments in-line with a few questions here;

1. “Atunnel will listen on port 443.. and forward to the actor’s port 80”” – does it mean we only support 443 for ingress?  can an actor receive on 443? (in-app tls)?

2. “A redirect rule conditionally captures all traffic leaving the actor” – what is the condition? is it only when a policy exist? I guess for the context of this PR is it when the req contains egressGateway address?

3. Can connect + mTLS be pluggable/configurable? If yes, how?

4. for “configured egress proxy” why not per atespace? I can see its on the actor now, how does this gets configured? And this means we cant change it while the actor is running, correct? I dont think we should be worried about supporting it now but would be better to document it if my assumption is correct.

4. "Egress proxying is configured only if egress policies are set for the actor”
I know API is TBD still but what would be the component that watches is and activates the egress proxy - the controller?

5. "The Egress component is expected to accept CONNECT requests over mTLS" – I think there were prior discussions to make that pluggable as well - how can we support different tunnels that are not only CONNECT?

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+	ServerName           string
+	CredentialBundlePath string
+	TrustBundlePath      string
+
+	// DialContext is injectable for tests. When nil, a net.Dialer is used.
+	DialContext func(context.Context, string, string) (net.Conn, error)
+}
+
+// EgressMetadata is attached to an egress CONNECT request. BearerToken is
+// optional until actor JWT issuance is wired into ateom.
```
LiorLieberman: >until actor JWT issuance is wired into ateom.

meaning until the project  adds supports for it? Lets add a TODO instead

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	if cfg.ServerName == "" {
+		return nil, fmt.Errorf("atunnel: egress gateway server name is required")
+	}
+	if cfg.CredentialBundlePath == "" {
+		return nil, fmt.Errorf("atunnel: credential bundle path is required")
+	}
+	if cfg.TrustBundlePath == "" {
+		return nil, fmt.Errorf("atunnel: trust bundle path is required")
+	}
```
LiorLieberman: nit: should we just wrap this and all other cfg validation in a helper?

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+
+	req := &http.Request{
+		Method: http.MethodConnect,
+		URL:    &url.URL{Host: destination},
+		Host:   destination,
+		Header: http.Header{
+			ActorAtespaceHeader: []string{metadata.Atespace},
+			ActorNameHeader:     []string{metadata.ActorName},
+			ActorVersionHeader:  []string{strconv.FormatInt(metadata.ActorVersion, 10)},
```
LiorLieberman: not sure I get that version header, is this set randomly?

internal/atunnel/egress.go
```diff
@@ -0,0 +1,207 @@
...
+type EgressDialer interface {
+	DialContext(context.Context, string, EgressMetadata) (net.Conn, error)
+}
+
+// OriginalDestination returns the address that a transparently intercepted
+// connection originally targeted.
+type OriginalDestination func(net.Conn) (string, error)
+
+// Egress proxies actor TCP connections through an egress CONNECT dialer. It is
+// long-lived across actor activations, but only carries traffic while an actor
```
LiorLieberman: >It is long-lived across actor activations

Does that mean that if I have an actor suspended and then resumed it seamlessly connect to the same connection it started? e.g sent a req to an llm, went to sleep and have then is resumed to read a response?

Also, would it only work if the actor is resumed on the same node?

internal/atunnel/server.go
```diff
@@ -0,0 +1,286 @@
...
+type Config struct {
+	CredentialBundlePath string
+	TrustBundlePath      string
+	AllowedClientID      string
+	Upstream             *url.URL
+}
+
+// Server is an activation-aware HTTPS reverse proxy. It is long-lived across
+// actor activations, but only routes requests for the actor currently assigned
+// to its worker.
```
LiorLieberman: is this only for Ingress?

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	if metadata.ActorVersion < 1 {
+		return nil, fmt.Errorf("atunnel: actor version must be positive")
+	}
+
+	rawConn, err := c.dialContext(ctx, "tcp", c.gatewayAddress)
+	if err != nil {
+		return nil, fmt.Errorf("atunnel: connecting to egress gateway: %w", err)
+	}
+	tlsConn := tls.Client(rawConn, c.tlsConfig.Clone())
```
LiorLieberman: nit: do we really need to clone tlsConfig? ServerName is fixed. is this changed somewhere?

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, fmt.Errorf("atunnel: reading trust bundle: %w", err)
+	}
+	rootCAs := x509.NewCertPool()
+	if !rootCAs.AppendCertsFromPEM(trustPEM) {
+		return nil, fmt.Errorf("atunnel: trust bundle %q contains no certificates", cfg.TrustBundlePath)
+	}
+
+	dialContext := cfg.DialContext
+	if dialContext == nil {
+		dialContext = (&net.Dialer{}).DialContext
```
LiorLieberman: do we need to add timeout in the dialer?

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+	host, port, err := net.SplitHostPort(destination)
+	if err != nil {
+		return fmt.Errorf("atunnel: invalid egress destination %q: %w", destination, err)
+	}
+	if host == "" {
+		return fmt.Errorf("atunnel: invalid egress destination %q: host is empty", destination)
+	}
+	if net.ParseIP(host) == nil {
+		return fmt.Errorf("atunnel: invalid egress destination %q: host must be an IP address", destination)
+	}
```
LiorLieberman: does that mean that the Gateway (PEP) never sees a dialed DNS name?

I may be confused here - but to enforce hostname based ACLs -- is the PEP expected to read SNI from TLS and re-resolve it to IP?

cmd/ateom-microvm/net.go
```diff
@@ -333,12 +310,16 @@ func enableIPv4Forwarding() error {
...
-	//   * prerouting: DNAT pod-IP:80/tcp to the actor veth IP.
+	//   * prerouting: redirect actor TCP to atunnel, preserving SO_ORIGINAL_DST.
+	//   * postrouting: masquerade traffic not handled by the TCP tunnel, notably
+	//     DNS over UDP.
 	//   * forward: accept forwarded packets between the actor veth and pod eth0.
-	// Mirrors cmd/ateom-gvisor (same compatibility-bridge caveats and TODOs).
+	// Mirrors cmd/ateom-gvisor.
+	//
+	// TODO: Restrict the compatibility masquerade to DNS traffic sent to the
+	// configured cluster resolver and drop all other non-tunneled actor egress.
```
LiorLieberman: Are we all in agreement that the target posture should be that when an egress gateway is set, default-DENY non-tunneled egress, with only DNS to the configured resolver permitted?

Agreed with the TODO but more of a question to the broad and to figure out whether there were any discussions on this posture in the summit - it requires documentation ofc.

/cc @bowei @thockin

internal/atunnel/egress.go
```diff
@@ -0,0 +1,207 @@
...
+		defer downstream.Close()
+
+		destination, err := e.originalDestination(downstream)
+		if err != nil {
+			slog.WarnContext(active.ctx, "atunnel failed to resolve original egress destination", slog.Any("err", err))
+			return
+		}
+		upstream, err := active.dialer.DialContext(active.ctx, destination, active.metadata)
+		if err != nil {
+			slog.WarnContext(active.ctx, "atunnel failed to open egress tunnel", slog.String("destination", destination), slog.Any("err", err))
```
LiorLieberman: nit: assuming this could be pretty noisy. I am assuming agents may attempt to access destinations the policy would block. maybe this should be a metric instead + debug log?

cmd/ateom-gvisor/main.go
LiorLieberman: atunnelEgress.Deactivate()?

cmd/ateom-gvisor/main.go
```diff
@@ -816,26 +899,30 @@ func ipPayloadEqual(offset uint32, ip string) []expr.Any {
...
-		&expr.Cmp{
-			Op:       expr.CmpOpEq,
+	}
+}
+
+func actorEgressRedirectRule(table *nftables.Table, chain *nftables.Chain, port uint16) *nftables.Rule {
+	if port == 0 {
+		return nil
+	}
+	exprs := append(ipSourceEqual(actorVethIP), tcpProtocol()...)
```
LiorLieberman: what about traffic to 169.254.169.254? do we need to exclude it? 

UPDATE: do we even have cases (or do we want to allow cases) for actors to talk to metadata?

PR #462, opened 2026-07-18
Update ActorTemplate's container spec to mirror k8s semantic for `command` and `args`

### LINKED ISSUE #456: ActorTemplate's container spec should mirror k8s semantics for `command` and `args`.
Today, atelet doesn't seem to implement the same entry point resolution in the same way k8s does. It sets the process argv directly from `ActorTemplate.spec.containers.command`, and never reads the image's ENTRYPOINT / CMD fields.

This introduces some friction as in order to run any image, I need to reverse engineer the image (e.g. docker inspect) and add the entire argv into the `command` field. I think also having a semantic which is different than k8s Pod spec is not a good idea.

We should think about:

* Fix the argv resolution to consider image's CMD and ENTRYPOINT.
* Add an `args []string` field to `ActorTemplateSpec.Container` and plumb it into the argv resolution.

I think this will also allows us to remove the explicit `command: ["/ko-app/<binary-name>"]` fields we need to set in all our demos (see for example [1])

[1]: https://github.com/agent-substrate/substrate/blob/381b3ad52ca7dbf04e453872a3acfe2d6bd4cbd0/demos/counter/counter.yaml.tmpl#L45

juli4n:
* Add `args` field in addition to `command`.
* Implement logic in atelet to resolve both argv from the actor template fields and the OCI image config (`entrypoint` and `cmd` fields).
* Update demos to not hardcode command as now we honor the image default, which makes it not necessary.

Fixes #456

### DIFF
--- a/pkg/api/v1alpha1/actortemplate_types.go
+++ b/pkg/api/v1alpha1/actortemplate_types.go
@@ -92,12 +92,28 @@ type Container struct {
 	// +kubebuilder:validation:XValidation:rule="self.contains('@')",message="All images must be pinned (changing the image invalidates snapshots)"
 	Image string `json:"image,omitempty"`
 
-	// Entrypoint array. Not executed within a shell.
+	// Entrypoint array. Not executed within a shell. The container image's
+	// ENTRYPOINT is used if this is not provided; if it is provided, the
+	// image's ENTRYPOINT and CMD are both ignored and the process argv is
+	// command + args.
+	//
+	// Unlike Kubernetes, variable references $(VAR_NAME) are NOT expanded.
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=64
 	// +listType=atomic
 	Command []string `json:"command,omitempty"`
+
+	// Arguments to the entrypoint. Not executed within a shell. The container
+	// image's CMD is used if this is not provided (unless command is set,
+	// which discards the image's CMD).
+	//
+	// Unlike Kubernetes, variable references $(VAR_NAME) are NOT expanded.
+	//
+	// +optional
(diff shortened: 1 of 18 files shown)


pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -92,13 +92,29 @@ type Container struct {
...
+	// Arguments to the entrypoint. Not executed within a shell. The container
+	// image's CMD is used if this is not provided (unless command is set,
+	// which discards the image's CMD).
+	//
+	// Unlike Kubernetes, variable references $(VAR_NAME) are NOT expanded.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=64
+	// +listType=atomic
+	Args []string `json:"args,omitempty"`
```
BenTheElder, line 116: Non-blocking: `docs/api-guide.md` wasn't updated: the Container field table doesn't mention `args` or the new ENTRYPOINT/CMD resolution, and its example still hardcodes `command: ["/app/server"]` — which this PR makes unnecessary (and the demos drop). The readyz PR set the precedent of documenting new container fields there.

juli4n: I can add that, but honestly, we should consider some way to auto-generate reference documentation from the resource types. Otherwise they will always get out of sync.

juli4n: Done.

BenTheElder: +1 to autogenerate in the future

BenTheElder (approved):

PR #466, opened 2026-07-20
fix(k8sjwt): parse EC keys from JWKS discovery

aramase:
The JWKS EC branch only ever returned an error, so an issuer publishing an EC key failed verification entirely (even for its RS256 tokens). Parse P-256/384/521 and add the package's first tests.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/k8sjwt/k8sjwt.go
+++ b/cmd/ateapi/internal/k8sjwt/k8sjwt.go
@@ -306,7 +306,7 @@ func verifySignature(algorithm string, selectedKey crypto.PublicKey, toBeSignedB
 	case "ES384":
 		ecdsaKey, ok := selectedKey.(*ecdsa.PublicKey)
 		if !ok || ecdsaKey.Curve != elliptic.P384() {
-			return fmt.Errorf("requested key ID is not an ECDSA P256 key")
+			return fmt.Errorf("requested key ID is not an ECDSA P384 key")
 		}
 		toBeSignedDigest := hashBytes(crypto.SHA384.New(), toBeSignedBytes)
 		if len(signatureBytes) != 2*48 {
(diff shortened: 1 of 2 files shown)


cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -394,10 +409,34 @@ func discoverKeysForIssuer(ctx context.Context, httpClient *http.Client, issuer
 
 		switch jwk.KeyType {
 		case "EC":
-			switch jwk.EllipticCurve {
-			default:
-				return nil, fmt.Errorf("unhandled elliptic curve %q", jwk.EllipticCurve)
+			curve, err := ellipticCurveForJWK(jwk.EllipticCurve)
```
BenTheElder, line 412: 🟡 🤖 An unknown `crv` (e.g. `secp256k1`) still fails the whole JWKS — the same whole-issuer poisoning this PR fixes, just for the next unsupported key type. The pre-existing `default:` KeyType case (`OKP`) has the same effect. The conventional behavior in go-oidc and jose is to skip unsupported keys at discovery and fail only when a token's `kid` matches no usable key. Fine as a follow-up.

BenTheElder (commented):
One finding inline; nothing merge-blocking.

Verified while reviewing: `ecdsa.Verify` does reject off-curve points (Go ≥1.20), so a malformed JWK can't bypass verification; the curve↔alg cross-checks prevent key-type confusion (test-covered); coordinate handling, including the P-521 widths, is correct; package tests pass locally.

---
_Drafted with the assistance of an automated review agent; reviewed by a human before submitting._

cmd/ateapi/internal/k8sjwt/k8sjwt.go
aramase: yes, I have that implemented already in a follow-up branch. I wanted to keep this PR scoped to just the EC fix. I'll open the follow-up with those changes once this is merged.

BenTheElder (approved):

Issue #463, opened 2026-07-19
Design: on-disk, layer-deduplicated OCI image cache for atelet


dberkov:
Consolidated design addressing #437, #120, #166, #228, and unblocking #223/#276/#235/#220. Supersedes the flattened-per-digest cache sketched in #228 with a per-layer store (same overlayfs consumption model, adds cross-image layer dedup).

## Problem

Image handling in the atelet has no persistent cache and duplicates work at every level. Current behavior (`cmd/atelet/internal/memorypullcache/memorypullcache.go`, `cmd/atelet/oci.go`):

- Images are pulled with go-containerregistry and immediately **flattened** via `mutate.Extract`, destroying layer identity.
- The only cache is an **in-memory** LRU of whole flattened tarballs: digest-keyed only (**tag refs are never cached**), bounded by entry count with no byte accounting, lost on atelet restart, shared with nothing.
- The rootfs is **fully re-untarred on every actor run/resume** (`prepareOCIDirectory` → `untar`; `resetActorDirs` wipes the bundle dir between runs).
- Layers shared between images are downloaded, stored, and unpacked once **per image**, not once per node.

Measured impact from existing issues:

- Resume latency is ~99% rootfs extraction: untar ~15–20 s vs. `runsc restore` ~268 ms (#166, #228).
- The in-memory cache's 100 MiB guard never fires — `v1.Image.Size()` returns *manifest* size — and the count-bounded LRU retains one flattened copy of every digest ever pulled: atelet RSS grew 0.8 → 15.3 GiB over 4 days, OOMKilled, stranding actors in `STATUS_RESUMING` (#437).
- `mutate.Extract` itself spikes memory on multi-GB images (all layer decompressors/response bodies held open); observed OOM at a 64 GiB pod limit (#120).
- Cold pulls of large images death-loop against the 28 s workflow deadline (#233).

At target scale (up to hundreds of images per node, ~1.5 GB compressed each, ~40% shared layers), flattened per-image storage needs ~1 TB where layer dedup needs ~650 GB — and every shared base layer is re-downloaded per image.

Related TODOs: `memorypullcache.go:51`; roadmap "Shared image cache", "Peer-to-peer OCI and snapshot sharing".

## Proposed design

A content-addressed, on-disk **layer pool** owned solely by the atelet (single-writer — no daemon, no leases DB; precedent: the digest-named asset cache in `cmd/atelet/sandbox_assets.go`).

### On-disk layout

```
[...]

PR #467, opened 2026-07-20
Add node-local OCI image layer cache: overlay-composed actor rootfs from shared unpacked layers (#463 Phase 1)

### LINKED ISSUE #166: Avoid re-untarring actor image rootfs on every restore

### LINKED ISSUE #228: Use a node-local overlayfs rootfs cache to eliminate per-restore untar during actor resume

### LINKED ISSUE #437: atelet: MemoryPullCache retains full image tarballs indefinitely (LRU bounded by entry count; the 100 MiB guard measures manifest size)
## Summary

Two bugs in `cmd/atelet/internal/memorypullcache`. Together they let atelet's heap grow without bound — it permanently retains one fully-extracted copy of every distinct image digest it has ever pulled — until the kernel OOM-kills it.

This is distinct from #120, which is a transient spike *inside* a single `mutate.Extract`. What follows is permanent retention *across* pulls.

## Bug 1: the size guard measures the wrong thing

```go
size, err := img.Size()
if size > 100*1024*1024 {
    slog.InfoContext(ctx, "Image is too large to cache", ...)
    return mutate.Extract(img), &imageCfg, err   // not cached
}
```

`v1.Image.Size()` returns the size of the **manifest** (a few KB), not the size of the image content. The guard therefore never fires: `Image is too large to cache` appears **0 times** in our logs, while a ~740 MiB (extracted) image is cached on every pull.

## Bug 2: the LRU is bounded by entry count, not by bytes

```go
cache: lru.New(256)

type cachedImage struct {
    tar []byte
    cfg v1.Config
}
```

Each entry holds the **entire flattened, uncompressed image tarball**. Bounded at 256 *entries*, atelet will retain 256 × image-size before evicting anything. The existing TODO already anticipates this:

> `// TODO: Need a smarter cache with bounds on total consumed size, not just number of entries.`

## Impact (measured)

A Node-based coding-agent image: 222 MiB compressed (4 layers) → **~740 MiB extracted**.

- `go_memstats_heap_inuse_bytes` = **1.455 GiB** with exactly 2 distinct image digests cached (2 × ~0.74 GiB). `go_goroutines` = 42 and `container_memory_cache` stays flat at 0.02–0.08 GiB, so this is live, reachable heap — not garbage, not page cache.
- Over ~4 days atelet RSS climbed monotonically from 0.8 GiB to **15.3 GiB** and was OOMKilled (exit 137). The sawtooth then repeats.

Growth is per *distinct digest*, not per actor — cache hits cost nothing. But since each actor pins the image digest it was created with, digest diversity accumulates as older actors are resumed, so any cluster that ships new image builds keeps introducing new permanently-retained entries.

### Collateral: actors stranded in `STATUS_RESUMING`

When atelet dies mid-restore, the in-flight `RestoreWorkload` is cancelled — `while creating pause container: while running 'runsc create': context canceled` — and the actor is left in `STATUS_RESUMING` indefinitely. An explicit `resume` recovers it, but it does not self-heal.

## Suggested fix

Make the cache size-aware: account `len(tar)` and bound total bytes (configurable budget), evicting LRU. Hot digests stay cached, so resume latency does not regress, while the heap is capped.

Fixing only Bug 1 would also stop the growth, but it would disable caching for exactly the images that matter and add a registry pull to every resume — which makes #228 worse. Bounding by bytes seems strictly better.

Happy to send a PR if this direction sounds right.

dberkov:
## Summary

Phase 1 of #463: replaces atelet's in-memory flattened-tarball pull cache (`memorypullcache`) with a node-local, on-disk, content-addressed pool of **unpacked image layers** (`internal/imagecache`), shared by every actor on the node. An actor's rootfs becomes an overlayfs mount — cached layers as read-only lowerdirs, a bundle-local upper for the actor's private writes — instead of a full image re-untar on every run/resume.

Fixes #437, fixes #166, fixes #228.

## What changes

- **`internal/imagecache`** (new): layer pool (`layers/sha256/<diffid>/fs`), digest-keyed image records, tag→digest resolution via one `remote.Head` (tag refs become cacheable), parallel streaming layer download+unpack (bounded, singleflight-collapsed per layer and per image), atomic tmpdir+rename with startup recovery, layout version marker. See `internal/imagecache/README.md` for the full design.
- **atelet**: `prepareOCIDirectory` calls `EnsureImage` and writes a per-bundle `rootfs-overlay.json` (+ empty `rootfs/`, `upper/`, `work/`) instead of fetch+untar. New `--image-cache-dir` flag (default `/var/lib/ateom-gvisor/image-cache`, on the shared hostPath). `memorypullcache` is deleted.
- **ateom-gvisor / ateom-microvm**: compose the bundle rootfs (`SetupBundleRootfs`) right before `runsc create`/`restore` / virtio-fs lower staging, and detach the mounts at teardown (`UnmountAllUnder`). Bundles without a spec are left untouched (compatibility with pre-imagecache bundles).
- **`tools/validate-image-cache`** (new): batch-validates that arbitrary registry images can be pulled, parsed, and unpacked by the store half; used to sweep 400+ real GB-scale eval images during development.
- **CI**: `run-tests` re-runs the imagecache package under `sudo` so the root-gated tests (mknod whiteouts, trusted xattrs, overlay mount round trip) execute instead of skipping.

## Deviations from the #463 design — please review deliberately

1. **Mounting lives in ateom, not atelet.** #463 sketched atelet mounting overlays, but atelet deliberately runs with **all capabilities dropped** ("atelet does no mounts" — `manifests/ate-install/atelet.yaml`). So the module is split along that existing boundary: atelet pulls/unpacks (pure file I/O) and records whiteouts in per-layer metadata; the privileged ateoms materialize whiteouts (`mknod` + `trusted.overlay.opaque`) once per layer and mount, in their own mount namespace — which is where runsc's gofer / virtiofsd resolve paths, so **no mount-propagation configuration is needed anywhere**.
2. **In-repo untar instead of `containerd/archive.Apply`.** containerd's applier assumes capabilities atelet doesn't have (e.g. writing children into image-defined read-only dirs relies on `CAP_DAC_OVERRIDE`). The existing hardened untar already solves that (dir-mode juggling for plain root, `os.Root` confinement, later-entry-wins); it's extended with whiteout capture and missing-parent-dir creation rather than replaced.

## Behavior preserved

- **Pristine rootfs per run**: the actor's upper is wiped by `resetActorDirs` between runs, same contract as the old full re-untar — just at mount cost (ms) instead of extraction cost (tens of seconds).
- **Micro-VM path nearly untouched**: the bundle rootfs it bind-mounts into virtiofsd is now an overlay mountpoint; the guest keeps building its own tmpfs upper. The find-paths requirement (byte-identical lower across reconstructions) holds — cached layers are more deterministic than per-node untars.

## Validation

- **kind e2e (gVisor)**: counter demo run → increment → suspend → resume; `runsc create/start/restore` all on overlay rootfs; `oci_unpack` **2.9–4.8 ms** on resume (was ~15–20 s per #166); durable state preserved across suspend/resume.
- **GKE (gVisor + micro-VM)**: counter and counter-microvm demos, external GCS checkpoints, repeated suspend/resume cycles; cache hits and ms-scale `oci_unpack` on both runtimes; cache survives atelet restarts.
- **Image corpus**: 411 real production eval images (~1–1.5 GB compressed, up to 19 layers) pulled/parsed/unpacked with **0 failures** via `tools/validate-image-cache`. Two real-image bugs found on GKE during development are fixed with regression tests: layer tars omitting parent-dir entries, and images listing the same layer twice (overlayfs rejects duplicate lowerdirs with `ELOOP`; compose dedups to the topmost occurrence).
- **Unit tests**: unpack security suite (path traversal, symlink/hardlink escape, whiteout capture), spec round-trips, mountinfo parsing, and root-gated tests for the privileged half — including a full mount → write-isolation → unmount round trip asserting actor writes land in the bundle upper, never in the shared pool.

## Known gaps / notes for reviewers

- **No GC yet** (Phase 2 of #463): layers are never evicted; disk grows with the set of distinct layers ever pulled. Documented in `internal/imagecache/README.md` with operator guidance.
- **Upgrade ordering**: a new atelet with old ateom worker pods yields an empty (unmounted) rootfs — deploy ateoms with/before atelet. Old bundles keep working under new ateoms.
- The `untar` tracing span no longer exists (untar is gone); `prepareOCIDirectory` and RPC spans remain. An `EnsureImage` span with hit/miss attributes is natural Phase 2 material alongside cache metrics.


internal/imagecache/spec.go
```diff
@@ -0,0 +1,142 @@
...
+		// assemble a corrupt option string.
+		if strings.ContainsAny(p, ":,") {
+			return "", fmt.Errorf("layer path %q contains overlay option separators", p)
+		}
+		lowers = append(lowers, p)
+	}
+	if strings.ContainsAny(upper, ":,") || strings.ContainsAny(work, ":,") {
+		return "", fmt.Errorf("bundle path %q contains overlay option separators", upper)
+	}
+	return "lowerdir=" + strings.Join(lowers, ":") + ",upperdir=" + upper + ",workdir=" + work, nil
```
BenTheElder: 🟡 🤖 No guard on the assembled option-string length. `mount(2)` copies options via a single page, so the whole string is capped at ~4096 bytes. Layer paths here are ~114 bytes each (`/var/lib/ateom-gvisor/image-cache/layers/sha256/<64-hex>/fs:`), which caps images at roughly **34 layers** — Docker permits 127, and heavily-layered CI/base-chain images in the 30–60 range exist in the wild. The validation corpus topped out at 19 layers, so this wasn't hit; when it is, the failure is an opaque `EINVAL` from the mount.

This is exactly why containerd's overlay snapshotter uses short integer snapshot IDs. Options: a preflight length check with a clear error (minimum), short per-layer symlinks/IDs in the pool, or the new mount API (`fsconfig` + `lowerdir+` appends, kernel ≥ 6.5) which avoids the single-page limit.

internal/imagecache/unpack.go
```diff
@@ -0,0 +1,259 @@
...
+		mode := hdr.FileInfo().Mode().Perm()
+
+		// A layer tar routinely omits entries for parent directories that
+		// exist in lower layers (e.g. just "etc/nsswitch.conf", with "etc/"
+		// declared only in the base layer). Unpacking per layer, those
+		// parents must be created here; overlayfs merges them with the
+		// lower layers' directories at compose time.
+		if parent := filepath.Dir(name); parent != "." {
+			if err := root.MkdirAll(parent, 0o755); err != nil {
+				return nil, fmt.Errorf("while creating parent directories for %q: %w", name, err)
```
BenTheElder, line 168: 🟡 🤖 Phantom parent dirs shadow lower-layer directory metadata in the merged view. In overlayfs, the **top-most** layer containing a directory supplies its attributes. A parent auto-created here (root:root `0755`) in an upper layer hides the real metadata a lower layer declared — e.g. a layer shipping `root/.bashrc` without a `root/` entry turns `/root` from `0700` into `0755`; a file added under `/tmp` without a `tmp/` entry drops the `1777` sticky bit. The old flattened extract didn't have this (parents declared in any layer kept their modes); the PR notes the omitted-parent case was hit on real GKE images, so this will occur. `mknodWhiteout`'s `MkdirAll` in bundle_linux.go has the same property.

Not easily fixable without lower-layer context at unpack time (containerd sidesteps it by applying through the mounted parent chain, so copy-up preserves metadata). At minimum worth documenting in the README as a known deviation; a fuller fix could consult the already-cached lower layers of the same image when creating phantom parents.

cmd/ateom-gvisor/main.go
```diff
@@ -292,6 +303,16 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 			"err", err)
 	}
 
+	// Detach the overlay rootfs mounts before atelet wipes the bundle dirs
+	// (deleting a bundle out from under a live mount in this namespace would
+	// leave the mount orphaned until the pod restarts). Best-effort, same as
+	// the container cleanup above.
+	if err := imagecache.UnmountAllUnder(ateompath.OCIBundleDir(req.GetActorUid())); err != nil {
```
BenTheElder, line 319: 🟢 🤖 Teardown asymmetry: this detaches bundle overlays after checkpoint, but the Run-failure defer (line ~197) and Restore-failure defer (~390) only clean up the network — overlays mounted by a partially-successful `SetupBundleRootfs` linger in this namespace until atelet's later bundle wipe lazily detaches them. Adding `UnmountAllUnder` to those failure defers would make cleanup symmetric.

internal/imagecache/imagecache.go
```diff
@@ -0,0 +1,546 @@
...
+}
+
+func (s *Store) recordPath(digest v1.Hash) string {
+	return filepath.Join(s.root, "manifests", digest.Algorithm, digest.Hex+".json")
+}
+
+// sweepTempDirs removes unpack temp dirs orphaned by a crash. A layer dir
+// without the temp prefix is always complete (it was moved into place with a
+// single rename), so this is the only recovery the pool needs.
+func (s *Store) sweepTempDirs() error {
```
BenTheElder, line 199: 🟢 🤖 This sweeps only orphaned layer temp dirs; `writeRecord`'s temp files (`.<digest>.json.tmp-*` under `manifests/sha256/`) orphaned by a crash are never cleaned. Tiny files, but the sweep exists for exactly this class — worth including the manifests dir.

BenTheElder (commented):
Quick AI review ... (continuing to refine that experiment on coworker PRs for now, intend to add a skill)

----

 Findings inline: two 🟡 (overlayfs mount-option page-size cap → images beyond ~34 layers can't mount; phantom parent dirs shadowing lower-layer dir metadata in the merged view) and two 🟢 (failure-path unmount asymmetry; record temp files not swept).

Checked and found no issues: unpack confinement (`os.Root`, tar-name validation, hardlink/symlink replacement) matches the old hardened untar; whiteout capture→materialize split is correct and `FinalizeLayer` is safely idempotent under concurrent ateoms; lowerdir dedup (topmost wins) is the right fix for the duplicate-layer `ELOOP`; the tag→digest HEAD resolution only affects non-pinned refs (actor images are digest-pinned by CRD validation); microvm covers both run and restore via `buildActorContainers`; FIFO/device tar entries still hard-fail but did in the old untar too (pre-existing, not a regression); `GOOS=linux` build and the unprivileged imagecache tests pass locally.

---
_Drafted with the assistance of an automated review agent; reviewed by a human before submitting._

internal/imagecache/spec.go
dberkov: Fixed in 77026523 by switching to the new mount API: `fsopen` + one `fsconfig` `lowerdir+` append per layer removes the aggregate option-string limit structurally (no symlink indirection needed), and failed mounts now surface the kernel's fs-context error log instead of a bare EINVAL. Added a 64-layer root-gated regression test that asserts the joined paths exceed a page before mounting. This sets a kernel floor of 6.5 (`lowerdir+`) — documented in the code and README; all current GKE channels ship ≥6.6, and we verified empirically that a 6.1 kernel (debian-12) fails fast with a clear kernel log line rather than corrupting anything.

internal/imagecache/unpack.go
dberkov: Implemented the fuller fix in 94fb9efe rather than document-only. Baking lower-derived attrs into the pool is unsound (layer trees are shared across images with different chains), so the repair is per-actor at compose time: unpack records implicitly-created dirs in the layer metadata (including parents that only exist for whiteout materialization — the `mknodWhiteout` MkdirAll case you noted), and `SetupBundleRootfs` resolves each shadowed dir's true mode+ownership from the top-most non-implicit layer in that image's chain, applying it through the mount so the copy-up lands in the bundle's private upper. The shared pool is never modified (asserted in the root-gated test). Pre-existing cached layers have no recorded metadata and are simply left as-is. Residual gaps (mtimes, xattrs, dirs implicit in every layer) are documented in the README.

cmd/ateom-gvisor/main.go
dberkov: Fixed in 91a2f0d6 (gvisor) and d5787a95 (microvm — buildActorContainers mounts the same way, so its Run and Restore failure defers had the identical gap). In gvisor the detach runs before cleanupActorNetworkOrExit, which exits the process on failure and would otherwise skip it.

internal/imagecache/imagecache.go
dberkov: Fixed in f7970cf9 — the startup sweep now also removes writeRecord's ".<digest>.json.tmp-*" orphans (finished records never start with a dot, so the check is exact), with the recovery test extended to cover both the orphan removal and a completed record surviving the sweep.

.github/workflows/pr-workflow.yaml
```diff
@@ -33,6 +33,11 @@ jobs:
       with:
         go-version-file: 'go.mod'
     - run: go test -v ./...
+    # The imagecache consumer half (overlay mounts, whiteout mknod, trusted.*
```
BenTheElder, line 36: I think we should generalize rootful "integration" tests as a follow-up TODO.

internal/imagecache/bundle_linux.go
```diff
@@ -0,0 +1,296 @@
...
+	return nil
+}
+
+// mountOverlay attaches an overlay of lowers (top-most first) with the given
+// upper/work dirs at mountpoint, using the new mount API rather than
+// mount(2): appending lowerdirs one fsconfig(2) call at a time sidesteps
+// mount(2)'s single-page option-string cap, which digest-derived layer paths
+// (~114 bytes each) would hit at roughly 34 layers.
+//
+// Minimum supported kernel: Linux 6.5, where overlayfs gained the
```
BenTheElder, line 106: 🟢 🤖 The COS citation covers GKE's default images, but not GKE Ubuntu node images or developer hosts running kind (Ubuntu 22.04 GA is kernel 5.15). On those, every mount now fails at `lowerdir+`. If they're in scope, consider a fallback to legacy `mount(2)` when the kernel lacks `lowerdir+`, or at least a startup probe that fails atelet/ateom with a clear "kernel too old" error instead of per-actor mount failures.

BenTheElder: I think we can follow-up on this one as well (green = non-blocking)

.github/workflows/pr-workflow.yaml
dberkov: Agreed — filed #501 to generalize rootful test execution (discovery via build tag/convention + a single sudo step, plus contributor docs), so this per-package step becomes the first consumer rather than a pattern.

internal/imagecache/bundle_linux.go
dberkov: Good catch on the non-COS environments — filed #500 for the follow-up: probe `lowerdir+` once at ateom startup, fall back to legacy `mount(2)` assembly when absent (graceful: pre-6.5 kernels can't mount >~34-layer chains by any mechanism, so the fallback only fails where the kernel itself is the limit), and emit one clear kernel-too-old log line instead of per-actor EINVALs.

BenTheElder (approved):

PR #468, opened 2026-07-20
dev-infra: make clang-format hermetic

### LINKED ISSUE #457: Dev scripts shouldn't require clang-format to be installed locally
We shouldn't assume tools are locally installed as that introduces friction, but more importantly, makes the dev env non hermetic. In general we use `go tools` (with a dedicated Go module under `hack/tools` for isolation) for tools that can be distributed with go modules, and for those that can't (like clang-format), we have as script that fetches a pinned version (see protoc as an example[1]).

[1]: https://github.com/agent-substrate/substrate/blob/381b3ad52ca7dbf04e453872a3acfe2d6bd4cbd0/hack/protoc.sh#L21

YQ-Wang:
## Summary

- add a pinned clang-format wrapper for Linux and macOS on x86-64 and ARM64
- verify downloaded wheels with platform-specific SHA-256 checksums and cache the installed binary by version
- use the hermetic wrapper for protobuf formatting instead of requiring clang-format on PATH

Fixes #457

## Testing

- cold-cache download and checksum verification
- cached clang-format invocation
- hack/update/proto-fmt.sh
- ShellCheck 0.11.0
- hack/verify/boilerplate.sh
- go test ./...


YQ-Wang:
@juli4n, would you be willing to review this implementation of #457?

hack/clang-format.sh
```diff
@@ -0,0 +1,77 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit -o nounset -o pipefail
+
+CLANG_FORMAT_VERSION="22.1.8"
+
+ROOT="$(git rev-parse --show-toplevel)"
+INSTALL_DIR="${ROOT}/bin/clang-format-install/${CLANG_FORMAT_VERSION}"
```
BenTheElder, line 26: 🤖 **should-fix** 🟡 – `hack/verify/proto-fmt.sh` runs the update script inside a throwaway worktree (`hack/third_party/kubernetes/verify-generated.sh` does `git worktree add "${tmpdir}"`), so `--show-toplevel` resolves to that temp directory and the install is deleted along with it. Every verify run re-downloads the wheel, and verify needs network even when the binary is already cached.

Anchoring on the main working tree instead keeps one copy for both paths:

```bash
ROOT="$(dirname "$(git rev-parse --path-format=absolute --git-common-dir)")"
```

Checked locally against this branch: two consecutive `hack/verify/proto-fmt.sh` runs both print `Downloading clang-format v22.1.8`, and no `bin/clang-format-install` survives either run.

hack/clang-format.sh
```diff
@@ -0,0 +1,77 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit -o nounset -o pipefail
+
+CLANG_FORMAT_VERSION="22.1.8"
```
BenTheElder, line 23: 🤖 **nit** 🟢 – Worth a line saying where the URLs and checksums come from (`https://pypi.org/pypi/clang-format/json`), and that both the hashed URL and the wheel's platform tag change on every bump. Otherwise the next person to move this pin has four opaque paths to reconstruct by hand.

BenTheElder:
Thanks, I agree with this, I think it need some slight fixes.

hack/clang-format.sh
YQ-Wang: Added comment for pypi

hack/clang-format.sh
YQ-Wang: Done

BenTheElder (approved):

PR #469, opened 2026-07-20
atelet: add successful RPC boundary tests

### LINKED ISSUE #217: Expand validation test coverage at atelet's RPC layer
Follow-up from review discussion on #206 (https://github.com/agent-substrate/substrate/pull/206#discussion_r3385245614).

#206 (in review) validates path-construction inputs at atelet's RPC boundary. The negative path is tested through the public surface (`TestRPCBoundariesReject` calls the real `Run`/`Checkpoint`/`Restore` handlers and asserts `InvalidArgument`), but the positive path cannot be asserted through the RPCs: after validation passes, the handlers immediately need object storage, the actor dirs under `/var/lib/ateom-gvisor` (`ateompath.BasePath` is a const), image pulls, and a live ateom.

Making the happy path testable means making those dependencies fakeable (injectable storage and dialer, a redirectable base path). That refactor is the bulk of this issue; the additional tests follow from it. It would also be a good foundation for unit testing the checkpoint/restore flows generally, which today are only covered by e2e.

YQ-Wang:
## Summary
- make ateom client dialing and image pulling injectable
- allow atelet filesystem paths to be redirected in tests
- cover a successful local Run, Checkpoint, and Restore lifecycle at the RPC boundary

## Testing
- `go test ./cmd/atelet ./internal/ateompath`
- `go test ./...`

Closes #217

### DIFF
--- a/cmd/atelet/main_test.go
+++ b/cmd/atelet/main_test.go
@@ -23,19 +23,43 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"syscall"
 	"testing"
 
 	"github.com/agent-substrate/substrate/internal/ateerrors"
 	"github.com/agent-substrate/substrate/internal/ateompath"
+	"github.com/agent-substrate/substrate/internal/imagecache"
 	"github.com/agent-substrate/substrate/internal/proto/ateletpb"
 	"github.com/agent-substrate/substrate/internal/proto/ateompb"
 	"github.com/google/go-cmp/cmp"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/testing/protocmp"
 )
+
+func newAteomPathMapperForTest(t *testing.T) pathMapper {
+	t.Helper()
+	root := t.TempDir()
+	prefix := ateompath.BasePath + string(os.PathSeparator)
+	paths := pathMapper(func(path string) string {
+		if path == ateompath.BasePath {
+			return root
+		}
+		relative, ok := strings.CutPrefix(path, prefix)
+		if !ok {
+			return path
+		}
+		return filepath.Join(root, relative)
+	})
+	if err := os.MkdirAll(paths.mapPath(ateompath.StaticFilesDir()), 0o700); err != nil {
+		t.Fatalf("creating static files dir: %v", err)
+	}
+	return paths
+}
 
 func TestWriteFileAtomic(t *testing.T) {
 	dir := t.TempDir()
@@ -260,21 +284,19 @@ func TestValidateRestoreRequest(t *testing.T) {
 // prove the ordering, it plants a real file at the exact path an invalid hash
 // resolves to: a correctly-ordered fetchAsset validates first and returns an
 // error, while a regression that stats first would find this file and return it
-// with a nil error, failing the test. StaticFilesDir is redirected to a temp
-// dir so the planted path is writable and isolated.
+// with a nil error, failing the test. The service uses a temp-rooted path
+// layout so the derived static-files path is writable and isolated.
 func TestFetchAssetRejectsBadHash(t *testing.T) {
-	orig := ateompath.StaticFilesDir
-	ateompath.StaticFilesDir = t.TempDir()
-	t.Cleanup(func() { ateompath.StaticFilesDir = orig })
+	paths := newAteomPathMapperForTest(t)
 
 	// Invalid (8 chars, not 64) but separator-free, so it resolves to a normal
-	// filename inside the temp StaticFilesDir.
+	// filename inside the temp static-files directory.
 	const badHash = "deadbeef"
-	if err := os.WriteFile(ateompath.RunSCBinaryPath(badHash), []byte("planted"), 0o755); err != nil {
+	if err := os.WriteFile(paths.mapPath(ateompath.RunSCBinaryPath(badHash)), []byte("planted"), 0o755); err != nil {
 		t.Fatalf("planting cache file: %v", err)
 	}
 
-	s := &AteomHerder{}
+	s := &AteomHerder{paths: paths}
 	_, err := s.fetchAsset(context.Background(), assetEntry{SHA256: badHash})
 	if err == nil {
 		t.Fatal("fetchAsset returned a cache hit for an invalid hash; validation must run before the os.Stat early return")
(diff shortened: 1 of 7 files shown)


YQ-Wang:
@dberkov ptal, thanks!

cmd/atelet/main_test.go
```diff
@@ -458,6 +461,156 @@ func TestRPCBoundariesReject(t *testing.T) {
...
+	return d.client, nil
+}
+
+// TestRPCBoundariesAccept runs a successful local lifecycle through the public
+// atelet RPC methods. The fake dependencies stop at the image-fetch and ateom
+// client boundaries; filesystem setup, request projection, checkpoint movement,
+// and sandbox record handling all use the production paths.
+func TestRPCBoundariesAccept(t *testing.T) {
+	origBasePath, origStaticFilesDir := ateompath.BasePath, ateompath.StaticFilesDir
+	ateompath.BasePath = t.TempDir()
```
dberkov: Mutable derived path vars are a footgun (main concern). StaticFilesDir is derived from BasePath at package init, so a test that overrides BasePath must remember to also override every derived var — this test does, but the invariant lives in the test, not the package. Concretely: the in-flight PR #467 adds ImageCacheDir = filepath.Join(BasePath, "image-cache") to that same var block. Once both merge, this test would silently leave ImageCacheDir pointing at
  /var/lib/ateom-gvisor/image-cache while BasePath points at a tempdir.

cmd/atelet/main_test.go
```diff
@@ -458,6 +461,156 @@ func TestRPCBoundariesReject(t *testing.T) {
 	})
 }
```
dberkov, line 482: Global mutation rules out t.Parallel. Fine today (no parallel tests in the package), but worth a one-line comment on the test so nobody adds t.Parallel() to a sibling test that touches ateompath paths.

cmd/atelet/oci.go
```diff
@@ -57,7 +56,11 @@ const (
 	ActorIDFileName = "actor-id"
 )
 
-func prepareOCIDirectory(ctx context.Context, pullCache *memorypullcache.MemoryPullCache, actorUID, containerName, ref string, command, args []string, env []string, annotations map[string]string, netns string, identityDir string, durableDirVolumeMounts []*ateletpb.VolumeMount) error {
+type imagePuller interface {
+	Fetch(context.Context, string) (io.ReadCloser, *v1.Config, error)
+}
+
+func prepareOCIDirectory(ctx context.Context, pullCache imagePuller, actorUID, containerName, ref string, command, args []string, env []string, annotations map[string]string, netns string, identityDir string, durableDirVolumeMounts []*ateletpb.VolumeMount) error {
```
dberkov: could you please wait for #467 be committed first.

cmd/atelet/main_test.go
```diff
@@ -458,6 +461,156 @@ func TestRPCBoundariesReject(t *testing.T) {
...
+	}
+	runscSum := fmt.Sprintf("%x", sha256.Sum256([]byte("fake runsc")))
+	runscPath := ateompath.RunSCBinaryPath(runscSum)
+	if err := os.WriteFile(runscPath, []byte("fake runsc"), 0o755); err != nil {
+		t.Fatalf("seeding runsc cache: %v", err)
+	}
+
+	client := &fakeAteomClient{}
+	dialer := &fakeAteomDialer{client: client}
+	s := &AteomHerder{ateomDialer: dialer, pullCache: fakeImagePuller{}}
```
dberkov: Nil GCS clients are safe only implicitly. The AteomHerder fixture omits anonGCSClient/gcsClient; that's sound because the runsc binary is pre-seeded (so ensureSandboxAssets cache-hits) and checkpoint type is LOCAL — but a future edit that causes a cache miss would nil-panic rather than fail with a message. A one-line comment stating why nil is OK would save the next person a debugging round.


 // GCS clients stay nil: the runsc asset is pre-seeded above (so
  // ensureSandboxAssets never downloads) and the checkpoint/restore legs
  // use CHECKPOINT_TYPE_LOCAL, so no code path touches object storage.

cmd/atelet/main_test.go
YQ-Wang: Updated the derived paths to functions of BasePath and added a regression test covering StaticFilesDir, ImageCacheDir, and RunSCBinaryPath.

cmd/atelet/main_test.go
YQ-Wang: Added a comment that the test must not run in parallel because it redirects the package-wide BasePath.

cmd/atelet/main_test.go
YQ-Wang: Added the requested comment explaining why the nil GCS clients are safe in this fixture.

cmd/atelet/oci.go
YQ-Wang: Rebased onto merged #467 and updated the test seam to the new EnsureImage API.

internal/ateompath/ateompath.go
```diff
@@ -19,25 +19,26 @@ import (
...
 
-const (
-	// The base path.  This is both the path of the root shared folder on the
-	// host filesystem, and when it is mounted into ateom and atelet containers.
-	BasePath = "/var/lib/ateom-gvisor"
-)
+// BasePath is both the path of the root shared folder on the host filesystem,
+// and where it is mounted into ateom and atelet containers. It is a variable
+// so tests can redirect filesystem operations to a temporary directory.
+var BasePath = "/var/lib/ateom-gvisor"
```
dberkov: making mutable global variable might introduce issues. Can it stay const or could you make it Method too, similar to StaticFilesDir()?

dberkov (commented):
minor comment, could you please fix it.

internal/ateompath/ateompath.go
YQ-Wang: Thanks for the feedback. I kept BasePath constant and added an instance-local path mapper for tests, so they no longer mutate global state.

BenTheElder (commented):
Are we injecting a full fake filesystem? If not, this isn't really a unit test, these methods may interact with other paths than ateom-path.


Ordinarily I'm all for test coverage ... but I'm not sure I understand what sort of bugs we intend to catch with this, while we're making the code more indirect.

I don't think we should "unit" test main(). And I'm not sure integration testing atelet makes much sense (vs e2e with ateom).

EDIT: keep in mind atelet now also does things like `mount` for OCI layers, which is not at all unit-test like, and I think it will be unnecessarily complex to fake out interfaces like this.

YQ-Wang:
> Are we injecting a full fake filesystem? If not, this isn't really a unit test, these methods may interact with other paths than ateom-path.
> 
> Ordinarily I'm all for test coverage ... but I'm not sure I understand what sort of bugs we intend to catch with this, while we're making the code more indirect.
> 
> I don't think we should "unit" test main(). And I'm not sure integration testing atelet makes much sense (vs e2e with ateom).

@BenTheElder Thanks for the feedback. Following #217, the goal was to cover atelet’s request handling and local checkpoint/restore flow. I checked the existing e2e suite and confirmed that it already covers the Run, Checkpoint, and Restore lifecycle with real atelet and ateom. Given that coverage, I agree the additional test seam may not justify the added indirection.

PR #474, opened 2026-07-20
Clear a never-written snapshot when suspending a dangling worker

HavenXia:
When the worker pod disappeared before Checkpoint was called, the suspend workflow skipped the atelet call but still let finalize promote the reserved InProgressSnapshot URI to LatestSnapshotInfo, overwriting the last good snapshot with one that was never written.  This could break the next resume call of this actor.

Now the dangling-worker path crashes the actor via `crashActor` too, and crashActor clears any reserved snapshot URI so `FinalizeSuspendedStep` won't overwrite the last good snapshot.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/crash.go
+++ b/cmd/ateapi/internal/controlapi/crash.go
@@ -51,6 +51,8 @@ func crashActor(ctx context.Context, st store.Interface, atespace, actorName str
 		return fmt.Errorf("while loading actor to crash: %w", err)
 	}
 	actor.Status = ateapipb.Actor_STATUS_CRASHED
+	// InProgressSnapshot is kept for debugging; failed workflow
+	// steps must never promote it to LatestSnapshotInfo.
 	// TODO(zoezhao):
 	// 1. If the Actor crashed because the worker is unhealthy,
 	//    free the worker and mark it as unhealthy(or delete it)
--- a/cmd/ateapi/internal/controlapi/workflow_suspend.go
+++ b/cmd/ateapi/internal/controlapi/workflow_suspend.go
@@ -119,8 +119,11 @@ func (s *CallAteletSuspendStep) Execute(ctx context.Context, input *SuspendInput
 	ateletConn, err := s.dialer.DialForWorker(state.Actor.GetAteomPodNamespace(), state.Actor.GetAteomPodName())
 	if err != nil {
 		if errors.Is(err, ErrWorkerPodNotFound) {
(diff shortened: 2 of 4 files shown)


cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -51,6 +51,8 @@ func crashActor(ctx context.Context, st store.Interface, atespace, actorName str
 		return fmt.Errorf("while loading actor to crash: %w", err)
 	}
 	actor.Status = ateapipb.Actor_STATUS_CRASHED
+	// No checkpoint can be in flight anymore, clean any reserved snapshot URI
+	actor.InProgressSnapshot = ""
```
zoez7: Can we leave this field for debugging purposes?

cmd/ateapi/internal/controlapi/workflow_suspend.go
```diff
@@ -119,8 +119,11 @@ func (s *CallAteletSuspendStep) Execute(ctx context.Context, input *SuspendInput
 	ateletConn, err := s.dialer.DialForWorker(state.Actor.GetAteomPodNamespace(), state.Actor.GetAteomPodName())
 	if err != nil {
 		if errors.Is(err, ErrWorkerPodNotFound) {
-			slog.Warn("Skipping suspend for dangling worker pod", "namespace", state.Actor.GetAteomPodNamespace(), "pod", state.Actor.GetAteomPodName())
-			return nil
+			slog.Warn("Worker pod gone before checkpoint, crashing actor", "namespace", state.Actor.GetAteomPodNamespace(), "pod", state.Actor.GetAteomPodName())
```
zoez7: nit: use slog.ErrorCtx instead?

zoez7 (dismissed):

cmd/ateapi/internal/controlapi/crash.go
zoez7: (keep the field as InProgressSnapshot, but do not promote it to LatestSnapshotInfo)

cmd/ateapi/internal/controlapi/workflow_suspend.go
HavenXia: Update both this and the above missing namespace scenario.

cmd/ateapi/internal/controlapi/crash.go
HavenXia: SG, done.

PR #475, opened 2026-07-21
Release the worker and clear the actor's binding when crashing an actor.

zoez7:
This will free up the worker to which the crashed actor was assigned, so it can be picked up by other actors.

Minor fix: 
* Added functional test to make sure a crashed actor can be deleted.

### DIFF
--- a/cmd/ateapi/internal/controlapi/crash.go
+++ b/cmd/ateapi/internal/controlapi/crash.go
@@ -96,14 +96,14 @@ func releaseWorker(ctx context.Context, st store.Interface, actor *ateapipb.Acto
 	if err != nil {
 		return fmt.Errorf("while getting worker to release: %w", err)
 	}
-
 	wass := worker.GetAssignment()
 	if wass == nil {
 		slog.WarnContext(ctx, "Worker's assignment is already nil, skipping release", slog.String("worker", podUid))
 		return nil
 	}
+	// Only free it if it still belongs to us
 	if wass.GetActor().GetAtespace() != actor.GetMetadata().GetAtespace() || wass.GetActor().GetName() != actor.GetMetadata().GetName() {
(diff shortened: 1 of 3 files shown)


cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -44,26 +45,59 @@ func maybeCrashActor(ctx context.Context, st store.Interface, atespace, actorNam
...
+			slog.WarnContext(ctx, "Worker already gone while crashing actor, skipping release", slog.String("worker", podUid))
+		} else {
+			errCollected = append(errCollected, fmt.Errorf("while getting worker to release: %w", err))
+		}
+		return errors.Join(errCollected...)
+	}
+	wass := worker.GetAssignment()
+	if wass == nil {
+		slog.WarnContext(ctx, "Worker's assignment is already nil, skipping release", slog.String("worker", podUid))
+	} else if wass.GetActor().GetAtespace() == atespace && wass.GetActor().GetName() == actorName {
```
HavenXia: nit: can we add a comment same as other workflows on the same logic - `// Only free it if it still belongs to us`

HavenXia (commented):
Checking the `FinalizeSuspendedStep`'s implementation in `workflow_suspend.go`, I think we should free the worker before free the actor to make it idempotent.

The current setup in crash.go, if the actor releasing succeeds and the worker releasing failed, then we can never find the worker since the `podNamespace` variables are local variables so next call of `crashActor` can never find the orphaned worker.

What do you think?

cmd/ateapi/internal/controlapi/crash.go
zoez7: Done.

HavenXia (approved):

PR #480, opened 2026-07-21
fix: both atelet and ate-apiserver leak gRPC connections

### LINKED ISSUE #470: Both atelet and ate-apiserver leak gRPC connections
Both `atelet` and `ate-apiserver` cache gRPC connections in a fixed-size LRU cache (k8s.io/utils/lru.Cache), but never close connections when they are evicted from the cache. When the LRU evicts the oldest entry, the underlying gRPC connection is silently dropped without calling conn.Close(), leaking the underlying connection.

ericdbishop:
Fixes #470 

- [X] Tests pass

Gen AI disclosure: Just used for adversarial review.


cmd/atelet/main.go
```diff
@@ -102,7 +102,13 @@ func main() {
...
 
 	ateomDialer := &AteomDialer{
-		conns: lru.New(256),
+		conns: lru.NewWithEvictionFunc(256, func(key lru.Key, value any) {
+			// Close connection when evicting from cache.
+			conn, ok := value.(*grpc.ClientConn)
+			if ok {
+				conn.Close()
+			}
+		}),
```
ericdbishop: Assuming this looks ok, not sure if there is a common util file I could move this function to so it's not duplicated across files.

cmd/atelet/main.go
```diff
@@ -102,7 +102,13 @@ func main() {
 	go serverboot.StartMetricsServer(ctx, serverboot.MetricsServerOptions{Addr: *metricsListenAddr})
 
 	ateomDialer := &AteomDialer{
-		conns: lru.New(256),
+		conns: lru.NewWithEvictionFunc(256, func(key lru.Key, value any) {
+			// Close connection when evicting from cache.
+			conn, ok := value.(*grpc.ClientConn)
+			if ok {
+				conn.Close()
```
ericdbishop: Add debug log if this fails?

cmd/ateapi/internal/controlapi/dialer.go
```diff
@@ -40,7 +40,13 @@ func NewAteletDialer(workerIndexer cache.Indexer, ateletIndexer cache.Indexer) *
 	return &AteletDialer{
 		workerIndexer: workerIndexer,
 		ateletIndexer: ateletIndexer,
-		ateletConns:   lru.New(1024),
+		ateletConns: lru.NewWithEvictionFunc(1024, func(key lru.Key, value any) {
+			// Close connection when evicting from cache.
+			conn, ok := value.(*grpc.ClientConn)
```
keithmattix: nit: should we log when either the assertion or Close() fails?

cmd/atelet/main.go
```diff
@@ -806,6 +812,7 @@ func toAteomReadyz(in *ateletpb.Readyz) *ateompb.Readyz {
 	return out
 }
 
+// AteomDialer handles gRPC connections to Ateom pods.
```
keithmattix, line 1800: Do we really need this comment?

keithmattix (commented):
Would be nice to have a test to confirm that the connections don't leak anymore

cmd/atelet/main.go
ericdbishop: I had just seen a similar comment in the ateapi file: https://github.com/ericdbishop/substrate/blob/5b11a10073e97f0c425bb7bee23955b2d5762dc3/cmd/ateapi/internal/controlapi/dialer.go#L31. I could remove it if preferred.

keithmattix: No worries just curious

keithmattix (approved):

cmd/atelet/main.go
juli4n: If we blindly close on eviction, can this close a connection which is already in use?

ericdbishop: Yes, how long lived are these connections? Maybe a delayed close might help? I also was considering if we should make the cache size configurable for both dialers.

PR #482, opened 2026-07-21
Rework the Lock API so that locks can be auto-renewed.

juli4n:
Instead of requiring callers to specify a fixed deadline for the lock, let the store return a `Lock` object that performs auto-renewal in background.

This fixes an issue when a workflow takes longer than the hard-coded TTL, most commonly when atelet takes long times to initialize an actor (e.g. pulling / untaring large OCI images).

Leases are taken for 30 seconds, but renewed every 10 seconds (TTL/3).

Removing that implicit TTL-based limit means a caller that sets no deadline of its own (e.g. kubectl-ate) could otherwise hang a Resume/Suspend/Pause indefinitely. MaxDeadlineUnaryInterceptor adds a 10-minute ceiling on every RPC as a hang-prevention backstop.

### DIFF
--- a/cmd/ateapi/internal/store/ateredis/ateredis.go
+++ b/cmd/ateapi/internal/store/ateredis/ateredis.go
@@ -75,7 +75,8 @@ type redisClient interface {
 
 // Persistence is a service that stores information about applications in Redis.
 type Persistence struct {
-	rdb redisClient
+	rdb     redisClient
+	lockTTL time.Duration
 }
 
 var _ store.Interface = (*Persistence)(nil)
@@ -83,7 +84,8 @@ var _ store.Interface = (*Persistence)(nil)
 // NewPersistence creates a new Persistence.
 func NewPersistence(redisClient *redis.ClusterClient) *Persistence {
 	return &Persistence{
-		rdb: redisClient,
+		rdb:     redisClient,
+		lockTTL: defaultLockTTL,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -767,24 +769,123 @@ func fetchProtos[M proto.Message](ctx context.Context, master *redis.Client, key
...
+	lastRenewed := time.Now()
+	timer := time.NewTimer(interval)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-timer.C:
+			callCtx, cancel := context.WithTimeout(ctx, interval)
```
BenTheElder: 🟡 🤖 If Redis is unreachable, the workflow can keep running past lease expiry. A retry's `callCtx` allows a full extra `interval`, and the elapsed-TTL check runs only after the attempt returns. So the lease context may not cancel until ~8s after the key expired, when another replica can already hold the lock. The old design cancelled the workflow 2s *before* the TTL.

Fix: declare the lease lost no later than `lastRenewed.Add(ttl)`, e.g. cap the retry `callCtx` at `time.Until(lastRenewed.Add(ttl))`.

cmd/ateapi/main.go
```diff
@@ -53,6 +53,9 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+// maxRPCDeadline is the max deadline for all RPC methods exposed by this server.
+const maxRPCDeadline = 10 * time.Minute
```
BenTheElder, line 57: 🟢 🤖 The PR description says the backstop is 5 minutes; this sets 10. Worth aligning one or the other.

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -767,24 +769,123 @@ func fetchProtos[M proto.Message](ctx context.Context, master *redis.Client, key
...
+			switch {
+			case err == nil && renewed:
+				lastRenewed = time.Now()
+				timer.Reset(interval)
+
+			case err == nil && !renewed:
+				slog.WarnContext(ctx, "lock renewal found lease no longer owned", "key", key)
+				return
+
+			case time.Since(lastRenewed) >= ttl:
```
Jefftree: k8s has precedence in that we explicitly accept that lease is lost well ahead of the ttl duration if renew is deemed unhealthy. Inflight operations generally take time stop (eg: I see that there was an explicit stop at 28s instead of stopping at the full 30s). Getting here means the renew failed multiple times already and another instance is taking over. I understand substate has a different set of requirements so it doesn't need to exactly follow k8s shape, just providing it as reference. K8s uses a [separate Deadline parameter](https://github.com/kubernetes/kubernetes/blob/v1.36.0/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go#L30-L36) kept below the lease duration (defaults to 2/3 of actual deadline) to account for things like clock skew so the holder gives up while there's still time left on a lease lock. This gives the renewal 2 tries before accepting that the renewal isn't working.

Suggestion: keep a padding here

cmd/ateapi/internal/store/ateredis/ateredis.go
Jefftree: +1. Also provided more context on the comment below, maybe also with some buffer time.

cmd/ateapi/internal/store/ateredis/ateredis_test.go
```diff
@@ -775,158 +756,261 @@ func TestListActors_Pagination(t *testing.T) {
...
-func TestAcquireLock_Conflict(t *testing.T) {
+func TestAcquireLock_RenewsUntilClosed(t *testing.T) {
 	mr, s, ctx := setupTest(t)
-	defer mr.Close()
 
 	key := "test-lock"
-	value1 := "token-1"
-	value2 := "token-2"
-	ttl := 10 * time.Second
+	ttl := 300 * time.Millisecond // renewed every ~100ms; overriding s.lockTTL lets the test avoid waiting out the real lockTTL.
```
Jefftree: nit: using a real clock has been sometimes brittle with CI and race conditions. Would recommend eventually transitioning to inject a fake clock that you can deterministically control to test for edge cases.

cmd/ateapi/internal/store/ateredis/ateredis_test.go
```diff
@@ -775,158 +756,261 @@ func TestListActors_Pagination(t *testing.T) {
...
+	// rather than deleting the new owner's lock out from under it.
+	lock.Close()
+
+	if !mr.Exists(key) {
+		t.Errorf("expected the new owner's lock to survive the old Lock's Close, but the key was deleted")
 	}
+}
 
-	// 2. Fast-forward time past TTL
-	mr.FastForward(6 * time.Second)
```
Jefftree, line 920: nit: I'd imagine you want to keep testing the path of actual TTL expiry (crash/give up) resulting in allowing another client to grab the lock? With renewals, maybe that involves two clients now.

Jefftree (commented):
👋 Ben asked me to take a look at this, I work on the leader election library in k8s core. Providing a couple suggestions based on my experience with k8s

cmd/ateapi/internal/store/ateredis/ateredis_test.go
BenTheElder: we _might_ be able to use https://go.dev/blog/testing-time

cmd/ateapi/main.go
juli4n: Done.

cmd/ateapi/internal/store/ateredis/ateredis.go
juli4n: Thanks for the context! Done.

cmd/ateapi/internal/store/ateredis/ateredis_test.go
juli4n: Good idea. Done.

cmd/ateapi/internal/store/ateredis/ateredis_test.go
juli4n: Done.

cmd/ateapi/internal/store/ateredis/ateredis.go
juli4n: Done.

Jefftree:
LGTM!

BenTheElder (approved):

PR #491, opened 2026-07-22
Make router health checks concurrent, bounded, and non-blocking

### LINKED ISSUE #490: Make router Kubernetes health checks bounded and non-blocking
## Expected Behavior

Router dependency health checks should be bounded by a timeout and stop when the router context is canceled.

The `/statusz` endpoint should remain responsive and return the most recently completed health report, even while a dependency check is slow or unavailable.

## Actual Behavior

The Kubernetes health check calls `Discovery().ServerVersion()`, which uses `context.TODO()` internally and therefore does not observe the router context or its cancellation.

The router also holds the health report mutex while performing dependency network calls. If the Kubernetes API accepts the request but never returns a response:

- the health-check goroutine remains blocked;
- `/statusz` blocks while waiting for the report mutex;
- subsequent health checks cannot run;
- router shutdown can remain blocked waiting for the health goroutine.

This was reproduced with an actual client-go `ServerVersion` request against an HTTP server that accepted `/version` but withheld its response. Canceling the health context did not terminate the request, and `Report()` remained blocked until the server response was released.

- Start a router health check using a Kubernetes client whose `/version` endpoint accepts the connection but does not return a response.
- Wait until the health check enters `Discovery().ServerVersion()`, then cancel or expire the router health context.
- Call `Report()` or request `/statusz` and observe that it remains blocked until the Kubernetes `/version` request eventually returns.

## Specifications

- Version: `main` at `45f20f6571f550e814f8df2e86990908921facef`
- Platform: Kind Kubernetes v1.36.1 (`kindest/node:v1.36.1`) on Colima/macOS arm64
- Fault injection: Proxy obtained the real Kind `/version` response and withheld it from the router client

YQ-Wang:
## Summary

- run the Envoy, Kubernetes API, and ATE API dependency checks concurrently, so each health cycle is bounded by the slowest check (about 500 ms) instead of their cumulative time (up to about 1.5 s)
- make the Kubernetes `/version` health check observe the router context and a 500 ms dependency timeout
- perform dependency network calls without holding the health report mutex, allowing `/statusz` to serve the most recently completed report
- preserve health timestamps and counters with an atomic report update after each concurrent check cycle
- add regression coverage for concurrency, timeout, cancellation, report/status responsiveness, and health accounting

## Root cause

The client-go `Discovery().ServerVersion()` helper uses `context.TODO()` internally. If the Kubernetes API accepted `/version` but withheld its response, canceling the router context could not terminate the request. The router also held the report mutex across dependency network calls, so the stalled request blocked `/statusz` and delayed shutdown.

The three dependency checks also ran sequentially. Even with a 500 ms timeout per dependency, one health cycle could therefore take up to about 1.5 seconds. Running them concurrently bounds the cycle by the slowest check, about 500 ms.

## Validation

- `make verify`
- `go test ./cmd/atenet/internal/router -run 'Test(CheckK8s|HealthCheck|HealthStart)' -count=20`
- `go test -race ./cmd/atenet/internal/router -count=1`
- Kind Kubernetes v1.36.1 fault injection: a proxy fetched the real cluster `/version` response and withheld it; `/statusz` remained responsive with HTTP 200, the request was canceled at the 500 ms deadline, and one Kubernetes health failure was recorded

Fixes #490

### DIFF
--- a/cmd/atenet/internal/router/health.go
+++ b/cmd/atenet/internal/router/health.go
@@ -48,6 +48,12 @@ type RouterHealthReport struct {
 	AteAPI ComponentHealth `json:"ate_api"`
 }
 
+type componentHealthCheckResult struct {
+	healthy   bool
+	message   string
+	checkedAt time.Time
+}
+
 // routerHealth periodically checks the dependent services of router to track health
 // status of this component.
 type routerHealth struct {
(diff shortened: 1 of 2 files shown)


haiyanmeng (commented):
@YQ-Wang , thanks for the great work!

Just one comment, should we consider running `checkEnvoy`, `checkK8s`, and `checkAteAPI` concurrently? This could limit the time of each `check` call to ~500ms instead of 1500ms.

HavenXia:
Just curious: should we add liveness/readiness for router itself?

YQ-Wang:
@haiyanmeng Good suggestion, thank you. I updated the three dependency checks to run concurrently, so each health cycle is now bounded by the slowest check (about 500 ms). I also added a regression test for this.

YQ-Wang:
@HavenXia Good point. Dedicated router liveness and readiness probes would be useful, but I think they are better handled in a follow-up with lightweight /livez and /readyz endpoints and clearly defined semantics, rather than reusing /statusz.

cmd/atenet/internal/router/health.go
```diff
@@ -95,31 +101,50 @@ func (rh *routerHealth) Start(ctx context.Context) {
...
+
+	if !envoyResult.healthy {
+		slog.ErrorContext(ctx, "Envoy health check failed", slog.String("msg", envoyResult.message))
+	}
+	if !k8sResult.healthy {
+		slog.ErrorContext(ctx, "Kubernetes API health check failed", slog.String("msg", k8sResult.message))
+	}
+	if !ateResult.healthy {
+		slog.ErrorContext(ctx, "ATE API gRPC health check failed", slog.String("msg", ateResult.message))
 	}
```
haiyanmeng: Can we move this logic to `runComponentHealthCheck`? It would look like:

```
func runComponentHealthCheck(ctx context.Context, failMsg string, check func(context.Context) (bool, string)) componentHealthCheckResult {
	healthy, message := check(ctx)
	if !healthy {
		slog.ErrorContext(ctx, failMsg, slog.String("msg", message))
	}
	return componentHealthCheckResult{
		healthy:   healthy,
		message:   message,
		checkedAt: time.Now(),
	}
}
```

haiyanmeng (commented):
Thanks for making the checks concurrent.

One tiny comment, otherwise looks good to me.

haiyanmeng:
@YQ-Wang , can you update the PR title and description to make it clear that it also makes the checks concurrent? Thanks!

YQ-Wang:
> Yiqing Wang (@YQ-Wang) , can you update the PR title and description to make it clear that it also makes the checks concurrent? Thanks!

@haiyanmeng done, thanks!

cmd/atenet/internal/router/health.go
YQ-Wang: Done

haiyanmeng (approved):
/lgtm
/approve

HavenXia:
Help merge before you have membership propagated :D

PR #492, opened 2026-07-22
feat(otel): add ate.workerpool.workers metric with e2e suite

krisztianfekete:
Part of the preferred split of #433, please check that out if you need wider context. 

This PR adds `ate.workerpool.workers` metric end to end so we have coverage for workers by pool / state / sandbox class, emitted by ateapi. Comes with the few attribute keys it actually uses, unit tests, and an e2e suite that drives an actor lifecycle and checks the metric (plus the two existing platform metrics) actually lands in the collector.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/metrics.go
+++ b/cmd/ateapi/internal/controlapi/metrics.go
@@ -1,0 +1,87 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


EItanya (dismissed):

cmd/ateapi/internal/controlapi/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+		ws, err := workers()
+		if err != nil {
+			// Worker cache unavailable (warmup/reconnect): skip the whole observation.
+			return nil
+		}
+		type key struct{ pool, state, class string }
+		tally := make(map[key]int64)
+		// Seed both states at 0 for every known pool so a saturated or empty pool
+		// reports 0, not an absent series that breaks idle==0 alerts.
+		if listPools != nil {
```
juli4n: Is `listPools == nil` even possible? Why handling it?

krisztianfekete: Fair, it cannot be `nil`, thanks!

juli4n (dismissed):

PR #493, opened 2026-07-22
Actor crash detection (part 1/3): crash an actor if ateom pod has terminated

zoez7:
Part of #292

This is the first CL for detecting a running Actor has crashed, and mark it as so in Redis. The overall plan is to
1. Have each ateom to monitor the Actor container inside
2. If a crash is detected, the Ateom will exit, this allows us to treat ateom and actor failures in the same way.
3. The workerpool syncer already watch all worker pods. It will delete the worker pod and crash the actor.

This CL implements the logic in Workerpool syncer to detect ateom container terminated. There are 2 more PRs coming which will implement the logic to monitor actor process in ateom-gvisor and ateom-microvm.


cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -185,21 +196,58 @@ func isWorkerEligible(pod *corev1.Pod) bool {
...
+	}
+	// Deleting the pod, because the actor network may not have been cleaned up from the Ateom Pod's net NS.
+	if err := s.kubeClient.CoreV1().Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{
+		Preconditions: metav1.NewUIDPreconditions(string(pod.UID)),
+	}); err != nil && !apierrors.IsNotFound(err) {
+		slog.ErrorContext(ctx, "Failed to delete crashed worker pod", slog.Any("err", err))
+	}
+}
+
+func (s *WorkerPoolSyncer) crashActorOnDeadWorker(ctx context.Context, namespace, pool, podName string) error {
```
EItanya: The naming of this function is a bit confusing, it's not really crashing the actor AFAICT, it's marking it crashed. How about `markActorCrashed` or something like that?

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -185,21 +196,58 @@ func isWorkerEligible(pod *corev1.Pod) bool {
...
+		if err := s.crashActorOnDeadWorker(ctx, pod.Namespace, pool, pod.Name); err != nil {
+			slog.ErrorContext(ctx, "Failed to crash actor bound to crashed worker; keeping pod so the resync retries", slog.Any("err", err))
+			return
+		}
+		if err := s.persistence.DeleteWorker(ctx, pod.Namespace, pool, pod.Name); err != nil {
+			slog.ErrorContext(ctx, "Failed to delete crashed worker from store", slog.Any("err", err))
+		}
+	}
+	// Deleting the pod, because the actor network may not have been cleaned up from the Ateom Pod's net NS.
+	if err := s.kubeClient.CoreV1().Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{
```
EItanya, line 243: Maybe I'm totally misunderstanding something, if the `ateom` crashed wouldn't that automatically rollout the pod because we're using a `Deployment` under the hood?

EItanya (commented):
Just some clarifying questions

cmd/ateapi/internal/controlapi/syncer.go
zoez7: Renamed in 290d5b74e5f76bf3f9dcd98068932cda233f983e

cmd/ateapi/internal/controlapi/syncer.go
zoez7: This is because the ateom container's crash will be detected by kubelet, which will restart the `ateom` container in the same pod. 
You are right that once we delete this pod, the `Deployment` will automatically create a new one.

EItanya: 🤦 

And the worry is that there will be dangling networking and/or other resources if we don't manually kill the pod?

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -210,24 +258,36 @@ func (s *WorkerPoolSyncer) releaseActorOnDeadWorker(ctx context.Context, namespa
...
 		if errors.Is(err, store.ErrNotFound) {
 			return nil
 		}
 		return err
 	}
-	// Skip if a concurrent SuspendActor already cleared the pointer.
-	if actor.GetAteomPodNamespace() != namespace || actor.GetAteomPodName() != podName {
+	// Only crash the actor if it's still assigned to the same worker.
+	if actor.GetAteomPodNamespace() == "" || actor.GetAteomPodName() == "" || actor.GetAteomPodUid() == "" {
+		slog.WarnContext(ctx, "cannot release actor on worker", slog.String("ateom pod namesace", actor.GetAteomPodNamespace()), slog.String("ateom name", actor.GetAteomPodName()), slog.String("ateom pod uid", actor.GetAteomPodUid()))
```
sfunkenhauser, line 276: nit: Could we clarify this warning? My understanding isn't that we can't release the actor, it's that the actor is no longer assigned to this worker.

Maybe something along the lines:
"Actor is no longer assigned to worker, skipping crash."

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -210,24 +258,36 @@ func (s *WorkerPoolSyncer) releaseActorOnDeadWorker(ctx context.Context, namespa
...
 		return nil
 	}
-	if actor.Status != ateapipb.Actor_STATUS_CRASHED {
-		actor.Status = ateapipb.Actor_STATUS_SUSPENDED
+	if actor.GetStatus() != ateapipb.Actor_STATUS_RUNNING {
+		slog.WarnContext(ctx, "actor is not in running state")
+		return nil
+	}
+	if actor.GetAteomPodNamespace() != worker.GetWorkerNamespace() || actor.GetAteomPodName() != worker.GetWorkerPod() || actor.GetAteomPodUid() != worker.GetWorkerPodUid() {
```
sfunkenhauser, line 283: nit: This is a similar class of warning / error as above, where the actor was suspended and no longer assigned to the worker. We should likely log in both cases or neither.

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -185,21 +196,58 @@ func isWorkerEligible(pod *corev1.Pod) bool {
...
+	switch {
+	case errors.Is(err, store.ErrNotFound):
+		// Worker already deleted from store, nothing to do.
+		slog.WarnContext(ctx, "worker is already deleted from store")
+	case err != nil:
+		slog.ErrorContext(ctx, "Failed to get worker for terminated ateom; keeping pod so the resync retries", slog.Any("err", err))
+		return
+	case worker.Assignment == nil:
+		// If the worker is not assigned to any actor, delete the worker from pool.
+		if err := s.persistence.DeleteWorker(ctx, pod.Namespace, pool, pod.Name); err != nil {
```
sfunkenhauser, line 228: I don't think this is a new issue, so I don't think we need to put a fix in this PR. I believe there are some race conditions around DeleteWorker. However unlikely, it's possible a different actor gets assigned to this worker after we check the state. So we could potentially be deleting a worker that has an actor assigned. We already check the expected version during updates, but maybe we need to do this for deletes as well?

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -185,21 +196,58 @@ func isWorkerEligible(pod *corev1.Pod) bool {
...
+		}
+	}
+	return false
+}
+
+func (s *WorkerPoolSyncer) handleTerminatedAteom(ctx context.Context, pod *corev1.Pod) {
+	slog.InfoContext(ctx, "Syncer: worker ateom terminated; crashing actor and recycling pod",
+		slog.String("worker", pod.Namespace+"/"+pod.Name))
+	pool := pod.Labels[workerPodLabel]
+	worker, err := s.persistence.GetWorker(ctx, pod.Namespace, pool, pod.Name)
```
sfunkenhauser, line 218: To avoid workers being re-used before we have a chance to clean them up, I wonder if it's worth doing something similar to how I'm going to handle graceful termination.

1) Mark the worker as 'CRASHED' (I'll use 'DRAINING' for graceful termination).
2) In the ateom, be able to detect that it crashed and reject any further suspend / resume attempts.

This mitigates the race conditions around state detection and clean up.

cmd/ateapi/internal/controlapi/syncer.go
zoez7: Yes, for example [cleanupActorNetwork](https://github.com/agent-substrate/substrate/blob/5f6f14c5c8117d1107176bac842481df11fe068f/cmd/ateom-microvm/net.go#L246C24-L246C43) won't be called if ateom exits unexpectedly.

PR #495, opened 2026-07-22
Add PR sizing guidance to CONTRIBUTING.md

HavenXia:
Balance large-PR breakdown against small/bulk-PR consolidation.

- [x] Tests pass

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,6 +47,24 @@ duplicating work that is already in flight.
 extensive review. We are not trying to be discouraging, but we need to make
 sure that we are focused on the most important work.
 
+### Sizing PRs for review
+
+We optimize PRs for easy review — large PRs get broken
+down, small PRs get merged.
+
+* **Large PRs**: split huge changes into a series of smaller PRs, each a
+  logically distinct feature. When the intermediate steps are not useful
+  on their own, keep the change as one PR split into commits at logical
+  break points, and preserve those commits on merge.
+* **Small and bulk PRs**: if you find a typo, review the whole file and
+  fix everything in one pass rather than sending the single edit. Group
+  related typo, doc, and single-line cleanup fixes into one PR rather
+  than opening several small ones for the same area. Maintainers may ask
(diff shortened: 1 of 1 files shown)


EItanya (commented):
Is there some relative tshirt size we can add here in terms of # of lines for clarity, otherwise everyone may come with a different estimation of what these mean

CONTRIBUTING.md
```diff
@@ -47,6 +47,18 @@ duplicating work that is already in flight.
 extensive review. We are not trying to be discouraging, but we need to make
 sure that we are focused on the most important work.
 
+### Sizing PRs for review
+
+We optimize PRs for easy review — large PRs get broken
+down, small PRs get merged.
+
+* **Large PRs**: split huge changes into a series of smaller, focused
```
BenTheElder: I think we need a little more nuance. Sometimes we may implement a large change like atespaces where we want to break it down to multiple _commits_ (and preserve those commits on merge) but we may not want to halfway merge it.

Breaking up large PRs into logically distinct features is encouraged though.

CONTRIBUTING.md
```diff
@@ -47,6 +47,18 @@ duplicating work that is already in flight.
...
+  PRs rather than submitting one large change.
+* **Small and bulk PRs**: group related typo, doc, and single-line cleanup
```
BenTheElder: https://www.kubernetes.dev/docs/guide/pull-requests/#trivial-edits also has some language we can consider

HavenXia: integrated the k8s doc into it.

CONTRIBUTING.md
HavenXia: rephrased to mention logical dinstinction.

HavenXia:
> Is there some relative tshirt size we can add here in terms of # of lines for clarity, otherwise everyone may come with a different estimation of what these mean

Searched around k8s doc and found the size in prow(https://github.com/kubernetes-sigs/prow/blob/main/pkg/plugins/size/size.go), I borrowed that to here, >500 lines are good candidate for breaking down.

PR #498, opened 2026-07-22
fix: use background context for async worker release

mesutoezdil:
AssignWorkerStep starts a background goroutine to release a stale
worker assignment. It builds its context from the request ctx, but
that ctx is cancelled right after the workflow returns (releaseLock
runs on defer). If the goroutine has not called UpdateWorker yet, the
call fails with a cancelled context and the worker keeps the stale
assignment forever. Nothing else clears it while the pod is alive.

Fix: use context.Background() for the release call, same pattern used
in cmd/atenet/internal/router/resumer.go.

Tested with go test ./cmd/ateapi/internal/controlapi/...


EItanya (commented):
This change definitely makes sense to me, I'd love @juli4n or @zoez7 to take a look as they are more familiar with this path

PR #499, opened 2026-07-23
fix(k8sjwt): skip unusable keys in JWKS discovery

aramase:
`discoverKeysForIssuer` now parses each JWK via `parseJWK` and **skips** keys the verifier can't use (unsupported `kty`/curve, malformed params, or no `kid`), failing only when no usable key remains — with distinct errors for `empty JWKS` vs `no usable keys ... (N skipped)`. Matches the go-oidc/jose convention: skip at discovery, fail only when a token's `kid` matches no usable key. Selection stays by `kid` and signatures are still verified, so skipping cannot enable forgery.

## Behavior note

A key with no `kid` is now skipped instead of failing the whole set. Safe: tokens must carry a `kid` and match by exact `kid`, so a no-`kid` key was never selectable anyway.

- [x] Tests pass
- [ ] Appropriate changes to documentation are included in the PR — N/A

### DIFF
--- a/cmd/ateapi/internal/k8sjwt/k8sjwt.go
+++ b/cmd/ateapi/internal/k8sjwt/k8sjwt.go
@@ -402,71 +402,103 @@ func discoverKeysForIssuer(ctx context.Context, httpClient *http.Client, issuer 
 	slog.InfoContext(ctx, "Fetched JWK set", slog.Any("jwkSet", jwkSet))
 
 	var ret []*KeyAndID
+	skipped := 0
 	for _, jwk := range jwkSet.Keys {
-		if jwk.KeyID == "" {
-			return nil, fmt.Errorf("JWKs endpoint returned key without key ID")
+		key, err := parseJWK(jwk)
+		if err != nil {
+			// Skip an unusable key instead of failing the whole issuer; it's safe because
+			// keys are selected by kid and the signature is still verified, so a skipped
+			// key can't be abused. Debug, not Warn: unsupported key types are a normal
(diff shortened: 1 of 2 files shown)


cmd/ateapi/internal/k8sjwt/k8sjwt.go
```diff
@@ -402,73 +402,105 @@ func discoverKeysForIssuer(ctx context.Context, httpClient *http.Client, issuer
...
+	// that fails later as a vaguer "unknown key ID".
+	if len(ret) == 0 {
+		if len(jwkSet.Keys) == 0 {
+			return nil, fmt.Errorf("issuer %q published an empty JWKS", issuer)
 		}
+		return nil, fmt.Errorf("no usable keys in JWKS for issuer %q (%d skipped)", issuer, skipped)
+	}
+
+	if skipped > 0 {
+		slog.InfoContext(ctx, "Skipped unusable JWKs from issuer",
```
BenTheElder: 🤖 **nit** 🟢 – This Info summary fires on every `Verify` (discovery is per-request until key caching lands) — the same per-request noise the PR avoided by putting per-key skips at Debug. Consider Debug here too, or rate-limiting.

BenTheElder (commented):
[AI first pass experiment ...]

cmd/ateapi/internal/k8sjwt/k8sjwt.go
aramase: Yeah, I was on the fence on this one. Moved it to Debug for now, will bump it back to Info once discovery is cached (the TODO above).

zoez7 (approved):

PR #496, opened 2026-07-22
ateom gvisor: drop `privileged: true`

BenTheElder:
Fixes #288

x-ref #174

Also, by not running with `privileged: true` we should stop mounting all host devices, so we can start thinking about passing through specific devices. cc @dims @omeryahud

We do this by:
- Dropping `privileged: true`, which also ensures we're in a private cgroupns (we assume cgroups v2 container ecosystem which does this, cgroup v1 is broadly deprecated).
- Enabling a lot of capabilities and unconfining apparmor (which we had from privileged and still need)
- Setting up interior leaf cgroups and moving ateom itself into one.
- Configuring the gvisor containers to run in leaf cgroups.

To be clear: ateom-gvisor is still *highly* permissioned, it's just not literally `privileged: true` anymore, so we drop unwanted side-effects from that. 

> It's a good idea to open an issue first for discussion.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

This PR was AI assisted. I have tested everything on kind + GKE locally and reviewed the changes.

Issue #450, opened 2026-07-17
Actor lifecycle hooks


sfunkenhauser:
This issue tracks the design and implementation of actor lifecycle hooks.

These are analogous to Kubernetes [container lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/). At a minimum we should support 2 new hooks (names TDB).

1.  PostResume - Triggered right after an actor is resumed.
2. PreSuspend - Triggered right before an actor is suspended.

This area needs more investigation to determine exactly how these hooks should be implemented.

EItanya:
I'm also definitely interested in this, but there's a bit of a wrinkle here i think we need to account for in this design. Specifically, at what point in the snapshot lifecycle does this happen. Before/after makes a very large difference for pre-warming etc.

eliranw:
We are also interested in this.
We would need lifecycle hooks for running commands post-resume + pre-suspend for `cuda-checkpoint`ing GPU actors.
This connects to https://github.com/agent-substrate/substrate/issues/185.

HavenXia:
@sfunkenhauser I can take this one. The upgrade workstream (#473) has to define these semantics anyway: a blue/green model that batch-suspends every running actor, so the `PreSuspend` and `PostResume` are inevitable path on the upgrade.

sfunkenhauser:
> [sfunkenhauser](https://github.com/sfunkenhauser) I can take this one. The upgrade workstream ([#473](https://github.com/agent-substrate/substrate/issues/473)) has to define these semantics anyway: a blue/green model that batch-suspends every running actor, so the `PreSuspend` and `PostResume` are inevitable path on the upgrade.

Is this required for the initial release? Or is it sufficient to push this responsibility to the actor? The actor receives a SIGTERM when the pod is deleted and decides how to proceed. This was the assumption I was proceeding with for graceful termination, and it seems like upgrades would be very similar?

HavenXia:
> Or is it sufficient to push this responsibility to the actor? The actor receives a SIGTERM when the pod is deleted and decides how to proceed. 

I think SIGTERM only exists when termination happens, right? On the suspend path there's no signals to actors, and according to #503 it's a good addition for in memory states. 

>  "required for the initial release"


I agree, it is not for M2. We can implement post M2.

> This was the assumption I was proceeding with for graceful termination, and it seems like upgrades would be very similar?

[...]

PR #502, opened 2026-07-23
feat: CDI-based NVIDIA GPU passthrough into gVisor actor containers

### LINKED ISSUE #627: GPU support for gVisor actors
## Goal
Let actors use NVIDIA GPUs passthrough from their workers.

## How to achieve this 
We need to inject the relevant info to the actor's OCI spec, and then `runsc` proxies the driver. 

##  Suggested Approach
1. Install GPU-Operator on the K8s cluster.
2. Request GPU resources in a workerpool. This should mount the ateom with relevant NVIDIA binaries and tooling and expose the GPUs to the workers.
3. When starting, ateom runs `nvidia-ctk cdi generate`,and takes the device nodes, mounts, hooks, and env and emits them to a CDI file. Hooks run using nvidia-cdi-hook as part of createContainer.
4. Merge into the actor's OCI spec.
5. Ateom runs `runsc create` to build the sandbox with the relevant GPU devices.

## The core tension
Running NVIDIA CDI inside an **unprivileged** gVisor worker has the following issues:

1. **CDI createContainer hooks run in `runsc`'s gofer**, which drops to a minimal cap set. 
2. **`update-ldcache` hook mounts `/proc`** to run the hook. we need /proc to be unmasked to support this.
3. **The current ateom-gvisor worker image is distroless**, but the NVIDIA toolkit binaries are glibc-dynamic  so they can't be exec'd directly; they can be run via a staged glibc loader from a shared volume.

**Net:** GPU on the unprivileged worker requires `hostUsers: false` + `procMount: Unmasked` and that can break cgorup delegation. 

## Known costs / open issues
- **Toolkit host-path portability.** gpu-operator/EKS install at `/usr/local/nvidia/toolkit`; GKE keeps it under `/home/kubernetes/bin/nvidia`. Made configurable via env. 
- **Snapshot/restore with GPU** out of scope currently (suspend/resume of a GPU actor).

eliranw:
## Summary

- `atecontroller` propagates a pool's `nvidia.com/gpu` request onto the `ateom` container and mounts the host NVIDIA toolkit read-only (path overridable via `ATE_NVIDIA_TOOLKIT_HOST_PATH`)
- `ateom-gvisor` generates a CDI spec with `nvidia-ctk` and injects the device nodes, driver-library mounts, and env into each actor container's OCI spec
- runs the CDI `createContainer` hooks except `update-ldcache` which needs a privileged ateom, staging the SONAME symlinks it would create from each library's ELF `DT_SONAME`
- enables `runsc --nvproxy` at sandbox creation

Requesting `nvidia.com/gpu` on the pool is the only configuration needed; a pool that requests N GPUs makes all N usable.

Two details of the CDI spec are worth calling out, because getting either wrong fails at runtime rather than at parse time. `nvidia-ctk` leaves `major`/`minor` unset — CDI delegates that to the OCI runtime — so each device node is resolved by stat-ing the host; without it the actor gets `0,0` char devices and NVML reports it cannot communicate with the driver. And it emits per-index, per-UUID, and `all` devices that repeat the same nodes, so only `all` is applied. The spec is plain JSON, so `encoding/json` suffices and no CDI library is vendored.

`update-ldcache` is the one hook that cannot run here: its `ldconfig` unshares a mount namespace and mounts a private `/proc`, which `mount_too_revealing()` rejects under the pod's masked `/proc`. Permitting it would need `procMount: Unmasked`, which Kubernetes only allows with `hostUsers: false`, and that user namespace breaks the per-actor cgroup delegation from #496. Skipping it avoids the whole chain, so a GPU worker keeps the same posture as any other unprivileged gVisor worker. `create-symlinks` and `enable-cuda-compat` still run unmodified.

`--nvproxy` must be set when the sandbox is created — the `pause` container, which holds no GPU devices — so runsc's auto-detection never fires on its own; without the flag the GPU subcontainer crashes the sentry on start. GPU detection matches any device index rather than assuming `/dev/nvidia0`, since a worker sharing a multi-GPU node can be assigned `/dev/nvidia2` and `/dev/nvidia3`.

GPU pools must set `spec.ateomImage` to a glibc build (`KO_DEFAULTBASEIMAGE=debian:stable-slim ko build ./cmd/ateom-gvisor`) because the distroless default cannot exec `nvidia-ctk`; the default base is unchanged for every other pool. `atelet` also has to run on the GPU nodes to restore actors there, so its DaemonSet needs a toleration for whatever taint they carry. Both are documented in the API guide rather than defaulted.

## Testing

- `make test`
- `env -u NO_COLOR make verify`
- Real GPU, Tesla T4 / driver 580.65.06, through the full actor flow: `nvidia-smi`, `vectorAdd`, `nbody` at 3.77 TFLOP/s, PyTorch matmul via cuBLAS at 3.5 TFLOP/s (T4 peak FP32 is ~8.1, so no measurable sandbox penalty)
- Actor whose entrypoint runs the CUDA sample directly exits 0, confirming the injected env reaches the workload without help from the test harness
- Two workers holding two GPUs each on one 4-GPU node see disjoint device sets

Snapshot and restore work when the workload holds no CUDA context. A live CUDA context cannot be checkpointed - gVisor fails with `can't save with live nvproxy clients` and the failed checkpoint terminates the sandbox so a GPU actor can only be suspended between CUDA workloads. Documented as a known limitation in the API
guide; a follow-up issue will track lifting it via `cuda-checkpoint`.

Fixes #627 

- [x] Tests pass

### DIFF
--- a/cmd/atecontroller/internal/controllers/workerpool_apply.go
+++ b/cmd/atecontroller/internal/controllers/workerpool_apply.go
@@ -15,6 +15,8 @@
 package controllers
 
 import (
+	"os"
+
 	corev1 "k8s.io/api/core/v1"
 	appsv1ac "k8s.io/client-go/applyconfigurations/apps/v1"
 	corev1ac "k8s.io/client-go/applyconfigurations/core/v1"
@@ -145,6 +147,7 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool, otel ateomOTelSettin
 
 	applyWorkerPoolPodTemplate(podSpecAC, containerAC, wp.Spec.Template)
 	maybeApplyMicroVMPodShape(podSpecAC, containerAC, wp.Spec.SandboxClass)
+	maybeApplyGPUPodShape(podSpecAC, containerAC, wp.Spec.Template, wp.Spec.SandboxClass)
 	podSpecAC.WithContainers(containerAC)
 	podSpecAC.WithTerminationGracePeriodSeconds(workerTerminationGracePeriodSeconds)
 
@@ -221,7 +224,8 @@ func fieldRefEnv(name, fieldPath string) *corev1ac.EnvVarApplyConfiguration {
 // veth and nftables rules (NET_ADMIN/NET_RAW), and the OCI rootfs is unpacked
 // and device nodes created as root over image-owned trees
 // (DAC_OVERRIDE/FOWNER/CHOWN/MKNOD). This replaces the former privileged worker;
-// the default seccomp and AppArmor profiles are sufficient (no Unconfined).
+// seccomp stays at the runtime default, but AppArmor must be Unconfined (see
+// ateomSecurityContext) since runsc's own mounts trip the default profile.
 var ateomGvisorCapabilities = []corev1.Capability{
 	"NET_ADMIN", "SYS_ADMIN", "SYS_CHROOT", "SYS_PTRACE",
 	"SETUID", "SETGID", "SETPCAP", "DAC_OVERRIDE",
@@ -299,6 +303,92 @@ func maybeApplyMicroVMPodShape(
 		WithOperator(corev1.TolerationOpEqual).
 		WithValue(string(atev1alpha1.SandboxClassMicroVM)).
 		WithEffect(corev1.TaintEffectNoSchedule))
+}
+
+// nvidiaToolkitContainerPath is where the host toolkit is mounted inside the
+// worker; ateom-gvisor's toolkitDir must match this. It sits outside
+// /usr/local/nvidia because the GPU device plugin mounts that tree into the
+// container read-only, and a mount cannot create its own mountpoint there, so
+// mounting under it only works when the toolkit happens to live inside the
+// directory the plugin mounts.
+const nvidiaToolkitContainerPath = "/opt/nvidia-toolkit"
+
+// defaultNvidiaToolkitHostPath is where gpu-operator installs the toolkit
+// (toolkit.installDir defaults to /usr/local/nvidia). It is deliberately not the
+// container path above: the two are independent, since where the node keeps the
+// toolkit says nothing about where we can mount it.
+const defaultNvidiaToolkitHostPath = "/usr/local/nvidia/toolkit"
+
+// nvidiaDriverRootEnv names the directory the GPU device plugin mounts the driver
+// into a pod at. ateom derives the driver library and binary paths from it, both of
+// which nvidia-ctk needs to generate a CDI spec. Only set it when the cluster's
+// device plugin does not use the /usr/local/nvidia convention; the controller
+// forwards its own value onto GPU worker pods.
+const nvidiaDriverRootEnv = "ATE_NVIDIA_DRIVER_ROOT"
+
+// nvidiaToolkitHostPath is where the NVIDIA container toolkit lives on the node.
+// It is platform-specific: gpu-operator and EKS install it at
+// /usr/local/nvidia/toolkit, while GKE keeps NVIDIA assets under
+// /home/kubernetes/bin/nvidia, so it is overridable via the
+// ATE_NVIDIA_TOOLKIT_HOST_PATH env var on the controller. We mount it read-only
+// so nvidia-ctk / nvidia-cdi-hook match whatever toolkit/driver the cluster runs.
+var nvidiaToolkitHostPath = envOrDefault("ATE_NVIDIA_TOOLKIT_HOST_PATH", defaultNvidiaToolkitHostPath)
+
+func envOrDefault(key, def string) string {
+	if v := os.Getenv(key); v != "" {
+		return v
+	}
+	return def
+}
+
+// maybeApplyGPUPodShape shapes a gVisor worker pod that requests a GPU so ateom
+// can inject the GPU into actors via CDI. It mounts the host NVIDIA toolkit
+// (version-matched to the node) read-only, for the glibc-based ateom image to run
+// directly. The pod keeps the same security posture as any other gVisor worker.
+// No-op for non-GPU pools and non-gVisor classes; an empty class defaults to
+// gVisor (WorkerPoolSpec kubebuilder default).
+func maybeApplyGPUPodShape(
+	podSpecAC *corev1ac.PodSpecApplyConfiguration,
+	containerAC *corev1ac.ContainerApplyConfiguration,
+	tmpl *atev1alpha1.WorkerPoolPodTemplate,
+	sandboxClass atev1alpha1.SandboxClass,
+) {
+	if sandboxClass != atev1alpha1.SandboxClassGvisor && sandboxClass != "" {
+		return
+	}
+	if !templateRequestsGPU(tmpl) {
+		return
+	}
+	// Mount the host NVIDIA toolkit (version-matched to the node) read-only.
+	containerAC.WithVolumeMounts(corev1ac.VolumeMount().
+		WithName("nvidia-toolkit").
+		WithMountPath(nvidiaToolkitContainerPath).
+		WithReadOnly(true))
+	podSpecAC.WithVolumes(corev1ac.Volume().
+		WithName("nvidia-toolkit").
+		WithHostPath(corev1ac.HostPathVolumeSource().
+			WithPath(nvidiaToolkitHostPath).
+			WithType(corev1.HostPathDirectory)))
+	// Only propagated when set, so a default deployment adds no env to worker pods.
+	if root := os.Getenv(nvidiaDriverRootEnv); root != "" {
+		containerAC.WithEnv(corev1ac.EnvVar().WithName(nvidiaDriverRootEnv).WithValue(root))
+	}
+}
+
+// templateRequestsGPU reports whether the pool template requests one or more
+// nvidia.com/gpu devices (limits or requests).
+func templateRequestsGPU(tmpl *atev1alpha1.WorkerPoolPodTemplate) bool {
+	if tmpl == nil || tmpl.Resources == nil {
+		return false
+	}
+	const gpu = corev1.ResourceName("nvidia.com/gpu")
+	if q, ok := tmpl.Resources.Limits[gpu]; ok && !q.IsZero() {
+		return true
+	}
+	if q, ok := tmpl.Resources.Requests[gpu]; ok && !q.IsZero() {
+		return true
+	}
+	return false
 }
 
 func applyWorkerPoolPodTemplate(
--- a/cmd/ateom-gvisor/gpu.go
+++ b/cmd/ateom-gvisor/gpu.go
@@ -1,0 +1,516 @@
+//go:build linux
+
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"cmp"
+	"context"
+	"debug/elf"
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"golang.org/x/sys/unix"
+
+	"github.com/agent-substrate/substrate/internal/ateompath"
+)
+
+// toolkitDir is where the host's NVIDIA container toolkit (nvidia-ctk,
+// nvidia-cdi-hook) is mounted into the worker pod — read-only from the node, so
+// the binaries match whatever toolkit/driver the cluster installed. A var (not
+// const) so tests can point it at a fixture directory.
+var toolkitDir = "/opt/nvidia-toolkit"
+
+// gpuDeviceGlob matches the per-GPU device nodes. The device plugin can assign any
+// indices (a worker sharing a multi-GPU node may get /dev/nvidia2,3 with no
+// /dev/nvidia0), so detection must not assume index 0. The [0-9] excludes the control
+// nodes (/dev/nvidiactl, /dev/nvidia-uvm). A var so tests can point it at a fixture.
+var gpuDeviceGlob = "/dev/nvidia[0-9]*"
+
+// nvidiaDriverRoot is where the GPU device plugin mounts the driver into the pod.
+// GKE and gpu-operator both use /usr/local/nvidia, but that is a convention rather
+// than a contract, so it is overridable via ATE_NVIDIA_DRIVER_ROOT (propagated onto
+// GPU worker pods by the controller).
+var nvidiaDriverRoot = cmp.Or(os.Getenv("ATE_NVIDIA_DRIVER_ROOT"), "/usr/local/nvidia")
+
+// Both directories are load-bearing for CDI generation, not just for completeness:
+// without the library path nvidia-ctk cannot load libnvidia-ml.so.1 to enumerate the
+// GPUs and generation fails outright, and without the bin path (which it discovers
+// via PATH) the generated spec carries libraries but no nvidia-smi.
+var (
+	driverLibDir = filepath.Join(nvidiaDriverRoot, "lib64")
+	driverBinDir = filepath.Join(nvidiaDriverRoot, "bin")
+)
+
+const cdiOutputDir = "/run/ate-cdi"
+
+// enabledCDIHooks is the set of CDI createContainer hooks the actor runs. It is an
+// allowlist because the toolkit is mounted from the host, so its version is the
+// cluster's choice: a newer one can emit hooks that have never been reviewed
+// against this worker's unprivileged posture, and those must not run by default.
+// update-ldcache is absent deliberately — its ldconfig needs a private /proc
+// mount, which the pod's masked /proc rejects, so the SONAME symlinks it would
+// create are staged directly into the rootfs instead.
+var enabledCDIHooks = map[string]bool{
+	"create-symlinks":    true,
+	"enable-cuda-compat": true,
+}
+
+// toolkitBinary resolves a toolkit command to an executable path, preferring the
+// unwrapped ".real" binary the NVIDIA toolkit ships: the plain name is often a
+// /bin/sh wrapper. The ateom image is glibc-based (debian), so the glibc-dynamic
+// toolkit binaries run directly — no ld-linux loader shim is needed.
+func toolkitBinary(name string) string {
+	if real := filepath.Join(toolkitDir, name+".real"); fileExists(real) {
+		return real
+	}
+	return filepath.Join(toolkitDir, name)
+}
+
+func fileExists(p string) bool { _, err := os.Stat(p); return err == nil }
+
+// dropEnvVar returns env with every "KEY=..." entry for the given key removed.
+func dropEnvVar(env []string, key string) []string {
+	prefix := key + "="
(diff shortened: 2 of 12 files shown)


dims:
@eliranw is there a way to do this without introducing all the new vendored components? (21k lines of code is too much!)

eliranw:
It seems like we decided to commit vendor packages in this repo. I think it’s best to use the cdi package instead of reimplementing it. As far as I understand, the main downside of committing the vendor folder (instead of gitignoring it) is having a large diff. Do you have a different solution?

To make it easier to review, I moved the vendor changes into a separate commit. I noticed many other PRs do the same thing - does this approach make sense to you?

dims:
@eliranw it's not whether the vendor-ed stuff is a separate commit or not. just that we should avoid adding stuff if not really required.

see for example - https://github.com/agent-substrate/substrate/compare/main...dims:substrate:gpu-nocdi

BenTheElder:
thanks for working on this

I'd also like to keep imported deps minimal.

I started looking at uVM without privileged and we may have a bigger problem there ... ideally I don't want to require a plugin on the host but we need permission to talk to KVM.

I may have an answer to that wrinkle but broadly speaking I'm also hoping that even if not implemented yet we can figure out how we'll extend this to uVM later, as I'm also working on closing the existing gap (durable directory support) at higher priority ...

eliranw:
Thanks @dims and @BenTheElder. Dropped the vendored CDI as you both suggested and re-tested on a GPU node.

I’ve also gated the whole thing to gVisor-only for now so the NVIDIA toolkit mount only lands on gVisor pools, meaning a micro-VM pool that requests a GPU won’t pick up the nvidia toolkit.

On uVM - I'll start looking into that

eliranw:
@BenTheElder @dims @omeryahud - I've updated this PR to reflect #496 (unprivileged ateom).

I had a few issues getting the GPU CDI hooks working, but right now GPU pools keep exactly the same posture as any other gVisor worker - no user namespace, no procMount: Unmasked, and per-actor cgroup delegation still working. The issue was that only one CDI hook was failing: update-ldcache, which needs the private /proc; skipping just that hook (and writing the SONAME symlinks from DT_SONAME instead) avoids the issue I had. The other CDI hooks run unmodified. I also needed a debian:stable-slim gvisor image to run the cdi generation tool.
I also dropped the vendored CDI library per your earlier comment.

I did a full test end to end on a cluster with a T4 node workerpool on GKE with different NVIDIA tooling to check compatibility: nvidia-smi, vectorAdd, nbody, and a PyTorch cuBLAS matmul, plus two workers with two GPUs each on one 4-GPU node to check isolation betweek workers on a single node. I've added details in the updated description.

docs/api-guide.md
```diff
@@ -53,31 +59,48 @@ metadata:
...
       limits:
         cpu: "1"
         memory: 2Gi
+        # (2) claim a GPU — this request is what triggers GPU passthrough
+        nvidia.com/gpu: "1"
 ```
 
+`atecontroller` propagates the request onto the `ateom` container and mounts the
+host NVIDIA toolkit into the pod. `ateom-gvisor` then generates a CDI spec with
+`nvidia-ctk` and injects the GPU — device nodes, driver libraries, and env — into
```
omeryahud:
```suggestion
`nvidia-ctk` and injects the GPU device nodes, driver libraries, and env into
```

docs/api-guide.md
```diff
@@ -53,31 +59,48 @@ metadata:
...
+each actor container's OCI spec, and runs `runsc` with `--nvproxy` so CUDA and NVML
```
omeryahud, line 87: When there are multiple containers within an actor, all of them get GPU access? I assume the same if for other resources (multiple actor containers share the same resource pool).

Let's document that now, and maybe address it in the future when/if individual actor containers get resource limitation.

eliranw: Yes, all containers receive acess. I tried it with a 2-container ActorTemplate and both containers saw the same devices and ran CUDA fine, documented this as well.

cmd/ateom-gvisor/main.go
```diff
@@ -104,11 +104,19 @@ func do(ctx context.Context) error {
...
+	// it with real accounting.
 	if err := setupCgroupDelegation(ctx); err != nil {
 		return fmt.Errorf("while setting up cgroup delegation: %w", err)
 	}
 
+	// If this worker has a GPU, enable nvproxy so every sandbox boots GPU-ready
+	// (see nvproxyGlobalArgs). Set once here at boot to avoid a data race on the
+	// concurrent RunWorkload path.
+	if gpuPresent() {
+		runscNvproxy = true
```
omeryahud: I don't like passing arguments in a back channel, let's pass it as a function argument to `maybeInjectGPU`.

You can leave the check here for logging purposes

cmd/atecontroller/internal/controllers/workerpool_apply.go
```diff
@@ -206,6 +210,71 @@ func maybeApplyMicroVMPodShape(
...
+	if v := os.Getenv(key); v != "" {
+		return v
+	}
+	return def
+}
+
+// maybeApplyGPUPodShape shapes a gVisor worker pod that requests a GPU so ateom
+// can inject the GPU into actors via CDI. It mounts the host NVIDIA toolkit
+// (version-matched to the node) — the glibc-based ateom image runs those toolkit
+// binaries directly — and applies the user-namespace + unmasked-/proc posture the
```
omeryahud: Is this correct? A left-over from a previous implementation?

eliranw: Yes, the host path was defaulting to the container path, so they’re split now (/usr/local/nvidia/toolkit vs /opt/nvidia-toolkit, which has to sit outside the device plugin’s read-only tree), and I fixed a stale comment beside it still claiming the function applies a userns + unmasked /proc.

cmd/ateom-gvisor/main.go
eliranw: instead of plumbing it I dropped the global entirely and had that call gpuPresent() directly, kept the check

dims:
Nice work. The pause container getting --nvproxy at create, the cgroup device allow rules, and the NVIDIA_VISIBLE_DEVICES strip are all subtle and correct. 

One question before the details: what happens to the golden snapshot when the workload holds a live CUDA context at suspend time? Doesn't gVisor refuses to save a sandbox with open GPU state? can we cross check if the test plan accounts for this?

cmd/ateom-gvisor/gpu.go
```diff
@@ -0,0 +1,401 @@
...
+
+	// Apply the CDI createContainer hooks except update-ldcache: running ldconfig
+	// needs a private /proc mount, which the pod's default masked /proc rejects. The
+	// SONAME symlinks it would create are staged below instead. The remaining hooks
+	// (create-symlinks, enable-cuda-compat) run from the mounted host toolkit.
+	if spec.Hooks == nil {
+		spec.Hooks = &specs.Hooks{}
+	}
+	for _, h := range edits.Hooks {
+		if h.HookName != "createContainer" || len(h.Args) < 2 || h.Args[1] == "update-ldcache" {
```
dims: Should this be an allowlist and not a deny list? Newer toolkits could emit more hooks

docs/api-guide.md
```diff
@@ -53,31 +59,55 @@ metadata:
...
+through.
+
+Every container in the actor gets the GPU. An actor's containers share one sandbox
+and the worker's whole device set, and `ActorTemplate` has no per-container resource
+fields, so GPUs are shared at the actor level rather than assigned to one container,
+the same as cpu and memory. This differs from a Kubernetes Pod, where the GPU goes
+only to the container that requests it. If per-container resource limits are added
+later, GPU assignment should follow them.
+
+**Requirements**
```
dims, line 102: Do we need to say anything about `LD_LIBRARY_PATH`?

cmd/ateom-gvisor/gpu.go
```diff
@@ -0,0 +1,401 @@
...
+	return int64(unix.Major(rdev)), int64(unix.Minor(rdev)), nil
+}
+
+// injectGPUIntoBundle merges the CDI spec generated in cdiSpecDir into the actor's OCI
+// config.json in bundleDir: device nodes (major/minor resolved from the host), the
+// driver-library mounts, and env. It does NOT run the CDI hooks; instead it stages the
+// SONAME symlinks (libcuda.so.1 -> libcuda.so.580.x) into the actor rootfs, which is
+// what lets the GPU worker keep the plain unprivileged posture (no user namespace, no
+// unmasked /proc). The CDI spec is plain JSON, so no CDI library is needed.
+func injectGPUIntoBundle(ctx context.Context, bundleDir, cdiSpecDir string) error {
```
dims, line 243: Does atelet always rewrite the bundle before each Run/Restore attempt? If not, a second call here doubles every device, mount, env, and hook. A guard that skips injection when an nvidia device is already in spec.Linux.Devices would make it safe either way.

cmd/ateom-gvisor/runsc.go
```diff
@@ -131,9 +146,10 @@ func (r *runsc) cmdStart(ctx context.Context, out io.Writer, containerName strin
 		// "-strace",
 		"-allow-connected-on-save",
 		"-root", ateompath.RunSCStateDir(r.actorUID),
-		"start",
-		containerName, // Name of the container
-	)
+	}
+	startArgs = append(startArgs, nvproxyGlobalArgs()...)
```
dims: Is --nvproxy needed on start? do we want to be consistent where all we add `--nvproxy`?

cmd/ateom-gvisor/gpu.go
```diff
@@ -0,0 +1,401 @@
...
+// gpuDeviceGlob matches the per-GPU device nodes. The device plugin can assign any
+// indices (a worker sharing a multi-GPU node may get /dev/nvidia2,3 with no
+// /dev/nvidia0), so detection must not assume index 0. The [0-9] excludes the control
+// nodes (/dev/nvidiactl, /dev/nvidia-uvm). A var so tests can point it at a fixture.
+var gpuDeviceGlob = "/dev/nvidia[0-9]*"
+
+const (
+	// driverLibDir is where the CDI/GKE device plugin injects the driver user-mode
+	// libraries into a GPU pod; nvidia-ctk needs it on its library search path.
+	driverLibDir = "/usr/local/nvidia/lib64"
```
dims: do we need this to be overridable?

cmd/ateom-gvisor/gpu.go
```diff
@@ -0,0 +1,401 @@
...
+	// driverBinDir holds the driver binaries (nvidia-smi, ...). nvidia-ctk discovers
+	// them via PATH, so we prepend this when generating the CDI spec — otherwise the
+	// spec carries only libraries and the actor has no nvidia-smi.
+	driverBinDir = "/usr/local/nvidia/bin"
```
dims: same

cmd/atecontroller/internal/controllers/workerpool_apply.go
```diff
@@ -206,6 +210,81 @@ func maybeApplyMicroVMPodShape(
...
+	return def
+}
+
+// maybeApplyGPUPodShape shapes a gVisor worker pod that requests a GPU so ateom
+// can inject the GPU into actors via CDI. It mounts the host NVIDIA toolkit
+// (version-matched to the node) read-only, for the glibc-based ateom image to run
+// directly. The pod keeps the same security posture as any other gVisor worker.
+// No-op for non-GPU pools and non-gVisor classes; an empty class defaults to
+// gVisor (WorkerPoolSpec kubebuilder default).
+func maybeApplyGPUPodShape(
```
dims, line 350: Hmm, A microvm pool that requests a GPU silently gets none, should this fail harder? earlier?

cmd/ateom-gvisor/gpu.go
eliranw: Sounds very reasonable, so I switched to an allowlist with logging of skipped hooks so we see them on newer toolkits as/when they come.

docs/api-guide.md
eliranw: Yes, added.
For context, CDI doesn’t emit LD_LIBRARY_PATH at all, the stock mechanism is the `update-ldcache` hook, which we skip for the unprivileged posture, so nothing puts the driver dir on the loader path. Every image I’d tested was an `nvcr.io/nvidia/*` one that sets it itself; override it and CUDA sees 0 devices even though the GPU is fully injected. Now prepending the driver library directory, with a unit test and a docs note.

cmd/ateom-gvisor/gpu.go
eliranw: It does, but I added a guard so we don't depend on the atelet, plus a regression test, and confirmed the test fails without it.

cmd/ateom-gvisor/runsc.go
eliranw: I thought so, but I tested removing it and it worked, kept it on create and restore, which actually boot the sentry process. `start` acts on a sandbox that already exists, so the flag has no effect there.

cmd/ateom-gvisor/gpu.go
eliranw: Yes, both of the variables you commented on now derive from one `ATE_NVIDIA_DRIVER_ROOT` (default `/usr/local/nvidia`).

cmd/ateom-gvisor/gpu.go
eliranw: Yes, both of the variables you commented on now derive from one `ATE_NVIDIA_DRIVER_ROOT` (default `/usr/local/nvidia`).

cmd/atecontroller/internal/controllers/workerpool_apply.go
eliranw: So to mitigate microVM pools with GPUs until/when actors there get support, I added a CEL rule on WorkerPoolSpec + tests - https://github.com/agent-substrate/substrate/pull/502/changes#diff-1ede08f8c093df5328ddbd446e4621241be469d4970f73b824426b601b35bab3R54

eliranw:
> One question before the details: what happens to the golden snapshot when the workload holds a live CUDA context at suspend time? Doesn't gVisor refuses to save a sandbox with open GPU state? can we cross check if the test plan accounts for this?

@dims Good catch as the test plan didn’t cover it, and with a live context the checkpoint fails.
After this is merged I’d open a separate issue to track GPU suspend/resume using `cuda-checkpoint`, linked to #450 since utilizing that likely needs PreSuspend/PostResume lifecycle hooks. For now it’s documented as a known limitation: a GPU actor can only be suspended with no CUDA context open, otherwise the checkpoint fails and the sandbox is terminated. Do you think that’s enough for now?

dims:
> Do you think that’s enough for now?

@eliranw will leave this to the maintainers of this repo! :)

cmd/ateom-gvisor/gpu.go
```diff
@@ -0,0 +1,491 @@
...
+		soname, err := elfSoname(m.Source)
+		if err != nil {
+			slog.WarnContext(ctx, "Skipping SONAME symlink for driver library",
+				slog.String("library", m.Source), slog.Any("err", err))
+			continue
+		}
+		if soname == "" || soname == base {
+			continue
+		}
+		linkDir := filepath.Join(rootfs, filepath.Dir(m.Destination))
```
BenTheElder: 🤖 **blocking** 🔴 – These writes are not confined to the rootfs, and every path component under it comes from the actor image. `maybeInjectGPU` runs immediately after `imagecache.SetupBundleRootfs`, so `rootfs` is the mounted overlay with image content already visible. An image that ships the driver-mount's parent — `/usr/lib/x86_64-linux-gnu` for the usual CDI destinations — as a symlink to an absolute path or a `../` chain makes the kernel resolve it in ateom's mount namespace, outside the actor.

`os.Remove(link)` on line 466 then deletes whatever sits at the escaped path, and `os.Symlink` plants a root-owned symlink there. ateom's namespace has the `ateompath.BasePath` hostPath mounted, so that reaches the shared image cache and other actors' bundles — an untrusted image gets arbitrary delete-as-root across the worker.

The codebase already defends this exact case one call earlier: `createExtraDirs` in `internal/imagecache/spec.go` opens the rootfs with `os.OpenRoot` precisely so "a symlink planted by the image cannot redirect the write outside the rootfs". Same treatment here — `root.MkdirAll` / `root.Remove` / `root.Symlink` — closes it. Worth rejecting a `soname` that is not a bare filename too, since it is read out of a file and joined into a path unchecked.

BenTheElder (commented):
High level: I was talking about devices with @juli4n earlier, I think we also need to select devices in the actortemplate so it's decoupled from the worker a bit, and so we can control which containers in an actor receive the devices.

There are some actual CI failures, but the zizmor scan just needs syncing with `main`, I already fixed that one.

BenTheElder:
> For now it’s documented as a known limitation: a GPU actor can only be suspended with no CUDA context open, otherwise the checkpoint fails and the sandbox is terminated. Do you think that’s enough for now?

I think that's OK for now, we should probably plan for how we might fix it? It seems like this will be hard to avoid in practice and make the features less useful?

cmd/atecontroller/internal/controllers/workerpool_apply.go
BenTheElder: We are otherwise generally doing uVM first now for features, I haven't had a chance to dig into device passthrough though as from my end there are issues that are more apparently pressing. I suspect we have to do a little dance with vfio and passing through the libraries.

cmd/ateom-gvisor/gpu.go
eliranw: Fixed, os.OpenRoot now confines the writes like createExtraDirs, the SONAME must be a bare filename, and a regression test covers both escapes. Verified on a T4 that symlink staging still works.

eliranw:
@BenTheElder so the solution for that should be using `cuda-checkpoint`. The main thing there is we would need to stage that binary on nodes. I think adding that should be a separate PR + Issue that I can handle

cmd/atecontroller/internal/controllers/workerpool_apply.go
```diff
@@ -301,6 +305,92 @@ func maybeApplyMicroVMPodShape(
...
+			WithType(corev1.HostPathDirectory)))
+	// Only propagated when set, so a default deployment adds no env to worker pods.
+	if root := os.Getenv(nvidiaDriverRootEnv); root != "" {
+		containerAC.WithEnv(corev1ac.EnvVar().WithName(nvidiaDriverRootEnv).WithValue(root))
+	}
+}
+
+// templateRequestsGPU reports whether the pool template requests one or more
+// nvidia.com/gpu devices (limits or requests).
+func templateRequestsGPU(tmpl *atev1alpha1.WorkerPoolPodTemplate) bool {
```
BenTheElder, line 380: 🤖 **nit** 🟢 – Accepting the GPU in `requests` shapes a pod Kubernetes will not admit: extended resources have to appear in `limits`, so a pool that sets only a request gets the toolkit mount and env, and then the Deployment is rejected. The failure surfaces on the Deployment rather than on the WorkerPool the user wrote.

The new CEL rule already rejects `nvidia.com/gpu` on non-gVisor pools; requiring it in `limits` there too would catch this at the same place instead.

eliranw: Good catch, added another rule for `limits`

BenTheElder (dismissed):
sorry this will need a rebase, drop me a ping when that's ready

BenTheElder (approved):

klueska:
FYI: I noticed this PR has the `update-ldconfig` hook disabled in it's CDI spec generation. This can cause problems if the host OS and the sandbox image have different fs layouts (e.g. RedHat vs. Debian). I’ve opened the following PR to fix the root cause so this hook so this hook doesn't need to be disabled on gVisor anymore: https://github.com/NVIDIA/nvidia-container-toolkit/pull/2059.

eliranw:
Thanks @klueska, the masked `/proc` EPERM is exactly why I skipped the hook. Worked around it by staging the SONAME symlinks into the container rootfs and prepending a configurable driver lib dir to `LD_LIBRARY_PATH`, but I would much rather have the fix upstream in the hook itself.

PR #504, opened 2026-07-23
imagecache: fall back to legacy mount on pre-6.5 kernels

### LINKED ISSUE #500: imagecache: probe for overlayfs lowerdir+ support and fall back to legacy mount(2) on pre-6.5 kernels
Follow-up to #467 review feedback (https://github.com/agent-substrate/substrate/pull/467#discussion_r3634283963).

`SetupBundleRootfs` mounts actor rootfs overlays via the new mount API (`fsopen` + one `fsconfig` `lowerdir+` append per layer), which lifts the mount(2) single-page option cap (~34 digest-derived layer paths) but requires Linux ≥ 6.5. All current GKE COS channels qualify (Stable: COS 121 LTS, kernel 6.6), but GKE Ubuntu node images and developer hosts running kind on older kernels (Ubuntu 22.04 GA = 5.15) do not — and there, every actor mount fails with a per-actor `EINVAL`, which reads as flakiness rather than an environment problem.

Proposed fix:
- Probe `lowerdir+` support once at ateom startup (cheap `fsopen`/`fsconfig` dry run against a scratch dir).
- When absent, fall back to legacy `mount(2)` option-string assembly. This is graceful degradation: pre-6.5 kernels cannot mount >~34-layer chains by any mechanism, so the fallback only fails where the kernel itself is the limit.
- Log one clear line at startup ("kernel lacks overlayfs lowerdir+ (need >= 6.5); using legacy mount, images beyond ~34 layers will fail") and include the kernel hint in the legacy path's over-length preflight error.

~40 lines + tests (the legacy assembly existed before cf77497d/77026523 and can be resurrected from history).

mesutoezdil:
Fixes #500.

SetupBundleRootfs mounts overlay rootfs with the new mount API,
using fsconfig lowerdir+ per layer. That option only exists on
Linux 6.5+. On older kernels every actor mount fails with a bare
EINVAL, which looks like flakiness, not a kernel version problem.

Fix: mountOverlay tries lowerdir+ directly on the first layer. If
the kernel does not recognize it, that call fails immediately and
queues "Unknown parameter" on the fs context log; only that specific
diagnostic triggers the fallback to a plain mount(2) call built from
the same lowerdir list, so a bad lowerdir path (ENOENT) is reported
as-is instead of being misread as a kernel version problem. The
legacy path keeps the old ~34 layer cap and errors with a kernel
hint if the option string would go over the mount(2) page limit.

Tested with go test ./internal/imagecache/... on a real Linux
kernel (root, CAP_SYS_ADMIN, tmpfs upperdir), including a test that
forces the fallback with an unrecognized option name and one for
the option string builder.

### DIFF
--- a/internal/imagecache/bundle_linux.go
+++ b/internal/imagecache/bundle_linux.go
@@ -22,10 +22,12 @@ package imagecache
 import (
 	"errors"
 	"fmt"
+	"log/slog"
 	"os"
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 
 	"golang.org/x/sys/unix"
 
@@ -103,6 +105,14 @@ func SetupBundleRootfs(bundlePath string) error {
 	return nil
 }
 
+// lowerdirPlusOption is the fsconfig parameter name mountOverlay uses for
+// each lowerdir. It is a var, not a literal, only so tests can force it to
+// an unrecognized name and exercise the pre-6.5-kernel fallback below
+// without needing an actual old kernel.
+var lowerdirPlusOption = "lowerdir+"
+
+var warnLegacyMountOnce sync.Once
+
 // setupImageVolumes exposes each image volume's contents read-only at its
 // bundle-local mount point, for the OCI spec to bind into the container.
 func setupImageVolumes(bundlePath string, volumes []ImageVolumeOverlay) error {
@@ -145,10 +155,10 @@ func mountImageVolume(mountpoint string, layers []string) error {
 }
 
 // mountOverlay attaches an overlay of lowers (top-most first) with the given
-// upper/work dirs at mountpoint, using the new mount API rather than
-// mount(2): appending lowerdirs one fsconfig(2) call at a time sidesteps
-// mount(2)'s single-page option-string cap, which digest-derived layer paths
-// (~114 bytes each) would hit at roughly 34 layers.
+// upper/work dirs at mountpoint. It prefers the new mount API, appending
+// lowerdirs one fsconfig(2) call at a time to sidestep mount(2)'s
+// single-page option-string cap, which digest-derived layer paths (~114
+// bytes each) would hit at roughly 34 layers.
 //
 // An empty upper mounts the overlay without a writable layer, which overlayfs
 // makes read-only.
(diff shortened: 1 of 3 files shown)


internal/imagecache/bundle_linux.go
```diff
@@ -97,17 +99,35 @@ func SetupBundleRootfs(bundlePath string) error {
 	return nil
 }
 
+// lowerdirPlusSupported probes, once per process, whether the running
+// kernel understands overlayfs's incremental "lowerdir+" fsconfig option
+// (Linux >= 6.5). The probe opens a throwaway fs context and tries the
+// option against a directory that always exists; it never creates a
+// superblock or touches the filesystem otherwise.
+var lowerdirPlusSupported = sync.OnceValue(func() bool {
```
BenTheElder: generally init() and friends are an anti-pattern *especially* interacting with the host from them ..., do we even need this probe versus just handling the error and falling back?

mesutoezdil: thx for feedback, i removed the separate probe PTAL

internal/imagecache/bundle_linux.go
```diff
@@ -97,21 +99,40 @@ func SetupBundleRootfs(bundlePath string) error {
...
+// signal, rather than a separate startup probe: a bad lowerdir path would
+// only surface later, at FsconfigCreate, so this first call failing is an
+// unambiguous kernel-version signal.
 func mountOverlay(mountpoint string, lowers []string, upper, work string) error {
 	fsfd, err := unix.Fsopen("overlay", unix.FSOPEN_CLOEXEC)
 	if err != nil {
 		return fmt.Errorf("while opening overlay fs context: %w", err)
 	}
+
+	if err := unix.FsconfigSetString(fsfd, lowerdirPlusOption, lowers[0]); err != nil {
```
BenTheElder, line 188: 🤖 **should-fix** 🟡 – Every error from this first call is read as "kernel too old", including errors that have nothing to do with the kernel version. On 6.5+ overlayfs resolves the path while parsing the parameter (`ovl_parse_layer` → `kern_path`), so a missing or unreadable layer directory fails right here with ENOENT. The operator then gets "kernel lacks overlayfs lowerdir+ (need >= 6.5)" followed by a second, unrelated failure from the legacy mount, and goes off chasing a kernel upgrade.

The kernel distinguishes the two cases for you: an unsupported parameter queues `Unknown parameter 'lowerdir+'` on the fs context log, which `fsContextLog` already reads — but this path closes the fd without draining it, discarding exactly the diagnostics #467 added to replace bare EINVALs. Gating the fallback on that message, or on the one-time probe against a known-good directory that the PR description describes, would keep real errors reporting themselves.

internal/imagecache/bundle_linux.go
```diff
@@ -97,21 +99,40 @@ func SetupBundleRootfs(bundlePath string) error {
...
-// it (Stable runs COS 121 LTS on kernel 6.6; Regular and Rapid run COS
-// 125/129 on 6.12).
+// Kernels older than 6.5 do not recognize the incremental "lowerdir+"
+// parameter at all, so the very first attempt to set it fails immediately
+// with an unrecognized-parameter error, before any path is looked at or any
+// superblock created. That failure is used directly as the fall-back
+// signal, rather than a separate startup probe: a bad lowerdir path would
+// only surface later, at FsconfigCreate, so this first call failing is an
+// unambiguous kernel-version signal.
 func mountOverlay(mountpoint string, lowers []string, upper, work string) error {
```
BenTheElder, line 178: 🤖 **nit** 🟢 – `lowers[0]` below makes this panic on an empty slice, where the old `range` loop was safe. Only `SetupBundleRootfs` calls it and it returns early for zero layers, so nothing hits it today — but a one-line guard is cheap insurance for a privileged per-actor path.

BenTheElder (commented):
description is out of date.

agentic review pass (human reviewed before submission) comments inline

internal/imagecache/bundle_linux.go
mesutoezdil: fixed in 078aa59f

internal/imagecache/bundle_linux.go
mesutoezdil: fixed in 078aa59f

internal/imagecache/bundle_linux.go
mesutoezdil: @BenTheElder thx, done

BenTheElder:
Please prefer rebase to merge, and squash these, so the commit message will be useful.

we started using volatile, which may have bumped up the requirements, but its load bearing and I don't think we're targeting old kernels

cc @igooch

mesutoezdil:
@BenTheElder thx, squashed into one commit.

PR #506, opened 2026-07-23
microvm: bump runtime assets to kata 4.0.0 + virtiofsd v1.14.0

BenTheElder:
- We no longer need to compile virtiofsd when installing for amd64, we still need to for arm64 (no upstream release)
- Drop-in compatible, only updating the scripts that manage obtaining the binaries, and the versions / hashes in the config.

- [x] Tests pass

### DIFF
--- a/.github/workflows/pr-workflow.yaml
+++ b/.github/workflows/pr-workflow.yaml
@@ -20,7 +20,7 @@ on:
   schedule:
     # Weekly run on the default branch keeps the micro-VM asset cache warm (GitHub
     # evicts caches idle for 7 days). The push-to-main run populates the cache that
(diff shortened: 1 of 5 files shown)


EItanya (dismissed):

BenTheElder:
Local testing was arm64 ... virtiofsd download on amd64 needed fixing ... fixed and tested that.

EItanya:
> Local testing was arm64 ... virtiofsd download on amd64 needed fixing ... fixed and tested that.

Is it worth adding tests for both architectures in a follow-up to catch it in the future?

BenTheElder:
github actions only supports KVM on amd64 (and even that is best-effort, not really supported)

juli4n (approved):

PR #514, opened 2026-07-24
feat(otel): add actor lifecycle + scheduler duration metrics

krisztianfekete:
This PR is the second slice of the platform-metrics split (#433).

It adds two duration histograms emitted by ateapi:
- `ate.actor.lifecycle.operation.duration`: create/resume/suspend/pause/delete, labeled by operation, template, pool, sandbox class, and (on resume) snapshot kind
  - I'd like to use this for user-facing latency for e.g. showing when suspended actors can serve requests again. The existing `rpc.server.call.duration` metric covers this, but the meaningful dimensions are missing, so it's not really actionable. Extending its labels with extra, domain-specific labels is an OTel anti-pattern, hence the new metric. 
- `ate.scheduler.assignment.duration`: worker-assignment step, labeled by outcome (assigned / no_free_worker / error) and pool
  -  I'd like to use this to alert on no free worker situation, and have proper SLOs via various percentiles for assigning latencies. There's no RPC around this, so it's not something existing RPC metrics cover.

Tested e2e on a local kind cluste where: both metrics reach the otel-system collector with the expected labels (resume shows `snapshot_kind=golden`, scheduler shows `outcome=assigned`, no `error.type` on success).

- [x] Tests pass


cmd/ateapi/internal/controlapi/create_actor.go
```diff
@@ -29,17 +31,25 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
-func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequest) (*ateapipb.Actor, error) {
-	if err := validateCreateActorRequest(req); err != nil {
+func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequest) (created *ateapipb.Actor, err error) {
```
EItanya, line 34: why are we swapping everything to be named return args, that's not a very common pattern. If you're trying to capture `err` can you just do `var err error`. I don't feel super duper strongly about this, but I personally find it a bit harder to read.

cmd/ateapi/internal/controlapi/delete_actor.go
```diff
@@ -18,29 +18,43 @@ import (
...
+func (s *Service) DeleteActor(ctx context.Context, req *ateapipb.DeleteActorRequest) (deleted *ateapipb.Actor, err error) {
+	start := time.Now()
+	// The request only names the actor, so tmpl (for the metric's template dims)
+	// is filled from whichever record we manage to load.
+	var tmpl *ateapipb.Actor
+	defer func() {
+		s.instruments.recordLifecycleOp(ctx, ateattr.OperationDelete, start, err,
+			ateattr.TemplateNameKey.String(tmpl.GetActorTemplateName()),
+			ateattr.TemplateNamespaceKey.String(tmpl.GetActorTemplateNamespace()),
+		)
```
EItanya: i think the behavior of this might be confusing. If this method returns before `tmpl` is assigned it may create a record with empty name, namespace. I think it's better to just use the incoming values even if they don't exist, what do you think?

cmd/ateapi/internal/controlapi/metrics.go
```diff
@@ -84,3 +91,98 @@ func RegisterWorkerCount(meter metric.Meter, workers func() ([]*ateapipb.Worker,
 	}
 	return nil
 }
+
+// Instruments holds ateapi's actor-lifecycle and scheduler duration histograms.
+// A nil *Instruments is a valid no-op, so call sites need no guard. Worker-count
+// is registered separately (RegisterWorkerCount): a callback-driven observable,
+// not a synchronous instrument.
+type Instruments struct {
```
EItanya, line 100: Is there any reason this is public, seems like we could construct it locally rather than exporting it

cmd/ateapi/internal/controlapi/create_actor.go
krisztianfekete: It's intentional, because with a `var err error` the defer would capture the pre-mapping error and also misqualify `error.type`. Would you say a comment would make this clearer?

cmd/ateapi/internal/controlapi/delete_actor.go
krisztianfekete: Fair, although `DeleteActorRequest` is an `ObjectRef` so there's no template. I'll just drop template labels from delete entirely.

cmd/ateapi/internal/controlapi/metrics.go
krisztianfekete: It's exported because main.go builds it and passes it to `NewService` like e.g. `RegisterWorkerCount`, so this seemed like the consistent way.

cmd/ateapi/internal/controlapi/delete_actor.go
krisztianfekete: Dropped this one.

krisztianfekete:
cc. @git286, @juli4n for review. This has been up for a good while now, can you please review when you have a chance?

zoez7:
@baizhenyu @JeffLuoo

cmd/ateapi/internal/controlapi/workflow.go
```diff
@@ -161,18 +165,27 @@ func NewActorWorkflow(
...
 
-	ctx, lock, err := w.acquireActorLock(ctx, actorRef)
+	// Recorded before the lock so lock contention still counts as an attempt; the
+	// incoming ctx stays valid where acquireActorLock returns nil on failure.
+	defer func() {
+		w.instruments.recordLifecycleOp(ctx, ateattr.OperationResume, start, err,
+			lifecycleOpAttrs(state.Actor, state.ActorTemplate, state.SnapshotKind)...)
+	}()
+
+	lockCtx, lock, err := w.acquireActorLock(ctx, actorRef)
```
JeffLuoo: ResumeActor is called on every event, and ActorResumer has no local "already running" cache — it always issues the ResumeActor RPC. That's exactly why `ResumeState.WasRunning` and the Resumed response field exist (ResumeOutcomeNone = "actor already running" in the router's own docs table).

This defer records unconditionally, so the sample rate of `ate.actor.lifecycle.operation.duration{operation="resume"}` equals **router QPS**, and the population is dominated by sub-millisecond no-op resumes that fast-forward every step. The p50/p99 you'd read off this is the warm-path no-op latency, not "when a suspended actor can serve requests again" — which is not the stated goal of the metric.

Consider adding a check here:
```
       if err == nil && state.WasRunning {
           return // already running: not a resume, and the router calls this per request
       }
```

cmd/ateapi/internal/controlapi/service.go
```diff
@@ -30,11 +30,12 @@ type Service struct {
 	actorTemplateLister listersv1alpha1.ActorTemplateLister
 	workerPoolLister    listersv1alpha1.WorkerPoolLister
 	actorWorkflow       *ActorWorkflow
+	instruments         *Instruments
```
JeffLuoo, line 33: The instrument is threaded into two different layers:

   - create / delete are recorded in the gRPC handler (create_actor.go, delete_actor.go), so they include request validation, lister lookups, and the final gRPC status.
   - resume / suspend / pause are recorded inside `ActorWorkflow` (workflow.go), so they exclude the handler's validation.

Can we handle all metrics in gRPC handlers to align?

JeffLuoo (changes requested):

cmd/ateapi/internal/controlapi/service.go
krisztianfekete: The split is intentional to some extent. The workflow owns the labels, so recording in handlers would drop those dims from exactly the failure datapoints they're for. But I agree about this not being fully consistent as validation failures currently count create/delete and not the workflow operations.

I'd rather fix that by moving the create/delete recording after validation, so all five operations measure a validated operation, and keep malformed requests visible in `rpc.server.call.duration`. Wdyt?

JeffLuoo: sg

cmd/ateapi/internal/controlapi/workflow.go
krisztianfekete: Added your guard in the latest push + a test pinning it.

JeffLuoo (dismissed):

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -519,6 +543,7 @@ func (s *CallAteletRestoreStep) Execute(ctx context.Context, input *ResumeInput,
 
 	if local := state.Actor.GetLocalSnapshotInfo(); local != nil {
 		slog.InfoContext(ctx, "Actor has snapshot; Restoring from snapshot")
+		state.SnapshotKind = ateattr.SnapshotKindLatest
```
baizhenyu: Should `snapshot_kind` separate local restores from durable ones?

This branch sets `latest` for a local in-node restore. `LoadActorForResumeStep` sets the same `latest` for a durable snapshot pulled from object storage. The two paths have very different latencies, so a p99 over `snapshot_kind=latest` mixes them together. Since resume latency is the main thing this metric is for, would you consider a fourth value such as `local`?

Separately, the kind is set in two places from different inputs. Could we set it only here? This is the only step that knows which restore path actually ran. A resume that fails before this step would then emit no kind label, which is how pool and sandbox class already behave.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -327,6 +348,7 @@ func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, stat
 		pickedWorker, err := s.scheduler.Schedule(ctx, constraints)
 		if err != nil {
 			if errors.Is(err, scheduling.ErrNoCapacity) {
+				outcome = ateattr.SchedulerOutcomeNoFreeWorker
```
baizhenyu, line 353: Could the `no_free_worker` outcome carry a capacity label?

Today this outcome has only `ate.scheduler.outcome`. The PR description says you want to alert on the no-free-worker case, but with only that label an alert can say that something ran out of capacity, not what ran out.

The pool is not available here, since `schedulingConstraints` matches on selectors rather than a pool name. But `constraints.SandboxClass` is available on all three outcome paths. Adding it would also make `assigned` and `no_free_worker` comparable, which is not possible today because they share no label. Would that work?

cmd/ateapi/internal/controlapi/metrics.go
```diff
@@ -85,3 +92,98 @@ func RegisterWorkerCount(meter metric.Meter, workers func() ([]*ateapipb.Worker,
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("create %s histogram: %w", lifecycleOpDurationMetric, err)
+	}
+
+	schedulerAssignmentDuration, err := meter.Float64Histogram(
+		schedulerAssignmentMetric,
+		metric.WithUnit("s"),
+		metric.WithDescription("Duration of the scheduler's worker-assignment step during actor resume."),
+		metric.WithExplicitBucketBoundaries(0.0005, 0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1),
```
baizhenyu: Is a 1 second ceiling high enough?

The comment above describes this as a sub-second in-memory pick. The timer in `AssignWorkerStep.Execute` covers the whole step: `workerCache.Workers()`, `Schedule`, `UpdateWorker`, and `UpdateActor`, plus a `GetActor` on the conflict path. That is two or three Redis round trips. If Redis latency rises, every sample lands in the `+Inf` bucket and the percentile stops being useful.

Could we extend the boundaries to 2.5 and 5? Changing histogram boundaries after the metric has data is harder than setting them now. The comment above would need a small update as well.

docs/observability.md
```diff
@@ -112,13 +112,18 @@ Agent Substrate emits foundational OpenTelemetry system and server metrics to mo
...
 | `atelet.snapshot.size` | atelet | histogram | uncompressed size in bytes of each gVisor snapshot image written during checkpoint (labels `kind`, `actor_template_namespace`, `actor_template_name`) |
+| `ate.workerpool.workers` | ateapi | up/down counter | live worker count per pool, split by state (`idle`/`assigned`) and sandbox class to provide fleet capacity and saturation at a glance |
+| `ate.actor.lifecycle.operation.duration` | ateapi | histogram | how long each actor operation (create/resume/suspend/pause/delete) takes and whether it failed (`error.type` present = failure, absent = success); labeled by operation, template, pool, sandbox class, and snapshot kind on resume; already-running resume no-ops are not recorded so the histogram tracks actual activations, not router traffic |
+| `ate.scheduler.assignment.duration` | ateapi | histogram | time it takes for an actor to be assigned to a worker, with the outcome (`assigned` / `no_free_worker` / `error`) to catch scheduling latency and capacity starvation problems |
```
baizhenyu: Could this row state whether the duration is per attempt or end to end?

`start` is declared inside `AssignWorkerStep.Execute`, and `runStep` re-runs `Execute` on `ErrVersionConflict`. A retried assignment therefore records only the final attempt, not the total time. That is a reasonable choice, but the current wording reads as end to end. Adding the words 'per attempt' here would make it unambiguous.

cmd/ateapi/internal/controlapi/metrics.go
```diff
@@ -85,3 +92,98 @@ func RegisterWorkerCount(meter metric.Meter, workers func() ([]*ateapipb.Worker,
...
+// dimensions (template, pool, class, snapshot kind).
+func (i *Instruments) recordLifecycleOp(ctx context.Context, op string, start time.Time, err error, extraAttrs ...attribute.KeyValue) {
+	if i == nil || i.lifecycleOpDuration == nil {
+		return
+	}
+	attrs := make([]attribute.KeyValue, 0, len(extraAttrs)+2)
+	attrs = append(attrs, ateattr.ActorOperationNameKey.String(op))
+	attrs = append(attrs, extraAttrs...)
+	if err != nil {
+		attrs = append(attrs, ateattr.ErrorTypeKey.String(status.Code(err).String()))
```
baizhenyu, line 147: Could `error.type` use the same format as `rpc.server.call.duration`?

`status.Code(err).String()` produces the string `Internal`. `rpc.server.call.duration` reports `rpc.response.status_code` as a number. The same failure appears in two formats, so no query can join the two metrics.

Would the numeric code work here instead? Lowercase names would also work if you prefer strings. The same change would apply to `recordSchedulerAssignment` below.

baizhenyu (commented):
Thanks for this. The `WasRunning` skip in the latest commit is a good fix, and reusing the existing signal means the metric and the `resumed` return value cannot disagree.

I have left some questions inline about the metric definitions. I would like to settle these before we build dashboards on top of them.

docs/observability.md
```diff
@@ -112,13 +112,18 @@ Agent Substrate emits foundational OpenTelemetry system and server metrics to mo
...
 
 The table lists the OpenTelemetry instrument names. How a name appears in a query depends on the backend (Cloud Monitoring (GMP) / Kind collector).
 
 For `atenet.router.route.duration`:
 * `ate.router.outcome` categorizes the route attempt result: `ok`, `cancelled`, `timeout`, `no_capacity`, `lock_conflict`, `not_found`, `unavailable`, `rate_limited`, or `resume_error`.
 * `ate.router.resume` indicates the singleflight execution state of actor resumption: `none` (actor already running), `triggered` (initiated cold activation), or `joined` (parked on in-flight activation).
 
+The `ate.*` control-plane metrics all have the same bounded labels (operation, outcome, pool, sandbox class, snapshot kind) so they aggregate cleanly, and they deliberately keep high-cardinality actor identity (name/uid/atespace) outside of metrics. These live on logs and traces instead.
```
baizhenyu: Two corrections to this sentence.

First, the labels are not all bounded. `ate.template.name` and `ate.template.namespace` are created by users, so they are the largest source of cardinality on these metrics.

Second, the label sets are not the same across operations: create has template only, delete has none, and resume has all five. Both are reasonable design choices, but this sentence is what a dashboard author will rely on, so it should describe the actual behavior.

On cardinality: `ate.sandbox.class` is derived from the template, and the template is already a label, so carrying both multiplies the series count without adding information. At roughly 50 templates and 5 pools that is about 850k datapoints across the 15 buckets. If the goal is to aggregate by class without knowing template names, keeping it makes sense. Could you record that reason here, or drop the label?

cmd/ateapi/internal/controlapi/delete_actor.go
```diff
@@ -16,20 +16,27 @@ package controlapi
...
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
-func (s *Service) DeleteActor(ctx context.Context, req *ateapipb.DeleteActorRequest) (*ateapipb.Actor, error) {
+func (s *Service) DeleteActor(ctx context.Context, req *ateapipb.DeleteActorRequest) (deleted *ateapipb.Actor, err error) {
 	if errs := validateDeleteActorRequest(req); len(errs) > 0 {
 		return nil, toGRPCStatusError(errs)
 	}
+	start := time.Now()
+	// No template dims: a delete request names only the actor.
```
baizhenyu: Could delete keep a template label?

I see this was added and then reverted. The revert was correct, since `tmpl` is nil on the failure paths and would produce empty-string series.

Could you use the same conditional append that `lifecycleOpAttrs` uses, and add the template only when `deleted` is not nil? At the moment delete is the only operation with no template label, so grouping this metric by template silently drops deletes.

cmd/ateapi/internal/controlapi/metrics.go
krisztianfekete: I'd keep the gRPC code strings. OTel's `error.type` semconvs recommends meaningful, low-cardinality string values, and numeric codes make alerts hard to read and reason about.

docs/observability.md
krisztianfekete: Fixed the sentence.

Regarding dropping `sandbox.class`: 
 - it functionally depend on template (each template has exactly one class), so it multiplies nothing.
 - it exists so dashboards can sum by (class) without enumerating templates. I've added this rationale in the doc.

JeffLuoo (approved):

cmd/ateapi/internal/controlapi/workflow_resume.go
baizhenyu: Thanks, this is better than what I suggested. Setting the kind only in this step and adding `local` makes the resume histogram readable per restore path. The golden/latest re-derivation in the durable branch matches `LoadActorForResumeStep`'s ordering, since that step checks the actor's own snapshot first, so a non-nil `LatestSnapshot` always means the location came from it.

One non-blocking note: the kind is now also absent when this step is skipped entirely, not just on failures, so a successful crash-recovery resume that finds the restore already done would record without a `snapshot_kind`. That is consistent with the omit-until-known rule, so I mention it only so it is a known behavior rather than a surprise later.

Resolving.

cmd/ateapi/internal/controlapi/workflow_resume.go
baizhenyu: Thanks, this does exactly what I was after. `assigned` and `no_free_worker` now share `ate.sandbox.class`, so a per-class success ratio is queryable and the starvation alert can name what ran out. Resolving.

cmd/ateapi/internal/controlapi/metrics.go
baizhenyu: Thanks, the 5s ceiling and the corrected comment both look right. Resolving.

docs/observability.md
baizhenyu: Thanks, the per-attempt wording plus the retry note makes it unambiguous. Resolving.

cmd/ateapi/internal/controlapi/metrics.go
baizhenyu: You are right, and I withdraw this. Semconv does ask for meaningful low-cardinality strings, and my joinability argument does not hold anyway: `rpc.server.call.duration` reports `rpc.response.status_code`, not `error.type`, so the two were never joinable on a shared label regardless of format. Keeping the gRPC code strings is the better choice. Resolving.

docs/observability.md
baizhenyu: The rewritten sentence covers what I was asking for, thanks.

On `ate.sandbox.class` you are right and my analysis was wrong. Class is functionally dependent on the template, so next to the template labels it adds no new series at all. I had treated it as an independent dimension and multiplied by it, which is where the 850k figure came from. That number should be ignored. The aggregate-by-class argument is a good reason to keep the label on its own merits. Resolving.

cmd/ateapi/internal/controlapi/delete_actor.go
baizhenyu: Thanks, this is the pattern I had in mind. Resolving.

baizhenyu (approved):

PR #517, opened 2026-07-24
Handle graceful worker termination

sfunkenhauser:
Part of [#23](https://github.com/agent-substrate/substrate/issues/23)

The overall plan is:
1) Trap SIGTERM in the ateom, and forward SIGTERM to all application containers.
2) Once all application containers exit, ateom will exit itself.
3) The workerpool syncer will mark workers as 'DRAINING' when a pod deletion is triggered (pod.DeletionTimestamp) to prevent actors from being scheduled. On the pod Deleted event, the worker and actor (if it was assigned) will be cleaned up.

This PR implements 3), there are 2 more PRs coming which will implement 1) and 2) for each ateom (gvisor and uVM).


sfunkenhauser:
@juli4n @zoez7 PTAL

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -106,13 +113,13 @@ func (s *WorkerPoolSyncer) syncWorkerToStore(ctx context.Context, pod *corev1.Po
...
-		}
-		err := s.persistence.DeleteWorker(ctx, pod.Namespace, pod.Labels[workerPodLabel], pod.Name)
-		if err != nil {
-			slog.ErrorContext(ctx, "Failed to delete worker from store during update event (deleting)", slog.Any("err", err))
+		// The pod has entered Terminating: mark the worker DRAINING so the
+		// scheduler stops routing new actors to it. We deliberately do NOT touch
+		// the bound actor here — inside the pod ateom has received SIGTERM and is
+		// gracefully shutting the actor down. Actor cleanup happens on the Pod
+		// Deleted event.
+		if err := s.markWorkerDraining(ctx, pod.Namespace, pod.Labels[workerPodLabel], pod.Name); err != nil {
```
zoez7, line 121: We might fail to mark the worker as draining due to version mismatch, if it was selected by another actor "won" and marked it as assigned during resuming? 
Maybe we need to retry this, and add another step in the ResumeActor workflow to confirm the worker is not in Draining state before returning "success"?

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -185,20 +193,90 @@ func isWorkerEligible(pod *corev1.Pod) bool {
...
+// reconcileOrphanedWorkers cleans up stored worker records whose pods no longer
+// exist. It runs once after the informer cache has synced, when the indexer is a
+// fresh, authoritative snapshot of the live worker pods, so a worker missing
+// from the indexer (or present under a different pod UID, i.e. name reuse) is
+// stale.
+func (s *WorkerPoolSyncer) reconcileOrphanedWorkers(ctx context.Context) {
+	var workers []*ateapipb.Worker
+	var pageToken string
+	for {
+		wPage, nextToken, err := s.persistence.ListWorkers(ctx, 1000, pageToken)
```
zoez7, line 238: This will needed to be sharded once we have multiple instances of ate-apiserver.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -182,7 +182,7 @@ func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, stat
 		if err != nil {
 			return fmt.Errorf("while checking worker eligibility: %w", err)
 		}
-		if eligible {
+		if eligible && worker.GetState() != ateapipb.Worker_STATE_DRAINING {
```
zoez7: should this be `worker.GetState() == Worker_STATE_ACTIVE` instead, in case we add other worker states?

cmd/ateapi/internal/controlapi/syncer.go
sfunkenhauser: This will be retried on subsequent informer updates. We could add retry logic to every update call, but I think that will get complicated. I'd rather rely on the informer (and in the future a workqueue) to retry failed operations.

Marking a worker as draining will always be best effort, so the ateom needs to handle the case where a resume/suspend operation lands on a pod that's getting deleted.

ResumeActor already checks in a few places to make sure the worker isn't draining (AssignWorker), but there is always going to be a race between reading worker state and performing some other action (AteletRestore for example).

If we end up calling AteletRestore on a pod that's being deleted, the ateom will return an error and fail the resume workflow.

If the AteletRestore succeeds before the pod is deleted, we should consider the resume successful.

cmd/ateapi/internal/controlapi/syncer.go
zoez7: I'm not sure if the sharding needs to be addressed now, but looks like @juli4n already has https://github.com/agent-substrate/substrate/pull/538 open.

sfunkenhauser: This is just an optimization right? Ultimately I think we should move the syncer out of the ate-apiserver. Maybe we can wait for that decision before optimizing this.

cmd/ateapi/internal/controlapi/syncer.go
zoez7: > If we end up calling AteletRestore on a pod that's being deleted, the ateom will return an error and fail the resume workflow.

I see, so we'll need to (1) update ateom to reject Restore requests after receiving SIGTERM, and (2) update ateapi to return a retriable error, so client retries the resume operation. Did I understand that correctly?

cmd/ateapi/internal/controlapi/workflow_resume.go
sfunkenhauser: Yes, good catch!

cmd/ateapi/internal/controlapi/syncer.go
sfunkenhauser: (1) That's correct.
(2) Ultimately yes, but this is going to take a bit more work. Right now if you resume an actor with status Actor_STATUS_RESUMING, and the worker is in a bad state (NOT_FOUND and I'm adding DRAINING to this as well) the actor will be crashed. I just don't know if it's safe to try and resume the actor on another worker. We need to take a closer look at the different resume workflow states and figure it out. Right now I think it's safer to just crash in this case.

Concretely for this change, I would have the ateom reject the Restore request with a non-retriable error and the resume workflow would crash the actor. Let me know if you think this is being overly cautious.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -182,7 +191,7 @@ func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, stat
 		if err != nil {
 			return fmt.Errorf("while checking worker eligibility: %w", err)
 		}
-		if eligible {
+		if eligible && worker.GetState() == ateapipb.Worker_STATE_ACTIVE {
```
BenTheElder: 🤖 **should-fix** 🟡 – This requires `STATE_ACTIVE`, but `findFreeWorker` below only excludes `STATE_DRAINING`. The two disagree about `STATE_UNSPECIFIED`, and that is the state every pre-existing worker record has: `State` is set only on `CreateWorker`, and `syncWorkerToStore`'s update path never backfills it, so records written before this PR stay unspecified indefinitely. `ValidateWorker` tolerates that value explicitly "for backward compatibility", so it is an expected state, not a corrupt one.

After an upgrade, this loop then rejects a legacy worker that holds a valid claim for the actor, releases the claim, and re-schedules — and if the pool has no other free worker, the resume fails even though the claimed worker was healthy. Using the same predicate at both sites (`!= STATE_DRAINING`) would fix it, or backfill `State` in the update path so legacy records converge on `ACTIVE`.

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -221,16 +299,18 @@ func (s *WorkerPoolSyncer) releaseActorOnDeadWorker(ctx context.Context, namespa
...
 		return nil
 	}
-	if actor.Status != ateapipb.Actor_STATUS_CRASHED {
-		actor.Status = ateapipb.Actor_STATUS_SUSPENDED
+	// If the actor is suspended, it's already been released.
+	if actor.Status == ateapipb.Actor_STATUS_SUSPENDED {
+		return nil
 	}
+
+	actor.Status = ateapipb.Actor_STATUS_CRASHED
```
BenTheElder, line 308: 🤖 **question** 🟢 – This is the right end state once ateom suspends the actor on SIGTERM, but steps 1 and 2 are still to come. Until they land, a running actor is no longer marked `SUSPENDED` when its worker pod goes away, so every routine worker rollout leaves its actors reading `CRASHED`.

Nothing gates resume on status, so they stay resumable and this is a labeling change rather than data loss — and the old `SUSPENDED` arguably implied a clean snapshot that was never taken. Is that interim acceptable, or would this transition be better landed alongside step 2?

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -185,20 +193,90 @@ func isWorkerEligible(pod *corev1.Pod) bool {
...
+}
+
+// releaseActorOnDeadWorker resets the actor bound to a vanishing worker pod. An
+// actor that already reached STATUS_SUSPENDED (it saved its state cleanly during
+// graceful termination) is left untouched and remains resumable. An actor that
+// was still running when the pod disappeared is moved to STATUS_CRASHED and its
+// pod pointers are cleared.
 //
 // UpdateActor uses optimistic version checking. A concurrent SuspendActor
 // or ResumeActor wins; we drop this attempt silently.
```
BenTheElder: 🤖 **nit** 🟢 – Stale: the `ErrPersistenceRetry` tolerance is gone, so contention is no longer dropped silently — it propagates and stops `reconcileDeadWorker` from deleting the worker record. That is deliberate for real errors, but on contention the actor was already updated by whoever won, and the dead worker's record now lingers until the next startup reconcile.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -319,7 +328,7 @@ func (s *AssignWorkerStep) findFreeWorker(
 		if err != nil {
 			return nil, err
 		}
-		if !eligible {
+		if !eligible || worker.GetState() == ateapipb.Worker_STATE_DRAINING {
```
zoez7: This should be && worker.GetState() == ateapipb.Worker_STATE_ACTIVE as well? 

(It's not backwards compatible, but at this stage I think requiring devs to rebuild ate-system is still ok, soon that may not be the case.)

zoez7 (dismissed):
LGTM. Need to resolve conflicts before merging.

cmd/ateapi/internal/controlapi/syncer.go
zoez7: Re (2) I think this is a good starting point. Pick-another-worker-on-retry is an SLO improvement we can later make.

cmd/ateapi/internal/controlapi/syncer.go
sfunkenhauser: This was a concious change to mark an actor as CRASHED when it's worker pod goes away. Marking it as SUSPENDED is misleading as there is currently no mechanism to SUSPEND the actor on shutdown. Marking as CRASHED is the safer option and we can relax this constraint as we add support for self-suspend and lifecycle hooks.

cmd/ateapi/internal/controlapi/workflow_resume.go
sfunkenhauser: Done. Moved to scheduling.go.

cmd/ateapi/internal/controlapi/syncer.go
sfunkenhauser: Updated comment. This is a deliberate change, we don't know the state of the actor after the contentious update, and I don't think it's safe to just assume the actor is in a good state. We should fail this attempt, keep the worker record and retry.

Right now the only retry mechanism is on startup reconciliation, but we need to add a workqueue to retry failed informer events.

cmd/ateapi/internal/controlapi/workflow_resume.go
sfunkenhauser: scheduling.go has consolidated this logic so the same state is used for scheduling and finding free workers.

This is a breaking change, but that is currently acceptable.

BenTheElder:
[needs rebase again]

sfunkenhauser:
Merge conflicts fixed. PTAL.

BenTheElder (approved):

Issue #503, opened 2026-07-23
OpenTelemetry Application Telemetry Data Loss and Delivery Delays Across Actor Lifecycle Events


git286:
---

## 1. Issue Background & Context

User applications running as Agent Substrate actors frequently emit custom metrics and distributed trace spans using OpenTelemetry (OTel) push exporters (`otlpmetricgrpc`, `otlptracegrpc`, or `otlpmetrichttp`).

In typical production workloads:
- **OTel Push Exporter Interval:** e.g., 60 seconds (standard batch export timer).
- **Active Execution Duration Before Idle Suspension:** e.g., 30 seconds (active request handling window before an idle timeout or on-demand pause occurs).

When the push exporter interval (e.g., 60s) is longer than the active execution duration before suspension (e.g., 30s), telemetry events generated during the active phase may remain buffered in the OTel SDK's in-memory queues (`PeriodicReader` accumulators and `BatchSpanProcessor` ring buffers).

This issue summarizes the core challenges, empirical findings across Substrate snapshot modes, short-term mitigations for actor authors, and proposed long-term system-level solutions.

---

## 2. Identified Telemetry Issues Across Actor Lifecycle Events

Through architectural analysis and lifecycle benchmarking, five primary telemetry issues were identified:

### Issue 1: In-RAM Telemetry Loss in Data Snapshot Mode (`onPause: Data`) — **HIGH**
- **Root Cause:** When `snapshotsConfig.onPause` is set to `Data`, Substrate excludes process memory (RAM) from the snapshot and saves only attached `DurableDir` disk volumes. The container process is killed upon pause.
- **Impact:** **100% loss of pre-pause in-memory telemetry**. All metrics and trace spans generated during the 30s before pause that were awaiting the 60s push timer are permanently lost when the process is killed.

### Issue 2: Telemetry Abandonment for Un-Resumed Suspended Actors — **HIGH**
- **Root Cause:** In Full Snapshot Mode (`onPause: Full`), process RAM is saved to snapshot storage. However, if an actor is suspended and subsequently **deleted or garbage-collected** (`kubectl ate delete actor`) without ever being resumed, the buffered in-memory metrics remain stuck in the snapshot file.
- **Impact:** Un-pushed telemetry is never delivered to the OTel Collector.

### Issue 3: Delivery Delays & Wall-Clock Timer Expiration — **MEDIUM**
[...]

PR #522, opened 2026-07-24
docs(observability): add actor application telemetry best practices

git286:
The observability guide covers telemetry emitted by Substrate's own control plane, but says nothing to actor authors whose application code emits OpenTelemetry metrics and spans. Those signals sit in the OTel SDK's in-memory buffers (PeriodicReader accumulators, BatchSpanProcessor queues) and are subject to actor lifecycle events in ways a normal long-lived process never is.

Add a section documenting the actual behavior:

  * onPause: Full preserves the buffer; onPause: Data discards it, since process memory is excluded from the snapshot and the containers are deleted at the end of every checkpoint.
  * The workload receives no SIGTERM and no pre-suspend callback, so a flush can only happen where the actor itself initiates the suspend.
  * On resume, gVisor's CLOCK_MONOTONIC includes the suspended interval, so overdue export timers fire immediately.
  * Metric data points are stamped at collection time and so are attributed to the resume; span timestamps are recorded when the span ran and so go stale across a long suspension.
  * The OTLP connection does not survive a pod migration.

Along with the corresponding recommendations: flush before self-suspend, stay on onPause: Full, shorten the metric and span export timers separately, configure gRPC keepalives, and use a DurableDir volume under onPause: Data.

Known gaps are tracked in #450 (PreSuspend/PostResume hooks), #23 (graceful termination) and #503.

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -191,3 +191,97 @@ Dashboard definitions live in [`tools/setup-gcp/dashboards/`](../tools/setup-gcp
 ```sh
 go run ./tools/setup-gcp create dashboards   # also part of: bootstrap
 ```
+
+---
+
+## 6. Actor Application Telemetry (Best Practices)
+
+Everything above describes telemetry emitted by Substrate itself. This section is for **actor authors** whose application code emits its own OpenTelemetry metrics and spans.
+
+An actor is not an ordinary long-lived process. It is checkpointed, its containers are destroyed, and it is later restored — possibly onto a different worker pod, possibly much later. OpenTelemetry's push exporters buffer signals in memory and flush them on a timer (`PeriodicReader` for metrics, `BatchSpanProcessor` for spans), so whether that buffer survives a suspension depends on how the actor is configured and on whether the actor gets a chance to flush.
+
+### What happens to buffered telemetry on suspend
+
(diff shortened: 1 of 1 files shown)


EItanya (commented):
I think these notes will be important, but I think it's worth waiting until we have some specific direction on the self-suspend and graceful-suspend before putting this doc out there as it will be superseded by that

git286:
Thanks — I want to push back gently on the timing.

We discussed this with @sfunkenhauser and there's currently no plan for #450. So "wait for direction" isn't a short hold, it's indefinite — and in the meantime anyone running a workload on Substrate is silently losing telemetry with no documented way to find out why. The purpose of this section isn't to describe the eventual design; it's to tell people what they can do *today* (stay on `onPause: Full`, shorten both export timers, flush before self-suspend) so their telemetry works at all in the interim.

I'd also note that only part of this gets superseded. The mitigations do. But the behavioral notes don't — the `CLOCK_MONOTONIC` jump across restore, metric points being stamped at collection time while span timestamps go stale, and the dead OTLP connection after migration are gVisor and OTel SDK properties that hold no matter what we land in #450/#23.

I've pushed 8773431, which adds an explicit "interim guidance" callout at the top of the section naming #450 and #23, so nobody mistakes these for a stable contract. I'll own the follow-up PR once those have direction. Does that address the concern?

Issue #181, opened 2026-06-04
Define substrate HA story


juli4n:
This includes how we will run ate-apiserver in HA mode. This includes not just the grpc server but also whatever database we decide to use. How will ate-apiserver scale horizontally? Do we care about vertical scaling? How do we scale the underlying database? Are all replicas active? Is there any sharding mechanism? Any leader election? etc...

PR #523, opened 2026-07-24
Instrument ate-apiserver so that it can shutdown gracefully.

juli4n:
Make the server trap SIGTERM and initiate a graceful shutdown process. This involves reporting the Pod as not ready, and draining all in-flight gRPC requests so they can complete.

Part of #181

### DIFF
--- a/cmd/ateapi/main.go
+++ b/cmd/ateapi/main.go
@@ -24,7 +24,9 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/controlapi"
@@ -76,6 +78,9 @@ var (
 	podIdentityCACerts     = pflag.String("pod-identity-ca-certs", "", "The file that contains the pod-identity CA bundle, used both for verifying client certificates presented to the gRPC server and for verifying atelet serving certificates when dialing atelet. If empty, client-cert verification is disabled and atelet dials will fail.")
 	ateletClientCredBundle = pflag.String("atelet-client-cred-bundle", "", "Credential bundle presented as the client certificate when dialing atelet.")
 
+	drainDelay   = pflag.Duration("drain-delay", 13*time.Second, "How long to keep accepting new work after SIGTERM, before starting the gRPC drain.")
+	drainTimeout = pflag.Duration("drain-timeout", 15*time.Second, "Deadline for the graceful gRPC drain on shutdown. In-flight RPCs still running past it are forcefully cancelled.")
+
 	showVersion     = pflag.Bool("version", false, "Print version and exit.")
 	clientJWTCAFile = pflag.String("client-jwt-ca-cert", ateapiauth.DefaultServiceAccountCAFile, "CA cert file used to verify TLS when fetching the OIDC discovery document and JWKS for JWT authentication. Defaults to the in-cluster service account CA.")
 )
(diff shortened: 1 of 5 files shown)


cmd/ateapi/main.go
```diff
@@ -204,14 +211,39 @@ func main() {
...
+		Readiness:     readiness,
+		EnableHealthz: true,
 	})
 
+	go drainOnShutdown(shutdownCtx, mux, readiness)
+
 	if err := mux.Serve(lis); err != nil {
 		serverboot.Fatal(ctx, "Failed to serve", err)
 	}
+	slog.InfoContext(ctx, "Shutdown complete")
```
michaelasp, line 225: Nothing waits on this to complete right? Should the main function wait for the drainOnShutdown to complete?

e.g
```
drainFinished := drainOnShutdown(shutdownCtx, mux, readiness)

if err := mux.Serve(lis); err != nil {
	serverboot.Fatal(ctx, "Failed to serve", err)
}
<-drainFinished
```

HavenXia:
Thanks, this is what I am defining for the upgrade model - the old instances need to be gracefully drained before shutdown.

manifests/ate-install/ate-api-server.yaml
```diff
@@ -141,6 +146,15 @@ spec:
             port: 9090
           initialDelaySeconds: 5
           periodSeconds: 2
```
michaelasp, line 154: Won't this run 3 times before the pod is marked unhealthy in the apiserver? That means that it would take ~6s to mark the pod unhealthy. Won't we start rejecting traffic at 5s in the ate-apiserver? That would mean we still aren't graceful unless we tweak some limits. I'd just test this in a kind cluster and make sure that the graceful shutdown timeouts are enough.

manifests/ate-install/ate-api-server.yaml
```diff
@@ -91,6 +93,9 @@ spec:
           - --session-id-jwt-pool=/run/session-id-jwt-pool/pool.json
           - --session-id-ca-pool=/run/session-id-ca-pool/pool.json
           - --workerpool-ca-certs=/run/workerpool-ca-certs/trust-bundle.pem
+          # Graceful shutdown knobs. The sum must fit within terminationGracePeriodSeconds.
+          - --drain-delay=5s
```
BenTheElder: 🤖 **should-fix** 🟡 – The readiness signal can't land within this 5s window. The readinessProbe above sets `periodSeconds: 2` and leaves `failureThreshold` at its default of 3, so kubelet needs ~6s of consecutive 503s before it marks the pod NotReady — after the drain has already closed the listener. As configured, `MarkNotReady` contributes nothing and removal from Endpoints rests entirely on the deletionTimestamp path.

Setting `failureThreshold: 1` on the readinessProbe makes the probe react in ~2s, inside the drain delay.

cmd/ateapi/main.go
```diff
@@ -200,14 +211,39 @@ func main() {
...
+}
+
+func drainOnShutdown(ctx context.Context, srv *grpc.Server, readiness *serverboot.Readiness) {
```
BenTheElder: 🤖 **nit** 🟢 – `drainOnShutdown` reads the flag globals directly, so the sequence it orchestrates (mark unready → delay → graceful stop → force stop on timeout) can't be unit tested. Passing the two durations as parameters would make it testable, like `metricsMux` was split out for the same reason.

manifests/ate-install/ate-api-server.yaml
BenTheElder: hah, your comment hadn't loaded yet, AI review suggests ~ the same https://github.com/agent-substrate/substrate/pull/523#discussion_r3648390810

BenTheElder: From manual follow-up this is what jumps out the most.

cmd/ateapi/main.go
juli4n: Added. FWIW, I don't think this is a correctness thing as `.Serve` will only unblock once we are done with the shutdown, but there is one subtlety, the last log line in `drainOnShutdown`. 

Not sure if it's worth the complexity tbh, but I added it as I like the log line discriminating a interrupted drain from a drain that completed gracefully.

manifests/ate-install/ate-api-server.yaml
juli4n: Adjusted.

manifests/ate-install/ate-api-server.yaml
juli4n: Adjusted. It was off by a few seconds.

BenTheElder (approved):

PR #526, opened 2026-07-24
Fix orphaned golden actors by using deterministic actor IDs in `ActorTemplateReconciler`

### LINKED ISSUE #525: `ActorTemplateReconciler` orphans golden actors in `ateapi` on reconciliation retries
## Expected Behavior
Golden actor creation should be idempotent and deterministic per `ActorTemplate` resource lifecycle so retries recover safely without leaving orphaned actors in `ateapi`.

## Actual Behavior
In `ActorTemplateReconciler.Reconcile` (`PhaseInitial`), golden actor IDs are generated non-deterministically using `uuid.NewString()`. 
If `CreateActor` succeeds but the subsequent Kubernetes status update fails, `Reconcile` returns an error. On retry, `at.Status.Phase` is still `PhaseInitial`. The reconciler generates a **new** random UUID and calls `CreateActor` again, leaving the actor created in the prior attempt orphaned in `ateapi`.

troychiu:
This PR fixes an issue where reconciliation retries in `PhaseInitial` left orphaned actors in `ateapi` when `CreateActor` succeeded but status update failed.

Fixes #525 

- [X] Tests pass

### DIFF
--- a/cmd/atecontroller/internal/controllers/actortemplate_controller.go
+++ b/cmd/atecontroller/internal/controllers/actortemplate_controller.go
@@ -22,7 +22,6 @@ import (
 	"github.com/agent-substrate/substrate/internal/resources"
 	atev1alpha1 "github.com/agent-substrate/substrate/pkg/api/v1alpha1"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
-	"github.com/google/uuid"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	k8errors "k8s.io/apimachinery/pkg/api/errors"
@@ -80,7 +79,7 @@ func (r *ActorTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	switch at.Status.Phase {
 	case atev1alpha1.PhaseInitial:
-		actorID := uuid.NewString()
+		actorName := string(at.UID)
 
 		// Golden actors live in the reserved ate-golden system atespace.
 		_, err := r.AteClient.CreateAtespace(ctx, &ateapipb.CreateAtespaceRequest{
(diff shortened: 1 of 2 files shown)


cmd/atecontroller/internal/controllers/actortemplate_controller.go
```diff
@@ -80,7 +79,7 @@ func (r *ActorTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	switch at.Status.Phase {
 	case atev1alpha1.PhaseInitial:
-		actorID := uuid.NewString()
+		actorName := string(at.UID)
```
troychiu, line 82: Kubernetes UIDs (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids) should be compatible with resource names, which are enforced by standard RFC-1123 DNS label rules.

EItanya (approved):
Great catch!

zoez7:
This changes the behaviour of Reconcile - if a previous attempt created a actor with `at.UID` and failed to produce a golden snapshot, this will reuse the same actor and may never produce a usable golden snapshot.

Also, now updates to that same actor template doesn't create a new snapshot. I think this bug is part of the complexity mentioned in https://github.com/agent-substrate/substrate/issues/536.

EItanya:
> This changes the behaviour of Reconcile - if a previous attempt created a actor with `at.UID` and failed to produce a golden snapshot, this will reuse the same actor and may never produce a usable golden snapshot.

Functionally I think that's correct, otherwise like mentioned above we may have a bunch of orphaned actors. Do you think a more correct way would be to leave them orphaned and have a secondary GC process pick them up?

> Also, now updates to that same actor template doesn't create a new snapshot. I think this bug is part of the complexity mentioned in #536.

I thought the `ActorTemplate` was immutable today anyway?

troychiu:
> This changes the behaviour of Reconcile - if a previous attempt created a actor with at.UID and failed to produce a golden snapshot, this will reuse the same actor and may never produce a usable golden snapshot.

Even before this PR, once `at.Status.GoldenActorID` was saved, all subsequent reconciliation phases already reused that single actor ID upon retries. The orphan leak happened specifically when `CreateActor` succeeded in ateapi but the kubernetes status update failed.

> I thought the ActorTemplate was immutable today anyway?

Yes the `ActorTemplate` spec is immutable https://github.com/agent-substrate/substrate/blob/1afaaf5725638b328d8fdd6023fe13c65da79520/pkg/api/v1alpha1/actortemplate_types.go#L387-L388

PR #527, opened 2026-07-24
Handle partial volume creation failures

msau42:
Part of #232

1. During CreateActor, the volumes to be created get instantiated in the Actor API in PENDING status. The Actor is still instantiated as SUSPENDED.

However, the CreateVolume operation is moved to the beginning of ResumeActor. This is because CreateActor is intentionally designed to not be idempotent. The downside is that the user won't know dependent resource creation failed until the first resume.

2. Changes volume id to use actor uid instead of atespace+actorname and adds "substrate" prefix

3. Adds a new DELETING status that gets persisted before we start to delete volumes.

This shifts the existing DB DELETE precondition checks to the new DELETING status. And now you can only delete the actor in DB from DELETING status.

4. Refactor DeleteActor to use a workflow

In the future, we will also add a workflow that will do mandatory node cleanup of volumes before starting to delete the volumes.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/delete_actor.go
+++ b/cmd/ateapi/internal/controlapi/delete_actor.go
@@ -19,8 +19,6 @@ import (
 
 	"github.com/agent-substrate/substrate/internal/resources"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
--- a/cmd/ateapi/internal/controlapi/delete_actor.go
+++ b/cmd/ateapi/internal/controlapi/delete_actor.go
@@ -42,8 +42,29 @@ func (s *Service) DeleteActor(ctx context.Context, req *ateapipb.DeleteActorRequ
 		return nil, fmt.Errorf("while fetching actor: %w", err)
 	}
 
+	if actor.GetStatus() != ateapipb.Actor_STATUS_SUSPENDED &&
+		actor.GetStatus() != ateapipb.Actor_STATUS_CRASHED &&
+		actor.GetStatus() != ateapipb.Actor_STATUS_DELETING {
+		return nil, status.Errorf(codes.FailedPrecondition, "Actor %s is not in a deletable status (status: %v)", name, actor.GetStatus())
+	}
+
+	if actor.GetStatus() != ateapipb.Actor_STATUS_DELETING {
+		actor.Status = ateapipb.Actor_STATUS_DELETING
+		for _, vol := range actor.GetActorVolumes() {
+			vol.Status = ateapipb.ExternalVolume_DELETING
+		}
+		updated, err := s.persistence.UpdateActor(ctx, actor, actor.GetMetadata().GetVersion())
+		if err != nil {
+			if errors.Is(err, store.ErrPersistenceRetry) {
+				return nil, status.Error(codes.Aborted, "concurrent update conflict, please retry")
+			}
+			return nil, fmt.Errorf("while setting actor status to DELETING: %w", err)
+		}
+		actor = updated
+	}
+
 	// Delete associated volumes
-	if err := s.deleteActorVolumes(ctx, req.GetActor(), actor.GetActorVolumes()); err != nil {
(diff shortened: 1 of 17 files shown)


cmd/ateapi/internal/controlapi/volumes.go
```diff
@@ -29,54 +29,109 @@ import (
...
-				return nil, status.Errorf(codes.Internal, "failed to create volume %q: %v", vol.Name, err)
-			}
 			volumes = append(volumes, &ateapipb.ExternalVolume{
-				ActorVolumeId:   uniqueVolName,
-				StorageVolumeId: storageVolumeID,
-				VolumeType:      "mock", // TODO fix when we support multiple plugins
-				Status:          ateapipb.ExternalVolume_CREATED,
+				VolumeName: vol.Name,
+				VolumeType: "mock", // TODO fix when we support multiple plugins
+				Status:     ateapipb.ExternalVolume_CREATING,
```
juli4n: Should this be something more like `NOT_PROVISIONED` (or maybe `PENDING`)?

cmd/ateapi/internal/controlapi/delete_actor.go
```diff
@@ -44,8 +44,29 @@ func (s *Service) DeleteActor(ctx context.Context, req *ateapipb.DeleteActorRequ
...
+				return nil, status.Error(codes.Aborted, "concurrent update conflict, please retry")
+			}
+			return nil, fmt.Errorf("while setting actor status to DELETING: %w", err)
+		}
+		actor = updated
+	}
+
 	// Delete associated volumes
-	if err := s.deleteActorVolumes(ctx, req.GetActor(), actor.GetActorVolumes()); err != nil {
+	if err := s.deleteActorVolumes(ctx, actor.GetMetadata().GetUid(), actor.GetActorVolumes()); err != nil {
```
juli4n: Now that delete requires orchestrating a workflow (which needs idempotency, etc), should we add a worfklow_delete.go file for it and align with resume/suspend/pause?

cmd/ateapi/internal/controlapi/volumes.go
msau42: renamed to PENDING

cmd/ateapi/internal/controlapi/delete_actor.go
msau42: Refactored into a workflow

msau42:
I created new commits to address the review comments. I can squash before merge

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -115,19 +115,24 @@ message ResourceMetadata {
 }
 
 message ExternalVolume {
-  // actor_id + the volume name specified in the actor template.
-  string actor_volume_id = 1;
+  // Name of the volume specified in the actor template.
```
juli4n, line 118: We should add validation for these fields. If you do it in a follow up maybe add a TODO comment in create_actor.go?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -115,19 +115,24 @@ message ResourceMetadata {
...
   string storage_volume_id = 2;
 
   // Internal volume plugin name or CSI driver name.
   string volume_type = 3;
 
   enum Status {
-    PROVISIONING = 0;
-    CREATED = 1;
-    DELETING = 2;
+    STATUS_UNKNOWN = 0;
```
juli4n: Nit: This should be `STATUS_UNSPECIFIED`.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -115,19 +115,24 @@ message ResourceMetadata {
...
+    // Volume creation pending in the storage system.
+    PENDING = 1;
```
juli4n: Let's add `STATUS_` prefix here and on the other values below for consistency with other enums.

pkg/proto/ateapipb/ateapi.proto
msau42: added todo and linked in my followup tasks list

pkg/proto/ateapipb/ateapi.proto
msau42: done

pkg/proto/ateapipb/ateapi.proto
msau42: done

juli4n (approved):

PR #531, opened 2026-07-24
atenet: pick up refreshed pod certificate

zoez7:
Otherwise, envoy's serving certificate expires after 24hours.


cmd/atenet/internal/router/xds.go
```diff
@@ -571,16 +589,29 @@ func (x *XdsServer) buildHttpsListener() *listenerv3.Listener {
...
+					},
+				},
+				PrivateKey: &corev3.DataSource{
+					Specifier: &corev3.DataSource_Filename{
+						Filename: x.certPath,
+					},
+				},
+				// By specifying WatchedDirectory, we tell envoy to watch changes to the mounted pod certificate file.
+				// See documentation in https://pkg.go.dev/github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3#:~:text=This%20only%20applies%20when%20a%20%E2%80%9CTlsCertificate%E2%80%9C%20is%20delivered%20by%20SDS
+				WatchedDirectory: &corev3.WatchedDirectory{
```
BenTheElder, line 643: 🤖 **question** 🟢 – Was the reload verified end to end, rather than just the config shape? The tests pin the SDS wiring well, but nothing here exercises Envoy actually re-reading the bundle, and the failure mode is silent until the cert expires a day later. Forcing an early rotation (or touching the projected volume and watching Envoy's `secret` stats / admin `/certs`) would confirm the watch fires for a `podCertificate` volume specifically.

BenTheElder (commented):
🤖

cmd/atenet/internal/router/xds.go
zoez7: I only verified it by deploying atenet locally and forcing kubelet to resign the cert.  🤖 was right that the unit tests here only checks the expected config is delivered. I think to test envoy is actually re-reading the bundle, we would need to set up an e2e test with real envoy instance, and modify the podcertificate controller to issue certs with very short lifespan.

bowei:
@yufan-su

yufan-su:
Verified that envoy is picking up the refreshed cert. 

Test logs in https://github.com/yufan-su/substrate/blob/router-cert-sds-reload/pr-531-local-verification.md

yufan-su (dismissed):

PR #533, opened 2026-07-25
microvm durabledir support

### LINKED ISSUE #530: DurableDir support for microvm
https://github.com/agent-substrate/substrate/pull/295 added DurableDir / DATA snapshots, but only implemented for gvisor sandboxClass.

I'm implementing this for microvm, I consider this essential for milestone 2.

BenTheElder:
Fixes https://github.com/agent-substrate/substrate/issues/530

NOTE: This temporarily shares the "only one durabledir" limit, but there's no reason we couldn't lift that for uVM, and TODOs are added for this.

- [x] Tests pass

cc @dberkov

### DIFF
--- a/cmd/ateom-microvm/checkpoint.go
+++ b/cmd/ateom-microvm/checkpoint.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"time"
 
@@ -198,10 +199,13 @@ func (s *AteomService) teardownActor(ctx context.Context, id string, ra *running
 			_ = ra.chCmd.Process.Kill()
 			_, _ = ra.chCmd.Process.Wait()
 		}
-		// Kill the virtiofsd serving the overlay RO lower (after CH, its only client).
-		if ra.vfsdCmd != nil && ra.vfsdCmd.Process != nil {
-			_ = ra.vfsdCmd.Process.Kill()
-			_, _ = ra.vfsdCmd.Process.Wait()
+		// Kill the virtiofsds (after CH, their only client): the overlay RO lower's
+		// and, when the actor has durable-dir volumes, the writable share's.
+		for _, cmd := range []*exec.Cmd{ra.vfsdCmd, ra.durableVfsdCmd} {
+			if cmd != nil && cmd.Process != nil {
+				_ = cmd.Process.Kill()
+				_, _ = cmd.Process.Wait()
+			}
 		}
 	}
 
--- a/cmd/ateom-microvm/checkpoint.go
+++ b/cmd/ateom-microvm/checkpoint.go
@@ -30,19 +30,31 @@ import (
 	"github.com/agent-substrate/substrate/internal/ateompath"
 	"github.com/agent-substrate/substrate/internal/imagecache"
 	"github.com/agent-substrate/substrate/internal/proto/ateompb"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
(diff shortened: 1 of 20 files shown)


cmd/ateom-microvm/checkpoint.go
```diff
@@ -81,6 +94,72 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 		return nil, fmt.Errorf("while creating checkpoint dir %q: %w", checkpointDir, err)
 	}
 
+	// Durable-dir volumes are host-backed, so they are captured the same way under
+	// either scope — and are the ONLY thing a Data-scope snapshot captures.
+	durable := hasDurableVolumes(req.GetSpec().GetContainers())
+	scope := req.GetScope()
+	if scope == ateompb.SnapshotScope_SNAPSHOT_SCOPE_DATA && !durable {
+		return nil, status.Error(codes.FailedPrecondition,
```
dberkov: nit: line 83 
```
if err := client.Pause(ctx); err != nil {
		return nil, fmt.Errorf("while pausing guest: %w", err)
	}
```
pauses container, while line 102 throws logical error. The`if scope == ateompb.SnapshotScope_SNAPSHOT_SCOPE_DATA && !durable {` can be moved to the early stage of the method

cmd/ateom-microvm/internal/tarutil/tarutil.go
```diff
@@ -0,0 +1,360 @@
...
+	cleaned = filepath.Clean(filepath.FromSlash(name))
+	cleaned = strings.TrimPrefix(cleaned, string(filepath.Separator))
+	if cleaned == "" || cleaned == "." {
+		return "", true, nil
+	}
+	if !filepath.IsLocal(cleaned) {
+		return "", false, fmt.Errorf("not a local path: %q", name)
+	}
+	return cleaned, false, nil
+}
```
dberkov, line 374: following comment generated by Claude:

The comment above says ownership is applied before mode "so a restored setuid bit is not cleared by the chown" — but Perm() masks to 0o777, so setuid/setgid/sticky never survive extraction in the first place. Create does record them (tar.FileInfoHeader keeps the special bits), so the round trip is silently asymmetric.

The case that seems most likely to bite for durable volumes is a setgid data directory (0o2775): the workload relies on group inheritance for files created by different uids, and after a suspend/resume the bit is gone and the failure only shows up when the next file is created.

Two consistent options:
1. Restore the bits — the chown-before-chmod ordering then becomes genuinely load-bearing:
```
     if err := root.Chmod(name, hdr.FileInfo().Mode()&(fs.ModePerm|fs.ModeSetuid|fs.ModeSetgid|fs.ModeSticky)); err != nil {
```
2. Drop them deliberately and say so in the comment (defensible hardening: the tar is workload-authored and extraction runs as root, so not honoring workload-set setuid bits narrows what a compromised workload can carry across suspend/resume — at the cost of breaking setgid data dirs).

Either way, a round-trip test with a 0o2775 directory would pin the chosen behavior — the existing fixtures all use modes within 0o777, which is why the asymmetry isn't caught. (If you go with option 1, restoreDirMeta gets it via this same function; the mode at L226 used for creation can stay Perm() — it's fine for files to be born without the bits as long as the final chmod restores them.)

cmd/ateom-microvm/checkpoint.go
BenTheElder: we now validate before pausing, broadly.

cmd/ateom-microvm/internal/tarutil/tarutil.go
BenTheElder: We fixed restoring this and added test coverage for special mode bits.

PR #535, opened 2026-07-25
Refresh the actor when AssignWorkerStep's write hits a version conflict.

HavenXia:
Refresh the actor when `AssignWorkerStep`'s write hits a version conflict so that the workflow won't kept retry with a stale version actor and never succeed.

AI assisted on testing. 

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_resume.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume.go
@@ -222,7 +222,23 @@ func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, stat
 
 	updatedActor, err := s.store.UpdateActor(ctx, state.Actor, state.Actor.GetMetadata().GetVersion())
 	if err != nil {
-		return err
+		if !errors.Is(err, store.ErrPersistenceRetry) {
+			return err
+		}
+		// refresh the version of actor to avoid always failure in rest retries.
+		fresh, gerr := s.store.GetActor(ctx, input.Atespace, input.ActorName)
+		if gerr != nil {
+			slog.WarnContext(ctx, "Failed to refresh actor after assignment conflict", slog.Any("err", gerr))
+			return err
+		}
+		switch fresh.GetStatus() {
+		case ateapipb.Actor_STATUS_SUSPENDED, ateapipb.Actor_STATUS_PAUSED:
+			slog.InfoContext(ctx, "Retrying assignment due to actor version conflict", slog.String("actor", input.Atespace+"/"+input.ActorName))
+			state.Actor = fresh
+			return err
+		default:
+			return status.Errorf(codes.Aborted, "actor %s is %s and can no longer be resumed", input.ActorName, fresh.GetStatus())
+		}
 	}
 	state.Actor = updatedActor
 	state.Worker = assignedWorker
--- a/cmd/ateapi/internal/controlapi/workflow_resume_test.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume_test.go
@@ -16,10 +16,13 @@ package controlapi
 
 import (
 	"context"
+	"errors"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/scheduling"
+	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store/storetest"
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/workercache"
 	atev1alpha1 "github.com/agent-substrate/substrate/pkg/api/v1alpha1"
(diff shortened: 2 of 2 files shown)


botengyao (approved):

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -244,7 +244,22 @@ func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, stat
...
+			return err
+		}
+		// refresh the version of actor to avoid always failure in rest retries.
+		fresh, gerr := s.store.GetActor(ctx, input.Atespace, input.ActorName)
+		if gerr != nil {
+			slog.WarnContext(ctx, "Failed to refresh actor after assignment conflict", slog.Any("err", gerr))
+			return err
+		}
+		switch fresh.GetStatus() {
+		case ateapipb.Actor_STATUS_SUSPENDED, ateapipb.Actor_STATUS_PAUSED:
```
zoez7, line 235: Nit: add a log entry for "retrying assignment due to version conflict"?

cmd/ateapi/internal/controlapi/workflow_resume_test.go
```diff
@@ -404,6 +407,133 @@ func TestAssignWorkerStep_RetryAfterConflictPicksFreshWorker(t *testing.T) {
...
+	return actor, wc
+}
+
+// TestAssignWorkerStep_ConflictRefreshesActor verifies the actor write's
+// conflict handling within a single Execute: a concurrent spec write leaves
+// ErrPersistenceRetry with state.Actor refreshed, so the engine's backoff
+// retry applies on top of the concurrent write instead of replaying the same
+// stale version until the backoff is exhausted; an actor that left the
+// resumable states aborts the retry instead of being resurrected to RESUMING.
+func TestAssignWorkerStep_ConflictRefreshesActor(t *testing.T) {
```
zoez7, line 314: For readability of tests, can we use table-driven tests, and run them in parallel if possible?

zoez7 (dismissed):

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -244,7 +244,22 @@ func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, stat
 
 	updatedActor, err := s.store.UpdateActor(ctx, state.Actor, state.Actor.GetMetadata().GetVersion())
 	if err != nil {
-		return err
+		if !errors.Is(err, store.ErrPersistenceRetry) {
```
EItanya, line 225: This isn't necessarily something we need to solve in this PR, but I wonder if a `ErrVersionConflct` is in order for this scenario. Refetching the actor on EVERY error here doesn't seem quite right to me

cmd/ateapi/internal/controlapi/workflow_resume.go
HavenXia: Added

cmd/ateapi/internal/controlapi/workflow_resume.go
HavenXia: We are not refetching the actor on every error, current `ErrPersistenceRetry` means right "version mismatch", [code](https://github.com/agent-substrate/substrate/blob/main/cmd/ateapi/internal/store/store.go#L55).

cmd/ateapi/internal/controlapi/workflow_resume_test.go
HavenXia: updated

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -222,7 +222,23 @@ func (s *AssignWorkerStep) Execute(ctx context.Context, input *ResumeInput, stat
...
+		// refresh the version of actor to avoid always failure in rest retries.
+		fresh, gerr := s.store.GetActor(ctx, input.Atespace, input.ActorName)
+		if gerr != nil {
+			slog.WarnContext(ctx, "Failed to refresh actor after assignment conflict", slog.Any("err", gerr))
+			return err
+		}
+		switch fresh.GetStatus() {
+		case ateapipb.Actor_STATUS_SUSPENDED, ateapipb.Actor_STATUS_PAUSED:
+			slog.InfoContext(ctx, "Retrying assignment due to actor version conflict", slog.String("actor", input.Atespace+"/"+input.ActorName))
+			state.Actor = fresh
```
zoez7, line 237: Just a note - the correctness here relies on the fact that no other step before AssignWorkStep modifies `state.Actor`.

EItanya:
I’m ok that this merged, but honestly I don’t think this answers my question https://github.com/agent-substrate/substrate/pull/535#discussion_r3661060436

It checks a negative case on an error because that function happens to only use one other one, which is making too many assumptions about the behavior of the underlying function. The errors should describe the behavior. 


This comment is also fairly scary https://github.com/agent-substrate/substrate/pull/535#discussion_r3661853703

HavenXia:
> It checks a negative case on an error because that function happens to only use one other one, which is making too many assumptions about the behavior of the underlying function. The errors should describe the behavior.

Understood, to avoid we fold other errors into  `ErrPersistenceRetry` I will rename this specfic path error to reflect the version mismatch `ErrVersionConflict`.


> This comment is also fairly scary [#535 (comment)](https://github.com/agent-substrate/substrate/pull/535#discussion_r3661853703)

Fair - it's only safe because today no step before `AssignWorker` touches `state.Actor` and the engine retries only the failing step. Nothing enforces that.  We actually have #554 created to track it, adding to #554 for vis.

Issue #459, opened 2026-07-17
cache credential bundle instead of reading from file path on each request


zoez7:
In internal/credbundle/credbundle.go , we should cache the credential bundle instead of reading from file path on every request.

References: 
* PodIdentity signer issues certs that are valid for 24 hours: https://github.com/agent-substrate/substrate/blob/381b3ad52ca7dbf04e453872a3acfe2d6bd4cbd0/cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go#L108

zoez7:
Fixed by https://github.com/agent-substrate/substrate/pull/537

PR #537, opened 2026-07-25
Cache parsed credential bundles in credbundle loaders

botengyao:
Resolves the `TODO: Introduce caching.` on `credbundle.Loader` / `credbundle.ClientLoader`. The TODO dates back to the package's introduction; since #237 established mTLS between all ate system components, every inter-component handshake is on this path.

## Problem

`Loader`/`ClientLoader` re-read and re-parse the credential bundle from disk on every TLS handshake so that pod-certificate rotations get picked up. That costs ~125µs and ~120 allocations per handshake (RSA bundle) for a file that almost never changes.

## Approach: stat-validated cache

Each loader keeps the last successful parse together with the `os.FileInfo` taken just before that parse. A handshake stats the file (~1µs) and serves the cached certificate while file identity (`os.SameFile`, i.e. device+inode on Unix), mtime, and size all match; any mismatch re-reads and re-parses.

Invalidation and eviction:

- The kubelet rotates projected-volume contents (including pod certificates) by writing a fresh timestamped directory and atomically swapping a symlink, so a rotation always surfaces as a change of file identity and is picked up on the **next handshake** — no staleness window. mtime and size additionally cover in-place rewrites.
- **No TTL**: freshness is derived from the file, never from time, so an unchanged bundle is never re-parsed and a changed one is never served. (A refresh-interval floor to throttle stats was considered and rejected: it would introduce a staleness window to save a ~1µs syscall that only runs at connection-establishment rate. fsnotify was also considered and rejected: watching through symlink swaps is error-prone and adds watcher lifecycle for no latency benefit over the stat.)
- **Memory**: exactly one parsed certificate per loader (one loader per `tls.Config`), replaced on rotation — nothing unbounded to evict.
- **Failure semantics unchanged**: stat/read/parse errors are returned to the caller (never a stale certificate), the previous entry stays in place, and every later handshake retries until a parse succeeds — exactly the uncached behavior.
- The file can change between the stat and the read; the parse is then stored against the older stat, so the next handshake detects the mismatch and re-reads. The cache never lags a rotation by more than one handshake.

## Benchmarks

Apple M5 Pro, RSA-2048 bundle:

| | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| `BenchmarkLoaderCached` | 1,001 | 304 | 2 |
| `BenchmarkParseUncached` | 124,843 | 19,544 | 120 |

`Parse` stays exported and uncached; `Loader`/`ClientLoader` signatures are unchanged, so no call sites move.

- [x] Tests pass (`go test ./internal/credbundle/ -race`: cache-hit proof via unchanged-stat corruption, atomic-rename rotation, in-place rewrite, missing-file error, error recovery, concurrent handshakes under the race detector)
- [x] Appropriate changes to documentation are included in the PR (doc comments on `Loader`/`ClientLoader`/`certCache` describe the caching, rotation, and failure semantics)

### DIFF
--- a/internal/credbundle/credbundle.go
+++ b/internal/credbundle/credbundle.go
@@ -89,7 +89,7 @@ func (c *certCache) get() (*tls.Certificate, error) {
 
 	fi, err := os.Stat(c.path)
 	if err != nil {
-		return nil, fmt.Errorf("while examining credential bundle: %w", err)
+		return nil, fmt.Errorf("while getting file info for credential bundle %q: %w", c.path, err)
 	}
 	if c.cert != nil && os.SameFile(c.fi, fi) && fi.ModTime().Equal(c.fi.ModTime()) && fi.Size() == c.fi.Size() {
 		return c.cert, nil
--- a/internal/credbundle/credbundle.go
+++ b/internal/credbundle/credbundle.go
@@ -25,28 +25,82 @@ import (
 	"encoding/pem"
 	"fmt"
 	"os"
+	"sync"
 )
 
 // Loader reads a private key and certificate chain from a credential bundle file as written by the
 // Kubernetes Pod Certificates mechanism.
 //
-// Returns a function that can be used as GetCertificate in a tls.Config
+// Returns a function that can be used as GetCertificate in a tls.Config. The parsed bundle is
+// cached: each handshake stats the file and re-reads it only when the file has changed, so
+// pod-certificate rotations are picked up on the next handshake without paying the read and
+// parse cost when nothing changed.
 func Loader(path string) func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
-	// TODO: Introduce caching.
+	c := &certCache{path: path}
 	return func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
-		return Parse(path)
+		return c.get()
 	}
 }
 
 // ClientLoader is the client-side counterpart to Loader. It returns a function
-// suitable for use as GetClientCertificate in a tls.Config, re-reading the
-// bundle on each handshake so that in-place pod-certificate rotations are
-// picked up.
+// suitable for use as GetClientCertificate in a tls.Config, caching the parsed
+// bundle in the same way so that pod-certificate rotations are picked up on
+// the next handshake.
 func ClientLoader(path string) func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
-	// TODO: Introduce caching.
+	c := &certCache{path: path}
 	return func(_ *tls.CertificateRequestInfo) (*tls.Certificate, error) {
-		return Parse(path)
+		return c.get()
 	}
+}
+
+// certCache holds the parse of a credential bundle file together with the stat
+// of the file it was parsed from, so unchanged files are not re-parsed on
+// every TLS handshake.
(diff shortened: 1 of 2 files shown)


internal/credbundle/credbundle_test.go
```diff
@@ -87,11 +294,58 @@ func generateCertificate(t *testing.T) []byte {
...
+
+func leafSerial(t *testing.T, cert *tls.Certificate) int64 {
+	t.Helper()
+	if cert == nil || cert.Leaf == nil {
+		t.Fatalf("parsed bundle has no leaf certificate")
+	}
+	return cert.Leaf.SerialNumber.Int64()
+}
+
+func BenchmarkLoaderCached(b *testing.B) {
```
zoez7: Nit: remove unused functions BenchmarkLoaderCached and BenchmarkParseUncached

internal/credbundle/credbundle.go
```diff
@@ -25,30 +25,84 @@ import (
...
+// than one handshake. Errors leave the previous entry in place and are
+// returned to the caller: handshakes fail exactly as they would without
+// caching, and every later call retries until a parse succeeds.
+func (c *certCache) get() (*tls.Certificate, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	fi, err := os.Stat(c.path)
+	if err != nil {
+		return nil, fmt.Errorf("while examining credential bundle: %w", err)
```
zoez7: Nit: `fmt.Errorf("while getting file info for certpath %q: %w", c.path, err)`?

internal/credbundle/credbundle_test.go
```diff
@@ -61,7 +64,211 @@ func TestParseRejectsNonPKCS8PrivateKeyBlock(t *testing.T) {
...
+	cert, err := getCert(nil)
+	if err != nil {
+		t.Fatalf("Loader() first call error = %v", err)
+	}
+	if got := leafSerial(t, cert); got != 1 {
+		t.Fatalf("Loader() leaf serial = %d, want 1", got)
+	}
+
+	// Rotate the way the kubelet does: write the new bundle to a fresh file
+	// and atomically rename it over the old path, giving it a new inode.
```
zoez7: The comment here is inaccurate. The certPath is a symlink to the `..data` dir, kubelet rotates by writing to a new tmp dir first, which is then renamed to `..data`. 
Can we follow https://github.com/kubernetes/kubernetes/blame/master/pkg/volume/util/atomic_writer.go#L114 to mimic what kubelet does?

zoez7 (dismissed):

internal/credbundle/credbundle_test.go
botengyao: done

internal/credbundle/credbundle.go
botengyao: done

internal/credbundle/credbundle_test.go
botengyao: good point, done.

botengyao:
@zoez7 thanks for the review! PTAL.

zoez7 (approved):

zoez7:
This fixes https://github.com/agent-substrate/substrate/issues/459

PR #538, opened 2026-07-25
Run ate-api-server in HA mode

juli4n:
This PR updates `ate-apiserver` deployment to run with 2 replicas.

Some changes were needed to enable this, and are separate in a series of independent commits.

* Refactor port-forwarding logic: A few components were doing this, in different ways, but most importantly, the `kubectl-ate` binary was not checking for Pod readiness. With multiple replicas, there is a risk of picking a not ready Pod and failing the request.
* Delete `agent-secret` demo: This demo is broken now that `ate-apiserver` requires authn. I don't see an easy way to fix it, and the demo was not really adding much value.
* Change mock volume plugin so that it doesn't require global in-memory state: This just doesn't work when running multiple replicas. Luckily, we don't need to keep any state, so change the mock to be a no-op.
* Enable round-robin load balancing across replicas: Switch gRPC clients to keep a channel with a sub-channel per replica, and then round robin traffic across them. The current approach of connecting to the Service ClusterIP VIP doesn't really load balance as gRPC is session based and keeps connection open for a long time. The fix is to switch to a headless service and let the gRPC client create the connections by discovering the DNS A records.
* Set `ate-apiserver` deployment to `replicas: 2`:  Additionally, configure the deployment rolling update so that we update one instance at a time (make before break) and PDB so that we never lose more than one replica during eviction events.

Part of #181


cmd/ateapi/main.go
```diff
@@ -195,6 +196,14 @@ func main() {
 	mux := grpc.NewServer(
 		grpc.Creds(serverCreds),
 		grpc.StatsHandler(otelgrpc.NewServerHandler()),
+		// Bounds every connection's lifetime so round_robin clients
+		// periodically re-resolve DNS and pick up replicas added since they
+		// last connected - without this, an existing connection never
+		// notices new replicas on its own (see https://github.com/grpc/grpc/issues/12295).
+		grpc.KeepaliveParams(keepalive.ServerParameters{
```
michaelasp, line 208: This is a bit concerning to me, effectively we are trading off session stability(1m keepalive) in order to obtain the endpoints. I see in the referenced issue that people get periodic unavailable errors with this approach. From the aforementioned issue:

>The Go client has [this resolver](https://pkg.go.dev/google.golang.org/grpc/resolver) thing. You can import the package and Register a resolver builder or whatever. The client automatically picks it up, so I guess it's a global.

>My coworker hacked something together to watch endpoints in the Kubernetes API and we have perfect load balancing when pods go in an out of service.

I think this is fine for now as we iterate, but seems like there's a way to pass endpoints into the resolver. We can probably make this a follow up bug, I don't mind hacking on this if wanted :)

juli4n: I think we will eventually end up implementing our own resolver that watches endpoints. Let me add a TODO comment.

msau42:
mock volume changes lgtm

manifests/ate-install/ate-api-server.yaml
```diff
@@ -62,7 +62,12 @@ metadata:
   name: ate-api-server
   namespace: ate-system
 spec:
-  replicas: 1
+  replicas: 2
```
HavenXia, line 65: There will be 2 syncers with this change, are syncers' store update behavior idempotent?

HavenXia: btw this is another proof that why we should not do inplace ugrade for substrate otherwise old and new ateapis are certainly interleaved.

manifests/ate-install/ate-api-server.yaml
```diff
@@ -208,14 +213,29 @@ spec:
               - key: "pool"
                 path: "pool.json"
```
HavenXia, line 215: should we add podAntiAffinity to avoid replicas being deployed on same node for a better HA behavior?

internal/ateclient/builder.go
```diff
@@ -159,81 +155,15 @@ func dialPortForward(ctx context.Context, kubeconfigPath, k8sContext string, tra
 		return nil, fmt.Errorf("failed to create k8s client: %w", err)
 	}
 
-	// Look up the 'api' Service to dynamically get its pod selector
-	svc, err := clientset.CoreV1().Services("ate-system").Get(ctx, "api", metav1.GetOptions{})
+	localPort, stopForward, err := portforward.ServicePortForward(ctx, config, clientset, "ate-system", "api", 443)
```
EItanya, line 160: Should we check for an LB address first?

manifests/ate-install/ate-api-server.yaml
EItanya: > There will be 2 syncers with this change, are syncers' store update behavior idempotent?

I think that is exactly why he is bumping it to 2, if it does not work with more than 1 replica we need to figure that out now, this is an important aspect of scaling substrate.

> btw this is another proof that why we should not do inplace ugrade for substrate otherwise old and new ateapis are certainly interleaved.

This is a problem we need to solve regardless. When in Alpha slightly less so, but API breaks are a natural part of any system that needs to be handled

HavenXia: > This is a problem we need to solve regardless. When in Alpha slightly less so, but API breaks are a natural part of any system that needs to be handled

I agree, it's actually a M2 P0 I am working on, blue/green cutover upgrade can reduce the pain of api breaks.

manifests/ate-install/ate-api-server.yaml
juli4n: Yes. I intentionally left spread constraints out of scope as we are running on Kind, with a single node. We will definitely need to add rules to spread across nodes and zones.

HavenXia: That makes senses, thank you.

HavenXia (dismissed):

internal/ateclient/builder.go
juli4n: Added a TODO to consider handling that.

PR #540, opened 2026-07-25
kubectl-ate: keep client-go unhandled-error reports off stderr

### LINKED ISSUE #26: Sporadic error copying from local connection to remote stream ... connection reset by peer"
## Expected Behavior

No error

## Actual Behavior

```
$ kubectl ate create actor ctr2 --template ate-demo-counter/counter
NAMESPACE          TEMPLATE   ID     STATUS             ATEOM POD   ATEOM IP   VERSION
ate-demo-counter   counter    ctr2   STATUS_SUSPENDED   <none>                 1
E0520 15:40:33.154830 3507355 portforward.go:502] "Unhandled Error" err="error copying from local connection to remote stream: writeto tcp4 127.0.0.1:42521->127.0.0.1:42648: read tcp4 127.0.0.1:42521->127.0.0.1:42648: read: connection reset by peer" logger="UnhandledError"
```

1. hack/create-kind-cluster.sh && hack/install-ate-kind.sh --deploy-ate-system && hack/install-ate-kind.sh --deploy-demo-counter && go install ./cmd/kubectl-ate
1. kubectl ate create actor ctr1 --template ate-demo-counter/counter
1. kubectl ate create actor ctr2 --template ate-demo-counter/counter

NOTE: I did not run port-forward

## Specifications

- Version: HEAD
- Platform: Linux

YQ-Wang:
Fixes #26

- [x] Tests pass

## Summary

A *successful* `kubectl ate create actor` can print this:

```
E0520 15:40:33.154830 3507355 portforward.go:502] "Unhandled Error" err="error copying from local connection to remote stream: ... read: connection reset by peer"
```

The user never asked for port-forwarding — it is an implementation detail of this plugin — so a message naming `portforward.go` is not something they can act on.

client-go reports problems it cannot return to a caller through `utilruntime.HandleError`, whose default handler writes to stderr. This replaces that handler for `kubectl-ate`: those reports are dropped by default and printed under a new `--verbose` flag. klog is deliberately left alone, so warnings a user can act on directly — a missing kubeconfig, an exec credential plugin that failed to refresh — keep reaching the terminal. That is enough because every port-forward error that reaches a user goes through `HandleError`, while the package's direct klog calls are all `V(4)` or higher.

#539 proposes removing the localhost hop that produces this particular reset. That is a ~200 line transport change for one message, though, and the next `HandleError` call inside client-go would read just as strangely. Handling it here covers the class and leaves #539 to be weighed on its own merits.

## Notes for reviewers

- Reflector and informer errors travel through the same channel. `kubectl-ate` makes one-shot RPCs today, so nothing actionable is lost, but a watch-based command should revisit this.
- The flag is `--verbose` rather than `--v=N` because `-v` is already cobra's `--version` shorthand here, and this does not wire klog verbosity. Happy to match that or the server binaries' `--log-level` if you prefer.

## Validation

`go test -race -count=5 ./cmd/kubectl-ate/...` and the `hack/verify` scripts pass, with no changes to `go.mod` or `vendor/modules.txt`. New tests assert the exact text from #26 produces no output by default and is printed under `--verbose`, and the flag is documented in the CLI README.

The reset in #26 is timing-dependent and did not reproduce on macOS in 1,500 invocations against a kind cluster, so I forced the same code path by pointing the forwarder at a port with nothing listening. On `main`:

```
E0725 02:52:41.650594 63626 portforward.go:522] "Unhandled Error" err="an error occurred forwarding 54823 -> 9999: ... connect: connection refused" logger="UnhandledError"
Error: failed to list atespaces: rpc error: code = Unavailable ...
```

With this change only the second line is printed, and `--verbose` restores the first. A `klog.Warning` raised with the quiet handler installed still reaches stderr, and normal commands and error paths are identical to `main`.

### DIFF
--- a/cmd/kubectl-ate/README.md
+++ b/cmd/kubectl-ate/README.md
@@ -26,6 +26,8 @@ By default, `kubectl-ate` will automatically read your `~/.kube/config`, discove
 
 If you prefer to route traffic directly (e.g., through a LoadBalancer or when running natively inside a cluster pod), simply provide the `--endpoint` flag to bypass the tunnel.
 
+Because the tunnel is an implementation detail, diagnostics that the Kubernetes client reports about it are not printed. Pass `--verbose` to see them when troubleshooting a connection.
+
 ## Tracing
 The CLI supports on-demand tracing using the `--trace` flag. When enabled, the CLI will generate a trace ID and signal to the server that it wants the request to be traced.
 
@@ -67,6 +69,7 @@ These flags can be appended to any command:
 | `--endpoint` | | Manual gRPC endpoint override (e.g., `localhost:8080`) | |
 | `--output` | `-o` | Output format (`table`, `json`, `yaml`) | `table` |
 | `--trace` | | Enable on-demand tracing for the request | `false` |
+| `--verbose` | | Print internal Kubernetes client diagnostics that are otherwise suppressed | `false` |
 
 ---
 
--- a/cmd/kubectl-ate/internal/cmd/clientlogging.go
+++ b/cmd/kubectl-ate/internal/cmd/clientlogging.go
@@ -1,0 +1,51 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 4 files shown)


YQ-Wang:
Could you take a look? @thockin

PR #543, opened 2026-07-27
Build the locust image for the cluster's architecture

chuangw6:
`build_and_push.sh` ran `docker build` with no `--platform`, so an arm64
workstation produced an arm64 image while GKE nodes (c3-standard-4) are
amd64. The push succeeds and nothing fails until the kubelet pulls, two
steps later:

    no match for platform in manifest: not found

Default to linux/amd64, overridable via LOCUST_IMAGE_PLATFORM for arm64
node pools (T2A, Axion).

Verified: pushed manifest reports linux/amd64 and the locust pod reaches
3/3 Running.

Stacked on #542.


chuangw6:
Reopening from my fork instead of a branch on this repo. No content change.

PR #544, opened 2026-07-27
Pace boomer iterations that exit early

chuangw6:
`iterate()` only slept after a full resume/ping/suspend cycle. The early returns — `startUser` failing, `resume` failing — skipped the wait, and boomer re-enters `iterate` immediately, so a VU that could not get a worker spun at gRPC round-trip speed.

That is backwards from what you want under load: run more VUs than there are workers and every starved VU hammers ResumeActor at full rate, inflating the reported request counts and adding load to the ateapi that the healthy VUs are waiting on.

Move the wait into a `defer` so every path pays it.

Stacked on #543.


chuangw6:
Not load-bearing for anything we measured: ResumeActor was 0% fail in every run, so the early-return path never fired. Closing to keep the stack to changes we actually diagnosed.

PR #545, opened 2026-07-27
Add a readyz probe to the benchmark glutton actor

chuangw6:
## Problem

ResumeActor returned as soon as the resume workflow finished, not when glutton's port 80 listener was accepting connections. A ping issued right after resume could land in that gap, get ECONNREFUSED, and come back from the router as a 503. At 50 VUs this was roughly 42% of pings.

ateom already has a gate for exactly this: `internal/readyz.WaitAll` blocks `RestoreWorkload` and `RunWorkload` until every container that declares a readyz probe returns 200. But containers without a probe are skipped entirely, and glutton declared none, so nothing waited on it.

## Solution

Add `/readyz` to glutton's HTTP mux and point the ActorTemplate at it.

The route goes on the same listener as `/ping` rather than the metrics port: the metrics server starts independently, so a 200 there would say nothing about whether `/ping` is reachable yet.

## Notes

`ActorTemplate.spec` is immutable, so picking this up needs a delete and recreate of the template, not an apply.

Stacked on #543.

### DIFF
--- a/benchmarking/workloads/manifests/workloads.yaml.tmpl
+++ b/benchmarking/workloads/manifests/workloads.yaml.tmpl
@@ -70,6 +70,15 @@ spec:
     - "--metrics-listen-addr=:9090"
     - "--data-dir=/tmp/glutton"
     - "--mode=http"
+    # Hold ResumeActor open until port 80 actually answers. Starting the
+    # process is not the same as it accepting connections, and without this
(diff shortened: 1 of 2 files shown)


chuangw6:
Reopening from my fork instead of a branch on this repo, and rebased to stand alone on main rather than stacking on #543. No content change.

PR #549, opened 2026-07-27
Stop Envoy reusing upstream connections to actors

chuangw6:
> **Note from Claude (AI assistant), not from Chuang :)** this branch should have lived on Chuang's fork, and the PR should have been opened from `chuangw6:bench/envoy-no-connection-reuse` into `main`. I created it here on the shared repo instead, which was my mistake. Deleting the branch now would close this PR and take the existing review threads with it, so Chuang agreed to leave this one as it is.

## Problem

Envoy pools HTTP/1.1 connections by destination address. That is safe when an address means one stable server, and for actors it does not. A worker pod keeps its IP for its whole lifetime, but the actor sandbox behind port 80 is torn down on every Suspend and a different actor takes the slot. A pooled connection therefore belongs to an actor that may already be gone, and reusing it races the far end's close.

## Symptom

The request goes into a dead socket and Envoy synthesizes a 503. At 50 actors churning across 50 workers, ~42% of pings failed, with Envoy reporting 450 `upstream_cx_destroy_remote_with_active_rq`.

## Solution

Set `max_requests_per_connection` to 1 on the dynamic forward proxy cluster. Failures and the reset counter both went to 0.

## Notes

The handshake this costs is sub-millisecond in-cluster, against pings that take ~22ms. It removes the race rather than recovering from it after the fact, which is why this is preferred over a route-level retry policy.

Carrying `HttpProtocolOptions` on a `dynamic_forward_proxy` cluster also requires `allow_insecure_cluster_options`. Envoy guards that because a DFP cluster forwards to whatever host the request names, so a TLS upstream must validate against it. This cluster has no transport socket at all and the authority is an IP literal, so there is nothing to validate. Without the flag Envoy NACKs every CDS push and drops the cluster, which presents as "all actor traffic 503s" rather than as a config error, so there is a test pinning it.

### DIFF
--- a/cmd/atenet/internal/router/xds.go
+++ b/cmd/atenet/internal/router/xds.go
@@ -66,6 +66,12 @@ const (
 	RouteName            = "substrate_routes"
 	ClusterName          = "ate-cluster"
 	OtlpClusterName      = "otel_collector_cluster"
+
+	// httpProtocolOptionsName is the well-known extension key Envoy looks for in
+	// a cluster's typed_extension_protocol_options. It must match the message's
+	// full proto type name exactly; a typo is silently ignored rather than
+	// rejected, so the options simply never take effect.
+	httpProtocolOptionsName = "envoy.extensions.upstreams.http.v3.HttpProtocolOptions"
 )
 
 // XdsServer implements an aggregated discovery service server for dynamic Envoy router nodes.
@@ -262,7 +268,7 @@ func (x *XdsServer) buildCluster() *clusterv3.Cluster {
 			},
 		},
 		TypedExtensionProtocolOptions: map[string]*anypb.Any{
-			"envoy.extensions.upstreams.http.v3.HttpProtocolOptions": h2Opts,
+			httpProtocolOptionsName: h2Opts,
 		},
 	}
 }
@@ -324,7 +330,7 @@ func (x *XdsServer) buildOtlpCollectorCluster() *clusterv3.Cluster {
 			},
 		},
 		TypedExtensionProtocolOptions: map[string]*anypb.Any{
-			"envoy.extensions.upstreams.http.v3.HttpProtocolOptions": h2Opts,
+			httpProtocolOptionsName: h2Opts,
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


cmd/atenet/internal/router/xds.go
```diff
@@ -334,10 +340,46 @@ func (x *XdsServer) buildDynamicForwardProxyCluster() *clusterv3.Cluster {
...
 
+	// Disable upstream connection reuse.
+	//
+	// Envoy pools HTTP/1.1 connections by destination address, which is safe
+	// when an address means one stable server. Here it does not. A worker pod
+	// keeps its IP for its whole lifetime, but the actor sandbox behind port 80
+	// is destroyed on every Suspend and a different actor takes the slot. A
+	// pooled connection therefore belongs to an actor that may already be gone,
+	// and reusing it races the far end's close: the request goes into a dead
+	// socket and Envoy synthesizes a 503. Measured at ~40% of pings with 50
```
maxsmythe: We should avoid citing specific latency/measurement references as those are coupled with infrastructure and may quickly go stale as project progresses. Noting the shape of the concern and that it has been observed should be enough to help with Chesterton's Fence in the future.

cmd/atenet/internal/router/xds.go
```diff
@@ -334,10 +340,46 @@ func (x *XdsServer) buildDynamicForwardProxyCluster() *clusterv3.Cluster {
...
+	// One request per connection costs a TCP handshake — sub-millisecond
+	// in-cluster, against pings that take ~22ms — and removes the race outright
+	// rather than recovering from it after the fact.
+	httpOpts, _ := anypb.New(&httpv3.HttpProtocolOptions{
+		CommonHttpProtocolOptions: &corev3.HttpProtocolOptions{
+			MaxRequestsPerConnection: wrapperspb.UInt32(1),
+		},
+		UpstreamProtocolOptions: &httpv3.HttpProtocolOptions_ExplicitHttpConfig_{
+			ExplicitHttpConfig: &httpv3.HttpProtocolOptions_ExplicitHttpConfig{
+				// HTTP/1.1 upstream, matching what actors serve on port 80.
```
maxsmythe, line 363: Do we know what actors will be serving on port 80? Currently we are assuming HTTP/1.1 as a simplification. That may not always be true.

Not familiar enough with envoy to know how limiting this configuration will be on future updates. Are we precluded from supporting streaming connections, proto servers, etc. in the future with this choice?

maxsmythe (commented):
Mostly nits, but the hard-coding of HTTP/1.1 is something I'd like to understand the impact of more.

@bowei thoughts?

cmd/atenet/internal/router/xds.go
keithmattix: I think these are all defaults so nothing is locking us into http 1.1

cmd/atenet/internal/router/xds.go
chuangw6: SG. Removed!

cmd/atenet/internal/router/xds.go
chuangw6: Keith's right that nothing changes today — Envoy's default upstream protocol is already HTTP/1.1. And I can't leave it out: `HttpProtocolOptions` requires the `upstream_protocol_options` oneof, so there's no way to set `max_requests_per_connection` without naming a protocol. It's a required field, not a new opinion. (The other two clusters (ate-cluster and otel_collector_cluster) set h2 because they carry gRPC and have to override the default; this one is spelling the default in.)

maxsmythe: Thanks for the explanation, SGTM

maxsmythe (approved):
LGTM

PR #556, opened 2026-07-27
Add a readyz probe to the benchmark glutton actor

chuangw6:
## Problem

ResumeActor returned as soon as the resume workflow finished, not when glutton's port 80 listener was accepting connections. A ping issued right after resume could land in that gap, get ECONNREFUSED, and come back from the router as a 503.

ateom already has a gate for exactly this: `internal/readyz.WaitAll` blocks `RestoreWorkload` and `RunWorkload` until every container that declares a readyz probe returns 200. But containers without a probe are skipped entirely, and glutton declared none, so nothing waited on it.

## Solution

Add `/readyz` to glutton's HTTP mux and point the ActorTemplate at it.

The route goes on the same listener as `/ping` rather than the metrics port: the metrics server starts independently, so a 200 there would say nothing about whether `/ping` is reachable yet.

## Notes

`ActorTemplate.spec` is immutable, so picking this up needs a delete and recreate of the template, not an apply.

### DIFF
--- a/cmd/benchmarking/glutton/main.go
+++ b/cmd/benchmarking/glutton/main.go
@@ -136,6 +136,11 @@ func main() {
 		// (re-exposable as additional routes if/when needed).
 		mux := http.NewServeMux()
 		mux.HandleFunc("/ping", httpPingHandler(svc))
+		// ateom blocks RestoreWorkload until this returns 200, so ResumeActor
+		// cannot report success before /ping is reachable.
+		mux.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+		})
 		// otelhttp at the mux level + per-handler span follows
 		// docs/dev/best-practices/tracing.md: extract incoming context,
 		// then name the span after the operation in each handler.
--- a/cmd/benchmarking/glutton/main.go
+++ b/cmd/benchmarking/glutton/main.go
@@ -29,6 +29,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/benchmarking/glutton/main.go
```diff
@@ -136,6 +136,18 @@ func main() {
...
 		mux.HandleFunc("/ping", httpPingHandler(svc))
+		// Readiness gate for the actor lifecycle. ateom blocks RestoreWorkload
+		// on this returning 200 (internal/readyz), so ResumeActor does not
+		// report success until the socket the router forwards to is genuinely
+		// serving. Without it a ping can beat the listener and get
+		// ECONNREFUSED, which the router turns into a 503.
+		//
+		// It must live on this listener rather than the metrics port: the
+		// metrics server starts independently, so a 200 there would say
+		// nothing about whether /ping is reachable yet.
```
maxsmythe: Comments seem excessive here. Discussion of metrics port seems like a non-sequitur

maxsmythe (changes requested):
In general, streamline comments, otherwise LGTM

bowei (dismissed):

bowei:
remove the comments and you should be good

cmd/benchmarking/glutton/main.go
chuangw6: updated.

chuangw6:
@maxsmythe I've updated the comment. Are we good to merge this?

cmd/benchmarking/glutton/main.go
```diff
@@ -136,6 +136,11 @@ func main() {
 		// (re-exposable as additional routes if/when needed).
 		mux := http.NewServeMux()
 		mux.HandleFunc("/ping", httpPingHandler(svc))
+		// ateom blocks RestoreWorkload until this returns 200, so ResumeActor
+		// cannot report success before /ping is reachable.
+		mux.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
```
maxsmythe: We probably need to serve /readyz regardless of mode (grpc vs. http). Apologies for not spotting this earlier.

maxsmythe (changes requested):

maxsmythe:
Thanks for the ping, found 1 more issue

cmd/benchmarking/glutton/main.go
chuangw6: Done.

maxsmythe (approved):
LGTM, thanks for update!

Issue #484, opened 2026-07-21
router: Future-Proof Non-HTTP actor protocols


keithmattix:
# Motivation
atenet router's primary function is to receive an incoming request, parse the actor and ate space name from the Host header, and resume the actor before forwarding the request to the worker IP hosting that actor. While this is currently HTTP-only, the current hostname-style addressing scheme leaves open the possibility of supporting non-HTTP actor protocols by using TLS (the router uses SNI to parse the `atespace:actor` tuple). Even this has issues though as it's not possible to front a single wildcard TLS certificate for actors as the DNS name has two wildcarded components (*.*.actors.resources.substrate.dev). In order to pave the way for HTTP-specific actor addressing schemes (e.g. paths or headers), this issue proposes a strategy to support actors with non-HTTP protocols.

# Proposal

Substrate ingress standardizes on HTTP, HTTPS, and HTTP/2 (TLS and h2c) as the supported transports for calling an actor. Clients will explicitly dial the ingress load balancer (likely via a DNS name provisioned out-of-band) since the encoding will occur outside of the host header (exact shape tbd; we can discuss in #445). To support actors that don't speak HTTP, the ingress must support terminating CONNECT and forwarding the opaque TCP packets from tunnel. In this model, the ingress will be able to have a single parsing scheme for determining the atespace;actor tuple regardless of actor protocol. 

Happy to take a stab at POCing this if we think it's a good idea

bowei:
1) Need for double wildcard in DNS -- this feels like a separate issue from the non-HTTP actor protocol support? We might want to split that one off to its own discussion.

2) This is an interesting approach -- I'm assuming this would be in addition to the standard 80, 443 that we expose today. CONNECT would allow arbitrary ports, protocols without having to play games with the router setup. The router setup using K8s Service for instance becomes very awkward due to lack of support for ALL_PORTS etc.

This would enable something like?

```bash
$ export http_proxy=my-ate-router.ate.dev:8443
$ my-protocol-client my-actor.my-atespace.router.ate.dev:12345
```

We would end up with:

* 80, 443 on the router remains an easy to use front door. Note: we need to fix having 443 go to 443 on the Actor for consistency.
* You can use standard HTTP CONNECT (v1, v2, ...) to 8443 to use CONNECT tunneling for an opaque stream.

Right now the fields that would be interpreted would just be the `:authority` and related fields.

bowei:
@LiorLieberman

keithmattix:
[...]

PR #557, opened 2026-07-27
atenet: map HTTPS requests to actor port 443 with TLS re-origination

yufan-su:
# Problem
The atenet router terminates client TLS on its HTTPS listener (0.0.0.0:8443) and then forwards the request cleartext to podIP:80 — exactly like a plaintext request. It was flagged in #484 

# Fix
Requests arriving over TLS get TLS re-originated by Envoy to the actor on port 443. Plaintext requests still go to port 80 unchanged.

 - xds.go — the HTTPS listener gets its own route config and its own DFP cluster with an UpstreamTlsContext. All three resources are gated together on the existing httpsPort > 0 check so snapshot.Consistent() never sees a dangling reference.
 - SNI — ext_proc rewrites :authority to a bare pod IP, so auto_sni would offer an IP literal. The TLS cluster sets OverrideAutoSniHeader: x-substrate-sni and ext_proc emits that header with the actor's mesh DNS name. OVERWRITE_IF_EXISTS_OR_ADD, not the default append, so a caller can't slip its own value into the handshake.
- ateom-gvisor / ateom-microvm — one DNAT rule per port instead of a hardcoded 80.

# Notes
 - Port selection is based on the field `connection.tls_version`, for HTTP request it is empty.
 - demos/counter was updated for local testing on HTTPs requests.

### DIFF
--- a/cmd/atenet/internal/router/extproc.go
+++ b/cmd/atenet/internal/router/extproc.go
@@ -176,8 +176,9 @@ func (s *ExtProcServer) handleRequestHeaders(
 			"actor %q routing failed", actorName)
 	}
 
-	// Actors serve plaintext on 80. A request that arrived over TLS is
-	// re-originated as TLS to 443, so the actor sees the scheme the client used.
+	// Actors serve plaintext on 80. A request whose downstream connection was
+	// TLS is re-originated as TLS to 443, so the actor is reached over the same
+	// transport the client used.
 	// TODO(bowei) -- handle actor ports beyond this fixed 80/443 pair.
 	targetAddr := net.JoinHostPort(workerIP, actorPortForIngress(tlsIngress))
 
--- a/cmd/atenet/internal/router/extproc.go
+++ b/cmd/atenet/internal/router/extproc.go
@@ -31,7 +31,9 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/propagation"
 	"google.golang.org/grpc"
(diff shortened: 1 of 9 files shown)


haiyanmeng:
@yufan-su , FYI, https://github.com/agent-substrate/substrate/pull/548 refactors ateom networking to shared internal/ateomnet package.

haiyanmeng:
@yufan-su , can you squash the three commits in the PR into a single one?

haiyanmeng:
/hold

@yufan-su , https://github.com/agent-substrate/substrate/pull/559 deletes the dynamic forward proxy entirely. 

cc @LiorLieberman  and @bowei

EItanya (commented):
Frankly I think we should hold off on this for a bit. I don't think we have a clear idea of how ports/protocols should be configured for incoming traffic to Actors, so rather than adding more, let's take a step back and decide where we want to go.

See https://github.com/agent-substrate/substrate/issues/484, https://github.com/agent-substrate/substrate/issues/265, and https://github.com/agent-substrate/substrate/pull/296 for previous discussion

howardjohn:
This seems concerning to me. If we start with a premise that an actor has one port (I say this because its a current limitation, but also because other sandboxing platforms do this!) then what you would want is to terminate on 443 at the ingress and then route to 80 in the actor. This makes it impossible to do that simple case.

Once we then make it more complex and move towards supporting multiple exposed ports in the actor this muddles things even further.

PR #558, opened 2026-07-28
Authenticate the benchmark's ateapi clients

chuangw6:
The benchmark has been unrunnable on `main` since #237 landed. Every ateapi
call fails with:

```
rpc error: code = Unauthenticated desc = missing bearer token
```

`ateapiauth`'s interceptor takes identity from an mTLS client certificate, or
failing that an `authorization: Bearer <jwt>` header, and rejects anything
else. The benchmark's clients still dialed with server-side TLS only.

#237 gave the ate-system components a pod certificate and the flags to present
it. The benchmark has its own Deployment, outside `manifests/ate-install/`, so
it was missed.

## What this does

- Projects a `podidentity.podcert.ate.dev/identity` certificate into the locust
  pod, mounted in all three containers.
- Go boomer: `DialControl` now builds its options through
  `ateapiauth.DialOptions` with `ClientCredBundle`, so `credbundle.ClientLoader`
  re-reads the bundle on every handshake and rotations need no restart. This
  also drops the `InsecureSkipVerify` it was using — the servicedns trust bundle
  was already mounted and nothing was reading it.
- Python user classes: a new `common/ateapi_channel.py` replaces the five-line
  channel setup duplicated across all five of them.

Certificate rather than token because that is what the base install gives the
other components — the token path is the opt-in `ATE_ATEAPI_CLIENT_AUTH=token`
overlay — and because it yields a SPIFFE ID rather than an opaque JWT subject
for a future authz layer to key on.

One asymmetry worth knowing about: Python's gRPC has no per-handshake reload
hook, so it reads the bundle once per channel. Each locust `User` builds its own
channel in `on_start`, so a respawn picks up a rotated certificate, and an
established connection is unaffected by the old one expiring.

## Verification

Deployed to a GKE cluster running `main`. Before: every call rejected, boomer
retrying `on_start` in a loop. After: every call authenticated as
`spiffe://cluster.local/ns/benchmarking/sa/default`, and no `Unauthenticated`
in the api-server log.

The SPIFFE ID is the part that matters. Only `mtlsPeerIdentity` logs
`Authentication successful`; the JWT branch is silent. So an absence of errors
alone would not distinguish the certificate being accepted from a token falling
through behind it — seeing the SPIFFE ID does.

### DIFF
--- a/benchmarking/locust/common/ateapi_channel.py
+++ b/benchmarking/locust/common/ateapi_channel.py
@@ -1,0 +1,89 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Authenticated gRPC channel to ateapi, shared by every locust user class.
+
+ateapi rejects calls that carry no credential: its interceptor takes identity
+from an mTLS client certificate, or failing that from an `authorization:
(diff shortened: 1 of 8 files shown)


maxsmythe (approved):
LGTM

PR #559, opened 2026-07-28
Adding atunnel - networking component within ateom

LiorLieberman:
[Will rebase after its lgtm] 

Introducing atunnel, a component that sits within ateom and handles all inbound and outbound traffic to the actor. With this change in place there is no direct traffic outside and inside the actor with the exception of DNS and non-TCP which traffic which is still TBD.

this PR; 

- Introduces atunnel, on ingress, it terminates tls and forward to the actor on port 80 (We'll have separate discussions on supporting multiple ports). On egress, it initiates CONNECT over mtls.
- Currently atunnel is using the pod/worker identity for mtls, which is wrong. We should likely use the actor identity once @ahmedtd work is done on the SessionIdentity (likely to be renamed to something else) to be able to mint certs and JWTs.
- RunRequest/RestoreRequest has actor_version and egress_gateway_address here because ateom and atunnel consume them in the same functions that install the nftables rules. Nothing populates either field in this PR,  so they arrive zero-valued, so the egress client is never built and no REDIRECT rule is installed. A fast follow-up egress gateway PR adds the producer side that configures these fields.

Huge part of it uses and is built-upon @howardjohn 's work on #460 

---
More specific notes:

- New internal/atunnel package — an mTLS ingress server plus a transparently-intercepted egress TCP proxy. Only the ingress half is live in this PR.
- Every worker pod hosts an atunnel ingress server on :443. The server authorizes the caller and checks the Actor is the one currently assigned to that worker.
- atenet-router dials the tunnel instead of the Actor. The dynamic_forward_proxy cluster is replaced by an ORIGINAL_DST cluster driven by an ext_proc-set x-ate-original-dst header, with upstream mTLS and SPIFFE URI SAN validation. :authority is left untouched, so atunnel can authorize on the original Host.
- The pod-IP:80 → actor-veth:80 DNAT is removed. Worker port 80 is no longer an ingress path; the mTLS listener is the only way in.
- The egress proxy, the nftables REDIRECT rule and the two new ateompb fields are all gated on an egress gateway address that nothing populates yet, so Actor egress still uses the existing masquerade, unchanged. A pluggable-egress PR is a fast follow.

- [x] Tests pass


haiyanmeng:
FYI, https://github.com/agent-substrate/substrate/pull/557 maps HTTPS requests to actor port 443 with TLS re-origination- #557

docs/architecture.md
```diff
@@ -338,15 +338,24 @@ A `WorkerPool` selects a **sandbox class** (`spec.sandboxClass`), and each class
...
 
   * **Uniform DNS Mesh**: Substrate provides a location-transparent actor discovery scheme via a global DNS suffix (`<actor-name>.<atespace>.actors.resources.substrate.ate.dev`).
 
-  * **Routing**: The `atenet` router (powered by Envoy and an External Processing server) intercepts traffic destined for the mesh. It extracts the actor name from the `Host` header, queries the Control Plane to determine the actor's current location, and triggers a `ResumeActor` workflow if the session is currently suspended.
+  * **Ingress Routing**: `atenet-router` runs Envoy with an `ext_proc` external
+    processor and accepts HTTP traffic for the Actor DNS suffix. The ext_proc
+    extracts the Actor name and Atespace from the `Host` header and calls the
+    Control Plane to resume the Actor and resolve its current worker assignment.
+
+  * **Worker Tunnel**: After resolving the assignment, `atenet-router` opens an
```
haiyanmeng, line 354: Change this to `Worker Ingress Tunnel`

docs/architecture.md
```diff
@@ -338,15 +338,24 @@ A `WorkerPool` selects a **sandbox class** (`spec.sandboxClass`), and each class
...
+    authenticated TLS tunnel to the worker's `atunnel` listener on port 443.
+    `atunnel`, hosted by `ateom`, forwards the request to the active Actor over
+    its private veth interface. Worker pod port 80 is not a direct Actor ingress
```
haiyanmeng, line 357: `pod` -> `Pod`

docs/architecture.md
```diff
@@ -360,26 +369,29 @@ suspended (UML sequence diagram):
 sequenceDiagram
     actor Client
     participant DNS as atenet DNS
-    participant Router as atenet router
+    participant Gateway as atenet-router
     participant API as ate-api-server
     participant Atelet as atelet
-    participant Ateom as ateom
+    participant Tunnel as ateom / atunnel
+    participant Actor
```
haiyanmeng: This does not render correctly, because Actor is a reserved keyword in Mermaid diagrams. Renaming it works:
```
sequenceDiagram
    actor Client
    participant DNS as atenet DNS
    participant Gateway as atenet-router
    participant API as ate-api-server
    participant Atelet as atelet
    participant Tunnel as ateom / atunnel
    participant A as Actor
    participant Store as snapshot storage

    Client->>DNS: resolve actor DNS name
    DNS-->>Client: ingress gateway address
    Client->>Gateway: HTTP request (Host = actor)
    Gateway->>API: ResumeActor(atespace, actor name)
    API->>Atelet: Restore
    Store-->>Atelet: download snapshot
    Atelet->>Tunnel: RestoreWorkload
    Note over Tunnel: restore sandbox and activate Actor network
    Tunnel-->>Atelet: ready
    Atelet-->>API: worker pod IP
    API-->>Gateway: worker assignment
    Gateway->>Tunnel: mTLS tunnel to worker port 443
    Tunnel->>A: forward request to Actor port
    A-->>Tunnel: response
    Tunnel-->>Gateway: response
    Gateway-->>Client: response
    Note over API,Store: later: an explicit SuspendActor checkpoints back to storage and frees the worker
```

docs/architecture.md
haiyanmeng, line 354: Worthy adding a one-line like `egress redirection through atunnel is implemented but not yet enabled`, which would stop the readers from concluding either that it's live or that it doesn't exist.

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+	// for policy decisions.
+	ActorAtespaceHeader = "X-Ate-Atespace"
+	// ActorNameHeader identifies the actor that opened an egress tunnel.
+	ActorNameHeader = "X-Ate-Actor"
+	// ActorVersionHeader is the Actor resource version observed when the worker
+	// was assigned. Gateways use it as a lower bound on cached Actor metadata.
+	ActorVersionHeader = "X-Ate-Actor-Version"
+)
+
+// ClientConfig configures an egress CONNECT client.
```
howardjohn, line 47: I think we should probably use CONNECT on both directions.

1. This makes it symmetrical. Least important reason but we already have a 'sunk cost' on CONNECT.
2. This more clearly future proofs us to multi-port actors. Because right now we have to send to atunnel port 443 which is then making multiple ports hard. Vs sending to some <ingress CONNECT port> which then can easily forward to any port on the actor.
3. We can pass additional metadata more clearly. We already are ~abusing the same HTTP channel to send actor responses OR atunnel responses (StaleAssignmentHeader). CONNECT lets these be clearly split

LiorLieberman: I agree we should. and I think we will. Was working on getting something that unifies all discussions (see https://github.com/agent-substrate/substrate/pull/296#issuecomment-5086745729 for ref). 

For the purpose of getting something in, I think we should split it but I will try to come up with issue/doc/comment today about solving the multi-port (and DNS with double wildcard).

cmd/atenet/internal/router/cmd.go
```diff
@@ -53,13 +56,15 @@ func NewRouterCmd() *cobra.Command {
 	cmd.Flags().IntVar(&cfg.StatusPort, "status-port", 4040, "Port to serve /statusz on (set <= 0 to disable serving status)")
 	cmd.Flags().DurationVar(&cfg.HealthInterval, "health-interval", 1*time.Second, "Interval for checking health of dependent services")
 	cmd.Flags().IntVar(&cfg.HttpsPort, "port-https", 8443, "TCP port for HTTPS workload traffic entering through the Envoy Router")
-	cmd.Flags().StringVar(&cfg.EnvoyCertPath, "envoy-cert-path", "", "Path to the Envoy certificate file.")
+	cmd.Flags().StringVar(&cfg.EnvoyCertPath, "envoy-cert-path", "", "Path to the Envoy certificate file (if empty, a self-signed cert will be generated for testing)")
+	cmd.Flags().StringVar(&cfg.UpstreamClientCertPath, "upstream-client-cert-path", "/run/podidentity.podcert.ate.dev/credential-bundle.pem", "PEM credential bundle (cert+key) the router presents as the client cert when dialing the actor's atunnel ingress server over mTLS. Empty disables upstream mTLS (legacy plaintext pod-IP:80).")
```
haiyanmeng: Can we use the same flag surffix as the `--atunnel-credential-bundle` flag of ateom-gvisor|microvm? Consistency matters more than exactly which one you pick.

cmd/atenet/internal/router/cmd.go
```diff
@@ -53,13 +56,15 @@ func NewRouterCmd() *cobra.Command {
...
+	cmd.Flags().StringVar(&cfg.UpstreamTrustPath, "upstream-trust-bundle-path", "/run/podidentity.podcert.ate.dev/trust-bundle.pem", "PEM trust bundle used to validate the actor's atunnel ingress server certificate.")
```
haiyanmeng: Can we use the same flag surffix as the `--atunnel-trust-bundle` flag of ateom-gvisor|microvm? Consistency matters more than exactly which one you pick.

docs/threat-model.md
LiorLieberman, line 1: Review that before we merge - this is ai generated fully

docs/architecture.md
LiorLieberman: need to take a whole look at the diagram before we merge. It will be fixed

cmd/atenet/internal/router/config.go
```diff
@@ -48,8 +48,17 @@ type routerConfig struct {
...
-	MetricsAddr    string
+	// UpstreamClientCertPath is the router's podidentity credential bundle
+	// (cert+key) presented as the client cert when dialing the actor's atunnel
+	// ingress server over mTLS. UpstreamTrustPath is the CA bundle used to
+	// validate that server. Empty UpstreamClientCertPath disables upstream mTLS.
+	UpstreamClientCertPath string
+	UpstreamTrustPath      string
+	// UpstreamSpiffePrefix validates the actor's atunnel server cert by its
+	// SPIFFE URI SAN prefix (trust domain) instead of the dialed pod IP.
+	UpstreamSpiffePrefix string
```
bowei, line 61: nit: put a new line to separate this in its own block

cmd/atenet/internal/router/extproc.go
```diff
@@ -170,14 +170,22 @@ func (s *ExtProcServer) handleRequestHeaders(
 			"actor %s routing failed", actorRef)
 	}
 
+	// The actor is reached through the in-worker atunnel ingress server, which
+	// listens on :443 (mTLS) and forwards to the actor's :80. The worker no
```
bowei, line 188: I'm thinking we should use not 443, but own port just in case we need the standard HTTPS port later.

cmd/ateom-gvisor/main.go
```diff
@@ -51,11 +58,30 @@ import (
 var (
 	podUID = pflag.String("pod-uid", "", "The UID of the current pod")
 
+	atunnelListenAddress       = pflag.String("atunnel-listen-address", "0.0.0.0:443", "Address for actor ingress HTTPS")
```
bowei, line 58: consider factoring this out into a common library internal/atunnel/ateom 

that has a common flag registration func etc.

cmd/ateom-gvisor/main.go
```diff
@@ -51,11 +58,30 @@ import (
...
+const (
+	hostVethName      = "ateom0"
+	actorVethName     = "eth0"
+	actorVethTempName = "ateom1"
+	hostVethCIDR      = "169.254.17.1/30"
+	actorVethCIDR     = "169.254.17.2/30"
+	actorVethGateway  = "169.254.17.1"
+	actorVethIP       = "169.254.17.2"
+	actorNftTableName = "ateom_actor"
+	actorHTTPUpstream = "http://169.254.17.2:80"
```
bowei: this should be constructed from existing constants

cmd/ateom-gvisor/main.go
```diff
@@ -147,6 +182,50 @@ func do(ctx context.Context) error {
 	return nil
 }
 
+func runAtunnel(ctx context.Context, upstream *url.URL) (*atunnel.Server, *atunnel.Egress, uint16, error) {
```
bowei, line 181: is this going to be same or dupe for ateom-uvm

cmd/ateom-microvm/checkpoint.go
```diff
@@ -61,6 +59,9 @@ import (
 func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.CheckpointWorkloadRequest) (*ateompb.CheckpointWorkloadResponse, error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
+	if err := s.deactivateActorNetworking(ctx); err != nil {
```
bowei, line 64: is this just a bug that it's missing on the CheckpointWorkload path? Specifically, we should just make this change outside of this change.

cmd/ateom-microvm/main.go
```diff
@@ -139,12 +148,55 @@ func do(ctx context.Context) error {
 	// logWriter with the runtime logger so the two streams to os.Stdout are
 	// serialized through one SyncedWriter and never interleave-corrupt lines.
 	actorLogger := actorlog.NewActorLogger(logWriter, metadata.OnGCE())
+	upstream, err := url.Parse("http://169.254.17.2:80")
```
bowei: Let's see how to make it so this code is not diff between different ateoms

cmd/ateom-microvm/run.go
```diff
@@ -207,6 +208,9 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
 		templateName: req.GetActorTemplateName(),
 		containers:   req.GetSpec().GetContainers(),
 		assetPaths:   req.GetRuntimeAssetPaths(),
+
+		actorVersion:         req.GetActorVersion(),
```
bowei, line 215: actor version or actor UUID

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
```diff
@@ -47,9 +48,18 @@ const (
 	ateletServiceAccount = "atelet"
 )
 
-func extKeyUsages(namespace, serviceAccount string) []x509.ExtKeyUsage {
+// workerPoolLabel marks pods created by atecontroller for a WorkerPool. Worker
+// pods host the atunnel ingress server, presenting this cert as a TLS server
```
bowei, line 52: Can we split this off and get these changes in separately?

docs/threat-model.md
bowei: text changes look ok to me

internal/proto/ateompb/ateom.proto
```diff
@@ -64,6 +66,10 @@ message RunWorkloadRequest {
   // to the local on-disk path atelet fetched it to (content-addressed, like
   // runsc_path). Empty for the gVisor runtime, which uses runsc_path.
   map<string, string> runtime_asset_paths = 8;
+
+  // Remote egress gateway selected for this activation. When absent, actor
+  // traffic uses direct egress instead of being redirected through atunnel.
+  optional string egress_gateway_address = 10;
```
bowei, line 72:
```
oneof egress {
  string via_gateway = 10; 
}
```

bowei, line 72: see comment about making this oneof

manifests/ate-install/atenet-router.yaml
```diff
@@ -211,6 +220,12 @@ spec:
               signerName: podidentity.podcert.ate.dev/identity
               keyType: ECDSAP256
               credentialBundlePath: credential-bundle.pem
+          - clusterTrustBundle:
+              signerName: podidentity.podcert.ate.dev/identity
+              labelSelector:
+                matchLabels:
+                  podcert.ate.dev/canarying: live
```
LiorLieberman, line 226: TODO: not sure if we need this or not. Discuss with other community members

internal/atunnel/original_dst_linux.go
```diff
@@ -0,0 +1,72 @@
...
+	tcpConn, ok := conn.(*net.TCPConn)
+	if !ok {
+		return "", fmt.Errorf("atunnel: original destination requires a TCP connection, got %T", conn)
+	}
+	rawConn, err := tcpConn.SyscallConn()
+	if err != nil {
+		return "", fmt.Errorf("atunnel: acquiring TCP syscall connection: %w", err)
+	}
+
+	var addr unix.RawSockaddrInet4
```
bowei, line 43: please leave a TODO or implement support for IPv6

bowei: I would prefer you add support for IPv6

internal/atunnel/original_dst_linux.go
```diff
@@ -0,0 +1,72 @@
+// Copyright 2026 Google LLC
```
bowei, line 1: doesn't hurt to put this in its own package.

bowei: internal/atunnel/sys

internal/atunnel/client.go
bowei: Probably leaving a TODO saying that would help people understand this is an incremental step.

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/agent-substrate/substrate/internal/resources"
+)
+
+const (
+	// ActorAtespaceHeader identifies the atespace whose actor opened an egress
+	// tunnel. The egress gateway must authenticate this metadata before using it
+	// for policy decisions.
+	ActorAtespaceHeader = "X-Ate-Atespace"
+	// ActorNameHeader identifies the actor that opened an egress tunnel.
+	ActorNameHeader = "X-Ate-Actor"
```
bowei: X-Ate-Actor-Name

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+
+// ClientConfig configures an egress CONNECT client.
+type ClientConfig struct {
+	GatewayAddress       string
+	ServerName           string
+	CredentialBundlePath string
+	TrustBundlePath      string
+
+	// DialContext is injectable for tests. When nil, a net.Dialer is used.
+	DialContext func(context.Context, string, string) (net.Conn, error)
```
bowei: It's a little gross to mix the func in with a plain 'ol data struct

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+// optional until actor JWT issuance is wired into ateom.
+type EgressMetadata struct {
+	Atespace     string
+	ActorName    string
+	ActorVersion int64
+	BearerToken  string
+}
+
+// Client opens actor egress streams through an mTLS-authenticated gateway.
+type Client struct {
```
bowei, line 82: nit: would be good to make this more consistent, calling everything egress related egress vs a very generic sounding `Client`.

internal/atunnel/client.go
```diff
@@ -0,0 +1,213 @@
...
+	gatewayAddress string
+	tlsConfig      *tls.Config
+	dialContext    func(context.Context, string, string) (net.Conn, error)
+}
+
+var _ EgressDialer = (*Client)(nil)
```
bowei, line 89: You should leave a comment that this is a interface check

// EgressDialer implements Client.

internal/proto/ateompb/ateom.proto
LiorLieberman: why??

internal/atunnel/egress.go
```diff
@@ -0,0 +1,207 @@
...
+		dialer: dialer,
+		ctx:    ctx,
+		cancel: cancel,
+	}
+	return nil
+}
+
+// Deactivate rejects new egress, closes active streams, and waits for their
+// forwarding goroutines to exit.
+func (e *Egress) Deactivate(ctx context.Context) error {
```
bowei, line 125: I would prefer if per-Actor state management involved just throwing away the whole object graph and closing connection, but we can take that up in a follow PR.

cmd/atenet/internal/router/extproc.go
LiorLieberman: Discussed offline but commeting for posterity. This port is standard https. 

When we introduce CONNECT on Ingress (client -> atenet/pluggable ingress) - we may also have connect to atunnel, and in that case, we can consider having that a 8443 or any other special port.

cmd/ateom-microvm/checkpoint.go
LiorLieberman: deactivateActorNetworking is a new function this PR added. Ideally we are trying to stop admitting traffic and drain active streams before the Actor network is torn down. Since (1) we introduced atunnel and it takes care of all networking connections for the actor and (2) CheckpointWorkload suspends the actor. I think the best home for this addition is within this PR.

cmd/ateom-microvm/main.go
LiorLieberman: I think we should have seprate task to have dedup the logic between ateoms. For adressing this comment - I added the same const.

cmd/ateom-microvm/run.go
LiorLieberman: actor version, discussed offline.

internal/atunnel/client.go
LiorLieberman: added

internal/atunnel/original_dst_linux.go
LiorLieberman: TODO is already there, we can take this as a follow up to parallelize work.

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
LiorLieberman: discussed offline, should be with this change as atunnel needs it to server tls on ingress until we replace with "real actor identitie"

internal/atunnel/egress.go
LiorLieberman: Discussed offline given that only one actor can be active at any given worker, it does not really matter. 

We deconstruct the state with `e.active = nil` so no risk that one actor connections leak to another

internal/proto/ateompb/ateom.proto
bowei: We may want to add more ways than just ip. Also is it ip or DNS address or both

internal/atunnel/client.go
LiorLieberman: good point. fixed

internal/atunnel/client.go
LiorLieberman: I originally did it and then reverted, there is nothing egress-only in client. Client implements EgressDialer though.. Suggest we leave it this way

cmd/ateom-gvisor/main.go
LiorLieberman: will move it as part of a atunnel repackaging in a follow up PR

cmd/ateom-gvisor/main.go
LiorLieberman: added a comment, will be part of a follow up

internal/atunnel/server.go
```diff
@@ -0,0 +1,284 @@
...
+		credentialBundlePath: cfg.CredentialBundlePath,
+		proxy:                proxy,
+	}
+	s.tlsConfig = &tls.Config{
+		MinVersion: tls.VersionTLS12,
+		GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+			return loadCredentialBundle(s.credentialBundlePath)
+		},
+		ClientAuth: tls.RequireAndVerifyClientCert,
+		ClientCAs:  clientCAs,
```
BenTheElder: 🤖 **should-fix** 🟡 – The client CA pool is frozen at process start while the leaf beside it reloads per connection. `GetCertificate` above re-reads the credential bundle on every handshake, with a comment explaining that this is what lets kubelet's rotation take effect — but `clientCAs` comes from a single `os.ReadFile` in `NewServer`.

Both files live in the same projected volume: `workerpool_apply.go` gives it a `podCertificate` source for `credential-bundle.pem` and a `clusterTrustBundle` source for `trust-bundle.pem`. A `ClusterTrustBundle` projection exists precisely so kubelet can push CA changes into a running pod, and ateom outlives many activations, so after a signer CA rotation this server keeps verifying against the old pool and rejects the router until the worker pod restarts.

`GetConfigForClient` is the usual fix: return a `*tls.Config` per connection with a freshly built `ClientCAs`, mirroring what `GetCertificate` already does for the leaf. `client.go` reads its `RootCAs` once too, though `NewClient` runs per activation, so the window there is much shorter.

BenTheElder (commented):
One Plausible AI review finding.

internal/atunnel/server.go
bowei: Lior can you take a TODO to follow up on this

LiorLieberman: added a todo.. do we have rotation today?

bowei: yes, we should have it...otherwise substrate will stop working after 1 - 2 days due to the certs having very quick expiry.

LiorLieberman: sg, added a todo and will open an issue for it as well. Rebased now, hopefully all should be good to go

docs/architecture.md
LiorLieberman: applied your suggestion.Thanks

PR #560, opened 2026-07-28
rename ErrPersistenceRetry to ErrVersionConflict to reflect the actual behavior

HavenXia:
The error is only returned on an optimistic-concurrency version mismatch. Rename it for that behavior so other retryable errors don't get folded into it (but should introduce their own error name).

Followup for post-merge comment in #535 

- [x] Tests pass


EItanya (approved):
LGTM

PR #458, opened 2026-07-17
Make kubectl-ate log  and worker output filtered by atespace

HavenXia:
The current implementation still only use actor name to filter, which could leads to interleaved information from actors with same name in different atespaces.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

PR #561, opened 2026-07-28
docs: fix kubectl ate logs examples and flag reference

botengyao:
Doc-only fix, no issue filed — happy to open one if preferred.

`-a/--atespace` became a required flag on `kubectl ate logs actors` in #458, but the examples in the CLI README and `docs/observability.md` were never updated. Every documented invocation currently fails:

```
$ go run ./cmd/kubectl-ate logs actors my-actor
Error: required flag(s) "atespace" not set
```

While in those files, two smaller inaccuracies in `cmd/kubectl-ate/README.md`:

- The logs example says it "follows by default", but `--follow` defaults to `false`. Split into a one-shot and a `-f` example.
- The global flags table was missing `--context`.

- [x] Tests pass

### DIFF
--- a/cmd/kubectl-ate/README.md
+++ b/cmd/kubectl-ate/README.md
@@ -64,6 +64,7 @@ These flags can be appended to any command:
 | Flag | Short | Description | Default |
 |---|---|---|---|
 | `--kubeconfig` | | Path to your kubeconfig file | `~/.kube/config` |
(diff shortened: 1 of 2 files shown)


HavenXia:
Thanks! I forgot to update those.

HavenXia (approved):

Issue #550, opened 2026-07-27
Feature Request: Resource-usage telemetry at actor / ActorTemplate granularity


git286:
## Problem

Substrate cannot answer *"how much CPU and memory is this actor using?"* — nor the same question rolled up per ActorTemplate.

Resource usage is observable only at worker-pod granularity, via `kubectl ate top workers`. actor is the unit users create, name, bill for, debug, and reason about — and it is invisible.

## Why this matters

Concretely, four groups are blocked:

| Who | What they cannot do today |
|---|---|
| **Agent developers** | Diagnose why an actor was OOM-killed. Size an ActorTemplate's resource requests with anything but guesswork. Tell which of their actors is the expensive one. |
| **Platform operators** | Detect noisy neighbours. Attribute cost or chargeback to a tenant, when atespace *is* the tenant boundary. Set quotas informed by observed usage. |
| **Substrate developers** | Validate the density claim, or tune multiplexing against real numbers. #44 (concurrent actors per worker, P0) has no way to measure whether it succeeded. |
| **Load testing** | Explain bottlenecks or confirm fleet health under load — the original motivation for #174, and still open in #168 / #176. |

This also gates scheduling work. Demand-reactive capacity (#198) and suspend-on-inactivity (#424) both make better decisions with a real actor resource footprint than without one.

---

## Related

- #174 Observability At-Scale — parent discussion, granularity taxonomy, cardinality decisions
- #433 Initial set of platform metrics — the metrics workstream this complements
- #288 All actors are under the `pause` cgroup — **closed**, was the stated blocker
- #496 ateom gvisor: drop `privileged: true` — delivered the per-actor cgroup leaves
- #161 fix(atelet): set OCI CgroupsPath — open, superseded by #496
- #515 / #516 `kubectl ate top workers` — the pod-level precedent
- #44 High Concurrent Active Actors per Worker — will invalidate pod-as-proxy-for-actor
- #198 WorkerPool autoscaling / #424 suspend on inactivity — consumers of actor usage data
- #503 OpenTelemetry data loss across actor lifecycle events — adjacent, application-side

krisztianfekete:
+1 for driving this!
Most important question to officially decide I think is to what signals to use for what purpose.

[...]

PR #562, opened 2026-07-28
(otel): onboard ateom to the OTLP metrics path

krisztianfekete:
This PR wires `ateom` into the OTLP push path so it can emit metrics, and unblock unblocking https://github.com/agent-substrate/substrate/issues/550

- serverboot: new InitMetricsPushOnly (OTLP periodic reader, no Prometheus pull surface) for binaries that run no metrics HTTP server
- ateom-gvisor + ateom-microvm call it, so the existing otelgrpc server handler now feeds `rpc.server.*` to the collector
- atecontroller propagates its `OTEL_EXPORTER_OTLP_ENDPOINT` + `OTEL_RESOURCE_ATTRIBUTES` into the ateom worker pods
- fixes ateom telemetry not reaching the collector 
- e2e: metrics suite asserts an `ateom` service reaches the collector

cc. @git286 

- [x] Tests pass

### DIFF
--- a/internal/serverboot/serverboot.go
+++ b/internal/serverboot/serverboot.go
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 // Package serverboot collects the startup boilerplate shared by the
-// long-running substrate server binaries (ateapi, atelet, ateom-gvisor):
-// slog wiring, OTel tracer + meter providers, a Prometheus + /readyz
-// HTTP surface, and a couple of small helpers for startup fail-fast.
+// long-running substrate server binaries (ateapi, atelet, ateom-gvisor,
+// ateom-microvm): slog wiring, OTel tracer + meter providers, a Prometheus +
+// /readyz HTTP surface, and a couple of small helpers for startup fail-fast.
 package serverboot
 
 import (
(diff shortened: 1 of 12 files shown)


internal/serverboot/serverboot.go
```diff
@@ -123,13 +123,24 @@ func InitTracing(ctx context.Context, opts TracingOptions) (*sdktrace.TracerProv
 // reader (exposed via StartMetricsServer's /metrics handler) and an
 // OTLP periodic reader.
 func InitMetrics(ctx context.Context, serviceName string) (*sdkmetric.MeterProvider, error) {
-	if serviceName == "" {
-		return nil, fmt.Errorf("serviceName is required")
-	}
 	promExporter, err := prometheus.New()
```
baizhenyu, line 129: Minor: the `serviceName == ""` guard moved into `newMeterProvider`, so it now runs after `prometheus.New()`.

`New()` registers a collector into `prometheus.DefaultRegisterer` as a side effect of construction (`config.go:46-47` defaults the registerer, `exporter.go:143` registers) — that's what lets `StartMetricsServer`'s `promhttp.Handler()` serve it. Nothing unregisters it on the error path, so `InitMetrics(ctx, "")` returns the right error but strands a live collector, and the next `InitMetrics` call fails with a duplicate-registration error pointing nowhere near the actual cause.

Not reachable in production — all four callers pass non-empty constants, and each exits the process on failure, so there's no second call to collide with. It bites in tests, where the whole package shares one process: the symmetric `TestInitMetricsRequiresServiceName` to go with your `TestInitMetricsPushOnlyRequiresServiceName` would break every later test in the file.

```go
func InitMetrics(ctx context.Context, serviceName string) (*sdkmetric.MeterProvider, error) {
	// Check before prometheus.New(), which registers a collector into the
	// default registry and cannot be undone if construction later fails.
	if serviceName == "" {
		return nil, fmt.Errorf("serviceName is required")
	}
	promExporter, err := prometheus.New()
	...
```

Worth keeping the copy in `newMeterProvider` as well — that one is the invariant covering every construction path, including `InitMetricsPushOnly` and anything added later. This one is just a precondition on the side effect.

krisztianfekete: Fair, addressed this just now.

baizhenyu (approved):

Issue #529, opened 2026-07-24
Add first-class ActorSnapshot lifecycle APIs


EItanya:
## Summary

Substrate creates a durable checkpoint whenever it suspends an Actor. Make each successful checkpoint a first-class, immutable `ActorSnapshot` that callers can get, list, tag, and use to initialize another Actor.

The design changed during implementation. Superseded decisions remain crossed out so the history is visible:

- ~~A caller explicitly promotes a suspended Actor's private checkpoint.~~ Every successful `SuspendActor` automatically creates an `ActorSnapshot` and exposes it through `Actor.latest_snapshot`.
- ~~Tags are globally unique aliases.~~ Tags are owned and addressed by Atespace as `atespace/name`; the same name may exist in different Atespaces.
- ~~Visibility is mutable state on the snapshot.~~ Reuse policy belongs to each tag: `ATESPACE` permits local reuse and `PUBLISHED` permits reuse from any Atespace without changing the tag's address.
- ~~Tags are embedded in `ActorSnapshot`.~~ `ActorSnapshotTag` is a separate resource and O(1) index entry.
- ~~Deleting an Atespace cascades to its tags.~~ Atespace deletion is blocked while any Actors or ActorSnapshotTags remain, matching existing non-empty Atespace behavior.
- ~~Expose explicit snapshot deletion and protect tagged snapshots from it.~~ Explicit snapshot deletion and automatic retention are deferred until the GC policy is designed.
- ~~`CreateActor(source_snapshot)` accepts a canonical snapshot reference or a tag.~~ Explicit reuse accepts only a tag reference, ensuring reuse always has an explicit policy.

## Motivation

Control planes need to preserve an Actor at a known state and later create an independent Actor from it. Substrate must own the storage layout, compatibility checks, authorization boundary, and stable references rather than requiring callers to copy private storage fields.

## Proposed API

Snapshots remain immutable resources:

```protobuf
message ActorSnapshot {
  ResourceMetadata metadata = 1;
  ObjectRef source_actor = 2;
  string source_actor_uid = 3;
  int64 source_actor_version = 4;
  string actor_template_namespace = 5;
  string actor_template_name = 6;
  string actor_template_uid = 7;
  SnapshotContentScope content_scope = 8;
}
```

Tags are Atespace-owned aliases and retention pins. `ATESPACE` is intentionally the zero/default scope:

```protobuf
enum ActorSnapshotTagScope {
[...]

PR #568, opened 2026-07-28
ateapi: store snapshots under stable UUID paths

EItanya:
## Summary

- store snapshots under `<configured-root>/snapshots/<uuid>`
- avoid encoding actor or atespace lifecycle in the GCS object prefix
- validate the snapshot prefix and UUID format

This is a small storage-layout prerequisite for #529. Snapshot metadata, tagging, and retention remain separate follow-up work.

## Testing

- `make test`
- `env -u NO_COLOR make verify`
- Kind E2E, token auth: gVisor suite and micro-VM demo suite
- Kind E2E, certificate auth: gVisor suite and micro-VM demo suite

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_suspend.go
+++ b/cmd/ateapi/internal/controlapi/workflow_suspend.go
@@ -94,7 +94,7 @@ func (s *MarkSuspendingStep) CheckPrerequisite(ctx context.Context, input *Suspe
 func (s *MarkSuspendingStep) Execute(ctx context.Context, input *SuspendInput, state *SuspendState) error {
 	state.Actor.Status = ateapipb.Actor_STATUS_SUSPENDING
 	snapshotID := time.Now().Format(time.RFC3339) + "-" + rand.Text()
-	state.Actor.InProgressSnapshot = strings.TrimSuffix(state.ActorTemplate.Spec.SnapshotsConfig.Location, "/") + "/" + input.ActorRef.Name + "/" + snapshotID
+	state.Actor.InProgressSnapshot = strings.TrimSuffix(state.ActorTemplate.Spec.SnapshotsConfig.Location, "/") + "/snapshots/" + snapshotID
 	updatedActor, err := s.store.UpdateActor(ctx, state.Actor, state.Actor.GetMetadata().GetVersion())
 	if err != nil {
 		return err
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -1553,7 +1553,7 @@ func TestSuspendActor(t *testing.T) {
 		LatestSnapshotInfo: &ateapipb.SnapshotInfo{
 			Data: &ateapipb.SnapshotInfo_External{
 				External: &ateapipb.ExternalSnapshotInfo{
-					SnapshotUriPrefix: fmt.Sprintf("gs://fake-fake-fake/%s/", name),
+					SnapshotUriPrefix: "gs://fake-fake-fake/snapshots/",
 				},
 			},
 		},
--- a/cmd/ateapi/internal/controlapi/workflow_suspend_test.go
+++ b/cmd/ateapi/internal/controlapi/workflow_suspend_test.go
@@ -16,12 +16,14 @@ package controlapi
 
 import (
 	"context"
+	"strings"
 	"testing"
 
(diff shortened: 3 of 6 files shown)


cmd/ateapi/internal/controlapi/workflow_suspend.go
```diff
@@ -94,7 +94,7 @@ func (s *MarkSuspendingStep) CheckPrerequisite(ctx context.Context, input *Suspe
 func (s *MarkSuspendingStep) Execute(ctx context.Context, input *SuspendInput, state *SuspendState) error {
 	state.Actor.Status = ateapipb.Actor_STATUS_SUSPENDING
 	snapshotID := time.Now().Format(time.RFC3339) + "-" + rand.Text()
-	state.Actor.InProgressSnapshot = strings.TrimSuffix(state.ActorTemplate.Spec.SnapshotsConfig.Location, "/") + "/" + input.ActorName + "/" + snapshotID
+	state.Actor.InProgressSnapshot = strings.TrimSuffix(state.ActorTemplate.Spec.SnapshotsConfig.Location, "/") + "/snapshots/" + snapshotID
```
HavenXia, line 97: Why removing atespace while #529 designs to keep it?

cmd/ateapi/internal/controlapi/workflow_suspend.go
```diff
@@ -94,7 +94,7 @@ func (s *MarkSuspendingStep) CheckPrerequisite(ctx context.Context, input *Suspe
 func (s *MarkSuspendingStep) Execute(ctx context.Context, input *SuspendInput, state *SuspendState) error {
 	state.Actor.Status = ateapipb.Actor_STATUS_SUSPENDING
 	snapshotID := time.Now().Format(time.RFC3339) + "-" + rand.Text()
```
HavenXia, line 96: This isn't technically a UUID right while PR description says that

cmd/ateapi/internal/controlapi/workflow_suspend.go
HavenXia: resolved, #529 clarified not storing atespace in url anymore.

cmd/ateapi/internal/controlapi/workflow_suspend.go
EItanya: I wanted to keep the scope of this change as small as possible, and that's the current storage ID. Happy to change it. The requirement is a Unique ID, this is also that

HavenXia (dismissed):

dberkov:
### Concern:
This concept introduces a tight coupling between the snapshot layer and the persistence layer. If an issue occurs with the persistence layer, data cannot be recreated. Relying on backups is insufficient because, between the most recent backup and the failure point, we cannot determine which snapshot belongs to which actor.

### Proposals:
Option 1: Keep the directory layout flat as proposed in this PR, but store an additional metadata file inside the snapshot folder indicating its associated actor.
Option 2: Structure the directory hierarchy, for example: /snapshots/atespaces/ATE_SPACE/actorTemplates/ACTOR_TEMPLATE/ACTOR_ID/SNAPSHOT_ID (or something similar).
Option 3: Combine Options 1 & 2.

EItanya:
> ### Concern:
> This concept introduces a tight coupling between the snapshot layer and the persistence layer. If an issue occurs with the persistence layer, data cannot be recreated. Relying on backups is insufficient because, between the most recent backup and the failure point, we cannot determine which snapshot belongs to which actor.
> 
> ### Proposals:
> Option 1: Keep the directory layout flat as proposed in this PR, but store an additional metadata file inside the snapshot folder indicating its associated actor. Option 2: Structure the directory hierarchy, for example: /snapshots/atespaces/ATE_SPACE/actorTemplates/ACTOR_TEMPLATE/ACTOR_ID/SNAPSHOT_ID (or something similar). Option 3: Combine Options 1 & 2.

I'm happy to add the metadata file, but at a higher level is persistence rebuilding a goal of the project. From everything I've seen so far data loss is essentially existential for the system.

EItanya:
Added the metadata mapping while keeping the storage layout flat. Each snapshot already contains a `manifest.json`, so instead of adding another file or directory hierarchy, the existing manifest now records `atespace`, `actorName`, `actorUid`, `actorTemplateNamespace`, and `actorTemplateName` alongside the snapshot file list and sandbox metadata. This makes the snapshot independently attributable to its actor without coupling identity to the object path.

HavenXia (approved):

PR #576, opened 2026-07-28
Add autoscaled-workerpool demo using HPA and prometheus-adapter on Kind

laoj2:
This introduces a new demo (`demos/autoscaled-workerpool`) demonstrating how to dynamically autoscale a WorkerPool using HPA + `prometheus-adapter` on a local Kind cluster.

As mentioned in https://github.com/agent-substrate/substrate/issues/198, this approach may be too slow for some use cases, but this is a good first milestone. 

Plus, there are some limitations with scaling workerpools that still need to be addressed, for example:

- Scale-down can strand paused actors. When a worker pod is removed, a local-snapshot paused actor keeps its node pin pointing at a node that may no longer have (free) worker pods.
- Scale-up gives no locality guarantee: If there are multiple local-snapshot paused actors that can't resume because no worker pods are available on their node, upscaling will not unblock them: the added capacity can land anywhere, so the pool can grow without ever placing a worker where a stranded actor needs it.
- Scale-down picks victims blind to actor assignment (can kill busy pods).

- [x] Tests pass: n/a

### DIFF
--- a/demos/autoscaled-workerpool/README.md
+++ b/demos/autoscaled-workerpool/README.md
@@ -1,0 +1,174 @@
+<!--
+Copyright 2026 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


juli4n:
Needs rebase.

laoj2:
> Needs rebase.

Done

juli4n (approved):

PR #578, opened 2026-07-29
ateapi/controlapi: retry worker updates on version conflict in syncer

troychiu:
This PR adds retry logic with exponential backoff when `WorkerPoolSyncer` encounters optimistic locking conflicts (`store.ErrVersionConflict`) while updating worker state in the store.

- **Before**: `syncWorkerToStore` attempted a single `UpdateWorker` write. If a concurrent write modified the worker in Redis, it failed with `store.ErrVersionConflict`, logged an error, and silently dropped the metadata update (`SandboxClass`, `Labels`).
- **Now**: On `store.ErrVersionConflict`, `WorkerPoolSyncer` retries up to 5 times with exponential backoff, re-fetching the latest worker state from Redis on each attempt before re-applying changes.

- [X] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/syncer.go
+++ b/cmd/ateapi/internal/controlapi/syncer.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"log/slog"
 	"maps"
+	"time"
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
 	"github.com/agent-substrate/substrate/internal/resources"
@@ -27,6 +28,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -156,28 +158,68 @@ func (s *WorkerPoolSyncer) syncWorkerToStore(ctx context.Context, pod *corev1.Po
...
-	if w.SandboxClass != string(pool.Spec.SandboxClass) {
-		slog.InfoContext(ctx, "Syncer: updating worker in store (SandboxClass changed)", slog.String("worker", pod.Namespace+"/"+pod.Name))
-		w.SandboxClass = string(pool.Spec.SandboxClass)
-		changed = true
-	}
-	if !maps.Equal(w.Labels, pool.GetLabels()) {
-		slog.InfoContext(ctx, "Syncer: updating worker in store (labels changed)", slog.String("worker", pod.Namespace+"/"+pod.Name))
-		w.Labels = pool.GetLabels()
-		changed = true
+	// TODO: We perform an inline retry loop here using wait.ExponentialBackoff
```
EItanya, line 161: Should we just make that change now? How large would it be? IIRc The `ActorTemplate` controller already uses requeue

troychiu: Yes `ActorTemplateController` is built using `controller-runtime`, which automatically provides a workqueue and handles retries when `Reconcile()` returns an error or `RequeueAfter`.

The refactor appears moderate, involving approximately 80–100 lines of changes in `syncer.go` to implement a `PodLister`, workqueue lifecycle, and worker goroutine loop.

We could keep this PR focused on the small inline backoff fix to handle version conflicts immediately, and I can create a separate PR for the refactor.

What do you think?

EItanya: Given that most of this will be replaced by that change I’d vote to just close this in favor of that, or just change this PR

troychiu: Sounds good. I can repurpose this PR. Thanks for the feedback.

troychiu: I created https://github.com/agent-substrate/substrate/pull/674 instead.

troychiu:
Closed this PR since we are going for https://github.com/agent-substrate/substrate/pull/674

PR #582, opened 2026-07-29
Implement endpoint slice resolution for ate api connections for HA

### LINKED ISSUE #565: Consider replacing DNS-based round robin + forced reconnects with a k8s-aware gRPC resolver for ateapi clients
Today clients load balance across ate-apiserver replicas using a headless k8s service. K8s creates A records per ready pods. Clients dial with `dns:///` and use `round_robin` load-balancing policy.

When a Pod goes away gRPC re-resolves DNS, but when a new pod is added nothing triggers a re-resolve. Our current mitigation forces every connection to close after a given period of time so clients are forced to re-resolve. It works, but feels pretty hacky.

A probably better approach is to build a k8s-aware gRPC resolver (ala [github.com/sercand/kuberesolver](url)) that watches `EndpointSlices` for a given `Service` and handles `Pod` updates.

michaelasp:
Fixes #565

Replaces standard DNS resolution (`dns:///`) and the forced 1-minute `MaxConnectionAge` reconnect workaround with a native, event-driven Kubernetes gRPC resolver (`internal/k8sresolver`). The resolver uses client-go's `SharedInformerFactory` and `EndpointSliceLister` to watch `discoveryv1.EndpointSlice` resources in real time.

### Verification
Created a kind cluster to test changes by scaling the deployments up and down and verified the new servers were connected to by clients.

#### 1. Kind Cluster Scale-Out & EndpointSlice Discovery
  Click to view kubectl scale and EndpointSlice output
    $ kubectl scale deployment/ate-api-server -n ate-system --replicas=4
    deployment.apps/ate-api-server scaled

    $ kubectl get pods -n ate-system -l app=ate-api-server
    NAME                              READY   STATUS    RESTARTS   AGE
    ate-api-server-6bcd9fddbb-gxxth   1/1     Running   0          3m8s
    ate-api-server-6bcd9fddbb-jgwbw   1/1     Running   0          3m8s
    ate-api-server-6bcd9fddbb-m8nbx   1/1     Running   0          7m20s
    ate-api-server-6bcd9fddbb-vwm2g   1/1     Running   0          7m20s

    $ kubectl get endpointslices -n ate-system -l kubernetes.io/service-name=api
    NAME        ADDRESSTYPE   PORTS   ENDPOINTS
    api-dmzfc   IPv4          443     10.244.0.15,10.244.0.19,10.244.0.31,10.244.0.32

#### 2. Proof of Active gRPC Load-Balancing Across All Replicas
```
    [pod/ate-api-server-6bcd9fddbb-gxxth] {"time":"2026-07-29T13:20:49Z","level":"INFO","msg":"Handle RPC","method":"/ateapi.Control/ListActors","principal":{"ID":"spiffe://cluster.local/ns/ate-system/sa/atenet-router"}}
    [pod/ate-api-server-6bcd9fddbb-jgwbw] {"time":"2026-07-29T13:20:46Z","level":"INFO","msg":"Handle RPC","method":"/ateapi.Control/ListActors","principal":{"ID":"spiffe://cluster.local/ns/ate-system/sa/atenet-router"}}
    [pod/ate-api-server-6bcd9fddbb-m8nbx] {"time":"2026-07-29T13:20:48Z","level":"INFO","msg":"Handle RPC","method":"/ateapi.Control/ListActors","principal":{"ID":"spiffe://cluster.local/ns/ate-system/sa/atenet-router"}}
    [pod/ate-api-server-6bcd9fddbb-vwm2g] {"time":"2026-07-29T13:20:47Z","level":"INFO","msg":"Handle RPC","method":"/ateapi.Control/ListActors","principal":{"ID":"spiffe://cluster.local/ns/ate-system/sa/atenet-router"}}
```

### DIFF
--- a/cmd/atecontroller/internal/controllers/actortemplate_controller.go
+++ b/cmd/atecontroller/internal/controllers/actortemplate_controller.go
@@ -59,6 +59,7 @@ type ActorTemplateReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups=discovery.k8s.io,resources=endpointslices,verbs=get;list;watch,namespace=ate-system
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
--- a/internal/k8sresolver/resolver.go
+++ b/internal/k8sresolver/resolver.go
@@ -1,0 +1,214 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8sresolver
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"net"
+	"slices"
+	"strconv"
+	"strings"
+
+	"google.golang.org/grpc/resolver"
+	v1 "k8s.io/api/discovery/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	discoverylisters "k8s.io/client-go/listers/discovery/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+const Scheme = "k8s"
+
+// Builder implements resolver.Builder for Kubernetes EndpointSlices backed by client-go Informers.
+type Builder struct{ client kubernetes.Interface }
+
+// NewBuilder creates a new gRPC resolver Builder for EndpointSlices using client-go Informers.
+func NewBuilder(client kubernetes.Interface) *Builder { return &Builder{client: client} }
+func (b *Builder) Scheme() string                     { return Scheme }
+
+func (b *Builder) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (resolver.Resolver, error) {
+	if b == nil || b.client == nil {
+		return nil, fmt.Errorf("k8sresolver: kubernetes client is required")
+	}
+
+	ns, svc, port, err := ParseTarget(target)
+	if err != nil {
+		return nil, fmt.Errorf("k8sresolver: %w", err)
+	}
+
+	cli := b.client
+
+	selector := labels.Set{"kubernetes.io/service-name": svc}.AsSelector()
+	factory := informers.NewSharedInformerFactoryWithOptions(cli, 0,
+		informers.WithNamespace(ns),
+		informers.WithTweakListOptions(func(o *metav1.ListOptions) { o.LabelSelector = selector.String() }),
+	)
+
+	inf := factory.Discovery().V1().EndpointSlices()
+	ctx, cancel := context.WithCancel(context.Background())
+	r := &k8sResolver{
+		ctx:      ctx,
+		cancel:   cancel,
+		service:  svc,
+		port:     port,
+		cc:       cc,
+		informer: inf.Informer(),
+		lister:   inf.Lister().EndpointSlices(ns),
+		selector: selector,
+	}
+
+	_ = inf.Informer().SetWatchErrorHandler(func(_ *cache.Reflector, err error) {
+		if ctx.Err() == nil {
+			slog.Error("k8sresolver watch error", slog.String("namespace", ns), slog.String("service", svc), slog.Any("error", err))
+			cc.ReportError(fmt.Errorf("k8sresolver: watch error for %s/%s: %w", ns, svc, err))
+		}
+	})
+
+	reg, err := inf.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(_ any) { r.updateState() },
+		UpdateFunc: func(_, _ any) { r.updateState() },
+		DeleteFunc: func(_ any) { r.updateState() },
+	})
+	if err != nil {
+		return nil, fmt.Errorf("k8sresolver: failed to add event handler for %s/%s: %w", ns, svc, err)
+	}
+
+	r.registration = reg
+
+	factory.Start(ctx.Done())
+	go func() {
+		if cache.WaitForCacheSync(ctx.Done(), inf.Informer().HasSynced) {
+			r.updateState()
+		} else if ctx.Err() == nil {
+			slog.Error("k8sresolver failed to sync EndpointSlice cache", slog.String("namespace", ns), slog.String("service", svc))
+			cc.ReportError(fmt.Errorf("k8sresolver: failed to sync EndpointSlice cache for %s/%s", ns, svc))
+		}
+	}()
+	return r, nil
+}
+
+// ParseTarget parses resolver.Target in canonical format k8s:///<namespace>/<service>[:port] or k8s:///<service>.<namespace>[:port].
+func ParseTarget(target resolver.Target) (ns, svc, port string, err error) {
+	ep := target.Endpoint()
+	if target.URL.Host != "" && target.URL.Host != "localhost" {
+		ep = target.URL.Host + "/" + strings.TrimPrefix(target.URL.Path, "/")
+	}
+
+	if parts := strings.SplitN(ep, "/", 2); len(parts) == 2 {
+		ns, ep = parts[0], parts[1]
+	} else {
+		ns = "default"
+	}
+
+	svc, port, err = net.SplitHostPort(ep)
+	if err != nil {
+		svc, port = ep, "443"
+	}
+
+	if parts := strings.Split(svc, "."); len(parts) > 1 {
+		svc, ns = parts[0], parts[1]
+	}
+
+	if ns == "" || svc == "" {
+		return "", "", "", fmt.Errorf("invalid target %q, expected k8s:///<namespace>/<service>[:port]", target.String())
+	}
+	return ns, svc, port, nil
+}
+
+type k8sResolver struct {
+	ctx          context.Context
+	cancel       context.CancelFunc
+	service      string
+	port         string
+	cc           resolver.ClientConn
+	informer     cache.SharedIndexInformer
+	registration cache.ResourceEventHandlerRegistration
+	lister       discoverylisters.EndpointSliceNamespaceLister
+	selector     labels.Selector
+}
+
+func (r *k8sResolver) ResolveNow(resolver.ResolveNowOptions) { r.updateState() }
+
+func (r *k8sResolver) Close() {
+	r.cancel()
+	if r.registration != nil {
+		_ = r.informer.RemoveEventHandler(r.registration)
+	}
+}
+
+func (r *k8sResolver) updateState() {
+	if r.ctx.Err() != nil {
+		return
+	}
+
+	eps, err := r.lister.List(r.selector)
+	if err != nil {
+		if r.ctx.Err() == nil {
+			r.cc.ReportError(err)
+		}
+		return
+	}
+
+	targetPort, _ := strconv.Atoi(r.port)
+	var addrs []resolver.Address
+	for _, slice := range eps {
+		p := targetPort
+		if p == 0 && len(slice.Ports) > 0 && slice.Ports[0].Port != nil {
+			p = int(*slice.Ports[0].Port)
+		}
+		if p == 0 {
+			p = 443
+		}
+		for _, ep := range slice.Endpoints {
+			if !isHealthy(&ep) {
+				continue
+			}
+			for _, ip := range ep.Addresses {
+				addrs = append(addrs, resolver.Address{Addr: net.JoinHostPort(ip, strconv.Itoa(p))})
+			}
+		}
+	}
+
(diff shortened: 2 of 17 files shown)


michaelasp:
TLS error caused e2e to fail, changed the url to match.

EItanya:
Initial agent review:

`dns:///` remains supported by gRPC, and adding the `k8s` resolver does not cause DNS targets to read EndpointSlices. However, this PR makes `K8sClient` unconditionally required by `ateapiauth.DialOptions`, even when the caller continues to use a `dns:///` target. That unnecessarily couples out-of-cluster/DNS clients to Kubernetes client configuration, despite them never using the `k8s:///` resolver.

Could the Kubernetes client requirement be made conditional/opt-in so only callers selecting `k8s:///` need Kubernetes API access and EndpointSlice RBAC? External clients should remain able to use DNS, a load balancer, or another stable endpoint without Kubernetes API credentials.

Additional findings from the initial pass:

- `boomer-glutton` now defaults to `k8s:///`, but the benchmark pod uses the default service account and its manifest does not grant EndpointSlice `list/watch`. Resolution will therefore fail in the existing benchmark deployment.
- The resolver uses the port from the dial target for each pod IP and only consults `EndpointSlice.Ports` when that target port parses as zero. A Kubernetes Service port can differ from its endpoint/target port, so this can dial the wrong pod port. The current Service happens to use 443 for both.
- `ctrl.SetLogger(zap.New(zap.UseDevMode(true)))` was removed from `atecontroller` while making the resolver changes. This appears unrelated and may suppress controller-runtime logging; it should likely be retained.
- EndpointSlice authorization/list-watch failures happen inside the informer after `Build` succeeds. Without surfacing cache-sync failure to the caller, a misconfigured client may sit unresolved instead of failing startup with a useful RBAC error.

michaelasp:
> Initial agent review:
> 
> `dns:///` remains supported by gRPC, and adding the `k8s` resolver does not cause DNS targets to read EndpointSlices. However, this PR makes `K8sClient` unconditionally required by `ateapiauth.DialOptions`, even when the caller continues to use a `dns:///` target. That unnecessarily couples out-of-cluster/DNS clients to Kubernetes client configuration, despite them never using the `k8s:///` resolver.
> 
> Could the Kubernetes client requirement be made conditional/opt-in so only callers selecting `k8s:///` need Kubernetes API access and EndpointSlice RBAC? External clients should remain able to use DNS, a load balancer, or another stable endpoint without Kubernetes API credentials.

Ok I made it so that if we provide a kubernetes client we create a resolver. Although on the apiserver side of things, we used to set the keepalive in order to close connections to allow for us to pick up new endpoints, however that is not optimal. This PR removes that in order to prevent spurious connection drops. This change would make DNS clients not be able tell if new apiservers have come up until they restarted. Not sure if we should account for that. @juli4n made the change in order to add the keepalive, wdyt?

EItanya:
> > Initial agent review:
> > `dns:///` remains supported by gRPC, and adding the `k8s` resolver does not cause DNS targets to read EndpointSlices. However, this PR makes `K8sClient` unconditionally required by `ateapiauth.DialOptions`, even when the caller continues to use a `dns:///` target. That unnecessarily couples out-of-cluster/DNS clients to Kubernetes client configuration, despite them never using the `k8s:///` resolver.
> > Could the Kubernetes client requirement be made conditional/opt-in so only callers selecting `k8s:///` need Kubernetes API access and EndpointSlice RBAC? External clients should remain able to use DNS, a load balancer, or another stable endpoint without Kubernetes API credentials.
> 
> Ok I made it so that if we provide a kubernetes client we create a resolver. Although on the apiserver side of things, we used to set the keepalive in order to close connections to allow for us to pick up new endpoints, however that is not optimal. This PR removes that in order to prevent spurious connection drops. This change would make DNS clients not be able tell if new apiservers have come up until they restarted. Not sure if we should account for that. Julian Gutierrez Oschmann (@juli4n) made the change in order to add the keepalive, wdyt?

What if we have 2 code paths:
1. keep the old path when no `k8s` resolver can be built
2. use the new one when it can

michaelasp:
> What if we have 2 code paths:
> 
> 1. keep the old path when no `k8s` resolver can be built
> 2. use the new one when it can

The issue is that this is on 2 sides and we can have many clients to one server. The server can be configured with how long to keep the connection alive(default infinity) but was turned down to kill a connection every minute to allow dns based clients to find new replicas. This is bad for keeping a live connection, and was reported as an issue in GRPC, https://github.com/grpc/grpc/issues/12295

The endpoint resolver fixes that but doesn't help if clients use the dns backend without the resolver. I think that we should expect that if clients connect with no resolver then they do not get the ability to dynamically obtain backends. Maybe we could create a MaxConnectionAge of something large, ~1 hr or so just to ensure that eventually those clients would obtain backends?

EItanya:
> > What if we have 2 code paths:
> > 
> > 1. keep the old path when no `k8s` resolver can be built
> > 2. use the new one when it can
> 
> The issue is that this is on 2 sides and we can have many clients to one server. The server can be configured with how long to keep the connection alive(default infinity) but was turned down to kill a connection every minute to allow dns based clients to find new replicas. This is bad for keeping a live connection, and was reported as an issue in GRPC, [grpc/grpc#12295](https://github.com/grpc/grpc/issues/12295)
> 
> The endpoint resolver fixes that but doesn't help if clients use the dns backend without the resolver. I think that we should expect that if clients connect with no resolver then they do not get the ability to dynamically obtain backends. Maybe we could create a MaxConnectionAge of something large, ~1 hr or so just to ensure that eventually those clients would obtain backends?

That sounds ok to me. But maybe stepping back a second what would be the typical way to solve something like this?

michaelasp:
> That sounds ok to me. But maybe stepping back a second what would be the typical way to solve something like this?

This is mostly an issue that discovery of endpoints happens at the beginning of a connection instantiation with DNS and grpc creates a long lived connection that unless dropped will not reobtain endpoints. We could maybe also create a DNS resolver that obtains endpoints via polling or some other mechanism, but if we consider that most clients will be able to obtain native kubernetes endpoints then this should be enough to eventually balance out any non native client and allow for clients that do want/need immediate access to all endpoints the ability to do so.

I added a hour connection drop so it should reconnect infrequently enough to not cause any issues for clients but allow for any DNS based client to ~eventually obtain IPs.

internal/k8sresolver/resolver.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	update := func(_ any) { r.updateState() }
+	_, _ = inf.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: update, UpdateFunc: func(_, n any) { update(n) }, DeleteFunc: update,
+	})
+
+	factory.Start(ctx.Done())
+	go func() {
+		if cache.WaitForCacheSync(ctx.Done(), inf.Informer().HasSynced) {
```
BenTheElder, line 96: 🤖 **should-fix** 🟡 – Informer failures are invisible to gRPC, so a broken resolver looks like a hang rather than an error. `ReportError` is only reachable from `lister.List`, which reads a local cache and effectively cannot fail. If the watch is rejected — missing `endpointslices` RBAC being the obvious case, since this PR is adding it — `WaitForCacheSync` never returns true, `updateState` never runs, and the `ClientConn` sits with no addresses and no error while RPCs block to their deadline. `dns:///` surfaced resolution failures instead.

Reporting the `false` return here, and wiring `SetWatchErrorHandler` to `cc.ReportError`, would turn that into a diagnosable failure. Worth also surfacing the `UpdateState` return below: it is discarded, and an empty address set makes gRPC return `ErrBadResolverState`, which is exactly the "service has no ready endpoints" signal.

internal/k8sresolver/resolver.go
```diff
@@ -0,0 +1,191 @@
...
+	service  string
+	port     string
+	cc       resolver.ClientConn
+	cancel   context.CancelFunc
+	lister   discoverylisters.EndpointSliceNamespaceLister
+	selector labels.Selector
+}
+
+func (r *k8sResolver) ResolveNow(resolver.ResolveNowOptions) { r.updateState() }
+func (r *k8sResolver) Close()                                { r.cancel() }
```
BenTheElder: 🤖 **nit** 🟢 – `Close` cancels the context but does not wait for in-flight informer handlers, so `updateState` can still call `r.cc` afterwards. gRPC's resolver contract asks that no `ClientConn` method be called once `Close` returns. Tracking the handler registration and removing it here, or guarding on the cancelled context before touching `r.cc`, would close the window.

BenTheElder (commented):
Thanks! 

AI first pass with some minor issues:

BenTheElder:
Should also fix the CI / linter findings.

internal/k8sresolver/resolver.go
michaelasp: Done! Made it so that we guard on context cancellation.

internal/k8sresolver/resolver.go
michaelasp: Done, reported the WaitForCacheSync as an error.

internal/k8sresolver/resolver.go
```diff
@@ -0,0 +1,214 @@
...
+		selector: selector,
+	}
+
+	_ = inf.Informer().SetWatchErrorHandler(func(_ *cache.Reflector, err error) {
+		if ctx.Err() == nil {
+			cc.ReportError(fmt.Errorf("k8sresolver: watch error for %s/%s: %w", ns, svc, err))
+		}
+	})
+
+	update := func(_ any) { r.updateState() }
```
juli4n: Maybe just:

```
  reg, err := inf.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
    AddFunc:          func(_ any) { r.updateState() },
    UpdateFunc:     func(_, _ any) { r.updateState() },
    DeleteFunc:      func(_ any) { r.updateState() },
  }
```

There is no much value in the update wrapper as `UpdateFunc` needs to also inline the closure.

michaelasp: Done! Also added an error check here, that might have been bad.

internal/k8sresolver/resolver.go
```diff
@@ -0,0 +1,219 @@
...
+		cancel:   cancel,
+		service:  svc,
+		port:     port,
+		cc:       cc,
+		informer: inf.Informer(),
+		lister:   inf.Lister().EndpointSlices(ns),
+		selector: selector,
+	}
+
+	_ = inf.Informer().SetWatchErrorHandler(func(_ *cache.Reflector, err error) {
```
BenTheElder, line 76: 🤖 **nit** 🟢 – Installing a handler here replaces client-go's default one, which was the thing logging "failed to list/watch". Watch errors now reach only `cc.ReportError`, so they surface through failing RPCs and nowhere else — an idle `ClientConn` produces no operator-visible signal at all, which is the case where a missing-RBAC misconfiguration is easiest to miss. A `slog.Error` alongside the `ReportError` keeps both.

michaelasp: Ah good catch, made it so this logs an error as well. Thanks!

michaelasp:
Seems like we hit #675, rerun should probably pass.

manifests/ate-install/atenet-router.yaml
```diff
@@ -33,6 +33,14 @@ rules:
   - get
   - watch
   - list
+- apiGroups:
```
juli4n, line 56: Do we need to update atelet as well so it can watch the endpoints?

michaelasp: Probably, let me update this too.

manifests/ate-install/generated/role.yaml
```diff
@@ -80,6 +80,14 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
```
juli4n: Can / should we scope this down to the `ate-system` namespace using a Role instead? We don't need to watch endpoints in other namespaces.

internal/k8sresolver/resolver.go
```diff
@@ -0,0 +1,222 @@
...
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	discoverylisters "k8s.io/client-go/listers/discovery/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+)
+
+const Scheme = "k8s"
+
+func init() { resolver.Register(NewBuilder(nil)) }
```
juli4n: How about removing this (and the check in line 55) and always require the Builder to hold a client?

manifests/ate-install/generated/role.yaml
michaelasp: Yeah makes sense. I've run into issues before where the cache doesn't get populated when we pare the permissions down but that's on controller-runtime usually. Shared informer shouldn't run into that. I'll double check in kind to make sure we don't run into rbac issues.

internal/k8sresolver/resolver.go
michaelasp: Done!

manifests/ate-install/generated/role.yaml
michaelasp: Double checked in kind, no rbac issues when initializing.

manifests/ate-install/atenet-router.yaml
michaelasp: Done!

manifests/ate-install/generated/role.yaml
juli4n: We also want this one to be down scoped to the namespace.

cmd/atecontroller/internal/controllers/actortemplate_controller.go
```diff
@@ -59,6 +59,7 @@ type ActorTemplateReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups=discovery.k8s.io,resources=endpointslices,verbs=get;list;watch
```
juli4n: namespace=ate-system?

michaelasp: Done

manifests/ate-install/generated/role.yaml
michaelasp: Fixed!

juli4n (dismissed):

michaelasp:
#799 hit this time, passed locally.

juli4n (approved):

PR #590, opened 2026-07-29
Add unit tests for request validators in ate-apiserver.

juli4n:
Add unit tests that exercise all the validation cases and assert on the k8s `ErrorMatcher` instead of comparing raw error strings.

Moved most validations from functional tests to unit tests, and kept one validation for each gRPC method to validate that handlers properly map Go errors to gRPC statuses.

### DIFF
--- a/cmd/ateapi/internal/controlapi/create_actor.go
+++ b/cmd/ateapi/internal/controlapi/create_actor.go
@@ -30,8 +30,8 @@ import (
 )
 
 func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequest) (*ateapipb.Actor, error) {
-	if err := validateCreateActorRequest(req); err != nil {
-		return nil, err
+	if errs := validateCreateActorRequest(req); len(errs) > 0 {
+		return nil, toGRPCStatusError(errs)
 	}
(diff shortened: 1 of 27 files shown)


EItanya:
I haven't done a full review yet, but I have been thinking out of band that this package is getting really really large, what do you think?

EItanya (approved):
LGTM

juli4n:
> I haven't done a full review yet, but I have been thinking out of band that this package is getting really really large, what do you think?

I think having all gRPC handlers in this package makes sense (as well as the tests). We also have the workflows (which *really* need a rewrite) in this same package and I think that would be a great candidate to be moved to another package. I recently moved the scheduling logic under `cmd/ateapi/internal/scheduling`, which is a step towards this direction.

EItanya:
> > I haven't done a full review yet, but I have been thinking out of band that this package is getting really really large, what do you think?
> 
> I think having all gRPC handlers in this package makes sense (as well as the tests). We also have the workflows (which _really_ need a rewrite) in this same package and I think that would be a great candidate to be moved to another package. I recently moved the scheduling logic under `cmd/ateapi/internal/scheduling`, which is a step towards this direction.

100%. I was mostly referring to the workflows

PR #591, opened 2026-07-29
feat: extend edge routing telemetry with resume and outcome labels

### LINKED ISSUE #564: Feature request: More system metrics for debuggability (Activation SLI, crash accounting, and capacity signals)

## Summary

|---|---|---|---|

### At a glance

|---|---|---|---|---|

---

---

## Prerequisite

## What's next

Angelawork:
Fixes #564 (Part 1)

- [x] Tests pass

## Description
This PR implements the first part of #564 by extending metric `atenet.router.route.duration` with singleflight activation state (`ate.router.resume`) and outcome status labels (`ate.router.outcome`). All telemetry attribute keys and label value sets are centralized in `internal/ateattr` to ensure consistency across components.

## Key Changes:
   * Added `resumed` boolean to `ResumeActorResponse` in `ateapi.proto` indicating whether a cold activation workflow occurred (`!state.WasRunning`).
   * Updated `workflow_resume_test.go` to assert `resumed == true` on cold activations and `resumed == false` for already-running actors.
   * Extended `classifyOutcome(err)` in Envoy `ext_proc` to classify errors into metric labels: `ok`, `cancelled`, `timeout`, `no_capacity`, `lock_conflict`, `not_found`, `unavailable`, `rate_limited`, and `resume_error`.
   * Added explicit handling for `codes.Unavailable` (`"unavailable"`) and `codes.ResourceExhausted` / `StatusCode_TooManyRequests` (`"rate_limited"`).
   * Consolidated metric label names on `atenet.router.route.duration` using standard keys from `internal/ateattr`:
     * `ate.template.namespace` (`TemplateNamespaceKey`)
     * `ate.template.name` (`TemplateNameKey`)
     * `ate.router.outcome` (`RouterOutcomeKey`)
     * `ate.router.resume` (`RouterResumeKey`)
   * Updated `manifests/ate-install/kind/kustomization.yaml` to export `atenet-router` OTLP metrics to `opentelemetry-collector`.

## Testing
* `make test`
## E2E  Test


cmd/atenet/internal/router/extproc.go
```diff
@@ -197,29 +200,49 @@ func (s *ExtProcServer) handleRequestHeaders(
...
+		return "timeout"
+	}
+	switch status.Code(err) {
+	case codes.FailedPrecondition:
+		return "no_capacity"
+	case codes.Aborted:
+		return "lock_conflict"
+	case codes.NotFound:
+		return "not_found"
+	}
```
JeffLuoo, line 239: Please also add case `codes.Unavailable` is returned, and map it to return "unavailable".

cmd/atenet/internal/router/extproc.go
```diff
@@ -104,11 +108,11 @@ func (s *ExtProcServer) Process(stream extprocv3.ExternalProcessor_ProcessServer
 				} else {
 					resp = immediateResponse(envoy_type.StatusCode_InternalServerError, err.Error())
 				}
-				s.recordRouteDuration(stream.Context(), elapsed, tmplNs, tmplName, classifyOutcome(err))
+				s.recordRouteDuration(stream.Context(), elapsed, tmplNs, tmplName, outcomeStr, resumeStr)
 				s.recorder.AddRouterRequest(start, elapsed, "Error", "-", rqm)
 			} else {
 				resp.Response = &extprocv3.ProcessingResponse_RequestHeaders{RequestHeaders: hResponse}
-				s.recordRouteDuration(stream.Context(), elapsed, tmplNs, tmplName, "ok")
+				s.recordRouteDuration(stream.Context(), elapsed, tmplNs, tmplName, outcomeStr, resumeStr)
```
JeffLuoo, line 115: If the condition runs into this block, the condition is err = nil, have you tested the outcomeStr and resumeStr are always healthy return like "ok"?

cmd/atenet/internal/router/extproc.go
```diff
@@ -197,29 +200,49 @@ func (s *ExtProcServer) handleRequestHeaders(
...
-func (s *ExtProcServer) recordRouteDuration(ctx context.Context, d time.Duration, tmplNs, tmplName, outcome string) {
+func (s *ExtProcServer) recordRouteDuration(ctx context.Context, d time.Duration, tmplNs, tmplName, outcome, resume string) {
 	if s.routeDuration == nil {
 		return
 	}
 	s.routeDuration.Record(ctx, d.Seconds(), metric.WithAttributes(
 		attribute.String("actor_template_namespace", tmplNs),
 		attribute.String("actor_template_name", tmplName),
 		attribute.String("outcome", outcome),
+		attribute.String("resume", resume),
```
JeffLuoo: We want to use namespaced attributes per comment: https://github.com/agent-substrate/substrate/issues/564#issuecomment-5118641350

Maybe replace it with `ate.router.resume`, @krisztianfekete can you help confirm this is the attribute you expect?

JeffLuoo: Also, please update docs/observability.md to include new attribute

cmd/atenet/internal/router/resumer.go
```diff
@@ -167,8 +190,8 @@ func (r *ActorResumer) ResumeActor(ctx context.Context, actorRef resources.Actor
 			}
 
 			if r.retryable(err) {
-				lastRetryErr = err // remember it in case the budget elapses
-				return false, nil  // park: retry until the budget elapses
+				lastRetryErr = err
+				return false, nil
```
JeffLuoo: Is removing the in-line comments intentional? If not, let's add it back.

cmd/atenet/internal/router/resumer.go
```diff
@@ -64,6 +65,25 @@ type budgetExhaustedError struct{ lastErr error }
...
+	// ResumeOutcomeTriggered indicates this request won the singleflight lock and initiated cold activation.
+	ResumeOutcomeTriggered ResumeOutcome = "triggered"
+	// ResumeOutcomeJoined indicates this request parked on an in-flight singleflight resume.
+	ResumeOutcomeJoined ResumeOutcome = "joined"
+)
+
+type resumeCallResult struct {
+	actor    *ateapipb.Actor
+	resumed  bool
+	leaderID uint64
```
JeffLuoo, line 85: Add comments to non-obvious semantic - resumed and leaderID

cmd/atenet/internal/router/resumer.go
```diff
@@ -78,6 +98,7 @@ type ActorResumer struct {
 	budget time.Duration
 	// backoff paces the retries within the budget.
 	backoff wait.Backoff
+	nextID  uint64
```
JeffLuoo: ditto, add comment to the nextID

cmd/atenet/internal/router/resumer.go
```diff
@@ -187,21 +210,44 @@ func (r *ActorResumer) ResumeActor(ctx context.Context, actorRef resources.Actor
...
+			return nil, ResumeOutcomeNone, callRes.err
+		}
+
+		outcome := ResumeOutcomeNone
+		if callRes.resumed {
+			if callRes.leaderID == reqID {
+				outcome = ResumeOutcomeTriggered
+			} else {
+				outcome = ResumeOutcomeJoined
+			}
```
JeffLuoo, line 260: I think this whole block is the core logic to determine the resume attribute, and it's not straightforward to read in the beginning, let's also add more comments in this block to explain how the logic is implemented.

JeffLuoo (changes requested):

cmd/atenet/internal/router/extproc.go
Angelawork: Yes. Since err is nil, classifyOutcome(err) will consistently return "ok" as the value assigned to outcomeStr. To confirm this behavior, the test case "TestRecordRouteDuration_Attributes" is added

JeffLuoo:
Overall it LGTM, but since this change touches `ResumeActor` function, and also this is the first PR I review, I would like to have it also reviewed by folks who are more experienced on this project.

@juli4n @maxsmythe @zoez7

cmd/atenet/internal/router/extproc.go
```diff
@@ -197,29 +200,51 @@ func (s *ExtProcServer) handleRequestHeaders(
...
-func (s *ExtProcServer) recordRouteDuration(ctx context.Context, d time.Duration, tmplNs, tmplName, outcome string) {
+func (s *ExtProcServer) recordRouteDuration(ctx context.Context, d time.Duration, tmplNs, tmplName, outcome, resume string) {
 	if s.routeDuration == nil {
 		return
 	}
 	s.routeDuration.Record(ctx, d.Seconds(), metric.WithAttributes(
 		attribute.String("actor_template_namespace", tmplNs),
 		attribute.String("actor_template_name", tmplName),
 		attribute.String("outcome", outcome),
+		attribute.String("ate.router.resume", resume),
```
krisztianfekete: Can you please define this key (and its three values) in internal/ateattr instead of inlining the string here? That package is documented as the single source of truth for `ate.*` keys, and it keeps the value set from drifting.

cmd/atenet/internal/router/extproc.go
```diff
@@ -197,29 +200,51 @@ func (s *ExtProcServer) handleRequestHeaders(
...
 
-func (s *ExtProcServer) recordRouteDuration(ctx context.Context, d time.Duration, tmplNs, tmplName, outcome string) {
+func (s *ExtProcServer) recordRouteDuration(ctx context.Context, d time.Duration, tmplNs, tmplName, outcome, resume string) {
 	if s.routeDuration == nil {
 		return
 	}
 	s.routeDuration.Record(ctx, d.Seconds(), metric.WithAttributes(
 		attribute.String("actor_template_namespace", tmplNs),
 		attribute.String("actor_template_name", tmplName),
 		attribute.String("outcome", outcome),
```
krisztianfekete: We have a mix of naming styles here. Can we consolidate with what we started in https://github.com/agent-substrate/substrate/pull/412, and use `ate.template.namespace`/`ate.template.name`, and either an `ate.*`-namespaced outcome or the semconv `error.type` (absent on success instead of outcome=ok). We can do this as a follow-up, but this is the first time we ship them of to a collector so might be better to do here. 

What do others think?

cmd/atenet/internal/router/extproc.go
```diff
@@ -197,29 +200,51 @@ func (s *ExtProcServer) handleRequestHeaders(
...
+		attribute.String("ate.router.resume", resume),
 	))
 }
 
 func classifyOutcome(err error) string {
```
krisztianfekete, line 218: We probably need to handle rate-limited requests here, right?

cmd/ateapi/internal/controlapi/workflow_resume_test.go
```diff
@@ -460,6 +454,9 @@ func TestResumeActorWorkflow_RejectedAndIdempotentPaths(t *testing.T) {
 				if actor.GetStatus() != tc.wantStatus {
 					t.Errorf("returned status = %v, want %v", actor.GetStatus(), tc.wantStatus)
 				}
+				if tc.seedStatus == ateapipb.Actor_STATUS_RUNNING && resumed {
```
krisztianfekete: Could you please also assert resumed == true?

cmd/atenet/internal/router/extproc.go
Angelawork: Sure, I've renamed these inline attribute str with keys from internal/ateattr

cmd/atenet/internal/router/extproc.go
Angelawork: I added the cases codes.ResourceExhausted and envoy_type.StatusCode_TooManyRequests to map to "rate_limited" as the outcomeStr

JeffLuoo (dismissed):
Please rebase

Angelawork:
> Please rebase

Done!

JeffLuoo:
@krisztianfekete Can you please take another look of this PR? Thanks!

krisztianfekete (approved):

JeffLuoo:
Please rebase and I can merge it.

Angelawork:
> Please rebase and I can merge it.

Done! Thanks for helping with the merge

PR #595, opened 2026-07-29
fix: match actor DNS names case-insensitively at the router

mesutoezdil:
DNS lookups are case-insensitive (RFC 4343), so a client that resolves `MyActor.MySpace.actors.resources.substrate.ate.dev` reaches the router with that spelling preserved in `Host`/`:authority` (Go's http.Client and `curl -H` both send the host as written). `ParseActorDNSName` compared it byte for byte against a strictly lower-case suffix and name pattern, so the request was rejected with a 404 by `invalidHostErr` even though its DNS lookup had resolved to that very actor.

Fold the name to lower case before parsing. Actor and atespace names are always lower case (`ResourceNameRegexPattern`), so a folded name either addresses the same actor its lookup resolved to, or fails validation exactly as before.

### DIFF
--- a/cmd/atenet/internal/router/ingress/actorref_test.go
+++ b/cmd/atenet/internal/router/ingress/actorref_test.go
@@ -52,6 +52,12 @@ func TestParseActorRef(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			name:    "mixed-case host",
+			host:    "My-Actor.Team-A.actors.resources.substrate.ate.dev",
+			want:    resources.ActorRef{Atespace: "team-a", Name: "my-actor"},
+			wantErr: false,
+		},
+		{
 			name:    "missing atespace label",
 			host:    "my-actor.actors.resources.substrate.ate.dev",
 			wantErr: true,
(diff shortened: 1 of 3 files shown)


EItanya:
Optional note (AI-generated):

`K` and `K` look the same, but they are different characters. `K` is the Unicode Kelvin sign (U+212A), while `K` is the ASCII letter K (U+004B).

Go converts `K` to the ASCII letter `k` when calling `strings.ToLower`. As a result, a malformed non-ASCII hostname could become a valid hostname during parsing and be routed to the wrong actor.

Consider either rejecting non-ASCII hostnames before lowercasing or lowercasing only ASCII letters A-Z. If an earlier layer already guarantees that hostnames are ASCII, no change may be needed here; this would be defense in depth.

mesutoezdil:
> Optional note (AI-generated):
> 
> `K` and `K` look the same, but they are different characters. `K` is the Unicode Kelvin sign (U+212A), while `K` is the ASCII letter K (U+004B).
> 
> Go converts `K` to the ASCII letter `k` when calling `strings.ToLower`. As a result, a malformed non-ASCII hostname could become a valid hostname during parsing and be routed to the wrong actor.
> 
> Consider either rejecting non-ASCII hostnames before lowercasing or lowercasing only ASCII letters A-Z. If an earlier layer already guarantees that hostnames are ASCII, no change may be needed here; this would be defense in depth.

hm, oki makes sense, i did, ptal

BenTheElder (approved):
LGTM.

We're probably going to wind up needing this in other places down the line.

PR #599, opened 2026-07-29
ateapi: backfill unset worker state during pod sync

mesutoezdil:
Follow-up to #517.

#517 added a scheduler filter that skips any worker whose state is not `STATE_ACTIVE`:

https://github.com/agent-substrate/substrate/blob/87698ea9c6a22774cd55d685d9a6b3ad264d6f39/cmd/ateapi/internal/scheduling/scheduling.go#L113-L115

But the syncer writes `State` only when it *creates* a worker record. The update path in `syncWorkerToStore` diffs `Ip`, `SandboxClass` and `Labels` and never touches `State`, and no other code path writes `State` onto an existing record.

The store is Redis backed, so worker records outlive an ate-api-server rollout, and worker pods are not recreated by that rollout either. Every worker record written before the `State` field existed therefore stays at `STATE_UNSPECIFIED` and is skipped by the scheduler for the whole life of its pod. On an existing cluster upgrading past #517, actor placement fails with `ErrNoCapacity` until each worker pod is deleted and recreated. The startup sync does not help, because the field diff computes `changed == false` and never writes.

### DIFF
--- a/cmd/ateapi/internal/controlapi/syncer.go
+++ b/cmd/ateapi/internal/controlapi/syncer.go
@@ -160,7 +160,7 @@ func (s *WorkerPoolSyncer) Start(ctx context.Context) {
 
 		// Reconcile the other direction: enqueue every stored worker so records
 		// whose pods no longer exist are cleaned up. This recovers delete events
-		// missed while ate-api-server was down — neither the watch relist nor
+		// missed while ate-api-server was down, neither the watch relist nor
 		// the resync period can replay a delete across a process restart,
 		// because the informer cache starts empty. Runs after the cache sync so
 		// the indexer is an authoritative snapshot of live pods.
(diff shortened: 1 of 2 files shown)


cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -185,6 +185,18 @@ func (s *WorkerPoolSyncer) syncWorkerToStore(ctx context.Context, pod *corev1.Po
 		w.Labels = pool.GetLabels()
 		changed = true
 	}
+	if w.State == ateapipb.Worker_STATE_UNSPECIFIED {
+		// A worker record written before the State field existed. The store
```
zoez7: Backwards incompatible changes at this stage of the project is ok. Also, I don't think it's safe to always set `UNSPECIFIED` to  `_ACTIVE`, we can have a bug somewhere that unset the state field by mistake.

mesutoezdil: @zoez7 thx, done

mesutoezdil: done, ptal

mesutoezdil:
Closing this. The original fix was already dropped after review (see e663516, unsafe unset-state backfill), leaving only comment cleanup here. The syncer itself has since moved to cmd/atecontroller/internal/workersync (#1203, #1243), so there's nothing left for this branch to contribute.

PR #601, opened 2026-07-29
ateapi: fail cleanly on a malformed actor JWT signing pool

mesutoezdil:
`MintJWT` indexes `Authorities[0]` from the actor JWT signing pool file, and `actoridjwt.Sign` type-asserts the signing key without checking it, so a pool file carrying no authorities (which `localjwtauthority.Unmarshal` accepts without error) or an algorithm that does not match its key panics instead of returning an error. Nothing in the server installs a panic-recovery interceptor, so that configuration error takes down ate-api-server on the first `MintJWT` call, and `MintCert` already guards the same situation for its own pool. This adds the missing checks plus the first tests for both packages, which also pin the behavior of the paths that already worked.

### DIFF
--- a/cmd/ateapi/internal/actoridentity/actoridentity.go
+++ b/cmd/ateapi/internal/actoridentity/actoridentity.go
@@ -134,8 +134,13 @@ func (s *Server) MintJWT(ctx context.Context, req *ateapipb.MintJWTRequest) (*at
 		return nil, fmt.Errorf("while making actor JWT claims: %w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


ahmedtd:
Apologies, I think this may be superseded by #1313

mesutoezdil:
Closing, the file this PR touches will be deleted by #1313 so this diff cannot land as-is. The algorithm/key-type mismatch panic in sign() is still unfixed after #1313 though, I will open a fresh PR against internal/localjwtauthority/localjwtauthority.go once it merges.

PR #632, opened 2026-07-30
ateapi: key the workerpool metric by namespace

mesutoezdil:
`ate.workerpool.workers` tallies by pool name, state, and sandbox class, but a WorkerPool is namespaced, so two pools that share a name sum into a single series. This is reachable today: the counter demo and the autoscaled-workerpool demo both create a pool named `counter`, and with both installed the demo HPA scales on the other pool's assigned workers. Adds `ate.workerpool.namespace` to the tally key and attributes, and pins it in the demo's adapter query and HPA selector.

### DIFF
--- a/cmd/ateapi/internal/controlapi/metrics.go
+++ b/cmd/ateapi/internal/controlapi/metrics.go
@@ -36,7 +36,7 @@ func RegisterWorkerCount(meter metric.Meter, workers func() ([]*ateapipb.Worker,
 	counter, err := meter.Int64ObservableUpDownCounter(
 		workerpoolWorkersMetric,
 		metric.WithUnit("{worker}"),
-		metric.WithDescription("Number of workers by pool, worker state, and sandbox class."),
+		metric.WithDescription("Number of workers by pool namespace, pool, worker state, and sandbox class."),
 	)
 	if err != nil {
 		return fmt.Errorf("create %s updowncounter: %w", workerpoolWorkersMetric, err)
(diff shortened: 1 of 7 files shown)


cmd/ateapi/internal/controlapi/metrics.go
```diff
@@ -48,7 +48,10 @@ func RegisterWorkerCount(meter metric.Meter, workers func() ([]*ateapipb.Worker,
 			// Worker cache unavailable (warmup/reconnect): skip the whole observation.
 			return nil
 		}
-		type key struct{ pool, state, class string }
+		// A WorkerPool is namespaced, so the namespace is part of the key: without
+		// it two pools that share a name in different namespaces sum into one
+		// series, and an HPA reading that series scales on a foreign pool's load.
```
EItanya: Nit: This comment is a PR description, not necessary to live next to the code IMO

cmd/ateapi/internal/controlapi/metrics_test.go
```diff
@@ -134,6 +134,41 @@ func TestWorkerCountSkipsWhenCacheNotReady(t *testing.T) {
 	}
 }
 
+// TestWorkerCountSeparatesPoolsByNamespace pins the namespace dimension. A
+// WorkerPool is namespaced, so two pools that share a name must stay two series:
+// merged, an HPA selecting on the name alone scales on a foreign pool's load.
+func TestWorkerCountSeparatesPoolsByNamespace(t *testing.T) {
```
EItanya: Rather than adding a new test for this, shouldn't namespace be incorporated into all existing tests

EItanya (changes requested):
Good catch, changes make sense to me

mesutoezdil:
@EItanya thx! done
ops sorry for force push..

EItanya:
> Eitan Yarmush (@EItanya) thx! done ops sorry for force push..

I think right now commits are maintained in substrate so force pushing may be more appropriate here 😂

krisztianfekete (approved):

EItanya (approved):

PR #635, opened 2026-07-30
atenet: drop redundant deployment mode

howardjohn:
This is intended to be essentially a dead code cleanup, not a controversial change. There were two ways to deploy the same proxy component; I assume no one is customizing the deployment mechanic, and this follows how other components work (just deployed with YAML).

This simplifies the setup so we only need to maintain one path. The default option is retained so there is no behavioral changes.

- [x] Tests pass


bowei:
conflicts?

howardjohn:
> "No conflicts with base branch
> Changes can be cleanly merged."

@bowei

bowei:
Don't know why github is giving me this message:

image

Issue #563, opened 2026-07-28
Need more guidance on the configurability and scalability for OTel Collector setup


git286:
## Summary

Substrate hardcodes its OTLP endpoint to the **GKE Managed OpenTelemetry** collector (`opentelemetry-collector.gke-managed-otel.svc.cluster.local:4317`) in at least **nine places** across manifests, a CLI flag, and a best-practices doc. That collector is a centrally-deployed **Deployment** (single gateway tier), the feature is currently in **Preview**, and it is GKE-only.

We have no documented answer to any of these:

1. Is a single centralized collector Deployment going to scale for Substrate's telemetry shape?
2. Should we instead recommend operators run their own collector, and should it be a DaemonSet?
3. How does an operator point Substrate at their own collector today?
4. What *is* the recommended collector topology and config?

Today the honest answer to (3) is "patch six manifests and accept that TLS is impossible." That's not a supportable story for anyone running Substrate outside our GKE dev setup.

This issue asks for a written **Collector Setup Guide** plus the small amount of code/manifest work needed to make that guide actually followable.

---

### Two potential blockers for a bring-your-own collector

1. **TLS is impossible.** `serverboot.go:109,133` hardcode `otlptracegrpc.WithInsecure()` /`otlpmetricgrpc.WithInsecure()`, with the comment *"GKE managed traces doesn't support validating the TLS certs of the collector."* `WithInsecure()` overrides scheme inference, so setting `OTEL_EXPORTER_OTLP_ENDPOINT=https://…` will **not** produce a TLS connection — it will silently stay plaintext. Any operator pointing at an authenticated or off-cluster collector is blocked outright.
2. **Protocol is fixed to gRPC.** The `otlptracegrpc` / `otlpmetricgrpc` exporters are compiled in, so `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` is ignored.

PR #637, opened 2026-07-30
Configure Envoy tracing from OTEL_EXPORTER_OTLP_ENDPOINT

JeffLuoo:
`--otlp-collector-address` defaults to `os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")`, unify all stacks to use environment variables.

#563 

- [x] Tests pass

### DIFF
--- a/cmd/atenet/internal/router/xds.go
+++ b/cmd/atenet/internal/router/xds.go
@@ -184,20 +184,29 @@ const otlpDefaultPort = "4317"
 // collector. addr empty disables tracing. See normalizeOtlpCollector for the
 // accepted forms.
 func (x *XdsServer) SetOtlpCollector(addr string) error {
-	x.mu.Lock()
-	defer x.mu.Unlock()
 	if addr == "" {
-		x.otlpHost = ""
-		x.otlpPort = 0
+		x.DisableOtlpCollector()
 		return nil
 	}
+	// normalizeOtlpCollector reads nothing off x, so it runs unlocked.
 	host, port, err := normalizeOtlpCollector(addr)
 	if err != nil {
 		return err
 	}
+	x.mu.Lock()
+	defer x.mu.Unlock()
 	x.otlpHost = host
 	x.otlpPort = port
 	return nil
+}
+
+// DisableOtlpCollector turns Envoy-side tracing off. The router's own exporter
+// is independent of this and keeps reporting spans.
+func (x *XdsServer) DisableOtlpCollector() {
+	x.mu.Lock()
+	defer x.mu.Unlock()
+	x.otlpHost = ""
+	x.otlpPort = 0
 }
 
 // normalizeOtlpCollector resolves a collector endpoint to the bare host and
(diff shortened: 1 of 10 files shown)


manifests/ate-install/kind/kustomization.yaml
```diff
@@ -70,3 +70,34 @@ patches:
...
+        namespace: ate-system
+      spec:
+        template:
+          spec:
+            containers:
+            - name: atenet-router
+              env:
+              - name: OTEL_EXPORTER_OTLP_ENDPOINT
+                value: http://opentelemetry-collector.otel-system.svc:4317
+  # Envoy's tracing collector is passed as a container arg rather than an env
```
juli4n: Why not just remove the `otlp-collector-address` flag for the atenet and make it honor the OTEL_EXPORTER_OTLP_ENDPOINT env variable? That would solve this problem and make it consistent with other substrate components (ate-apiserver, atelet).

EItanya: 100%, in fact I think we should favor these wellknown env vars everywhere

JeffLuoo: The flag isn't the router's own exporter endpoint. atenet-router already honors OTEL_EXPORTER_OTLP_ENDPOINT for its own spans and this PR adds overlay to it.

`otlp-collector-address` is what Envoy xDS logic expects, it's not reading environment variable to configure the endpoint, I can probably refactor it to also read from env var to align with other components.

Ref: https://sourcegraph.com/r/github.com/agent-substrate/substrate@46adcb8017852fa4e322798828f3b2ea361fc4cf/-/blob/cmd/atenet/internal/router/xds.go?L197-198

I will update it to both read from env var.

juli4n: Right, the comment was about not having the flag and using the value in the ENV variable as the value we propagate through xDS.

JeffLuoo: Updated to let xDS read the configuration from env var like other system components.

EItanya (commented):
LGTM but will leave to @juli4n for final review

cmd/atenet/internal/router/xds.go
```diff
@@ -185,18 +191,64 @@ func (x *XdsServer) SetOtlpCollector(addr string) error {
...
+	hostport := addr
+	if strings.Contains(addr, "://") {
+		u, err := url.Parse(addr)
+		if err != nil {
+			return "", 0, fmt.Errorf("parse OTLP collector endpoint %q: %w", addr, err)
+		}
+		switch u.Scheme {
+		case "http":
+		case "https":
+			return "", 0, fmt.Errorf("OTLP collector endpoint %q uses https, which Envoy-side tracing does not support: the tracer cluster is plaintext h2c. Point it at an http:// endpoint, or set --otlp-collector-address explicitly to disable (empty) or override", addr)
```
krisztianfekete: Could this crash the pod?

JeffLuoo: Good catch, I refactor it to make the tracing not fail the pod and it will print warning logs.

krisztianfekete (approved):

PR #639, opened 2026-07-30
atelet: return destination Close error from copyFile

### LINKED ISSUE #611: [P2] copyFile discards destination.Close() error and has no fsync — local checkpoint files may be silently incomplete
**Severity:** P2 (silent data corruption in local checkpoint files; actor crashes on demand-page fault)
**Component:** Node Layer — `cmd/atelet/main.go`
**Audit ID:** NL-8

---

## Summary

`copyFile` (called during every LOCAL-type restore to copy checkpoint files from
`LocalCheckpointsDir` to `RestoreStateDir`) defers `destination.Close()` and discards
its return error. If the OS has buffered writes that fail to flush on close (e.g.,
delayed allocation failure, I/O error), `copyFile` returns `nil` with a correct byte
count while the destination file is silently incomplete. Additionally, there is no
`Sync()` call before close. The live VM demand-pages from these files for its entire
lifetime (see ISSUE-003) — a silently incomplete copy causes a `SIGBUS` mid-execution.

---

## Root Cause

**File:** `cmd/atelet/main.go` lines 647–670

```go
func copyFile(src, dst string) (int64, error) {
    source, err := os.Open(src)
    if err != nil { return 0, err }
    defer source.Close()

    destination, err := os.Create(dst)
    if err != nil { return 0, err }
    defer destination.Close()  // ← error DISCARDED

    n, err := io.Copy(destination, source)
    return n, err  // ← returns BEFORE destination.Close() flushes
    // No destination.Sync() call
}
```

The deferred `destination.Close()` runs after `return n, err` but its error value is
not captured anywhere in the defer. On kernels using delayed allocation (ext4 with
`data=ordered`, XFS), the actual disk write may not happen until close. If it fails
at that point, the caller of `copyFile` receives `nil` error but the destination file
is missing pages.

---

## Steps to Reproduce

1. Mount the atelet node's volume on a filesystem that triggers close-time I/O errors
   (e.g., inject an I/O error at the block layer using `dm-flakey` during close).
2. Trigger a LOCAL-type Restore.
3. `copyFile` returns `nil`; atelet calls `ateom.RestoreWorkload`.
4. VM starts; on first access to the incomplete memory region:
   - gVisor: `runsc` receives `SIGBUS` from the kernel on the demand-page fault.
   - microvm: `cloud-hypervisor` returns an I/O error on the first memory access from
     the guest.
5. Actor crashes with no clear error in atelet logs — the crash originates in the guest.

---

## Expected Behavior

`copyFile` should:
1. Call `destination.Sync()` after `io.Copy` succeeds and before closing.
2. Capture and return `destination.Close()` error (if `io.Copy` already returned `nil`).

---

## Suggested Fix

```go
func copyFile(src, dst string) (int64, error) {
    source, err := os.Open(src)
    if err != nil { return 0, err }
    defer source.Close()

    destination, err := os.Create(dst)
    if err != nil { return 0, err }
    defer func() {
        if cerr := destination.Close(); cerr != nil && err == nil {
            err = cerr  // capture close error
        }
    }()

    n, err := io.Copy(destination, source)
    if err != nil { return n, err }

    if err = destination.Sync(); err != nil {  // fsync before close
        return n, err
    }
    return n, nil
}
```

HavenXia:
Otherwise if the deferred `Close()` failed the caller will still received a successful response from `copyFile()`.

Fixes #611.

- [x] Tests pass

### DIFF
--- a/cmd/atelet/main_test.go
+++ b/cmd/atelet/main_test.go
@@ -86,6 +86,63 @@ func TestWriteFileAtomic(t *testing.T) {
 	})
 }
 
+func TestCopyFile(t *testing.T) {
+	dir := t.TempDir()
+	src := filepath.Join(dir, "src")
+	want := []byte("checkpoint pages")
+	if err := os.WriteFile(src, want, 0o600); err != nil {
+		t.Fatalf("seeding src: %v", err)
+	}
+
+	dst := filepath.Join(dir, "dst")
+	n, err := copyFile(src, dst)
+	if err != nil {
+		t.Fatalf("copyFile: %v", err)
+	}
+	if n != int64(len(want)) {
+		t.Errorf("copied %d bytes, want %d", n, len(want))
+	}
+	got, err := os.ReadFile(dst)
+	if err != nil {
+		t.Fatalf("reading dst: %v", err)
+	}
+	if !bytes.Equal(got, want) {
+		t.Errorf("dst content = %q, want %q", got, want)
+	}
+
+	if _, err := copyFile(dir, filepath.Join(dir, "dst2")); err == nil {
+		t.Error("copyFile(directory, ...) succeeded, want error")
+	}
+}
+
+type failingCloseFile struct{ *os.File }
+
+func (f failingCloseFile) Close() error {
+	_ = f.File.Close()
+	return errors.New("deferred flush failed")
+}
+
+func TestCopyFile_CloseError(t *testing.T) {
+	orig := createDestFile
+	createDestFile = func(name string) (io.WriteCloser, error) {
+		f, err := os.Create(name)
+		if err != nil {
+			return nil, err
+		}
+		return failingCloseFile{f}, nil
+	}
+	t.Cleanup(func() { createDestFile = orig })
+
+	dir := t.TempDir()
+	src := filepath.Join(dir, "src")
+	if err := os.WriteFile(src, []byte("checkpoint pages"), 0o600); err != nil {
+		t.Fatalf("seeding src: %v", err)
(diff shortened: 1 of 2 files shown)


cmd/atelet/main_test.go
```diff
@@ -86,6 +86,35 @@ func TestWriteFileAtomic(t *testing.T) {
...
+	}
+	got, err := os.ReadFile(dst)
+	if err != nil {
+		t.Fatalf("reading dst: %v", err)
+	}
+	if !bytes.Equal(got, want) {
+		t.Errorf("dst content = %q, want %q", got, want)
+	}
+
+	if _, err := copyFile(dir, filepath.Join(dir, "dst2")); err == nil {
```
BenTheElder, line 113: 🤖 **nit** 🟢 – This case never reaches the code the PR changes. `copyFile` rejects a directory up front via the `IsRegular` check, so it returns before `os.Create`, `io.Copy`, or `Close`. The new test therefore covers the happy path and an early-return error, but nothing exercises the `Close` error propagation the change exists for.

That path is genuinely awkward to reach — `copyFile` takes paths, so there is no seam to inject a failing `Close`, and provoking a real flush error needs a full filesystem. Fine to leave as-is; worth knowing the regression this fixes is not currently guarded.

BenTheElder: Human take: We could consider dependency injecting a filesystem interface.

HavenXia: Claude helped me to create a replaceable helper so test can reach it : D

PR #642, opened 2026-07-30
feat: Add control plane telemetry - actor crashes counter and operation labels

### LINKED ISSUE #564: Feature request: More system metrics for debuggability (Activation SLI, crash accounting, and capacity signals)

Angelawork:
Fixes #564 (Part 2)

- [x] Tests pass

## Description
This PR implements the second part of #564 by adding telemetry counter instrumentation for `ate.actor.crashes` in `ateapi` and establishing bounded label validation for operations and crash failure causes.

## Key Changes:
  - Updated `maybeCrashActor` and `crashActor` to reference `ateattr.OperationName*` constants and normalize `opName`.
  - Passed explicit failure reasons (`ReasonCorruptedAssignment`, `ReasonWorkerPodGone`, `ReasonWorkerReassigned`) across `workflow_resume.go`, `workflow_suspend.go`, and `workflow_pause.go`.
  - Instrumented `WorkerPoolSyncer.releaseActorOnDeadWorker` in `syncer.go` to record `recordActorCrash` when background pod deletion crashes an actor.
  - Instrumented `FinalizePausedStep` in `workflow_pause.go` to record `recordActorCrash` when `nodeName` is missing during pause finalization.
  - Added `ate_actor_crashes` to `PlatformMetricPrefixes` in `collector_metrics.go`.
  - Added pre-creation resource cleanup in `metrics_test.go` to prevent `AlreadyExists` errors on test reruns.
  - Updated `metrics_test.go` to trigger an actor crash via `UpdateActor` and `ResumeActor`, asserting `ate_actor_crashes` emission and label presence in OTel collector scrape outputs.

## Testing
* `make test`
## E2E  Test


internal/ateattr/ateattr.go
```diff
@@ -56,6 +56,17 @@ const (
 	WorkerPoolNameKey      = attribute.Key("ate.workerpool.name")
 	WorkerStateKey         = attribute.Key("ate.worker.state")
 	SandboxClassKey        = attribute.Key("ate.sandbox.class")
+	OperationNameKey       = attribute.Key("operation.name")
```
krisztianfekete: Please namespace `ate.*`-specific attrs.

```suggestion
	OperationNameKey       = attribute.Key("ate.operation.name")
```

cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -48,12 +65,37 @@ func maybeCrashActor(ctx context.Context, st store.Interface, actorRef resources
...
 	if err != nil {
 		return fmt.Errorf("while loading actor to crash: %w", err)
 	}
 
+	opName := ateattr.OperationNameUnknown
+	reason := ateattr.ReasonUnknown
+
+	if len(args) == 1 {
+		reason = args[0]
+	} else if len(args) >= 2 {
```
krisztianfekete: Can we increment after a successful `UpdateActor`, and only when the previous status wasn't CRASHED? Same ordering issue in syncer too and FinalizePausedStep, where the record happens before an update that the code expects to retry

cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -48,12 +65,37 @@ func maybeCrashActor(ctx context.Context, st store.Interface, actorRef resources
 
 // crashActor moves the actor to CRASHED state and frees the worker it was
 // assigned to, if any, so the worker can host other actors.
-func crashActor(ctx context.Context, st store.Interface, actorRef resources.ActorRef) error {
+//
+// args is variadic to support optional telemetry metadata:
+//   - 0 args: defaults opName and reason to OperationNameUnknown and ReasonUnknown
+//   - 1 arg:  (reason)
```
krisztianfekete: Is this used?

cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -48,12 +65,37 @@ func maybeCrashActor(ctx context.Context, st store.Interface, actorRef resources
 
 // crashActor moves the actor to CRASHED state and frees the worker it was
 // assigned to, if any, so the worker can host other actors.
-func crashActor(ctx context.Context, st store.Interface, actorRef resources.ActorRef) error {
+//
+// args is variadic to support optional telemetry metadata:
+//   - 0 args: defaults opName and reason to OperationNameUnknown and ReasonUnknown
```
krisztianfekete: What do you think about using explicit `opName` and `reason` string params instead?

internal/ateerrors/ateerrors.go
```diff
@@ -127,3 +128,48 @@ func ActorCrashRequested(err error) bool {
...
+	}
+	switch s {
+	case ateattr.ReasonCorruptedAssignment,
+		ateattr.ReasonWorkerReassigned,
+		ateattr.ReasonWorkerPodGone,
+		ateattr.ReasonUnknown:
+		return true
+	}
+	return false
+}
```
krisztianfekete, line 157: Hm, two things:

1.  `ateerrors` importing `ateattr` means the error package now depends on the telemetry registry. The control-plane causes are domain concepts, so what do you think about defining them here next to `Reason` and let `ateattr` reference them?
2. The allowlist is manually maintained, so the next time we add a `Reason` const will silently degrade to UNKNOWN. I am thinking about co-locating the list with the const block to avoid these drifting.

internal/ateattr/ateattr.go
```diff
@@ -56,6 +56,17 @@ const (
...
+	FailureReasonKey       = attribute.Key("ate.failure.reason")
+	ReasonKey              = FailureReasonKey
```
krisztianfekete: Doesn't seem like we use this alias, maybe we should drop it?

cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -48,12 +65,37 @@ func maybeCrashActor(ctx context.Context, st store.Interface, actorRef resources
...
+	} else if len(args) >= 2 {
+		opName = args[0]
+		reason = args[1]
+	}
+
+	opName = ateattr.NormalizeOperationName(opName)
+
+	var sandboxClass string
+	if actor.AteomPodNamespace != "" && actor.WorkerPoolName != "" && actor.AteomPodName != "" {
+		if worker, werr := st.GetWorker(ctx, actor.AteomPodNamespace, actor.WorkerPoolName, actor.AteomPodName); werr == nil && worker != nil {
```
krisztianfekete: We do this here, then repeat it below for `releaseWorker`, can we do it only once instead?

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -309,6 +310,9 @@ func (s *WorkerPoolSyncer) releaseActorOnDeadWorker(ctx context.Context, namespa
 		return nil
 	}
 
+	// Record crash counter metric when syncer crashes an actor due to a dead worker pod.
+	recordActorCrash(ctx, actor, worker.GetSandboxClass(), ateattr.OperationNameUnknown, ateattr.ReasonWorkerPodGone)
```
krisztianfekete: Hm, this isn't really unknown, right?

internal/ateattr/ateattr.go
```diff
@@ -56,6 +56,17 @@ const (
...
+)
+
+// Control-plane failure reasons for ate.actor.crashes metric.
+const (
+	ReasonCorruptedAssignment = "CORRUPTED_ASSIGNMENT"
+	ReasonWorkerReassigned    = "WORKER_REASSIGNED"
+	ReasonWorkerPodGone       = "WORKER_POD_GONE"
+	ReasonUnknown             = "UNKNOWN"
```
krisztianfekete: We probably want these to be consistently lower case?

internal/e2e/suites/metrics/metrics_test.go
```diff
@@ -82,13 +90,42 @@ func TestPlatformMetricsEmitted(t *testing.T) {
 		missing = e2e.MissingPlatformMetrics(scrape, e2e.PlatformMetricPrefixes)
 		ateomSeen = e2e.CollectorHasService(scrape, "ateom-gvisor", "ateom-microvm")
 		if len(missing) == 0 && ateomSeen {
+			// Verify ate_actor_crashes metric carries low-cardinality labels: operation_name and ate_failure_reason.
+			if !strings.Contains(scrape, `operation_name=`) {
```
krisztianfekete: Let's make sure to filter these to `ate_actor_crashes`.

docs/observability.md
```diff
@@ -110,6 +110,7 @@ Agent Substrate emits foundational OpenTelemetry system and server metrics to mo
 | Metric | Emitted by | Type | Measures |
 |--------|------------|------|----------|
 | `rpc.server.call.duration` | ateapi & atelet (gRPC servers, via `otelgrpc`) | histogram | per-method gRPC latency, request rate, and errors (labels `rpc.method`, `rpc.response.status_code`) |
+| `ate.actor.crashes` | ateapi | counter | Total number of actors that transitioned to `STATUS_CRASHED` (labels `operation_name`, `ate.failure.reason`, `ate.template.namespace`, `ate.template.name`, `ate.workerpool.name`, `ate.sandbox.class`) |
```
krisztianfekete: Let's not mix Prometheus naming with OTel.

```suggestion
| `ate.actor.crashes` | ateapi | counter | Total number of actors that transitioned to `STATUS_CRASHED` (labels `operation.name`, `ate.failure.reason`, `ate.template.namespace`, `ate.template.name`, `ate.workerpool.name`, `ate.sandbox.class`) |
```

krisztianfekete: Nit on the description: an actor can crash multiple times, and this counts the number os crashes.

internal/e2e/suites/metrics/metrics_test.go
```diff
@@ -82,13 +90,42 @@ func TestPlatformMetricsEmitted(t *testing.T) {
...
+			}
 			return
 		}
 		time.Sleep(3 * time.Second)
+
 	}
 	t.Fatalf("platform telemetry never reached the collector: missing metrics %v, ateom pushed=%v", missing, ateomSeen)
 }
 
+func triggerActorCrash(t *testing.T, ctx context.Context, clients *e2e.Clients, actorID string) {
```
JeffLuoo, line 167: Can you use the waitForStatus() function here https://sourcegraph.com/r/github.com/agent-substrate/substrate@main/-/blob/internal/e2e/suites/metrics/metrics_test.go?L102 to make the test self-verifying, confirm the actor reaches the crash status here?

internal/e2e/suites/metrics/metrics_test.go
```diff
@@ -82,13 +90,42 @@ func TestPlatformMetricsEmitted(t *testing.T) {
...
+			if !strings.Contains(scrape, `operation_name=`) {
+				t.Fatalf("ate_actor_crashes metric missing required operation_name label in collector scrape output:\n%s", scrape)
+			}
+			if !strings.Contains(scrape, `ate_failure_reason=`) {
+				t.Fatalf("ate_actor_crashes metric missing required ate_failure_reason label in collector scrape output:\n%s", scrape)
+			}
 			return
 		}
 		time.Sleep(3 * time.Second)
+
```
JeffLuoo, line 160: blank line not needed here

internal/ateattr/ateattr.go
JeffLuoo: The attribute will also receive reason value from https://sourcegraph.com/r/github.com/agent-substrate/substrate@main/-/blob/internal/ateerrors/ateerrors.go?L44-46. And those are upper snake. How should we align here?

krisztianfekete: Good point, I'd keep them upper snake as-is then.

cmd/ateapi/internal/controlapi/syncer.go
Angelawork: Are you suggesting that we should use `suspend` as the `opName`?

My understanding is that this pod crash can happen while the actor was running and not during any active resume/suspend/pause API call
- If the actor is already in `STATUS_SUSPENDED`, we exit `releaseActorOnDeadWorker`; 
- Any actor in another state whose worker pod vanished out-of-band is transitioned to `STATUS_CRASHED`.

internal/ateerrors/ateerrors.go
Angelawork: thanks for the review! this makes more sense
I moved definitions into ```ateerrors``` and co-located the AllReasons list (https://github.com/agent-substrate/substrate/pull/642/changes/fa2c10ceb54c20fbf43302297f4d19c01bd00375#diff-2f97874f95a04a589f9efe61b79e8bef5ff2dab5ec7a194525dda516dc2c4ec6R66) with the constants to serve as the single source of truth for valid failure reasons

cmd/ateapi/internal/controlapi/crash.go
Angelawork: Thanks for the catch! they're indeed clearer and safer than variadic args, I've refactored `crashActor` to accept explicit `opName, reason string` params

cmd/ateapi/internal/controlapi/syncer.go
Angelawork: On second thought, I realized we can infer `opName` directly from `actor.Status` 
I've updated `syncer.go`'s part at https://github.com/agent-substrate/substrate/pull/642/changes#diff-52e7595bd4adfcb5aa2f7bd2207f04cdfcfb5f6cb34508f7a54f41c98c36ef52R312-R322

cmd/ateapi/internal/controlapi/crash_test.go
```diff
@@ -350,3 +354,68 @@ func TestMaybeCrashActor(t *testing.T) {
...
+	}
+	if _, err := st.CreateActor(ctx, actor); err != nil {
+		t.Fatalf("CreateActor: %v", err)
+	}
+
+	if err := crashActor(ctx, st, actorRef, ateattr.OperationNameResume, ateattr.ReasonCorruptedAssignment); err != nil {
+		t.Fatalf("crashActor: %v", err)
+	}
+
+	assertCrashMetricDatapoint(t, reader, ateattr.OperationNameResume, ateattr.ReasonCorruptedAssignment, 1)
```
JeffLuoo: We need to assert all attributes in this function, not just op name and reason.

cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -317,6 +329,10 @@ func (s *WorkerPoolSyncer) releaseActorOnDeadWorker(ctx context.Context, namespa
 	actor.InProgressSnapshot = ""
 	actor.WorkerPoolName = ""
 
-	_, err = s.persistence.UpdateActor(ctx, actor, actor.GetMetadata().GetVersion())
+	updatedActor, err := s.persistence.UpdateActor(ctx, actor, actor.GetMetadata().GetVersion())
+	if err == nil && !wasAlreadyCrashed {
+		recordActorCrash(ctx, updatedActor, worker.GetSandboxClass(), opName, ateattr.ReasonWorkerPodGone)
```
JeffLuoo: Ditto, the metadata is reset above. We should capture it in the unit test as well to make sure the metric attributes are correct.

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -280,6 +283,9 @@ func (s *FinalizePausedStep) Execute(ctx context.Context, input *PauseInput, sta
 		latestActor.AteomPodIp = ""
 		latestActor.WorkerPoolName = ""
 		updatedActor, err := s.store.UpdateActor(ctx, latestActor, latestActor.GetMetadata().GetVersion())
+		if err == nil && updatedActor.GetStatus() == ateapipb.Actor_STATUS_CRASHED && !wasAlreadyCrashed {
+			recordActorCrash(ctx, updatedActor, "", ateattr.OperationNamePause, ateattr.ReasonCorruptedAssignment)
```
JeffLuoo: ditto, let's make sure attributes are not empty string

cmd/ateapi/internal/controlapi/metrics.go
```diff
@@ -25,7 +25,34 @@ import (
...
+	)
+	if err != nil {
+		return fmt.Errorf("create %s counter: %w", actorCrashesMetric, err)
+	}
+	actorCrashesCounter = counter
+	return nil
+}
+
+// recordActorCrash records a crash event on ate.actor.crashes with low-cardinality attributes.
+func recordActorCrash(ctx context.Context, actor *ateapipb.Actor, sandboxClass, opName, reason string) {
```
JeffLuoo: We can update the function to take attributes instead of an Actor, which might make it easier to resolve the empty metadata label issue I mentioned in other comments.

```
func recordActorCrash(ctx context.Context, attrs []attribute.KeyValue)
```

cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -69,48 +93,59 @@ func crashActor(ctx context.Context, st store.Interface, actorRef resources.Acto
...
-	if _, err := st.UpdateActor(ctx, actor, actor.GetMetadata().GetVersion()); err != nil {
+	updatedActor, err := st.UpdateActor(ctx, actor, actor.GetMetadata().GetVersion())
+	if err != nil {
 		errCollected = append(errCollected, fmt.Errorf("while marking actor crashed: %w", err))
+		return errors.Join(errCollected...)
 	}
+
+	// Increment metric only after a successful UpdateActor, and only if the actor was not already crashed.
+	if !wasAlreadyCrashed {
+		recordActorCrash(ctx, updatedActor, sandboxClass, opName, reason)
```
JeffLuoo: From L90-L94 above, metadata for the actor is reset to empty string, if we record metrics here, I think all attributes related to metadata will be empty string.

You can snapshot the attributes when it crashes and record it with saved attributes.

```
  sandboxClass, err := releaseWorker(ctx, st, actor)
  ...
  // Snapshot the crash labels before the pod and pool pointers are cleared
  // below; the counter itself is emitted only after the transition commits.
  crashAttrs := ateattr.ActorMetricAttributes(actor, sandboxClass, opName, reason)

  actor.Status = ateapipb.Actor_STATUS_CRASHED
  ...
  actor.WorkerPoolName = ""

  if _, err := st.UpdateActor(ctx, actor, actor.GetMetadata().GetVersion()); err != nil {
        errCollected = append(errCollected, fmt.Errorf("while marking actor crashed: %w", err))
        return errors.Join(errCollected...)
  }
  if !wasAlreadyCrashed {
        recordActorCrash(ctx, crashAttrs)
  }
```

JeffLuoo (changes requested):

internal/ateattr/ateattr.go
```diff
@@ -56,8 +58,18 @@ const (
...
+)
+
+// Control-plane failure reasons for ate.actor.crashes metric.
+const (
+	ReasonCorruptedAssignment = string(ateerrors.ReasonCorruptedAssignment)
+	ReasonWorkerReassigned    = string(ateerrors.ReasonWorkerReassigned)
+	ReasonWorkerPodGone       = string(ateerrors.ReasonWorkerPodGone)
+	ReasonUnknown             = string(ateerrors.ReasonUnknown)
+	RouterResumeKey           = attribute.Key("ate.router.resume")
+	RouterOutcomeKey          = attribute.Key("ate.router.outcome")
```
krisztianfekete: Can we move them back up to the metric-label keys block where they came from?

internal/e2e/suites/metrics/metrics_test.go
```diff
@@ -95,13 +104,68 @@ func TestPlatformMetricsEmitted(t *testing.T) {
...
+						return
+					}
+				}
+			}
 		}
+
 		time.Sleep(3 * time.Second)
 	}
+
 	t.Fatalf("platform telemetry never reached the collector: missing metrics %v, ateom pushed=%v", missing, ateomSeen)
```
krisztianfekete, line 164: Can you please surface more information about wrong/missing labels here?

cmd/ateapi/internal/controlapi/metrics.go
```diff
@@ -21,11 +21,39 @@ import (
...
+	actorCrashesMetric      = "ate.actor.crashes"
+)
+
+var actorCrashesCounter metric.Int64Counter
+
+// RegisterActorCrashes initializes the ate.actor.crashes counter instrument.
+func RegisterActorCrashes(meter metric.Meter) error {
+	counter, err := meter.Int64Counter(
+		actorCrashesMetric,
+		metric.WithUnit("{actor}"),
```
JeffLuoo: nit - metric.WithUnit("{crash}") -> because we are tracking crashes not tracking actor.

cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -21,23 +21,40 @@ import (
...
-// maybeCrashActor inspects err returned by an atelet RPC, it crashes
-// the actor if the err carries the actorCrashed=true metadata directive.
-func maybeCrashActor(ctx context.Context, st store.Interface, actorRef resources.ActorRef, err error, wrapMsg string) error {
+// maybeCrashActor inspects err returned by an atelet RPC and crashes the actor
+// if err carries the actorCrashed=true metadata directive.
+//
+// opNames is variadic (...string) to provide backward compatibility for generic callers
+// where the operation context is not explicitly supplied. If provided, opNames[0] is
+// normalized against the bounded operations {resume, suspend, pause, unknown}.
+func maybeCrashActor(ctx context.Context, st store.Interface, actorRef resources.ActorRef, err error, wrapMsg string, opNames ...string) error {
```
JeffLuoo: Why do we use `opNames ...string`, not just pass opNames as a single string to this function?

We are only using `opNames[0]` in the following code block:
```
		if len(opNames) > 0 {
			opName = ateattr.NormalizeOperationName(opNames[0])
		}
```

I may miss some context in previous round of review.

cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -48,17 +65,28 @@ func maybeCrashActor(ctx context.Context, st store.Interface, actorRef resources
...
 	actor, err := st.GetActor(ctx, actorRef)
 	if err != nil {
 		return fmt.Errorf("while loading actor to crash: %w", err)
 	}
 
+	wasAlreadyCrashed := actor.GetStatus() == ateapipb.Actor_STATUS_CRASHED
+	opName = ateattr.NormalizeOperationName(opName)
+	if reason == "" {
+		reason = ateattr.ReasonUnknown
+	}
```
JeffLuoo, line 66: You can remove L75-78, NormalizeOperationName would never return empty string.

JeffLuoo (changes requested):

cmd/ateapi/internal/controlapi/crash.go
Angelawork: I found that we should actually remove L44-48 from maybeCrashActor, since it always calls crashActor which will normalize the values

cmd/ateapi/internal/controlapi/crash.go
Angelawork: it being variadic was originally used to make opName optional when the operation context was not explicitly supplied; I've updated maybeCrashActor to accept an explicit opName as param instead so its consistent with crashActor and opName is always explicitly provided

JeffLuoo:
Overall it LGTM, will let @krisztianfekete to also take another round of look.

JeffLuoo:
Please rebase

internal/ateattr/ateattr.go
```diff
@@ -62,6 +66,16 @@ const (
 	SchedulerOutcomeKey    = attribute.Key("ate.scheduler.outcome")
 	RouterResumeKey        = attribute.Key("ate.router.resume")
 	RouterOutcomeKey       = attribute.Key("ate.router.outcome")
+	OperationNameKey       = attribute.Key("ate.operation.name")
```
krisztianfekete: Now that #514 is merged we have two keys. Can the crash counter reuse `ActorOperationNameKey` and drop this one, so crashes and lifecycle latency join on a single key?

krisztianfekete (commented):
Looking good, added one comment!

docs/observability.md
```diff
@@ -110,6 +110,7 @@ Agent Substrate emits foundational OpenTelemetry system and server metrics to mo
 | Metric | Emitted by | Type | Measures |
 |--------|------------|------|----------|
 | `rpc.server.call.duration` | ateapi & atelet (gRPC servers, via `otelgrpc`) | histogram | per-method gRPC latency, request rate, and errors (labels `rpc.method`, `rpc.response.status_code`) |
+| `ate.actor.crashes` | ateapi | counter | Number of times actors transitioned to `STATUS_CRASHED` with failure reasons (labels `ate.operation.name`, `ate.failure.reason`, `ate.template.namespace`, `ate.template.name`, `ate.workerpool.name`, `ate.sandbox.class`) |
```
JeffLuoo: The key name is `ate.actor.operation.name`

JeffLuoo (changes requested):
One last change needed on documentation and I can merge it.

JeffLuoo (approved):

PR #650, opened 2026-07-30
imagecache: record layer sizes at unpack and image last-use on cache hit

igooch:
First slice of #463 Phase 2 (GC): the two facts eviction will need, persisted in the
filesystem so they survive atelet restarts. Inert on its own — no behavior change for
any current caller, and nothing deletes anything.

- **Layer sizes**: unpack counts the uncompressed tar stream and writes a per-layer
  `size` file before the atomic rename (best-effort; a missing file is healed by a
  lazy backfill). Backfill covers layers unpacked by older atelets: one walk, once
  ever, skipping unreadable dirs (atelet has no `CAP_DAC_READ_SEARCH`) and preserving
  the dir mtime.
- **Last use**: cache hits touch the image record's mtime — the persisted LRU
  timestamp — under a new `hitMu` (shared side only; the eviction pass will be its
  exclusive holder).
- **`Store.CacheSize()`**: sums recorded sizes; the accounting behind the future
  `--image-cache-max-bytes`.

On-disk changes are purely additive, so no cache-layout version bump and no
migration: an older atelet reads this cache dir fine (ignores `size` files), and
this atelet reads a pre-upgrade dir fine (backfill). Rollback-safe.

**Testing**: unit tests for size recording, backfill (incl. the unreadable-dir
contract and mtime preservation), hit-path touch, and `CacheSize`; `go test -race`
green standalone; the demo e2e suite passes unchanged on Kind against this build,
with `size` files and record-mtime updates verified on the node.

### DIFF
--- a/internal/imagecache/imagecache.go
+++ b/internal/imagecache/imagecache.go
@@ -78,7 +78,10 @@ const (
 	layerFinalizedMarkerName = "finalized"
 	// layerSizeFileName holds the layer's byte count, recorded at unpack so
 	// sizing the pool never walks a tree. Absent for layers unpacked by
-	// older atelets (backfilled lazily, see layerSize).
+	// older atelets (backfilled lazily, see layerSize). An estimate: the
+	// value is the tar-stream length when written at unpack, or the summed
+	// file sizes when backfilled, so it may differ from disk usage and
+	// across nodes.
 	layerSizeFileName = "size"
 
 	// layerPullConcurrency bounds concurrent layer download+unpack streams per
--- a/internal/imagecache/imagecache.go
+++ b/internal/imagecache/imagecache.go
@@ -108,8 +108,8 @@ type Store struct {
 	layerSF singleflight.Group
 
 	// hitMu makes the hit path's record read + last-use touch atomic with
-	// respect to eviction (#463 Phase 2), which will hold it exclusive
-	// around record removal. No exclusive holder yet, so effectively free.
+	// respect to eviction, which will hold it exclusive around record
+	// removal. No exclusive holder yet, so effectively free.
 	hitMu sync.RWMutex
 }
 
@@ -270,7 +270,7 @@ func (s *Store) EnsureImage(ctx context.Context, ref string) (*Image, error) {
 	s.hitMu.RLock()
 	img, err := s.cachedImage(digest)
 	if err == nil && img != nil {
-		// Record last-use for eviction's LRU ordering (#463 Phase 2).
+		// Record last-use for eviction's LRU ordering.
 		s.touchRecord(digest)
 	}
 	s.hitMu.RUnlock()
--- a/internal/imagecache/imagecache.go
+++ b/internal/imagecache/imagecache.go
@@ -56,7 +56,9 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/authn"
@@ -74,6 +76,10 @@ const (
 	layerFSDirName           = "fs"
 	layerWhiteoutsFileName   = "whiteouts.json"
 	layerFinalizedMarkerName = "finalized"
+	// layerSizeFileName holds the layer's byte count, recorded at unpack so
+	// sizing the pool never walks a tree. Absent for layers unpacked by
+	// older atelets (backfilled lazily, see layerSize).
+	layerSizeFileName = "size"
 
 	// layerPullConcurrency bounds concurrent layer download+unpack streams per
 	// image pull. Memory use is O(stream buffers) per slot, independent of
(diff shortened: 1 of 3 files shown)


internal/imagecache/sizes.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		if d.Type().IsRegular() {
+			if info, err := d.Info(); err == nil {
+				total += info.Size()
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return 0, fmt.Errorf("while backfilling size of %q: %w", layerDir, err)
```
dberkov: nit:
this is a dead code and never executed.

filepath.WalkDir reports errors:
  - Error case, unreadable subdirectory → return fs.SkipDir (swallowed by WalkDir, not an error)
  - Error case, anything else → return nil
  - Success case, regular file → return nil (note that even the inner d.Info() failure is deliberately ignored)
  - Success case, everything else → return nil

---
  // The callback swallows every error (skip-and-under-count contract),
  // so WalkDir cannot return one.
  _ = filepath.WalkDir(fsRoot, func(...) error { ... })

  or keep the check but as an invariant guard with a comment saying it's unreachable by construction. Either is fine; keeping it silently as-is just leaves a small landmine for whoever refactors the callback later and assumes the error handling below is load-bearing.

internal/imagecache/imagecache.go
```diff
@@ -74,6 +76,10 @@ const (
 	layerFSDirName           = "fs"
 	layerWhiteoutsFileName   = "whiteouts.json"
 	layerFinalizedMarkerName = "finalized"
+	// layerSizeFileName holds the layer's byte count, recorded at unpack so
+	// sizing the pool never walks a tree. Absent for layers unpacked by
+	// older atelets (backfilled lazily, see layerSize).
```
dberkov: add following comment to the existing comment
"an estimate; the value is the tar-stream length when written at unpack, or the summed file sizes when backfilled, so it may differ from disk
  usage and across node"


---
#### Path 1 — recorded at unpack time (imagecache.go, in unpackLayerToPool):
  Here cr.n is the number of bytes in the uncompressed tar stream. That includes tar framing overhead: a 512-byte header per entry, padding of each file body up to a 512-byte boundary, and the end-of-archive blocks. On the other hand, a hardlink in a tar archive is just a header with no body, so a layer that hardlinks one big file many times contributes its content bytes only once.

#### Path 2 — backfilled later (sizes.go, in layerSize, for layers unpacked by an older atelet or when the file is missing/corrupt):
  Here the number is the sum of on-disk file content sizes. No tar framing at all — so it skews smaller than path 1 for layers with many small files. But hardlinked files show up as multiple regular directory entries that each report the full size, so the same content gets counted once per link — skewing larger than path 1 for hardlink-heavy layers. And if the walk hits an unreadable directory, its contents are skipped entirely — skewing smaller again.

internal/imagecache/sizes.go
```diff
@@ -0,0 +1,124 @@
...
+// layerSize returns the layer's recorded byte count, backfilling the size
+// file for layers unpacked before sizes were recorded.
+func (s *Store) layerSize(layerDir string) (int64, error) {
+	b, err := os.ReadFile(filepath.Join(layerDir, layerSizeFileName))
+	if err == nil {
+		n, perr := strconv.ParseInt(strings.TrimSpace(string(b)), 10, 64)
+		if perr == nil {
+			return n, nil
+		}
+		// Corrupt size file: fall through to backfill.
```
dberkov, line 67: nit (test coverage): This corrupt-file fall-through is the one recovery branch in layerSize without a test — nothing pins that a garbage size file triggers the backfill walk, or that the walk's result overwrites the corrupt file so the repair happens once rather than on every CacheSize call.
  
It's exactly the kind of branch a later "simplification" could silently break (e.g. collapsing the fast path to return strconv.ParseInt(...) would make a corrupt file skip the layer from CacheSize forever, with all current tests still green).
  
Cheap to cover — slots into TestLayerSizeRecordedAndBackfilled right after the delete-and-backfill sequence, reusing its fixture:

// Corrupt size file: healed the same way as a missing one.
if err := os.WriteFile(sizePath, []byte("not-a-number\n"), 0o600); err != nil {
     t.Fatal(err)
}
repaired, err := store.layerSize(img.LayerDirs[0])
if err != nil {
    t.Fatalf("layerSize on corrupt size file: %v", err)
}
if repaired < 4096 {
    t.Errorf("layerSize = %d, want >= 4096: corrupt file not healed by backfill", repaired)
}
if b, _ := os.ReadFile(sizePath); strings.TrimSpace(string(b)) == "not-a-number" {
    t.Error("backfill did not overwrite the corrupt size file")
}

internal/imagecache/sizes.go
igooch: Updated to remove the error and add the message that this function cannot return an error.

internal/imagecache/sizes.go
igooch: Good call, added the test.

internal/imagecache/imagecache.go
igooch: Added a comment about this being an estimate.

dberkov (approved):

PR #652, opened 2026-07-30
atenet: add option to run agentgateway

howardjohn:
Agentgateway is a popular [AAIF](https://aaif.io/) project designed for agentic systems. It is of particular interest for substrate users as it was designed specifically to serve the role of an egress proxy for agents and other AI-adjacent workloads, with strong support for credential injection, token exchanges, TLS MITM, and other authentication and authorization schemes.

Agentgateway supports the ext_proc protocol, like Envoy, so we use that here. There has been some debate in the community around the long term end state of ext_proc vs other options, but for now this maintains the status quo.

While Agentgateway can be configured dynamically over XDS, it does not take the same configuration as Envoy. However, none of the config is dynamic anyways, so we use just a static configuration file (configmap) for now to keep things simple.

While we don't have specific per-component/file OWNERS in the project at the moment, I can informally commit to myself and Eitan maintaining this integration. Given the goals around velocity in the project we can commit to either rapidly fixing any issues that may arise or removing/disabling the integration if this ends up slowing things down.

- [x] Tests pass


LiorLieberman:
John - this would require #559 to be merged first, wouldnt it?

howardjohn:
> John - this would require #559 to be merged first, wouldnt it?

No, there is no dependency and it works end to end currently. With 559 we will need a couple line change to make it send to the atunnel instead of directly but either one can merge first

EItanya:
> While Agentgateway can be configured dynamically over XDS, it does not take the same configuration as Envoy. However, none of the config is dynamic anyways, so we use just a static configuration file (configmap) for now to keep things simple.

I was actually wondering if we could do the same for Envoy as well, the dynamic config seems like more complexity than it's worth right now

cmd/atenet/internal/router/cmd.go
```diff
@@ -41,6 +41,7 @@ func NewRouterCmd() *cobra.Command {
 	cmd.Flags().StringVar(&cfg.LogLevel, "log-level", "info", "Log level: debug, info, warn, error")
 	cmd.Flags().StringVar(&cfg.MetricsAddr, "metrics-listen-addr", ":9090", "Address and port the prometheus metrics server should listen on.")
 	cmd.Flags().BoolVar(&cfg.Standalone, "standalone", false, "Run in standalone mode, bypassing creation of managed deployment and services in Kubernetes cluster")
+	cmd.Flags().StringVar(&cfg.AnetRouter, "anetrouter", anetRouterEnvoy, "Router dataplane: envoy or agentgateway")
```
EItanya: unfortunately I do think naming matters, and FWIW I actually kinda like `anet`, but is that the name we want to go with, or should we stick to `atenet`?

cmd/atenet/internal/router/config.go
```diff
@@ -70,6 +76,17 @@ type routerConfig struct {
...
 }
 
+func (c routerConfig) anetRouter() string {
+	if c.AnetRouter == "" {
+		return anetRouterEnvoy
+	}
+	return c.AnetRouter
+}
+
+func (c routerConfig) usesEnvoy() bool {
```
EItanya: any specific reason you went with `usesEnvoy` rather than just using a go style enum, or a type alias string with switch cases?

EItanya (commented):
Just a couple of small naming things on my end, overall LGTM

howardjohn:
> John - this would require #559 to be merged first, wouldnt it?

Rebased on 559

cmd/atenet/internal/router/config.go
```diff
@@ -109,9 +124,17 @@ func (c routerConfig) extProcMaxRequests() int {
...
+	case atenetRouterEnvoy, atenetRouterAgentgateway:
+	default:
+		return fmt.Errorf("--atenet-router must be %q or %q, got %q", atenetRouterEnvoy, atenetRouterAgentgateway, c.AtenetRouter)
+	}
 	if err := c.ParkedRequest.validate(); err != nil {
 		return err
 	}
+	if c.atenetRouter() == atenetRouterAgentgateway {
+		return nil
+	}
```
LiorLieberman: add a comment why this is not relevant for agw and/or what it does instead?

cmd/atenet/internal/router/config_test.go
```diff
@@ -29,6 +29,15 @@ func TestRouterConfigValidate(t *testing.T) {
 			name: "defaults are valid (auto breaker)",
 			cfg:  routerConfig{ExtProcMaxRequests: 0, ParkedRequest: ParkedRequestConfig{Max: defaultParkedRequestMax}},
 		},
+		{
+			name: "agentgateway is valid",
+			cfg:  routerConfig{AtenetRouter: string(atenetRouterAgentgateway), ParkedRequest: ParkedRequestConfig{Max: defaultParkedRequestMax}},
```
LiorLieberman: add a testcase for atenetRouterEnvoy? (both as explicit, and both as a default). Maybe add a atenetRouter field in the test case or this could be better tested on a different func.

nit: change the name of the test to atenet-router set to agentgateway is valid, atenet-router set to envoy is valid.

cmd/atenet/internal/router/extproc_out.go
```diff
@@ -53,6 +54,31 @@ func addOriginalDstMutation(dst string, mut *extproc.HeaderMutation) {
...
+		Header: &corev3.HeaderValue{
+			Key:      atunnel.OriginalHostHeader,
+			RawValue: []byte(actorHost),
+		},
+	})
+	if routeViaAuthority {
+		mut.SetHeaders = append(mut.SetHeaders, &corev3.HeaderValueOption{
+			AppendAction: corev3.HeaderValueOption_OVERWRITE_IF_EXISTS_OR_ADD,
+			Header: &corev3.HeaderValue{
+				Key:      ":authority",
```
LiorLieberman: nit: we have a const for authority likely?

internal/atunnel/server.go
```diff
@@ -258,6 +267,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		cancel()
 	}()
 
+	// Do not expose the router-only routing header to actor code. Restore Host
+	// so dataplanes that route dynamically on worker IP still give the actor its
+	// stable mesh hostname.
```
LiorLieberman: mesh?

manifests/ate-install/base/kustomization.yaml
LiorLieberman, line 1: why embedding this under a gke specific director?

LiorLieberman (commented):
overall looks good to me, lots of switch statements/if statements throughout. Is there a way we can make it more self-contained? that _most_ dataplanes logic happens within specific packages?

LiorLieberman (commented):
Can you make the tests happy? this is LGTM. 
/cc @bowei for approval

howardjohn:
> Can you make the tests happy? this is LGTM. /cc Bowei Du (@bowei) for approval

It was an unrelated test flake; I rebased and now its green

bowei:
Which test flaked? We need to fix flakes ASAP..

howardjohn:
opened https://github.com/agent-substrate/substrate/issues/716

Issue #619, opened 2026-07-30
TestDurableDirLifecycle e2e test flaking


BenTheElder:
spotted in post-merge test runs, which should all be passing:
https://github.com/agent-substrate/substrate/commits/main/

https://github.com/agent-substrate/substrate/actions/runs/30482958208/job/90681436378

BenTheElder:
this appears to be surfacing a flake in the actual system

BenTheElder:
Will send a PR ~monday. (need to refine further and I'm out tomorrow-sunday).

PR #654, opened 2026-07-30
e2e,ci: keep a failed run's namespaces and dump their worker logs

BenTheElder:
When an e2e test failed, the evidence was deleted before anyone could read it. The suite deleted every namespace it created on the way out, taking the worker pods with it, and the workflow's post-failure dump only looked at three fixed namespaces — never the suites' randomly-named ones. So a failure inside an actor (#619: a micro-VM resume where the guest died at boot) left nothing behind but the RPC error the test printed.

Keep the namespaces when the suite failed, and dump every worker pod in every namespace, so the ateom logs — which carry the guest's console tail — reach the failed run's output.

Kept namespaces are nobody's to reclaim, and each holds a WorkerPool's worth of running pods, so they now carry an ate.dev/e2e label and hack/cleanup-e2e.sh deletes them once the logs have served their purpose. CI throws its cluster away, but a development cluster accumulates them run after run.

Built while debugging https://github.com/agent-substrate/substrate/issues/619

- [x] Tests pass

### DIFF
--- a/.github/workflows/pr-workflow.yaml
+++ b/.github/workflows/pr-workflow.yaml
@@ -114,12 +114,20 @@ jobs:
       if: failure()
       run: |
         kubectl --context kind-kind get actortemplate,workerpool,pods -A -o wide || true
-        for ns in ate-demo-counter-microvm ate-demo-counter ate-system; do
-          for p in $(kubectl --context kind-kind get pods -n "$ns" -o name 2>/dev/null); do
-            echo "=== logs: $ns/$p ==="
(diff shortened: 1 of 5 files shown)


BenTheElder:
ugh, we have to fix some existing issues with new restrictions on the actions first

BenTheElder:
https://github.com/agent-substrate/substrate/pull/657

LiorLieberman:
rebase?

LiorLieberman (approved):

PR #656, opened 2026-07-31
imagecache: record image digest in bundle overlay specs; atomic WriteSpec

igooch:
Second foundation slice of #463 Phase 2 (GC), independent of #650 and inert on its
own: nothing reads the new field and nothing deletes anything.

- **`OverlaySpec.ImageDigest`** (optional, `omitempty`): the manifest digest the
  bundle's image ref resolved to, populated in `prepareOCIDirectory`. The GC
  root-set scan will use it to root the whole image record while the bundle
  exists. Compatible in both directions: ateom consumers never read the field
  (old binaries drop the unknown JSON key), and specs written by older atelets
  keep parsing with the field empty — so there is no deploy-ordering constraint.
- **Atomic `WriteSpec`** (temp file + rename in the bundle dir): the root-set scan
  will read specs concurrently with bundle preparation, and a torn spec would
  under-report the layers an actor is using — an error in the dangerous
  direction (toward deleting in-use layers). With the rename, a reader sees the
  complete old spec or the complete new one, never a fragment.

Landing early matters: every bundle written from now on carries the digest,
shrinking the digestless-spec population the GC's compatibility fallback has to
cover.

**Testing**: round-trip + old-spec compat and no-temp-files-left unit tests;
`go test -race` green on both affected packages.

### DIFF
--- a/internal/imagecache/spec.go
+++ b/internal/imagecache/spec.go
@@ -38,9 +38,11 @@ const OverlaySpecFileName = "rootfs-overlay.json"
 type OverlaySpec struct {
 	Version int `json:"version"`
 	// ImageDigest is the manifest digest the bundle's image ref resolved to
-	// ("sha256:<hex>"). The cache GC's root-set scan uses it to protect the
-	// image while the bundle exists; consumers ignore it. Optional: older
-	// specs lack it.
+	// ("sha256:<hex>"). For a multi-arch ref this is the index digest; the
+	// cache may hold a twin record under the platform-child digest, and GC
+	// must treat the pair as one image. The GC's root-set scan uses it to
+	// protect the image while the bundle exists; consumers ignore it.
+	// Optional: older specs lack it.
 	ImageDigest string `json:"imageDigest,omitempty"`
 	// Layers are the cached layer directories (each holding its tree under
 	// fs/), bottom-most layer first — the order the image manifest lists
--- a/internal/imagecache/spec.go
+++ b/internal/imagecache/spec.go
@@ -55,9 +55,10 @@ type OverlaySpec struct {
 
 // WriteSpec writes spec into the bundle at bundlePath.
 //
-// The write is atomic (temp file + rename): the cache GC's root-set scan
-// reads these files concurrently, and a torn spec would under-report the
-// layers an actor is using — failing toward deletion.
+// The write is atomic (temp file + rename): concurrent readers — notably
+// the cache GC's root-set scan — must never see a partial spec, which
+// could parse with layers missing and leave them eligible for eviction
+// while the actor is using them.
 func WriteSpec(bundlePath string, spec *OverlaySpec) error {
 	spec.Version = 1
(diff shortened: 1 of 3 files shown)


internal/imagecache/spec_test.go
```diff
@@ -131,3 +134,50 @@ func TestCreateExtraDirs(t *testing.T) {
...
+	// Round trip with the new field.
+	if err := WriteSpec(dir, &OverlaySpec{ImageDigest: "sha256:" + strings.Repeat("ef", 32), Layers: []string{"/pool/layers/sha256/aa"}}); err != nil {
+		t.Fatalf("WriteSpec: %v", err)
+	}
+	spec, err := ReadSpec(dir)
+	if err != nil || spec == nil {
+		t.Fatalf("ReadSpec: %v, %v", spec, err)
+	}
+	if spec.ImageDigest == "" {
+		t.Error("ImageDigest lost in round trip")
```
dberkov: nit (test strength): asserting non-emptiness misses field-mapping bugs that equality would catch for free (wrong JSON tag, value bleeding from another field). Since the written value is right there:

if want := "sha256:" + strings.Repeat("ef", 32); spec.ImageDigest != want {
    t.Errorf("ImageDigest = %q, want %q", spec.ImageDigest, want)
}

internal/imagecache/spec.go
```diff
@@ -37,6 +37,11 @@ const OverlaySpecFileName = "rootfs-overlay.json"
 // bundle path by the consumer rather than trusted from the file.
 type OverlaySpec struct {
 	Version int `json:"version"`
+	// ImageDigest is the manifest digest the bundle's image ref resolved to
+	// ("sha256:<hex>"). The cache GC's root-set scan uses it to protect the
+	// image while the bundle exists; consumers ignore it. Optional: older
+	// specs lack it.
+	ImageDigest string `json:"imageDigest,omitempty"`
```
dberkov, line 46: doc suggestion (for the GC slice, not this PR's behavior): for a multi-arch ref this is the index digest, but pull() also records the image under the platform-child digest — so rooting only this digest protects one of the two twin records; the child twin still looks unrooted/cold to a naive record scan. Layers are safe regardless (the spec's Layers list roots them directly, and both twins share diffIDs), so this only affects record-file GC — but worth a sentence here so the root-set scan treats twins as one image, e.g.:

// ImageDigest is the manifest digest the bundle's image ref resolved to
// ("sha256:<hex>"). For a multi-arch ref this is the index digest; the
// cache may hold a twin record under the platform-child digest, and GC
// must treat the pair as one image. The GC's root-set scan uses it to
// protect the image while the bundle exists; consumers ignore it.
// Optional: older specs lack it.

internal/imagecache/spec_test.go
igooch: Added

internal/imagecache/spec.go
igooch: Updated the comment

dberkov (approved):

PR #658, opened 2026-07-31
imagecache: two-phase layer retirement primitives + .rm-* startup sweep

igooch:
Third foundation slice of #463 Phase 2 (GC), independent of #650 and #656. The
primitives are uncalled until the eviction engine PR; the only live behavior is
the startup sweep, which only ever sees `.rm-*` dirs the engine will create.

- **`retireLayer`**: evicts a layer with one atomic rename to a `.rm-*` name
  inside the layer singleflight; the slow `RemoveAll` is the caller's job,
  outside all locks. Existence/mtime pre-flight runs *outside* the flight (a
  retire must never block behind an in-progress download) and both checks are
  re-run inside it before the rename. Returns gone/vetoed/retired so the engine
  can distinguish "nothing there" from "must keep".
- **Reuse interlock**: `ensureLayer` now refreshes the layer dir mtime inside
  the same flight, so a retirement and a reuse cannot interleave — either the
  retire wins (the pull re-unpacks) or the touch wins (the retire vetoes).
- **Startup sweep**: `.rm-*` dirs (a crash between rename and removal) are
  reclaimed alongside the existing `.tmp-*` dirs, via `RemoveAllWritable` since
  layer trees legitimately contain read-only content.
- **`isLayerDirName`** validates names read from disk or records.

Divergence from the prototype: `retireLayer` reports no freed size — size
crediting belongs to the engine's call sites and arrives with it.

**Testing**: unit tests for the status contract, startup-sweep recovery and
non-interference, and a retire-vs-pull race test; `go test -race` green.

### DIFF
--- a/internal/imagecache/imagecache.go
+++ b/internal/imagecache/imagecache.go
@@ -192,23 +192,33 @@ func (s *Store) recordPath(digest v1.Hash) string {
 	return filepath.Join(s.root, "manifests", digest.Algorithm, digest.Hex+".json")
 }
 
-// sweepTempDirs removes unpack temp dirs and manifest-record temp files
-// orphaned by a crash. A layer dir without the temp prefix and a record
-// without a leading dot are always complete (both are moved into place with
-// a single rename), so this is the only recovery the pool needs.
+// sweepTempDirs removes unpack temp dirs, retired layer dirs, and
+// manifest-record temp files orphaned by a crash. A layer dir without the
+// temp/retired prefix and a record without a leading dot are always
+// complete (both are moved into place with a single rename), so this is
+// the only recovery the pool needs.
 func (s *Store) sweepTempDirs() error {
 	entries, err := os.ReadDir(s.layersDir())
 	if err != nil {
 		return fmt.Errorf("while listing layer pool: %w", err)
 	}
+	swept := 0
 	for _, e := range entries {
-		if !strings.HasPrefix(e.Name(), ".tmp-") {
+		// ".tmp-": an unpack in flight at crash. ".rm-": a layer retirement
+		// renamed aside but not yet removed. Either way, unreferenced
+		// garbage.
+		if !strings.HasPrefix(e.Name(), ".tmp-") && !strings.HasPrefix(e.Name(), retiredPrefix) {
 			continue
 		}
 		p := filepath.Join(s.layersDir(), e.Name())
+		slog.Info("Image cache sweeping orphaned layer dir", slog.String("dir", e.Name()))
 		if err := RemoveAllWritable(p); err != nil {
 			return fmt.Errorf("while sweeping orphaned layer temp dir %q: %w", p, err)
 		}
+		swept++
+	}
+	if swept > 0 {
+		slog.Info("Image cache startup sweep removed orphaned layer dirs", slog.Int("count", swept))
 	}
 
 	// writeRecord's temp files are ".<hex>.json.tmp-<rand>"; finished records
@@ -386,6 +396,12 @@ func (s *Store) ensureLayer(ctx context.Context, diffID v1.Hash, layer v1.Layer)
 	dir := s.layerDir(diffID)
 	_, err, _ := s.layerSF.Do(diffID.String(), func() (any, error) {
 		if _, err := os.Stat(filepath.Join(dir, layerFSDirName)); err == nil {
+			// Refresh the dir mtime inside the flight: retireLayer re-checks
+			// the mtime in this same flight, so a layer reused here can
+			// never be renamed away between this stat and the image record
+			// that will re-reference it.
+			now := time.Now()
+			_ = os.Chtimes(dir, now, now)
 			return nil, nil
 		}
 		return nil, s.unpackLayerToPool(ctx, diffID, layer)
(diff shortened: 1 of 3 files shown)


internal/imagecache/retire.go
```diff
@@ -0,0 +1,137 @@
...
+		return "", retireVetoed, err
+	}
+	if fi.ModTime().After(cutoff) {
+		slog.Info(logMsgLayerRetireVetoed, slog.String("diffid", hex), slog.Time("last_used", fi.ModTime()))
+		return "", retireVetoed, nil
+	}
+
+	var retired string
+	status := retireGone
+	_, err, _ = s.layerSF.Do(v1.Hash{Algorithm: "sha256", Hex: hex}.String(), func() (any, error) {
```
dberkov: nit:
singleflight.Do shares one execution across concurrent callers on the same key — and this key is deliberately shared with ensureLayer. If this call joins a flight already in progress (an ensureLayer reuse-touch, or another retire), our closure never runs, so status/retired keep their zero-init values and we report retireGone + "" regardless of what actually happened — e.g. "gone, nothing stranded" while the layer sits in the pool with a fresh mtime (truthful answer: retireVetoed).

 Consistency is never violated (nothing renamed or lost), so this is an accounting/reporting wart — but Do's third return value is exactly the discarded signal, and a shared flight means "someone was concurrently active on this layer," which is what a veto is:

 _, err, shared := s.layerSF.Do(layerFlightKey(hex), func() (any, error) { ... })
 if shared {
     // Our closure didn't run — a concurrent ensureLayer or retire owned
     // the flight. Concurrent activity is precisely a veto.
     return "", retireVetoed, nil
 }

internal/imagecache/imagecache.go
```diff
@@ -386,6 +396,12 @@ func (s *Store) ensureLayer(ctx context.Context, diffID v1.Hash, layer v1.Layer)
 	dir := s.layerDir(diffID)
 	_, err, _ := s.layerSF.Do(diffID.String(), func() (any, error) {
 		if _, err := os.Stat(filepath.Join(dir, layerFSDirName)); err == nil {
+			// Refresh the dir mtime inside the flight: retireLayer re-checks
+			// the mtime in this same flight, so a layer reused here can
+			// never be renamed away between this stat and the image record
+			// that will re-reference it.
+			now := time.Now()
+			_ = os.Chtimes(dir, now, now)
```
dberkov: nit: this silent discard swallows the error the reuse-veto depends on. Failure is benign (a wrongly-retired layer self-heals — cachedImage returns nil on a missing layer dir and the image re-pulls), but the package convention for best-effort metadata writes is warn-and-continue:

if err := os.Chtimes(dir, now, now); err != nil {
  slog.WarnContext(ctx, "Failed to refresh layer mtime on reuse", slog.String("diffid", diffID.String()), slog.Any("err", err))
}

internal/imagecache/imagecache.go
```diff
@@ -192,23 +192,33 @@ func (s *Store) recordPath(digest v1.Hash) string {
...
+		// ".tmp-": an unpack in flight at crash. ".rm-": a layer retirement
+		// renamed aside but not yet removed. Either way, unreferenced
+		// garbage.
+		if !strings.HasPrefix(e.Name(), ".tmp-") && !strings.HasPrefix(e.Name(), retiredPrefix) {
 			continue
 		}
 		p := filepath.Join(s.layersDir(), e.Name())
+		slog.Info("Image cache sweeping orphaned layer dir", slog.String("dir", e.Name()))
 		if err := RemoveAllWritable(p); err != nil {
 			return fmt.Errorf("while sweeping orphaned layer temp dir %q: %w", p, err)
```
dberkov: nit: this branch now also handles retired dirs, for which "temp dir" is misleading:

 return fmt.Errorf("while sweeping orphaned layer dir %q: %w", p, err)

internal/imagecache/retire.go
dberkov: nit: v1.Hash{Algorithm: "sha256", Hex: hex}.String() must stay in lockstep with ensureLayer's diffID.String() — true today, but nothing enforces it. A shared helper makes the invariant structural:

 // layerFlightKey is the singleflight key shared by ensureLayer and
// retireLayer; the retire/reuse interlock depends on both using it.
func layerFlightKey(hex string) string { return "sha256:" + hex }

internal/imagecache/imagecache.go
igooch: Added.

internal/imagecache/retire.go
igooch: Fixed, but not with the `shared` flag patch. The subtlety: `singleflight.Do`'s shared return is also true for the caller whose closure did run when others joined it mid-flight. Gating on shared would therefore report `retireVetoed` after actually renaming the dir — losing the retired path (leaking the `.rm-*` dir until the next restart) and telling the future engine "kept" about a layer that's gone, which would trigger a record restore pointing at a missing layer. Instead, a ran flag set inside the closure distinguishes the case: closure never executed → status/retired are stale zero values → return retireVetoed.

igooch: Added.

internal/imagecache/imagecache.go
igooch: updated.

dberkov (approved):

PR #659, opened 2026-07-31
fix flaky TestPlatformMetrics test

### LINKED ISSUE #634: E2E test flake: TestPlatformMetricsEmitted
e2e test flake:

* failure: https://github.com/agent-substrate/substrate/actions/runs/30498992688/job/90734205524

```
--- FAIL: TestPlatformMetricsEmitted (121.23s)
FAIL
FAIL	github.com/agent-substrate/substrate/internal/e2e/suites/metrics	121.396s
```

LiorLieberman:
Cut the OTel metric export interval to 10s in the kind stack, plumbing OTEL_METRIC_EXPORT_INTERVAL from ate-controller through to the ateom worker pods it creates. ateom registers no instruments of its own, so it stays invisible to the collector until its first export tick - which consumed ~50s of TestPlatformMetricsEmitted's 120s budget and made the suite flaky on loaded CI runners. 

also removes a dead `service_name=` in CollectorHasService, since the Prometheus exporter only ever emits that resource attribute as the job label.

- [ ] Tests pass

### DIFF
--- a/cmd/atecontroller/internal/controllers/workerpool_apply.go
+++ b/cmd/atecontroller/internal/controllers/workerpool_apply.go
@@ -43,6 +43,9 @@ type ateomOTelSettings struct {
 	// first export tick fires. Shortening the interval is what keeps that
 	// startup gap inside an e2e budget; production leaves it unset.
 	MetricExportInterval string
+	// MetricExportTimeout overrides the SDK's  per-export timeout, in the same
+	// whole-millisecond form as MetricExportInterval. Empty keeps the default.
+	MetricExportTimeout string
 }
 
 // buildDeploymentApplyConfig constructs the SSA apply configuration for the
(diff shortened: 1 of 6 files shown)


krisztianfekete (commented):
Ah, opened #661 for the same flake before spotting this one, sorry! Same approach in both, so happy to close mine. One thing worth pulling in from #661: in the failed run both export ticks actually fired and failed (first one used the whole 30s default export timeout, second died on a DNS timeout), so interval alone still leaves workers at ~1 attempt per 30s during a blip. Can you also set + propagate `OTEL_METRIC_EXPORT_TIMEOUT=10000` the same way? Then this covers everything. 

The `CollectorHasService` cleanup looks right too, the kind collector has no resource_to_telemetry_conversion so service_name= never matches.

EItanya:
LGTM, but will wait for @krisztianfekete to give an approval

LiorLieberman:
Thanks! @krisztianfekete just added - can you take a look?

krisztianfekete (approved):
Thanks, LGTM!

Issue #643, opened 2026-07-30
Recovery from failed resume, crash scenarios


msau42:
Related issue: https://github.com/agent-substrate/substrate/issues/641

Once an Actor is in RESUMING status, the only allowed transition paths are:
- Success scenario: Actor becomes RUNNING
- Crash scenario: Actor becomes CRASHED

That means if some error happened during Resume, the only way out is for users to keep on retrying until it succeeds, or abandon and leak the actor.

https://github.com/agent-substrate/substrate/issues/119 proposes a "revert" API, which could help with restoring to the last known state, but even that could also run into errors.

I think as a final escape hatch, we should provide some  "force delete" option. where we can 1) trigger (perhaps periodically) a cleanup sequence on nodes and CP 2) still keep track of that actor and potentially leaked resources, but allow for new actors with the same name to be created.

To accomplish this, we probably have to:
- Track force deleting actors under a different key, maybe we prefix the actor UID with some keyword like "forcedeleting"
- Have periodic garbage collectors in both CP and atelet. 
- Atelet only GCs actor UIDs that have been marked as forcedeleting. 
- CP GC waits for atelet GC for that actor UID to complete. This implies that we need to keep track of which atelet the actor was running on and also that we wouldn't have allowed the actor to leak on multiple nodes.

Any other thoughts/ideas?

msau42:
Untested POC here: https://github.com/agent-substrate/substrate/compare/main...msau42:substrate:force-delete

For now I decided to focus on adding just a cleanup workflow, but not solve for the "allow creating same actor name while deletion is in progress" problem.

zoez7:
> but allow for new actors with the same name to be created.

yeah this is going to be challenging, for now maybe we can just return an error with "previous actor is still in wipeout, cannot create actor with same name".

Sneha-at:
There are 2 issues here
1. Allow an escape hatch to force "delete" an actor from any state. This will transition the state to DELETE state (It might still fail in DELETE state)
2. Have a GC running in the background that cleans up all the leaked resources for deleted actors.

https://github.com/agent-substrate/substrate/pull/788 tracks the 1st point in this issue.

PR #666, opened 2026-07-31
Make UnmountVolume a hard crash failure during suspend/pause.

msau42:
Don't try to rollback mounts during Resume/Run. Instead should introduce a way to cleanup actor resources when an actor is in a failed state. Discussion in #643.

- [x] Tests pass

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -258,12 +258,6 @@ func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (resp *
 	if err := s.mountExternalVolumes(ctx, actorUID, req.GetSpec().GetVolumes()); err != nil {
 		return nil, err
 	}
-	defer func() {
-		if err != nil {
-			// TODO cleanup orphaned volumes
-			_ = s.unmountExternalVolumes(ctx, actorUID, req.GetSpec().GetVolumes())
-		}
-	}()
 
 	// Record the sandbox binaries this actor is running so a later Checkpoint
 	// (whose request no longer carries the sandbox config) can re-fetch the same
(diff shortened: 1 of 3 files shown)


cmd/atelet/main.go
```diff
@@ -409,8 +403,9 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 		return nil, fmt.Errorf("unexpected checkpoint type: %v", req.GetType())
 	}
 
-	// TODO cleanup orphaned volumes
-	_ = s.unmountExternalVolumes(ctx, actorUID, req.GetSpec().GetVolumes())
+	if err := s.unmountExternalVolumes(ctx, actorUID, req.GetSpec().GetVolumes()); err != nil {
```
zoez7, line 406: This looks like a simple filesystem operation in the mock impl. https://github.com/agent-substrate/substrate/blob/860250bf7105261295533e8c19e6f5409d1a46f2/internal/volume/mock.go#L112

But I'm wondering in the future if we fail to unmount the volume, do we need to set the Actor as "crashed" (is there a risk of losing data that actor has written?)

cmd/atelet/main.go
```diff
@@ -258,12 +258,6 @@ func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (resp *
 	if err := s.mountExternalVolumes(ctx, actorUID, req.GetSpec().GetVolumes()); err != nil {
 		return nil, err
 	}
-	defer func() {
-		if err != nil {
-			// TODO cleanup orphaned volumes
-			_ = s.unmountExternalVolumes(ctx, actorUID, req.GetSpec().GetVolumes())
-		}
-	}()
```
zoez7, line 266: Wait... this was in AteomHerder.Run, so we were removing external volumes right after telling ateom to run the workload?

zoez7 (dismissed):

cmd/atelet/main.go
msau42: Previous behavior tried to rollback operations on failure to try to cleanup node state.

So if Run() encountered some failure,  then we would try to undo the mounts to cleanup.

But now I'm deciding that it's better to just leave the node in a partial state, and have an explicit cleanup flow to handle the cleanup.

cmd/atelet/main.go
msau42: As part of this PR, I am changing unmount failure to be a hard failure of pause/suspend. So the user has to retry Pause/Suspend again until it succeeds.

There are many causes of unmount failure, some are temporary, some could be permenent.

Do you think it's better to just crash on any failure even if it's temporary?

zoez7: yeah, I'm thinking we can start with crashing on any failure, and only exempt the ones we know are not permanant and are retriable

msau42: done

zoez7 (approved):

PR #667, opened 2026-07-31
ateom: add GetWorkloadStats RPC and retain actor attribution

### LINKED ISSUE #594: Phase 0: StatsWorkload RPC on ateom (proto, both runtime reads, identity retention)
Part of #550. First of the four phases in the [design comment](https://github.com/agent-substrate/substrate/issues/550#issuecomment-5106174074):

| Phase | Scope | Unblocks |
|---|---|---|
| **0** | **Stats method: proto, both reads, identity retention in ateom** | **Independently testable, nothing user-facing** |

## Why the read lives in ateom

atelet is a node-level DaemonSet whose only host mount is `/var/lib/ateom-gvisor` — no `/sys/fs/cgroup`, no `hostPID`. The per-actor cgroup leaves are only addressable as `/sys/fs/cgroup/<containerName>` from inside the worker pod's own cgroup namespace, and the kata-agent vsock is only reachable from the ateom that booted the VMM. So the read has to originate in ateom.

Phase 0 delivers only the RPC that later phases call. atelet's reader, the aggregation, and the OTel emission are Phase 1+.

## Scope

Add `StatsWorkload` to `ateom.Ateom` (`internal/proto/ateompb/ateom.proto`) — a pure read that does not change the ateom's available/executing state and is safe to call on a timer:

```proto
rpc StatsWorkload(StatsWorkloadRequest) returns (StatsWorkloadResponse) {}

message StatsWorkloadRequest {
  // The actor the caller believes is running here. ateom rejects a mismatch
  // with FAILED_PRECONDITION rather than reporting another actor's numbers.
  string actor_uid = 1;
}
```

`StatsWorkloadResponse` as specified in the design comment: the five identity fields, `sandbox_class`, `source`, `memory_current_bytes`, `memory_peak_bytes`, `memory_working_set_bytes`, `cpu_usage_usec`, `observed_at_unix_nano`.

**gVisor read** — from `/sys/fs/cgroup/pause`: `memory.current`, `memory.peak`, `memory.stat:inactive_file`, `cpu.stat:usage_usec`. `memory.peak` must be optional (kernel 5.19+). `pause/` and not `main/` because `setupCgroupForSubcontainer` installs application-container leaves with empty resources and every application process actually runs inside the sentry, which is a single process in `pause/`.

**Micro-VM read** — via the kata guest agent's `StatsContainer`, not the host cgroup. Guest RAM is a fixed allocation (default 2048 MiB), so the host cgroup is roughly constant regardless of actor behaviour, and `Shared: true` backs it with a memfd — putting those pages on the file LRU, which makes the working-set subtraction cancel out idle guest RAM and understate the footprint. `StatsContainerRequest`/`Response`/`CgroupStats` are already in the vendored `cmd/ateom-microvm/internal/third_party/kata/agentpb`.

**Identity retention** — ateom must be able to answer with the identity from the original `RunWorkloadRequest`.

## Two corrections to the design comment

1. It says *"`ateom-microvm` already keeps them in `running`."* It does not. `runningActor` (`cmd/ateom-microvm/run.go:48-83`) holds `baseID`, `chCmd`, `vfsdCmd`, `durableVfsdCmd`, `apiSocket`, `restoreSourceDir`, `logAgent` — no identity fields. The identity *is* already assembled as `actorBootParams` (`run.go:224-231`, exactly the five fields) and flows into `coldBootActor` / `restoreFullScope`, so retention is cheap — but it is real work on **both** runtimes, not just gVisor.
2. `runningActor.logAgent` is a live kata-agent ttrpc client held open for the actor's lifetime (closed only by `teardownActor`), so the micro-VM read needs no new connection — it reuses `logAgent`.

## Planned PRs

1. **Proto, identity retention, `Unimplemented` stub.** `ateom.proto` + regen; `workloadIdentity` on the gVisor `AteomService` (set in `RunWorkload`/`RestoreWorkload`, cleared in `CheckpointWorkload` and on the `retErr` cleanup paths) and on `runningActor` for micro-VM. Retention covered by unit tests; not meaningfully testable in a live cluster until PR 2.
2. **gVisor cgroup read.** New `cmd/ateom-gvisor/internal/cgroupstats` with a caller-supplied root directory, so the parser is unit testable from a fixture tree without root.
3. **Micro-VM guest-agent read.** `AgentClient.StatsContainer` alongside the existing thin wrappers; one call per container keyed by `overlayWorkloadID(name)` (each workload has its own guest cgroup `/ateomchv/<workloadID>`), summed into the flat response. Note the unit conversion: the agent reports CPU in **nanoseconds**, the response field is **microseconds**.

## Out of scope

Per-container breakdown for gVisor (needs `runsc events` / `ContainerUsage` from the sentry — Phase 3). `ate.workerpool.name` as a label: agreed in the thread to include it early, but `ateletpb` carries no workerpool field today, so plumbing it would widen this into atelet and its callers. Adding it to `ateompb` later is a backward-compatible proto change — deferring to Phase 1/2.

baizhenyu:
Phase 0 of #550, tracked by #594. First of three stacked PRs.

Adds `ateom.Ateom/GetWorkloadStats`, the RPC atelet will poll for per-actor resource
usage, plus the actor attribution both ateom runtimes need to label a sample. The measurement half of each runtime lands in the two follow-ups;
`GetWorkloadStats` returns `Unimplemented` until then, so this change puts nothing
half-populated on the wire.

### The RPC

`GetWorkloadStats` is a **pure read**: unlike `RunWorkload` / `CheckpointWorkload` /
`RestoreWorkload` it does not move the ateom between "available" and "executing",
so it is safe to call on a timer for a workload's whole lifetime.

The request carries the actor UID the caller believes is executing here. A worker
can be recycled between atelet's view of the world and the call landing, so a
mismatch is rejected with `FAILED_PRECONDITION` rather than reporting a different
actor's numbers under the requested actor's identity. `FAILED_PRECONDITION` is
also the answer when the ateom is available — there is nothing to measure.

The response is one sample, measured at **sandbox** granularity (which today
equals the actor; per-container attribution would need the gVisor sentry's own
accounting). It echoes the measured actor's identity so the caller can attribute
the sample without holding its own worker→actor mapping, and carries
`sandbox_class` and `source` so two differently-measured numbers are not silently
compared as though they were the same thing. Both are enums — closed sets the
ateom binary picks from, where a typo in a free-form string would silently split
a metric in two downstream.

### Attribution retention

Neither runtime kept the actor's identifying fields past the call that started the
workload — they arrive on Run/Restore and nothing downstream needed them. A usage
sample is only useful once attributed, so both now hold them for as long as they
are executing:

* **ateom-gvisor** gains `AteomService.activeActor`, set by `RunWorkload` and
  `RestoreWorkload`, cleared by `CheckpointWorkload` and by both boot-failure
  paths — tracking exactly the available/executing state machine on the service.
* **ateom-microvm** gains `runningActor.activeActor`, populated from the existing
  `actorBootParams` on both the cold-boot and restore paths; the existing delete
  from `s.running` in `teardownActor` clears it.

The extraction from the request is shared in `internal/ateomstats`, since both
binaries need the same mapping. `ActorAttribution` composes the existing
`resources.ActorRef` rather than repeating its two fields, so the actor's
`Atespace` stays visibly attached to the actor — worth noting because
`TemplateNamespace` beside it is an unrelated namespace (a Kubernetes one;
`ActorTemplate` is a namespaced CRD, while an atespace is Substrate's own
tenancy unit).

It is deliberately **not** called `ActorIdentity`: "actor identity" already means
a credential in this repo — the `ateapi.ActorIdentity` service, `substratex509`,
`ateompath.ActorIdentityDirPath`, and #670 building on all three. Nothing in this
type is a secret or is presented as proof of anything; it is the tuple a usage
sample is labeled with.

### Reviewer notes

* **`activeActor` has no reader in this PR.** It is written and cleared but never
  read, because both `GetWorkloadStats` bodies are still stubs. The readers arrive
  with the measurement in the two follow-ups; the stub comments point there.
* **No metric labels are added here.** The response carries actor and atespace
  identity, but that is a sample, not a datapoint. Per the decisions in #174,
  actor/atespace identity belongs in logs and traces rather than as TSDB labels,
  and only template-level dimensions become metric labels. That conversion is
  atelet's, in Phase 1/2.
* **`cpu_usage_usec` resets on restore.** A restore recreates the sandbox, so the
  counter restarts at zero while the actor UID stays the same. Documented on the
  field; callers computing deltas must treat a decrease as a reset. If that proves
  too subtle in practice, an explicit epoch marker is a backward-compatible
  addition later.
* **`workerpool_name` is deliberately absent.** atelet does not know it today;
  adding the field later is backward compatible.

### Testing

* `internal/ateomstats` — `TestActorAttributionFromRequest` pins the mapping from
  both request types, including the empty and nil cases (the callers are not
  defensive about the request pointer, so the nil-safety of the generated getters
  is load-bearing). Five deliberately distinct placeholder values, so a field
  wired to the wrong source is visible.
* `cmd/ateom-microvm` — `TestActorBootParamsAttribution` pins the
  `actorBootParams` → `ActorAttribution` mapping, and
  `TestActorBootParamsAttributionMatchesRequest` checks that the two hops (request
  → boot params → attribution, written in different files) compose back into what
  the caller sent.
* Both runtimes — `TestGetWorkloadStatsUnimplemented` pins the stub's advertised
  contract, and `TestAteomServiceStartsAvailable` checks a fresh gVisor service
  retains no attribution (a non-nil zero value there would make an idle ateom
  report an empty actor's usage instead of refusing).
* **Not unit-tested:** the gVisor set/clear transitions. `RunWorkload`,
  `RestoreWorkload` and `CheckpointWorkload` each reach for netlink, runsc and the
  worker pod's netns within a few lines of entry, so there is no seam to drive
  them from `go test`. Noted in `cmd/ateom-gvisor/stats_test.go` rather than
  covered with a fake; the transitions get verified end to end once
  `GetWorkloadStats` returns real data.

Both ateom packages are `//go:build linux`, so their tests do not execute on
darwin. They were run natively on a Linux host in addition to the local
compile-check, along with `hack/verify/shellcheck.sh` (which needs docker).

### Still ahead

* **PR 2** — gVisor: read the sandbox cgroup. Reads `/sys/fs/cgroup/pause`, not
  `main/`: every application process runs inside the sentry, which is a single
  process in `pause/`.
* **PR 3** — micro-VM: read the guest agent's `StatsContainer` over the existing
  vsock connection, not the host cgroup (guest RAM is a fixed allocation, so the
  host cgroup barely moves with the workload). Closes #594.

### DIFF
--- a/cmd/ateom-gvisor/main.go
+++ b/cmd/ateom-gvisor/main.go
@@ -256,9 +256,9 @@ type AteomService struct {
 
 	// activeActor is the actor whose workload this ateom is currently running,
 	// or nil when it is "available". An ateom serves one actor at a time, so a
-	// single slot is enough (the micro-VM ateom holds the same field on each
-	// entry of its s.running map). Guarded by lock, like every other RPC-visible
-	// field.
+	// single slot is enough (the micro-VM ateom holds the same field, set and
+	// cleared at the same points). Guarded by lock, like every other
+	// RPC-visible field.
 	//
 	// Set by RunWorkload / RestoreWorkload and cleared by CheckpointWorkload, so
 	// it tracks exactly the available/executing state machine described on the
--- a/cmd/ateom-gvisor/main.go
+++ b/cmd/ateom-gvisor/main.go
@@ -35,6 +35,7 @@ import (
 	"github.com/agent-substrate/substrate/internal/ateinterceptors"
 	"github.com/agent-substrate/substrate/internal/ateomnet"
 	"github.com/agent-substrate/substrate/internal/ateompath"
+	"github.com/agent-substrate/substrate/internal/ateomstats"
 	"github.com/agent-substrate/substrate/internal/atunnel"
 	"github.com/agent-substrate/substrate/internal/contextlogging"
 	"github.com/agent-substrate/substrate/internal/imagecache"
@@ -49,7 +50,9 @@ import (
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/reflection"
+	"google.golang.org/grpc/status"
 )
 
 var (
@@ -250,6 +253,17 @@ type AteomService struct {
 	podIdentityTrustBundlePath string
 	// egressGatewayTrustBundlePath verifies the remote gateway's serving cert.
 	egressGatewayTrustBundlePath string
+
+	// activeActor is the actor whose workload this ateom is currently running,
+	// or nil when it is "available". An ateom serves one actor at a time, so a
+	// single slot is enough (the micro-VM ateom holds the same field on each
+	// entry of its s.running map). Guarded by lock, like every other RPC-visible
+	// field.
+	//
+	// Set by RunWorkload / RestoreWorkload and cleared by CheckpointWorkload, so
+	// it tracks exactly the available/executing state machine described on the
+	// Ateom service. GetWorkloadStats reads it to attribute its sample.
+	activeActor *ateomstats.ActorAttribution
 }
 
 var _ ateompb.AteomServer = (*AteomService)(nil)
@@ -278,6 +292,12 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
 	actorRef := resources.ActorRef{Atespace: req.GetAtespace(), Name: req.GetActorName()}
 	s.actorLogger.EmitLifecycleLog("Actor starting", actorRef, req.GetActorUid(), req.GetActorTemplateNamespace(), req.GetActorTemplateName())
 
+	// Retain the attribution before the boot rather than after it, so a sample
+	// taken against a workload that dies mid-boot is still attributable. The
+	// cleanup below drops it again if the boot fails outright.
+	attribution := ateomstats.ActorAttributionFromRequest(req)
+	s.activeActor = &attribution
+
 	// Contract with atelet:
 	//
 	//   * Correct runsc version is downloaded and placed on disk.
@@ -292,6 +312,9 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
 		DumpNetInfo:        true,
 		EgressRedirectPort: s.egressRedirectPort(req.GetEgressGateway() != nil),
 	}); err != nil {
+		// Cleared here as well as in the deferred cleanup below, because that
+		// defer is not registered until after this check.
+		s.activeActor = nil
 		return nil, fmt.Errorf("while setting up actor network: %w", err)
 	}
 	rcmd := &runsc{
@@ -301,6 +324,7 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
 	var containersToDelete []string
 	defer func() {
 		if retErr != nil {
+			s.activeActor = nil
 			cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
 			defer cancel()
 			if err := s.deactivateActorNetworking(cleanupCtx); err != nil {
@@ -420,6 +444,19 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 		return nil, fmt.Errorf("unsupported snapshot scope: %v", req.GetScope())
 	}
 
+	// The sandbox is gone as of the checkpoint above, so the ateom is back to
+	// "available" from here on: there is nothing left to measure, and holding
+	// the attribution would let a later GetWorkloadStats report a checkpointed
+	// actor as though it were still running.
+	//
+	// Cleared here rather than at the end of the function because everything
+	// below is bookkeeping over a dead sandbox and can still fail (listing the
+	// snapshot files returns an error), which would otherwise leave the
+	// attribution behind. Conversely nothing above this point clears it: a
+	// checkpoint that failed may well have left the workload running, and
+	// reporting its usage is then the honest answer.
+	s.activeActor = nil
+
 	// After checkpointing the sandbox root, runsc may no longer have a usable
 	// control server for state/delete calls. Keep this as best-effort cleanup:
 	// atelet resets the actor runsc, bundle, pidfile, and checkpoint
@@ -455,6 +492,16 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 	s.actorLogger.EmitLifecycleLog("Actor checkpointed", actorRef, req.GetActorUid(), req.GetActorTemplateNamespace(), req.GetActorTemplateName())
 
 	return &ateompb.CheckpointWorkloadResponse{SnapshotFiles: snapshotFiles}, nil
+}
+
+// GetWorkloadStats implements ateompb.Ateom/GetWorkloadStats.
+//
+// The attribution half is wired up here; the measurement half is not. Reading the
+// sandbox's cgroup (/sys/fs/cgroup/pause) lands in the follow-up to
+// https://github.com/agent-substrate/substrate/issues/594, at which point this
+// stops returning Unimplemented.
+func (s *AteomService) GetWorkloadStats(ctx context.Context, req *ateompb.GetWorkloadStatsRequest) (*ateompb.GetWorkloadStatsResponse, error) {
+	return nil, status.Error(codes.Unimplemented, "GetWorkloadStats is not implemented yet")
 }
 
 // listSnapshotFiles returns the (relative) names of regular files directly under
@@ -510,6 +557,10 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
 	actorRef := resources.ActorRef{Atespace: req.GetAtespace(), Name: req.GetActorName()}
 	s.actorLogger.EmitLifecycleLog("Actor restoring", actorRef, req.GetActorUid(), req.GetActorTemplateNamespace(), req.GetActorTemplateName())
 
+	// Same as RunWorkload: retain before the boot, drop again if it fails.
+	attribution := ateomstats.ActorAttributionFromRequest(req)
+	s.activeActor = &attribution
+
 	// Contract with atelet:
 	//
 	//   * Correct runsc version is downloaded and placed on disk.
@@ -525,6 +576,8 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
 		DumpNetInfo:        true,
 		EgressRedirectPort: s.egressRedirectPort(req.GetEgressGateway() != nil),
 	}); err != nil {
+		// Same as the Run path: the defer below is not registered yet.
+		s.activeActor = nil
 		return nil, fmt.Errorf("while setting up actor network: %w", err)
 	}
 	rcmd := &runsc{
@@ -534,6 +587,7 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
 	var containersToDelete []string
 	defer func() {
 		if retErr != nil {
+			s.activeActor = nil
 			cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
 			defer cancel()
 			if err := s.deactivateActorNetworking(cleanupCtx); err != nil {
--- a/cmd/ateom-gvisor/main.go
+++ b/cmd/ateom-gvisor/main.go
@@ -28,6 +28,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
@@ -257,13 +258,27 @@ type AteomService struct {
 	// activeActor is the actor whose workload this ateom is currently running,
 	// or nil when it is "available". An ateom serves one actor at a time, so a
 	// single slot is enough (the micro-VM ateom holds the same field, set and
-	// cleared at the same points). Guarded by lock, like every other
-	// RPC-visible field.
+	// cleared at the same points).
 	//
 	// Set by RunWorkload / RestoreWorkload and cleared by CheckpointWorkload, so
 	// it tracks exactly the available/executing state machine described on the
 	// Ateom service. GetWorkloadStats reads it to attribute its sample.
-	activeActor *ateomstats.ActorAttribution
+	//
+	// Atomic rather than guarded by lock, unlike every other RPC-visible field
+	// here. The three writers already hold lock for their whole bodies and keep
+	// doing so; the point is the reader. lock is held across an entire boot,
+	// restore, or checkpoint, so a lock-guarded read would park a poller for the
+	// full duration of each -- going quiet during exactly the phases whose usage
+	// is most interesting -- and holding it across the read would put a
+	// CheckpointWorkload behind telemetry instead. The field is only ever
+	// assigned or cleared as a whole pointer, never mutated in place, which is
+	// exactly what atomic.Pointer is for.
+	//
+	// The type makes a lock-free read possible; it does not make one happen.
+	// GetWorkloadStats must not take lock at all, including around whatever it
+	// does with the value. A regression test pins that once there is a handler
+	// with a body to pin.
+	activeActor atomic.Pointer[ateomstats.ActorAttribution]
 }
 
 var _ ateompb.AteomServer = (*AteomService)(nil)
@@ -296,7 +311,7 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
 	// taken against a workload that dies mid-boot is still attributable. The
 	// cleanup below drops it again if the boot fails outright.
 	attribution := ateomstats.ActorAttributionFromRequest(req)
-	s.activeActor = &attribution
+	s.activeActor.Store(&attribution)
 
 	// Contract with atelet:
 	//
@@ -314,7 +329,7 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
 	}); err != nil {
 		// Cleared here as well as in the deferred cleanup below, because that
 		// defer is not registered until after this check.
-		s.activeActor = nil
+		s.activeActor.Store(nil)
 		return nil, fmt.Errorf("while setting up actor network: %w", err)
 	}
 	rcmd := &runsc{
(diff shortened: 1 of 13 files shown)


internal/proto/ateompb/ateom.proto
```diff
@@ -45,6 +45,15 @@ service Ateom {
...
   rpc RestoreWorkload(RestoreWorkloadRequest) returns (RestoreWorkloadResponse) {}
+
+  // StatsWorkload returns a point-in-time resource-usage sample for the
+  // workload this ateom is currently executing.
+  //
+  // It is a pure read: unlike the three calls above it does not move the ateom
+  // between "available" and "executing", so it is safe to call on a timer for
+  // the whole lifetime of a workload. Returns FAILED_PRECONDITION when the
+  // ateom is "available" (nothing to measure).
+  rpc StatsWorkload(StatsWorkloadRequest) returns (StatsWorkloadResponse) {}
```
git286: I would recommend to name it GetWorkloadStats (and rename the request and response as well) to follow the AIP standard: https://google.aip.dev/131

internal/proto/ateompb/ateom.proto
```diff
@@ -185,3 +194,50 @@ message RestoreWorkloadRequest {
...
+  // the caller can attribute the sample without holding its own mapping from
+  // worker to actor.
+  string atespace = 1;
+  string actor_name = 2;
+  string actor_uid = 3;
+  string actor_template_namespace = 4;
+  string actor_template_name = 5;
+
+  // "gvisor" | "microvm".
+  string sandbox_class = 6;
```
git286: Should we make it an enum?

internal/proto/ateompb/ateom.proto
```diff
@@ -185,3 +194,50 @@ message RestoreWorkloadRequest {
...
+
+  // How the measurements below were obtained: "cgroup" (read from the sandbox's
+  // cgroup on the host) or "guest-agent" (read from inside the guest). The two
+  // are not measured the same way and should not be compared against each other
+  // as though they were.
+  //
+  // Empty when no measurement was taken.
+  string source = 7;
```
git286: Same, should we make the source an enum?

cmd/ateom-gvisor/main.go
```diff
@@ -336,11 +360,26 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 		return nil, fmt.Errorf("while listing checkpoint files: %w", err)
 	}
 
+	// The workload is gone and the ateom is back to "available": there is
+	// nothing left to measure, and holding the identity would let a later
+	// StatsWorkload report a checkpointed actor as though it were still running.
+	s.activeActor = nil
```
git286: Is it too late to set nil here? What if listSnapshotFiles failed above?

internal/proto/ateompb/ateom.proto
baizhenyu: Done — `GetWorkloadStats`, `GetWorkloadStatsRequest`, `GetWorkloadStatsResponse`. Agreed that `StatsWorkload` read as if "Stats" were a verb.

One note on how far I took AIP-131: §3.1 of `docs/api-style-guide.md` also says the response **must** be the resource itself rather than a `...Response` wrapper, which would make this `returns (WorkloadStats)`. I kept the `Request`/`Response` pair as you wrote it, since the rest of this internal service (`RunWorkload`, `CheckpointWorkload`, `RestoreWorkload`) is paired that way and a sample is not really a addressable resource. Happy to go the other way if you prefer the strict reading.

internal/proto/ateompb/ateom.proto
baizhenyu: Done — `SandboxClass` enum with `SANDBOX_CLASS_UNSPECIFIED` / `_GVISOR` / `_MICROVM`, following §5.1 of the style guide (prefixed values, `_UNSPECIFIED` zero) and matching the existing `SnapshotScope` in this file. The values mirror the `ate.dev/v1alpha1` `SandboxClass` the WorkerPool and ActorTemplate are configured with.

internal/proto/ateompb/ateom.proto
baizhenyu: Done, same treatment — `StatsSource` with `STATS_SOURCE_UNSPECIFIED` / `_CGROUP` / `_GUEST_AGENT`. `_UNSPECIFIED` carries the "no measurement was taken" meaning the empty string used to, so the "all four measurements are zero" case is still distinguishable from a real zero.

cmd/ateom-gvisor/main.go
baizhenyu: Good catch — that was a real bug, fixed.

`listSnapshotFiles` can return early, and by that point `cmdCheckpoint` / `cmdFsCheckpoint` has already destroyed the sandbox, so the ateom would have kept advertising a checkpointed actor as live for whatever `GetWorkloadStats` calls arrived before atelet recycled the worker.

Moved the clear to immediately after the snapshot `switch`, which is the exact moment the sandbox stops existing. Everything below it (`cleanupContainersAfterCheckpoint`, `UnmountAllUnder`, `CleanupActorNetwork`, `listSnapshotFiles`) is bookkeeping over a dead sandbox. Deliberately nothing *above* it clears: a checkpoint that failed may well have left the workload running, and reporting its usage is the honest answer there.

I checked the micro-VM `CheckpointWorkload` for the same shape and it does not have it — the guest is only `Pause`d until `teardownActor`, so every early return leaves a live VM still in `s.running`, and `teardownActor` / `delete(s.running, ...)` are adjacent with nothing fallible in between.

git286:
LGTM for the o11y side but please also get a LGTM from @BenTheElder  on the ateom side

git286 (dismissed):

cmd/ateom-microvm/run.go
```diff
@@ -458,7 +478,7 @@ func (s *AteomService) coldBootActor(ctx context.Context, p actorBootParams) (re
 		return fmt.Errorf("while waiting for container readyz: %w", err)
 	}
 
-	ra := &runningActor{chCmd: chCmd, vfsdCmd: vfsdCmd, durableVfsdCmd: durableVfsdCmd, apiSocket: apiSocket, baseID: actorUID, logAgent: ac}
+	ra := &runningActor{chCmd: chCmd, vfsdCmd: vfsdCmd, durableVfsdCmd: durableVfsdCmd, apiSocket: apiSocket, baseID: actorUID, logAgent: ac, activeActor: p.actorAttribution()}
```
BenTheElder: 🤖 **question** 🟢 – The two runtimes disagree about when a booting actor becomes attributable. Attribution attaches here, on the `runningActor` built after readyz, so nothing is registered until the boot succeeds. The gVisor side sets `s.activeActor` before its boot and says why: "so a sample taken against a workload that dies mid-boot is still attributable" (`cmd/ateom-gvisor/main.go:287`).

Against the proto's state machine that difference is visible: a poll during a micro-VM boot finds no `runningActor` and gets FAILED_PRECONDITION, i.e. "the ateom is available (nothing to measure)", when it is really mid-activation. The RPC is documented as "safe to call on a timer for the whole lifetime of a workload", so a caller polling across a resume would see that blip on micro-VM and not on gVisor.

Nothing observable yet, since both return Unimplemented — but this PR is the foundation the measurement half builds on, and the semantics are cheaper to settle now. Is the gVisor rationale meant to apply to both, or is post-boot-only the intended contract?

BenTheElder (commented):
claude raises a good question:

cmd/ateom-microvm/run.go
baizhenyu: Both should attach before the boot — the gVisor reason applies to micro-VM too. Post-boot-only wasn't a decision. Fixed in `9bc7fb6e`.

It followed from where the field lived. gVisor has a slot on `AteomService` whose only job is attribution, so `RunWorkload` sets it before the boot. Micro-VM put attribution on `runningActor`, which is assembled from the boot's outputs (`chCmd`, `vfsdCmd`, `apiSocket`, `logAgent`) and so isn't constructed until just after `readyz.WaitAll`. The gVisor field's comment already claimed micro-VM "holds the same field", so the code asserted a symmetry it didn't have.

Nothing is observable while both handlers return `Unimplemented`. It bites once measurement lands: an actor that dies mid-boot or loops never reaches readyz, and that's the actor whose usage you'd most want.

**Change.** Micro-VM gains `AteomService.activeActor`, set before the boot in `RunWorkload` and `RestoreWorkload` and cleared by a deferred check on the error paths — the same points as gVisor. `runningActor.activeActor` is gone, so there's one source of truth. A single slot because an ateom serves one actor at a time; `running` is keyed by UID for lookup, not concurrency.

**The alternative I rejected** was inserting a half-built `runningActor` into `running` before the boot and filling it in after. It wouldn't break anything today — `teardownActor` and `snapshotVMState` nil-guard every field, because `ra` can already be nil after an ateom restart — and that's the problem. `ra != nil` currently means "we own this actor's processes"; a half-built entry takes that branch and silently no-ops. The guards that save it are per-field and incidental, not enforced. It's also only unreachable because `s.lock` spans both RPCs, and shrinking that hold (the readiness wait) is a follow-up I've already flagged.

**Checkpoint clears at the teardown, not the snapshot** — the one place the runtimes should differ. runsc's checkpoint takes the sandbox down, so gVisor clears as soon as it returns. The micro-VM guest is only paused until `teardownActor`, so a checkpoint that failed earlier has left it present, and reporting its usage is the honest answer. Clearing next to the `delete` from `running` keeps the two views of "is an actor here" in agreement.

**Proto.** The contract sentence equated `FAILED_PRECONDITION` with "available". There's now a third state: executing and attributable, but no sandbox to read yet. Same code on purpose — a timer-driven caller skips the sample either way — so what needed saying is that it means "no numbers right now", not "the actor is gone".

Neither field is atomic yet; each becomes an `atomic.Pointer` when its runtime's measurement half lands. Keeping the retention points identical now is what makes both mechanical.

cmd/ateom-gvisor/main.go
```diff
@@ -243,6 +246,17 @@ type AteomService struct {
...
+	// activeActor is the actor whose workload this ateom is currently running,
+	// or nil when it is "available". An ateom serves one actor at a time, so a
+	// single slot is enough (the micro-VM ateom holds the same field, set and
+	// cleared at the same points). Guarded by lock, like every other
+	// RPC-visible field.
+	//
+	// Set by RunWorkload / RestoreWorkload and cleared by CheckpointWorkload, so
+	// it tracks exactly the available/executing state machine described on the
+	// Ateom service. GetWorkloadStats reads it to attribute its sample.
+	activeActor *ateomstats.ActorAttribution
```
BenTheElder: 🤖 **should-fix** 🟡 – Guarding this with `lock` is what will stop `GetWorkloadStats` being safe on a timer. `RunWorkload`, `RestoreWorkload` and `CheckpointWorkload` each take `s.lock` on their first line and hold it for the whole handler, so once the follow-up reads `activeActor` under that same mutex the stats call blocks for the full duration of a boot, a restore, or a checkpoint — including the snapshot write. The proto says the RPC "does not move the ateom between available and executing, so it is safe to call on a timer for the whole lifetime of a workload", but not moving the state machine is not the same as not blocking on it, and a poller would go quiet during exactly the phases whose usage is most interesting.

`ateom-microvm` inherits the same shape, since `runningActor` lives in `s.running` under its own `s.lock`.

Cheap to avoid now while the field is new and has no readers: give the attribution its own mutex, or hold it in an `atomic.Pointer[ActorAttribution]`. Doing it in the follow-up means either revisiting the locking or shipping a poller that stalls.

internal/proto/ateompb/ateom.proto
```diff
@@ -208,3 +226,68 @@ message RestoreWorkloadRequest {
 
 message RestoreWorkloadResponse {
 }
+
+message GetWorkloadStatsRequest {
+  // The actor the caller believes is executing here. A worker can be recycled
+  // between the caller's view of the world and this call, so ateom rejects a
+  // mismatch with FAILED_PRECONDITION rather than reporting a different
+  // actor's numbers under the requested actor's identity.
+  string actor_uid = 1;
```
BenTheElder, line 251: 🤖 **question** 🟢 – A UID mismatch and a workload still booting both answer `FAILED_PRECONDITION`, but they call for different behaviour from the caller. Booting means "skip this sample, take the next one", which the RPC doc argues well. A mismatch means the caller's worker-to-actor mapping is stale, and retrying on the same timer just keeps asking about an actor that is not here — the caller wants to re-resolve, and cannot tell that from the code alone.

Since this is the PR that fixes the contract, is a distinct code worth it for the recycled-worker case (`NOT_FOUND`, or `ABORTED` to signal "re-read your state and retry")? The response already echoes the measured actor, so a caller could compare after the fact, but only on the success path.

cmd/ateom-gvisor/main.go
baizhenyu: Agreed, and taken now. `9376fc9f`.

Both fields become `atomic.Pointer[ateomstats.ActorAttribution]`.

The reason this works: `lock` isn't making the read slow, it's the only thing making it safe. A plain field read alongside a concurrent write is a data race, so `GetWorkloadStats` would have no choice but to take `lock` — and `lock` is held for a whole boot, restore, or checkpoint. Reading one pointer would mean waiting out the boot. `atomic.Pointer` makes the read safe on its own, so the reader doesn't queue behind the writers. Writers keep taking `lock`; they have other state to protect, and the point here is the reader.

You're right that the micro-VM has the same shape, and worse: its `lock` spans a cold boot including the retry, a snapshot write, and a restore.

One thing the type doesn't buy: a handler can still take `s.lock` for some other reason and put the stall right back. So both field comments now say it explicitly — `GetWorkloadStats` must not take `lock` at all — and each runtime's measurement PR carries the regression test (hold `s.lock` across the call; a handler that takes it deadlocks). Not added here because against a stub it passes vacuously.

On the proto wording you quoted: "does not move the ateom between available and executing" was a narrower claim than it reads as. Not blocking on the state machine is a separate property, and the doc now states it separately.

internal/proto/ateompb/ateom.proto
baizhenyu: Yes, worth it. `9376fc9f` — went with `NOT_FOUND`, and slightly wider than the mismatch case.

Splitting by what the caller should *do*, rather than by what the ateom's state machine calls itself, gives a two-way:

- **`NOT_FOUND`** — this ateom is not executing the actor you asked about. Available, or recycled onto a different one. Retrying on the same timer won't change it; re-resolve.
- **`FAILED_PRECONDITION`** — executing the requested actor, no numbers yet. Transient; skip and take the next sample.

"Available" moves to the first side too. It tells the caller the same thing a mismatch does — the actor isn't here — so the transient code would have a poller keep asking a worker that has nothing for it. `NOT_FOUND` over `ABORTED` mostly because "not here" is the plainer reading, and after a recycle there's nothing to retry into.

On comparing the echoed actor after the fact: agreed that only helps on the success path, which is why the codes have to carry it.

Worth noting this split is only expressible because of `9bc7fb6e` in this PR. While micro-VM attached attribution after readyz, "available" covered both "no actor here" and "actor booting", and no pair of codes could separate them — your two comments turned out to be the same problem from different ends.

Free to change: both handlers still return `Unimplemented`, so nothing reads either code yet.

BenTheElder (approved):

PR #669, opened 2026-07-31
ateapi: garbage collect actor snapshots

### LINKED ISSUE #664: Add retention policy and garbage collection for ActorSnapshots
## Problem

Every successful suspend creates an immutable `ActorSnapshot`, but superseded and deleted-Actor snapshots are never removed. An overnight soak produced 1,603 generations for one Actor, of which 1,602 were no longer referenced. Bucket lifecycle rules are unsafe because they cannot distinguish an old live snapshot from garbage.

## Retention semantics

A snapshot is protected when any of these is true:

- an `Actor.latest_snapshot` references it;
- an `ActorSnapshotTag` references it;
- it is among the newest `minimumCount` snapshots produced by its live source Actor UID; or
- it is younger than `minimumAgeSeconds` and its source Actor still exists.

Add user-tunable policy to `ActorTemplate.spec.snapshotsConfig`:

```yaml
retentionPolicy:
  minimumCount: 3
  minimumAgeSeconds: 86400
```

Both thresholds use OR semantics. `minimumCount` includes the Actor's latest snapshot. Explicit zero disables that threshold; omitted fields default to `3` and `86400`. Policy updates apply to existing history on the next GC pass. Actor UID, rather than name, defines history so delete/recreate cannot inherit it.

Deleting an Actor removes count/age protection from its history immediately. Tags or references from other Actors still pin those snapshots.

## Collection design

- Run a periodic collector in `ateapi`. Use Kubernetes Lease leader election so only one replica collects at a time; the existing Redis distributed lock remains an alternative if Lease coordination proves unsuitable.
- Determine retention from authoritative Actor, tag, snapshot, and ActorTemplate state. Before deletion, prevent new references and recheck roots so tagging/cloning cannot race GC.
- Delegate deletion of the exact private snapshot prefix to any healthy `atelet`; `atelet` already owns the GCS/S3 clients and credentials. Add an internal idempotent delete-prefix RPC and backend implementations.
- Delete object-storage bytes first, then snapshot metadata. On any failure, retain metadata and retry on a later pass. Never delete metadata first, which would make leaked bytes undiscoverable.
- Keep GC interval/concurrency as operator settings; retention remains workload policy.
- Emit basic attempted/deleted/failed counters and structured failure logs.

## Out of scope

- Partial failed uploads with no `ActorSnapshot` metadata ([#610](https://github.com/agent-substrate/substrate/issues/610)).
- Node-local Actor directories and pause snapshots ([#641](https://github.com/agent-substrate/substrate/issues/641)).
- Per-Atespace policies or storage quotas.
- Manual `DeleteActorSnapshot` API.

## Acceptance criteria

- [ ] CRD defaults retain the newest 3 snapshots and all snapshots younger than 24 hours for each live Actor.
- [ ] Users can tune either threshold, including setting either or both to zero.
- [ ] Latest-snapshot and tag references are never collected.
- [ ] Deleting an Actor makes its otherwise-unreferenced history collectable.
- [ ] Concurrent tag/clone operations cannot produce a reference to deleted data.
- [ ] GCS and S3-compatible prefixes are deleted idempotently; metadata is removed only afterward.
- [ ] Kubernetes Lease leader election makes collection safe with multiple `ateapi` replicas, and collection resumes after process/storage failures.
- [ ] Unit tests cover policy boundaries, references, races, and retry ordering; E2E verifies superseded and deleted-Actor cleanup while tagged snapshots survive.

Follow-up to [#529](https://github.com/agent-substrate/substrate/issues/529), [#568](https://github.com/agent-substrate/substrate/pull/568), and [#570](https://github.com/agent-substrate/substrate/pull/570).

EItanya:
## Summary
- add per-ActorTemplate minimum-count and minimum-age snapshot retention with tunable zero values
- elect one ateapi GC worker with a Kubernetes Lease and delete snapshot storage through ready atelets
- pin Actor latest snapshots, in-progress snapshots, golden snapshots, and tagged snapshots; tombstone deletions to prevent new references
- add unit, storage-prefix, leader-election, and kind E2E coverage

## Known tradeoff

`ateapi` currently calls `DialAny()` to route each storage deletion through an arbitrary ready `atelet`. Snapshot deletion is not node-affine; this exists only because `atelet` already has the object-store credentials while `ateapi` does not. This is intentionally minimal but creates an unnecessary availability dependency on `atelet`. Before making the PR ready, we should decide whether `ateapi` should receive narrowly scoped object-delete privileges and delete directly instead.

## Testing
- env -u NO_COLOR GOCACHE=/tmp/substrate-issue-664-gocache make verify
- hack/run-e2e-kind.sh ./internal/e2e/suites/demo -run '^TestActorSnapshotGarbageCollection$' -count=1 -args --no-color

Fixes #664


EItanya:
Closing while the retention configuration and GC indexing design are revised.

PR #674, opened 2026-07-31
ateapi/controlapi: process syncer events via rate-limiting workqueue

troychiu:
Refactors `WorkerPoolSyncer` to process events through a rate-limiting workqueue instead of processing them synchronously inline.

Previously, store errors during Informer event handling, such as 
- optimistic locking version conflicts (`store.ErrVersionConflict`)
- create races (`store.ErrAlreadyExists`)
- `WorkerPool` not yet present in the lister cache

were logged and dropped, leaving store state out of sync until a subsequent pod event occurred.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/syncer.go
+++ b/cmd/ateapi/internal/controlapi/syncer.go
@@ -17,25 +17,46 @@ package controlapi
 import (
 	"context"
 	"errors"
+	"fmt"
 	"log/slog"
 	"maps"
+	"time"
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
 	"github.com/agent-substrate/substrate/internal/resources"
 	listersv1alpha1 "github.com/agent-substrate/substrate/pkg/client/listers/api/v1alpha1"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 )
+
+// syncerWorkerCount is the number of goroutines draining the work queue. The
+// queue never hands the same key to two workers concurrently, so per-key
+// ordering is preserved.
+const syncerWorkerCount = 2
+
+// workerKey identifies a worker row in the store. pool is captured at enqueue
+// time because once the pod is gone from the informer cache, the
+// ate.dev/worker-pool label cannot be recovered from a namespace/name key.
+type workerKey struct {
+	namespace string
+	pool      string
+	name      string
+}
 
 // WorkerPoolSyncer reconciles the state of worker pods from Kubernetes Informer
 // into the store.
+//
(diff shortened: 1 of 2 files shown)


troychiu:
cc @EItanya

troychiu:
friendly pinging for review @EItanya @sfunkenhauser @zoez7

EItanya:
Sorry about the misunderstanding, I should've been more clear. When I said using the manager I meant to use the same reconcile style as `ActorTemplate` https://github.com/agent-substrate/substrate/blob/a0e4c787323fe947d7a59cd2ef28a1b9b0f36856/cmd/atecontroller/internal/controllers/actortemplate_controller.go#L65

We have multiple reconcile styles throughout the codebase which is already confusing, and controller-runtime controllers already implement rate limiting queues. 

I mentioned this to @sfunkenhauser in #728 as well

troychiu:
I agree and thanks for the context!

Eventually we would like to move `WorkerPoolSyncer` out of `ate-api-server` (#729), which requires #730. In the meantime, #728 can be temporarily solved by either this PR or #578 as a stop-gap before #730 is ready.

Given that we want a minimal stop-gap ([comment](#issuecomment-5182364620)), #578 might be a better candidate since it avoids introducing a new workqueue pattern here.

Let me know if you'd prefer I reopen #578 instead! @EItanya @sfunkenhauser

At the same time, I am happy to help with #730

sfunkenhauser:
My thought was that we can address the different reconciliation styles when moving the WorkerPoolSyncer to the ate-controller.

This PR isn't introducing a new style, and just tries to fix gaps in the existing one. I'd rather not block this improvement on reimplementing the syncer using the controller-runtime.

@EItanya Feel free to push back if you disagree.

EItanya:
I really don’t wanna be the person blocking this, of course I think incremental change is important. My broader point was that a lot of the logic added in this PR is done automatically by controller runtime, so if we’re going that direction we should just do it. But definitely don’t block on me if that’s not the plan, or if we plan to move away from that style.

sfunkenhauser:
@EItanya I appreciate you calling this out, and I agree that we should consolidate on the same reconciliation mechanism.

That being said I don't want to ask @troychiu to re-write this PR for the third time. I don't think this PR will make it anymore difficult to switch to using controller runtime (it might actually make it easier). In the interest of incremental improvement let's get this in as in, and we can handle the controller runtime refactor when moving to ate-controller (which should hopefully happen in the next week or so).

sfunkenhauser (approved):

EItanya:
So I spoke to @juli4n, and it sounds like the direction is to move away from controller-runtime generally, so either way I think this was the correct direction in the end, sorry for the delay

troychiu:
Thanks @sfunkenhauser @EItanya! Both of your suggestions were really helpful.

troychiu:
> move away from controller-runtime generally

Do you mind providing more details?

EItanya:
> > move away from controller-runtime generally
> 
> Do you mind providing more details?

Basically, controller-runtime is very opinionated, and contains more than is strictly necessary for many controllers. This sometimes leads to bloated controllers which don't behave exactly the way you expect. client-go machinery is slightly more complicated, but gives much more freedom to achieve the exact desired effect and performance.

Issue #675, opened 2026-07-31
E2E test flaky: TestRequestParking


yufan-su:
The E2E test TestRequestParking added in #221 was flaky in ci

Sample failures: 
- https://github.com/agent-substrate/substrate/actions/runs/30657328354/job/91244870748
-  https://github.com/agent-substrate/substrate/actions/runs/30477187339/job/90661727410
- https://github.com/agent-substrate/substrate/actions/runs/30476804701/job/90660441335

```
--- FAIL: TestRequestParking (13.43s)
    --- FAIL: TestRequestParking/ParkThenServed (6.11s)
    --- PASS: TestRequestParking/BudgetExhaustion (5.06s)
FAIL
```

zoez7:
Hey @omeryahud Do you mind taking a look at this since you added the test? This is quite flaky and blocking many PRs.

omeryahud:
> Hey @omeryahud Do you mind taking a look at this since you added the test? This is quite flaky and blocking many PRs.

Hi @zoez7 !

I've replied on the linked PR (see above references), waiting for @yufan-su's reply

zoez7:
Thanks! Looks like yufan-su is out of office for the next week, do you mind taking over the PR @omeryahud ?

BenTheElder:
seeing this one a lot

haiyanmeng:
Another instance of the test flakiness: https://github.com/agent-substrate/substrate/actions/runs/32727304540/job/97464580196?pr=1156

aditya-shantanu:
No failures for TestRequestParking (e2e-gvisor) since 3a3ec62f (#1160) across 56 runs in the last 7 days — marking resolved on the flake dashboard. The e2e-microvm lane also has zero post-merge failures; its dashboard row stays open only until the pre-merge failures age out of the 7-day window.

aditya-shantanu:
Follow-up: no failures for TestRequestParking (e2e-microvm) since 3a3ec62f (#1160) across 51 runs — at the prior 12.6% flake rate that's ~0.1% probability by chance, so marking the microvm lane resolved on the flake dashboard as well (supersedes my earlier note about waiting for the window to age out). Both lanes now resolved.

PR #676, opened 2026-07-31
router: grant an in-flight resume a bounded grace past the park budget when a worker is already assigned

yufan-su:
#675 

# Context

In a flaky [run](https://github.com/agent-substrate/substrate/actions/runs/30657328354/job/91244870748) of  TestRequestParking E2E, it parks a request for actor B behind a deliberately 1-worker pool, frees the worker by suspending actor A, and expects a 200.

What we got instead was a `503 "no free workers available"` at just over the 5s park budget, where Actor B was left in RESUMING with the worker already assigned. In this case, CI runs 7 e2e suites against a single 4-vCPU kind cluster; under that contention the gVisor snapshot restore routinely runs a few hundred milliseconds past the 5s budget. So the router cancelled a resume that was about to succeed.

# Proposal

When the park budget elapses with a resume attempt still in flight, the router issues one `GetActor`:

 - Actor is `RESUMING` or `RUNNING` → a worker is assigned; the in-flight attempt gets a bounded 3s grace to land.
 - Anything else (including a failed or timed-out probe) → cancel and 503 exactly as before.

The retry loop always stops at the budget. The grace covers the single attempt already underway, never another round. The probe's 1s timeout is spent inside the grace, not on top of it, so the router's worst-case hold on a parked request is exactly budget + grace.

- [x] Tests pass

### DIFF
--- a/cmd/atenet/internal/router/resumer.go
+++ b/cmd/atenet/internal/router/resumer.go
@@ -36,6 +36,26 @@ import (
 // conflicts are retried; capacity errors fail immediately.
 const failFastResumeBudget = 15 * time.Second
 
+// resumeAttemptGrace is the extra time a resume attempt still in flight at the
+// budget gets to land, but only if a worker is already assigned to the actor
+// (see workerAssigned). Retries always stop at the budget; the grace covers the
+// one attempt underway, never another round.
+//
+// Canceling an assigned attempt is the case worth avoiding: it does not
+// release the worker, so the pool the request was waiting on stays starved,
+// and a restore that was about to succeed is discarded. Under node contention
+// those overshoot the budget by a few hundred milliseconds routinely. With no
+// worker assigned there is nothing to lose, so the request fails on its budget.
+//
+// budget+grace is the router's worst-case hold on a parked request, which the
+// ext_proc message timeout must cover; see extProcMessageTimeoutFor.
+const resumeAttemptGrace = 3 * time.Second
+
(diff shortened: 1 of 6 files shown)


cmd/atenet/internal/router/resumer.go
```diff
@@ -185,9 +233,35 @@ func (r *ActorResumer) ResumeActor(ctx context.Context, actorRef resources.Actor
...
+		attemptCtx := bgCtx
+		if r.grace > 0 {
+			var cancelAttempt context.CancelFunc
+			attemptCtx, cancelAttempt = context.WithCancel(context.Background())
+			defer cancelAttempt()
+
+			stopWatchdog := context.AfterFunc(bgCtx, func() {
+				graceExpiry := time.After(r.grace) // started before the probe, so the probe is spent inside the grace
+				if !r.workerAssigned(actorRef) {
+					cancelAttempt()
```
omeryahud, line 251: I don't think this resolves the race.

Even if the actor was deemed suspended in line 250, with an active RPC still in-flight, it is still possible for the actor to switch phases to resuming/running by the time the cancellation is propagated to ateapi.

I think the only pragmatic approach to this is to let the in-flight RPC run to completion, without a grace period, and document it.

omeryahud (commented):
Thanks for reporting and submitting a fix!

haiyanmeng:
@yufan-su , any updates on this PR?

yufan-su:
closed as fix was in #1160

PR #677, opened 2026-07-31
serverboot: dynamic --log-level flag for the server binaries

igooch:
Adds a dynamic `--log-level` flag (debug, info, warn, error; case-insensitive) to
the server binaries, backed by a `slog.LevelVar` in `serverboot`. Defaults to
`info` — no behavior change unless the flag is set.

- **serverboot**: the LevelVar behind `InitLogger`'s handler; `SetLogLevel(string)`
  to parse and set it (empty string = unset, a documented no-op); `LogLevel()`
  returns a `slog.Leveler` for binaries that build their own handler.
- **atelet, ateapi**: new pflag, fatal on an invalid value.
- **ateom-gvisor, ateom-microvm**: flag accepted by the binary; note their
  container args are built by the WorkerPool controller with no pass-through yet,
  so wiring it is a follow-up — and that follow-up must roll out only after this
  change's ateom images are deployed (older images reject unknown flags and would
  crashloop).
- **atenet (router + dns)**: replaces two pre-existing hand-rolled `--log-level`
  parsers that silently fell back to `info` on a bad value; both now use
  `serverboot.SetLogLevel` and `serverboot.InitLogger`, which also adds
  `ate.dev/trace-id` correlation to atenet logs.
- **Kind manifests**: atelet runs at `--log-level=debug` on kind (dev/CI), so e2e
  suites can assert on per-item log lines; production installs keep `info`.

Motivation: groundwork for #463 Phase 2 observability — per-item GC log lines can
ship at Debug and be enabled per-node instead of landing at Info.

**Testing**: unit tests pin the untouched default at exactly `info`, dynamic
raise/lower, case-insensitivity, invalid-value rejection, and the empty-string
no-op; `go test -race` green across serverboot, atenet, atelet, ateapi.

### DIFF
--- a/cmd/ateapi/main.go
+++ b/cmd/ateapi/main.go
@@ -83,6 +83,7 @@ var (
 	drainTimeout = pflag.Duration("drain-timeout", 15*time.Second, "Deadline for the graceful gRPC drain on shutdown. In-flight RPCs still running past it are forcefully cancelled.")
 
 	showVersion     = pflag.Bool("version", false, "Print version and exit.")
+	logLevelFlag    = pflag.String("log-level", "info", "Minimum log level: debug, info, warn, or error.")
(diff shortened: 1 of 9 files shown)


maxsmythe (approved):
LGTM

PR #678, opened 2026-07-31
Add validation to ateapi ExternalVolume

msau42:
Part of #232

CSI driver name and volume id syntax is referenced from https://github.com/container-storage-interface/spec

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/actor.go
+++ b/cmd/ateapi/internal/controlapi/actor.go
@@ -103,7 +103,7 @@ func (s *ServiceImpl) CreateActor(ctx context.Context, inActor *ateapipb.Actor) 
 		LatestSnapshot: sourceSnapshotStatus.GetSnapshot(),
 		SourceSnapshot: sourceSnapshotStatus,
 	}
-	if errs := validateActorUpdate(ctx, field.NewPath("actor"), outActor, inActor, true); len(errs) > 0 {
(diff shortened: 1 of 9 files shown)


cmd/ateapi/internal/controlapi/create_actor.go
```diff
@@ -164,6 +164,9 @@ func validateCreateActorRequest(req *ateapipb.CreateActorRequest) field.ErrorLis
 	if val := actor.GetWorkerSelector(); val != nil {
 		errs = append(errs, validateSelector(val, actorPath.Child("worker_selector"))...)
 	}
+	if val := actor.GetActorVolumes(); len(val) > 0 {
```
juli4n: Why are we validating actor volumes here? When a client calls `CreateActor`, we provision a set of volumes from the ActorTemplate CRD (initialized by `initialActorVolumes` so this must be always valid.

I was expecting some validations on the `ActorVolume` CRD though.

msau42: good point, moved to ActorTemplate CRD

juli4n: Why do we still need this?

msau42: I moved the validation of each individual field to right before we update the Actor object with the field.

juli4n:
LGTM. This needs rebase now.

msau42:
This has been implemented in https://github.com/agent-substrate/substrate/pull/1244

PR #679, opened 2026-07-31
Right-size actor sandboxes to declared ActorTemplate resource limits

chw120:
Actors previously ran in sandboxes sized to the whole node; there was no way to declare how much CPU/memory a given actor should get. This adds an explicit, immutable sizing knob on the ActorTemplate and plumbs it into the sandbox's OCI spec for both the gVisor and micro-VM runtimes.

**API** — `ActorTemplate.spec.resources` (`*corev1.ResourceRequirements`). The `limits` size the sandbox and are baked into the immutable spec; the CRD and generated code are regenerated accordingly.

**internal/sizing** (shared by both runtimes) — new `SandboxSize` value (`FromLimits` / `VCPUs` / `ApplyToOCISpec`). `ApplyToOCISpec` writes CPU quota+period and the memory limit onto the OCI spec, and is a no-op when neither dimension is set, so 0 means "unconstrained". `VCPUs` rounds milliCPU up to whole
vCPUs.

**Plumbing** — ateapi reads the template limits (`actorResourceLimits` → `tmpl.Spec.Resources`) and supplies `CpuMilli`/`MemoryBytes` over the actor RPCs (ateapi → atelet → ateom); ateom applies them — gVisor via the cgroup leaf (`runsc --cpu-num-from-quota` provisions the sentry vCPU count), micro-VM via the guest VmConfig. The two fields are carried on the proto messages.

**Scheduling** — worker capacity is taken from the WorkerPool's per-worker limits and advertised on the Worker; the scheduler only places an actor on a worker whose capacity >= the actor's declared limits. A missing worker or actor
dimension is treated as unconstrained, so placement is never blocked by absent data.

**Docs & demos** — document the model in `api-guide.md`; the counter, sandbox, and micro-VM demos declare actor limits, and their WorkerPool comments now describe the real model (worker limits size the worker pod + advertise scheduling capacity; the sandbox itself is sized by `ActorTemplate.spec.resources`).

- [x] Tests pass (unit tests for sizing + scheduling; e2e suite in `internal/e2e/suites/sizing` resumes an actor and asserts, via the probe fixture's `/resources` endpoint, that the running sandbox observes the declared CPU/memory from the inside)


BenTheElder:
This PR looks like it implements exactly what we proposed, but I think the premise was flawed (my bad!)

I think maybe instead what we should actually do is:
1. ActorTemplate starts specifying resource limits.
2. Scheduling considers this to select a worker with >= resource limits.
3. We supply these to the actor RPCs and pass them through to the sandbox.

In the current state we will wind up heavily pinning to the worker it first ran on but maybe not considering it when scheduling again?

cc @juli4n 

3\) should be reusable from the current state, but I think 1/2 might need to replace the downward API approach.

cmd/ateom-microvm/run.go
```diff
@@ -335,6 +347,22 @@ func (s *AteomService) coldBootActor(ctx context.Context, p actorBootParams) (re
...
+	// vCPUs round up; VM RAM reserves a fixed margin for the VMM + virtiofsd, which
+	// share the pod cgroup with the guest RAM. NB: a FULL-scope snapshot restore
+	// reuses the size baked into the snapshot (restoreFullScope), so resizing an
+	// existing actor takes effect on its next cold boot.
+	sz := p.size
+	if v := sz.VCPUs(); v > 0 {
+		vcpus = v
+	}
+	if sz.MemoryBytes > 0 {
+		if m := int(sz.MemoryBytes/(1024*1024)) - s.memReserveMiB; m > 0 {
```
BenTheElder: 🤖 **should-fix** 🟡 – When the reserve swallows the whole limit this silently falls back to the kata default, which is *larger* than what the actor declared. `guestConfig` defaults to 2048 MiB, so an actor declaring 256Mi — exactly the default reserve — takes the `m > 0` branch as false and boots a 2 GiB VM, 8x its declared size. The scheduler already placed it against 256Mi, so a worker sized for that actor is overcommitted by the same factor and the pod OOMs rather than the actor being capped.

The band just above the reserve is the other half: 320Mi leaves 64 MiB of guest RAM, which will not boot a kata guest, and the failure surfaces as a boot hang rather than a sizing error.

Both cases are quiet. Failing the run with a message naming the declared limit and the reserve would make them obvious, and a minimum on `spec.resources` at admission would catch them before an actor is ever scheduled — there is no validation on that field today.

BenTheElder (commented):
quick agentic pass on the latest version:

presubmit also surfaces some typos

cmd/ateom-microvm/run.go
chw120: Thanks Ben for reviewing the PR. I updated the code with the detailed error message and the validation check. Please review again.

BenTheElder:
[needs a rebase, a lot changed]

demos/counter/counter-microvm.yaml.tmpl
```diff
@@ -38,6 +38,18 @@ spec:
   sandboxClass: microvm
   sandboxConfigName: microvm
   ateomImage: ko://github.com/agent-substrate/substrate/cmd/ateom-microvm
+  template:
+    # Per-worker resources: these size the worker POD and advertise its
+    # scheduling capacity (the per-actor ceiling). The micro-VM sandbox itself is
+    # sized by the ActorTemplate's spec.resources below, not by these; keep the
+    # actor memory below this limit so the VMM reserve fits (see internal/sizing).
+    resources:
+      requests:
```
BenTheElder: given that we snapshot/restore, I'm not sure we should support requests, vs only limits for now?

unless we play with the balloon driver, uVM has to have max memory allocated up front, and then on restore it must have the same size.

from my past experiments, the balloon can be pretty slow.

cmd/ateom-gvisor/runsc.go
```diff
@@ -106,7 +113,10 @@ func (r *runsc) cmdCreate(ctx context.Context, out io.Writer, containerName stri
 		// "-debug-to-user-log",
 		// "-log-packets",
 		// "-strace",
-		"-root", ateompath.RunSCStateDir(r.actorUID),
+		// Provision the sentry's vCPU count from the cgroup CPU quota written by
+		// sizing.ApplyToOCISpec, so the sandbox is sized to the pod's limit (runsc
+		// otherwise sizes to all host CPUs). Global flag: before the subcommand.
+		"--cpu-num-from-quota",
```
BenTheElder, line 120: 🤖 **blocking** 🔴 – `-root` was dropped from `cmdCreate`. This `--cpu-num-from-quota` line replaced `"-root", ateompath.RunSCStateDir(r.actorUID)` instead of being added alongside it.

Every other subcommand still passes it — `cmdStart` (line 161), `cmdCheckpoint`, `cmdFsCheckpoint`, `cmdRestore`, `cmdDelete`, `cmdState` — so `runsc create` now writes container state to runsc's default root while `runsc start` looks in `<actor>/runsc-state`. They no longer agree on where state lives, which should break gVisor cold boot at the start step.

`cmdRestore` at line 267 shows the intended shape: `-root` kept, the new flag added after it.

cmd/ateom-microvm/run.go
```diff
@@ -366,6 +386,23 @@ func (s *AteomService) coldBootActor(ctx context.Context, p actorBootParams) (re
...
+	// share the pod cgroup with the guest RAM. A declared memory limit the reserve
+	// leaves too small to boot is rejected (resolveGuestMemMiB) rather than silently
+	// falling back to the larger kata default. NB: a FULL-scope snapshot restore
+	// reuses the size baked into the snapshot (restoreFullScope), so resizing an
+	// existing actor takes effect on its next cold boot.
+	sz := p.size
+	if v := sz.VCPUs(); v > 0 {
+		vcpus = v
+	}
+	memMiB, err = resolveGuestMemMiB(sz.MemoryBytes, s.memReserveMiB, memMiB)
```
BenTheElder, line 404: 🤖 **should-fix** 🟡 – The guest container's memory cgroup and the VM's RAM get sized from different numbers. `buildActorContainers` ran above with `p.size`, so `ApplyToOCISpec` already set the container's `Memory.Limit` to the full declared limit, while this line gives the VM that limit *minus* the VMM reserve. A 1Gi actor ends up with a 1024Mi cgroup cap inside a VM holding 768Mi of RAM.

The cap can then never bind. The workload exhausts guest RAM before it reaches the cgroup limit, so instead of a clean OOM kill you get the guest kernel's OOM killer or a hang — the failure mode `minGuestMemMiB` exists to prevent. Sizing the container cgroup to the post-reserve guest RAM would make the limit meaningful.

`restoreFullScope` calls `buildActorContainers` with `p.size` too, so the same gap applies to a FULL-scope restore.

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -414,6 +424,19 @@ type ActorTemplateSpec struct {
...
+	// CPU/memory limits (cgroup caps, and for the micro-VM the VM's vCPU count and
+	// memory), the scheduler only places the actor on a worker whose capacity is
+	// >= these limits, and the limits are supplied to the sandbox over the actor
+	// RPCs. Because the size is baked into snapshots, it is part of the immutable
+	// spec. Requests are not consulted today (an actor occupies its whole worker).
+	// A zero or absent limit leaves the sandbox at the runtime default (unlimited
+	// for gVisor, the kata config for the micro-VM).
+	//
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```
BenTheElder, line 385: Human edit: upgrading to **should-fix** 🟡 

----

🤖 **question** 🟢 – The field takes a full `corev1.ResourceRequirements`, but only `Limits` is ever read (`actorResourceLimits` in workflow_resume.go). Nothing rejects `requests`, so a user can set them, watch them persist into the spec, and get no effect. `Claims` is accepted on the same terms. The doc comment says requests aren't consulted, which only helps the people who read it.

If your requests-vs-limits question on the demo template lands on "limits only for now", a CEL rule rejecting `resources.requests` would turn a silent no-op into an error at create time. Worth settling either way, since this is in the immutable spec and can't be corrected on an existing template.

internal/proto/ateompb/ateom.proto
```diff
@@ -238,9 +245,15 @@ message RestoreWorkloadRequest {
   optional EgressGateway egress_gateway = 12;
 
   // The object storage URI of the ActorTemplate's golden snapshot.
-  // Set only when scope is SNAPSHOT_SCOPE_DATA_ON_GOLDEN. Mirrors the
   // snapshot_uri contract (field 8).
```
BenTheElder, line 270: 🤖 **nit** 🟢 – A line went missing here. This previously read "Set only when scope is SNAPSHOT_SCOPE_DATA_ON_GOLDEN. Mirrors the snapshot_uri contract (field 8)." and now reads "...golden snapshot." followed by a dangling "snapshot_uri contract (field 8)." The deletion looks incidental to this PR, and it drops the only statement of when the field is set.

demos/counter/counter-microvm.yaml.tmpl
chw120: SG! `limits` is only being used in scheduling. Removed `requests` in demos to avoid the confusion.

cmd/ateom-gvisor/runsc.go
chw120: Good catch. Restored it.

cmd/ateom-microvm/run.go
chw120: Updated code to size guest container memory cgroup to post-reserve guest RAM.

pkg/api/v1alpha1/actortemplate_types.go
chw120: Added more validations to reject it.

internal/proto/ateompb/ateom.proto
chw120: Restored.

cmd/ateom-microvm/run.go
```diff
@@ -593,6 +636,43 @@ func (s *AteomService) guestConfig(rr resolvedRuntime) (memMiB, vcpus int, kpara
...
+// limits: CPU passes through unchanged (kata-agent sets the CFS quota), while
+// memory is reduced by the VMM reserve so the container cgroup limit inside the
+// guest matches the guest VM's actual RAM rather than the (larger) outer limit.
+func (s *AteomService) guestSize(sz sizing.SandboxSize) sizing.SandboxSize {
+	if sz.MemoryBytes <= 0 {
+		return sz
+	}
+	memMiB, err := resolveGuestMemMiB(sz.MemoryBytes, s.memReserveMiB, 0)
+	if err != nil {
+		return sz
```
BenTheElder: 🤖 **should-fix** 🟡 – This swallows the error and returns the size unreduced, which is the mismatch the fix was for. When `resolveGuestMemMiB` fails, `sz.MemoryBytes` is still the outer declared limit, so the guest container cgroup goes back to being larger than the VM's RAM.

Cold boot is protected because it checks `resolveGuestMemMiB`'s error itself and aborts. `restoreFullScope` doesn't — the VM size comes from the snapshot — so the reachable case is raising `--vmm-mem-reserve-mib` after an actor was snapshotted: the next restore quietly gets the bad pairing instead of a clear failure. `TestGuestSize` covers the reduce and unset paths but not this one.

Returning the error, or at least logging, would keep it from failing silently.

Related: `coldBootActor` at line 403 does the same translation inline (`guestSize := sz; guestSize.MemoryBytes = memMiB * MiB`) rather than calling this helper. Two implementations that agree on the happy path and disagree on this one.

cmd/ateom-microvm/run.go
```diff
@@ -593,6 +636,43 @@ func (s *AteomService) guestConfig(rr resolvedRuntime) (memMiB, vcpus int, kpara
...
+			declaredMiB, reserveMiB, m, minGuestMemMiB)
+	}
+	return m, nil
+}
+
+// guestSize translates an actor's declared limits into the effective in-guest
+// limits: CPU passes through unchanged (kata-agent sets the CFS quota), while
+// memory is reduced by the VMM reserve so the container cgroup limit inside the
+// guest matches the guest VM's actual RAM rather than the (larger) outer limit.
+func (s *AteomService) guestSize(sz sizing.SandboxSize) sizing.SandboxSize {
```
BenTheElder: 🤖 **question** 🟢 – With this change the container cgroup limit lands at exactly the guest's total RAM, so it still can't bind before the guest kernel OOMs — the kernel, agent and init need some of that RAM, so the workload hits the guest OOM killer just short of its cgroup limit. Closer than before, but the clean per-container OOM kill is still out of reach.

It also applies uniformly: every container in a multi-container actor gets a limit equal to the whole guest RAM, so they aren't bounded relative to each other. That may well be the intent — the VM's RAM is the real envelope and the per-container cgroup is belt-and-braces — but if so it's worth saying in the `guestSize` comment, since "right-sized to the actor's limits" reads as if the cgroup is doing the enforcing.

BenTheElder (commented):
mostly resolved now, two more agent findings. thanks for continuing to work on this

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -479,6 +479,14 @@ message Worker {
     STATE_DRAINING = 2;
   }
   State state = 11;
+
+  // Capacity is the worker pod's compute capacity available to host an actor
+  // sandbox, taken from the ateom container's resource limits. The scheduler
+  // only places an actor on a worker whose capacity is >= the actor's declared
+  // resource limits. Zero means "unknown/unset": treated as unconstrained so
+  // placement is not blocked (matching the pre-capacity behavior).
+  int64 cpu_milli_capacity = 12;    // CPU capacity in millicores.
```
juli4n: maybe wrap these two fields in a message for better logical grouping?

cmd/ateom-microvm/run.go
chw120: Added more comments to clarify it. If it's better to subtract the guest-side overhead, I can make changes in this or following PR.

cmd/ateom-microvm/run.go
chw120: Updated line 403 to call guestSize().

pkg/proto/ateapipb/ateapi.proto
chw120: Created message WorkerCapacity.

cmd/ateom-microvm/run.go
```diff
@@ -599,6 +642,56 @@ func (s *AteomService) guestConfig(rr resolvedRuntime) (memMiB, vcpus int, kpara
...
+// guest RAM — so containers are not bounded relative to each other, and because
+// the guest kernel, agent and init consume part of that RAM the workload reaches
+// the guest OOM killer just before the cgroup limit binds. Sizing the cgroup to
+// the guest's RAM keeps the two numbers from contradicting each other; enforcing
+// a per-container share would need those overheads subtracted first.
+//
+// An error means the declared limit cannot be honored (see resolveGuestMemMiB);
+// callers must not fall back to the unreduced size, which is the mismatch this
+// translation exists to avoid.
+func (s *AteomService) guestSize(sz sizing.SandboxSize) (sizing.SandboxSize, error) {
```
BenTheElder, line 708: 🤖 **question** 🟢 – Memory gets a VMM reserve, a guest floor and an admission check; CPU gets none of that, so the same declared number means different things on the two runtimes.

On gVisor the sentry is the workload and sits in the same cgroup leaf, so a 2000m limit covers the sandbox too. On the micro-VM the guest cgroup is capped at 2000m, but cloud-hypervisor's vCPU threads, virtiofsd and ateom are host processes sharing the worker pod's CPU quota alongside it — the same reason memory needs `vmmMemReserveMiB`. The scheduler compares the declared limit against worker capacity with `>=`, so an actor may be placed with the whole worker's CPU and leave nothing for the VMM.

CPU is compressible, so this throttles rather than fails, which is presumably why there is no reserve. Worth saying so in this comment though — "passes through unchanged" reads as "the workload gets what it asked for", and on this runtime it gets somewhat less.

BenTheElder: [we should defer this to follow-up but I think it probably needs considering]

BenTheElder:
this PR might be affecting e2e?

cmd/ateom-microvm/run.go
chw120: Updated the comment.

chw120:
FYI, the CI failure is due to known flaky tests (tracked in #675 and #619) and is unrelated to the changes in this PR.

chw120:
> FYI, the CI failure is due to known flaky tests (tracked in #675 and #619) and is unrelated to the changes in this PR.

@BenTheElder

BenTheElder (dismissed):

BenTheElder (approved):

PR #680, opened 2026-08-01
Security status report SKILL

mtaufen:
This is a (partially) vibe-coded skill that generates a report for each threat in the threat model, and renders a png graph to give a high level gut-feel overview of Substrate's current security posture, based on how well each threat model item has been addressed. 

Now we can all run these reports to keep tabs on the rough status of the repo :). Expect some variation between runs, depending on model, harness, etc.

Example report json excerpt:
```
[
...
{
  "threat_id": "T-08",
  "threat": "Attacker who can create ActorTemplates can read or write any storage buckets atelet has access to.",
  "quality": 0.2,
  "strengths": "ActorTemplate spec is immutable post-creation, preventing retroactively modifying SnapshotsConfig.Location on existing templates.",
  "weaknesses": "ActorTemplate allows specifying arbitrary GCS bucket URIs in SnapshotsConfig.Location without validation or restricted bucket mapping. Atelet performs object storage operations using shared node/pod credentials rather than actor identity-derived credentials.",
  "citations": [
    "pkg/api/v1alpha1/actortemplate_types.go",
    "cmd/ateapi/internal/controlapi/workflow_suspend.go",
    "cmd/atelet/internal/ategcs/gcs.go"
  ]
}
...
]
```

And we can get nice charts like this. The amount of red is how bad we're doing ;). Everything else is "better than red." 

chart

<details>
<summary>Original chart for reference. Click to expand.</summary>
Original chart for reference below. The newer one above (e5b9bcd729b39ec7a44edb0dedc00d9a6200bcf4) does a better job of using the total surface area of a given color to indicate security posture.
Screenshot 2026-07-31 at 4 56 12 PM
</details>


mtaufen:
@dberkov addressed offline feedback, PTAL

.agents/skills/security-status-report/scripts/template_dispatch.py
```diff
@@ -0,0 +1,66 @@
...
+```'''
+        # TODO: Agent, update this part to ensure it matches the correct schema for you to invoke sub-agents via a tool call.
+        subagents.append({
+            "Prompt": prompt,
+            "Role": "Security Reviewer",
+            "TypeName": "self",
+            "Workspace": "inherit"
+        })
+
+    os.makedirs(os.path.dirname(output_path), exist_ok=True)
```
dberkov: If output_path is a filename without directory components (e.g. python3 template_dispatch.py subagents.json), os.path.dirname(output_path) returns "" (empty string). Calling os.makedirs("", exist_ok=True) raises FileNotFoundError: [Errno 2] No such file or directory: ''.

output_dir = os.path.dirname(output_path)
if output_dir:
    os.makedirs(output_dir, exist_ok=True)

.agents/skills/security-status-report/scripts/compile_report.py
```diff
@@ -0,0 +1,100 @@
...
+    
+    print(f"Report compiled successfully to {output_path}")
+    print(f"Summary: {total_count} total threats | {succeeded_count} succeeded | {len(failed_ids)} failed")
+    print(f"Average Quality Score: {avg_quality:.2f}")
+
+    if failed_ids:
+        print(f"Warning: The following {len(failed_ids)} threat(s) failed evaluation: {', '.join(failed_ids)}", file=sys.stderr)
+
+if __name__ == '__main__':
+    compile_report(sys.argv[1], sys.argv[2], sys.argv[3])
```
dberkov, line 102: nit:
if __name__ == '__main__':
    if len(sys.argv) < 4:
        print(f"Usage: {sys.argv[0]} <threats_json_path> <results_dir> <output_path>", file=sys.stderr)
        sys.exit(1)
    compile_report(sys.argv[1], sys.argv[2], sys.argv[3])

.agents/skills/security-status-report/scripts/compile_report.py
```diff
@@ -0,0 +1,100 @@
...
+        final_report.append(t)
+            
+    output_dir = os.path.dirname(output_path)
+    if output_dir:
+        os.makedirs(output_dir, exist_ok=True)
+    with open(output_path, 'w') as f:
+        json.dump(final_report, f, indent=2)
+
+    total_count = len(final_report)
+    avg_quality = (total_quality / total_count) if total_count > 0 else 0.0
```
dberkov: nit: total_quality only sums succeeded threats, but divides by total_count , treating failed/timed-out evaluations as 0.0. If 5 sub-agents fail due to a temporary network timeout, the reported average quality score is penalized.

.agents/skills/security-status-report/SKILL.md
```diff
@@ -0,0 +1,42 @@
...
+2. Complete the TODO in `.agents/scratch/security-status-report/template_dispatch.py`, meeting the following requirements:
+  a. Iterate over each threat from `docs/threats.json` to produce a list of invocations that matches your tool for invoking sub-agents.
+  b. Each invocation must address a single threat, use the below prompt, and instruct the sub-agent to output the correct schema.
+3. Execute the script using `run_command` from the repository root, specifying `.agents/scratch/security-status-report/subagents.json` as the output file argument (`python3 .agents/scratch/security-status-report/template_dispatch.py .agents/scratch/security-status-report/subagents.json`).
+4. Read `.agents/scratch/security-status-report/subagents.json` and copy its exact JSON array into your tool for invoking subagents. DO NOT manually craft or bypass the invocations. Run ALL generated sub-agents concurrently in a single tool call. You already updated the script in step 2 to output exactly what you need, so no modifications to the output should be necessary unless you made a mistake.
```
dberkov: nit:
"Run ALL generated sub-agents concurrently in a single tool call." With 43 threats in docs/threats.json, spawning 43 sub-agents simultaneously can hit platform rate limits or concurrency limits on some LLM harnesses. Adding a note or batching fallback (e.g. "If your environment limits concurrency, run in batches of 10-15") would make the skill more resilient across different runners.

.agents/skills/security-status-report/scripts/compile_report.py
mtaufen: The review agents keep going back and forth on this one :).
These are just summary statistics printed out from the script anyway, probably best to delete them.

.agents/skills/security-status-report/SKILL.md
mtaufen: Updated.

.agents/skills/security-status-report/scripts/compile_report.py
mtaufen: Fixed

.agents/skills/security-status-report/scripts/template_dispatch.py
mtaufen: Fixed

.agents/skills/security-status-report/scripts/compile_report.py
mtaufen: Removed the average

.agents/skills/security-status-report/SKILL.md
```diff
@@ -0,0 +1,42 @@
...
+  a. Iterate over each threat from `docs/threats.json` to produce a list of invocations that matches your tool for invoking sub-agents.
+  b. Each invocation must address a single threat, use the below prompt, and instruct the sub-agent to output the correct schema.
+3. Execute the script using `run_command` from the repository root, specifying `.agents/scratch/security-status-report/subagents.json` as the output file argument (`python3 .agents/scratch/security-status-report/template_dispatch.py .agents/scratch/security-status-report/subagents.json`).
+4. Read `.agents/scratch/security-status-report/subagents.json` and copy its exact JSON array into your tool for invoking subagents. DO NOT manually craft or bypass the invocations. Run ALL generated sub-agents concurrently in a single tool call by default unless bound by model rate limts or harness consurrency limits (use batching intelligently if needed). You already updated the script in step 2 to output exactly what you need, so no modifications to the output should be necessary unless you made a mistake or hit limits.
```
dberkov: typos: 
* rate limts → rate limits
* harness consurrency limits → harness concurrency limits

.agents/skills/security-status-report/SKILL.md
```diff
@@ -0,0 +1,42 @@
...
+# Task
+Generate a security status report by evaluating threats listed in `docs/threats.json`.
+
+# Workflow
+
+1. Copy `.agents/skills/security-status-report/scripts/template_dispatch.py` to `.agents/scratch/security-status-report/template_dispatch.py`.
+2. Complete the TODO in `.agents/scratch/security-status-report/template_dispatch.py`, meeting the following requirements:
+  a. Iterate over each threat from `docs/threats.json` to produce a list of invocations that matches your tool for invoking sub-agents.
+  b. Each invocation must address a single threat, use the below prompt, and instruct the sub-agent to output the correct schema.
+3. Execute the script using `run_command` from the repository root, specifying `.agents/scratch/security-status-report/subagents.json` as the output file argument (`python3 .agents/scratch/security-status-report/template_dispatch.py .agents/scratch/security-status-report/subagents.json`).
```
dberkov: This step hardcodes the tool name `run_command`, while the rest of the doc deliberately stays harness-neutral ("your tool for invoking sub-agents"). Suggest the same phrasing here for portability across harnesses.

.agents/skills/security-status-report/SKILL.md
```diff
@@ -0,0 +1,42 @@
...
+4. Read `.agents/scratch/security-status-report/subagents.json` and copy its exact JSON array into your tool for invoking subagents. DO NOT manually craft or bypass the invocations. Run ALL generated sub-agents concurrently in a single tool call by default unless bound by model rate limts or harness consurrency limits (use batching intelligently if needed). You already updated the script in step 2 to output exactly what you need, so no modifications to the output should be necessary unless you made a mistake or hit limits.
+5. Wait for all sub-agents to complete.
+6. Run `python3 .agents/skills/security-status-report/scripts/compile_report.py docs/threats.json .agents/scratch/security-status-report .agents/scratch/security-status-report/final.json` to produce the final report.
+7. Run `python3 .agents/skills/security-status-report/scripts/render_chart.py .agents/scratch/security-status-report/final.json .agents/scratch/security-status-report/chart.png` to render a bar chart of the scores.
+
+## Prompt template for each sub-agent that should be used in the script
```
dberkov: This prompt template is duplicated in `template_dispatch.py`, and the two can drift (the script version also interpolates the actual ID into the explanatory clause, producing e.g. "where T-08 matches the id you were initially provided", which reads oddly). Consider making the script the single source of truth and having this doc point at it.

.agents/skills/security-status-report/scripts/compile_report.py
```diff
@@ -0,0 +1,100 @@
...
+import os
+
+def compile_report(threats_json_path, results_dir, output_path):
+    with open(threats_json_path, 'r') as f:
+        threats = json.load(f)
+        
+    final_report = []
+    succeeded_count = 0
+    failed_ids = []
+    total_quality = 0.0
```
dberkov: `total_quality` is accumulated but never used. Either drop it, or print an average score in the summary line — the latter would be a nice addition.

.agents/skills/security-status-report/scripts/compile_report.py
```diff
@@ -0,0 +1,100 @@
...
+        threat_id = t.get("threat_id", "unknown")
+        result_file = os.path.join(results_dir, f"{threat_id}.json")
+        
+        if os.path.exists(result_file):
+            try:
+                with open(result_file, 'r') as rf:
+                    res = json.load(rf)
+
+                # Check threat_id matching
+                if res.get("threat_id") and res.get("threat_id") != threat_id:
```
dberkov: A result file with no `threat_id` at all silently passes this check. If that leniency is intentional, fine; otherwise consider requiring the field since the schema asks for it.

.agents/skills/security-status-report/scripts/compile_report.py
```diff
@@ -0,0 +1,100 @@
...
+                            if isinstance(c, list):
+                                t["citations"] = [str(item) for item in c]
+                            elif isinstance(c, str):
+                                t["citations"] = [c]
+                            else:
+                                t["citations"] = []
+            except Exception as e:
+                t["error"] = f"Failed to parse agent JSON: {e}"
+        else:
+            t["error"] = "The evaluation sub-agent timed out or failed to produce a valid JSON."
```
dberkov: This message over-claims the cause — the sub-agent may simply have written to the wrong path. Suggest something neutral like `"no result file found at <path>"`.

.agents/skills/security-status-report/scripts/render_chart.py
```diff
@@ -0,0 +1,112 @@
...
+        errors.append(is_error)
+
+        try:
+            score = float(raw_score) if not is_error else 0.0
+        except (ValueError, TypeError):
+            score = 0.0
+        score = max(0.0, min(1.0, score))
+        scores.append(score)
+
+        if score > 0.85:
```
dberkov: These thresholds are policy, so a brief comment would help. Also note the asymmetric boundary handling: exactly 0.85 is yellow (`> 0.85` vs `>= 0.7`) — fine if intended, just worth documenting.

.agents/skills/security-status-report/scripts/render_chart.py
```diff
@@ -0,0 +1,112 @@
...
+            ax.text(x, y_pos, f"{s:.2f}", color='black', ha='center', va=va, fontsize=8, fontweight='bold', 
+                    bbox=dict(facecolor='white', alpha=0.7, edgecolor='none', pad=1.5))
+
+    # Annotate errors with vertical "ERROR" text above x-axis
+    for i, err in enumerate(errors):
+        if err:
+            ax.text(i, 0.02, "ERROR", rotation=90, ha='center', va='bottom', fontsize=8, fontweight='bold', color='white')
+
+    ax.set_ylim(0.0, 1.0)
+    ax.set_xlim(-0.5, len(threat_ids) - 0.5)
```
dberkov, line 94: If the report is empty this becomes `ax.set_xlim(-0.5, -0.5)`, which matplotlib rejects/warns on. A guard with a friendly "no threats to plot" exit would be cleaner.

hack/export-threats.py
```diff
@@ -0,0 +1,107 @@
...
+        line = line.strip()
+        if not line.startswith('|'):
+            in_table = False
+            continue
+            
+        line_escaped = line.replace('\\|', '\x00PIPE\x00')
+        cells = [c.strip().replace('\x00PIPE\x00', '\\|') for c in line_escaped.split('|')[1:-1]]
+        if not in_table:
+            lower_cells = [unescape_md(c.lower()) for c in cells]
+            if 'threat' in lower_cells:
```
dberkov, line 44: Header detection re-arms on any pipe-row containing a cell exactly equal to "threat", even mid-document, so a data row could theoretically be misparsed as a new header. Low risk given the current doc — not blocking, just noting.

docs/threat-model.md
```diff
@@ -4,9 +4,6 @@
 
 Last updated: Jun 25, 2026
 
-TODO:
-- [ ] file GitHub issues for each threat
```
dberkov, line 8: The removed TODO included "file GitHub issues for each threat", and the "GitHub Issue" column is gone too. Threat IDs are a better stable handle, but if issue-linking is still intended, that intent is now recorded nowhere — consider tracking it as an issue instead of deleting outright.

.agents/skills/security-status-report/scripts/compile_report.py
```diff
@@ -0,0 +1,100 @@
...
+            try:
+                with open(result_file, 'r') as rf:
+                    res = json.load(rf)
+
+                # Check threat_id matching
+                if res.get("threat_id") and res.get("threat_id") != threat_id:
+                    t["error"] = f"Mismatched threat_id in result file: expected {threat_id}, got {res.get('threat_id')}"
+                else:
+                    # Validate quality score presence, type, and range
+                    if "quality" not in res:
```
dberkov, line 43: nit: In Python, bool is a subclass of int (isinstance(True, int) is True). Consequently, float(True) returns 1.0 and float(False) returns 0.0 without raising a ValueError or TypeError. If a sub-agent outputs "quality": true in JSON, float(res["quality"]) will silently convert it to 1.0

if "quality" not in res:
    t["error"] = "Result JSON missing 'quality' score"
elif isinstance(res["quality"], bool):
    t["error"] = f"Invalid boolean quality score: {res['quality']}"
else:
    try:
        score = float(res["quality"])
        ...

.agents/skills/security-status-report/scripts/compile_report.py
mtaufen: I can change it to "likely timed out ..." but in practice this is almost always the cause. The models wrote the correct path 100% of the time in local testing (using both Gemini pro and flash).

.agents/skills/security-status-report/scripts/render_chart.py
mtaufen: I think you'd just get a weird looking chart in this case, which is signal enough.

docs/threat-model.md
mtaufen: No issue linking is still intended

.agents/skills/security-status-report/scripts/compile_report.py
mtaufen: sg, will fix

.agents/skills/security-status-report/scripts/render_chart.py
mtaufen: will update with a comment, these really are not a policy, just visual assist

.agents/skills/security-status-report/scripts/compile_report.py
mtaufen: will fix

.agents/skills/security-status-report/scripts/compile_report.py
mtaufen: will just drop it

.agents/skills/security-status-report/SKILL.md
mtaufen: will deduplicate

.agents/skills/security-status-report/SKILL.md
mtaufen: will do

.agents/skills/security-status-report/SKILL.md
mtaufen: ack will do

.agents/skills/security-status-report/scripts/compile_report.py
mtaufen: updated the error

hack/export-threats.py
mtaufen: I think we probably need this, since the doc contains multiple tables. It needs to go from "not in_table" to "in_table" state and re-grab the headers multiple times.

dberkov (approved):

PR #682, opened 2026-08-01
feat: Add control plane telemetry - scheduler eligible workers histogram

Angelawork:
Part of #564 (Part 3)

- [x] Tests pass

## Description
This PR implements Part 3 of #564 by adding telemetry histogram instrumentation for `ate.scheduler.eligible_workers` in `ateapi`. It measures unassigned free worker capacity remaining after all scheduling constraint filters are applied, sampled at every scheduling decision. 

## Key Changes:
* Updated `Scheduler.Schedule()` to record eligible candidate workers per pool (`recordEligibleWorkers`).
Defined `SchedulingConstraintKey` (`ate.scheduling.constraint`) and constraint classification values (`none`, `required_nodes`, `selector`) in `internal/ateattr/ateattr.go`.
* Added unit tests in `scheduling_test.go` covering candidate counts, namespaced attributes, zero-capacity fleet states, empty fleets, sandbox class mismatches, draining workers, and constraint classifications.

## Testing
* `make test`
## E2E  Test


JeffLuoo:
Please rebase and fix e2e tests

internal/e2e/suites/metrics/metrics_test.go
```diff
@@ -126,3 +198,17 @@ func waitForStatus(t *testing.T, ctx context.Context, clients *e2e.Clients, acto
 	}
 	t.Fatalf("actor %q never reached %v", actorID, want)
 }
+
+func extractPrometheusLabelValue(line, labelName string) string {
```
JeffLuoo: Let's rename it to extractLabelValue as we are not always using Prometheus in the stack. We are using OTel SDK.

cmd/ateapi/internal/scheduling/scheduling.go
```diff
@@ -77,6 +86,25 @@ func WithIntn(intn func(n int) int) Option {
...
+func WithMeter(meter metric.Meter) Option {
+	return func(s *scheduler) {
+		if meter == nil {
+			return
+		}
+		hist, err := meter.Int64Histogram(
+			eligibleWorkersMetric,
+			metric.WithUnit("{worker}"),
+			metric.WithDescription("Number of eligible workers available during scheduling given the constraint filters."),
+		)
```
JeffLuoo: Every other histogram in this repo sets buckets explicitly. The SDK default is [0, 5, 10, 25, 50, 75, 100, 250, ...], so 1 vs 5 eligible workers is one bucket and every percentile above p50 is noise for a typical fleet.

Let's define our own bucket here like:
```
metric.WithExplicitBucketBoundaries(0, 1, 2, 3, 5, 10, 20, 50, 100, 250)
```

to start with.

cmd/ateapi/internal/scheduling/scheduling.go
```diff
@@ -86,25 +114,105 @@ func New(source WorkerSource, opts ...Option) Scheduler {
 	return s
 }
 
+// Schedule filters the current worker fleet to find unassigned candidates matching the given constraints,
+// records the ate.scheduler.eligible_workers metric, and picks a random candidate if available.
```
JeffLuoo: Let's avoid being too specific in function comment. In the case of future refactoring, if we don't record `ate.scheduler.eligible_workers` metric here, we may forget to update the comment. Let's keep the comment generic like:
```
Schedule filters the current worker fleet to find unassigned candidates matching the given constraints
```

cmd/ateapi/internal/scheduling/scheduling.go
```diff
@@ -86,25 +114,105 @@ func New(source WorkerSource, opts ...Option) Scheduler {
...
+	// Filter for candidate workers that are unassigned and meet all scheduling constraints
 	var candidates []*ateapipb.Worker
 	for _, worker := range workers {
 		if worker.GetAssignment() == nil && s.Applies(worker, constraints) {
 			candidates = append(candidates, worker)
 		}
 	}
 
+	// Record telemetry on the number of eligible workers per pool/namespace before returning
+	s.recordEligibleWorkers(ctx, workers, candidates, constraints)
```
JeffLuoo:
```suggestion
        // Filter for candidate workers that are unassigned and meet all scheduling constraints
        matching := make([]*ateapipb.Worker, 0, len(workers))
        var candidates []*ateapipb.Worker
        for _, worker := range workers {
                if !s.Applies(worker, constraints) {
                        continue
                }
                matching = append(matching, worker)
                if worker.GetAssignment() == nil {
                        candidates = append(candidates, worker)
                }
        }

        // Record telemetry on the number of eligible workers per pool/namespace before returning
        s.recordEligibleWorkers(ctx, matching, constraints)
  ```

this loop already computed Applies for every worker and discarded it for the assigned ones. Keeping both subsets from one pass removes a second round of computation.

File: cmd/ateapi/internal/scheduling/scheduling.go [hunk 5]
```diff
@@ -86,25 +114,105 @@ func New(source WorkerSource, opts ...Option) Scheduler {
...
+	// Emit histogram observation for each worker pool key using standard ateattr keys
+	for k, count := range eligibleByPool {
+		s.eligibleWorkers.Record(ctx, count, metric.WithAttributes(
+			ateattr.WorkerPoolNamespaceKey.String(k.namespace),
+			ateattr.WorkerPoolNameKey.String(k.pool),
+			ateattr.SandboxClassKey.String(k.sandboxClass),
+			ateattr.SchedulingConstraintKey.String(k.constraint),
+		))
+	}
+}
```
JeffLuoo:
```suggestion
  func (s *scheduler) recordEligibleWorkers(ctx context.Context, matching []*ateapipb.Worker, constraints Constraints) {
        if s.eligibleWorkers == nil {
                return
        }
  
        // Sandbox class and constraint are constant across every key: Applies requires
        // an exact class match, and the classification is per call. They belong on the
        // Record call, not in the key.
        type key struct{ namespace, pool string }
        eligibleByPool := make(map[key]int64)
        for _, w := range matching {
                k := key{w.GetWorkerNamespace(), w.GetWorkerPool()}
                if _, ok := eligibleByPool[k]; !ok {
                        eligibleByPool[k] = 0
                }
                if w.GetAssignment() == nil {
                        eligibleByPool[k]++
                }
        }

        // No pool matched the constraints at all. Emit a single zero-valued series so
        // "nothing is schedulable" stays visible; empty namespace/pool marks it. The
        // label set matches the per-pool series, so dashboards need no special case.
        if len(eligibleByPool) == 0 {
                eligibleByPool[key{}] = 0
        }

        constraintStr := classifyConstraint(constraints)
        for k, count := range eligibleByPool {
                s.eligibleWorkers.Record(ctx, count, metric.WithAttributes(
                        ateattr.WorkerPoolNamespaceKey.String(k.namespace),
                        ateattr.WorkerPoolNameKey.String(k.pool),
                        ateattr.SandboxClassKey.String(constraints.SandboxClass),
                        ateattr.SchedulingConstraintKey.String(constraintStr),
                ))
        }
  }
  ```

The suggested changes suggest two improvement:

1. Avoid calling the apply on constraint again.
2. When the `len(eligibleByPool) == 0` condition is met, not all attributes are emitted in metrics. This will be confusing when we plot the chart in the dashboard. Suggestion is to set the value to 0, and let the emitting logic below to handle the emitting with all 4 labels. A general rule for metric labels (attributes) is never set empty string if possible.

File: cmd/ateapi/internal/scheduling/scheduling.go [hunk 6]
```diff
@@ -77,6 +86,25 @@ func WithIntn(intn func(n int) int) Option {
 	return func(s *scheduler) { s.intn = intn }
 }
 
+// WithMeter configures the meter used to create telemetry instruments for the scheduler.
+func WithMeter(meter metric.Meter) Option {
```
JeffLuoo: `WithMeter` will silently drop the error when instrument-creation errors are seen.

You can reference https://sourcegraph.com/r/github.com/agent-substrate/substrate@main/-/blob/cmd/ateapi/internal/controlapi/metrics.go?L30-35 and follow how `RegisterWorkerCount` registers the metric, it will explicitly print out errors when seen.

JeffLuoo: Also, consider creating a `metrics.go` file in scheduling folder to align with other metrics implementation. You can move all metrics-related logic to the new `metrics.go` file.

JeffLuoo (changes requested):

cmd/ateapi/internal/scheduling/metrics.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// Metric identifier for tracking eligible worker counts.
+const eligibleWorkersMetric = "ate.scheduler.eligible_workers"
+
+// RegisterEligibleWorkers registers the ate.scheduler.eligible_workers histogram instrument against meter.
+// Returns (nil, nil) if meter is nil. If registration fails, an error is returned.
+func RegisterEligibleWorkers(meter metric.Meter) (metric.Int64Histogram, error) {
```
JeffLuoo: You don't need to make `RegisterEligibleWorkers` a public function because it's only used in this file.

JeffLuoo: Also consider using the name `newEligibleWorkers` because the term `Register` usually infer a callback.

JeffLuoo (changes requested):

JeffLuoo (dismissed):

JeffLuoo:
Please do not use `Fixes #564 (Part 3)` in the description as it will close the bug. Use `Part of #564 (Part 3)`

JeffLuoo (approved):

JeffLuoo:
For some reasons the CI is not running..  I close and reopen the PR to see if it will trigger the presubmit again.

Issue #477, opened 2026-07-21
Actor Image Upgrade Design


zoez7:
This issues tracks the design for how to upgrade actor's base image during the lifetime of an actor

* **CUJ**: As an actor developer, I would like to ship new binary versions or patch CVEs of existing actor instances.
* **Rule** (agreed upon during the summit): Actors cannot upgrade using an existing memory snapshot.
* **Design idea discussed during the summit**: 
   * Users should be able to create multiple versions of an ActorTemplate, once a ActorTemplateVersion is created, its actor image version is immutable.
   * Users should be able to define upgrade strategy, such as actor termination policy, how the upgrade is applied. For M2, the minimalist default upgrade strategy could be "on next resume, use new ActorTemplateVersion".

This requires more investigation for what the ActorTemplateVersion API should look like with different SnapshotConfigs, and what happens exactly during an upgrade.

zoez7:
Initial design draft https://docs.google.com/document/d/1OsHMah0M7vvl1OBtHFtYJngItHWDYcenSCQ1gxhYseU/edit?tab=t.0

zoez7:
2026-08-21 Update : Below is the updated design proposal for actor upgrade:

## Objective

As an actor developer, I would like to ship new features in my container image, or patch CVEs, for existing actor instances. How to do this in Substrate?

## Design decisions

For M2, start with building primitives that both an automated rollout system and substrate user can use. Substrate itself will not trigger suspend or resume of any actors, but will provide abilities to


1. Move the `ActorTemplate` (AT) CRD into Substrate without change. The new Substrate resource is **immutable**; shipping a new image or sandbox config means creating a new `ActorTemplate`. There is no separate `ActorTemplateVersion` resource.  
2. Upgrade each actor by first updating its template reference (`UpdateActor`), then resuming it (`ResumeActor`), and rollback to the previous template on failure.  
3. Get template distribution of all actors, to confirm rollout completeness.


Post M2, when we have more data of the shape of rollout policies that will be useful to customers, we can explore building rollout controllers in substrate.

## Proposed API

### Immutable ActorTemplate

`ActorTemplate` will be a Substrate resource managed through ate-api, instead of a Kubernetes custom resource (see discussion in [issues/368](https://github.com/agent-substrate/substrate/issues/368)). Substrate owns the lifecycle of each AT. The CRD schema moves over without change; the only semantic change is that the resource is immutable after creation.


[...]

Issue #451, opened 2026-07-17
Support resuming actors with golden memory snapshot and DurableDirectory


zoez7:
In the Actor Upgrades discussion during summit, we defined a rule that
"Substrate cannot upgrade into an existing memory snapshot".

When resuming an actor after ActorTemplateVersion upgrade, we will need to use
1. The memory portion from the GoldenSnapshot of the upgraded ActorTemplate Version and
2. The DurableDirectory from the previous run.

zoez7:
Thinking about it more, I think this could be useful as a `SnapshotScope` type of its own https://github.com/agent-substrate/substrate/blob/e1fa0040a694c84ef04450c78e9d779d32f9b452/pkg/api/v1alpha1/actortemplate_types.go#L253-L265 not just as "magic" used during upgrades. ("As a substrate user, I want to use memory snapshot only for the purpose of avoiding cold boot")

But if we drop the idea of "mandatory golden".... having a `GOLDEN_MEMORY_WITH_DUR_DIR` enum value won't make much sense. cc: @juli4n @EItanya

PR #683, opened 2026-08-01
microvm: onResume fromData Golden - resume Data snapshots on the golden snapshot (#451, phase 1)

dberkov:
Phase 1 of #451, micro-VM only. gVisor support will follow in a separate PR.

## What

Adds a `snapshotsConfig.onResume` block to the ActorTemplate, selecting per snapshot situation what supplies the guest state at resume: each field names what is being resumed *from*, the value names the boot source. `onPause`/`onCommit` remain pure capture scopes (`Full | Data`), and the `ActorSnapshot` records plain `Full`/`Data` content; the golden-combine is strictly a restore-time behavior.

```yaml
snapshotsConfig:
  onPause: Data
  onCommit: Data
  onResume:
    fromData: Golden   # ColdBoot (default) | Golden — Data snapshots resume as
  location: gs://…     # golden memory + fs delta with the actor's data layered on
```

Under `fromData: Golden`, a data-only restore combines the template's **golden snapshot** (guest memory + full fs delta) with the actor's captured durable data — the actor comes back with the golden's warm state over its own data instead of cold-booting. `fromData` applies to every resume of a Data snapshot — commit or pause checkpoint. Future situations (e.g. resuming a Full snapshot across a template upgrade, #477) get sibling fields under `onResume` rather than overloading this one.

## How

- **CRD** (`pkg/api/v1alpha1`): new `OnResumeConfig`/`ResumeSource` types; `fromData: Golden` is CEL-gated to `sandboxClass: microvm`.
- **Control plane** (`ateapi`): resume derives the behavior from the template's `onResume` configuration — for a Data durable snapshot or a Data pause checkpoint it resolves the golden snapshot's location and sends the restore-only `DATA_ON_GOLDEN` wire scope, failing early (with an actionable error) if the golden snapshot is missing or not Full. Golden actors always commit `Full` regardless of `onCommit` — their snapshot is the base the combine needs.
- **Wire APIs** (`ateletpb`/`ateompb`): restore-only `SNAPSHOT_SCOPE_DATA_ON_GOLDEN`, rejected on checkpoints; `golden_snapshot_uri_prefix` is a top-level `RestoreRequest` field because the actor's snapshot may be a local pause checkpoint while the golden snapshot is always external.
- **atelet**: stages a single combined folder — the actor's files (the durable-dir tar) win name collisions, the golden snapshot supplies the rest. External restores download both halves concurrently; local (pause) restores copy the actor's files from the local checkpoint dir while the golden's files download in parallel. The golden manifest's pinned sandbox binaries run the restored guest and are recorded on-node for later checkpoints.
- **ateom-microvm**: `DATA_ON_GOLDEN` restores through the existing Full path — cloud-hypervisor relaunches from the golden's guest files while the durable virtio-fs share serves the actor's re-materialized data. gVisor rejects the scope (defense-in-depth behind the CRD gate).

## Testing

- Unit: envtest CEL cases, converter tables, atelet request validation (checkpoint rejection, golden-URI rules incl. local+golden), combined-download test against a fake object store, resume-workflow golden-resolution cases for both the durable and pause paths.
- e2e (`suites/demo`): OnGolden lifecycle cases for the commit path, the pause path, and a two-durable-volume variant (micro-VM lane only), asserting counters across pause/suspend and the recorded Data content scope.
- Manually verified on a live GKE cluster: Data commit and Data pause both resume with the golden guest's memory over the actor's data, on different pods.


demos/counter/counter-microvm.yaml.tmpl
```diff
@@ -131,6 +131,11 @@ spec:
     matchLabels:
       workload: counter-microvm
   snapshotsConfig:
+    onPause: Full
+    onCommit: Data
+    # A Data commit resumes by combining the durable data with the golden
+    # snapshot's guest state (memory + fs delta) instead of cold-booting.
+    dataResumePolicy: OnGolden
```
zoez7: Nit: should this be
```
onResume:
    fromData: useGolden
```

to be consistent with the OnCommit and OnPause above? then upgrade can support

```
onResume:
    fromData: useGolden
    fromFullOnUpgrade: useGolden
```

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -101,10 +123,45 @@ func (s *LoadActorForResumeStep) Execute(ctx context.Context, input *ResumeInput
...
+	// OnGolden, resolve the golden snapshot's location so the restore can
+	// combine the golden's guest state with the actor's data. The pending
+	// restore is data-only when the actor is paused with a Data pause scope
+	// (the local snapshot takes precedence at restore), or when its durable
+	// snapshot holds Data. Valid Full snapshots restore from their own
+	// content and ignore the policy.
+	if actorTemplate.Spec.SnapshotsConfig.DataResumePolicy == atev1alpha1.DataResumePolicyOnGolden {
+		dataOnly := false
+		if actor.GetLocalSnapshotInfo() != nil {
+			dataOnly = actorTemplate.Spec.SnapshotsConfig.OnPause == atev1alpha1.SnapshotScopeData
```
zoez7, line 145: This is making me think - should paused actor's LocalSnapshot also be represented as a `ActorSnapshot` object for consistency?

cmd/atelet/main.go
```diff
@@ -566,20 +596,42 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 		t := time.Now()
 		switch req.GetType() {
 		case ateletpb.CheckpointType_CHECKPOINT_TYPE_EXTERNAL:
-			if err := s.downloadExternalCheckpoint(gctx, req.GetExternalConfig().GetSnapshotUriPrefix(), checkpointDir, sandboxRec.SnapshotFiles); err != nil {
+			if goldenRec != nil {
```
zoez7: Nit: This should check `if req.GetScope() == ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA_ON_GOLDEN` ?

cmd/atelet/main.go
```diff
@@ -566,20 +596,42 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+			// snapshot: the actor's files come from the local checkpoint dir,
+			// the golden's from object storage, concurrently.
+			gLocal, gLocalCtx := errgroup.WithContext(gctx)
+			gLocal.Go(func() error {
+				if err := s.copyLocalCheckpoint(gLocalCtx, req.GetLocalConfig().GetSnapshotPrefix(), ateompath.LocalCheckpointsDir(actorUID), checkpointDir, sandboxRec.SnapshotFiles); err != nil {
+					return ateerrors.CrashIfReason(ctx, err, ateerrors.ReasonTerminalFileSystemError)
+				}
+				return nil
+			})
+			if goldenRec != nil {
```
zoez7: Same here: This should check if `req.GetScope() == ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA_ON_GOLDEN` and throw error if goldenRec is nil

cmd/atelet/main.go
```diff
@@ -566,20 +596,42 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+				gLocal.Go(func() error {
+					if err := s.downloadExternalCheckpoint(gLocalCtx, req.GetGoldenSnapshotUriPrefix(), checkpointDir, goldenOnlyFiles(sandboxRec.SnapshotFiles, goldenRec.SnapshotFiles)); err != nil {
```
zoez7, line 623: we need to cache golden snapshot on the node for Paused Actors? If not, this means we download golden snapshot each time we un-Pause an actor which hurts latency.

cmd/atelet/main.go
```diff
@@ -553,6 +555,34 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+			return nil, ateerrors.CrashIfReason(ctx, fmt.Errorf("while unmarshalling golden sandbox record: %w", err), ateerrors.ReasonInvalidSandboxAsset)
+		}
+		if goldenRec.SandboxClass != sandboxRec.SandboxClass {
+			return nil, status.Errorf(codes.FailedPrecondition, "golden snapshot sandbox class %q does not match actor snapshot sandbox class %q", goldenRec.SandboxClass, sandboxRec.SandboxClass)
+		}
+	}
+
+	// The record whose pinned binaries run the restored workload: the golden's
+	// for a DATA_ON_GOLDEN restore, the snapshot's own otherwise.
+	runtimeRec := sandboxRec
```
zoez7, line 578: Can filesystem snapshots also be sensitive to runsc binary versions?  I assume it's possible? So if down the road `runsc fscheckpoint` changes the params it requires in a new runsc version, and the sandbox records don't match, we can't restore either way?

cmd/atelet/main.go
```diff
@@ -553,6 +555,34 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+		}
+		if goldenRec, err = unmarshalSandboxRecord(manifest); err != nil {
+			return nil, ateerrors.CrashIfReason(ctx, fmt.Errorf("while unmarshalling golden sandbox record: %w", err), ateerrors.ReasonInvalidSandboxAsset)
+		}
+		if goldenRec.SandboxClass != sandboxRec.SandboxClass {
+			return nil, status.Errorf(codes.FailedPrecondition, "golden snapshot sandbox class %q does not match actor snapshot sandbox class %q", goldenRec.SandboxClass, sandboxRec.SandboxClass)
+		}
+	}
+
+	// The record whose pinned binaries run the restored workload: the golden's
```
zoez7, line 572: I think this comment should explain why it chose goldenRec over runtimeRec.

zoez7 (changes requested):

cmd/ateapi/internal/controlapi/workflow_resume.go
dberkov: I dont think so, since PAUSED snapshots are temporary and not supposed to be persistence. It is not possible to clone from it, etc...

demos/counter/counter-microvm.yaml.tmpl
dberkov: Adopted in 6d51901e — reworked as a structured `onResume` block: `onResume: { fromData: Golden }`, values `ColdBoot` (default) | `Golden`. Fields name the snapshot situation being resumed from, values name the boot source, so future situations (e.g. resuming a Full snapshot across a template upgrade) become sibling fields instead of overloading one knob. The gvisor CEL gate moved onto `onResume.fromData`.

cmd/atelet/main.go
dberkov: Done in 6d51901e — the branch now checks `req.GetScope() == SNAPSHOT_SCOPE_DATA_ON_GOLDEN` explicitly, with a defensive error if the golden record is somehow missing at that point.

cmd/atelet/main.go
dberkov: Done in 6d51901e — same explicit scope check plus the missing-goldenRec error. One deviation from the suggested placement: the check is hoisted above the errgroup so a violated invariant fails before the local-copy goroutine starts, rather than returning while it is still writing into the restore dir.

cmd/atelet/main.go
dberkov: Agreed — un-pause should not pay a golden download every time. Golden snapshots are immutable and shared across a template's actors, so a node-local content-addressed cache (keyed by the golden snapshot URI, alongside atelet's existing asset cache) would fix this for every golden restore path — including the pre-existing golden-fallback boot, which re-downloads today too. Filed #690 to track it as a follow-up.

cmd/atelet/main.go
dberkov: Done in 6d51901e — the comment now explains the choice: the guest state being resumed *is* the golden snapshot's memory image, which must be resumed by the exact binary versions that produced it; the actor's snapshot contributes only a plain durable-data tar that no binary version reads back.

cmd/atelet/main.go
dberkov: This is not just runc. micro VM has the similar issue. We will deal with it when will support updating sandboxConfig.

zoez7 (approved):

PR #691, opened 2026-08-02
Enable MicroVM Benchmarking

maxsmythe:
Currently we are unable to easily automate the benchmarking of micro VMs.

This PR:

* Refactors the microVM setup scripts to allow discrete installation of uVM sanbox configs separate from counter demo
* Pipes a MicroVM container test  config through the benchmarking system
* Updates default atelet daemonset to tolerate all kinds of sandboxClass taints

### DIFF
--- a/hack/run-microvm-demo.sh
+++ b/hack/run-microvm-demo.sh
@@ -18,6 +18,13 @@
 # local kind cluster use hack/run-microvm-demo-kind.sh (which sets the kind env
 # and calls this script), mirroring install-ate.sh / install-ate-kind.sh.
 #
+# Composes:
+#   1. hack/install-ate.sh --deploy-ate-system  (control plane)
+#   2. hack/install-microvm-deps.sh --install   (asset build/stage + cluster-wide
+#                                                microvm-default SandboxConfig)
+#   3. Apply the counter-microvm demo manifest (namespace + WorkerPool +
+#      ActorTemplate; resolves to microvm-default via sandboxClass=microvm).
+#
 # Like the other hack scripts, this sources .ate-dev-env.sh for the cluster /
 # registry / bucket settings unless NO_DEV_ENV is set.
 #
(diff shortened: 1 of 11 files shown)


hack/run-microvm-demo.sh
```diff
@@ -18,6 +18,13 @@
 # local kind cluster use hack/run-microvm-demo-kind.sh (which sets the kind env
 # and calls this script), mirroring install-ate.sh / install-ate-kind.sh.
 #
+# Composes:
+#   1. hack/install-ate.sh --deploy-ate-system  (control plane)
+#   2. hack/install-microvm-deps.sh --install   (asset build/stage + cluster-wide
+#                                                microvm-default SandboxConfig)
+#   3. Apply the counter-microvm demo manifest (namespace + WorkerPool +
+#      ActorTemplate; resolves to microvm-default via sandboxClass=microvm).
```
BenTheElder: 🤖 **nit** 🟢 – This describes the mechanism the PR deliberately rejected. The config is named `microvm`, not `microvm-default`, and `sandboxconfig-microvm.yaml.tmpl` says it "is deliberately NOT marked default:true" so a dirty teardown cannot silently bind new pools — binding is by explicit `sandboxConfigName`, which is what the demo manifest and `deploy.sh` both do. "Resolves to microvm-default via sandboxClass=microvm" reads as default-by-class resolution, so a reader tracing why a WorkerPool picks this config would draw the wrong conclusion.

The stale name also appears at line 24 here and in three `orchestrator.py` docstrings (lines 288, 294, 441).

BenTheElder (commented):
generally lgtm, I think, one confusing nit

hack/run-microvm-demo.sh
maxsmythe: Oops, thank you for spotting these. Removed/renamed.

BenTheElder:
lgtm, modulo rebase

maxsmythe:
@BenTheElder Thanks for review! Rebased.

BenTheElder (dismissed):

BenTheElder (approved):

Issue #592, opened 2026-07-29
Define mutability/validity/cacheability for API responses


bowei:
Our implementation currently has a workflows that put API server calls on the hot path. 

A  example is that the a Router request to an Actor  calls `ResumeActor`. This puts ATE API at the center of the operation which will be hard to scale. We should add some notion of mutability/cacheability/validity of state to the system so state can be cached. These are not documented in the .proto today.

In addition, we need to be more specific as the RPC response has fine grain validity within the message. Example:

```proto
// From ateapi.proto
message Actor {
 ...
 string actor_template_namespace; // <-- this field can be cached indefinitely as it will never change
 ...
 Status status = 4; // <-- this field would have an expiry of ??? ms?

 // Fields that will change, but have TTL of ??? ms
 string ateom_pod_namespace = 5;
 string ateom_pod_name = 6;
 string ateom_pod_ip = 7;
 ...
```

Proposal:

We should minimally document which fields are immutable and can be cached forever.

We should identify state that could/should be cached and what the strategy is for indicating validity. 

Some potential (bad) ideas: separate cacheable data into a nested message. Add some standard set of cache metadata (strategy, valid until, TTL etc). I think one requirement is that the server set these fields explicitly vs rely on implicit constants baked into the codebase. We may want to consider having this metadata on a per-response basis a la HTTP. The issue with using HTTP response caching semantics is that it is very broad and some pieces of state may have different caching requirements. As our API structures are somewhat designed to avoid round trips, we may not be able to split resources in a way that matches their ability to be cached.

PR #693, opened 2026-08-02
egress: add egress gateway support - atenet-egress

LiorLieberman:
Added pluggable egress PEP support.  (Feedback on atenet-egress name is welcomed - **will open a separate PR to rename atenet-router to atenet-ingress** )

#559 shipped the actor egress data path without any egress gateway. This adds an Envoy deployment that terminates actor CONNECTs. It requires downstream mTLS, so only a worker's atunnel can reach it.

The gateway consists of an Envoy and an `atenet router --standalone` ext_proc sidecar.

The same ext_proc binary now can serve both directions. It has a mode=ingress|egress|all. 

  *  --egress-gateway-address flag, This turns on pluggable actor egress cluster-wide. ateapi stamps this address onto every atelet Run/Restore, ateom hands it to atunnel, and actor TCP egress is transparently redirected into atunnel, which wraps it in mTLS + HTTP CONNECT to the egress gateway.
  
  **We should remove this when egress API pr is merged**

One more FYI

The egress listener's is pinned to the actor-identity CA root alone, so the trust anchor is a property of the filter chain rather than the request, any client certificate not signed by the actor-identity CA fails the TLS handshake, and only traffic with actor-identities can open a tunnel through the egress gateway.

- [x] Tests pass


LiorLieberman:
FYI @dberkov

manifests/ate-install/atenet-egress.yaml
```diff
@@ -0,0 +1,348 @@
...
+          mountPath: /run/servicedns.podcert.ate.dev
+          readOnly: true
+        - name: podidentity
+          mountPath: /run/podidentity.podcert.ate.dev
+          readOnly: true
+      # Co-located ext_proc server (the atenet router, ext_proc-only). The egress
+      # Envoy calls it over localhost to authenticate actor identity against the
+      # ate API on every CONNECT. This mirrors the ingress gateway topology
+      # (Envoy + ext_proc in one pod); a shared/standalone ext_proc is a future step.
+      - name: ext-proc
```
haiyanmeng, line 286: Let us make `ext-proc` a native sidecar (`initContainers` + `restartPolicy: Always`).
    Regular containers get SIGTERM together, `ext_proc` has no drain and will exit first,
    and with `failure_mode_allow: false` every CONNECT arriving during Envoy's drain
    window fails closed with a 503.

manifests/ate-install/atenet-egress.yaml
```diff
@@ -0,0 +1,348 @@
...
+        - name: servicedns
+          mountPath: /run/servicedns.podcert.ate.dev
+          readOnly: true
+        - name: podidentity
+          mountPath: /run/podidentity.podcert.ate.dev
+          readOnly: true
+      # Co-located ext_proc server (the atenet router, ext_proc-only). The egress
+      # Envoy calls it over localhost to authenticate actor identity against the
+      # ate API on every CONNECT. This mirrors the ingress gateway topology
+      # (Envoy + ext_proc in one pod); a shared/standalone ext_proc is a future step.
```
haiyanmeng: `a shared/standalone ext_proc is a future step`: can we first move the `exe_proc` logic into a separate binary?

bowei: ext_proc is already separate

haiyanmeng: I don't think so. See Line 268 in this file:

```
image: ko://github.com/agent-substrate/substrate/cmd/atenet
```

cmd/atelet/main.go
```diff
@@ -76,10 +76,39 @@ var (
 	localhostRegistryReplacement = pflag.String("localhost-registry-replacement", "", "The replacement registry endpoint for localhost and/or loopback IP addresses, useful for local development. for example kind-registry:5000")
 	imageCacheDir                = pflag.String("image-cache-dir", ateompath.ImageCacheDir, "Directory for the node-local OCI image layer cache. Must be on the volume shared with the ateom pods (the cached layers are their overlay lowerdirs), and on a disk sized for both capacity and IOPS: unpack throughput is gated by the volume's IOPS.")
 
+	// SEE(lior): both sides kept — main added --log-level here while this branch
```
bowei: What is this?

cmd/atelet/main.go
```diff
@@ -76,10 +76,39 @@ var (
...
+	// added --egress-gateway-address. Independent flags, no interaction.
```
bowei: probably want to scrub your comments

cmd/atelet/main.go
```diff
@@ -76,10 +76,39 @@ var (
...
+	//
+	// egressGatewayAddress turns on pluggable actor egress cluster-wide (POC).
+	// When set, actors whose Run/Restore request does not already carry an egress
+	// gateway address have this address injected, causing ateom to redirect actor
+	// TCP egress through atunnel to the egress gateway. Empty keeps egress off.
+	egressGatewayAddress = pflag.String("egress-gateway-address", "", "Address (host:port) of the egress gateway. When set, actor TCP egress is transparently tunneled through atunnel to this gateway. Empty disables egress.")
```
bowei: Your comment and the help text are different.

As I understand it -- this sets a default egress if not specified on the Actor resource explicitly?

Can you fix it to match up?

cmd/atenet/internal/router/extproc.go
```diff
@@ -96,7 +96,16 @@ func (s *ExtProcServer) Process(stream extprocv3.ExternalProcessor_ProcessServer
 		switch reqType := req.Request.(type) {
 		case *extprocv3.ProcessingRequest_RequestHeaders:
 			start := time.Now()
-			hResponse, rqm, target, tmplNs, tmplName, resumeOutcome, err := s.handleRequestHeaders(stream.Context(), reqType.RequestHeaders)
+			// One ext_proc server handles both directions: actor egress
```
bowei: Let's factor this so you could run it in this way, but in a real deployment, I think you would keep them as separate deployments as they will likely have different scaling parameters.

cmd/atenet/internal/router/extproc_egress.go
```diff
@@ -0,0 +1,185 @@
...
+const (
+	// EgressFilterChainName is the Envoy filter chain that terminates actor
+	// egress CONNECTs. It must stay in sync with the filter chain name in
+	// manifests/ate-install/atenet-egress.yaml.
+	EgressFilterChainName = "egress"
+	// FilterChainNameAttribute is the CEL attribute carrying the name of the
+	// filter chain that accepted the request. The egress Envoy asks for it via
+	// request_attributes on its ext_proc filter.
+	//
+	// SEE(lior): this was xds.listener_name, which reads more naturally but
```
bowei: you should scrub for these.

bowei:
implementation itself looks ok, but would like to fix a few organization things:

* Make it so that egress, ingress for ext_proc can be enabled separately (add flags and gated deployment). We currently deploy both in the same deployment, but at scale, you would probably want to have separate ingress and egress deployments as they will be allocated resources very differently.
* Put ingress,egress logic in .../atenet/router/{ingress,egress} packages.
* router/ext_proc just has mux logic

probably something like:

```
switch {
case EgressEnabled && IsEgressRequest(...):
case IngressEnabled:
default:
   // send 4xx nothing is enabled.
}
```

bowei (commented):
You need to scrub through all of the comments -- there seems to some internal notes to yourself and your friend Claude.

bowei:
Needs rebase

cmd/atelet/main.go
LiorLieberman: Removed.

cmd/atelet/main.go
LiorLieberman: removed POC language but we wont need that when we are properly implementing the egress api

cmd/atelet/main.go
LiorLieberman: not directly. We are using this flag because we have not yet designed the egress API properly and we havent wired it up yet. 

When we do, we dont need any egressgateway flag

LiorLieberman: BTW Eitan took some of it to his identity PR so if that one ends up in first, I will remove it from here and rebase

manifests/ate-install/atenet-egress.yaml
LiorLieberman: with what we have now - atenet == extproc server 
atenet-egress deployment == envoy + atenet (handling egress releated extproc).

atenet-ingress deployment == envoy + atenet(handling ingress releated extproc).

manifests/ate-install/atenet-egress.yaml
LiorLieberman: good point. created an issue for that - https://github.com/agent-substrate/substrate/issues/760

cmd/atenet/internal/router/extproc.go
LiorLieberman: Comment is confusing, they are currently separate deployments but egress and ingress is supported with the same atenet binary (different extproc handlers) in the same binary. I will reword the comment and will also make a mode to be able to launch the atenet in ingress only or egress only mode.

cmd/atenet/internal/router/extproc_egress.go
```diff
@@ -0,0 +1,176 @@
...
+	if actor.GetStatus() != ateapipb.Actor_STATUS_RUNNING {
+		return nil, metadata, "", "", "", ResumeOutcomeNone, newReqError(envoy_type.StatusCode_Forbidden,
+			"egress denied: actor %q/%q is %s, not running", atespace, actorName, actor.GetStatus())
+	}
+
+	// X-Ate-Actor-Version is the Actor version the worker observed when it was
+	// assigned, and atunnel documents it as a lower bound on trustworthy actor
+	// metadata. If our authoritative view is older than what the worker asserts,
+	// we cannot yet vouch for the identity, so reject rather than allow blindly.
+	if assertedVersion != "" {
```
EItanya: Will not be necessary once my PR goes in, this version is removed

EItanya (commented):
Offline we discussed cleaning up the locations in which the egressgateway address needs to be set, right now it's set on quite a few pods, but we can definitely simplify.

EItanya:
Here is the list of specific PEP todos as a result of #708 
```
  - Require mTLS client certificates signed by the actor-identity CA.
  - Verify validity period, ClientAuth EKU, and IsCA == false.
  - Require exactly one valid ActorIdentity extension.
  - Require non-empty atespace, actor name, and actor UID.
  - Require Purpose == "atunnel"; reject generic, missing, or unknown purposes.
  - Authorize using the verified actor UID—not CONNECT headers or actor-provided metadata.
```

manifests/ate-install/atenet-egress.yaml
shrutiyam-glitch: I've actually opened a PR to handle the graceful termination of the atenet-router (#774, which should help mitigate the Envoy drain window issue mentioned here.

manifests/ate-install/atenet-egress.yaml
```diff
@@ -0,0 +1,361 @@
...
+              codec_type: HTTP1
+              upgrade_configs:
+              - upgrade_type: CONNECT
+              # SEE(lior): this pair is how the actor's certificate reaches the
+              # ext_proc handler. ext_proc can request Envoy attributes, but none
+              # of them carry a custom X.509 extension, so the only way to check
+              # ActorIdentity in Go is to have Envoy hand over the raw chain.
+              # SANITIZE_SET drops whatever x-forwarded-client-cert the client
+              # sent and writes Envoy's own view of the verified peer, and
+              # chain: true puts the full URL-encoded PEM chain in it.
```
LiorLieberman: verify this is what we want

@botengyao @ahmedtd

bowei:
Probably closes https://github.com/agent-substrate/substrate/pull/338

manifests/ate-install/atenet-egress.yaml
bowei: You can also just have ext-proc register the SIGTERM but not quit (alternatively)

manifests/ate-install/atenet-egress.yaml
bowei: It would be ideal to have a cleaner solution ... and ideally not in the config itself, but we should not block on getting it working.

hack/verify-egress-demo.sh
```diff
@@ -0,0 +1,87 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Preconditions:
+#   hack/create-kind-cluster.sh
+#   hack/install-ate-kind.sh --deploy-demo-egress
+#
+# The egress demo Actor accepts {"url":"..."} and performs an HTTP GET. With
+# egress turned on (ate-api-server --egress-gateway-address, which ateapi stamps
```
bowei: probably detail is no needed and could change in the future, can be confusing for humans and agents.

demos/egress/test-egress.sh
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env bash
```
bowei, line 1: I didn't review this script in great detail. At some point, we may want to switch to writing this stuff in go, otherwise it's not going to be maintainable except by agents...

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -0,0 +1,418 @@
...
+// the gateway tunnels it out, using the actor certificate atunnel presented in
+// the mTLS handshake. Nothing the actor can write — no CONNECT header, no
+// request metadata — contributes to the identity; the only inputs are the
+// certificate the actor-identity CA signed and the control plane's own view of
+// that actor.
+//
+// Authorization by destination and credential/token injection are deliberately
+// a TODO once we have SessionIdentity RPC service figured out.
+//
+// Egress never resumes an actor — it requires one already RUNNING — and picks
```
bowei: Comment seems unnecessary

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -0,0 +1,418 @@
...
+// no upstream of its own, so the Result carries neither a resume outcome nor a
+// target.
+func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestMetadata) (extproc.Result, error) {
+	// Dispatch is by filter chain, so reaching here means the egress listener
```
bowei: Alternative suggest for easier to understand:

```
// Sanity check that we were called on the Egress listener filter chain with
// a CONNECT.
```

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -0,0 +1,418 @@
...
+		// The body stays generic on purpose: an actor that fails authentication
+		// has not proven it is anyone, so it gets no detail about why. The
+		// specific reason rides along as the wrapped cause, which only the
+		// server-side log below reads.
+		slog.WarnContext(ctx, "egress denied: actor certificate rejected", slog.Any("err", err))
+		return extproc.Result{}, extproc.WrapReqError(envoy_type.StatusCode_Forbidden, err,
+			"egress denied: invalid actor certificate")
+	}
+
+	atespace := identity.Atespace
```
bowei: suggest put these in helper fun:

func validateIdentity() error

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -0,0 +1,418 @@
...
+	// The CA only ever mints these from control-plane state, so a name that is
+	// not a legal resource name means the CA or its inputs are compromised.
+	if !resources.IsValidResourceName(atespace) || !resources.IsValidResourceName(actorName) {
+		return extproc.Result{}, extproc.NewReqError(envoy_type.StatusCode_Forbidden,
+			"egress denied: invalid actor identity %q/%q", atespace, actorName)
+	}
+
+	// Confirm the certified actor still exists. The name is only a lookup key
+	// here; the UID below is what actually authorizes.
+	actor, err := h.apiClient.GetActor(ctx, &ateapipb.GetActorRequest{
```
bowei, line 165: we should be careful here as this results in load to the ATEAPI server

file a TODO as a sub issue of the actor state caching issue

https://github.com/agent-substrate/substrate/issues/592

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -0,0 +1,418 @@
...
+		Actor: &ateapipb.ObjectRef{Atespace: atespace, Name: actorName},
+	})
+	if err != nil {
+		return extproc.Result{}, mapEgressIdentityError(atespace, actorName, err)
+	}
+
+	// Authorize on the UID, not the name. Names are reused: delete an actor and
```
bowei: If we can put this code in helper 

func validateActor

will be easier to reason about

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -0,0 +1,418 @@
...
+//
+// The chain is verified here even though Envoy already did it at the handshake
+// (require_client_certificate with the actor-identity CA as trusted_ca). We have
+// to parse the certificate anyway to read the ActorIdentity extension, which
+// Envoy cannot see, and trusting a parsed-but-unverified certificate is a
+// well-worn source of CVEs. It also keeps the handler safe if the Envoy config
+// is ever loosened, and costs one signature check per CONNECT rather than per
+// request. The IsCA, ClientAuth-EKU, and purpose checks below have no Envoy-side
+// equivalent at all.
+func (h *Handler) verifyActorCertificate(chain []*x509.Certificate) (*substratex509.ActorIdentity, error) {
```
bowei, line 219: put this and the functions it calls in its own package 

egress/cert

cert.VerifyActor(...)

If a func doesn't really need to be a method, you should make it a free func for easier testing.

bowei (commented):
Mostly looks ok.

hack/verify-egress-demo.sh
LiorLieberman: I believe you meant for the thing in the "()". Removed.

manifests/ate-install/atenet-egress.yaml
LiorLieberman: verified with Boteng. This is ~fine. Added a TODO. Dont think an issue is needed since "cleaner" is undefined yet.

demos/egress/test-egress.sh
LiorLieberman: https://github.com/agent-substrate/substrate/issues/842

cmd/atenet/internal/router/egress/egress.go
LiorLieberman: added a ref to this issue here

cmd/atenet/internal/router/egress/egress.go
LiorLieberman: this PR has grew very big (partly because I addressed a similar comment from you to split ingress + egress under router). Lets take this as a follow up

bowei: Ok

cmd/atenet/internal/router/extproc/dispatch.go
```diff
@@ -0,0 +1,83 @@
...
+	// DirectionEgress is outbound traffic tunneled out of an actor.
+	DirectionEgress Direction = "egress"
+)
+
+const (
+	// EgressFilterChainName is the Envoy filter chain that terminates actor
+	// egress CONNECTs, and so the one that selects the egress handler. It must
+	// stay in sync with the filter chain name in
+	// manifests/ate-install/atenet-egress.yaml.
+	EgressFilterChainName = "egress"
```
bowei, line 38: Take this as a follow-up: it should be named something more unique to give a clue that there is something special about the name. Ideally, `git grep <name>` should just yield this and the configuration location.

bowei (approved):

bowei:
There are a number of follow ups. Let's make sure we get them in as well.

BenTheElder:
@bowei if we don't have clear logical commits, let's squash merge?

PR #698, opened 2026-08-02
install: always re-apply CRDs and RBAC on --deploy-ate-system

### LINKED ISSUE #697: install-ate.sh: --deploy-ate-system never re-applies CRDs/RBAC on upgrade
## Summary

`deploy_ate_system` gates `manifests/ate-install/generated/` behind `ensure_crds`, which returns early when the three CRDs already exist. On any upgrade of an existing cluster that skips the whole directory — and since `role.yaml` lives there and no other deploy path applies it, both CRD schemas and every ClusterRole stay frozen at first-install state.

A controller image that needs a newly added permission then deadlocks at startup: its informer cache cannot list the resource, so reconciliation never runs, while the rollout itself reports success.

## Observed on a real cluster

kind cluster installed 4 days ago, then upgraded via `install-ate-kind.sh --deploy-ate-system`:

- ate-controller (new image) logged
  `failed to list *v1.NetworkPolicy: ... User "system:serviceaccount:ate-system:ate-controller" cannot list resource "networkpolicies"`
  and stopped reconciling entirely — a deleted WorkerPool Deployment was never recreated, with no events anywhere.
- After manually applying `manifests/ate-install/generated/role.yaml` and restarting the controller, reconciliation resumed immediately.
- With the fix, the same upgrade run printed `actortemplates.ate.dev configured` / `workerpools.ate.dev configured` — the CRD schemas had also silently drifted.

## Reproduction

1. Install: `./hack/install-ate-kind.sh --deploy-ate-system`
2. Advance the tree to any revision whose controller requires a new RBAC rule (or edit the live ClusterRole to drop one).
3. Upgrade with the same command: rollout reports success, `deploy_crds` never runs, controller deadlocks on informer start.

## Suggested fix

Call `deploy_crds` unconditionally from `deploy_ate_system` — `kubectl apply` is idempotent, so re-applying on every install/upgrade is free. `ensure_crds` remains correct for the demo scripts and per-component deploy flags, where the semantic really is "make sure they exist".

orangeCatDeveloper:
Fixes #697

## Summary

`deploy_ate_system` gated `manifests/ate-install/generated/` behind `ensure_crds`, whose existence check skips the directory on any upgrade — stranding CRD schemas and, since `role.yaml` has no other apply path, all ClusterRoles at first-install state. A controller image needing a new permission then deadlocks on informer start while the rollout reports success.

`deploy_ate_system` now calls `deploy_crds` unconditionally (`kubectl apply` is idempotent). The demo scripts and per-component deploy flags keep `ensure_crds`, where "make sure they exist" is the intended semantic.

## Test plan

- Reproduced on a 4-day-old kind cluster: upgrading the control plane deadlocked ate-controller on `cannot list networkpolicies`; manually applying `role.yaml` unblocked it.
- With this fix, the same `install-ate-kind.sh --deploy-ate-system` run prints the `deploy_crds` step and re-applied the drifted manifests (`actortemplates.ate.dev configured`, `workerpools.ate.dev configured`, ClusterRoles applied); cluster reconciles normally.
- `bash -n hack/install-ate.sh`.

- [x] Tests pass

### DIFF
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -327,7 +327,9 @@ deploy_crds() {
 
 deploy_ate_system() {
   log_step "deploy_ate_system"
-  ensure_crds
+  # Not ensure_crds: its existence check skips upgrades, stranding stale CRD
+  # schemas and RBAC (role.yaml has no other apply path).
+  deploy_crds
 
   # Enforce per-class SandboxConfig asset requirements (applied before any
   # SandboxConfig so the defaults below are validated too).


maxsmythe (approved):
LGTM, thank you for the fix!

PR #699, opened 2026-08-02
atenet: make xDS snapshot versions restart-unique

### LINKED ISSUE #617: [P3] XDS versionCount resets to 1 on router restart — config changes may not reach Envoy
**Severity:** P3 (edge case; only affects first restart after exactly one prior snapshot)
**Component:** Network Plane — `cmd/atenet/internal/router/xds.go`
**Audit ID:** NET-9

---

## Summary

The XDS snapshot version counter is an in-memory `int64` starting at 0. Each router
restart begins versioning at 1. If Envoy previously ACKed snapshot version `"1"` and
the router restarts pushing a new config as version `"1"` again (same string), the
go-control-plane ADS server may not re-push the config (depending on its same-version
handling). This creates a small window where Envoy's config reflects the pre-restart
state rather than the new state.

---

## Root Cause

**File:** `cmd/atenet/internal/router/xds.go` lines 99–100

```go
var versionCount int64 = 0  // reset to 0 on every process start

func nextVersion() string {
    return strconv.FormatInt(atomic.AddInt64(&versionCount, 1), 10)
    // First call after restart always returns "1"
}
```

The go-control-plane cache's behavior when a client reconnects with `version_info="1"`
and the server's current version is also `"1"` depends on the implementation:
if the cache treats it as "client already has this version," it may not push an update.

---

## Steps to Reproduce

```bash
# Initial state: router running, Envoy has ACKed version "1" (just started)
# Confirm via Envoy admin
curl http://localhost:9901/config_dump | jq '.configs[] | select(.["@type"] | contains("ScopedRoutesConfig")) | .version_info'
# "1"

# Change router config and restart
kubectl set env -n ate-system deployment/atenet-router SOME_PARAM=new_value
kubectl rollout restart -n ate-system deployment/atenet-router

# New router pushes version "1" again (reset counter)
# Envoy may not receive the updated config if it reconnects and sends version_info="1"
curl http://localhost:9901/config_dump | jq '...'
# May still show old config
```

---

## Suggested Fix

Initialize `versionCount` with a monotonically increasing base that survives restarts:

```go
// Use Unix timestamp as base to ensure monotonic increase across restarts
var versionCount int64 = time.Now().Unix()
```

Or use a UUID-based version string that is always different from any prior version,
eliminating the same-version collision entirely:

```go
func nextVersion() string {
    return fmt.Sprintf("%d-%d", startupTimestamp, atomic.AddInt64(&versionCount, 1))
}
```

orangeCatDeveloper:
Fixes #617

## Summary

The xDS snapshot version was an in-memory counter starting at 0, so every router restart re-minted "1", "2", … go-control-plane's snapshot cache compares version strings for equality and skips the push when Envoy already reports that version, so a restarted router whose counter lands on the string Envoy still holds leaves Envoy silently stranded on pre-restart config.

Versions are now `<unix-seconds>-<8-char random>-<counter>` — a per-process epoch no incarnation can repeat, even if the clock jumps backwards across a restart. version_info is opaque to Envoy and only ever compared for equality, so the format change is safe.

## Test plan

- `TestSnapshotVersionsUniqueAcrossRestarts`: two XdsServer instances (simulating a restart) each deploy snapshots through the real SnapshotCache; asserts no version string repeats within or across instances.
- Live verification on a kind cluster: deployed the router, confirmed Envoy ACKed the new format via `config_dump`, then SIGKILLed only the router container (Envoy kept running holding the old version) — after the restart Envoy ACKed a new-epoch version, proving the re-push reaches a reconnecting Envoy.

- [x] Tests pass

### DIFF
--- a/cmd/atenet/internal/router/xds.go
+++ b/cmd/atenet/internal/router/xds.go
@@ -16,6 +16,7 @@ package router
 
 import (
 	"context"
+	"crypto/rand"
 	"fmt"
 	"log/slog"
 	"net"
@@ -152,6 +153,7 @@ type XdsServer struct {
 	snapshot     cachev3.SnapshotCache
 	srv          serverv3.Server
 	versionCount int64
+	versionEpoch string
 
 	mu sync.Mutex
 
@@ -204,6 +206,7 @@ func NewXdsServer(xdsPort int) *XdsServer {
 		xdsPort:               xdsPort,
 		snapshot:              cache,
 		srv:                   srv,
+		versionEpoch:          strconv.FormatInt(time.Now().Unix(), 10) + "-" + rand.Text()[:8],
 		extprocPort:           50051, // matches default extproc port
 		extprocAddr:           "127.0.0.1",
 		ingressPort:           8080,
@@ -408,7 +411,7 @@ func (x *XdsServer) UpdateSnapshot() error {
 	defer x.mu.Unlock()
 
 	x.versionCount++
-	ver := strconv.FormatInt(x.versionCount, 10)
+	ver := x.versionEpoch + "-" + strconv.FormatInt(x.versionCount, 10)
 
 	// connectEnabled is true when either CONNECT listener (plaintext or TLS) is
 	// configured; the main_internal cluster/listener only exist to serve them.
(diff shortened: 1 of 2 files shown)


bowei:
Change LGTM, needs rebase

bowei (dismissed):

orangeCatDeveloper:
> Change LGTM, needs rebase

Rebased

orangeCatDeveloper:
gentle bump @bowei

bowei (approved):

PR #703, opened 2026-08-02
install: route kind component deploys through the overlays

### LINKED ISSUE #702: install-ate.sh: kind component deploys use GKE OTel endpoint
## Problem

On kind, `--deploy-atenet` and `--deploy-ate-apiserver` apply the base manifests instead of the kind Kustomize overlay. The base manifests set:

```text
OTEL_EXPORTER_OTLP_ENDPOINT=http://opentelemetry-collector.gke-managed-otel.svc.cluster.local:4317
```

That collector exists on GKE, not in the kind stack, so the redeployed component cannot export OTLP telemetry. `--deploy-atelet` is not affected by this endpoint bug because it already renders a component-specific kind overlay.

## Reproduction

```bash
hack/install-ate-kind.sh --deploy-ate-system
hack/install-ate-kind.sh --deploy-atenet
kubectl get deployment/atenet-router -n ate-system \
  -o jsonpath='{.spec.template.spec.containers[?(@.name=="atenet-router")].env[?(@.name=="OTEL_EXPORTER_OTLP_ENDPOINT")].value}'
```

The command returns the `gke-managed-otel` endpoint. The kind deployment should use:

```text
http://opentelemetry-collector.otel-system.svc:4317
```

`--deploy-ate-apiserver` has the same problem.

## Fix

Add component-specific kind overlays for atenet and ate-api-server, and use them when `ATE_INSTALL_KIND=true`. Keep the existing base-manifest path for GKE.

## Acceptance criteria

- Both partial kind deploys render the `otel-system` endpoint.
- The ate-api-server partial overlay retains the kind metric interval and timeout settings.
- Non-kind deploy behavior is unchanged.
- The metrics E2E observes router telemetry after the SDK export interval.

orangeCatDeveloper:
Fixes #702

## Summary

`--deploy-atenet` and `--deploy-ate-apiserver` applied base manifests directly, skipping the kind overlay's OTel env patches, so components deployed through those flags silently exported telemetry to the GKE-only collector address. They now route through per-component kind overlays the way `--deploy-atelet` already does, with the root kind overlay composing the three so each patch exists in exactly one place. `hack/verify/kind-overlays.sh` (picked up by `verify-all.sh`) guards the routing, the render equivalence, and the per-container env values.

## Test plan

- Pre/post-refactor full kind renders are byte-identical across all 57 resources.
- Six negative injections each fail the verifier (helper bypassed, container renamed, env values changed or deleted, duplicate patch re-added at root).
- Real kind cluster: `--deploy-atenet` lands the otel-system endpoint without manual intervention; the metrics e2e suite went from a 120s timeout to a sub-second pass.
- `bash -n`, `hack/verify/shellcheck.sh`, and `hack/verify/boilerplate.sh` pass.

- [x] Tests pass

### DIFF
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -396,8 +396,32 @@ deploy_ate_apiserver() {
 
   ensure_apiserver_prerequisites
 
-  run_ko apply -f manifests/ate-install/ate-api-server.yaml
+  component_manifests ate-api-server | run_ko resolve -f - | run_kubectl apply -f -
   run_kubectl rollout status deployment/ate-api-server -n ate-system --timeout=120s
+}
+
+# Kind overlay vs base manifests is decided ONLY here; hack/verify/kind-overlays.sh
+# sources this to assert kind deploys can't bypass the overlay and silently
+# revert its env patches (e.g. pointing OTel exports at the GKE-only collector).
+component_manifests() {
+  local component="$1"
+  if [[ "${ATE_INSTALL_KIND:-false}" == "true" ]]; then
+    kubectl kustomize "manifests/ate-install/kind/${component}" --load-restrictor LoadRestrictionsNone
+    return
+  fi
+  case "${component}" in
+    atenet)
+      cat manifests/ate-install/atenet-router.yaml
+      printf '\n---\n'
+      cat manifests/ate-install/atenet-dns.yaml
+      ;;
+    ate-api-server) cat manifests/ate-install/ate-api-server.yaml ;;
+    atelet) cat manifests/ate-install/atelet.yaml ;;
+    *)
+      echo "component_manifests: unknown component ${component}" >&2
+      return 1
+      ;;
+  esac
 }
 
 deploy_atelet() {
@@ -408,15 +432,7 @@ deploy_atelet() {
   run_kubectl apply -f manifests/ate-install/ate-system-namespace.yaml \
     && run_kubectl wait --for=jsonpath='{.status.phase}'=Active namespace/ate-system --timeout=60s
 
-  local manifest=""
-  if [[ "${ATE_INSTALL_KIND:-false}" == "true" ]]; then
-    # Use Kustomize to build and resolve the atelet DaemonSet patch
-    manifest=$(kubectl kustomize manifests/ate-install/kind/atelet --load-restrictor LoadRestrictionsNone | run_ko resolve -f -)
-  else
-    # Use base manifest for GKE
-    manifest=$(run_ko resolve -f manifests/ate-install/atelet.yaml)
-  fi
-  echo "${manifest}" | run_kubectl apply -f -
+  component_manifests atelet | run_ko resolve -f - | run_kubectl apply -f -
   run_kubectl rollout status daemonset/atelet -n ate-system --timeout=120s
 }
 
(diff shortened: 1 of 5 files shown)


hack/verify/kind-overlays.sh
```diff
@@ -0,0 +1,198 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Guards against partial-deploy drift: kind-mode component deploys must emit exactly
+# the component overlay render, component and full-overlay renders must match, and
+# each container must carry its OTel env — value-checked per container because a
+# renamed container turns a strategic-merge patch into a silent ghost sibling.
+
+set -o errexit -o nounset -o pipefail
```
maxsmythe, line 22: I'm not sure we want to use hack/verify as a unit test for the deploy script.

@BenTheElder any opinions?

hack/verify/kind-overlays.sh
```diff
@@ -0,0 +1,198 @@
...
+python3 - "${OUT}" <<'EOF'
+import re
+import sys
+
+out = sys.argv[1]
+ENDPOINT = "http://opentelemetry-collector.otel-system.svc:4317"
+INTERVAL = ("OTEL_METRIC_EXPORT_INTERVAL", "10000")
+TIMEOUT = ("OTEL_METRIC_EXPORT_TIMEOUT", "10000")
+
+def docs(path):
```
maxsmythe, line 95: This looks like it is parsing the overlays and making sure all resources of a specific kind have the expected overlays, is that correct?

I like the idea of it, but why not parse the YAML directly? Maybe include this as a standalone Python script and avoid the bash validation?

Maintaining overlays seems like a more durable validation problem.

hack/install-ate.sh
```diff
@@ -428,11 +444,15 @@ deploy_atenet() {
   run_kubectl apply -f manifests/ate-install/ate-system-namespace.yaml \
     && run_kubectl wait --for=jsonpath='{.status.phase}'=Active namespace/ate-system --timeout=60s
 
-  local router_manifest=""
-  router_manifest="$(render_atenet_router_manifest)"
```
maxsmythe, line 432: should render_atenet_router_manifest be made to understand the kind switch? It looks like other similar functions do:

https://github.com/agent-substrate/substrate/blob/55f16fa9dec78e5218c95f4e89f438cee7989595/hack/install-ate.sh#L161-L198

maxsmythe (changes requested):
Thanks for the fixes! Had a few comments

hack/verify/kind-overlays.sh
orangeCatDeveloper: Good idea. Python would need PyYAML though, which the repo doesn't declare — it'd pass only where the dev happens to have it installed. Would Go under tools/, like validate-image-cache, work instead? sigs.k8s.io/yaml is already vendored, so no new dependency.

maxsmythe: I think that is a fine place for it.

orangeCatDeveloper:
Closing this. #746 centralized the OTLP endpoint into a ConfigMap, which removes the endpoint from the workload manifests entirely — the symptom this PR was chasing (kind component deploys reverting to the GKE collector) is gone, and #746 already fixed the targeted redeploys for it.

PR #704, opened 2026-08-02
feat(atenet): replicate router dataplane

### LINKED ISSUE #645: [P1] atenet-router and dns run as single replicas with no PDB — every restart is a guaranteed full data-plane outage
> Dedup note: verified against all open+closed issues 2026-07-30 — no issue
> or PR files the data-plane SPOF. #181 ("Define substrate HA story") is
> scoped entirely to ate-apiserver + its database and never mentions the
> router or dns. PR #538 already did HA for ate-api-server (2 replicas +
> round-robin), so a precedent exists. #617/#612/#614 are related but
> distinct (see below).

## Summary

All actor traffic flows through one `atenet-router` pod and all actor name
resolution through one `dns` pod:

- `manifests/ate-install/atenet-router.yaml:110` — `replicas: 1`, no PDB,
  no topologySpreadConstraints.
- `manifests/ate-install/atenet-dns.yaml:82` — `replicas: 1`, same.
- The router-managed Envoy Deployment is additionally hardcoded to one
  replica in code: `cmd/atenet/internal/router/envoyrunner.go:140`
  (`replicas := int32(1)`) — so this is a code change, not just a manifest
  bump.
- The only PDB in the repo belongs to ate-api-server
  (`manifests/ate-install/ate-api-server.yaml:220`).

Any voluntary or involuntary disruption — deploy, node upgrade, eviction,
OOM — takes the entire data plane down for its duration.

## Evidence (live, GKE substrate-poc, 2026-07-30)

Measured blast radius with an in-cluster probe (5 req/s against a RUNNING
actor) while force-deleting the router pod:

- **~5.5 s of hard connection failures** (curl code `000` — *not* graceful
  503s) in the best case: warm image, instant reschedule.
- During the 09:53Z GKE node-pool event the same failure mode produced
  **6,880 `dial tcp` errors and 80 EOFs** in the overnight soak — the full
  pod-reschedule + image-pull window.

## Impact

- Guaranteed user-visible outage on every router deploy or node drain;
  failures surface as raw connection errors, which HTTP clients cannot
  safely retry for non-idempotent requests (POST).
- Defeats the "retries heal everything" design story: there is nothing to
  retry against while the single router is down.

## Related but distinct (do not fold into this issue)

- #617 — xDS `versionCount` resets on router restart (a *config-staleness*
  bug triggered by restart; also a hint that multi-replica xDS needs design
  attention, since each replica would keep an independent version counter).
- #612 — atenet controller reconcile short-circuit prevents Envoy
  Deployment self-heal.
- #614 — lists "component crash + recovery (… atenet-router)" as an
  untested scenario; test-coverage gap, not the SPOF itself.
- #181 — HA umbrella; body covers only ate-apiserver and its store.

## Suggested fix

1. Follow the PR #538 pattern (which HA'd ate-api-server): ≥2 router
   replicas behind the Service, PDB `maxUnavailable: 1`, topology spread
   across nodes; same for dns. Make `envoyrunner.go`'s replica count
   configurable.
2. Resolve the multi-replica xDS question first (#617's version counter
   per-replica); if multi-replica is not yet safe, document the SPOF in the
   operations runbook (#581) as an interim known limitation.

orangeCatDeveloper:
Fixes #645

`atenet-router` is a single-replica data-plane SPOF: a deploy, node drain, or eviction takes actor traffic down. A router pod kill produced about 5.5 seconds of hard connection failures, and one GKE node-pool event produced 6,880 dial errors.

The Deployment now runs two replicas with surge-only rollouts (`maxUnavailable: 0`), soft hostname spreading, and a PodDisruptionBudget. An Envoy readiness probe on admin `/ready` prevents a new replica from receiving traffic before its initial xDS load. The drain verifier pins its request and probes to the pod it deletes so a second replica cannot mask the shutdown path.

The multi-replica xDS concern is already resolved by the restart- and replica-unique version epochs in #699. Each Envoy talks only to the router in the same pod, so replicas do not share an xDS stream.

The DNS controller and its manifest were removed from `main` while this PR was pending, so the rebased change deliberately drops the obsolete DNS replication, PDB, and preStop work.

## Evidence

On a kind cluster, with a 5 req/s in-cluster probe against a running actor:

- rolling restart under traffic: 278/278 requests OK (single-replica baseline: about 5.5 seconds of hard failures)
- force-delete one router replica under traffic: 183/183 requests OK
- strict kustomize rendering passes after the rebase

### DIFF
--- a/hack/verify-atenet-drain.sh
+++ b/hack/verify-atenet-drain.sh
@@ -128,14 +128,15 @@ done
 
 # --- Observers ---------------------------------------------------------------
 
-POD="$(run_kubectl get pods -n "${ROUTER_NS}" -l app=atenet-router --no-headers | awk '$3=="Running"{print $1}')"
-# wc -w pads its output on macOS; compare numerically.
-(( $(wc -w <<<"${POD}") == 1 )) || fail "expected exactly 1 running router pod, got: ${POD:-none}"
+# Pin every observer to one pod: with several replicas the Service would spread
+# the parked request and the probes across pods that are not being deleted.
+POD="$(run_kubectl get pods -n "${ROUTER_NS}" -l app=atenet-router --no-headers | awk '$3=="Running"{print $1; exit}')"
+[[ -n "${POD}" ]] || fail "no running router pod"
 log_step "router pod under test: ${POD}"
 
 run_kubectl logs -n "${ROUTER_NS}" "${POD}" -c atenet-router -f > "${LOG_FILE}" 2>&1 &
 BG_PIDS+=($!)
-run_kubectl port-forward -n "${ROUTER_NS}" svc/atenet-router "${LOCAL_HTTP_PORT}:80" >/dev/null 2>&1 &
+run_kubectl port-forward -n "${ROUTER_NS}" "${POD}" "${LOCAL_HTTP_PORT}:8080" >/dev/null 2>&1 &
 BG_PIDS+=($!)
 run_kubectl port-forward -n "${ROUTER_NS}" "${POD}" "${LOCAL_METRICS_PORT}:9090" >/dev/null 2>&1 &
 BG_PIDS+=($!)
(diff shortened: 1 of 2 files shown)


maxsmythe:
This looks reasonable, but sending to @bowei as atenet expert

PR #705, opened 2026-08-03
atelet: prune local pause snapshots

### LINKED ISSUE #668: Pause snapshots are not cleaned up
## Expected Behavior
When pausing ,we should only keep the latest snapshot per actor on local disk.

When suspending, we should delete the local snapshot.

## Actual Behavior
We don't clean up local snapshots at all, so disk usage will fill up.

1. Pause and resume an actor multiple times and inspect the local directories

orangeCatDeveloper:
## Summary

atelet never deletes local pause snapshots, so they accumulate until the node disk fills (#668). The control plane only ever references the latest one (`LocalSnapshotInfo` is single-valued), so older directories are dead weight by design. Checkpoint now prunes every existing snapshot before writing the new one — each is superseded by the snapshot about to be written, and pruning first caps disk usage at a single snapshot. Pruning is best-effort: a failed delete is logged and retried by the next prune, never failing the checkpoint.

Snapshot prefixes are now validated as single path segments at the Checkpoint/Restore RPC boundary (shared `ValidateLocalSnapshotPrefix`). Previously only checked non-empty: a nested prefix like `pause/2` would silently write nested directories, and `..` could escape the actor's directory entirely.

Fixes #668

## Test plan

- Unit: prune remove-all/missing-dir cases; validator table (nested, traversal, absolute, backslash); Checkpoint + Restore request rejection cases
- Live on kind: demo e2e suites pass (snapshot + durable-dir lifecycle: pause→resume→suspend→resume across pause-scope configs); atelet logs confirm each checkpoint prunes the prior snapshot before the write

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -446,6 +446,12 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	sandboxRec.ActorTemplateNamespace = req.GetActorTemplateNamespace()
 	sandboxRec.ActorTemplateName = req.GetActorTemplateName()
 
+	// No earlier pause snapshot can ever be restored again, so remove them
+	// all: the actor's current state was just captured by CheckpointWorkload,
+	// and the control plane tracks only a single local snapshot, which this
+	// checkpoint either overwrites (pause) or clears (suspend).
+	pruneLocalCheckpoints(ctx, actorUID)
+
 	switch req.GetType() {
 	case ateletpb.CheckpointType_CHECKPOINT_TYPE_EXTERNAL:
 		// TODO(#362): Because we do not cache the external snapshot files when upload fails, we have to mark the Actor as CRASHED.
@@ -1103,8 +1109,8 @@ func validateCheckpointRequest(req *ateletpb.CheckpointRequest) error {
 			return err
 		}
 	case ateletpb.CheckpointType_CHECKPOINT_TYPE_LOCAL:
-		if req.GetLocalConfig().GetSnapshotPrefix() == "" {
-			return fmt.Errorf("snapshot prefix must be non-empty for type %s", req.GetType().String())
+		if err := resources.ValidateLocalSnapshotPrefix(req.GetLocalConfig().GetSnapshotPrefix()); err != nil {
+			return err
 		}
 	default:
 		return fmt.Errorf("invalid checkpoint type: %v", req.GetType())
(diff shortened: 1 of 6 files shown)


cmd/atelet/main.go
```diff
@@ -399,10 +399,14 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 		if err := s.uploadExternalCheckpoint(ctx, req, checkpointDir, sandboxRec); err != nil {
 			return nil, ateerrors.NewGRPCError(ctx, codes.DataLoss, ateerrors.ReasonFaileSaveSnapshot, ateerrors.ActorCrashedMetadata(), fmt.Errorf("%w: while uploading external snapshot: %w", ateerrors.ReasonFaileSaveSnapshot, err))
 		}
+		// The durable snapshot supersedes local pause snapshots.
```
dberkov: There are 2 possible flows:
1. run->pause [1]->resume->pause [2]
2. run->pause[1] -> resume -> suspend

In both cases,  all old snapshots can be deleted, regardless if it is just pause[1] or any other old snapshot.


Current implementation of the "func (s *AteomHerder) copyLocalCheckpoint(ctx context.Context, snapshotPrefix string, srcDir, dstDir string, files []string) error {
" method, copied previous snapshot to new location that takes time. I did in one of my PoC following code:

```
func (s *AteomHerder) copyLocalCheckpoint(ctx context.Context, snapshotPrefix string, srcDir, dstDir string, files []string) error {
	for _, fileName := range files {
		if ctx.Err() != nil {
			return fmt.Errorf("context cancelled: %w", ctx.Err())
		}
		src := filepath.Join(srcDir, snapshotPrefix, fileName)
		dst := filepath.Join(dstDir, fileName)
		// Hardlink first — constant-time regardless of file size, safe here
		// because runsc restore only reads these files. Fall back to a byte
		// copy on any error (cross-device, permissions, etc.).
		if err := os.Link(src, dst); err == nil {
			span.SetAttributes(attribute.String("method", "link"))
			span.End()
			continue
		}
		n, err := copyFile(src, dst)
		if err != nil {
			return fmt.Errorf("failed to copy %s to %s: %w", src, dst, err)
		}
	}

	return nil
}
```

it significantly improved the performance. The point if, you will do the "os.link" optimization, the `pruneLocalCheckpoints(ctx, actorUID, "")` logic need to be executed before new snapshot been copied either to external storage or moved to local storage location.

It means the logic can be moved above the "switch req.GetType() {" (line 396)

orangeCatDeveloper: Thanks for the detailed suggestion — that simplified this nicely. Dropped the keep parameter entirely and moved the prune above the switch, right after CheckpointWorkload returns.

cmd/atelet/main.go
```diff
@@ -445,6 +445,7 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	sandboxRec.ActorUID = req.GetActorUid()
 	sandboxRec.ActorTemplateNamespace = req.GetActorTemplateNamespace()
 	sandboxRec.ActorTemplateName = req.GetActorTemplateName()
+	pruneLocalCheckpoints(ctx, actorUID)
```
dberkov, line 453: nit: could you please add a comment here explaining why all previously taken snapshots from PAUSE can be cleaned at this time.

dberkov (commented):
minor comment, otherwise LGTM

cmd/atelet/main.go
orangeCatDeveloper: Comment added

dberkov (approved):

PR #708, opened 2026-08-03
atunnel: broker actor certificates through atelet

### LINKED ISSUE #706: Broker actor JWTs from atelet to atunnel over an authenticated Unix socket
## Context

Atunnel carries actor traffic but must not trust identity asserted by the actor. Today, egress identifies the actor using unauthenticated `X-Ate-*` CONNECT headers while atunnel authenticates with the worker Pod certificate.

The egress policy enforcement point (PEP) needs two independently verifiable facts:

1. The connection originates from a trusted atunnel.
2. That atunnel is currently acting for a specific actor.

Use the atunnel's existing Pod certificate for the first identity and a short-lived actor JWT for the second.

This issue assumes `podcertcontroller` and the required Kubernetes certificate features are enabled.

### Atelet credential broker

Add a small internal gRPC service served by atelet over a Unix domain socket.

Atunnel requests an actor JWT and supplies only the intended audience. It does not supply actor identity.

The connection uses mTLS:

- atunnel presents its worker Pod certificate;
- atelet presents its atelet Pod certificate;
- both verify the corresponding Pod identity trust bundle and require the peer's signed Pod identity to be on their own node;
- atelet extracts the worker Pod UID from the authenticated certificate.

This prevents another worker on the same node from impersonating atunnel, even though workers share the hostPath containing the socket.

### Resolve actor identity

For each request, atelet:

1. Authenticates the worker Pod UID from the mTLS connection.
2. Calls `ListWorkers` and finds the current assignment for that exact Pod UID.
3. Calls `ActorIdentity.MintJWT` with the authoritative actor identity, worker Pod UID, and requested audience.

`MintJWT` must independently revalidate that:

- the caller is atelet;
- the worker Pod UID exists;
- the worker is on the calling atelet's node;
- the worker is still assigned to the requested actor;
- the actor UID matches;
- the requested audience matches the egress PEP audience configured in ateapi.

The second authorization check closes assignment changes between `ListWorkers` and `MintJWT`.

A future dedicated assignment lookup may replace `ListWorkers` if its cost becomes material.

### JWT contents and future PEP validation

The JWT contains:

- actor atespace, name, and UID;
- actor resource version;
- the selected PEP audience;
- the worker Pod UID;
- normal issuer and validity claims.

The worker Pod UID binds the actor JWT to the atunnel mTLS identity.

For each CONNECT request, atunnel sends only `Authorization: Bearer <actor JWT>`. Remove the `X-Ate-*` CONNECT headers and their corresponding metadata fields.

PEP-side verification is a follow-up. Its contract is to:

1. Authenticate the atunnel Pod certificate.
2. Verify the actor JWT signature, issuer, audience, and validity.
3. Extract the worker Pod UID from both credentials and require an exact match.
4. Use the verified JWT for actor authorization.

### Multiple PEPs

Carry an `egress_gateway_audience` alongside `egress_gateway_address`.

The audience is a logical PEP identifier rather than a mutable network address. The control plane selects it and passes it through atelet to ateom. Today there is one PEP audience configured in ateapi and all other requested audiences are rejected; the wire contract supports multiple PEPs later.

### Lifecycle

- Atunnel obtains the first JWT before activating tunneled egress.
- Initial issuance failure fails actor activation closed.
- Renew when approximately 10% of the JWT lifetime remains.
- Retry renewal while the current JWT remains valid.
- After expiry, reject new egress connections.
- Existing established tunnels continue until closed.
- Deactivation cancels renewal and drops the JWT from memory.
- Atelet remains stateless; reconnection re-derives assignment from ateapi.

## API changes

- Add the internal atelet UDS gRPC broker service.
- Make `ActorIdentity.MintJWT` atelet-only.
- Add exact worker Pod UID authorization to `MintJWT`.
- Add the actor resource version and worker Pod UID binding to actor JWT claims.
- Add `egress_gateway_audience` to the activation path.
- Remove the `X-Ate-*` headers and their metadata fields.
- Send the issued JWT as the CONNECT bearer token.

## Security properties

- Actor workloads never receive the atunnel Pod certificate, actor JWT, or atelet credentials.
- Actor-supplied identity is never trusted.
- A compromised sibling worker cannot mint for another worker.
- A stale or migrated assignment cannot mint or renew.
- The worker Pod UID binding enables the PEP to reject an actor JWT replayed through another atunnel once PEP verification is implemented.
- Audience restriction prevents use against another PEP.

## Scope

- `ateom-gvisor`
- `ateom-microvm`
- Atunnel egress authentication
- Renewal and fail-closed behavior
- Documenting the PEP verification contract

## Out of scope

- PEP-side JWT verification and authorization
- Actor certificates
- Credentials mounted into actor workloads
- Pod-Certificate-free/token-only deployments
- Replacing `ListWorkers` with a dedicated lookup RPC
- Using actor identity for ingress routing

## Acceptance criteria

- Atunnel and atelet mutually authenticate over the Unix socket and reject peers from another node.
- Atelet derives actor identity from the authenticated worker assignment.
- Minting fails for a sibling, stale, unassigned, suspended, or migrated worker.
- Minting rejects any audience other than the configured egress PEP audience.
- The JWT contains actor identity, actor resource version, worker Pod UID, audience, and validity claims.
- CONNECT sends the actor JWT and no `X-Ate-*` identity headers.
- Initial issuance failure prevents egress activation.
- Renewal replaces the JWT before expiry.
- New egress fails after an unrecoverable expiry while established tunnels remain active.
- Both gVisor and microVM runtimes pass unit and end-to-end tests.

EItanya:
Closes #706

## Summary
- broker short-lived actor certificates from atelet over a same-node mTLS Unix socket
- keep the actor private key in atunnel and renew the certificate before expiry
- authenticate egress CONNECT using the actor certificate instead of bearer tokens

## Testing
- `make verify`
- `go test -race ./internal/atunnel`


LiorLieberman:
Thanks @EItanya ! 

Quick question before I review more deeply today. 

With the goal of "having a JWT available to atunnel to append on a CONNECT request to the egress" -- What parts of the this PR addresses?

Here are the parts we needed (before this pr -- havent reviewed it yet); 

1. Improve the JWT issuance flow as part of MintJWT (properly authorize atelets, having the JWT formats solid, etc)
2. Implement an in-memory cache with go background routine that refreshes certs for every active actor.
3. Implement the UDS socket between atunnel and atelet to pass the certs+jwts.

From a very brief glance - I think your PR is attempting to do all three. Is that correct? 

Also, For (1), we likely need to sort out two things to make the JWTs useful:

* Get the issuer a publicly accessible DNS name that is also oidc-compliant (cc: @ahmedtd) 
* Agree on the format for the Subject field on JWTs 

See for ref - https://github.com/agent-substrate/substrate/blob/3ed6aa07e1676e178ed4f31d23d5484e9fc5e0e0/cmd/ateapi/internal/actoridentity/actoridentity.go#L135-L138

EItanya:
Yes, with two clarifications: this PR renews JWTs, not actor certificates, and each worker has only one active actor.

1. **JWT issuance/authorization:** implemented. Atunnel authenticates to atelet over mTLS using the worker Pod identity. Atelet derives the worker UID from that certificate and resolves its current actor assignment; the request does not supply actor identity. Ateapi then independently revalidates the atelet’s node, worker assignment, actor UID, and permitted audience before minting. The JWT includes the actor UID/resource version and worker Pod UID.

2. **In-memory renewal:** implemented for the worker’s single active actor. Atunnel mints before activation, stores the JWT in memory, renews with roughly 10% lifetime remaining, retries until expiry, and fails closed for new tunnels after expiry. Deactivation clears the JWT and stops renewal.

3. **UDS broker:** implemented. The UDS carries only JWT requests/responses—not certificates. The Pod certificate is used to mutually authenticate the atunnel↔atelet connection, including verifying both are on the same node.

I agree the issuer and subject contracts remain unresolved. This PR intentionally retains the existing issuer/subject behavior and does not implement PEP verification yet. Before verification lands, we need a stable OIDC issuer with discovery/JWKS reachable by the PEP.

cmd/ateapi/internal/actoridentity/actoridentity.go
```diff
@@ -125,27 +113,31 @@ func (s *Server) MintJWT(ctx context.Context, req *ateapipb.MintJWTRequest) (*at
...
 
 		Substrate: actoridjwt.SubstrateClaims{
-			Atespace:  req.GetAtespace(),
-			ActorName: req.GetActorName(),
-			ActorUid:  req.GetActorUid(),
+			Atespace:             req.GetAtespace(),
+			ActorName:            req.GetActorName(),
+			ActorUid:             req.GetActorUid(),
+			ActorResourceVersion: actor.GetMetadata().GetVersion(),
+			WorkerPodUid:         req.GetWorkerPodUid(),
```
LiorLieberman: why do we need these?

cmd/ateapi/main.go
```diff
@@ -71,9 +71,10 @@ var (
 	redisTLSServerName  = pflag.String("redis-tls-server-name", "", "The ServerName to use for Redis TLS hostname verification.")
 	redisClientCert     = pflag.String("redis-client-cert", "", "The file containing client TLS certificate/key credential bundle for Redis/Valkey.")
 
-	clientJWTIssuer    = pflag.String("client-jwt-issuer", "", "The expected issuer URL for client JWTs.")
-	clientJWTAudience  = pflag.String("client-jwt-audience", "", "The expected audience for client JWTs.")
-	actorIDJWTPoolFile = pflag.String("actor-id-jwt-pool", "", "The file that contains the serialized JWT authority pool for signing actor JWTs")
+	clientJWTIssuer      = pflag.String("client-jwt-issuer", "", "The expected issuer URL for client JWTs.")
+	clientJWTAudience    = pflag.String("client-jwt-audience", "", "The expected audience for client JWTs.")
+	actorIDJWTPoolFile   = pflag.String("actor-id-jwt-pool", "", "The file that contains the serialized JWT authority pool for signing actor JWTs")
+	egressGatewayAddress = pflag.String("egress-gateway-address", "", "Address of the egress PEP. Empty disables tunneled egress.")
```
LiorLieberman, line 76: why do need egressgateway on ateapi?

cmd/atelet/credentialbroker.go
```diff
@@ -0,0 +1,121 @@
...
+	"google.golang.org/grpc/peer"
+	"google.golang.org/grpc/status"
+)
+
+type credentialBroker struct {
+	ateletpb.UnimplementedCredentialBrokerServer
+	// control resolves the authenticated worker Pod to its current assignment.
+	control ateapipb.ControlClient
+	// identity revalidates that assignment and signs the actor certificate.
+	identity ateapipb.ActorIdentityClient
```
LiorLieberman: identity as "identityService"? feels weird to call a field identity that does not represent an identity

cmd/atelet/credentialbroker.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (b *credentialBroker) MintActorCertificate(ctx context.Context, req *ateletpb.MintActorCertificateRequest) (*ateletpb.MintActorCertificateResponse, error) {
+	// TODO: Before release, request an atunnel-specific MintCert purpose and
+	// require the egress PEP to reject generic actor certificates.
+	// The request deliberately carries no actor identity. The authenticated Pod
+	// UID is the only input used to select a worker and its current assignment.
+	workerUID, err := authenticatedWorkerUID(ctx)
```
LiorLieberman: not sure I am following the TODO comment. I thought we were saying that we need it to be an actor cert thats specifically for atunnel use.

cmd/atelet/credentialbroker.go
```diff
@@ -0,0 +1,121 @@
...
+		return "", status.Error(codes.Unauthenticated, "missing peer certificate")
+	}
+	identity, err := substratex509.PodIdentityFromCertificate(tlsInfo.State.PeerCertificates[0])
+	if err != nil || identity == nil {
+		return "", status.Error(codes.PermissionDenied, "invalid worker identity")
+	}
+	return identity.PodUID, nil
+}
+
+func restrictClientToNode(node *substratex509.PodIdentity) func(tls.ConnectionState) error {
```
LiorLieberman: can we get a more descriptive name and/or a comment that tells what you are doing in this func?

cmd/atelet/main.go
```diff
@@ -71,6 +74,9 @@ var (
 
 	grpcServerCredBundle = pflag.String("grpc-server-cred-bundle", "/run/podidentity.podcert.ate.dev/credential-bundle.pem", "Credential bundle atelet presents as its gRPC serving certificate.")
 	clientCACerts        = pflag.String("client-ca-certs", "/run/podidentity.podcert.ate.dev/trust-bundle.pem", "CA bundle used to verify gRPC client certificates.")
+	ateapiAddress        = pflag.String("ateapi-address", "dns:///api.ate-system.svc:443", "ateapi gRPC target used by the credential broker.")
+	ateapiCAFile         = pflag.String("ateapi-ca-file", "/run/servicedns.podcert.ate.dev/trust-bundle.pem", "CA bundle used to verify ateapi.")
+	ateapiServerName     = pflag.String("ateapi-server-name", "api.ate-system.svc", "DNS name expected on the ateapi certificate.")
```
LiorLieberman, line 81: without this change atelet did not know whats ateapi address?

cmd/ateom-gvisor/main.go
```diff
@@ -56,11 +58,11 @@ var (
 
 	// TODO(liorlieberman) have a sub package for all atunnel releated things like that
 	atunnelListenAddress       = pflag.String("atunnel-listen-address", "0.0.0.0:443", "Address for actor ingress HTTPS")
-	atunnelCredentialBundle    = pflag.String("atunnel-credential-bundle", "/run/podidentity.podcert.ate.dev/credential-bundle.pem", "PEM credential bundle for actor ingress HTTPS")
-	atunnelTrustBundle         = pflag.String("atunnel-trust-bundle", "/run/podidentity.podcert.ate.dev/trust-bundle.pem", "PEM trust bundle for actor ingress clients")
+	workerCredentialBundle     = pflag.String("atunnel-credential-bundle", "/run/podidentity.podcert.ate.dev/credential-bundle.pem", "Worker Pod credential bundle used by atunnel for inbound serving and outbound mTLS")
```
LiorLieberman, line 60: While we are using the pod identity right now for atunnel, maybe we would want a seprate atunnel identity - not sure but this was discussed at some point. 

whats the reason to change the var names?

cmd/ateom-gvisor/main.go
```diff
@@ -160,12 +162,12 @@ func do(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("while parsing atunnel upstream: %w", err)
 	}
-	atunnelServer, atunnelEgress, atunnelEgressPort, err := runAtunnel(ctx, upstream)
+	atunnelIngress, atunnelEgress, egressProxyPort, err := runAtunnel(ctx, upstream)
```
LiorLieberman: its still a port on atunnel, why calling it egressProxy?

atunnelIngress LGTM though

cmd/ateom-gvisor/main.go
```diff
@@ -275,15 +284,30 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
...
+	}
+	var containersToDelete []string
 	defer func() {
 		if retErr != nil {
+			cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+			defer cancel()
+			if err := s.deactivateActorNetworking(cleanupCtx); err != nil {
+				slog.WarnContext(cleanupCtx, "Failed to deactivate actor networking after Run failure", slog.Any("err", err))
+			}
+			deleteContainers(cleanupCtx, rcmd, containersToDelete, "Run")
```
LiorLieberman, line 309: confused by these changes - is it just a re-order?

cmd/ateom-gvisor/main.go
```diff
@@ -341,7 +362,7 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
 	if err := readyz.WaitAll(ctx, req.GetSpec().GetContainers(), ateomnet.ActorVethIP); err != nil {
 		return nil, fmt.Errorf("while waiting for container readyz: %w", err)
 	}
-	if err := s.activateActorNetworking(req.GetAtespace(), req.GetActorName(), req.GetActorVersion(), req.GetEgressGatewayAddress()); err != nil {
+	if err := s.activateActorNetworking(req.GetAtespace(), req.GetActorName(), egress); err != nil {
```
LiorLieberman, line 364: why removed actorversion?

cmd/ateom-gvisor/main.go
```diff
@@ -496,31 +517,41 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+	}
+	var containersToDelete []string
 	defer func() {
 		if retErr != nil {
+			cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+			defer cancel()
+			if err := s.deactivateActorNetworking(cleanupCtx); err != nil {
+				slog.WarnContext(cleanupCtx, "Failed to deactivate actor networking after Restore failure", slog.Any("err", err))
+			}
+			deleteContainers(cleanupCtx, rcmd, containersToDelete, "Restore")
```
LiorLieberman, line 542: same comment as above

cmd/ateom-gvisor/main.go
```diff
@@ -595,49 +630,77 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
...
+		SocketPath:           ateompath.CredentialBrokerSocket,
+		CredentialBundlePath: s.workerCredentialBundlePath,
+		TrustBundlePath:      s.podIdentityTrustBundlePath,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("while configuring actor certificate broker: %w", err)
+	}
+	// Mint before starting the workload so configured tunneled egress fails
+	// closed. The source retains the private key for mTLS and renewal.
+	expiresAt, err := certificateSource.Mint(ctx)
```
LiorLieberman, line 662: the actual changes we need are buried within many other esthetic changes.

Can you explain exactly what we are doing - calling mint on every activation?

cmd/ateom-microvm/restore.go
```diff
@@ -70,8 +70,7 @@ func (s *AteomService) RestoreWorkload(ctx context.Context, req *ateompb.Restore
 		containers:   req.GetSpec().GetContainers(),
 		assetPaths:   req.GetRuntimeAssetPaths(),
 
-		actorVersion:         req.GetActorVersion(),
```
LiorLieberman, line 73: whats the rationale for removing the actorVersion

internal/ateompath/ateompath.go
```diff
@@ -33,7 +33,8 @@ var (
 	// internal/imagecache). It lives under BasePath so the cached layer
 	// directories are visible at the same path in atelet (which writes them)
 	// and in every ateom pod (which mounts them as overlay lowerdirs).
-	ImageCacheDir = filepath.Join(BasePath, "image-cache")
+	ImageCacheDir          = filepath.Join(BasePath, "image-cache")
+	CredentialBrokerSocket = filepath.Join(BasePath, "credential-broker.sock")
```
LiorLieberman, line 40: can we add a comment on top of CredentialBrokerSocket?

internal/atunnel/credential.go
```diff
@@ -0,0 +1,184 @@
...
+	"time"
+
+	"github.com/agent-substrate/substrate/internal/credbundle"
+	"github.com/agent-substrate/substrate/internal/proto/ateletpb"
+	"github.com/agent-substrate/substrate/internal/substratex509"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+// BrokerCertificateSource owns atunnel's actor private key and obtains the
```
LiorLieberman, line 40: I could be wrong but dont we want atelet to be the one that generate the private keys?

internal/atunnel/egress.go
```diff
@@ -88,45 +95,99 @@ func (e *Egress) Serve(ctx context.Context, listener net.Listener) error {
 	}
 }
 
-// Activate allows egress for one actor. There can be only one active actor per
-// worker. bearerToken may be empty until actor JWT issuance is available.
-func (e *Egress) Activate(dialer EgressDialer, atespace, actorName string, actorVersion int64, bearerToken string) error {
+// Activate allows egress with a previously obtained actor certificate and
```
LiorLieberman, line 102: >with a previously obtained actor certificate

thinking out loud - do we care about actorVersion(s)?

internal/proto/ateletpb/atelet.proto
```diff
@@ -18,6 +18,22 @@ package atelet;
...
 option go_package = "github.com/agent-substrate/substrate/internal/proto/ateletpb";
 
+// CredentialBroker gives an authenticated worker its current actor credential.
+service CredentialBroker {
+  rpc MintActorCertificate(MintActorCertificateRequest) returns (MintActorCertificateResponse) {}
+}
+
+message MintActorCertificateRequest {
+  // DER-encoded PKCS #10 certificate signing request. Atunnel retains the
+  // corresponding private key.
```
LiorLieberman, line 28: My initial thinking was that atelet is the thing we trust, and hence it generates the private keys and hands them to atunnel. Like kubelet. But your approach may be bettter, though keys would have to be regenerated across suspend/resume?

/cc @ahmedtd for thoughts.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -549,6 +539,10 @@ message MintCertRequest {
   // The signer will ignore the contents of the CSR except to extract the
   // subject public key.
   bytes certificate_signing_request = 4;
+
+  // Exact worker Pod requesting the certificate. Ateapi verifies that this
+  // worker is still assigned to the actor before signing.
+  string worker_pod_uid = 5;
```
ahmedtd: Can you add an additional note here to the effect that it's safe to trust the worker_pod_uid specified in the request body because atelet is the client, and atelet is trusted to tell us which worker pod is requesting the certificate (as long as the worker pod is associated with that particular atelet).

cmd/ateapi/internal/actoridentity/actoridentity.go
ahmedtd, line 243: We need an additional indication here that this is an ateom or atunnel, not the actor using its own identity.

Maybe fields in the ActorIdentity extension for:
* SystemComponentName
* SystemComponentPodName
* SystemComponentPodUID

cmd/atelet/credentialbroker.go
ahmedtd: I think it would be fine to merge without that, and then add it in a follow up?  Either way works.

cmd/atelet/credentialbroker.go
```diff
@@ -0,0 +1,121 @@
...
+	// TODO: Before release, request an atunnel-specific MintCert purpose and
+	// require the egress PEP to reject generic actor certificates.
+	// The request deliberately carries no actor identity. The authenticated Pod
+	// UID is the only input used to select a worker and its current assignment.
+	workerUID, err := authenticatedWorkerUID(ctx)
+	if err != nil {
+		return nil, err
+	}
+	var assigned *ateapipb.Worker
+	for pageToken := ""; ; {
```
ahmedtd: Looping over all workers isn't going to work even in the medium term.  Do we have any way to fetch the correct worker in a single request?

cmd/atelet/main.go
ahmedtd: Normally ate-api-server dials out to atelet.  In this case, we need to dial back.

cmd/atelet/main.go
```diff
@@ -192,6 +211,40 @@ func main() {
...
+	}
+	ateletIdentity, err := substratex509.PodIdentityFromCertificate(ateletCert.Leaf)
+	if err != nil {
+		serverboot.Fatal(ctx, "Failed to load atelet Pod identity", err)
+	}
+	if ateletIdentity == nil {
+		serverboot.Fatal(ctx, "Failed to load atelet Pod identity", fmt.Errorf("credential bundle has no Pod identity"))
+	}
+	brokerTLS := tlsCfg.Clone()
+	brokerTLS.VerifyConnection = restrictClientToNode(ateletIdentity)
```
ahmedtd: I think over a UDS we should be able to get this level of control just from filesystem permissions?  Ie, we create a UDS socket file for each individual ateom, and make sure only that ateom can access it.

TLS over a UDS is a bit strange.

EItanya: Sure, we can do that. To get the same isolation from filesystem permissions, though, we would need a separate broker hostPath, mount only each worker’s own socket directory into its ateom, and manage one listener per worker. The current hostPath is shared by all root-running ateoms, so `0600` on a per-worker filename alone would not isolate them.

mTLS gives us the same worker-to-atelet binding with considerably less change in the current layout. Would you be okay with keeping mTLS here?

internal/atunnel/egress.go
```diff
@@ -88,45 +95,99 @@ func (e *Egress) Serve(ctx context.Context, listener net.Listener) error {
...
+	}
+}
+
+func renewAfter(expiresAt time.Time) time.Duration {
+	remaining := time.Until(expiresAt)
+	return remaining - remaining/10
+}
+
+func retryAfter(expiresAt time.Time) time.Duration {
+	remaining := time.Until(expiresAt)
```
thompsonmax, line 189: IIUC, the current logic will stop attempting refresh if the credential expires. Should we continue to attempt refresh after expiry to prevent wedging if we fail to refresh in the last 10% of the credential lifetime?

cmd/ateapi/internal/actoridentity/actoridentity.go
EItanya: The JWT-specific changes that prompted this are gone; this PR no longer changes MintJWT.

cmd/atelet/credentialbroker.go
EItanya: Done. MintCert now requires the ATUNNEL purpose, embeds it in the ActorIdentity certificate extension, and atunnel verifies it before installing the certificate. PEP-side purpose enforcement remains a before-release follow-up.

cmd/ateapi/main.go
EItanya: Ateapi owns the deployment-level gateway selection and passes it with each Run/Restore request. Actor has no egress field today; an empty deployment value disables redirected egress.

cmd/ateom-gvisor/main.go
EItanya: Agreed; renamed it back to atunnelEgressPort.

cmd/atelet/credentialbroker.go
EItanya: Renamed this to actorIdentityClient.

cmd/ateom-gvisor/main.go
EItanya: It is more than a reorder: the command and created-container list now exist before the deferred failure cleanup, so every post-create failure deactivates networking and deletes only the containers created by that attempt.

cmd/atelet/credentialbroker.go
EItanya: Renamed this to verifyClientOnSameNode and added a comment describing the exact node-incarnation check.

cmd/atelet/main.go
EItanya: Correct: the existing direction was ateapi dialing atelet. The broker adds the reverse path because atelet resolves the authenticated worker assignment and then calls GetWorker, GetActor, and MintCert.

cmd/ateom-gvisor/main.go
EItanya: Atunnel still uses the worker Pod identity only to authenticate to the node-local atelet. It presents the new purpose-scoped actor certificate to the egress gateway. I renamed these variables to make those distinct roles explicit.

cmd/ateom-gvisor/main.go
EItanya: Actor resource version remains in actor metadata for control-plane concurrency. It is no longer sent to egress: atunnel creates a new private key and purpose-scoped certificate per activation, and ateapi revalidates the current worker assignment when minting.

internal/atunnel/egress.go
EItanya: Good catch. Fixed: expiry still blocks new tunnels, but renewal now continues every 30 seconds after expiry and restores admission only after a valid certificate is installed.

internal/atunnel/credential.go
EItanya: I kept key generation in atunnel so the private key never crosses the broker boundary. Only the CSR goes to atelet; renewals reuse that key within one activation.

cmd/ateom-gvisor/main.go
EItanya: Same answer here: resource version remains a control-plane concern; a new key/certificate is minted for each atunnel activation and is bound to the current assignment.

cmd/ateapi/internal/actoridentity/actoridentity.go
EItanya: Added ActorCertificatePurpose with ATUNNEL, encoded it in the ActorIdentity extension, and made both ateapi and atunnel reject a missing or different purpose.

cmd/ateom-gvisor/main.go
EItanya: Yes: for configured tunneled egress we mint once before starting each activation, fail closed if that mint fails, reuse the private key while renewing at 90% lifetime, and deactivate/clear it with the actor.

cmd/atelet/credentialbroker.go
EItanya: Added exact GetWorker(namespace, pod) lookup backed by the existing worker cache, and changed atelet to use it instead of scanning ListWorkers.

internal/ateompath/ateompath.go
EItanya: Added a comment describing the node-local broker socket and its use by atunnel.

cmd/ateom-microvm/restore.go
EItanya: Actor resource version remains in actor metadata. The egress credential is instead activation-scoped by a fresh atunnel-held key/certificate plus current-assignment validation at mint time.

internal/proto/ateletpb/atelet.proto
EItanya: The implementation now generates a new private key for each activation and reuses it only for renewals during that activation. Suspend/deactivate drops the source; the next resume gets a new key and certificate.

internal/atunnel/egress.go
EItanya: Resource version stays on the Actor for control-plane concurrency. For egress, a fresh atunnel-held key and purpose-scoped certificate are minted per activation, so the credential itself is not shared across activations.

pkg/proto/ateapipb/ateapi.proto
EItanya: Added this trust note to worker_pod_uid. Atelet derives the UID from the authenticated worker certificate; ateapi still checks that exact UID, node, and current actor assignment before signing.

EItanya:
Correction to my earlier summary: this PR has been retooled from bearer JWTs to purpose-scoped actor certificates. Atunnel generates and retains a per-activation private key, sends only a CSR to the node-local atelet, and atelet asks ateapi to mint an ATUNNEL-purpose actor certificate after revalidating the exact worker assignment. Atunnel renews that certificate in memory; no actor JWT is transported, so JWT issuer/subject work is no longer in scope here. PEP-side purpose enforcement remains a before-release follow-up.

internal/atunnel/egress.go
```diff
@@ -88,45 +95,102 @@ func (e *Egress) Serve(ctx context.Context, listener net.Listener) error {
...
+	if remaining <= 0 {
+		return 30 * time.Second
```
thompsonmax: [nit] I think it's worth adding some jitter to this to avoid a synchronized thundering herd if multiple clients have their certificate expire.

internal/proto/ateompb/ateom.proto
```diff
@@ -51,9 +51,6 @@ message RunWorkloadRequest {
   string atespace = 1;
   string actor_name = 2;
   string actor_uid = 3;
-  // Actor resource version observed by ate-api when assigning this worker.
-  int64 actor_version = 9;
```
thompsonmax, line 55: Should we mark removed proto field numbers as reserved to prevent backwards compatibility issues?

internal/atunnel/egress.go
```diff
@@ -88,45 +95,102 @@ func (e *Egress) Serve(ctx context.Context, listener net.Listener) error {
...
+}
+
+func renewAfter(expiresAt time.Time) time.Duration {
+	remaining := time.Until(expiresAt)
+	return remaining - remaining/10
+}
+
+func retryAfter(expiresAt time.Time) time.Duration {
+	remaining := time.Until(expiresAt)
+	if remaining <= 0 {
```
thompsonmax, line 190: [nit] Is it worth logging when the cert expires and when it recovers from expiry to give some visibility into that? (ERROR on expiry, INFO on recovery)?

cmd/atelet/credentialbroker.go
```diff
@@ -0,0 +1,114 @@
...
+func (b *credentialBroker) MintActorCertificate(ctx context.Context, req *ateletpb.MintActorCertificateRequest) (*ateletpb.MintActorCertificateResponse, error) {
+	// TODO: Before release, require the egress PEP to reject actor certificates
+	// whose ActorIdentity purpose is not atunnel.
+	// The request deliberately carries no actor identity. The authenticated Pod
+	// identity is the only input used to select a worker and its current assignment.
+	workerIdentity, err := authenticatedWorkerIdentity(ctx)
+	if err != nil {
+		return nil, err
+	}
+	assigned, err := b.control.GetWorker(ctx, &ateapipb.GetWorkerRequest{
```
thompsonmax: Just thinking aloud here, since MintCert already re-validates everything via authorizeActor (it does its own GetActor/GetWorker plus the node/UID/assignment checks), the broker's GetWorker/GetActor calls are really just discovering which actor to name in the request, which the server then refuses to trust anyway. Could we instead pass the authenticated worker identity (namespace/name/uid from the pod cert we already extracted) in MintCertRequest and have ate-api resolve the current actor from the assignment itself?

We could also pass an expected_actor_uid that ate-api rejects with FAILED_PRECONDITION on mismatch , purely as a guard to keep the current deny-on-stale behavior if the actor is swapped while a mint is in flight.

This would drop the dependency on GetActor/GetWorker so that mint takes 1 round trip instead of 3 here and loses nothing security-wise since the handler makes the same checks we're making client-side here. Let me know if I'm missing something though.

internal/atunnel/egress.go
EItanya: Done. Post-expiry retries now use a 25–35 second jittered interval.

internal/proto/ateompb/ateom.proto
EItanya: Since this project is still pre-alpha and the wire API is not yet stable, I would prefer not to reserve these removed field numbers yet.

internal/atunnel/egress.go
EItanya: Done. Expiry logs once at WARN when new egress is blocked, and successful recovery logs once at INFO.

cmd/atelet/credentialbroker.go
EItanya: Done. Atelet now forwards the mTLS-authenticated worker namespace/name/UID and the expected actor UID in a single MintCert call. Ateapi exact-resolves the worker, derives the current actor from its assignment, verifies the reciprocal assignment and node/Pod UID, and treats expected_actor_uid only as a fail-closed stale-activation guard. The broker no longer calls GetWorker/GetActor, and the temporary public GetWorker RPC has been removed.

internal/atunnel/egress.go
```diff
@@ -88,45 +96,113 @@ func (e *Egress) Serve(ctx context.Context, listener net.Listener) error {
...
+	delay := renewAfter(expiresAt)
+	expired := false
+	for waitForRenewal(active.ctx, delay) {
+		if !expiresAt.After(time.Now()) && !expired {
+			slog.WarnContext(active.ctx, "Atunnel actor certificate expired; blocking new egress connections",
+				slog.Time("expiredAt", expiresAt))
+			expired = true
+		}
+		nextExpiry, err := active.certificateSource.Mint(active.ctx)
+		if err != nil {
```
thompsonmax, line 146: One more thought here, if Mint returns FailedPrecondition or PermissionDenied that would seem to indicate that the control plane no longer maps this worker to this actor and the activation is stale. Should we return immediately rather than retry for those errors?

internal/atunnel/credential.go
```diff
@@ -0,0 +1,194 @@
...
+			return (&net.Dialer{}).DialContext(ctx, "unix", s.socketPath)
+		}),
+	)
+	if err != nil {
+		return time.Time{}, err
+	}
+	defer conn.Close()
+	resp, err := ateletpb.NewCredentialBrokerClient(conn).MintActorCertificate(ctx, &ateletpb.MintActorCertificateRequest{
+		CertificateSigningRequest: csr,
+		ExpectedActorUid:          s.expectedActorUID,
```
thompsonmax, line 148: Should we add a defense in depth check that the returned certificate's ActorIdentity has the same UID as what we expect?

ahmedtd:
LGTM for the certificate bits.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -507,10 +497,9 @@ service ActorIdentity {
   // it on the actor's behalf, authenticating with its own client certificate
   // rather than a bearer token.
   //
-  // Authorization is decided on that client certificate: it must identify the
-  // atelet running on the same node as the worker Pod that currently hosts the
-  // requested actor, and the actor must still be running. Any other caller is
-  // rejected with PERMISSION_DENIED.
+  // Authorization is decided on that client certificate and the worker
+  // identity attested by atelet. Ateapi resolves the worker's current actor and
+  // verifies their reciprocal assignment before signing.
```
bowei: "reciprocal assignment" -- can we use more plain language? I had a hard time understanding what this comment is referring to

internal/proto/ateompb/ateom.proto
```diff
@@ -67,9 +64,14 @@ message RunWorkloadRequest {
...
-  // traffic uses direct egress instead of being redirected through atunnel.
-  optional string egress_gateway_address = 10;
+  // When absent, actor traffic uses direct egress instead of atunnel.
+  EgressGateway egress_gateway = 10;
+}
+
+// EgressGateway configures tunneled egress for one actor activation.
+message EgressGateway {
+  // address is the remote gateway's host:port.
+  string address = 1;
```
bowei, line 75: optional

internal/proto/ateompb/ateom.proto
```diff
@@ -67,9 +64,14 @@ message RunWorkloadRequest {
...
-  // Remote egress gateway selected for this activation. When absent, actor
-  // traffic uses direct egress instead of being redirected through atunnel.
-  optional string egress_gateway_address = 10;
+  // When absent, actor traffic uses direct egress instead of atunnel.
+  EgressGateway egress_gateway = 10;
+}
+
+// EgressGateway configures tunneled egress for one actor activation.
+message EgressGateway {
+  // address is the remote gateway's host:port.
```
bowei: host is IP:port or DNS name :Port

internal/proto/ateletpb/atelet.proto
```diff
@@ -48,6 +68,15 @@ message RunRequest {
...
   SandboxAssets sandbox_assets = 8;
+
+  // When absent, actor traffic uses direct egress instead of atunnel.
+  EgressGateway egress_gateway = 9;
+}
+
+// EgressGateway configures tunneled egress for one actor activation.
+message EgressGateway {
+  // address is the remote gateway's host:port.
+  string address = 1;
```
bowei, line 80: optional, see other comment

is this IP:port or host:port

bowei (commented):
look ok. I think we might want to clean up some of the API comments + make sure optional vs required is what we want.

internal/proto/ateompb/ateom.proto
EItanya: Done. Marked all Run/Restore `egress_gateway` fields explicitly optional in both internal APIs; absence continues to mean direct egress.

internal/proto/ateompb/ateom.proto
EItanya: Done. Clarified in both internal APIs that the required address is a DNS name or IP address followed by its port.

internal/atunnel/egress.go
EItanya: Done. `FailedPrecondition` and `PermissionDenied` now immediately block new egress connections and terminate renewal; existing admitted connections can still drain.

pkg/proto/ateapipb/ateapi.proto
EItanya: Done. Reworded this to say that ateapi verifies the worker is assigned to the actor and that the actor points back to that exact worker.

internal/atunnel/credential.go
EItanya: Done. Atunnel now rejects a returned certificate unless its embedded actor UID matches the expected actor UID for the activation.

internal/proto/ateletpb/atelet.proto
EItanya: Done. Marked this field explicitly optional and clarified that the required address is a DNS name or IP address followed by its port.

cmd/ateom-gvisor/main.go
LiorLieberman: interesting, there was not anything that cleaned orphan containers before?

cmd/atelet/main.go
LiorLieberman: [needs issue] lets open an for that

cmd/atelet/credentialbroker.go
LiorLieberman: [needs issue] @EItanya do we need an issue for that?

LiorLieberman:
LGTM
Will merge once workflows are done.

@EItanya please open the issues here (added [needs issue] tag)

Also we want to remove the egressgatewayadress flag and replace it with a proper proto on the actor.

cmd/atelet/credentialbroker.go
EItanya: I can create one, but really this should just go straight into your PR

cmd/atelet/main.go
EItanya: Tracking this investigation in #778: https://github.com/agent-substrate/substrate/issues/778

Issue #605, opened 2026-07-30
[P1] ENOSPC not classified as a terminal filesystem error — disk-full causes infinite retry loop


aditya-shantanu:
**Severity:** P1 (actor stuck, manual disk cleanup required)
**Component:** Node Layer — `cmd/atelet/sandbox_assets.go`
**Audit ID:** NL-4
**Maps to suspect:** N8

---

## Summary

`isTerminalFileSystemErr` (the function that determines whether a filesystem error
should crash the actor vs. be retried) lists several terminal conditions but omits
`syscall.ENOSPC`. When a node's disk fills up, every filesystem write during a
checkpoint or restore returns `ENOSPC`. Without the terminal classification, the control
plane retries the RPC immediately and repeatedly — each retry fails again with ENOSPC,
creating a tight loop that prevents any forward progress and consumes control-plane
resources.

---

## Impact

- Actor stuck in `STATUS_RESUMING` or `STATUS_SUSPENDING` indefinitely.
- Tight retry loop produces log spam and wastes ateapi goroutines.
- Disk-full condition is not surfaced as a clear error to the user — just repeated
  `FailedPrecondition` or transient errors on the RPC.
- On GKE, the node may OOM-kill the atelet pod due to log volume before any operator
  notice.
- Related: disk fills up partly because `LocalCheckpointsDir`, `StaticFilesDir`, and
  `ImageCacheDir` are never garbage-collected (see ISSUE-008, ISSUE-009).

---

## Root Cause

**File:** `cmd/atelet/sandbox_assets.go` lines 250–265 (`isTerminalFileSystemErr`)

```go
func isTerminalFileSystemErr(err error) bool {
    for _, terminalFileErr := range terminalFileErrs {
        if errors.Is(err, terminalFileErr) {
            return true
        }
    }
    return false
}

var terminalFileErrs = []error{
    os.ErrNotExist,
    os.ErrPermission,
    syscall.EISDIR,
    syscall.ENOTDIR,
    syscall.ENAMETOOLONG,
    syscall.ELOOP,
    syscall.EROFS,
    // syscall.ENOSPC is MISSING
}
```

When ENOSPC is returned by `writeSandboxRecord`, `ensureSandboxAssets`, or any of the
`MkdirAll`/`os.Create`/`io.Copy` calls in the restore/checkpoint path,
`wrapFileSystemErr` wraps it without the `ReasonTerminalFileSystemError` sentinel.
Callers treat it as a transient error and retry.

---

[...]

PR #710, opened 2026-08-03
fix(ateapi): mark atelet churn unavailable

### LINKED ISSUE #646: [P2] Transient atelet unavailability defaults to codes.Internal — outside the documented retryable set; strands actor in SUSPENDING and surfaces as http/500
> Dedup note: verified against all open+closed issues 2026-07-30 — nothing
> files the Internal-misclassification of transient infra conditions or the
> SUSPENDING-strand scenario. Closest relatives: #605 (inverse
> classification bug: terminal ENOSPC treated as retriable), #50 (same
> SUSPENDING symptom, different root cause — runsc exit 128, retry does NOT
> converge there), #554 (reentrancy umbrella), #600 (precedent for the
> "Terminating pod has no IP" window, worker-pod side).

## Summary

When the atelet DaemonSet pod on an actor's node is being replaced (routine
during node upgrades), SuspendActor fails with:

```
rpc error: code = Internal desc = internal server error: workflow failed at
step CallAteletSuspend: while getting atelet conn for worker pod: selected
atelet "ate-system/atelet-w78v7" has no assigned IPs
```

Mechanism (verified on main):
- `cmd/ateapi/internal/controlapi/dialer.go:112` produces the error as a
  plain `fmt.Errorf` — no gRPC code attached.
- `workflow_suspend.go:140` wraps it with `%w` — still no code.
- `internal/ateinterceptors/ateinterceptors.go:72` then defaults every
  non-status error to `codes.Internal "internal server error"`.

Why the code matters — this is not a style nit; the system documents
contracts around specific codes:
1. `docs/request-parking.md`: the router parks/retries only
   `FailedPrecondition`, `Aborted`, `Unavailable`; everything else is
   fail-fast. A self-healing ~10 s condition lands outside the retryable
   set.
2. Router HTTP mapping (PR #68): `Unavailable`/`FailedPrecondition` → 503,
   "anything else" → **500**. So Internal-classified churn errors surface
   to HTTP callers as 500s — which matches the **1,840 unexplained
   `http/500`** responses in the 2026-07-30 overnight soak (during
   ateapi/atelet chaos kills).

## Live reproduction (GKE substrate-poc, 2026-07-30)

`kubectl delete pod -n ate-system <atelet-on-actor-node> --grace-period=0`,
then SuspendActor on the RUNNING actor hosted there ~0.5 s later:
- SuspendActor returned the `Internal` error above.
- Actor left in `STATUS_SUSPENDING`. In-cluster HTTP probe went from steady
  200s to hanging until its 5 s client timeout — consistent with requests
  being parked on the resume path's `FailedPrecondition` (SUSPENDING is not
  resumable) with nothing repairing the state, rather than a fast 503.
- A retried SuspendActor after the replacement atelet got its IP converged
  cleanly (SUSPENDED → auto-resume on next request). Forward recovery
  works; the defect is purely that the error classification tells clients
  *not* to retry.

## Suggested fix

- In the dialer / workflow step, classify "no atelet conn / atelet has no
  assigned IPs / atelet not found" as `codes.Unavailable`.
- Audit other plain `fmt.Errorf` paths in workflow steps that hit the
  `ateinterceptors.go:72` Internal default for infrastructure-transient
  conditions (the #605 ENOSPC bug is the same audit from the opposite
  direction: both directions of misclassification exist).
- Router: ensure upstream `Unavailable` during wake maps to 503 +
  `Retry-After`, never a park-until-timeout hang.

orangeCatDeveloper:
Fixes #646

## Summary

When the atelet DaemonSet pod on a node is being replaced (routine during node upgrades), `DialForWorker` fails with a plain `fmt.Errorf`, so the interceptor defaults it to `codes.Internal` — clients are told not to retry an error that self-heals in seconds (#646). The two transient paths (atelet pod count ≠ 1, no assigned IPs) now return `codes.Unavailable`: the interceptor extracts the code through `%w` wrapping, the router parks and retries per docs/request-parking.md, and unrecoverable cases surface as 503 with `Retry-After: 1` (added to the router's immediate responses, 503 only).

Scope, per #646's own dedup notes: worker-pod-side lookups ("expected 1 pod match") stay as-is (#600's territory), credential/connection build failures remain Internal, and the full workflow-step audit is the inverse direction of #605.

## Test plan

- dialer: fake-informer scenarios (zero atelets, two atelets, atelet without IPs) assert `Unavailable`; a wrapping test asserts the code survives `fmt.Errorf("...: %w", err)` as the workflows wrap it; `ErrWorkerPodNotFound` sentinel unaffected
- router: `retry-after: 1` present on 503 immediate responses, absent on 500/504/403; RawValue encoding pinned

### DIFF
--- a/cmd/ateapi/internal/controlapi/dialer.go
+++ b/cmd/ateapi/internal/controlapi/dialer.go
@@ -31,7 +31,9 @@ import (
 	"github.com/spiffe/go-spiffe/v2/svid/x509svid"
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/utils/lru"
@@ -44,6 +46,16 @@ var ErrWorkerPodNotFound = errors.New("worker pod not found")
 // Distinct from ErrWorkerPodNotFound, which callers treat as crash-worthy;
 // this one is retryable.
 var ErrNoAteletOnNode = errors.New("no atelet pod found on node")
+
+// unavailableError keeps a sentinel reachable via errors.Is while carrying
+// codes.Unavailable, so the interceptor doesn't default it to Internal.
+type unavailableError struct{ error }
+
+func (e *unavailableError) GRPCStatus() *status.Status {
+	return status.New(codes.Unavailable, e.Error())
+}
+
+func (e *unavailableError) Unwrap() error { return e.error }
 
 // The SPIFFE identity that atelet serving certs carry, as minted by the
 // podidentity signer (cmd/podcertcontroller/internal/podidentitysigner).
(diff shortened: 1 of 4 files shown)


maxsmythe:
@bowei to evaluate networking impact and/or route review

PR #711, opened 2026-08-03
Define a production-ready default tracing policy

### LINKED ISSUE #584: Define a default tracing policy
## Expected Behavior

Users of substrate should be able to use tracing to identify problems in their systems by default.

See also https://github.com/agent-substrate/substrate/issues/580, https://github.com/agent-substrate/substrate/issues/581

## Actual Behavior

Today, users cannot really use tracing to investigate failures because 

- every components hardcodes its sampler at startup
- router is set to `NeverSample` 
- ateapi requires `--trace` via the CLI to be traces
- we never read `OTEL_TRACES_SAMPLER`

## Proposal

To follow OpenTelemetry recommendations, sampling should happen in the pipeline, not in the observed application itself.

So I am proposing the following:
- **Control-plane sampled by default:** 
  - ateapi, atelet, ateom set `ParentBased(AlwaysSample)`
  - Control plane request volume is small compared to the data plane, and these are the traces we need when suspend/resume/restore needs troubleshooting. 
  - We can extend this via exemplars on the upcoming lifecycle duration histograms.
 - **Data plane parent-based with a low default ratio at the root:** The router sets `ParentBased(TraceIDRatioBased(small))`. Clients sampled already stay sampled; the ratio defines how many unsampled requests we pull in.
 - **Leverage `OTEL_TRACES_SAMPLER`**: This is the OTel-native way of exposing this to end users

 ## Non-goals

 - Shipping or mandating a specific collector setup. We can have guidelines around this as per https://github.com/agent-substrate/substrate/issues/563
 - Any per-workload or per-tenant sampling configs

 ## Open questions

 1. Is a ratio at the router root acceptable for everyone as the data plane default, and what is a sane starting value? I would start conservative, e.g. 1%, it's configurable anyway.
 2. Do we want `always_on` for the lifecycle plane at target scale (1B actors means a lot of lifecycle ops), or should `ateapi` also default to a higher (than 1%) ratio?

krisztianfekete:
Today everything outside ateapi hardcodes `ParentBased(NeverSample)`, and setting `OTEL_TRACES_SAMPLER` does nothing because an explicit sampler silences the SDK's env handling. This makes troubleshooting production issues impossible via traces.

This PR gives every component a sane default and makes the standard OTel env vars work:

  - Control plane (ateapi, atelet, ateom) defaults to `parentbased_traceidratio` 0.1, the router (data plane root) to 0.01, per the discussion on #584.
  - `OTEL_TRACES_SAMPLER` / `OTEL_TRACES_SAMPLER_ARG` override any of this without a rebuild. Invalid values keep the component default and log a warning instead of inheriting the SDK's fall-open-to-100% behavior.
  - Envoy's `RandomSampling` is derived from the router's resolved policy, so the two root decisions cannot drift.
  - kubectl-ate without `--trace` no longer installs a tracer provider at all: the old `NeverSample` provider injected `sampled=0`, which pinned every parent based sampler downstream and would have defeated the server side ratios. `--trace` still forces a full end to end trace.
  - ate-controller propagates the two env vars to the ateom worker pods it creates, same as the metric export vars.
  - kind pins ateapi to `parentbased_always_on`, so the local Jaeger flow keeps showing every API call.
 - The agentgateway integration follows what we have above. its `randomSampling` changes from `true` to `0.01` to match the data plane default, though as static config it does not follow `OTEL_TRACES_SAMPLER` overrides, so the two need adjusting together.

Verified on a kind cluster e2e manually. It resolved samplers logged at startup, a `--trace` resume produced one trace across ateapi, atelet, and ateom, an unsampled CLI call got picked up server side, Envoy continued a sampled traceparent while sampling 0 of 30 parentless requests at the 1% default, and an invalid env value fell back to the component default.

Gating who may use `--trace` stays a separate follow-up (and a discussion), and the more sophisticaed tracing policies belongs in a collector, not in substrate.

Fixes #584
cc. @git286 

- [x] Tests pass

### DIFF
--- a/internal/serverboot/sampling.go
+++ b/internal/serverboot/sampling.go
@@ -1,0 +1,137 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package serverboot
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"os"
+	"strconv"
+	"strings"
+
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+const (
+	tracesSamplerEnv    = "OTEL_TRACES_SAMPLER"
+	tracesSamplerArgEnv = "OTEL_TRACES_SAMPLER_ARG"
+)
+
+// ControlPlaneTraceRatio is the default root sampling ratio for the control
+// plane binaries: low volume, and their traces are what lifecycle debugging
+// needs, so the default leans generous.
+const ControlPlaneTraceRatio = 0.1
+
+// TraceSampling is a resolved sampling policy. The root fraction is kept
+// alongside the sampler because the atenet router must mirror it into Envoy's
+// RandomSampling percent, which roots the data plane traces.
+type TraceSampling struct {
+	sampler   sdktrace.Sampler
+	rootRatio float64
+}
+
+func (s TraceSampling) Sampler() sdktrace.Sampler { return s.sampler }
+
+// RootSamplingPercent is the percentage (0 to 100) of parentless requests
+// sampled at the root.
+func (s TraceSampling) RootSamplingPercent() float64 { return s.rootRatio * 100 }
+
+// ParentRatioSampling is ParentBased(TraceIDRatioBased(ratio)), clamped to [0, 1].
+func ParentRatioSampling(ratio float64) TraceSampling {
+	ratio = min(max(ratio, 0), 1)
+	return TraceSampling{
+		sampler:   sdktrace.ParentBased(sdktrace.TraceIDRatioBased(ratio)),
+		rootRatio: ratio,
+	}
+}
+
+// ParentNeverSampling is ParentBased(NeverSample()).
+func ParentNeverSampling() TraceSampling {
+	return TraceSampling{sampler: sdktrace.ParentBased(sdktrace.NeverSample())}
+}
+
+// ResolveTraceSampling applies OTEL_TRACES_SAMPLER / OTEL_TRACES_SAMPLER_ARG on
+// top of the component default. serverboot resolves the env itself because an
+// explicit WithSampler silences the SDK's env handling, and the SDK falls open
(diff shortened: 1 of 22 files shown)


maxsmythe:
@git286 for observability

Also, there are merge conflicts that need resolving

krisztianfekete:
Thanks @maxsmythe, one usually has to resolve conflicts 5+ times before reviews, will do it shortly. 
Also, not sure what we can do about auto-adding folks for reviews, or just adding the ability for non-org members to request reviews as currently I cannot do that. Maybe it could improve the process as well.

maxsmythe:
Yeah, as far as adding reviewers... we are looking to at least get some manual coverage for that in the short term.

No worries if you want to wait for reviewer activity for rebase.

internal/serverboot/sampling.go
```diff
@@ -0,0 +1,137 @@
...
+		return def, nil
+	}
+	name = strings.ToLower(strings.TrimSpace(name))
+	// Unlike the SDK, treat set-but-empty as unset: templated manifests can
+	// render empty env vars.
+	if name == "" {
+		return def, nil
+	}
+
+	ratio := 1.0
```
JeffLuoo: My understanding is that if the arg is unset, we will set it to 100% here. It may conflict with we intent to not set it to 100% in this PR. Should we set the default value here to a lower value?

internal/serverboot/sampling.go
```diff
@@ -0,0 +1,137 @@
...
+// to 100% sampling on invalid values where this keeps the default and logs.
+func ResolveTraceSampling(ctx context.Context, def TraceSampling) TraceSampling {
+	name, nameSet := os.LookupEnv(tracesSamplerEnv)
+	arg, argSet := os.LookupEnv(tracesSamplerArgEnv)
+	resolved, err := resolveTraceSampling(name, nameSet, arg, argSet, def)
+	if err != nil {
+		slog.WarnContext(ctx, "Invalid trace sampler environment, keeping the component default",
+			slog.String("sampler", name),
+			slog.String("arg", arg),
+			slog.String("default", def.sampler.Description()),
```
JeffLuoo: def.sampler may be unset, can you handle it here and and a unit test to cover unset def.sampler?

JeffLuoo (changes requested):

git286 (dismissed):

internal/serverboot/sampling.go
krisztianfekete: Yep, will do this shortly!

internal/serverboot/sampling.go
krisztianfekete: The current behavior follows the OTel spec (missing `OTEL_TRACES_SAMPLER_ARG` means ratio 1.0) with a warning. But you have point that this is the one remaining fail open to 100% path, and the PR's approach is that invalid missing configs keeps the component default, so I can make this consistent with the rest, thank you!

JeffLuoo (approved):

Issue #465, opened 2026-07-19
Suspend-safe actor networking via injected in-sandbox ingress/egress proxies


ronlv10:
## Background
This is a follow-up to the summit, where we identified this pattern and validated it with a PoC that suspends a real agent actor (openclaw) mid WebSocket and LLM/MCP calls.

Today both ingress and egress traffic reach the actor over TCP connections that cross the sandbox boundary: inbound requests are forwarded from outside the sandbox to a port the actor listens on, and the actor's outbound connections have their peer endpoint outside the sandbox.

A TCP connection's state lives at both endpoints, so any connection crossing the boundary is broken by suspend, for two reasons: checkpointing freezes only the sandbox side, so the outside half is closed or lost and the actor's frozen half has no peer on restore; and the actor may resume on a different worker with a different IP, which TCP cannot follow. But connections with **both endpoints inside the sandbox** - loopback/localhost - survive checkpoint/restore intact, since both halves are snapshotted together and are insensitive to where the sandbox resumes. That idea is the basis of this proposal.
## Problem
An actor that is serving an inbound HTTP request, or waiting on an outbound HTTP call, cannot be safely suspended: the socket drops and the request fails. It forces actors to be suspend-aware, forces the platform to avoid suspending actors with in-flight traffic, and makes long-running calls (e.g. an agent waiting minutes on an LLM response) effectively pin the actor as running.

Being able to suspend an actor at _any_ point, in-flight traffic included, greatly boosts the value of suspend/resume and the system as a whole: it turns it from an idle-only optimization into a general platform mechanism:

- **Density and cost**: agents spend most of their wall-clock time waiting on slow calls (LLM inference, tools, humans). If waiting pins the actor as running, the workloads that would benefit most from suspension are exactly the ones that can never be suspended. Suspend-during-wait lets a node oversubscribe to actual CPU use rather than open connections.
  
- **Upgrades and maintenance**: draining a worker for a node upgrade, or rebalancing no longer needs to wait for actors to go quiet or fail their in-flight requests - network-wise, any actor can be checkpointed immediately and resumed elsewhere.
  
- **Scheduling freedom**: the scheduler can possibly preempt and migrate actors based on resource pressure alone, without reasoning about connection state or coordinating with the traffic path.
  
- **Failure recovery**: a resumed-from-checkpoint actor comes back with its in-flight work intact instead of with a lap full of broken sockets it must detect and retry.
  
## Proposal

[...]

PR #712, opened 2026-08-03
docs: convention for integration repository structure and naming

mayawang:
## Summary

Adds `docs/integration-repos.md`: where end-to-end integrations live, how their
repositories are named, and how the fixes they need flow back into core.

The convention in one line — trivial demos stay in the core repo, each
non-trivial integration gets one dedicated repo under the `agent-substrate`
org, and core gaps get closed by making core configurable with defaults
unchanged rather than by patching it downstream.

## Why now

We are about to create the first real, end-to-end integrations rather than
counter-style demos: a code-execution sandbox, and an always-on agent. Both are
large enough to need their own images, dependencies, and release cadence.

Whichever repository gets created first will set the precedent for every one
after it. This writes the convention down so that precedent is chosen
deliberately instead of inherited by accident.

## What it covers

- **Where code lives** — the core-repo/dedicated-repo split, the rough test for
  which side something falls on (API keys, external services, third-party
  accounts), and why this is a set of peer repos rather than a second org.
- **Naming** — capability-named for general capabilities
  (`code-execution-sandbox`), integration-named for specific third-party
  products, named for the product rather than the vendor behind it. Plus what to
  avoid: over-broad names, names that clone a vendor's API or brand, and the
  redundant `-integration` suffix.
- **Third-party names** — allowed descriptively, with a non-affiliation note in
  the repo README, and brand/policy edge cases cleared before the repo exists.
- **Upstreaming** — the part with teeth for this repo. Integration repos that
  accumulate local patches against core bitrot, and the gap they work around
  stays invisible to everyone else. So: prefer making core behavior configurable
  with defaults unchanged. #487 and #465 are linked as illustrations of that
  pattern — this PR does not depend on either, and branches from `main`.
- **Two worked examples** that validate the convention rather than just
  following it, including the third-party-name edge case.

## Review

This was announced at the community meeting and circulated as a shared design
doc with a 7-day review window, which has now closed. It synthesizes the
`#integrations` thread discussion. Comment history:

<https://docs.google.com/document/d/1Tb6u0b1XSvWrNpoyD4jdsQaJ58aAgDtQOM18uxujs-8/edit>

This PR is the trimmed version: doc-review scaffolding — status block, reviewer
list, self-link — is dropped, and only the durable convention is carried over.

## Left open

Two questions are deliberately out of scope, called out in the doc rather than
answered. Both are maintainer calls and neither blocks the first repositories:

- Governance tiers — whether to distinguish "official" from "community"
  integrations with different review bars, as Home Assistant and Obsidian do.
- Who creates integration repositories and grants per-integration maintainer
  access.

## Also in this PR

- README gets an entry in the docs list, matching every other file in `docs/`.
- `CONTRIBUTING.md` gets one sentence pointing there, since "where does my
  integration go?" is a question a contributor asks before opening a PR.

- [x] Tests pass

### DIFF
--- a/docs/integration-repos.md
+++ b/docs/integration-repos.md
@@ -58,7 +58,7 @@ repositories under `agent-substrate` does not.
 Two cases, depending on what the repository actually is:
 
 **Capability-named**, when it provides a general Substrate capability that
-happens to have one implementation today. Prefer `code-execution-sandbox` over
+happens to have one implementation today. Prefer `execution-sandbox` over
 `sandbox` (too broad) or a vendor's product name (too narrow).
 
 **Integration-named**, when it integrates one specific *open-source* project.
@@ -70,13 +70,19 @@ name refers to something anyone can read, run, and fork, so using it
 descriptively claims nothing. A proprietary product's name is a brand we would
 be borrowing, and borrowing it implies an endorsement or a compatibility promise
 that neither side has made. When the thing being integrated is proprietary, use
-a capability name instead: `code-execution-sandbox`, not the vendor's product
+a capability name instead: `execution-sandbox`, not the vendor's product
 name.
 
 Avoid:
 
 - **Generic names** such as `sandbox` or `plugins`, which claim far more ground
   than any one repository covers.
+- **Over-specific names**, which have the opposite failure: they steer people
+  away from a solution that would have worked for them. A sandbox that is often
+  used for code execution is still a general execution sandbox, and naming it
+  `code-execution-sandbox` invites a reader with a different workload to
+  conclude it is not for them. A name should be just long enough to describe
+  the thing, and no longer.
 - **Names that clone a proprietary API or brand**, which quietly commits the
   project to chasing someone else's naming decisions.
 - **The `-integration` suffix.** Every repository in this category is an
@@ -120,9 +126,11 @@ first" a short path rather than a blocking one.
 
 These two validate the convention rather than merely following it:
 
-- **`agent-substrate/code-execution-sandbox`** — capability-named. A sandboxed
-  code-execution service built on Substrate, in the spirit of existing
-  code-execution products but not modeled on any one of their APIs.
+- **`agent-substrate/execution-sandbox`** — capability-named. A sandboxed
+  execution service built on Substrate, in the spirit of existing sandbox
+  products but not modeled on any one of their APIs. The name stops at the
+  capability on purpose: code execution is one workload it serves, not the
+  boundary of what it is.
 
 - **`agent-substrate/always-on-agent`** — a connection-holding agent: a
   multi-tenant gateway plus a suspendable per-conversation actor. Its first
--- a/docs/integration-repos.md
+++ b/docs/integration-repos.md
@@ -1,0 +1,140 @@
+# Integration Repositories: Structure and Naming
+
+## Summary
+
+Substrate is acquiring its first end-to-end integrations — workloads that *run
+on* Substrate rather than demonstrate it. This document records where that code
+lives, how the repositories are named, and how fixes flow back into core.
+
+In short: trivial demos stay in the core repository, each non-trivial
+integration gets one dedicated repository under the `agent-substrate`
+organization, and gaps in core are closed by landing the change in core
+first — never by patching core downstream.
+
+## Motivation
+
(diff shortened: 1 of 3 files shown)


BenTheElder:
Meta point: the commit doesn't seem to be associating with your account? The git author info should include an email address that you have associated with your github account (you can associate mutliple).

docs/integration-repos.md
```diff
@@ -0,0 +1,114 @@
...
+
+**Integration-named**, when it integrates one specific third-party product. Name
+it for the product, not the vendor behind it — `hermes`, not the name of the
+organization that publishes Hermes.
+
+Avoid:
+
+- **Generic names** such as `sandbox` or `plugins`, which claim far more ground
+  than any one repository covers.
+- **Names that clone a vendor's API or brand**, which quietly commits the
```
BenTheElder: this is slightly at odds with the integration bit below, should we limit them to OSS projects or something?

I don't think we have to fully resolve that now, but ... bit tricky

docs/integration-repos.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Upstreaming: prefer configurability over forks
+
+Real integrations surface real gaps in core. Building a long-running agent on
+Substrate turned up the need for configurable timeouts and golden-snapshot
+warmup ([#487](https://github.com/agent-substrate/substrate/pull/487)), and ran
+into suspend-safe actor networking
+([#465](https://github.com/agent-substrate/substrate/issues/465)).
+
+The path for those fixes should be short and well-travelled. An integration
```
BenTheElder: I don't think integration repositories that we host should exist with *any* required patches to core, we should always land the patch to core first.

I think it's OK for independent POCs, but for official projects it's embarassing to depend on changes that don't exist in our own upstream yet. Let's just avoid that entirely. We are our own upstream if we're hosting both repos.

I'd drop this section or require no forking of core.

docs/integration-repos.md
```diff
@@ -0,0 +1,114 @@
...
+  implementation is built on a third-party agent runtime, and the repository is
+  capability-named rather than vendor-named while that name goes through the
+  brand check described above. It is the third-party-name edge case in practice.
+
+## Not settled yet
+
+These are open questions for the maintainers, deliberately left out of scope
+here so they do not block the first repositories:
+
+- **Governance tiers.** Whether to distinguish "official" from "community"
```
BenTheElder: IMHO: let community host their own outside the org. It's confusing to have things that are under our brand but not considered official. Let's just not do that.

docs/integration-repos.md
BenTheElder: This is the main thing I'd change.

mayawang:
> Meta point: the commit doesn't seem to be associating with your account? The git author info should include an email address that you have associated with your github account (you can associate mutliple).

Thanks Ben! I updated PR based on comments and also fixed the email. PTAL.

BenTheElder (dismissed):
LGTM

docs/integration-repos.md
```diff
@@ -0,0 +1,140 @@
...
+**Integration-named**, when it integrates one specific *open-source* project.
+Name it for the project, not the vendor behind it — `hermes`, not the name of
+the organization that publishes Hermes.
+
+Integration-naming is limited to open-source projects. An open-source project's
+name refers to something anyone can read, run, and fork, so using it
+descriptively claims nothing. A proprietary product's name is a brand we would
+be borrowing, and borrowing it implies an endorsement or a compatibility promise
+that neither side has made. When the thing being integrated is proprietary, use
+a capability name instead: `code-execution-sandbox`, not the vendor's product
```
thockin: A very specific name can sometimes steer people AWAY from a perfectly good solution.  There's not much that makes our "sandbox" specific to code-execution, but people may see that and assume that it is not "for them".

"execution-sandbox" is less biased.  Or "remote-exec-sandbox"?

There's also some value in brevity, when it doesn't come at the cost of clarity.

Names should be just long enough to describe the thing, and not longer.

mayawang: Agreed — went with execution-sandbox. Dropped "remote" since it's implied by it being a hosted service, per your brevity point. Also added the missing half to the Avoid list: it only warned about names that are too broad, and over-specific names are the opposite failure. Closed it on your line: "a name should be just long enough to describe the thing, and no longer".

PR #713, opened 2026-08-03
fix: detect -allow-connected-on-save support instead of assuming it

mayawang:
> Split out of #487, which bundled three unrelated changes. This one is a bug
> fix with no design question attached, so it is sent first and on its own.

## Summary

ateom passes `runsc start -allow-connected-on-save` unconditionally. Some runsc
builds do not define the flag and reject the invocation with `flag provided but
not defined`, which fails **every actor start** on those builds.

This probes the binary instead of assuming, so ateom works across runsc builds
with no operator configuration.

## Changes

`cmd/ateom-gvisor/runsc.go` — `supportsAllowConnectedOnSave` shells out to
`runsc flags` once per binary path (memoized in `runscFlagSupport`, since
`cmdStart` runs on every container start) and passes the flag only when that
build lists it. A probe that cannot run at all assumes the flag *is* supported,
preserving today's behavior wherever detection is inconclusive.

It has to be `runsc flags`, not `runsc help start`. `-allow-connected-on-save`
is a top-level flag, and the per-subcommand usage that `help start` prints lists
only `-h` and `-help` even on builds that define it — so probing the subcommand
would report "unsupported" everywhere and quietly stop passing a flag that
works. `runsc_test.go` asserts the probe invokes `flags` specifically, because
the stub binaries answer any argv and a regression back to `help start` would
otherwise pass.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` — all pass.
- `runsc_test.go` covers: flag in stdout usage, flag in stderr usage, flag
  absent, probe failure fails open, missing binary fails open, per-path
  memoization, and the `flags`-subcommand assertion above.
- **Validated against five real runsc binaries.** The probe's verdict was
  compared against ground truth — whether the flag is defined in the binary, and
  whether the binary actually accepts it. It agrees on all five:

  | runsc build | Defines the flag | Probe verdict |
  |---|---|---|
  | `google-873132245` (GKE-Sandbox) | no | drops the flag |
  | `release-20260504.0` | no | drops the flag |
  | `release-20260622.0` | yes | passes the flag |
  | `release-20260706.0` | yes | passes the flag |
  | `release-20260706.0-21-g6d73c10d` | yes | passes the flag |

  The two rejecting builds fail *any* invocation with `flag provided but not
  defined` — note both spellings, one build says `flags provided` — so string
  matching on the error text is not a reliable substitute for the flag listing.
- The original failure was hit e2e on a GKE-Sandbox build that rejects the flag.
  At that time it was worked around with an explicit opt-out; the automatic
  probe replaces it, validated as above.

## Out of scope

The gVisor `SandboxConfig` change that points `runsc.url` at a specific build is
deliberately not included — it is environment-specific (a private bucket and
build SHA) and does not belong upstream. The portable finding is this probe.

- [x] Tests pass

### DIFF
--- a/cmd/ateom-gvisor/runsc.go
+++ b/cmd/ateom-gvisor/runsc.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -25,6 +26,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sync"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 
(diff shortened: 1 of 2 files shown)


mayawang:
Closing this. I just learnt that the gVisor team has ongoing work on runsc save/restore flags, so rather than probing for capability on our side we'll move to a recent runsc build and pin a minimum supported version during GA.

Issue #593, opened 2026-07-29
Enable caching for the Router for ResumeActor


bowei:
We should understand how to cache the metadata about the Actor after a successful resume. Currently every request ends up hitting the ATE API server for the current Actor. Being able to cache the state and response will be able to cut down significantly on the API.

PR #715, opened 2026-08-03
Support CONNECT in atenet router

### LINKED ISSUE #689: Support multiple, arbitrary actor ports
Support routing to actors (ingress) on arbitrary ports and not only 80/443.

Related discussions (that will likely be solved with the same solution): 

- https://github.com/agent-substrate/substrate/issues/484
- https://github.com/agent-substrate/substrate/issues/445
- https://github.com/agent-substrate/substrate/issues/265

(do not self assign)

cc: @keithmattix @bowei @howardjohn @yanavlasov

keithmattix:
Fixes #689, #265 and starts to address #484. Modifies the ate router to:

1. Read :authority from extproc forwarded metadata instead of the requests authority (necessary for CONNECT since the inner http request may not have the proper host header). 
2. Always send to the port in the authority header (passed back to Envoy via metadata)  
3. Serve network extproc on a separate port (needed for TCP tunneled within CONNECT). Doesn't fully work though; see below.

#484 is currently blocked because Envoy's implementation of NetworkExternalProcessor doesn't support passing filter state to the extproc server. Dynamic metadata doesn't work because only Host (i.e. endpoint) and cluster metadata can propagate across internal listeners (a necessary implementation detail of doing CONNECT in Envoy). This PR still adds the xDS for it; the extproc calls just fail because there's no metadata. 

Open questions:
1. Do we need all combinations of [plaintext CONNECT, TLS CONNECT] x [http 1.1, h2]? I've implemented all 4 but wanted to confirm
2. This PR still has extproc relying on :authority header to determine actor+atespace since potential changes to DNS (e.g. maybe we use headers instead) are under discussion. Need to resolve that at some point, but doesn't have to block this.
3. ~~We're assuming that non HTTP actors don't get suspended during the lifetime of a TCP connection; we only run Resume on the initial connection establishment. We could potentially do network ext authz instead (it actually on gets invoked once during conn establishment vs. we hit extproc on every client write), but wanted to discuss. I will say that I highly doubt that we'll need ANYTHING in the TCP body.~~ Per @LiorLieberman's comments, there's Envoy work being done on network ext_proc, so I assume that's what we're doing. 
4. Not sure if we've agreed if we want to do CONNECT (+ TLS?) to atunnel or not. This PR doesn't do that yet. I have no problem adding that once this merges.

Tested 6 scenarios locally:

1. Plain HTTP ingress → hello from: 169.254.17.2 | preserved memory count: 14...
2. Plain HTTPS ingress → hello from: 169.254.17.2 | preserved memory count: 15...
3. CONNECT (plaintext) + HTTP/1.1 → port 9090 → hello from extra port 9090 on pod 169.254.17.2
4. CONNECT (plaintext) + h2c → port 9090 → hello from extra port 9090 on pod 169.254.17.2
5. CONNECT+TLS + HTTP/1.1 → port 9090 → hello from extra port 9090 on pod 169.254.17.2
6. CONNECT+TLS + h2c → port 9090 → hello from extra port 9090 on pod 169.254.17.2

~~Once #652 merges, I can add the agentgateway implementation~~ Added and tested agentgateway implementation as well

- [X] Tests pass


maxsmythe:
@bowei to route networking PR

bowei:
Can you put the vendor change in its own commit (make it obvious)

keithmattix:
@bowei good call - done

cmd/atenet/internal/router/extproc_network.go
```diff
@@ -0,0 +1,207 @@
...
+// the connection to that actor's worker.
+//
+// TODO(router): authority is always empty today (see SubstrateMetadataNamespace's
+// doc comment) -- dynamic metadata doesn't survive the connect_terminate ->
+// main_internal internal-listener hop, and NetworkExternalProcessor has no
+// filter-state-reading mechanism (no request_attributes) to fall back on the
+// way the HTTP leg's handleRequestHeaders does. Until Envoy adds one, every
+// CONNECT scenario that reaches this raw TCP leg -- non-HTTP payloads, and any
+// TLS-wrapped payload per buildMainInternalListener's transport-protocol
+// match -- fails here.
```
LiorLieberman: not sure I understand that. Can you elaborate? 

AFAIU the metadata fails to cross the hop likely because `passthrough_metadata` only forwards static Host/Cluster metadata and never carried the value in the first place. I dont think its a NetworkExternalProcessor gap. Unless I am missing something

LiorLieberman (commented):
Why do we need that network ext_proc? Can we resolve instead on the CONNECT itself. `connect_terminate` listener is already an HCM with no ext_proc on it - we can add the http ExtProc there. 

What are the concerns with that?

keithmattix:
@LiorLieberman at connect_terminate, we don't know whether the CONNECT is tunneling HTTP or not and we want different behavior depending on the inner protocol. If it's HTTP, we want to hit the ext proc on every request so that we Resume. If it's just TCP, we only want to hit the extproc once (on connection establishment). But we don't know which one we want until we get to main_internal and sniff the protocol of the CONNECT tunnel's contents.

cmd/atenet/internal/router/extproc_network.go
keithmattix: I call it a NetworkExternalProcessor gap because its HTTP equivalent has a way to pass filter state to the extproc proc server and we use that to get the original CONNECT authority over there. Because request level metadata isn't passed through the internal listener transport socket; we can't use metadata to propagate it; we have to use filter state. So the requirements for the Envoy team is to be either allow internal listeners to propagate request metadata OR to allow network extproc to pass filter state via connection attributes. The same limitation applies to network ext auth btw

LiorLieberman:
Documenting some of the things Keith and I discussed just now; 

* This PR does implement CONNECT from the client to atenet and currently do not implement connect from atenet-to-atunnel. Rest of the bullets below focus on CONNECT between client to atenet.

>can we do HTTP ext_proc at first CONNECT? 

To answer this we need to agree on one thing -- we want/need extProc to call resumeActor on **every ** (inner) http request.  With this as a given, we can use extProc on the outer connect, but we _could_ loose the ability to reuse a persistent connection (likely has to be h2) between the harness and atenet-ingress and just stream new inner http requests within it as they come. We would loose it cause if we do extProc on outer CONNECT we wont trigger extProc on inner new http req. Long way to say that we DO need inner http ext procs.

If I try to put a visual representation, I think that this PR currently attempts to do 

CONNECT  -> **HTTP ext_proc** (for httptraffic)
_________ -> **network ext_proc** (for non-http traffic)

To avoid using NetworkExtProc I think we can do; 

CONNECT ext_proc **-> HTTP ext_proc** (_yes - we will have two extProc calls one when the CONNECT is established, the second one in the inner http -  will return early though_)
_________________**-> no network ext_proc** (for non-http traffic)

LiorLieberman:
Also @botengyao just added a PR (merged) for filter_state propagation yesterday.. 

Keith - take a look - https://github.com/envoyproxy/envoy/pull/46551

keithmattix:
Agree with the summary except for one thing (emphasis mine):

> yes - we will have two extProc calls once when the CONNECT is established, **the second one will return early though**)

Not in the common case right? Again, assuming we want per request extproc triggers, the second extproc will be triggered onRequestHeaders like it does now. If anything, the first one would need to return early for HTTP, but we can't even do that because we don't know what's inside the CONNECT tunnel

LiorLieberman:
> Agree with the summary except for one thing (emphasis mine):
> 
> > yes - we will have two extProc calls once when the CONNECT is established, **the second one will return early though**)
> 
> Not in the common case right? Again, assuming we want per request extproc triggers, the second extproc will be triggered onRequestHeaders like it does now. If anything, the first one would need to return early for HTTP, but we can't even do that because we don't know what's inside the CONNECT tunnel

I messed up my language earlier - fixed the language. `one when the CONNECT is established, the second one in the inner http - will return early though)`. 

>If anything, the first one would need to return early for HTTP, but we can't even do that because we don't know what's inside the CONNECT tunnel

why? Becuase we dont know whats the inside in CONNECT my point was -- outer -- always extproc (since we need it both for TCP and HTTP to call resume), then for the inner, for TCP we dont need again (cause we know the outer did it), for HTTP we do need again, cause it may be the second http stream and not the first one. So the only "cost" we paid is that we did two extProcs for http request in the first CONNECT initiation.

keithmattix:
I'm still not sure I understand: what does "return early" mean here practically? Does that mean the extproc hop is quicker than normal or that we won't call Resume() for the actor? If it's the latter, that still doesn't work if we assume that we want to call extproc for every inner http request; the second ext proc would need to be a fully fledged call to Resume so we can wake up the actor. Even if we just scope it down to the very first CONNECT establishment where we'll have double extproc, you'd need to have some sort of really small cache or other coordination mechanism if you wanted to return early (honestly may not be worth the hassle if ResumeActor() is idempotent)

LiorLieberman:
>(honestly may not be worth the hassle if ResumeActor() is idempotent)

Exactly, its just a super minor optimizations I suggested (say to keep a 30 sec cache or something and if you just called resume actor dont call again for that actor). I agree its not necessary to have that optimization for now. We may be proved wrong if client harnesses wont maintain persistent connections too often as we think they will.

keithmattix:
I've done another pass to clean up the commits to make this a bit more reviewable. If we're all on the same page regarding the points in https://github.com/agent-substrate/substrate/pull/715#issuecomment-5208366275, this should be ready for review.

/cc @bowei

bowei:
Re: caching of the resume -- we really need to discuss the semantics. Ideally, the server tells the consumer how long to cache. I want to avoid baking timeouts as constants into the client -- this presents problems for flexibility in how the caching works, including disabling it in certain cases. https://github.com/agent-substrate/substrate/issues/593

bowei:
I see that the code has the Network ext_proc, but there was a lot of discussion above how we don't need it at the moment.

Is it possible to get a conclusive comment here on whether or not we have settled on the approach? It's not clear at the moment from what I can tell from the discussion.

keithmattix:
> Re: caching of the resume -- we really need to discuss the semantics

If we take the approach currently in the PR, we don't need to cache any resumes; we only call it once. @LiorLieberman was suggesting alternative approaches that would reduce the need for network extproc.

> Is it possible to get a conclusive comment here on whether or not we have settled on the approach?

@LiorLieberman correct me if I'm wrong, but my understanding from our discussion (above and synchronously) is that we're aligned on using network ext proc for now for a couple of reasons:

1. Removes the need for any consideration of caching the resume
2. Reduces ext proc hops (it's loopback but still slower than in process) as the alternate architecture would require two hops in the common HTTP case

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -25,42 +25,91 @@ package ingress
...
+	// same move (see network.go).
+	SubstrateMetadataNamespace = "dev.substrate"
+
+	// AuthorityFilterStateKey is the filter-state object key holding a
+	// CONNECT request's (or, for plain ingress, any request's) :authority,
+	// set by xds.go's authorityFilterStateFilter and shared with the
+	// upstream internal connection so main_internal's HTTP ext_proc leg can
+	// read it back -- via AuthorityFilterStateAttribute -- across the
+	// internal-listener hop that dynamic metadata did not survive.
+	AuthorityFilterStateKey = "dev.substrate.authority"
```
EItanya: The official DNS name is currently `ate.dev`. Do we want to mirror that?

cmd/atenet/internal/router/cmd.go
```diff
@@ -50,10 +50,13 @@ func NewRouterCmd() *cobra.Command {
...
 	cmd.Flags().IntVar(&cfg.HttpPort, "port-http", 8080, "TCP port for workload traffic entering through the Envoy Router")
+	cmd.Flags().IntVar(&cfg.ConnectPort, "port-connect", 8081, "TCP port for CONNECT-tunneled traffic entering through the router dataplane")
+	cmd.Flags().IntVar(&cfg.ConnectTLSPort, "port-connect-tls", 8443, "TCP port for CONNECT-tunneled traffic entering through the router dataplane over TLS")
 	cmd.Flags().IntVar(&cfg.XdsPort, "port-xds", 18000, "TCP port listening for the xDS dynamic Envoy connections")
 	cmd.Flags().IntVar(&cfg.ExtprocPort, "port-extproc", 50051, "Listen port for the External Processing (ext_proc) server the dataplane calls")
 	cmd.Flags().StringVar(&cfg.ExtprocAddr, "extproc-address", "127.0.0.1", "Host IP or address of the External Processing (ext_proc) server")
-	cmd.Flags().StringVar(&cfg.EnvoyImage, "envoy-image", "envoyproxy/envoy:v1.30-latest", "Image URI used for dynamically launched router instances")
+	cmd.Flags().IntVar(&cfg.NetworkExtprocPort, "port-network-extproc", 50052, "Listen port for the network (L4/TCP) External Processing server, serving CONNECT-tunneled traffic separately from --port-extproc's HTTP server")
```
EItanya: Any particular reason why this is on a different port?

cmd/atenet/internal/router/config.go
```diff
@@ -97,8 +97,21 @@ type routerConfig struct {
...
+	ConnectPort    int
+	ConnectTLSPort int
 	EnvoyCertPath  string
 
+	// NetworkExtprocPort is the listen port for the network (L4) External
+	// Processing gRPC server, separate from ExtprocPort's HTTP one: it serves
+	// CONNECT-tunneled TCP traffic reinjected through main_internal, which has
+	// no HTTP semantics for Envoy's HTTP ext_proc filter to hook into. See
+	// ingress.NetworkExtProcServer.
+	NetworkExtprocPort int
```
EItanya: Same question about separate port

cmd/atenet/internal/router/cmd.go
```diff
@@ -50,10 +50,13 @@ func NewRouterCmd() *cobra.Command {
 	cmd.Flags().StringVar(&cfg.Kubeconfig, "kubeconfig", "", "Absolute path to the kubeconfig configuration file")
 	cmd.Flags().StringVar(&cfg.AteapiAddr, "ateapi-address", "k8s:///api.ate-system.svc:443", "gRPC dial target for the cluster ateapi Control instance.")
 	cmd.Flags().IntVar(&cfg.HttpPort, "port-http", 8080, "TCP port for workload traffic entering through the Envoy Router")
+	cmd.Flags().IntVar(&cfg.ConnectPort, "port-connect", 8081, "TCP port for CONNECT-tunneled traffic entering through the router dataplane")
+	cmd.Flags().IntVar(&cfg.ConnectTLSPort, "port-connect-tls", 8443, "TCP port for CONNECT-tunneled traffic entering through the router dataplane over TLS")
```
EItanya: --port-connect-tls and --port-https both default to 8443

cmd/atenet/internal/router/ingress/network.go
```diff
@@ -0,0 +1,209 @@
...
+	errChan := make(chan error, 1)
+	go func() {
+		if err := grpcServer.Serve(lis); err != nil {
+			errChan <- err
+		}
+	}()
+
+	select {
+	case <-ctx.Done():
+		grpcServer.GracefulStop()
```
EItanya: According to AI:

```
GracefulStop() can block indefinitely on long-lived CONNECT streams, bypassing the existing bounded drain logic
```

internal/atunnel/ingress.go
```diff
@@ -103,12 +112,26 @@ func NewServer(cfg Config) (*Server, error) {
...
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(cfg.Upstream)
 	transport := http.DefaultTransport.(*http.Transport).Clone()
-	proxy.Transport = transport
-	proxy.ErrorHandler = func(w http.ResponseWriter, r *http.Request, err error) {
-		slog.WarnContext(r.Context(), "atunnel upstream request failed", slog.Any("err", err))
-		http.Error(w, "bad gateway", http.StatusBadGateway)
+	proxy := &httputil.ReverseProxy{
+		Rewrite: func(pr *httputil.ProxyRequest) {
```
EItanya, line 125: According to AI, worth double checking:

```
replacing NewSingleHostReverseProxy with Rewrite strips all Forwarded/X-Forwarded-* headers. Call pr.SetXForwarded() to preserve prior behavior
```

EItanya (commented):
LGTM overall, just some questions

cmd/atenet/internal/router/ingress/ingress.go
keithmattix: Oh yes good catch; it'll be reversed due to protobuf/envoy convention though

cmd/atenet/internal/router/cmd.go
keithmattix: I think I just made it completely separate for testing and experimentation and forgot to clean it up. Good catch though; we can definitely put this as the same port

cmd/atenet/internal/router/ingress/network.go
keithmattix: Will get fixed by the single port refactor

internal/atunnel/ingress.go
keithmattix: Good catch AI 😄

keithmattix: Fixed

cmd/atenet/internal/router/cmd.go
keithmattix: My bad; moved to 8444

cmd/atenet/internal/router/config.go
keithmattix: Collapsed to single port

LiorLieberman:
> Lior Lieberman (@LiorLieberman) correct me if I'm wrong, but my understanding from our discussion (above and synchronously) is that we're aligned on using network ext proc for now for a couple of reasons:
> 
> 1. Removes the need for any consideration of caching the resume
> 2. Reduces ext proc hops (it's loopback but still slower than in process) as the alternate architecture would require two hops in the common HTTP case

Sorry it took a day to respond. My understanding was that we are aligned on the 2nd approach from whats proposed in https://github.com/agent-substrate/substrate/pull/715#issuecomment-5208366275. 

We are aligned on not needing to cache, since its a minor optimization, but I I would rather do

CONNECT ext_proc -> HTTP ext_proc
_________________-> no network ext_proc (for non-http traffic).

> 2. Reduces ext proc hops (it's loopback but still slower than in process) as the alternate architecture would require two hops in the common HTTP case

Since this would only be 2 extproc once when connection is established (and only for http calls, for non-http there wont be 2 callouts) - I thought we said its very minimal and a super minor optimization can be to cache. 

are there any more reasons that I missed?

keithmattix:
@LiorLieberman responded in DMs: doing the extproc on the outer and inner CONNECTs just feels brittle to me; we're assuming `ResumeActor` is always going to be cheap to call multiple times in sequence and that just feels unnecessary. All of the complexity of needing to support network extproc is downstream of 1. not exposing Resume() as a callout on ate-api and 2. Envoy implementation details of not being able to do HTTP things without an HTTPConnectionManager. Happy to move to network_extauthz but that would need further envoy work to copy over the enhancements @botengyao is working on for network extproc

LiorLieberman:
For posterity, we have talked offline -- probably easier to start without supporting TCP traffic to actors. And add it laters as we see concrete usecases. This will allow us to only do extProc / resume call for the inner http stream. And revisit net extProc or other approaches to enable tcp down the road.

bowei:
+1 to Lior's comment -- I think the support for TCP is strictly additive and it would be good to get the HTTP connect in which we have a good idea on how it will work. Great that we explored all of the different options when we do add the TCP support.

LiorLieberman:
> +1 to Lior's comment -- I think the support for TCP is strictly additive and it would be good to get the HTTP connect in which we have a good idea on how it will work. Great that we explored all of the different options when we do add the TCP support.

Thats Keith idea - so credit goes to keith.

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -25,42 +25,81 @@ package ingress
...
+	// namespace xds.go's
+	// ORIGINAL_DST cluster reads from for Envoy. The exact same struct also
+	// reaches agentgateway, which surfaces ext_proc's dynamic_metadata field
+	// verbatim to CEL as `extproc` (see configmap.yaml's dynamic backend
+	// `target` expression) -- one response payload, two dataplanes reading it
+	// their own way. Reuses Envoy's own
+	// envoy.filters.listener.original_dst listener filter's namespace
+	// instead of inventing one; see xds.go's OriginalDstClusterName doc for
+	// why that does not collide with the listener filter's own, unrelated
+	// use of it.
```
LiorLieberman:
```suggestion
	// ORIGINAL_DST cluster reads from for the dataplane.
```

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -25,42 +25,81 @@ package ingress
...
+	OriginalDstMetadataKey = "envoy.filters.listener.original_dst"
+	// OriginalDstAddressKey is the field within OriginalDstMetadataKey
+	// carrying the resolved worker atunnel address (IP:443). Reuses the same
+	// field name (rather than "address") that the
+	// envoy.filters.listener.original_dst listener filter itself reads for
+	// its own, unrelated EnvoyInternal fallback path.
```
LiorLieberman: lots of "do X rather than Y... reuse Z" comments, mind clearing them a little? relevant for all parts of the code

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -80,10 +119,36 @@ func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestM
...
-	actorRef, err := parseActorRef(md.Host)
+	// The actor is always resolved from the forwarded filter-state authority
+	// attribute, never the Host/:authority header directly: that header is
+	// only reliable for the plain ingress_http/ingress_https listeners
+	// (which populate the filter state themselves via xds.go's
+	// authorityFilterStateFilter). For CONNECT-tunneled traffic reinjected
+	// through main_internal, the tunneled protocol's own :authority is
+	// unrelated to the actor's DNS name; the authoritative value is whatever
+	// connect_terminate captured at CONNECT time and shared with upstream via
+	// filter state.
```
LiorLieberman: (thinking out loud) should we describe main_internal and connect_terminate rather than using names?  e.g "the main (internal?) listener" and "the listener where connect is terminated"

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -80,10 +119,36 @@ func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestM
...
+		// Authority is invalid, respond with 404.
+		return extproc.Result{}, invalidHostErr(authority, err)
+	}
+
+	// The port to reach on the actor itself travels in the same authority:
+	// for CONNECT-tunneled traffic it's the arbitrary port the client asked
+	// for (e.g. ":9090"), and for plain ingress_http/ingress_https it's
+	// absent, defaulting to the actor's normal port 80. atunnel's Server
+	// can't learn this any other way -- its Config.Upstream is fixed for its
+	// whole lifetime -- so it's forwarded via atunnel.TargetPortHeader.
```
LiorLieberman:
```suggestion
	// The port to reach on the actor itself travels in the same authority.
	// For CONNECT-tunneled traffic it's the arbitrary port the client asked
	// for, and for plain ingress_http/ingress_https it's
	// absent, defaulting to the actor's default port. 
```

unrelated to the comment but the comment reminded me - do we support 443 AND 80 right now without connect or only 80?

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -80,10 +119,36 @@ func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestM
...
+	targetPort := 80
```
LiorLieberman: take from a const instead?

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -80,10 +119,36 @@ func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestM
...
+	if _, portStr, err := net.SplitHostPort(authority); err == nil {
+		if p, err := strconv.Atoi(portStr); err == nil && p > 0 && p <= 65535 {
```
LiorLieberman: whats the reason to validate port correctness here manually, dont we have any helper to do it

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -123,28 +188,64 @@ func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestM
...
-	// longer DNATs pod-IP:80 to the actor, so the router dials :443 and the
-	// ORIGINAL_DST cluster's upstream TLS context presents the router's
-	// podidentity client cert (see buildOriginalDstCluster and
-	// buildUpstreamTransportSocket).
-	// TODO(bowei) -- handle more than port 80 on the actor.
+	// listens on :443 (mTLS) and forwards to targetPort on the actor. The
+	// worker no longer DNATs pod-IP:80 to the actor, so the router dials :443
+	// and the ORIGINAL_DST cluster's upstream TLS context presents the
+	// router's podidentity client cert (see xds.go's buildOriginalDstCluster
+	// and buildUpstreamTransportSocket).
```
LiorLieberman:
```suggestion
	// listens on :443 (mTLS) and forwards to targetPort on the actor. The router
	// dials :443 and the ORIGINAL_DST cluster's upstream TLS context presents
	// the router's podidentity client cert (xds.go - buildOriginalDstCluster
	// and buildUpstreamTransportSocket).
```

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -123,28 +188,64 @@ func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestM
...
 	mutation := &extprocv3.HeaderMutation{}
-	addRoutingMutations(targetAddr, md.Host, h.routeViaAuthority, mutation)
+	// atunnel picks which port on the actor to reach from this header (the
+	// CONNECT authority's port, or 80 for plain ingress -- see targetPort
+	// above). For envoy mode this duplicates what xds.go's buildRoutes
+	// derives declaratively from OriginalDstPortKey via
+	// %DYNAMIC_METADATA(...)% -- harmless, since both compute the same
+	// value, and the route's OVERWRITE_IF_EXISTS_OR_ADD wins last.
+	// Agentgateway has no equivalent of that route-level mechanism, so it
+	// depends on this header mutation being set directly.
```
LiorLieberman: can we trim agent-lang a bit here? like "...harmless" etc. Lets focus on what we want to say. We may not even need a comment here

cmd/atenet/internal/router/cmd.go
```diff
@@ -49,6 +49,8 @@ func NewRouterCmd() *cobra.Command {
 	cmd.Flags().StringVar(&cfg.Kubeconfig, "kubeconfig", "", "Absolute path to the kubeconfig configuration file")
 	cmd.Flags().StringVar(&cfg.AteapiAddr, "ateapi-address", "k8s:///api.ate-system.svc:443", "gRPC dial target for the cluster ateapi Control instance.")
 	cmd.Flags().IntVar(&cfg.HttpPort, "port-http", 8080, "TCP port for workload traffic entering through the Envoy Router")
+	cmd.Flags().IntVar(&cfg.ConnectPort, "port-connect", 8081, "TCP port for CONNECT-tunneled traffic entering through the router dataplane")
+	cmd.Flags().IntVar(&cfg.ConnectTLSPort, "port-connect-tls", 8444, "TCP port for CONNECT-tunneled traffic entering through the router dataplane over TLS. Defaults away from --port-https's 8443 since both listeners are commonly enabled at once")
```
LiorLieberman: why not 8443? not sure I am follwoing

cmd/atenet/internal/router/config.go
```diff
@@ -95,6 +95,12 @@ type routerConfig struct {
 	StatusPort     int
 	HealthInterval time.Duration
 	HttpsPort      int
+	// ConnectPort and ConnectTLSPort are the plaintext and TLS listener ports
+	// for CONNECT-tunneled traffic (arbitrary-port ingress — see xds.go's
+	// connect_terminate listeners). Non-positive disables the corresponding
+	// listener.
+	ConnectPort    int
```
LiorLieberman:
```suggestion
	ConnectPlainTextPort    int
```

cmd/atenet/internal/router/config.go
```diff
@@ -95,6 +95,12 @@ type routerConfig struct {
 	StatusPort     int
 	HealthInterval time.Duration
 	HttpsPort      int
+	// ConnectPort and ConnectTLSPort are the plaintext and TLS listener ports
+	// for CONNECT-tunneled traffic (arbitrary-port ingress — see xds.go's
+	// connect_terminate listeners). Non-positive disables the corresponding
+	// listener.
```
LiorLieberman:
```suggestion
	// ConnectPlainTextPort and ConnectTLSPort are the plaintext and TLS listener
	// ports for CONNECT-tunneled traffic.
	// Non-positive disables the corresponding listener.
```

cmd/atenet/internal/router/dataplane.go
```diff
@@ -31,10 +31,6 @@ type dataplaneHealthCheck struct {
 	expectedBody string
 }
 
-func (r atenetRouter) routeViaAuthority() bool {
-	return r == atenetRouterAgentgateway
-}
-
```
LiorLieberman, line 37: why?

LiorLieberman: cause now both route via authority? is it not only if its connect and not the ingress tls/cleatext ports (80+443?). 

can you elaborate?

cmd/atenet/internal/router/xds.go
```diff
@@ -78,10 +87,23 @@ const (
...
+	MainInternalName   = "main_internal"
+
+	// dynamicMetadataPortFormat is the %DYNAMIC_METADATA(...)% header-value
+	// command operator (see buildRoutes) that derives atunnel.TargetPortHeader
+	// from ingress.OriginalDstMetadataKey/ingress.OriginalDstPortKey.
+	dynamicMetadataPortFormat = "%DYNAMIC_METADATA(" + ingress.OriginalDstMetadataKey + ":" + ingress.OriginalDstPortKey + ")%"
+
+	// httpExtProcFilterName is envoy.filters.http.ext_proc's own well-known
+	// name, used as the HttpFilter.Name in buildHcm.
+	httpExtProcFilterName = "envoy.filters.http.ext_proc"
```
LiorLieberman, line 106: dont we have it somewhere in a const? if not, we should. not a blocking comment though

cmd/atenet/internal/router/xds.go
```diff
@@ -133,8 +155,10 @@ type XdsServer struct {
 
 	mu sync.Mutex
 
-	httpsPort int
-	certPath  string
+	httpsPort      int
+	connectPort    int
```
LiorLieberman: s/connectPort/connectPlainTextProt/g

internal/atunnel/ingress.go
```diff
@@ -103,12 +112,27 @@ func NewServer(cfg Config) (*Server, error) {
...
+			pr.SetURL(cfg.Upstream)
+			// Retain the actor's stable mesh hostname rather than the
+			// upstream's, matching NewSingleHostReverseProxy's default
+			// behavior.
+			pr.Out.Host = pr.In.Host
+
+			port := pr.In.Header.Get(TargetPortHeader)
+			pr.Out.Header.Del(TargetPortHeader)
+			if p, err := strconv.Atoi(port); err == nil && p > 0 && p <= 65535 {
+				pr.Out.URL.Host = net.JoinHostPort(cfg.Upstream.Hostname(), strconv.Itoa(p))
```
LiorLieberman: this func is repeating in a few places, worth putting as a helper

manifests/ate-install/atenet-router.yaml
LiorLieberman, line 1: are teher any envoy CONNECT specific changes we need to do? could not find any change beyond this internal listener

LiorLieberman (dismissed):
LGTM modulo a few small comments. 

Can you please also update the docs / diagrams / guides to clarify what clients can do now to support arbitraty ports etc? 

Plus maybe a sentence about the TCP limitation and an invite to respond on TCP (or non-http) actor usecases if they need it


Thanks for the persistence here!

cmd/atenet/internal/router/cmd.go
keithmattix: 8443 already in use for Https

manifests/ate-install/atenet-router.yaml
keithmattix: Nope this is it

internal/atunnel/ingress.go
keithmattix: Done

cmd/atenet/internal/router/xds.go
keithmattix: Done

cmd/atenet/internal/router/xds.go
keithmattix: This is the const declaration

cmd/atenet/internal/router/dataplane.go
keithmattix: I think John added this in the early days of agentgateway; now everything routes via metadata vs. authority

LiorLieberman (dismissed):
Sweet, thanks @keithmattix !

Can you keep the non-http protocols issue opened and update with the status after this merges?

looks like one last conflict

PR #718, opened 2026-08-03
[DO NOT SUBMIT] take the Atespace lock across actor creation

### LINKED ISSUE #620: [P2] DeleteAtespace TOCTOU race — concurrent CreateActor can orphan actors in a deleted atespace
**Severity:** P2 (actor orphaned; no API path to resume or delete it)
**Component:** Control Plane — `cmd/ateapi/internal/store/ateredis/ateredis.go`
**Confirmed:** Yes — code-level analysis

---

## Summary

`DeleteAtespace` checks for actor existence with `ListActors`, then deletes the atespace
key in two separate Redis operations. A concurrent `CreateActor` call can succeed in
the window between the emptiness check and the deletion, creating an actor whose
atespace no longer exists. The actor record persists in Redis but its atespace is gone,
making it invisible to `ListActors` (which scans by atespace key) and permanently
unreachable.

---

## Impact

- Actors orphaned in a deleted atespace accumulate as invisible Redis keys.
- `ResumeActor` for an orphaned actor fails because there is no atespace validation at
  resume time — but `ListActors` will never return the actor without a valid atespace.
- `DeleteActor` requires knowing the actor exists; without `ListActors` visibility it
  cannot be targeted.
- The orphaned actor consumes a worker slot (if it was RUNNING at the time of the race).

---

## Root Cause

**File:** `cmd/ateapi/internal/store/ateredis/ateredis.go` lines 183–214

```go
func (s *Persistence) DeleteAtespace(ctx context.Context, name string) (*ateapipb.Atespace, error) {
    // Step 1: Read atespace record (not watched)
    currentVal, err := s.rdb.Get(ctx, dbKey).Bytes()
    ...
    // Step 2: Check emptiness — separate Redis call, NOT inside a WATCH
    actors, _, err := s.ListActors(ctx, name, 1, "")
    if len(actors) > 0 {
        return nil, store.ErrFailedPrecondition  // "not empty"
    }
    // RACE WINDOW: CreateActor can succeed here
    // Step 3: Delete atespace key
    if err := s.rdb.Del(ctx, dbKey).Err(); err != nil { ... }
}
```

And `CreateActor` (`create_actor.go:54-59`) checks `AtespaceExists` and then creates
the actor — but after the Del at step 3, `AtespaceExists` would fail for new callers.
The concurrent caller that is already past the `AtespaceExists` check (step 2) proceeds
to `CreateActor` successfully, creating an actor with no atespace backing it.

---

## Steps to Reproduce

```bash
# Setup: create an atespace with no actors
kubectl ate create atespace race-test

# Terminal 1: Loop creating and deleting an actor in the atespace
while true; do
  kubectl ate create actor orphan-candidate -a race-test \
    --template ate-demo-counter/counter 2>/dev/null
  sleep 0.1
done

# Terminal 2: Loop trying to delete the atespace
while true; do
  kubectl ate delete atespace race-test 2>/dev/null
  kubectl ate create atespace race-test 2>/dev/null
done

# After a few seconds, inspect Valkey directly for orphaned actor keys
kubectl exec -n ate-system valkey-cluster-0 -- \
  valkey-cli SCAN 0 MATCH "actor:race-test:*" COUNT 1000
# You may find actor keys with no corresponding atespace key
```

---

## Expected Behavior

`DeleteAtespace` should be atomic with its emptiness check. The implementation should
use a Redis `WATCH` on both the atespace key and the actors scan, or use a Lua script
that atomically checks for actor presence and deletes the key in one transaction.

---

## Suggested Fix

Wrap the check-and-delete in a Redis `WATCH` transaction:

```go
err := s.rdb.Watch(ctx, func(tx *redis.Tx) error {
    actors, _, err := s.ListActors(ctx, name, 1, "")
    if err != nil { return err }
    if len(actors) > 0 { return store.ErrFailedPrecondition }
    _, err = tx.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
        pipe.Del(ctx, dbKey)
        return nil
    })
    return err
}, dbKey)
```

This ensures that if any key changes (another actor is created) between the scan and
the delete, the transaction is aborted. Note that watching while using SCAN is not
fully atomic in a Redis Cluster, so a stronger fix would use a Lua script that atomically
checks for actor key existence and deletes the atespace key.

HavenXia:
Acquire an Atespace lock across the actor creation so concurrent DeleteAtespace cannot orphan the new actor.

This fixes #620.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/create_actor.go
+++ b/cmd/ateapi/internal/controlapi/create_actor.go
@@ -87,6 +87,17 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
 	atespace := in.GetMetadata().GetAtespace()
 	name := in.GetMetadata().GetName()
 
+	// Acquire an Atespace lock across the actor creation so concurrent DeleteAtespace cannot orphan the new actor.
+	atespaceLock, err := s.persistence.AcquireLock(ctx, "lock:atespace:"+atespace)
+	if errors.Is(err, store.ErrLockConflict) {
+		return nil, status.Error(codes.Aborted, "another operation is using this Atespace")
+	}
+	if err != nil {
+		return nil, fmt.Errorf("while locking Atespace: %w", err)
+	}
+	defer atespaceLock.Close()
+	ctx = atespaceLock.Context()
+
 	// The atespace must already exist.
 	exists, err := s.persistence.AtespaceExists(ctx, atespace)
 	if err != nil {
--- a/cmd/ateapi/internal/controlapi/create_actor_test.go
+++ b/cmd/ateapi/internal/controlapi/create_actor_test.go
@@ -261,3 +261,31 @@ func TestCreateActor_RejectsSnapshotWithExternalVolumes(t *testing.T) {
 		t.Fatalf("CreateActor status = %v, want FailedPrecondition", status.Code(err))
 	}
 }
+
+func TestCreateActor_AbortsWhileAtespaceLocked(t *testing.T) {
+	ns := namespaceForTest("ns-create-atespace-lock")
+	tc := setupTest(t, ns)
+	defer tc.cleanup()
+	createTemplate(t, tc, ns)
+
+	lock, err := tc.persistence.AcquireLock(context.Background(), "lock:atespace:"+testAtespace)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


HavenXia:
Close as this is not a good idea to lock on atespace during creation, we expect a high cardinality for actor creation in a same atespace.

PR #720, opened 2026-08-03
Close the current-netns handles in ateomnet.

HavenXia:
I think it need to be closed when looking at what `netns` is in https://github.com/vishvananda/netns.

- [x] Tests pass

### DIFF
--- a/internal/ateomnet/net.go
+++ b/internal/ateomnet/net.go
@@ -394,6 +394,8 @@ func CreateNetNSWithoutSwitching(name string) (netns.NsHandle, error) {
 	if err != nil {
 		return -1, fmt.Errorf("while getting current netns: %w", err)
 	}
+	// Registered before the restoring defer below since deferred calls are LIFO.
+	defer curNetNS.Close()
 	defer func() {
 		if err := netns.Set(curNetNS); err != nil {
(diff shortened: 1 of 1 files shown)


internal/ateomnet/net.go
haiyanmeng, line 405: Add a comment to make it clear that the order of these two `defer` matters

internal/ateomnet/net.go
```diff
@@ -419,6 +420,7 @@ func NetNSDo(ctx context.Context, targetNS netns.NsHandle, do func(context.Conte
 	if err != nil {
 		return fmt.Errorf("while getting current netns: %w", err)
 	}
+	defer curNetNS.Close()
 	defer func() {
 		if err := netns.Set(curNetNS); err != nil {
```
haiyanmeng, line 427: Add a comment to make it clear that the order of these two `defer` matters

internal/ateomnet/net.go
HavenXia: added, thank you.

internal/ateomnet/net.go
HavenXia: added, thank you.

haiyanmeng (approved):
/lgtm
/approve

Thanks!

PR #726, opened 2026-08-04
add hack tool to run Substrate against a locally-built runsc

milantracy:
atelet fetches runsc at runtime from a url + sha256 on a cluster-scoped SandboxConfig, so trying an unreleased build needs no image rebuild — but there was no supported way to stage one, and the resolution path it would rely on was untested.

the tool is modelled on hack/microvm-assets/: deploy.sh hashes a local runsc, stages it (rustfs on kind, GCS on GKE), applies a SandboxConfig pinning it, and optionally repoints a WorkerPool. gvisor-default is left untouched; pools opt in via sandboxConfigName. 

So an unreleased gVisor feature can be tested in a cluster without hand-staging binaries or hand-writing a SandboxConfig.

### DIFF
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -225,7 +225,7 @@ spec:
 
 ### Running a locally-built runsc
 
-Because the binary is fetched at runtime rather than baked into the worker image, testing an unreleased runsc is a matter of staging the binary into the cluster's object store and pointing a `SandboxConfig` at its URL + `sha256`. See [`hack/gvisor-hack/`](../hack/gvisor-hack/) for scripts that do this and leave `gvisor-default` untouched. Note that snapshots are version-locked to the runsc that wrote them, so a custom build cannot restore checkpoints taken with the released one.
+Because the binary is fetched at runtime rather than baked into the worker image, testing an unreleased runsc is a matter of staging the binary into the cluster's object store and pointing a `SandboxConfig` at its URL + `sha256`. See [`hack/gvisor-hack/`](../hack/gvisor-hack/) for scripts that do this. By default they add a separate named config and leave `gvisor-default` untouched; opting in to `MAKE_DEFAULT` instead clears `spec.default` on the incumbent so the new config becomes the cluster default. Note that snapshots are version-locked to the runsc that wrote them, so a custom build cannot restore checkpoints taken with the released one.
 
 ### Micro-VM SandboxConfig
 
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -223,6 +223,10 @@ spec:
         sha256: "1ba2366ae2efceba166046f51a4104f9261c9cb72c6db8f5b3fe2dc57dea86b9"
 ```
 
+### Running a locally-built runsc
+
+Because the binary is fetched at runtime rather than baked into the worker image, testing an unreleased runsc is a matter of staging the binary into the cluster's object store and pointing a `SandboxConfig` at its URL + `sha256`. See [`hack/gvisor-hack/`](../hack/gvisor-hack/) for scripts that do this and leave `gvisor-default` untouched. Note that snapshots are version-locked to the runsc that wrote them, so a custom build cannot restore checkpoints taken with the released one.
+
 ### Micro-VM SandboxConfig
 
 A `microvm` `SandboxConfig` supplies the [Kata Containers](https://katacontainers.io/) + [Cloud Hypervisor](https://www.cloudhypervisor.org/) toolchain instead of `runsc`. Each architecture must define the full asset set — `kata-shim`, `cloud-hypervisor`, `virtiofsd`, `kata-kernel`, `kata-image`, and `kata-config` — which a `ValidatingAdmissionPolicy` enforces at apply time. Worker pods for a micro-VM pool require `/dev/kvm` and nested-virtualization-capable nodes labeled `ate.dev/sandboxClass=microvm` (the controller adds the device mount and node placement automatically).
--- a/hack/gvisor-hack/deploy.sh
+++ b/hack/gvisor-hack/deploy.sh
@@ -66,6 +66,14 @@ if [[ -n "${WORKERPOOL}" && -z "${WORKERPOOL_NAMESPACE}" ]]; then
   exit 1
 fi
 
+# NAME becomes a Kubernetes object name. Check it here so a typo costs a second
+# rather than an upload, and so it cannot be the thing that fails the apply after
+# the incumbent default has already been demoted.
+if [[ ${#NAME} -gt 253 ]] || ! [[ "${NAME}" =~ ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ ]]; then
+  echo "error: NAME=${NAME} is not a valid Kubernetes object name (DNS-1123 subdomain)" >&2
+  exit 1
+fi
+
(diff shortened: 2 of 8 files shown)


hack/gvisor-hack/deploy.sh
```diff
@@ -0,0 +1,154 @@
...
+  log "Staging to in-cluster rustfs bucket ${BUCKET_NAME}..."
+  RUNSC="${RUNSC}" OBJECT="${OBJECT}" BUCKET="${BUCKET_NAME}" KUBECTL_CONTEXT="${KUBECTL_CONTEXT}" \
+    hack/gvisor-hack/stage-to-rustfs.sh
+else
+  log "Staging to gs://${BUCKET_NAME}/${OBJECT}..."
+  RUNSC="${RUNSC}" OBJECT="${OBJECT}" BUCKET="${BUCKET_NAME}" \
+    hack/gvisor-hack/stage-to-gcs.sh
+fi
+
+# Two defaults for one sandbox class make every actor launch fail to resolve its
```
BenTheElder, line 161: 🤖 **should-fix** 🟡 – Demoting before applying swaps a two-default cluster for a zero-default one, and if the apply on line 130 fails the cluster stays that way. `TestResolveSandboxAssetsErrors` pins both as failures ("multiple default SandboxConfigs" and "no default SandboxConfig"), so every gVisor pool relying on the default stops resolving assets either way — but the two-default window closes on the next patch, while a failed apply leaves zero defaults until someone notices.

That is reachable: `NAME` is an unvalidated env var, so a name the API server rejects gets you there, as does any transient apply failure. The script exits on the apply error without mentioning that it just cleared `gvisor-default`, and the README's Reverting section only covers the success path.

A server-side dry run before the demote loop would close it — the ValidatingAdmissionPolicy runs under `--dry-run=server`, so the config is known good before the incumbent is touched.

docs/api-guide.md
```diff
@@ -223,6 +223,10 @@ spec:
         sha256: "1ba2366ae2efceba166046f51a4104f9261c9cb72c6db8f5b3fe2dc57dea86b9"
 ```
 
+### Running a locally-built runsc
+
+Because the binary is fetched at runtime rather than baked into the worker image, testing an unreleased runsc is a matter of staging the binary into the cluster's object store and pointing a `SandboxConfig` at its URL + `sha256`. See [`hack/gvisor-hack/`](../hack/gvisor-hack/) for scripts that do this and leave `gvisor-default` untouched. Note that snapshots are version-locked to the runsc that wrote them, so a custom build cannot restore checkpoints taken with the released one.
```
BenTheElder: 🤖 **nit** 🟢 – "leave `gvisor-default` untouched" only holds for the default path. With `MAKE_DEFAULT=true`, `deploy.sh` patches `spec.default: false` onto the incumbent — `gvisor-default` in a stock cluster — which is exactly the case a reader most wants warned about. Worth saying the scripts leave it untouched *unless* you opt into making the hack config the cluster default.

milantracy: The README and `docs/api-guide.md` now say `gvisor-default` is
  untouched *by default*, and that `MAKE_DEFAULT` is the one path that clears
  its `spec.default`.

 Added an "If the script failed partway" section: the rollback behavior

hack/gvisor-hack/deploy.sh
milantracy: zero defaults doesn't self-heal the way two defaults does.

  The old order was: demote the incumbent, then apply. Anything that went wrong
  in the apply left the cluster with no default at all.

  The new order is:

  1. Validate `NAME` as a DNS-1123 subdomain.
  2. `kubectl apply --dry-run=server` the rendered config.
  3. Upload the binary.
  4. Demote the incumbent.
  5. Apply for real.

pkg/api/v1alpha1/sandboxconfig_validation_test.go
```diff
@@ -194,3 +195,94 @@ func TestSandboxConfigValidation(t *testing.T) {
...
+
+// TestGvisorHackTemplate renders the hack/gvisor-hack SandboxConfig template and
+// applies it to a real API server carrying the shipped CRD and
+// ValidatingAdmissionPolicy. The dev workflow's whole premise is that the
+// rendered object is accepted, so a schema or policy change that invalidates the
+// template should fail here rather than in someone's kind cluster. It also
+// checks that the template and its renderer agree on the placeholder set: a
+// placeholder deploy.sh does not substitute would reach the API server as a
+// literal "${...}".
+func TestGvisorHackTemplate(t *testing.T) {
```
BenTheElder, line 224: This seems unnecessary and unrelated to testing the validation? the hack script probably doesn't need to be e2e tested on every PR / change. it's not load-bearing.

hack/gvisor-hack/deploy.sh
```diff
@@ -0,0 +1,154 @@
...
+# assets, so demote the incumbent BEFORE applying rather than after.
+if [[ "${MAKE_DEFAULT}" == "true" ]]; then
+  log "Clearing spec.default on other gvisor SandboxConfigs..."
+  while read -r sc_name sc_class sc_default; do
+    [[ -z "${sc_name}" ]] && continue
+    if [[ "${sc_name}" != "${NAME}" && "${sc_class}" == "gvisor" && "${sc_default}" == "true" ]]; then
+      log "   demoting ${sc_name}"
+      "${KUBECTL[@]}" patch sandboxconfig "${sc_name}" --type=merge -p '{"spec":{"default":false}}'
+    fi
+  done < <("${KUBECTL[@]}" get sandboxconfigs -o jsonpath='{range .items[*]}{.metadata.name}{" "}{.spec.sandboxClass}{" "}{.spec.default}{"\n"}{end}')
```
BenTheElder, line 121: 🤖 **should-fix** 🟡 – A failure here is silent, and the failure mode is the one the demotion exists to prevent. Process substitution's exit status is never collected — neither `errexit` nor `pipefail` covers it — so if `kubectl get sandboxconfigs` fails for any reason (transient API error, a context pointing at a cluster without the CRD, RBAC), the loop body simply never runs. The script then applies the new config with `spec.default: true` and the incumbent default is still standing.

`TestResolveSandboxAssetsErrors` in this same PR pins what that costs: two defaults for a class is `multiple default SandboxConfigs`, so *every* actor launch on that class fails to resolve assets, not just the pool the developer is working on. `MAKE_DEFAULT=true` is precisely the mode where they are touching shared cluster state.

Capturing first makes the failure loud, since assignment from command substitution does trip `errexit`:

```bash
configs="$("${KUBECTL[@]}" get sandboxconfigs -o jsonpath='...')"
while read -r sc_name sc_class sc_default; do
  ...
done <<<"${configs}"
```

hack/gvisor-hack/sandboxconfig.yaml.tmpl
```diff
@@ -0,0 +1,44 @@
...
+  # SandboxConfigs of the same class claim it.
+  default: ${DEFAULT}
+  assets:
+    # Single-arch on purpose. The sandboxconfig-assets ValidatingAdmissionPolicy
+    # requires a "runsc" asset for EVERY architecture listed, and atelet only
+    # fetches the block matching its node's GOARCH — so listing just the arch you
+    # built for is both valid and sufficient. Add a second block (and stage a
+    # second binary) for a mixed-arch cluster.
+    ${ARCH}:
+      runsc:
```
EtiennePerot, line 40: Should be `gvisor` not `runsc`

milantracy:
I will close this PR for now as we are moving to use shim than runsc now.

PR #735, opened 2026-08-04
imagecache: eviction engine — root set, EvictUnused, startup orphan recovery

igooch:
Phase 2 of #463 (image-cache GC + observability): this is the eviction engine itself — the root set, the eviction pass, and startup orphan recovery — assembled from the foundations merged in #650 (sizes/last-use), #656 (spec digest), #658 (retirement primitives), and #717 (record-first pull). What remains after this is the watermark loop + flags that turn it on, then metrics and e2e.

**Inert in production**: nothing calls `EvictUnused` yet — the watermark GC loop and its flags come in the next PR. The only behavior `New` gains is the startup orphan scan.

## What

- **`Store.InUse()`** — the root set: scans bundle overlay specs under the actors dir (opt-in via `WithActorsDir`) for image digests, layer dirs, and exact layer-set signatures. The exact-set rule roots the multi-arch twin record and digestless (pre-`imageDigest`) specs' records without making every subset image unevictable. Returns `(RootSet, error)`: a non-nil error means the root set may be incomplete, and deleting callers must do nothing.
- **`Store.EvictUnused(ctx, targetBytes, dryRun)`** — LRU eviction of unprotected images and the layers their removal leaves unreferenced. Min-age vetoes; per-victim re-checks under `hitMu` held exclusive against the cache-hit touch; two-phase deletion via `retireLayer`, with the slow `RemoveAll` outside all locks. If any layer of a deleted record must be kept (still referenced, rooted, fresh, or failed to retire), the record is atomically restored from bytes captured at listing time — a kept layer is never left unreachable.
- **`Store.RecoverOrphans`**, called once from `New` — reclaims layers no record references. Startup is the one moment this scan is race-free (no pull can be in flight); a failure never fails `New`.
- **Everything fails toward retention.** Both passes skip entirely, at ERROR naming the culprit, when their inputs can't be fully enumerated: an unreadable/undecodable image record, a garbled diffID inside a valid record, or an unenumerable root set (actors dir, bundles dir, or bundle spec). Refcounts and roots from partial data would retire layers that unread records still reference or running actors still mount — and for a long-running actor, the bundle spec is the only protection (its record mtime can be arbitrarily old, so min-age is no backstop). Missing dirs/specs remain legitimate emptiness, not errors.
- **Eviction never writes into the live pool.** Retired layers are credited from the size file that rides along with the rename-aside (O(1), read-only, consistent with `CacheSize`); the size-file backfill exists only on the `CacheSize` path, where it bumps the dir mtime rather than restoring it (restoring could rewind a concurrent reuse-touch).
- Per-item log lines are Debug behind an `Enabled` gate (level control via #677); pass summaries, evictions, and restores are Info.
- `internal/imagecache/README.md` now documents the GC: pass structure, protection rules, failure direction, startup recovery, and two-phase deletion, plus the on-disk layout entries (`size`, `.tmp-*`, `.rm-*`) that shipped with #650/#658.

## Notes for reviewers

- The gc tests reuse the helpers merged with #650 (`backdate`, `mustEnsure`) and #717's gated-registry harness — no duplicated symbols. The slow-pull gated test gains the eviction-veto assertion deferred from #717.
- Dry-run mutates nothing, including size-file backfills (`layerSizeReadOnly`).
- A `Store` built without `WithActorsDir` has an empty root set (deliberate for tools/tests).
- Carried to the wiring PR: pass `WithActorsDir` in atelet; handle `EvictUnused`'s error before any shortfall accounting (a gated pass returns zero stats plus an error and must not be logged as a shortfall); and surface skipped passes as a counter in the observability PR so a wedged node alerts rather than scrolling ERROR lines.

## Testing

- `go test -race ./internal/imagecache/`: every veto, shared-layer survival, both restore-on-keep shapes, orphan reclaimed at startup and ignored by the periodic pass, all enumeration gates — unreadable record (shared-layer scenario), undecodable record, garbled diffID, and the three unreadable root-set shapes (actors dir / bundles dir / spec, each against a rooted image with an arbitrarily old record) — dry-run purity, 25-iteration ensure-vs-evict race loop.
- Kind, via the integration branch that stacks the watermark loop on this PR (the loop and flags are **not** part of this PR): full gVisor e2e suite green; a planted backdated orphan layer (no size file) was reclaimed at atelet startup with the expected log lines, its size credited read-only from the retired dir.
- GKE (6 nodes), same integration branch: demo lifecycle + snapshot-continuity suites green; under aggressive watermarks the idle pool was evicted and subsequent actor resumes cold-pulled cleanly. No enumeration-gate false positives on either cluster.

### DIFF
--- a/internal/imagecache/gc.go
+++ b/internal/imagecache/gc.go
@@ -222,7 +222,10 @@ type evictionCandidate struct {
 //
 // Failed deletions are skipped, not fatal: the error return aggregates
 // them, but the pass continues to the next candidate (each retries next
-// pass). Concurrent passes are serialized.
+// pass). Concurrent passes are serialized. If the record enumeration is
+// incomplete (an unreadable or undecodable record), the pass is skipped
+// entirely: refcounts from partial data would retire layers the unread
+// records still reference.
 func (s *Store) EvictUnused(ctx context.Context, targetBytes int64, dryRun bool) (EvictStats, error) {
 	var stats EvictStats
 	s.evictMu.Lock()
@@ -231,8 +234,16 @@ func (s *Store) EvictUnused(ctx context.Context, targetBytes int64, dryRun bool)
 	roots := s.InUse()
 	cutoff := time.Now().Add(-s.minAge)
 
-	candidates, refcount, _, listErr := s.listEviction(roots, cutoff, &stats)
-	// listErr is deferred, not fatal: evict what was listed, report at the end.
+	candidates, refcount, complete, listErr := s.listEviction(roots, cutoff, &stats)
+	if !complete {
+		// A layer shared with an unread record would hit refcount zero and
+		// be retired while that record still names it. Fail the whole pass
+		// toward retention; the error names the records to repair or
+		// delete, and every later pass retries.
+		slog.ErrorContext(ctx, "Image cache eviction pass skipped: image records could not be fully enumerated",
+			slog.Any("err", listErr))
+		return stats, listErr
+	}
 	stats.Candidates = len(candidates)
 
 	slog.InfoContext(ctx, "Image cache eviction pass",
@@ -244,9 +255,6 @@ func (s *Store) EvictUnused(ctx context.Context, targetBytes int64, dryRun bool)
 		slog.Duration("min_age", s.minAge))
 
 	var errs []error
-	if listErr != nil {
-		errs = append(errs, listErr)
-	}
 	dbg := slog.Default().Enabled(ctx, slog.LevelDebug) // see InUse
 	var retired []string                                // renamed-aside dirs awaiting RemoveAll
 
@@ -529,7 +537,9 @@ func (s *Store) sweepOrphanLayers(ctx context.Context, roots RootSet, refcount m
 }
 
 // dryRunRetire reports what retireLayer would do, without renaming:
-// the same pre-flight checks, plus the size credit.
+// the same pre-flight checks, plus the size credit. Sizing is read-only —
+// even the size-file backfill would break dry-run's mutate-nothing
+// contract.
 func (s *Store) dryRunRetire(hex string, cutoff time.Time) (int64, retireStatus) {
 	dir := filepath.Join(s.layersDir(), hex)
 	fi, err := os.Stat(dir)
@@ -539,7 +549,7 @@ func (s *Store) dryRunRetire(hex string, cutoff time.Time) (int64, retireStatus)
 	if fi.ModTime().After(cutoff) {
 		return 0, retireVetoed
 	}
-	size, err := s.layerSize(dir)
+	size, err := s.layerSizeReadOnly(dir)
 	if err != nil {
 		size = 0
 	}
@@ -552,9 +562,11 @@ func (s *Store) dryRunRetire(hex string, cutoff time.Time) (int64, retireStatus)
 // vetoes into stats.
 //
 // The returned complete flag reports whether every record was read and
-// decoded. Refcounts from a partial listing understate references, which
-// is safe for the record-driven pass (it only ever skips work) but fatal
-// for the orphan sweep, which would read "no references" as "garbage".
+// decoded. Refcounts from a partial listing understate references — a
+// layer shared with an unread record looks unreferenced, so it would be
+// retired (record-driven pass) or swept (orphan scan) while that record
+// still names it. Both callers therefore gate on the flag and do nothing
+// with a partial listing.
 func (s *Store) listEviction(roots RootSet, cutoff time.Time, stats *EvictStats) (cands []evictionCandidate, refcount map[string]int, complete bool, err error) {
 	refcount = map[string]int{}
 	entries, err := os.ReadDir(s.manifestsDir())
@@ -582,11 +594,13 @@ func (s *Store) listEviction(roots RootSet, cutoff time.Time, stats *EvictStats)
 		}
 		var rec imageRecord
 		if err := json.Unmarshal(b, &rec); err != nil {
+			// An undecodable record contributes no refcounts, so evicting it
+			// would strand its (unknown) layers as orphans until the next
+			// restart. Never a candidate: leave it in place and fail the
+			// enumeration — the error names the file for the operator.
 			complete = false
-			// An undecodable record can't produce refcounts; deleting it is
-			// safe (layers it referenced become orphans, collected next pass)
-			// but do so only via the normal candidate path so vetoes apply.
 			errs = append(errs, fmt.Errorf("while decoding image record %s: %w", digest, err))
+			continue
 		}
 
 		// Dedupe diffIDs per record (images may list a layer twice) so the
@@ -642,7 +656,11 @@ func (s *Store) listEviction(roots RootSet, cutoff time.Time, stats *EvictStats)
 // one of its layers: without the record the kept layer is unreachable by
 // the runtime pass. The rewrite bumps the record's mtime, so min-age keeps
 // it off the next pass's candidate list — the retry happens once the kept
-// layer is itself old enough to go.
+// layer is itself old enough to go. The cost is the record's true last-use:
+// a cold restored image now sorts as freshly used and can outlive genuinely
+// warmer images until it re-ages. Accepted — preserving the old mtime
+// instead would make the same doomed candidate churn through
+// delete-and-restore on every pass until its kept layer ages out.
 func (s *Store) restoreRecord(digest v1.Hash, raw []byte) error {
 	path := s.recordPath(digest)
 	tmp, err := os.CreateTemp(filepath.Dir(path), "."+filepath.Base(path)+".tmp-*")
--- a/internal/imagecache/gc.go
+++ b/internal/imagecache/gc.go
@@ -282,7 +282,7 @@ func (s *Store) EvictUnused(ctx context.Context, targetBytes int64, dryRun bool)
 			slog.Int("layers", len(cand.diffIDs)),
 			slog.Bool("dry_run", dryRun))
 
-		kept, candRetired, retireErrs := s.retireCandidateLayers(ctx, cand, refcount, roots, cutoff, dryRun, &stats)
+		kept, candRetired, retireErrs := s.retireCandidateLayers(ctx, cand, refcount, roots, cutoff, dryRun, dbg, &stats)
 		errs = append(errs, retireErrs...)
 		retired = append(retired, candRetired...)
 		if kept {
@@ -347,8 +347,7 @@ func (s *Store) removeStaleRecord(cand evictionCandidate, cutoff time.Time, dryR
 // layer must stay (still referenced, rooted, fresh, or failed to retire).
 // The caller must then restore the record: a kept layer with no record is
 // unreachable until the next restart.
-func (s *Store) retireCandidateLayers(ctx context.Context, cand evictionCandidate, refcount map[string]int, roots RootSet, cutoff time.Time, dryRun bool, stats *EvictStats) (kept bool, retired []string, errs []error) {
-	dbg := slog.Default().Enabled(ctx, slog.LevelDebug) // see InUse
+func (s *Store) retireCandidateLayers(ctx context.Context, cand evictionCandidate, refcount map[string]int, roots RootSet, cutoff time.Time, dryRun, dbg bool, stats *EvictStats) (kept bool, retired []string, errs []error) {
 	for _, hex := range cand.diffIDs {
 		refcount[hex]--
 		if refcount[hex] > 0 {
--- a/internal/imagecache/gc.go
+++ b/internal/imagecache/gc.go
@@ -191,9 +191,10 @@ func addSpecRoots(rs *RootSet, spec *OverlaySpec, bundle string, dbg bool) {
 
 // EvictStats reports what an eviction pass did (or, dry-run, would do).
 type EvictStats struct {
-	// FreedBytes sums retired layers' sizes, walked from the renamed-aside
-	// trees (dry-run: read from size files, or walked read-only). An
-	// estimate either way; the caller's next statfs self-corrects.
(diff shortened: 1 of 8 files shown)


igooch:
For reviewers:

How a pass decides, and specifically how running actors are protected. Protection is based on **use, not age**: an image is protected while any actor on this node references it (i.e., while that actor's bundle spec exists on disk). mtime only orders images that nothing references.

```mermaid
flowchart TD
    subgraph inuse ["Store.InUse() — LOOP over every container bundle of every actor on this node"]
        SPEC["read actors/&lt;uid&gt;/bundles/&lt;container&gt;/rootfs-overlay.json"]
        SPEC --> MAPS["ImageDigests: spec.imageDigest<br/>LayerHexes: each layer dir the spec names<br/>LayerSets: exact layer-set signature"]
    end

    inuse --> G0{"root set fully enumerated?<br/>(actors dir, bundles dirs, specs<br/>all readable)"}
    G0 -- "no" --> SKIP["⛔ skip entire pass (ERROR) —<br/>partial roots/refcounts would retire layers<br/>running actors still mount or<br/>unread records still name"]
    G0 -- "yes" --> listing

    subgraph listing ["Store.listEviction() — LOOP over every image record (image-cache/manifests/sha256/*.json)"]
        REC["read + decode record;<br/>refcount++ each of its layers<br/>(rooted and fresh records count too)"]
        REC --> G1{"this record rooted?<br/>ImageDigests[digest] OR<br/>LayerSets[signature(its diffIDs)]"}
        G1 -- "yes — running actors exit here;<br/>mtime never consulted" --> NXT1["next record ↩"]
        G1 -- "no" --> G2{"record mtime younger<br/>than min-age?"}
        G2 -- "yes" --> NXT1
        G2 -- "no" --> CAND["→ candidate list"]
    end

    listing --> COMPLETE{"was every record read + decoded?"}
    COMPLETE -- "no" --> SKIP
    COMPLETE -- "yes" --> SORT["sort candidates by mtime, oldest first<br/>(order among evictables — never permission)"]

    SORT --> evict

    subgraph evict ["Store.EvictUnused() — LOOP over candidates until targetBytes freed"]
        RM["removeStaleRecord: re-check under hitMu<br/>(gone? fresh again?) then delete record"]
        RM --> layers

        subgraph layers ["retireCandidateLayers — LOOP over the deleted record's layers"]
            L1{"refcount-- still &gt; 0?<br/>(another image uses it)"}
            L1 -- "yes" --> KEEP["layer kept"]
            L1 -- "no" --> L2{"LayerHexes[hex]?<br/>(a bundle spec names this layer)"}
            L2 -- "yes" --> KEEP
            L2 -- "no" --> L3{"layer dir mtime fresh?"}
            L3 -- "yes" --> KEEP
            L3 -- "no" --> MARK["retireLayer: rename → .rm-* (the mark)"]
        end

        layers --> KEPT{"anything kept?"}
        KEPT -- "yes" --> RESTORE["restoreRecord — record rewritten,<br/>image NOT evicted; next candidate ↩"]
        KEPT -- "no" --> EVICTED["image evicted ✓; next candidate ↩"]
    end

    evict --> RRD["removeRetiredDirs — slow RemoveAll of .rm-* dirs,<br/>outside all locks; crash debris swept at next startup"]
```

- A referenced image never reaches the mtime sort — its layers can be months old with
  zero risk; protection ends only when its bundle is removed (post-checkpoint), never by aging.
- "Marked for deletion" = renamed to `.rm-*`, possible only for a layer no record
  references and no bundle names.
- Every failure path lands on retention: unreadable image records **and** unreadable
  actors dirs / bundles dirs / specs each gate the whole pass at ERROR (missing
  dirs/specs are legitimate emptiness, not errors).

*Edited after d1c806a2: added the root-set enumeration gate (G0) and corrected the
last bullet — an unreadable spec previously rooted nothing with a WARN; it now gates
the pass like an unreadable record.*

internal/imagecache/gc.go
```diff
@@ -0,0 +1,671 @@
...
+			}
+			// WriteSpec renames its temp file into place, so a spec read here
+			// is always whole; a partial read would under-report an actor's
+			// layers and fail toward deleting them.
+			spec, err := ReadSpec(filepath.Join(bundlesDir, bundle.Name()))
+			if err != nil {
+				slog.Warn("Unreadable bundle overlay spec during image-cache root-set scan",
+					slog.String("bundle", filepath.Join(bundlesDir, bundle.Name())), slog.Any("err", err))
+				continue
+			}
```
dberkov, line 153: This fails toward deletion, asymmetric with how unreadable *records* gate the whole pass at ERROR. An unreadable spec here means a running actor's image and layers lose root-set protection — and since the record mtime can be arbitrarily old for a long-running actor that hasn't re-Ensured, min-age won't save it: the pass can retire lowerdirs under an active mount.

`WriteSpec` is rename-atomic, so a read failure here is the same operator-damage/permissions class as an unreadable record. Should `InUse` report incompleteness and have both passes skip, mirroring the `!complete` gate in `EvictUnused`? A permanently bad spec wedging GC is the same trade-off already accepted for records, and it fails safe.

internal/imagecache/gc.go
```diff
@@ -0,0 +1,671 @@
...
+	rs := RootSet{ImageDigests: map[string]bool{}, LayerHexes: map[string]bool{}, LayerSets: map[string]bool{}}
+	if s.actorsDir == "" {
+		return rs
+	}
+	actorEntries, err := os.ReadDir(s.actorsDir)
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			slog.Warn("Failed to list actors dir for image-cache root set", slog.String("dir", s.actorsDir), slog.Any("err", err))
+		}
+		return rs
```
dberkov: This is the strictly worse instance of the unreadable-spec asymmetry below: a non-`ErrNotExist` failure here (e.g. a permissions problem on the actors dir) returns an **empty root set**, so *every* placed actor loses protection and the pass still runs. If `InUse` grows a completeness flag, this branch should set it too — `ErrNotExist` staying a legitimate empty root set, everything else gating the pass.

internal/imagecache/gc.go
```diff
@@ -0,0 +1,671 @@
...
+	}
+	for _, actor := range actorEntries {
+		if !actor.IsDir() {
+			continue
+		}
+		bundlesDir := filepath.Join(s.actorsDir, actor.Name(), "bundles")
+		bundles, err := os.ReadDir(bundlesDir)
+		if err != nil {
+			continue // no bundles: actor not placed / already torn down
+		}
```
dberkov, line 141: The comment assumes `ErrNotExist`, but a real error (e.g. `EACCES`) is also swallowed here and treated as "actor has no bundles" — same failure direction as the unreadable-spec case below, and this one isn't even logged. Suggest distinguishing `errors.Is(err, os.ErrNotExist)` from other errors, with the latter at least logged (or gating the pass, if `InUse` grows a completeness flag).

dberkov (commented):
Could you please update internal/imagecache/README.md with the logic how the GC works.

internal/imagecache/gc.go
igooch: Fixed. InUse now returns (RootSet, error) and both passes skip at ERROR on a partial root set, mirroring the record gate. Tested in TestEvictUnusedSkipsPassOnUnreadableRoots.

internal/imagecache/gc.go
igooch: Fixed. non-ErrNotExist failures here now gate the pass; ErrNotExist stays a legitimate empty root set.

internal/imagecache/gc.go
igooch: Fixed. ErrNotExist still means "not placed / torn down"; anything else gates the pass.

internal/imagecache/sizes.go
```diff
@@ -58,17 +58,47 @@ func (s *Store) touchRecord(digest v1.Hash) {
...
 // file for layers unpacked before sizes were recorded.
 func (s *Store) layerSize(layerDir string) (int64, error) {
+	if n, ok, err := recordedLayerSize(layerDir); err != nil || ok {
+		return n, err
+	}
+	total := walkLayerSize(layerDir)
+	// Preserve the dir mtime (the eviction age signal) across the write.
+	if fi, statErr := os.Stat(layerDir); statErr == nil {
+		if err := os.WriteFile(filepath.Join(layerDir, layerSizeFileName), []byte(strconv.FormatInt(total, 10)+"\n"), 0o600); err == nil {
+			_ = os.Chtimes(layerDir, fi.ModTime(), fi.ModTime())
```
dberkov: Non-blocking: a concurrent `ensureLayer` reuse-touch that lands between this Stat and the Chtimes gets rewound, defeating the exact mtime veto `retireLayer` relies on. The window is microseconds and the pull's post-unpack verify converts the worst case into a failed pull + clean retry (and the call site in `retireCandidateLayers` acknowledges it), so this is bounded — but there's an alternative that removes the window entirely: don't preserve the mtime across the backfill. The bumped mtime reads as fresh → `retireLayer` vetoes once → the next pass reads the now-recorded size file without walking and evicts normally. Cost: one-pass eviction delay for pre-size-file layers, plus a one-time LRU perturbation when `CacheSize` backfills on upgraded nodes. Your call — the current form is defensible.

internal/imagecache/gc.go
```diff
@@ -0,0 +1,696 @@
...
+	for _, dir := range dirs {
+		g.Go(func() error {
+			if err := RemoveAllWritable(dir); err != nil {
+				return fmt.Errorf("while removing retired layer %q: %w", dir, err)
+			}
+			return nil
+		})
+	}
+	if err := g.Wait(); err != nil {
+		errs = append(errs, err)
```
dberkov: Non-blocking: `errgroup.Wait` returns only the *first* error, so if several `RemoveAll`s fail, the rest are silently dropped — those dirs sit as `.rm-*` (space unreclaimed, invisible to the pool) until the next restart's `sweepTempDirs`. Since there's no early-abort benefit here, a plain per-dir error collection (mutex + slice), or at least a per-failure log inside the goroutine, would surface all of them.

internal/imagecache/gc.go
```diff
@@ -0,0 +1,696 @@
...
+		}
+
+		// Dedupe diffIDs per record (images may list a layer twice) so the
+		// decrement in EvictUnused stays symmetric with this count.
+		seen := map[string]bool{}
+		var unique []string
+		for _, d := range rec.DiffIDs {
+			diffID, err := v1.NewHash(d)
+			if err != nil {
+				continue
```
dberkov, line 654: Non-blocking, consistency: a record with valid JSON but a garbled diffID is silently dropped here — it contributes no refcount for that entry, yet doesn't set `complete = false`, so it does not gate the pass. A layer genuinely referenced only through the garbled entry would look unreferenced. It's the one crack in the fail-toward-retention rule; setting `complete = false` here would match the undecodable-record handling above. Admittedly a very unlikely corruption shape.

internal/imagecache/gc.go
```diff
@@ -0,0 +1,696 @@
...
+	EvictedImages, EvictedLayers int
+	// Candidates is the number of LRU-ordered eviction candidates after all
+	// listing-stage vetoes.
+	Candidates int
+	// RootedImages counts image records excluded because a bundle overlay
+	// spec roots them (the "actively placed" protection).
+	RootedImages int
+	// SkippedRooted / SkippedFresh count vetoes, whether fired at listing
+	// time or by the per-victim re-check during the pass.
+	SkippedRooted, SkippedFresh int
```
dberkov: Doc nit: this says `SkippedRooted` counts vetoes "whether fired at listing time or by the per-victim re-check", but at listing time rooted images increment `RootedImages`, not `SkippedRooted` — nothing increments `SkippedRooted` during listing; it only fires in `retireCandidateLayers`.

internal/imagecache/gc.go
```diff
@@ -0,0 +1,696 @@
...
+				refcount[diffID.Hex]++
+			}
+		}
+
+		fi, err := e.Info()
+		if err != nil {
+			continue
+		}
+		// A record is rooted either by digest (a bundle spec naming it) or
+		// because every layer it lists is rooted. The latter covers the
```
dberkov: Doc nit: "because every layer it lists is rooted" describes the rule more loosely than the implementation — the check is an *exact-set* match against `roots.LayerSets`, so an all-layers-rooted **subset** record is deliberately not rooted (per the `RootSet` doc, to avoid weakening the byte cap under layer sharing). As written, this comment and the `RootSet` one read as contradicting each other; suggest "because its exact layer set matches a rooted bundle's".

internal/imagecache/sizes.go
igooch: Went one step further in ac194f3e: eviction paths now size the layer after retiring, from the size file that rides the rename (layerSizeReadOnly, O(1), read-only — 6ff97da5), so GC never writes into the live pool at all. Your bump-without-restore is applied on the CacheSize path — at the eviction call sites it would have made the one-shot startup scan veto exactly the pre-size-file orphans it exists to reclaim.

internal/imagecache/gc.go
igooch: Fixed in ac194f3e: per-dir error collection; Wait's result unused.

internal/imagecache/gc.go
igooch: Fixed in ac194f3e: sets complete = false and gates the pass, matching the undecodable-record handling.

internal/imagecache/gc.go
igooch: Fixed in ac194f3e — comment now says it only fires during the pass; listing-time rooted images count into RootedImages.

internal/imagecache/gc.go
igooch: Fixed in ac194f3e (reflowed in 6ff97da5): now states the exact-set match and points at RootSet.LayerSets.

PR #739, opened 2026-08-04
ateom-gvisor: measure the sandbox cgroup for GetWorkloadStats

### LINKED ISSUE #594: Phase 0: StatsWorkload RPC on ateom (proto, both runtime reads, identity retention)

baizhenyu:
Second of the three PRs for Phase 0 of #550. #667 has merged, so this is now rebased onto `main` and stands alone as a single commit.

Fills in the measurement half of `GetWorkloadStats` for the gVisor runtime, so it returns real numbers instead of `Unimplemented`. The micro-VM runtime keeps its stub until the guest-agent read lands in PR 3, which closes #594.

## Where the numbers come from

`/sys/fs/cgroup/pause`, the sandbox's cgroup v2 leaf.

gVisor runs every container of a sandbox inside one host process — the sentry — and runsc places that process in the cgroup of the container that created the sandbox. That is `pause`, the first container `RunWorkload` and `RestoreWorkload` create. The actor's own containers get leaves too, but their memory and CPU are the sentry's and are charged there instead. This is why a sample is attributed to the actor and not to a container, as the proto already says.

The path follows the `"/" + containerName` convention `runsc.ensureContainerCgroupsPath` writes into the OCI spec, resolved against the pod's own cgroup scope that `setupCgroupDelegation` prepares (the worker runs in a private cgroup namespace, so `/sys/fs/cgroup` is the pod's cgroup, not the host root).

## The read

New `cmd/ateom-gvisor/internal/cgroupstats`. It takes the cgroup directory as an argument rather than reaching for an absolute path, so the parsing is testable from a fixture tree without root or a live sandbox — and it carries no build tag, so unlike the rest of `cmd/ateom-gvisor` those tests run on every platform.

It fails only when `memory.current` is missing or unparseable, wrapping `fs.ErrNotExist` in the first case so the handler can tell "no cgroup to measure" from "the format is not what we parse".

Every other file degrades to zero on its own field, because each has a legitimate reason to be absent on a healthy node: `memory.peak` predates kernel 5.19, and `setupCgroupDelegation` enables controllers best-effort, so a cgroup with `memory` but no `cpu` is reachable. Reporting no memory numbers because the node could not report CPU seemed like the wrong trade — pushback welcome if you'd rather it be all-or-nothing.

Working set is `memory.current − memory.stat:inactive_file`, saturating at zero rather than wrapping. The two files are read a moment apart and are not a consistent snapshot, so `inactive_file` can legitimately exceed the `memory.current` read just before it; on `uint64` the naive subtraction gives ~1.8e19 instead of ~0.

## Proto: the epoch caveat (added after review)

The measurement block gains a note it was missing. It already said `cpu_usage_usec` is per-epoch; `memory_peak_bytes` accumulates the same way and said nothing, so a caller reading it as a lifetime peak would silently under-report after every restore — this source recreates the sandbox cgroup, and `memory.peak` restarts with it.

The note is phrased per source rather than as one rule, because the two sources do not agree. `STATS_SOURCE_GUEST_AGENT` reads counters the guest kernel keeps in its own RAM, and `restoreFullScope` brings those back from the memory snapshot, so an epoch there does not end where it ends here.

It also drops the old suggestion that watching for a decrease is enough to spot a boundary. An epoch can begin *above* the value last reported — `DATA_ON_GOLDEN` resumes the template's golden guest — and then no decrease ever appears.

Comment-only, plus the regenerated `ateom.pb.go`.

## Status codes

This is the first user of the `NOT_FOUND` / `FAILED_PRECONDITION` split #667 documented, and the split is by what the caller has to do about it.

**`NOT_FOUND`** — the ateom is not executing the requested actor: it is available, or it was recycled onto a different one. Retrying on the same timer will never change that; the caller's worker-to-actor mapping wants re-resolving.

**`FAILED_PRECONDITION`** — the requested actor *is* the one here, but there is nothing to read yet. Since #667 the attribution is retained from the moment the ateom accepts the actor, so a poll landing in the boot arrives before runsc has created the leaf. Transient: skip the sample and take the next one.

**`Internal`** — the cgroup is there and does not parse. Not a routine race, so it must not be reported as one.

The post-read pointer recheck returns `NOT_FOUND` for the same reason the first two do: it catches exactly the state a retry would land on, and one condition should not report two different codes depending on where in the handler it was noticed.

## Locking

`AteomService.activeActor` becomes an `atomic.Pointer`. The three lifecycle RPCs still hold `AteomService.lock` for their whole bodies and keep doing so — the change is for the reader.

`lock` is not what would make the read slow; it is the only thing that would make a plain field read *legal* while a writer is running. And it is held across entire boots, restores, and checkpoints (including the readiness wait), so "take the lock to read one pointer" becomes "wait out the boot", with pollers piling up behind it. `atomic.Pointer` makes the read safe on its own, so the reader never queues. The field is only ever assigned or cleared as a whole pointer, never mutated in place, which is what the type is for.

The type makes a lock-free read *possible*; it does not make one happen. `TestGetWorkloadStatsDoesNotTakeLock` pins it: it holds `s.lock` across the call, so a handler that ever reaches for the lock deadlocks.

The handler reloads the pointer after the read and compares identity, so a checkpoint plus a fresh run completing underneath the read is reported rather than misattributed to the wrong actor.

## Deliberately not here: panic recovery

grpc-go serves each RPC on its own goroutine and does not recover handler panics, and the Go runtime kills the process on an unrecovered panic in any goroutine. `grep -rn "recover()"` finds nothing in our non-test code, so today a nil dereference in any handler ends every other RPC the ateom is serving — including an in-flight checkpoint.

An earlier revision of this PR fixed that for the two ateoms. I have pulled it back out. The gap predates this change and is not made reachable by it: nothing calls `GetWorkloadStats` yet, and neither the handler (which nil-checks before every dereference) nor the parser (which length-checks before indexing) has a panic path. It deserves its own PR covering every server rather than the two ateoms a telemetry change happens to touch, and the real constraint is that it lands before the phase that adds a caller polling on a timer — not before this one.

## Follow-up worth filing separately

The readiness wait is held under `s.lock` in all four boot paths (`main.go:364`, `:639`, and the micro-VM equivalents). It is not a runsc subcommand, so it sits outside what the lock's own comment claims to protect, and it is the longest hold in the file. Shrinking it changes lifecycle behavior, so it does not belong in a telemetry PR — but it is the highest-leverage locking fix here.

## Testing

- `cgroupstats`: nine-case table over fixture trees — all files present, missing `memory.peak`, missing `cpu.stat`, missing `memory.stat`, `memory.stat` without `inactive_file`, `inactive_file` above `memory.current`, all-zero, malformed/blank/over-long lines, unparseable optional files — plus missing-cgroup and malformed-`memory.current` cases that pin which one matches `fs.ErrNotExist`.
- `GetWorkloadStats`: happy path against a fixture cgroup root, a five-case error table (empty `actor_uid` → `InvalidArgument`; available and UID mismatch → `NotFound`; vanished cgroup → `FailedPrecondition`; malformed cgroup → `Internal`), and the no-lock regression test.
- All nine local verifiers pass, `shellcheck.sh` and `proto-fmt.sh` included. No shell scripts changed.

The `cmd/ateom-gvisor` handler tests are `//go:build linux`, and they were run rather than only compile-checked: `go test ./...` is green for the whole repo inside a `golang:1.26.3` container (linux/arm64), and the `GetWorkloadStats` tests pass under `-race`.

`cgroupstats.Read` was also pointed at that container's live cgroup v2 tree as a one-off check that the parser agrees with a real kernel and not just with its own fixtures: all four files present, and the working set matching `memory.current` less `inactive_file`.

The one thing a container cannot cover — that the sentry lands in `/sys/fs/cgroup/pause` — was measured on a live worker. The path is derived from `ensureContainerCgroupsPath`, but the placement is a runsc behavior, so it needed a real node. This branch's own build was deployed to a GKE cluster, an actor was run on it from the counter ActorTemplate, and the worker's processes were read through an ephemeral container sharing the ateom's PID namespace:

```
1    ateom-gvisor                                    0::/…-<sandbox>.scope/ateom
28   runsc-sandbox … boot … pause                    0::/…-<sandbox>.scope/pause
27   runsc-gofer  --bundle=…/bundles/pause           0::/…-<sandbox>.scope/pause
119  runsc-gofer  --bundle=…/bundles/counter         0::/…-<sandbox>.scope/pause
```

Both halves of the assumption hold. The sentry is in the `pause` leaf, and the gofer for the *counter* container is in that same leaf rather than one of its own — which is the one-cgroup-per-sandbox property the handler relies on to report a whole actor from a single read. `setupCgroupDelegation`'s private namespace is visible too: ateom's own processes sit in the sibling `ateom` leaf, so the sandbox's usage is not mixed with the ateom's.

The kubelet's cadvisor endpoint reads the same leaf independently, and reports the four values this PR parses: `memory.current` 39,854,080, peak 41,914,368, working set 38,596,608, cpu 0.238517 s. The working set is consistent with the parser's definition — 39,854,080 less 38,596,608 leaves 1,257,472 of `inactive_file`.

Not covered: `GetWorkloadStats` answering over the wire. The ateom serves its control API on a unix socket inside a distroless image, so a live RPC needs a client mounted alongside it — which is the phase that adds the caller, not this one.

### DIFF
--- a/cmd/ateom-gvisor/stats.go
+++ b/cmd/ateom-gvisor/stats.go
@@ -1,0 +1,164 @@
+//go:build linux
+
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"errors"
+	"io/fs"
+	"path/filepath"
+	"time"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/agent-substrate/substrate/cmd/ateom-gvisor/internal/cgroupstats"
+	"github.com/agent-substrate/substrate/internal/proto/ateompb"
+)
+
+// defaultCgroupRoot is the worker pod's own cgroup scope. The worker runs in a
+// private cgroup namespace, so this path is the pod's cgroup rather than the
+// host root, and runsc's per-container leaves are its direct children (see
+// setupCgroupDelegation).
+const defaultCgroupRoot = "/sys/fs/cgroup"
+
+// sandboxCgroupContainer is the container whose cgroup leaf accounts for the
+// sandbox as a whole.
+//
+// gVisor runs every container of a sandbox as threads inside one host process,
+// the sentry. runsc starts that process from the root container's create and
+// from inside that container's cgroup — container.createRoot wraps the sandbox
+// and gofer spawn in cgroup.RunInCgroup — so the sentry lands in the leaf of
+// "pause", the first container RunWorkload and RestoreWorkload create.
+//
+// The leaf is a direct child of the delegated scope rather than of ateom's own
+// cgroup, because runsc resolves cgroupsPath against the parent of the cgroup
+// it is running in: cgroup v2 forbids a cgroup from holding processes and
+// delegating controllers to children at once, so runsc walks up one level to
+// find a directory it is allowed to create in. setupCgroupDelegation moves
+// ateom into /sys/fs/cgroup/ateom precisely so that one level up is the
+// delegated scope.
+//
+// The actor's own containers get leaves too, since cmdCreate calls
+// ensureContainerCgroupsPath for each of them, but those leaves stay empty by
+// design. gVisor's setupCgroupForSubcontainer creates them with empty resources
+// and explains why: "Since subcontainers run exclusively inside the sandbox,
+// subcontainer cgroups on the host have no effect on them. However, some tools
+// (e.g. cAdvisor) uses cgroups paths to discover new containers and report
+// stats for them." They are discovery markers, not accounting.
+//
+// So the actor's memory and CPU are the sentry's and are charged here. That is
+// why a sample is attributed to the actor rather than to a container, and why
+// this reads one leaf instead of summing them — the others have nothing in them
+// to sum.
+//
+// What the leaf holds besides the actor's own work: the sentry's own overhead
+// (its Go heap, page tables, netstack) and the gofers. Process listings taken
+// on a live node in #161 put runsc-sandbox and both gofers — the pause
+// container's and the actor container's — in the "pause" cgroup. Those runs
+// predate #496, so they establish the leaf name and the fact that everything
+// lands in one leaf, not the absolute path, which #496's delegation moved under
+// the pod scope.
+//
+// So this measures the sandbox, not the actor's processes in isolation, which
+// is what the proto means by "the unit of measurement is the SANDBOX". Sizing
+// and chargeback want that number, since the sentry's overhead is a cost the
+// node pays for running this actor, but it is not comparable to a container
+// figure from a runc-based runtime, and a mostly idle actor reads as a nonzero
+// floor. Splitting the actor's share out would need the sentry's own
+// accounting, which the proto already says is not reported here.
+//
+// The name has to agree with the cgroupsPath convention in
+// runsc.ensureContainerCgroupsPath, which is "/" + containerName relative to
+// the same scope.
+const sandboxCgroupContainer = "pause"
+
+// GetWorkloadStats implements ateompb.Ateom/GetWorkloadStats.
+//
+// Unlike the three lifecycle RPCs this does not take s.lock, and must not start:
+// it is polled on a timer for the whole life of a workload, while lock is held
+// across entire boots and checkpoints. Blocking on it would park every poll
+// behind a multi-second runsc call, and — worse in the other direction — holding
+// it across the cgroup read would put a CheckpointWorkload behind telemetry.
+// The attribution is read from an atomic instead, and the cgroup files are read
+// with no lock held at all.
+func (s *AteomService) GetWorkloadStats(ctx context.Context, req *ateompb.GetWorkloadStatsRequest) (*ateompb.GetWorkloadStatsResponse, error) {
+	if req.GetActorUid() == "" {
+		return nil, status.Error(codes.InvalidArgument, "actor_uid is required")
+	}
+
+	// Both of these are NOT_FOUND rather than FAILED_PRECONDITION: they tell the
+	// caller the requested actor is not here, which no amount of retrying on the
+	// same timer will change. Its worker-to-actor mapping wants re-resolving.
+	active := s.activeActor.Load()
+	if active == nil {
+		return nil, status.Errorf(codes.NotFound, "ateom is available; it is not executing actor %q", req.GetActorUid())
+	}
+	if active.UID != req.GetActorUid() {
+		return nil, status.Errorf(codes.NotFound, "ateom is executing actor %q, not the requested %q", active.UID, req.GetActorUid())
+	}
+
+	observedAt := time.Now()
+	sample, err := cgroupstats.Read(filepath.Join(s.cgroupRoot, sandboxCgroupContainer))
+	if err != nil {
+		// The requested actor is the active one but its cgroup is not there. Most
+		// often that is a poll landing in the boot: the ateom retains the
+		// attribution from the moment it accepts the actor, before runsc has
+		// created the leaf. The other way in is a sandbox that went away between
+		// the check above and the read, which the next CheckpointWorkload turns
+		// into the NOT_FOUND above. Either way it is "no numbers right now" and the
+		// caller should take the next sample, so FAILED_PRECONDITION. Anything else
+		// is a real read failure.
+		if errors.Is(err, fs.ErrNotExist) {
+			return nil, status.Error(codes.FailedPrecondition, "no sandbox cgroup to measure yet")
+		}
+		return nil, status.Errorf(codes.Internal, "reading sandbox cgroup: %v", err)
+	}
+
+	// Re-check that the same workload is still the active one. The read above
+	// holds no lock, so a checkpoint plus a fresh run can complete underneath it,
+	// and the numbers would then belong to an actor other than the one being
+	// reported. Pointer identity is enough: activeActor is stored as a new
+	// pointer on every Run and Restore and never mutated in place, so an
+	// unchanged pointer means no transition happened across the read.
+	//
+	// NOT_FOUND, like the two checks above and for the same reason: the requested
+	// actor is no longer the one here, so a retry lands on one of them and gets
+	// that answer anyway. The same state should not report two different codes
+	// depending on where in the handler it was noticed.
+	if s.activeActor.Load() != active {
+		return nil, status.Errorf(codes.NotFound, "ateom stopped executing actor %q while the sample was being taken", req.GetActorUid())
+	}
+
+	return &ateompb.GetWorkloadStatsResponse{
+		Atespace:               active.Ref.Atespace,
+		ActorName:              active.Ref.Name,
+		ActorUid:               active.UID,
+		ActorTemplateNamespace: active.TemplateNamespace,
+		ActorTemplateName:      active.TemplateName,
+
+		SandboxClass: ateompb.SandboxClass_SANDBOX_CLASS_GVISOR,
+		Source:       ateompb.StatsSource_STATS_SOURCE_CGROUP,
+
+		MemoryCurrentBytes:    sample.MemoryCurrentBytes,
+		MemoryPeakBytes:       sample.MemoryPeakBytes,
+		MemoryWorkingSetBytes: sample.MemoryWorkingSetBytes,
+		CpuUsageUsec:          sample.CPUUsageUsec,
+
+		ObservedAtUnixNano: observedAt.UnixNano(),
+	}, nil
+}
--- a/internal/proto/ateompb/ateom.proto
+++ b/internal/proto/ateompb/ateom.proto
@@ -300,12 +300,28 @@ message GetWorkloadStatsResponse {
 
   // Measurements. All four are zero when source is STATS_SOURCE_UNSPECIFIED,
   // which means "not measured" rather than "measured as zero".
+  //
+  // Two of them accumulate -- memory_peak_bytes and cpu_usage_usec -- and both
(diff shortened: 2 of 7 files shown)


internal/proto/ateompb/ateom.proto
```diff
@@ -301,7 +301,13 @@ message GetWorkloadStatsResponse {
   // Measurements. All four are zero when source is STATS_SOURCE_UNSPECIFIED,
   // which means "not measured" rather than "measured as zero".
   uint64 memory_current_bytes = 8;
+  // High-water mark of memory_current_bytes. Also zero when the runtime cannot
+  // report a peak at all: the cgroup source reads memory.peak, which only
+  // exists on Linux 5.19 and later.
   uint64 memory_peak_bytes = 9;
```
git286, line 319: memory_peak_bytes is per-epoch too and might worth calling it out here. A restore recreates the sandbox, so the pause cgroup is new and memory.peak restarts at zero: a caller computing "peak memory over the actor's lifetime" silently under-reports after every restore.

cmd/ateom-gvisor/stats.go
```diff
@@ -0,0 +1,164 @@
...
+	active := s.activeActor.Load()
+	if active == nil {
+		return nil, status.Errorf(codes.NotFound, "ateom is available; it is not executing actor %q", req.GetActorUid())
+	}
+	if active.UID != req.GetActorUid() {
+		return nil, status.Errorf(codes.NotFound, "ateom is executing actor %q, not the requested %q", active.UID, req.GetActorUid())
+	}
+
+	observedAt := time.Now()
+	sample, err := cgroupstats.Read(filepath.Join(s.cgroupRoot, sandboxCgroupContainer))
```
git286, line 116: AI findings:

Nothing here distinguishes "the leaf exists and is the sandbox" from "the leaf exists and is one of the empty discovery markers setupCgroupForSubcontainer creates." 

Consider reading cgroup.procs from the same dir and logging when the leaf you just measured holds no processes: not an error, since a checkpoint window can legitimately empty it, but it turns a silent misattribution into logs.

internal/proto/ateompb/ateom.proto
baizhenyu: Agreed, and done — but the fix is not quite the same sentence as `cpu_usage_usec` had, because that sentence turned out to be too strong.

You are right for this source. `runsc delete` destroys the cgroup it created, so a restore gets a new `pause` leaf and `memory.peak` restarts at zero.

The guest agent source does not behave that way. `restoreFullScope` relaunches cloud-hypervisor with `--restore` and resumes, and guest RAM comes back from the memory snapshot. The counters live in the guest kernel's own memory, so a restored guest brings them back rather than restarting them. `DATA_ON_GOLDEN` is stranger still: it resumes the *template's* golden guest, so an actor's first sample can start at whatever the golden had accumulated when it was snapshotted.

So the note is now written per source instead of as one rule, and it sits above the measurement block since it covers both cumulative fields:

> Two of them accumulate — `memory_peak_bytes` and `cpu_usage_usec` — and both are scoped to the current EPOCH rather than to the actor's lifetime. An epoch begins wherever the accounting behind the sample begins, which is not the same event for every source: `STATS_SOURCE_CGROUP` reads a sandbox cgroup that a restore recreates, so both restart at zero there, while `STATS_SOURCE_GUEST_AGENT` reads counters the guest kernel keeps in its own RAM, which a restored guest brings back with it. A caller that wants a lifetime figure has to accumulate one itself, and must read a decrease as a new epoch rather than emit a negative delta — but not the converse. An epoch can also begin at a value above the last one reported, so no comparison of consecutive samples detects every boundary.

The last clause is the part your comment made me check. The old wording told callers to watch for a decrease, which is sound advice here but not sufficient in general — `DATA_ON_GOLDEN` can begin an epoch *above* the last value reported, and then no decrease ever appears. `cpu_usage_usec` is now just "cumulative CPU time within the current epoch" and the shared paragraph carries the rest.

I have left the guest-agent specifics (which scope resumes and which cold-boots) out of this PR, since nothing implements that source yet. They belong with the micro-VM change.

cmd/ateom-gvisor/stats.go
baizhenyu: I do not think this one holds, and I would rather not add the read — but say the word and I will add it.

The handler does not discover a leaf. It reads one fixed path, `filepath.Join(s.cgroupRoot, sandboxCgroupContainer)`, where `sandboxCgroupContainer` is the constant `"pause"` — the root container, the one `RunWorkload` and `RestoreWorkload` create first. The empty discovery markers `setupCgroupForSubcontainer` creates are the *subcontainer* leaves (`/counter` and friends), and this never reads them. That is the distinction the constant's doc comment is drawing a few lines up:

> The actor's own containers get leaves too … but those leaves stay empty by design. gVisor's `setupCgroupForSubcontainer` creates them with empty resources … They are discovery markers, not accounting.

There is now a live measurement behind that. On a GKE worker running this branch's own build, an ephemeral container sharing the ateom's PID namespace put the sentry and both gofers in the same leaf:

```
1    ateom-gvisor                            0::/…-<sandbox>.scope/ateom
28   runsc-sandbox … boot … pause            0::/…-<sandbox>.scope/pause
27   runsc-gofer --bundle=…/bundles/pause    0::/…-<sandbox>.scope/pause
119  runsc-gofer --bundle=…/bundles/counter  0::/…-<sandbox>.scope/pause
```

The kubelet's cadvisor endpoint reads that same leaf and reports 39,854,080 B and 0.238517 s against it. It is the sandbox, and it is not empty.

On the narrower suggestion — log when the measured leaf holds no processes — the case that would actually hurt us is a leaf that lingers after teardown, and a `cgroup.procs` check does not catch it. A stale leaf keeps its `cpu.stat` and `memory.peak` from the previous epoch, and once the next sandbox starts, `cgroup.procs` refills with the new sentry. The check would read "healthy" at exactly the moment the numbers are inherited. That is the sandbox-cgroup-reuse problem, and it wants a fresh cgroup per activation rather than a log line.

The remaining case is a genuinely empty leaf mid-checkpoint. That reports near-zero against an idle sandbox's ~40 MB floor, so it is visible in the sample rather than silent, and the next poll gets `FAILED_PRECONDITION` once `runsc delete` has removed the directory.

git286 (approved):

Issue #732, opened 2026-08-04
Make Update<ResourceType> methods to follow API guidelines


juli4n:
This task is to update `UpdateActor` (and `UpdateActorSnapshotTag`) to follow https://github.com/agent-substrate/substrate/blob/main/docs/api-style-guide.md#34-update.

The main change is to make update methods take the resource and a field mask in the request payload.

PR #742, opened 2026-08-04
Make UpdateActor method to follow API guidelines

laoj2:
#732  

* `UpdateActorRequest` now carries the resource itself + `update_mask`
* `worker_selector` is now applied via the update mask (making it possible to clear the worker selector)
*  `update_mask` is validated against an allowlist of paths. Currently, only `worker_selector` is allowed. 
* Added `uid` and `version` as optional guards. 

I'll update UpdateActorSnapshotTag in a follow-up PR

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/functional_test.go
+++ b/cmd/ateapi/internal/controlapi/functional_test.go
@@ -50,6 +50,7 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/testing/protocmp"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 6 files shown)


cmd/ateapi/internal/controlapi/update_actor.go
```diff
@@ -41,9 +53,20 @@ func (s *Service) UpdateActor(ctx context.Context, req *ateapipb.UpdateActorRequ
...
+	}
+
+	expectedVersion := actor.GetMetadata().GetVersion()
+	if version := in.GetMetadata().GetVersion(); version != 0 {
+		expectedVersion = version
+	}
+
+	applyActorUpdateMask(actor, in, req.GetUpdateMask())
+
+	updated, err := s.persistence.UpdateActor(ctx, actor, expectedVersion)
```
juli4n, line 69: I think this is ok as a first implementation (as it makes progress towards stabilizing the API), but we need to consider reworking the `UpdateActor` API in the storage layer. A few issues with the current approach:

* The is a race between the time we check the UID and when we update.
* Similarly, if there is a racing request to update, and the request doesn't have a version precondition, the handler will return an spurious `Aborted` which could have been resolved. If we had a way to pass a closure function to `persistence.UpdateActor` which has the actor as a parameter and mutates it, the storage could handle this scenarios by just retrying the tx.

Something like this:

```
UpdateActor(ctx, actorRef, mutate func(*Actor) error) (*Actor, error)
```

cmd/ateapi/internal/controlapi/functional_test.go
juli4n, line 2285: Can we add two tests here for the following cases?

* Create an actor, update, and then update again with old version, assert it fails.
*  Create actor, delete, create actor with same name, delete using old UID, assert it fails.

cmd/ateapi/internal/controlapi/update_actor.go
laoj2: Thanks for catching this! We discussed offline, we'll do this in a follow up PR: https://github.com/agent-substrate/substrate/issues/763

cmd/ateapi/internal/controlapi/functional_test.go
laoj2: This was an unit test. But, made it a functional test.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -318,15 +316,18 @@ message CreateActorRequest {
...
+  // actor.metadata.atespace and actor.metadata.name identify which resource to
+  // update.
+  // actor.metadata.version and actor.metadata.uid are optional preconditions and
+  // zero values skip the check.
   Actor actor = 1;
+
+  // The set of fields to update. Required.
+  //
+  // Field paths are relative to the Actor message (e.g., "worker_selector").
+  // Only client-mutable fields may be named; "*" is not supported.
```
juli4n: Should we document here which fields are mutable>?

juli4n (dismissed):

pkg/proto/ateapipb/ateapi.proto
laoj2: Done. Documented what fields this method supports

juli4n (approved):

cmd/ateapi/internal/controlapi/update_actor.go
thockin: I am messing with this code for validation - I will rebase, jut let m know when you are done.

PR #743, opened 2026-08-04
docs: add guide for running the microVM runtime locally (KVM / Lima)

Oneimu:
### Why

  The README Quickstart covers getting a local Substrate cluster running with
  the default gVisor runtime, but there's no documentation for the micro-VM
  runtime, where the friction can't be scripted away: it requires `/dev/kvm`
  (bare metal, nested virtualization, or Lima on macOS), and on Apple Silicon
  the Lima configuration and asset assembly are non-obvious. New contributors
  have to reverse-engineer the `hack/` scripts.

  ### What
  
  Adds `docs/dev/microvm-local.md` — "Running the microVM runtime locally" — a
  focused guide covering only the microVM delta, based on notes from real
  onboarding runs. General setup is deferred to the README Quickstart (listed
  as the guide's prerequisite) rather than duplicated:

  - **Option A: Linux host with KVM** — verifying `/dev/kvm` and CPU virt
    support, the rootless-Docker caveat for the KVM probe, cluster creation,
    and the one-shot `run-microvm-demo-kind.sh` bring-up.
  - **Option B: Apple Silicon macOS via Lima** — nested virtualization with the
    guest image pinned to Ubuntu 25.10 (until the kernel issue in the default
    image is fixed), `vzNAT` networking, and assembling the arm64 assets inside
    the Lima guest (`assemble.sh` requires a Linux host of the target arch).
  - **Trying it out** — defers to the next steps the demo script prints and
    links the counter demo's micro-VM variant, instead of duplicating those
    commands.
  - **Troubleshooting** — symptom → root cause → fix entries actually hit
    during onboarding (rootless-Docker KVM probe failures, the `aws` CLI
    staging requirement until #804 lands, arm64 `virtiofsd` build deps, M1
    lacking FEAT_NV2).


docs/dev/onboarding.md
```diff
@@ -0,0 +1,380 @@
...
+```
+
+Then jump to [§5 Workload testing](#5-workload-testing--actor-lifecycle).
+
+## 4. Path B: MicroVM on kind
+
+The micro-VM runtime needs `/dev/kvm` visible to the Docker environment where
+the kind node runs. Two supported host setups:
+
+### 4.1 Option A: Linux host with KVM (recommended)
```
BenTheElder: FWIW I've done virtually all development with limactl on macOS :-)

docs/dev/onboarding.md
```diff
@@ -0,0 +1,380 @@
...
+```
+
+## 6. Troubleshooting / friction log
+
+Issues actually hit during onboarding, with root causes and fixes:
+
+| # | Symptom | Root cause | Fix |
+|---|---|---|---|
+| 1 | `permission denied while trying to connect to the docker API` | User not in the `docker` group (or group not active in this session) | `sudo usermod -aG docker "$USER"` then `newgrp docker` or re-login |
+| 2 | `/dev/kvm: permission denied` during the kind KVM probe | Default `/dev/kvm` permissions (`660 root:kvm`) block the non-root probe container | Add yourself to the `kvm` group and re-login, or `sudo chmod 666 /dev/kvm` before `./hack/create-kind-cluster.sh` (see known issue below) |
```
BenTheElder: let's fix this instead of suggesting that people modify KVM permissions?

CONTRIBUTING.md
```diff
@@ -25,6 +25,13 @@ sign a new one.
 This project follows [Google's Open Source Community
 Guidelines](https://opensource.google/conduct/).
 
+### Set up a local development environment
+
+New to the project? [docs/dev/onboarding.md](docs/dev/onboarding.md) walks
+through bringing up a local Substrate cluster on kind — for both the gVisor
```
BenTheElder:
```suggestion
through bringing up a local Substrate cluster on [kind](https://kind.sigs.k8s.io) — for both the gVisor
```

CONTRIBUTING.md
```diff
@@ -25,6 +25,13 @@ sign a new one.
...
+and micro-VM (KVM / Lima) runtimes — and exercising the full actor lifecycle,
```
BenTheElder: nit:
```suggestion
and microVM  runtimes — and exercising the full actor lifecycle,
```

1. let's be consistent about microVM / microvm in the docs, it's usually not hyphenated
2. lima isn't part of the runtime, and KVM isn't really adding much detail here.

docs/dev/onboarding.md
```diff
@@ -0,0 +1,380 @@
...
+## 7. Known issues / planned improvements
+
+Rough edges in the current scripts that surfaced during onboarding review;
+fixes welcome:
+
+- **KVM probe requires host permission changes.** The `/dev/kvm` probe in
+  `hack/create-kind-cluster.sh` runs an unprivileged container, so users must
+  `chmod`/group-modify the host device first. The probe should instead run as
+  a root container so no host mutation is needed.
+- **Micro-VM asset staging requires the `aws` CLI on the host.**
```
BenTheElder: I guess this is OK as a known issue, but the KVM permissions I don't think we should recommend and it should be pretty cheap.

Do you have a rootless docker install? We don't specify the user, so the user should be root .. (`docker run --rm --device /dev/kvm busybox`)

CONTRIBUTING.md
BenTheElder: I see that we missed this in some of the scripts, we can normalize those in a follow-up.

docs/dev/onboarding.md
```diff
@@ -0,0 +1,380 @@
...
+### 5.4 Send traffic through the router
+
+```sh
+kubectl port-forward -n ate-system svc/atenet-router 8080:80 &
+
+# The Host header encodes the actor and atespace — the router uses it to
+# locate (and, if needed, resume) the actor:
+curl -s -H "Host: my-counter-1.demo.actors.resources.substrate.ate.dev" http://localhost:8080/
+curl -s -H "Host: my-counter-1.demo.actors.resources.substrate.ate.dev" http://localhost:8080/
+# -> Greeting from pod ... requestCount=2, fileCount=2
```
BenTheElder: 🤖 **nit** 🟢 – The demo does not print this. `demos/counter/counter.go` formats its response as:

```
hello from: <ip> | preserved memory count: 2 | preserved file counter: 2
```

so neither "Greeting from pod" nor `requestCount=`/`fileCount=` appears, here or in the §5.5 suspend/resume example.

docs/dev/onboarding.md
BenTheElder: [moving the rootless discussion to the main point]

docs/dev/onboarding.md
BenTheElder: Ok so this is actually due to is due to rootless docker. The container *is* spawned as root already, but when you have a rootless docker installation it will be remapped.

I guess the permissions change is OK.

We could also suggest using rootful docker as an alternative option.

Oneimu: Rewrote the section and the troubleshooting entry to attribute this to rootless Docker, with root Docker as the primary suggestion, and removed the incorrect known-issue about the probe.

docs/dev/onboarding.md
```diff
@@ -0,0 +1,377 @@
...
+
+> **Hardware requirement:** the first-generation M1 lacks hardware support for
+> ARM nested virtualization (FEAT_NV2) — Lima will fail with
+> `[hostagent] Starting VZ ... FATA exiting`. Use a newer Apple Silicon chip,
+> or fall back to Option A on a Linux host.
+
+**Step 1 — install tooling:**
+
+```sh
+brew install lima docker kind ko go
```
sfunkenhauser: Install kubectl as well

docs/dev/onboarding.md
```diff
@@ -0,0 +1,377 @@
...
+```
+
+**Step 2 — launch Lima with nested virtualization:**
+
+```sh
+limactl start --name=docker-nested template:docker-rootful --nested-virt
```
sfunkenhauser: To pin to 25.10 until the kernel is fixed.

```
limactl start --name=docker-nested template://docker-rootful --nested-virt --set '.images = [
{"location":"[https://cloud-images.ubuntu.com/releases/questing/release/ubuntu-25.10-server-cloudimg-arm64.img](https://www.google.com/url?q=https://cloud-images.ubuntu.com/releases/questing/release/ubuntu-25.10-server-cloudimg-arm64.img&sa=D&source=docs&ust=1786387046716559&usg=AOvVaw1ewVyz0gU1aVZ27bq0fEFN)","ar
ch":"aarch64"},
{"location":"[https://cloud-images.ubuntu.com/releases/questing/release/ubuntu-25.10-server-cloudimg-amd64.img](https://www.google.com/url?q=https://cloud-images.ubuntu.com/releases/questing/release/ubuntu-25.10-server-cloudimg-amd64.img&sa=D&source=docs&ust=1786387046716729&usg=AOvVaw0_JmGpGOi87gtYaFnubuvn)","ar
ch":"x86_64"}
]'
```

docs/dev/onboarding.md
```diff
@@ -0,0 +1,377 @@
...
+```
+
+When prompted to edit the configuration, set at least:
+
+```yaml
```
sfunkenhauser: I also needed to switch to vzNAT.

```
networks:
  - vsNAT: true
```

docs/dev/onboarding.md
```diff
@@ -0,0 +1,377 @@
...
+
+```sh
+export DOCKER_HOST="unix://${HOME}/.lima/docker-nested/sock/docker.sock"
+echo 'export DOCKER_HOST="unix://${HOME}/.lima/docker-nested/sock/docker.sock"' >> ~/.zprofile
+
+cd <your substrate checkout>
+
+# Cluster + control plane + microVM demo
+./hack/create-kind-cluster.sh
+./hack/install-ate-kind.sh --deploy-ate-system
```
sfunkenhauser: Do we need to run `./hack/install-ate-kind.sh --deploy-ate-system`? It looks like `./hack/run-microvm-demo-kind.sh` should run this.

https://github.com/agent-substrate/substrate/blob/07da819da1e8149859255bf52551887e4040f0bf/hack/run-microvm-demo.sh#L105

docs/dev/onboarding.md
BenTheElder: you don't need to install kind or ko, we manage those with the hack/tools modules. you do however need docker (and lima here) and go + kubectl, see also README.md

docs/dev/onboarding.md
BenTheElder: I think your suggestion should say `vzNat: true`? 👍

docs/dev/onboarding.md
Oneimu: fixed.

docs/dev/onboarding.md
Oneimu: added.

docs/dev/onboarding.md
Oneimu: You're right — done, removed the step. Our branch predated the reorder (the older script staged assets before deploying the control plane, which is how the localhost:9000 entry in the troubleshooting log came about — that entry now only covers running stage-to-rustfs.sh manually)

docs/dev/onboarding.md
```diff
@@ -0,0 +1,375 @@
...
+Substrate decouples compute infrastructure from stateful agent execution:
+
+1. **WorkerPool** — pre-warmed pools of worker pods (running `ateom`) ready to
+   accept workloads.
+2. **ActorTemplate** — reusable definitions of an actor: container images,
+   snapshot policy, sandbox class.
+3. **Actor** — a stateful instance that can be created, activated (*resume*),
+   frozen into object storage (*suspend*), and re-activated on any free worker.
+4. **atenet-router** — ingress gateway that routes HTTP to whichever worker pod
+   currently hosts the actor, resuming it on demand.
```
BenTheElder: we may need to revisit this section, it seems likely to get out of sync with broader docs ... ok to try it for now though.

docs/dev/onboarding.md
```diff
@@ -0,0 +1,375 @@
...
+### Runtime comparison
+
+| Dimension | gVisor | microVM |
+|---|---|---|
+| Isolation boundary | Application kernel / syscall interception (Sentry) | Hardware VM (KVM + Cloud Hypervisor) |
+| Hardware requirement | None | `/dev/kvm` (bare metal, nested virt, or Lima) |
+| Node label | None (default) | `ate.dev/sandboxClass=microvm` (applied automatically by `hack/create-kind-cluster.sh` when KVM is detected) |
+| Runtime assets | `runsc`, fetched per `SandboxConfig` | 5 assets (`cloud-hypervisor`, `virtiofsd`, `vmlinux`, `rootfs.img`, `configuration-clh.toml`) staged into the in-cluster rustfs S3 bucket |
+| Demo namespace | `ate-demo-counter` | `ate-demo-counter-microvm` |
+| Demo template | `ate-demo-counter/counter` | `ate-demo-counter-microvm/counter-microvm` |
```
BenTheElder: This again duplicates information (e.g. the list of sandbox assets), which you also don't really need to know here. Let's focus on how to get a cluster up and running, and avoid duplicating information.

docs/dev/onboarding.md
```diff
@@ -0,0 +1,375 @@
...
+
+## 2. Prerequisites (all paths)
+
+You need `docker`, `go`, and `kubectl`. (`kind` and `ko` are pinned in
+`hack/tools/` and invoked automatically by the hack scripts — no install
+needed.)
+
+```sh
+# Docker: follow https://docs.docker.com/engine/install/ for your distro, then:
```
BenTheElder: This only makes sense on a linux host. You don't need to usermod on macOS / limactl.

CONTRIBUTING.md
```diff
@@ -25,6 +25,13 @@ sign a new one.
 This project follows [Google's Open Source Community
 Guidelines](https://opensource.google/conduct/).
 
+### Set up a local development environment
+
+New to the project? [docs/dev/onboarding.md](docs/dev/onboarding.md) walks
+through bringing up a local Substrate cluster on [kind](https://kind.sigs.k8s.io)
```
BenTheElder: We cover this in the README already, even for end users, but we need a guide for getting uVM up and running because it has higher friction.

docs/dev/onboarding.md
```diff
@@ -0,0 +1,375 @@
...
+## 6. Troubleshooting / friction log
+
+Issues actually hit during onboarding, with root causes and fixes:
+
+| # | Symptom | Root cause | Fix |
+|---|---|---|---|
+| 1 | `permission denied while trying to connect to the docker API` | User not in the `docker` group (or group not active in this session) | `sudo usermod -aG docker "$USER"` then `newgrp docker` or re-login |
+| 2 | `/dev/kvm: permission denied` during the kind KVM probe | Rootless Docker: the probe container's root is remapped to your user, which can't open the device (`660 root:kvm`) | Use rootful Docker, or `sudo chmod 666 /dev/kvm` before `./hack/create-kind-cluster.sh` |
+| 3 | `kubectl: command not found` from install scripts | `kubectl` not installed | Install per [kubernetes.io/docs/tasks/tools](https://kubernetes.io/docs/tasks/tools/) |
+| 4 | `error: the 'aws' CLI is required but was not found in PATH` | `stage-to-rustfs.sh` uses `aws s3 cp` to stage microVM assets | Install the AWS CLI, or use the dockerized alias in §4.1 (see known issue below) |
```
BenTheElder: https://github.com/agent-substrate/substrate/pull/804 will fix this

docs/dev/onboarding.md
```diff
@@ -0,0 +1,375 @@
...
+`hack/tools/` and invoked automatically by the hack scripts — no install
+needed.)
+
+```sh
+# Docker: follow https://docs.docker.com/engine/install/ for your distro, then:
+sudo usermod -aG docker "$USER"
+newgrp docker          # or log out/in to pick up the group
+docker ps              # should succeed without sudo
+
+# Go 1.26+ (see go.mod for the exact version)
```
BenTheElder: You actually only need go 1.21+, because it will automatically fetch the correct go version

https://go.dev/doc/toolchain

docs/dev/onboarding.md
```diff
@@ -0,0 +1,375 @@
...
+go version
+
+# kubectl: https://kubernetes.io/docs/tasks/tools/
+
+# Go-installed binaries land in $(go env GOPATH)/bin — put it on PATH now,
```
BenTheElder: This is standard go setup instructions, let's just point at Go's docs and not inline this. They may change upstream.

docs/dev/onboarding.md
```diff
@@ -0,0 +1,375 @@
...
+export PATH="$(go env GOPATH)/bin:${PATH}"
+echo 'export PATH="$(go env GOPATH)/bin:${PATH}"' >> ~/.bashrc
+
+# Sanity check
+for tool in docker go kubectl; do
+  which "$tool" >/dev/null && echo "OK  $tool" || echo "MISSING  $tool"
+done
+```
+
+## 3. Path A: gVisor on kind (start here)
```
BenTheElder: This is covered by README.md, let's not duplicate.

BenTheElder (commented):
I think we can reframe this as "how to get uVM running locally", information like "what is the architecture of substrate" and "get a kind cluster with substrate running" are already covered by existing guides, duplicating that info just gives us N+1 places to keep up to date.

The guidance for getting uVM running is really valuable though.

Even in the readme for end users OR contributors, we want getting a cluster with substrate to be ~easy. But we can't totally eliminate the KVM friction so we need a guide for that.

Also, instead of working around the aws cli in the scripts in the docs, I filed a PR to address it in the utility scripts.

docs/dev/onboarding.md
```diff
@@ -0,0 +1,375 @@
...
+
+- **Build and test:** `make build`, `make test`, and `make verify` (codegen /
+  lint checks). Root-gated tests run via `hack/run-root-tests.sh` — see
+  [CONTRIBUTING.md](../../CONTRIBUTING.md).
+- **End-to-end tests on your kind cluster:** `hack/run-e2e-kind.sh`.
+- **Repo conventions and layout:** [AGENTS.md](../../AGENTS.md) and
+  [docs/dev/code-layout.md](code-layout.md).
+- **More demos:** [demos/](../../demos/) — multiplexing Claude Code agents,
+  request parking, autoscaled worker pools, and more.
+- **Getting help:** the [ate-dev](https://groups.google.com/g/ate-dev) Google
```
BenTheElder: Again, this is covered in the README, and if/when we update the groups / slack channels we're likely to miss updating it in more places.

Oneimu: Removed.

docs/dev/onboarding.md
Oneimu: Removed rather than risk the drift you predicted.

docs/dev/onboarding.md
Oneimu: Resolved by deletion — the doc no longer makes any Go version claim; prerequisites defer to the README Quickstart.

docs/dev/onboarding.md
Oneimu: Removed. perequisites section now defers to the README.

docs/dev/onboarding.md
Oneimu: Ack — dropped the dockerized-alias workaround from the doc and kept just a one-line "requires the aws CLI" note, which we can remove in a follow-up once #804 merges.

docs/dev/microvm-local.md
```diff
@@ -0,0 +1,181 @@
...
+
+### 1. Install Lima and the Docker CLI
+
+```sh
+brew install lima docker
+```
+
+### 2. Launch Lima with nested virtualization
+
+The guest image is pinned to Ubuntu 25.10 until the kernel issue in the
```
BenTheElder: We should probably specify that this is an issue with *all* kernels > 6.19 with nested virt on arm64, we don't need to go into all the details, but it's not just the current default distro/kernel. You need to use a <6.19 Kernel, 25.10 is just one option.

Also FYI @sfunkenhauser 25.10 is already EOL, we should probably pin to 24.04 LTS, which we can comfortably use for the immediate future
https://endoflife.date/ubuntu

docs/dev/microvm-local.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+## Option B: Apple Silicon macOS via Lima
+
+Lima can run a Linux VM with **nested virtualization**, exposing `/dev/kvm` to
+Docker (and therefore to the kind node) inside the VM. This is a well-trodden
+path — much of Substrate's development happens on macOS via limactl.
+
+> [!IMPORTANT]
+> The first-generation M1 lacks hardware support for ARM nested
```
BenTheElder: I think it may actually be M3+

https://developer.apple.com/documentation/virtualization/vzgenericplatformconfiguration/isnestedvirtualizationsupported#:~:text=Nested%20virtualization%20is%20available%20for%20Mac%20with%20the%20M3%20chip%2C%20and%20later.

CONTRIBUTING.md
```diff
@@ -25,6 +25,14 @@ sign a new one.
 This project follows [Google's Open Source Community
 Guidelines](https://opensource.google/conduct/).
 
+### Set up a local development environment
+
+The [Quickstart (Development)](README.md#quickstart-development) in the README
+covers bringing up a local cluster with the default (gVisor) runtime. To run
+the higher-friction microVM runtime locally — which needs `/dev/kvm`, or Lima
```
BenTheElder: ouch, maybe let's just say "To run MicroVM runtime locally", it _is_ higher friction but let's not slam our own work :^) ... and we hope to continue making that easier

BenTheElder (commented):
thanks, last comments ...

CONTRIBUTING.md
Oneimu: :( yeah that was harsh. 

Reworded.

docs/dev/microvm-local.md
Oneimu: Good find! thanks.

docs/dev/microvm-local.md
Oneimu: Ack.

sfunkenhauser (approved):

PR #746, opened 2026-08-04
manifests: centralize the OTLP endpoint in an ate-otel-config ConfigMap

### LINKED ISSUE #745: OTLP collector endpoint is defined in 9 places
`OTEL_EXPORTER_OTLP_ENDPOINT` is hardcoded in the four workload manifests, then re-patched five more times in the kind overlays:

```
manifests/ate-install/ate-api-server.yaml:123
manifests/ate-install/ate-controller.yaml:95
manifests/ate-install/atelet.yaml:108
manifests/ate-install/atenet-router.yaml:157
manifests/ate-install/kind/kustomization.yaml:60,80,98,116
manifests/ate-install/kind/atelet/kustomization.yaml:42
```

Pointing a cluster at a different collector means editing nine spots across six files, and knowing which install path renders which.

Two of the targeted redeploy flags apply the base manifests raw, with no
kustomize, even in kind mode:

- `deploy_ate_apiserver` → `run_ko apply -f manifests/ate-install/ate-api-server.yaml`
- `deploy_atenet` → `run_ko resolve -f manifests/ate-install/atenet-router.yaml`

So on a kind cluster:

```
hack/install-ate-kind.sh --deploy-ate-system     # endpoint = otel-system       (right)
hack/install-ate-kind.sh --deploy-ate-apiserver  # endpoint = gke-managed-otel  (wrong)
```

The second command silently repoints ateapi at a Service that doesn't exist in kind. No error, no failed pod — traces and metrics just stop showing up. You find out when a dashboard is empty.

`deploy_atelet` doesn't have the bug because it branches on kind and renders the overlay. That inconsistency is the tell.

## Suggested fix

One checked-in ConfigMap per environment (`ate-otel-config`), pulled in by every workload via `envFrom`. Two definitions instead of nine, and the targeted redeploys just apply the right ConfigMap for the environment.

`configMapGenerator` won't work here: the GKE path applies `manifests/ate-install/` as a raw directory and the targeted paths apply single files, so the name has to stay stable — no hash suffix.

There's precedent already: `ate-api-server.yaml` pulls in the optional `ate-api-server-envvars` ConfigMap exactly this way.

git286:
OTEL_EXPORTER_OTLP_ENDPOINT is written in 9 places: hardcoded inline in 4 base workload manifests, then re-patched in 5 spots across the kind overlays. Changing the collector address means editing all 9 and knowing which install path renders which.

Replace them with one checked-in ConfigMap per environment, consumed by every component via envFrom:

  manifests/ate-install/ate-otel-config.yaml       (GKE)
  manifests/ate-install/kind/ate-otel-config.yaml  (kind, same name)

The GKE copy is listed in manifests/ate-install/base, so token-client, agentgateway and agentgateway-token-client all inherit it; the kind overlay lists its own copy of the same ConfigMap name. No overlay builds on both, so the two never collide.

A kustomize-only fix does not work here. The GKE path applies the base directory raw and hack/install-ate.sh's targeted redeploys apply single files with no Kustomize, so the mechanism has to survive `kubectl apply -f <one-file>` -- which rules out configMapGenerator (hash-suffixed names) and replacements. envFrom on a stable name reaches every path.

deploy_ate_system applies the ConfigMap before the rendered bundle, the same way it already applies the namespace. A container whose envFrom target is missing does not start, and a raw directory apply is ordered by filename, so ate-api-server.yaml and ate-controller.yaml would otherwise be created first and sit in CreateContainerConfigError until the ConfigMap caught up.

This also fixes a latent bug in the targeted redeploys. deploy_ate_apiserver, deploy_atelet and deploy_atenet apply raw files even in kind mode, silently reverting the endpoint to the GKE value. They now apply the environment's ConfigMap via a new apply_otel_config helper, which picks the file by ATE_INSTALL_KIND rather than applying the base copy unconditionally -- applying the base one on kind would break telemetry for every component at once.

atenet-router needs no special handling: since a98f85b1 its --otlp-collector-address defaults to $OTEL_EXPORTER_OTLP_ENDPOINT, so Envoy's own spans follow the ConfigMap along with everything else.

OTEL_TRACES_SAMPLER deliberately stays an inline kind patch on ate-api-server rather than moving into the ConfigMap. The ConfigMap is shared by every component via envFrom, so putting parentbased_always_on there would pin the router and the rest of the control plane to 100% and undo the per-component ratios from 15eecd02.

Note that a ConfigMap edit does not roll the consuming pods the way an inline env change did, since the pod template is unchanged. Callers must follow a change with `kubectl rollout restart`. This is documented in both ConfigMaps, in docs/observability.md, and in the tracing best practices, which previously told authors to hardcode the variable.

Fixes #745

- [ ] Tests pass

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -189,6 +189,8 @@ Telemetry is emitted the same way everywhere; only the backend differs between a
 > In Kind, `ateapi`, `atelet`, `ate-controller`, and `atenet-router` are pointed at the in-cluster collector, and the controller propagates the endpoint to the ateom worker pods it creates, so all component telemetry lands locally.
 >
 > Every component reads that endpoint from the shared `ate-otel-config` ConfigMap ([`manifests/ate-install/ate-otel-config.yaml`](../manifests/ate-install/ate-otel-config.yaml), with a Kind replacement of the same name under [`manifests/ate-install/kind/`](../manifests/ate-install/kind/ate-otel-config.yaml)). Editing it does not restart the pods that consume it — follow a change with `kubectl rollout restart`.
+>
+> ateom workers don't read the ConfigMap at all — `ate-controller` copies the value into each worker pod at creation. A new endpoint reaches them only once the controller itself restarts, and that restart then rolls every WorkerPool Deployment, replacing the running workers along with the actors on them.
 
 ---
 
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -187,6 +187,8 @@ Telemetry is emitted the same way everywhere; only the backend differs between a
 | Dashboards | Not supported | Google Cloud Monitoring (see [Dashboards](#5-dashboards)) |
 
 > In Kind, `ateapi`, `atelet`, `ate-controller`, and `atenet-router` are pointed at the in-cluster collector, and the controller propagates the endpoint to the ateom worker pods it creates, so all component telemetry lands locally.
+>
+> Every component reads that endpoint from the shared `ate-otel-config` ConfigMap ([`manifests/ate-install/ate-otel-config.yaml`](../manifests/ate-install/ate-otel-config.yaml), with a Kind replacement of the same name under [`manifests/ate-install/kind/`](../manifests/ate-install/kind/ate-otel-config.yaml)). Editing it does not restart the pods that consume it — follow a change with `kubectl rollout restart`.
 
 ---
 
--- a/manifests/ate-install/base/kustomization.yaml
+++ b/manifests/ate-install/base/kustomization.yaml
@@ -15,6 +15,9 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
+# ate-otel-config.yaml carries the OTLP settings every component above consumes
+# via envFrom. The kind overlay lists its own copy of the same ConfigMap name
+# instead of building on this directory, so the two never collide.
 resources:
   - ../ate-api-server.yaml
   - ../ate-controller.yaml
@@ -23,3 +26,4 @@ resources:
   - ../atenet-router.yaml
   - ../valkey.yaml
   - ../pod-certificate-controller.yaml
+  - ../ate-otel-config.yaml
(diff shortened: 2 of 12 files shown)


manifests/ate-install/base/kustomization.yaml
```diff
@@ -23,3 +26,4 @@ resources:
   - ../atenet-router.yaml
   - ../valkey.yaml
   - ../pod-certificate-controller.yaml
+  - ../ate-otel-config.yaml
```
krisztianfekete, line 29: Maybe not for this PR, but `AGENTGATEWAY_OTLP_ADDRESS` is also hardcoded in multiple places.

docs/observability.md
```diff
@@ -192,6 +187,8 @@ Telemetry is emitted the same way everywhere; only the backend differs between a
 | Dashboards | Not supported | Google Cloud Monitoring (see [Dashboards](#5-dashboards)) |
 
 > In Kind, `ateapi`, `atelet`, `ate-controller`, and `atenet-router` are pointed at the in-cluster collector, and the controller propagates the endpoint to the ateom worker pods it creates, so all component telemetry lands locally.
+>
+> Every component reads that endpoint from the shared `ate-otel-config` ConfigMap ([`manifests/ate-install/ate-otel-config.yaml`](../manifests/ate-install/ate-otel-config.yaml), with a Kind replacement of the same name under [`manifests/ate-install/kind/`](../manifests/ate-install/kind/ate-otel-config.yaml)). Editing it does not restart the pods that consume it — follow a change with `kubectl rollout restart`.
```
krisztianfekete, line 191: Maybe we can add that ateom workers get these values at pod creation, so a rollout-restart of the controller doesn't change existing workers (until they are restarted).

krisztianfekete (approved):
This LGTM, added some non-blocking comments!

manifests/ate-install/base/kustomization.yaml
git286: Thanks! Will leave it as a follow-up.

docs/observability.md
git286: Added. One correction while verifying it: the workers don't need a separate restart. The controller SSA-applies the WorkerPool Deployment on every reconcile with the endpoint as a literal, so a controller restart changes the pod template and rolls the workers automatically. Confirmed on kind.

juli4n (approved):

PR #747, opened 2026-08-04
Validate install prerequisites

brendandburns:
Add checks for three prerequisites which I hit when I was installing using a fresh system.

### DIFF
--- a/hack/check-prerequisites.sh
+++ b/hack/check-prerequisites.sh
@@ -1,0 +1,146 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 The Agent Substrate Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_command() {
+  local command_name="$1"
+  if ! command -v "${command_name}" >/dev/null 2>&1; then
+    echo "Error: ${command_name} is required but was not found in PATH." >&2
+    return 1
+  fi
+}
+
+check_git() {
+  require_command git
+  if ! git --version >/dev/null 2>&1; then
+    echo "Error: git is installed but not working (failed to run 'git --version')." >&2
+    return 1
+  fi
+}
+
+check_git_worktree() {
+  local status=""
+  status="$(git status --porcelain --untracked-files=normal)"
+  if [[ -n "${status}" ]]; then
+    echo "Error: this operation requires a clean Git worktree, but local changes were found:" >&2
+    echo "${status}" >&2
+    echo "Hint: commit, stash, or remove these changes before running the installer." >&2
+    return 1
+  fi
+}
+
+check_go() {
+  require_command go
+
+  if ! go version >/dev/null 2>&1; then
+    echo "Error: go is installed but not working (failed to run 'go version')." >&2
+    return 1
+  fi
+}
+
+check_go_tool_ko() {
+  if ! ./hack/run-tool.sh ko version >/dev/null 2>&1; then
+    echo "Error: ko is required but not working through hack/run-tool.sh." >&2
+    echo "Hint: run './hack/run-tool.sh ko version' to verify your setup." >&2
+    return 1
+  fi
+}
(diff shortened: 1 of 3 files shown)


hack/install-ate.sh
```diff
@@ -55,6 +60,135 @@ function log_step() {
...
+    echo "Error: ${command_name} is required but was not found in PATH." >&2
+    return 1
+  fi
+}
+
+check_git_worktree() {
+  local status=""
+  status="$(git status --porcelain --untracked-files=normal)"
+  if [[ -n "${status}" ]]; then
+    echo "Error: ko requires a clean Git worktree, but local changes were found:" >&2
```
thockin: s/ko/this operation/ ?

hack/install-ate.sh
```diff
@@ -55,6 +60,135 @@ function log_step() {
...
+    echo "${status}" >&2
+    echo "Hint: commit, stash, or remove these changes before running the installer." >&2
+    return 1
+  fi
+}
+
+check_go() {
+  if ! command -v go >/dev/null 2>&1; then
```
thockin: `require_command go` ?

hack/install-ate.sh
```diff
@@ -55,6 +60,135 @@ function log_step() {
...
+
+  if ! go version >/dev/null 2>&1; then
+    echo "Error: go is installed but not working (failed to run 'go version')." >&2
+    return 1
+  fi
+}
+
+check_go_tool_ko() {
+  if ! (
+    cd "${ROOT}/hack/tools/ko" &&
```
thockin: given that this is in the repo, in a tool that we manage, how would it possibly fail here ?

Also `./hack/run-tool.sh ko`

thockin (commented):
Thanks for this

hack/install-ate.sh
brendandburns: This failed for me when I tried in a clean directory. Probably because my gopath wasn't set correctly or somesuch. I'm open to other validation that makes sure that this works, but `go tool ko` didn't work for me out of the box.

brendandburns:
@thockin sorry for the delays here, rebased, comments addressed, cleaned up a little, please take another look

hack/install-ate.sh
```diff
@@ -16,6 +16,14 @@
 
 set -o errexit -o nounset -o pipefail
 
+SCRIPT_DIR="${BASH_SOURCE[0]%/*}"
```
thockin, line 19: This is getting nitpicky but I think you can replace all this with:

```
ROOT="$(check_git && git rev-parse --show-toplevel)"
```

It seems to work (I know bash is persnickety about when errexit applies)

```
$ demosh

$ PS1="outer-shell $ "
outer-shell $ demosh

$ PS1="inner-shell $ "

inner-shell $ require_command() {   local command_name="$1";   if ! command -v "${command_name}" >/dev/null 2>&1; then     echo "Error: ${command_name} is required but was not found in PATH." >&2;     return 1;   fi; }

inner-shell $ check_git() {   require_command git; }

inner-shell $ check_foo() {   require_command foo; }

inner-shell $ unset ROOT

inner-shell $ ROOT="$(check_git && git rev-parse --show-toplevel)"; echo $?; echo $ROOT
0
/home/thockin/src/substrate

inner-shell $ ROOT="$(check_foo && git rev-parse --show-toplevel)"; echo $?; echo $ROOT
Error: foo is required but was not found in PATH.
1

inner-shell $ set -o errexit

inner-shell $ ROOT="$(check_foo && git rev-parse --show-toplevel)"; echo $?; echo $ROOT
Error: foo is required but was not found in PATH.

outer-shell $ 
```

I'll merge without, but we can fixup later.

hack/check-prerequisites.sh
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env bash
+
+# Licensed under the Apache License, Version 2.0 (the "License");
```
thockin, line 5: Fails boilerplate check because missing copyright.

```
# Copyright 2026 The Agent Substrate Authors
```

?

brendandburns: Can you comment on this? The boilerplate appears to validate that it is "Copyright Google" which I'm not comfortable with since I'm not a Google Employee. It doesn't seem that "The Agent Substrate Authors" is currently recognized?

thockin: Oh, yeah, that's F'ed up.  Either you can fix the script in this PR or I can fix it in the next few days when I get a second.  I hate seeing this simple PR bounce over and over.

brendandburns: Ok, I will send a separate targetted PR and then rebase.

brendandburns: @thockin https://github.com/agent-substrate/substrate/pull/1181

PR #749, opened 2026-08-04
benchmarking: add observability benchmarking for the OTel collector

JeffLuoo:
Part of #563

Substrate had no way to measure how much telemetry it sends. This PR adds one, and builds it from the parts the benchmark suite already has.

The managed collector sends its self-metrics to Cloud Monitoring only, and gives no data for each source. Thus you cannot see which component makes the volume. A second collector, the telemetry meter, counts the spans and the datapoints by `service.name`.

## Contents

- **`benchmarking/telemetry/meter.yaml`** — the meter. A tee: it counts the telemetry by service, then sends it on to the managed collector. Thus one run gives the volume for each service and the cost of the load together.
- **`benchmarking/telemetry/README.md`** — the method: install, measure, read, remove.
- **`hack/install-ate.sh --otlp-endpoint`** — sends all control plane telemetry to the meter. Each component reads the endpoint from the `ate-otel-config` ConfigMap through `envFrom`, and `ate-controller` copies the value to the ateom worker pods, thus one patch is sufficient for all of them.
- **`benchmarking/workloads/deploy.sh --otlp-endpoint`** — the same for actor containers, which substrate configures separately.
- **`benchmarking/monitoring.yaml`** — scrapes the meter and cAdvisor.
- **`benchmarking/observability.md`** — prerequisites and the scenario ladder.
- **`docs/dev/best-practices/otel-collector.md`** — the volume model and the size guidance, corrected for the `ParentBased` defaults from #711.

## Correction

The `glutton` ActorTemplate had no `env`. Substrate puts no OTLP configuration in an actor container, thus the exporter used the SDK default `localhost:4317`, no process listens on that address in the sandbox, and substrate discarded each actor span with no error message. The template now sets `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_RESOURCE_ATTRIBUTES`.

`ActorTemplate.spec` is immutable, thus `deploy.sh --deploy` deletes the templates before it applies them.

## How I verified this

On a GKE cluster with managed OTel. The sequence:

```bash
# 1. Start from a clean state.
./hack/install-ate.sh --delete-benchmarks
./hack/install-ate.sh --delete-ate-system

# 2. Install substrate from this branch.
./hack/install-ate.sh --deploy-ate-system

# 3. Install the meter and Prometheus.
kubectl apply -f benchmarking/telemetry/meter.yaml
kubectl apply -f benchmarking/monitoring.yaml

# 4. Send the control plane to the meter.
METER=http://telemetry-meter.benchmarking.svc.cluster.local:4317
./hack/install-ate.sh --deploy-ate-system --otlp-endpoint "${METER}"

# 5. Install the workloads and locust, with the actors on the meter also.
./benchmarking/deploy_locust.sh --deploy --worker-count 10 \
    --otlp-endpoint "${METER}"

# 6. Read the volume.
kubectl port-forward -n benchmarking svc/prometheus 9090:9090
```

## Result

The idle floor, from one measurement on 10 workers:

| Service | datapoints/min |
|---|---|
| `ate-controller` | 234.8 |
| `ateapi` | 54.3 |
| `atelet` | 44.9 |
| `atenet-router` | 10.4 |
| `ateom-gvisor` | 5.2 |
| `glutton-actor` | 4.2 |
| **Total** | **353.7** |

The `glutton-actor` row is the correction above. Before it, actor telemetry could not arrive at all.

Positive control for the same window: `otelcol_receiver_accepted_metric_points` 10.5/s, and each `otelcol_receiver_refused_*` counter at zero. Thus the zero values in the table are true zeros, and not data that did not arrive.

Two values differ from the earlier manual measurement. `ate-controller` is now the largest source, and `ateom-gvisor` is no longer zero. The new imagecache counters from #831 are a probable cause. This difference is the reason to keep results out of the repository.

### Traces

Trace volume follows the root sample decision, not the rate of each component. `atelet` and `ateom-gvisor` sent zero spans in each run, because they inherit the decision that `ateapi` makes and start no trace of their own. At idle each component is near zero: a trace needs an operation, but a metric pushes on a timer.

This run gives no span volume for a usual load. The actors crashed on suspend, and the failed retries made an artificial rate of 13.7k spans/s. That rate is not a measurement of substrate. It does show that the meter accepted the spans with no refusals.

## Follow-up

A separate PR adds the ladder to `automation/tests.yaml` and writes the results as artifacts of each run. The open items are at the end of `observability.md`.

- [x] Tests pass

### DIFF
--- a/benchmarking/monitoring.yaml
+++ b/benchmarking/monitoring.yaml
@@ -50,6 +50,33 @@ subjects:
   name: prometheus
   namespace: benchmarking
 ---
+# A node is a cluster-scoped resource. Thus the namespaced Role above cannot
+# give access to node discovery or to the kubelet proxy. The cadvisor scrape
+# job needs the two permissions.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: prometheus-benchmarking-nodes
+rules:
+- apiGroups: [""]
+  resources: ["nodes", "nodes/proxy", "nodes/metrics"]
+  verbs: ["get", "list", "watch"]
+- nonResourceURLs: ["/metrics"]
+  verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: prometheus-benchmarking-nodes
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: prometheus-benchmarking-nodes
+subjects:
+- kind: ServiceAccount
+  name: prometheus
+  namespace: benchmarking
+---
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -82,6 +109,64 @@ data:
             regex: ([^:]+)(?::\d+)?;(\d+)
             replacement: $1:$2
             target_label: __address__
+
+      # The telemetry-meter gives its substrate_* counts on port 8889, which
+      # the annotation job above collects. It gives its own otelcol_* counters
+      # on port 8888. The annotation relabel can name only one port, thus
+      # scrape port 8888 here. When a service reports zero volume,
+      # otelcol_receiver_accepted_* shows the difference between "sent nothing"
+      # and "arrived nowhere".
+      - job_name: 'telemetry-meter-self'
+        kubernetes_sd_configs:
+          - role: pod
+            namespaces:
+              names:
+                - benchmarking
+        relabel_configs:
+          - source_labels:
+              - __meta_kubernetes_pod_label_app
+              - __meta_kubernetes_pod_container_port_number
+            action: keep
+            regex: telemetry-meter;8888
+          - source_labels: [__meta_kubernetes_pod_name]
+            action: replace
+            target_label: pod
+
+      # The container memory directly from the cAdvisor endpoint of the
+      # kubelet. `kubectl top` reads metrics-server, which reports an average
+      # across its own window and hides short peaks.
+      # container_memory_working_set_bytes is the value that the OOM killer
+      # uses.
+      - job_name: 'kubernetes-cadvisor'
+        scheme: https
+        tls_config:
+          ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+          insecure_skip_verify: true
+        bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
+        kubernetes_sd_configs:
+          - role: node
+        relabel_configs:
+          - action: labelmap
+            regex: __meta_kubernetes_node_label_(.+)
+          - target_label: __address__
+            replacement: kubernetes.default.svc:443
+          - source_labels: [__meta_kubernetes_node_name]
+            regex: (.+)
+            target_label: __metrics_path__
+            replacement: /api/v1/nodes/$1/proxy/metrics/cadvisor
+        metric_relabel_configs:
+          # cAdvisor sends a very large number of metrics. Keep only the
+          # metrics that the benchmarks read.
+          - source_labels: [__name__]
+            action: keep
+            regex: container_(memory_working_set_bytes|cpu_usage_seconds_total|spec_memory_limit_bytes)
+          # Drop the series that have an empty container label, not the label
+          # itself. cAdvisor gives a rollup for the whole pod cgroup next to
+          # the series for each container. Keeping both counts the memory of
+          # each pod two times.
+          - source_labels: [container]
+            action: drop
+            regex: ""
 ---
 apiVersion: apps/v1
 kind: Deployment
--- a/benchmarking/observability.md
+++ b/benchmarking/observability.md
@@ -1,0 +1,72 @@
+# Observability benchmarking
+
+This page shows how much telemetry substrate sends, and what that telemetry
+costs the OTel collector.
+
+- To make a measurement, read [telemetry/README.md](telemetry/README.md).
+- For the volume model and the collector size guidance, read
+  [OTel Collector](../docs/dev/best-practices/otel-collector.md).
+- This page contains the scenario ladder.
+
+This page holds no results. The automation writes the values of each run to
+GCS. A result in this repository becomes incorrect after a change to the
+sample rates or to the instruments, and no reader can then see which copy is
+correct.
+
+---
+
+## 1. Scenario ladder
+
+`SleepUser` ([`locust/tests/sleep.py`](locust/tests/sleep.py)) does one suspend
+operation and one resume operation for each task. Thus
+`cycles/sec ≈ users / (1.0s + avg_wait_time)`. Change the number of users. Do
+not change the wait time.
+
+| # | Scenario | Duration | Setup |
+|---|---|---|---|
+| **S0** | Idle floor | 5m | No load |
+| **S1** | Suspend and resume sweep | 3m for each step | `SleepUser`, `--trace-probability 0.1`, users 5 → 15 → 30 |
+| **S2** | Sample-rate sweep | 3m for each step | The number of users stays the same. `--trace-probability` 0 → 0.1 → 1.0 |
+| **S3** | Soak | 15m | 80% of the maximum load above |
+
+Do not make a step shorter than 3 minutes. The metrics push interval is 60
+seconds. A shorter step gives less than three datapoints for each series. You
+cannot then tell the difference between a trend and noise.
+
+Do these runs with the headless runner in
+[automation/tests.yaml](automation/tests.yaml). In headless mode, `runner.py`
+sends `--trace-probability` to the boomer worker at start, and writes the value
+in the run config header. Thus each run records the rate that it used.
+
+The locust web UI is for manual examination only. Two conditions apply there.
+The `boomer-glutton` sidecar makes its own load for each user class that you
+select in the form. Also, the form changes the sample rate of the boomer worker
+but not of the Python workers: `locust.yaml` gives boomer `--master-web-port`,
+thus boomer reads `/boomer-config` from the master at each spawn message.
+
+**Pass criteria.** `otelcol_receiver_refused_*` must be zero.
+`otelcol_exporter_enqueue_failed_*` must be zero. `queue_full` on the client
+must be zero. The number of collector replicas must stay below `maxReplicas`.
+In S3, the slope of the collector working set and the slope of
+`otelcol_exporter_queue_size` must be equivalent to zero.
+
+Examine the positive control also. `otelcol_receiver_accepted_*` must be more
+than zero for each service that must send data. If you do not examine it, an
+exporter that sends nothing looks the same as a good run.
+
+---
+
+## 2. Open items
+
+- Add the ladder to [automation/tests.yaml](automation/tests.yaml). Write the
+  results as artifacts of each run.
+- Measure the collector CPU and memory from the working set. Do not use
+  `kubectl top`, which calculates an average and hides peaks.
+- Record `otelcol_receiver_accepted_*` with each volume table as a positive
+  control.
+- Record the number of collector replicas with the CPU and the memory. Without
+  the number of replicas, a collector with no capacity and a collector with
+  much capacity give the same table.
+- Measure the actor volume. The `glutton` template now sets an OTLP endpoint,
+  thus actor telemetry arrives for the first time.
+- Measure the DaemonSet configuration in the collector size guidance.
--- a/benchmarking/telemetry/README.md
+++ b/benchmarking/telemetry/README.md
@@ -128,7 +128,7 @@ the pass criteria:
 sum(rate(otelcol_receiver_refused_spans[5m]))   # must be 0
 sum(rate(otelcol_exporter_sent_spans[5m]))      # must be more than 0
 max_over_time(otelcol_exporter_queue_size[5m])  # must stay level
-absent(otelcol_exporter_queue_size)             # must give no result
+absent(otelcol_exporter_sent_spans)             # must give no result
 ```
 
 Do not use `otelcol_exporter_send_failed_spans` alone. The default
--- a/benchmarking/telemetry/README.md
+++ b/benchmarking/telemetry/README.md
@@ -1,0 +1,184 @@
+<!--
+Copyright 2026 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Telemetry meter
+
+The meter measures how much telemetry substrate sends, and which component
+sends it.
+
+The managed collector sends its self-metrics to Cloud Monitoring only, and
+gives no data for each source. Thus [`meter.yaml`](meter.yaml) is a second
+collector that counts the spans and the datapoints by `service.name`. The meter
+is a tee: it counts the data and then sends it on to the managed collector, thus
+the managed collector keeps the load of the run.
+
+For the prerequisites and the scenario ladder, read
+[observability.md](../observability.md). For the volume model, read
+[OTel Collector](../../docs/dev/best-practices/otel-collector.md).
+
+## Make a measurement
+
+Install the meter, then send the control plane to it:
+
+```bash
+kubectl apply -f benchmarking/telemetry/meter.yaml
+
+METER=http://telemetry-meter.benchmarking.svc.cluster.local:4317
+./hack/install-ate.sh --deploy-ate-system --otlp-endpoint "${METER}"
+```
+
+`--otlp-endpoint` patches the `ate-otel-config` ConfigMap and restarts the
+workloads that read it. One patch is sufficient for `ateapi`, `ate-controller`,
+`atelet`, and `atenet-router`, because each one reads the ConfigMap through
+`envFrom`. `ate-controller` also copies the values to the ateom worker pods
+that it creates.
+
+The actor containers are different. Substrate puts no OTLP configuration in
+them, thus they use the `env` of the ActorTemplate:
+
+```bash
+./benchmarking/workloads/deploy.sh --deploy
+```
+
+Without `--otlp-endpoint`, the script reads the endpoint from the
+`ate-otel-config` ConfigMap. The step above already changed that ConfigMap,
+thus the actors follow the control plane to the meter. Give `--otlp-endpoint`
+only to send the actors to a different address than the control plane.
+
+Keep the load generator on the usual collector. If you do not, the meter counts
+the telemetry of the load generator as telemetry from substrate.
+
+## Read the numbers
+
+[`monitoring.yaml`](../monitoring.yaml) scrapes the meter. Read the values
+through Prometheus. Do not compare raw scrapes manually.
+
+```bash
+kubectl apply -f benchmarking/monitoring.yaml
+kubectl port-forward -n benchmarking svc/prometheus 9090:9090
+```
+
+| To find | Query |
+|---|---|
+| Spans/sec for each service | `sum by (svc) (rate(substrate_spans_total[5m]))` |
+| Datapoints/min for each service | `60 * sum by (svc) (rate(substrate_datapoints_total[5m]))` |
+| If the data arrived | `sum(rate(otelcol_receiver_accepted_spans[5m]))` |
+| If the collector rejected the data | `sum(rate(otelcol_receiver_refused_spans[5m]))` |
+
+Keep the range at three times the 60s push interval or more. `[5m]` is a good
+default. A range less than 3m gives noise.
+
+Always read the last two counters. A service that reports zero volume gives no
+data by itself: `accepted` must be more than zero and `refused` must be zero
+for the same window. If they are not, the zero shows that the data did not
+arrive, not that the service sent no data.
+
+## The meter is a tee
+
+The meter counts the telemetry and then sends it on to the managed collector.
+Thus one run gives both measurements: the volume for each service, from the
+meter, and the cost of the load, from the managed collector.
+
+```promql
+max_over_time(
+  container_memory_working_set_bytes{namespace="gke-managed-otel", container!=""}[5m]
+)
+```
+
+Use the working set for the memory. `kubectl top` reads metrics-server, which
+calculates an average across its own window and hides short peaks. The working
+set is the value that the OOM killer uses. Divide the working set by
+`container_spec_memory_limit_bytes` to get the fraction of the limit in use.
+
+Record the number of replicas with the CPU and the memory. The managed
+collector is an HPA-scaled Deployment. Without the number of replicas, a
+collector at its maximum replica count and a collector with much margin give
+the same table.
+
+### Three conditions of the tee
+
+**The managed collector attributes the telemetry to the meter.** It applies
+`k8sattributes` with `from: connection`, and each forwarded item now comes from
+the meter pod. The volume and the CPU stay correct, but a query that groups by
+pod on the managed side does not. Group by `service.name` from the resource,
+which stays correct after the hop.
+
+**The shape of the load changes.** Usually many processes each keep a
+connection to the managed collector. With the tee there is one sender and
+larger batches. Thus the total is correct, but the distribution across the
+replicas is not, and the effect of connection stickiness disappears.
+
+**The meter can become the limit.** It is one replica. Add its own counters to
+the pass criteria:
+
+```promql
+sum(rate(otelcol_receiver_refused_spans[5m]))   # must be 0
+sum(rate(otelcol_exporter_sent_spans[5m]))      # must be more than 0
+max_over_time(otelcol_exporter_queue_size[5m])  # must stay level
+absent(otelcol_exporter_queue_size)             # must give no result
+```
+
+Do not use `otelcol_exporter_send_failed_spans` alone. The default
+`retry_on_failure.max_elapsed_time` is 5 minutes, thus the exporter retries for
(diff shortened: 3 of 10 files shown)


benchmarking/observability.md
```diff
@@ -0,0 +1,336 @@
...
+```
+
+Scrape `:8889` for `substrate_spans_total{svc=…}` and
+`substrate_datapoints_total{svc=…}`; take two scrapes and divide by the window.
+**Use ≥3 minutes** — at a 60s push interval anything shorter is noise.
+
+`ate-controller` propagates the endpoint to worker pods, so this covers actors
+too, at the cost of rolling the pool. Scale the pool down first if it is large.
+Leave the load generator pointed at the *managed* collector so its own telemetry
+stays out of the numbers.
```
krisztianfekete: > so this covers actors too

It doesn't cover actors, right? See 3.5

benchmarking/telemetry/meter.yaml
```diff
@@ -0,0 +1,167 @@
...
+            statements:
+              - set(attributes["svc"], resource.attributes["service.name"])
+
+    connectors:
+      count:
+        spans:
+          substrate_spans:
+            description: Spans received, by emitting service.
+            attributes:
+              - key: svc
```
krisztianfekete, line 92: Can you please set `default_value` to make sure we don't miss anything due to missing attributes?

benchmarking/observability.md
```diff
@@ -0,0 +1,336 @@
...
+## 3. Results
+
+### 3.1 Idle floor
+
+| Service | Processes | datapoints/min | per process |
+|---|---|---|---|
+| `atelet` | 27 (DaemonSet) | 81 | 3 |
+| `ateapi` | 2 | 23 | 12 |
+| `atenet-router` | 1 | 19 | 19 |
+| `ateom-gvisor` | 30 | **0** | 0 |
```
krisztianfekete: It's worth using `otelcol_receiver_accepted_spans`, so we can distinguish what was never sent vs. what never arrived.

benchmarking/observability.md
```diff
@@ -0,0 +1,336 @@
...
+METER=http://telemetry-meter.benchmarking.svc.cluster.local:4317
+for d in deployment/ate-api-server deployment/atenet-router \
+         deployment/ate-controller daemonset/atelet; do
+  kubectl set env -n ate-system $d OTEL_EXPORTER_OTLP_ENDPOINT=$METER
+done
+```
+
+Scrape `:8889` for `substrate_spans_total{svc=…}` and
+`substrate_datapoints_total{svc=…}`; take two scrapes and divide by the window.
+**Use ≥3 minutes** — at a 60s push interval anything shorter is noise.
```
krisztianfekete: We have Prometheus, let's just do:
```
sum by (svc) (rate(substrate_spans_total[5m]))
```

benchmarking/observability.md
```diff
@@ -0,0 +1,336 @@
...
+replicas — not a per-actor or per-workload-pod cost.
+
+| Condition | Collector CPU | Collector memory |
+|---|---|---|
+| Idle, 30 worker pods | 2–3m | 85–107Mi |
+| Idle, **1500** worker pods | 15–25m | 140–164Mi |
+| Substrate load, ~700 spans/s | 4–16m | 111–141Mi |
+| `telemetrygen` 10k spans/s | 200–410m | ~190Mi |
+| `telemetrygen` 50k spans/s | 700–1600m | ~250Mi |
+| `telemetrygen` 150k spans/s | ~1670m | 744Mi (1283Mi transient) |
```
krisztianfekete: Can we use the standard `container_memory_working_set_bytes` metric to get the actual peaks instead of the ~black box that is kubectl top + metrics-server combo?

benchmarking/telemetry-meter.yaml
```diff
@@ -0,0 +1,167 @@
...
+            description: Metric datapoints received, by emitting service.
+            attributes:
+              - key: svc
+
+    exporters:
+      prometheus:
+        endpoint: 0.0.0.0:8889
+        # Counters are the whole point; do not let them reset silently.
+        resource_to_telemetry_conversion:
+          enabled: false
```
krisztianfekete: `resource_to_telemetry_conversion` is not relevant for counter resets and it's also false. Do we need this?

benchmarking/telemetry/meter.yaml
```diff
@@ -0,0 +1,167 @@
...
+      transform/svc:
+        error_mode: ignore
+        trace_statements:
+          - context: span
+            statements:
+              - set(attributes["svc"], resource.attributes["service.name"])
+        metric_statements:
+          - context: datapoint
+            statements:
+              - set(attributes["svc"], resource.attributes["service.name"])
```
krisztianfekete: Do we need this? We should have the `job` label by default, which should be identical with this.

krisztianfekete (commented):
Added some comments! Do we want to make this repeatable locally (and in CI?) via a script?

benchmarking/telemetry/meter.yaml
JeffLuoo: Good catch.

Done — default_value: unknown on both.

benchmarking/observability.md
JeffLuoo: Done

benchmarking/observability.md
JeffLuoo: Done.

benchmarking/observability.md
JeffLuoo: Updated

benchmarking/telemetry-meter.yaml
JeffLuoo: Removed

benchmarking/telemetry/meter.yaml
JeffLuoo: I've added a comment recording the reasoning — happy to switch to honor_labels: true + exported_job if you'd rather not carry the transform

benchmarking/observability.md
JeffLuoo: Correct — it covers ateom-gvisor in the worker pods, not the actor containers. Updated.

JeffLuoo:
> Added some comments! Do we want to make this repeatable locally (and in CI?) via a script?

Might do it in a follow-up and figure out the best way to trigger it. Running it against each CL might be expensive, can consider running it as a daily job..? Can connect with people who work on benchmarking and CI/CD to discuss about it.

JeffLuoo:
Hi Max, can you please review it? It doesn't generate any report yet, just set up things we need to benchmark observability. @maxsmythe

benchmarking/telemetry/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+# Telemetry meter
+
+The meter measures how much telemetry substrate sends, and which component
+sends it.
+
+The managed collector sends its self-metrics to Cloud Monitoring only, and
+gives no data for each source. Thus [`meter.yaml`](meter.yaml) is a second
+collector that counts the spans and the datapoints by `service.name`. The meter
+is a terminal sink: it counts the data and forwards nothing.
```
maxsmythe: Is this true? The config states it tee-s the metric to managed otel

benchmarking/telemetry/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+Keep the range at three times the 60s push interval or more. `[5m]` is a good
+default. A range less than 3m gives noise.
+
+Always read the last two counters. A service that reports zero volume gives no
+data by itself: `accepted` must be more than zero and `refused` must be zero
+for the same window. If they are not, the zero shows that the data did not
+arrive, not that the service sent no data.
+
+## The meter is a tee
```
maxsmythe, line 89: This contradicts the earlier statement that the meter is a sink

benchmarking/telemetry/README.md
```diff
@@ -0,0 +1,170 @@
...
+```promql
+max_over_time(
+  container_memory_working_set_bytes{namespace="gke-managed-otel", container!=""}[5m]
+)
+```
+
+Use the working set for the memory. `kubectl top` reads metrics-server, which
+calculates an average across its own window and hides short peaks. The working
+set is the value that the OOM killer uses. Divide the working set by
+`container_spec_memory_limit_bytes` to get the margin.
```
maxsmythe: "get the margin" you mean "get the % of memory capacity used"?

benchmarking/telemetry/meter.yaml
```diff
@@ -0,0 +1,202 @@
...
+      # and each scrape shows only the last batch.
+      deltatocumulative:
+
+      # The count connector uses the attributes of the span or the datapoint,
+      # not the attributes of the resource. Thus copy service.name to a
+      # location where the connector can read it.
+      #
+      # The Prometheus exporter calculates job and instance from the resource.
+      # But the Prometheus server in benchmarking/monitoring.yaml scrapes
+      # without honor_labels. Thus those labels become exported_job and
```
maxsmythe: We can change "honor_labels" if necessary?

benchmarking/telemetry/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+Record the number of replicas with the CPU and the memory. The managed
+collector is an HPA-scaled Deployment. Without the number of replicas, a
+collector at its maximum replica count and a collector with much margin give
+the same table.
+
+### Three conditions of the tee
+
+**The managed collector attributes the telemetry to the meter.** It applies
```
maxsmythe, line 113: Can this be adjusted in the future? I can imagine a world where knowing the load on specific pods (not just service) is important.

Also, is there any issue with the fact of this overriding? If the meter grabs three metrics:

my_metric{pod=A,svc=Z}
my_metric{pod=B,svc=Z}
my_metric{pod=C,svc=Z}

Will that collapse into overlapping time series? Will that cause an issue? Are there other labels that differentiate the source that aren't clobbered?

benchmarking/monitoring.yaml
```diff
@@ -82,6 +109,60 @@ data:
...
+            target_label: __metrics_path__
+            replacement: /api/v1/nodes/$1/proxy/metrics/cadvisor
+        metric_relabel_configs:
+          # cAdvisor sends a very large number of metrics. Keep only the
+          # metrics that the benchmarks read.
+          - source_labels: [__name__]
+            action: keep
+            regex: container_(memory_working_set_bytes|cpu_usage_seconds_total|spec_memory_limit_bytes)
+          - source_labels: [container]
+            action: drop
```
maxsmythe, line 168: Why drop the container label?

benchmarking/observability.md
```diff
@@ -0,0 +1,72 @@
...
+---
+
+## 1. Scenario ladder
+
+`SleepUser` ([`locust/tests/sleep.py`](locust/tests/sleep.py)) does one suspend
+operation and one resume operation for each task. Thus
+`cycles/sec ≈ users / (1.0s + avg_wait_time)`. Change the number of users. Do
+not change the wait time.
+
+| # | Scenario | Duration | Setup |
```
maxsmythe, line 25: Note that Python doesn't play well with gRPC when the number of users gets high enough, causing seconds worth of latency and throttling QPS. This is why the switch to boomer occurred.

Sleep currently uses Python worker. Might want to switch to glutton user to avoid this.

benchmarking/observability.md
```diff
@@ -0,0 +1,72 @@
...
+
+Examine the positive control also. `otelcol_receiver_accepted_*` must be more
+than zero for each service that must send data. If you do not examine it, an
+exporter that sends nothing looks the same as a good run.
+
+---
+
+## 2. Open items
+
+- Add the ladder to [automation/tests.yaml](automation/tests.yaml). Write the
```
maxsmythe, line 61: The ladder seems like something that locust/runner.py should execute in some way.

The non-linear traffic pattern is something that is ideal to provide as a custom shape (locust/shapes)

Changing trace probability dynamically is an interesting wrinkle.

The Boomer worker does dynamically fetch config from master:

https://github.com/agent-substrate/substrate/blob/febeb7afe7e720e00f4755c71be1e9b1b43d52cf/internal/benchmarking/boomer/dynconfig/dynconfig.go#L89-L109

So it is possible. Reloads occur whenever worker receives a spawn event from locust, though that may not always be an intuitive trigger.

docs/dev/best-practices/otel-collector.md
```diff
@@ -508,16 +508,110 @@ Regardless of topology, these are the ones that tell you it is struggling:
...
+```
+
+Each substrate binary pushes on a `PeriodicReader` at the SDK default of 60s,
+with no jitter (`newMeterProvider` in
+[`serverboot.go`](../../../internal/serverboot/serverboot.go)). Thus processes
+that start together also push together. The request rate changes this term
+only a small quantity. A large increase in the actor lifecycle rate moves the
+datapoint volume very little.
+
+Two terms are easy to count too high. First, an idle worker pod sends **no**
```
maxsmythe: This may not be entirely true anymore... I believe we are forwarding Otel data from ateom via unix sockets?

docs/dev/best-practices/otel-collector.md
```diff
@@ -508,16 +508,110 @@ Regardless of topology, these are the ones that tell you it is struggling:
...
+because of the size of the `k8sattributes` cache:
+
+| Mode | Cache in each instance | When the cluster becomes larger |
+|---|---|---|
+| Deployment | O(pods in the cluster) | Increases with no limit. The HPA cannot decrease it. |
+| DaemonSet, filter scoped to the node | O(pods on the node) | Stays in a limit, because a node holds a limited number of pods |
+
+The DaemonSet is better only **with** `k8sattributes` scoped to the node
+(`filter: node_from_env_var`). Without this scope, a DaemonSet is worse: it
+keeps one full copy of the cluster cache on each node, and not in each replica.
```
maxsmythe: "not in each replica" you mean... "not sharded per-node"?

maxsmythe (changes requested):
This looks awesome! Couple questions that may just need clarificatoin.

benchmarking/telemetry/README.md
```diff
@@ -0,0 +1,170 @@
...
+larger batches. Thus the total is correct, but the distribution across the
+replicas is not, and the effect of connection stickiness disappears.
+
+**The meter can become the limit.** It is one replica. Add its own counters to
+the pass criteria:
+
+```promql
+sum(rate(otelcol_receiver_refused_spans[5m]))       # must be 0
+sum(rate(otelcol_exporter_send_failed_spans[5m]))   # must be 0
+sum(rate(otelcol_exporter_sent_spans[5m]))          # must be more than 0
```
krisztianfekete: `retry_on_failure.max_elapsed_time` is 5m, so these won't increment until we run out of all retries. We'd need `otelcol_exporter_queue_size` and use `absent()`

benchmarking/telemetry/meter.yaml
```diff
@@ -0,0 +1,202 @@
...
+          receivers: [otlp]
+          processors: [transform/svc]
+          exporters: [count]
+        metrics/forward:
+          receivers: [otlp]
+          exporters: [otlp/managed]
+        metrics/counts:
+          receivers: [count]
+          processors: [deltatocumulative]
+          exporters: [prometheus]
```
krisztianfekete, line 142: Can we add memory_limiter, and configure `GOMEMLIMIT`?

benchmarking/workloads/deploy.sh
```diff
@@ -38,6 +38,11 @@ fi
 
 WORKER_COUNT=1
 SANDBOX_CLASS="gvisor"
+# The address to which an instrumented actor container sends its telemetry. The
+# default is the collector that the control plane uses, thus a usual deploy
+# needs no flag. Set this to the telemetry meter to measure the actor volume
+# separately.
+OTLP_ENDPOINT="http://opentelemetry-collector.gke-managed-otel.svc.cluster.local:4317"
```
krisztianfekete: Can we make this work for kind clusters as well?

benchmarking/telemetry/meter.yaml
```diff
@@ -0,0 +1,202 @@
...
+      # then sends it on to the collector that the cluster uses. Thus the
+      # managed collector keeps the load of the run, and its CPU, memory, and
+      # replica count stay correct while you measure the volume.
+      #
+      # To make the meter terminal, delete the two forward pipelines below.
+      # The counts continue to work. A terminal meter keeps the telemetry of a
+      # long run out of Cloud Monitoring, but it makes the resource values of
+      # the managed collector those of an idle collector.
+      otlp/managed:
+        endpoint: opentelemetry-collector.gke-managed-otel.svc.cluster.local:4317
```
krisztianfekete, line 115: Same as in deploy.sh, can we make this work with local clusters as well?

krisztianfekete: Maybe we should also make this use the centralized OTel config introduced in https://github.com/agent-substrate/substrate/pull/746?

docs/dev/best-practices/otel-collector.md
```diff
@@ -508,16 +508,110 @@ Regardless of topology, these are the ones that tell you it is struggling:
...
+
+Make the decision in this sequence:
+
+1. **Do you need trace-complete processing?** Tail sampling and `spanmetrics`
+   need all the spans of a trace in one instance. A DaemonSet breaks the two
+   functions unless you add a second gateway tier. This condition frequently
+   makes the decision.
```
krisztianfekete: > Tail sampling and `spanmetrics` need all the spans of a trace in one instance.

`spanmetrics` doesn't need this.

> unless you add a second gateway tier

`loadbalancingexporter` worth mentioning here.

benchmarking/telemetry/README.md
JeffLuoo: Updated - this is left over from previous commit that it's originally a sink but I updated it.

benchmarking/telemetry/README.md
JeffLuoo: Corrected it and aligned in all places. The new collector is a tee not a sink.

benchmarking/telemetry/README.md
JeffLuoo: Yes. I reworded it to make it clear.

benchmarking/telemetry/README.md
JeffLuoo: Good catch, updated with your recommendation.

benchmarking/telemetry/meter.yaml
JeffLuoo: This is just a tee needed by GKE Managed OTel collector.

This tee collector is not needed for kind cluster and it should only process telemetry data and stream it to the GKE Managed collector, which means this is not a "real" collector. I am open to add `memory_limiter` and configure `GOMEMLIMIT` but just think it's not necessary.

docs/dev/best-practices/otel-collector.md
JeffLuoo: Good catch. And since the architecture is changing, I updated the doc to include less details, and let the benchmark result to tell the volume of telemetry data being sent out from the system.

docs/dev/best-practices/otel-collector.md
JeffLuoo: Updated it to make it clear. Each collector on the node will have "full" copy of cluster state if the filter is not used in daemonset mode.

docs/dev/best-practices/otel-collector.md
JeffLuoo: Updated

benchmarking/observability.md
JeffLuoo: Yeah, I can update it to glutton in a follow-up that adds the ladder to the config.

benchmarking/monitoring.yaml
JeffLuoo: Updated the comment - It doesn't drop the label but timeseries to avoid double count.

benchmarking/telemetry/meter.yaml
JeffLuoo: Rewritten the comment to explain it.

benchmarking/telemetry/README.md
JeffLuoo: I think the easiest way is to expose self-metrics from managed collector so we can curl the metrics endpoint and get it directly. With that, we can get rid of this collector to just fetch self-metrics..

The count connector aggregates rather than emitting duplicates, so three pods of service Z sum into one `substrate_spans_total{svc="Z"}`. That's intended, we count data for each server regardless of how many pods running the same service. This should be a good starting point given the current scale of ate system. But as you also call out the need for per-pod tracking, this is not sufficient, and can be resolved by support scraping self-metrics from managed otel collector.

Nothing is clobbered either. The transform runs only in the count pipelines; the forward pipelines have no processor.

benchmarking/observability.md
JeffLuoo: Agree, and this is already the planned follow-up. This PR focus on getting manifest ready.

benchmarking/telemetry/meter.yaml
JeffLuoo: We don't need the meter.yaml for kind cluster because the otel collector is customized. Instead, I think I can have a follow-up to update the manifest of the otel collector deployed to the kind cluster and integrate it with the benchmark.

I can do it in a follow-up.

benchmarking/workloads/deploy.sh
JeffLuoo: Updated it to read from configmap.

benchmarking/observability.md
maxsmythe: Makes sense, should have said this was an informational comment rather than a blocker.

benchmarking/telemetry/README.md
maxsmythe: > I think the easiest way is to expose self-metrics from managed collector so we can curl the metrics endpoint and get it directly. With that, we can get rid of this collector to just fetch self-metrics..
> 

Sorry I don't quite follow. This count connector tee-s the metrics to managed collector, which it needs to do to gather stats. Are you saying that this arrangement can be altered later?

Are you saying eventually the managed collector can calculate/emit what counter is already emitting, eliminating having counter MITM at all?

> The count connector aggregates rather than emitting duplicates, so three pods of service Z sum into one `substrate_spans_total{svc="Z"}`. That's intended, we count data for each server regardless of how many pods running the same service. This should be a good starting point given the current scale of ate system. But as you also call out the need for per-pod tracking, this is not sufficient, and can be resolved by support scraping self-metrics from managed otel collector.

Makes sense.

> 
> Nothing is clobbered either. The transform runs only in the count pipelines; the forward pipelines have no processor.

Actually, this is my concern. If the manager collector substitutes all origin tagging with that of the counter collector, I was wondering if we wind up with duplicate streams whenever we are collecting the same metric from multiple pods -- they all get re-tagged as coming from counter collector

benchmarking/monitoring.yaml
maxsmythe: One other comment then.... does an empty regex not match all label values (equivalent to ".*"), do we need "^$"?

Not 100% sure what the matching semantics are in this case.

benchmarking/telemetry/README.md
krisztianfekete: > Nothing is clobbered either. The transform runs only in the count pipelines; the forward pipelines have no processor.

> > Actually, this is my concern. If the manager collector substitutes all origin tagging with that of the counter collector, I was wondering if we wind up with duplicate streams whenever we are collecting the same metric from multiple pods -- they all get re-tagged as coming from counter collector

This shouldn't be a problem as we have:
https://github.com/agent-substrate/substrate/blob/2b3a4715c6b7af4debe42c32791ae7307f18ca1b/manifests/ate-install/ate-api-server.yaml#L125-L126

benchmarking/monitoring.yaml
krisztianfekete: Prometheus uses RE2 syntax, so everything is anchored and `regex: ""` matches empty value, this looks good to me.

benchmarking/telemetry/README.md
```diff
@@ -0,0 +1,184 @@
...
+replicas is not, and the effect of connection stickiness disappears.
+
+**The meter can become the limit.** It is one replica. Add its own counters to
+the pass criteria:
+
+```promql
+sum(rate(otelcol_receiver_refused_spans[5m]))   # must be 0
+sum(rate(otelcol_exporter_sent_spans[5m]))      # must be more than 0
+max_over_time(otelcol_exporter_queue_size[5m])  # must stay level
+absent(otelcol_exporter_queue_size)             # must give no result
```
krisztianfekete: Now `absent` is on `otelcol_exporter_queue_size`, but that won't disappear during an outage, `otelcol_exporter_sent_spans` will.
```suggestion
absent(otelcol_exporter_sent_spans)             # must give no result
```

docs/dev/best-practices/otel-collector.md
krisztianfekete: Unix socket is the atelet -> ateom, but ateom is still using network for its telemetry.

benchmarking/telemetry/README.md
JeffLuoo: Thank you Krisztian, I also confirmed it on my cluster and no telemetry data was rejected. send_failed and enqueue_failed never incremented.

> Sorry I don't quite follow. This count connector tee-s the metrics to managed collector, which it needs to do to gather stats. Are you saying that this arrangement can be altered later?

Please ignore my initial comment about it. The managed collector doesn't have processor to count telemetry data aggregated by services and self-metrics also don't have it. We will still need the extra collector, but this is not needed for kind cluster because we can customize the collector as we want.

benchmarking/monitoring.yaml
JeffLuoo: Thanks for your review. Resolve it but feel free to reopen if you have other concern, max.

maxsmythe (approved):
LGTM

benchmarking/telemetry/meter.yaml
krisztianfekete: I these are still beneficial for any perf/loadtesting setups to ensure reliability, but we can add them as a follow-up.

JeffLuoo: I will keep this comment open to allow me track it.

git286 (approved):

Issue #741, opened 2026-08-04
OTLP exporters hardcode WithInsecure(), so the SDK's TLS config never applies


git286:
Related: #563

`serverboot.go:141` builds the trace exporter with an unconditional `otlptracegrpc.WithInsecure()`, and `serverboot.go:182` does the same for metrics. `internal/benchmarking/boomer/trace/trace.go:89` repeats it.

The OTel SDK reads env config first and applies explicit options after, so that one option overrides everything the exporter already knows how to do:

```go
cfg = ApplyGRPCEnvConfigs(cfg)   // OTEL_EXPORTER_OTLP_CERTIFICATE, CLIENT_CERTIFICATE,
for _, opt := range opts {       // CLIENT_KEY, INSECURE, and http:// vs https:// inference
    cfg = opt.ApplyGRPCOption(cfg)   // ...all discarded here
}
```

So `OTEL_EXPORTER_OTLP_ENDPOINT=https://collector:4317` silently exports plaintext. No error, no warning. Already noted in otel-collector.md:458 as a known constraint.

The comment justifies it with GKE managed OTel, which is fair — that collector genuinely has no TLS receiver you can configure. But `serverboot` boots every binary in every environment, and we document a self-managed DaemonSet path (otel-collector.md:64) where TLS and mTLS work fine. A GKE Preview limitation shouldn't be compiled in for everyone.

## Fix

Delete the `WithInsecure()` calls and let scheme inference plus the standard env vars do their job. Log the resolved transport next to the existing "Tracing initialized" line so this is visible without reading source. Then fix the "TLS is not supported" text in otel-collector.md:458 and tracing.md:49.

Not proposing we enable TLS by default or wire pod certs into the exporter — managed OTel can't accept either. This is just about not blocking people whose collector can.

## Compat

No manifest changes needed. Every Go-SDK endpoint we ship already has an explicit `http://` scheme (the four in `ate-install/`, the five under `kind/`), which infers plaintext. Unchanged on GKE and Kind. The scheme-less `AGENTGATEWAY_OTLP_ADDRESS` is agentgateway's, not ours.

One thing to check first: with no `OTEL_EXPORTER_OTLP_ENDPOINT` set, the SDK defaults to `localhost:4317` with `Insecure: false`, so those processes would start attempting TLS. Need to confirm before merging:

- [ ] do ateom-gvisor / ateom-microvm inherit the env from atelet? They call `InitTracing` but have no manifest of their own
[...]

PR #750, opened 2026-08-05
demos: add actor telemetry continuity reference demo (#503)

sanskar-singh-2403:
Add a reference demo for actor telemetry continuity across suspend/resume, the
follow-up work requested in #503.

The demo is a small Go HTTP server that emits OpenTelemetry metrics and trace
spans over OTLP on every request. It reproduces the #503 loss: an OTel push
exporter batches telemetry in memory and flushes on a timer, so when the push
interval (default 60s) is longer than the active window before idle suspension,
telemetry is still buffered in the SDK when the actor is suspended. With
`snapshotsConfig.onPause: Data` (the mode the template uses) Substrate drops
process RAM on pause, so that buffered telemetry is lost.

The mitigation is isolated behind a `flushTelemetry(ctx)` seam that ForceFlushes
both the MeterProvider and TracerProvider. Today it is called from the
self-suspend path via an idle watcher; when the PreSuspend hook (#450) lands,
`flushTelemetry` does not change, only its call-site moves into the PreSuspend
handler. This keeps #450 a call-site move rather than a rewrite.

The ActorTemplate sets `OTEL_EXPORTER_OTLP_ENDPOINT` to the in-cluster
collector, since the SDK default (`localhost:4317`) has nothing listening inside
the actor sandbox netns.

Includes `install-demo-actor-telemetry.sh` wired into `install-ate.sh`, a
rendered-and-validated template, unit tests for the idle-flush watcher, and a
README documenting the reproduce/verify flow against Jaeger and Prometheus on
Kind.

## Scope

This covers the "reference demo for actor telemetry continuity" action item in
#503 only. The `docs/observability.md` update is tracked separately, and the
`PreSuspend`/`PostResume` alignment is #450.

Part of #503

## Testing

- `gofmt`, `go vet`, `go build`, and `go test ./demos/actor-telemetry/` all pass
  (4/4 unit tests on the idle-flush watcher).
- Template renders via the install script's `sed` substitution and parses as
  valid Kubernetes manifests.
- End-to-end suspend/resume verification against a live Kind cluster with
  ate-system is described in the README.

- [ ] Tests pass

### DIFF
--- a/demos/actor-telemetry/README.md
+++ b/demos/actor-telemetry/README.md
@@ -1,0 +1,141 @@
+# Actor Telemetry Continuity Demo
+
+This directory contains a reference demo for **actor telemetry continuity across
+suspend/resume**, the follow-up work discussed in
+[#503](https://github.com/agent-substrate/substrate/issues/503).
+
+It deploys a small Go HTTP server (`main.go`) that emits OpenTelemetry (OTel)
+metrics and trace spans over OTLP on every request, and shows how an actor
+author can avoid losing telemetry that is still buffered in the OTel SDK when an
+actor is suspended.
+
+## The problem it demonstrates
+
+OTel push exporters batch telemetry in memory and flush on a timer (the
+`PeriodicReader` for metrics, the `BatchSpanProcessor` for spans). When the push
+interval (e.g. 60s) is longer than the active-execution window before idle
+suspension (e.g. 30s), telemetry generated during the active phase is still
+sitting in those in-memory queues when the actor is suspended.
+
+With `snapshotsConfig.onPause: Data` (the mode this demo uses), Substrate
+excludes process memory from the snapshot and kills the container process on
+pause, so that buffered telemetry is **permanently lost** (this is Issue 1,
+HIGH, in #503).
+
+## The mitigation it shows
+
+Until the `PreSuspend` lifecycle hook ([#450](https://github.com/agent-substrate/substrate/issues/450))
+lands, an actor author can flush the OTel providers themselves before going
+idle. The workload runs an idle watcher: after `--idle-flush` of request
+inactivity it calls `ForceFlush` on both the `MeterProvider` and
+`TracerProvider`, exporting everything buffered before a suspend can kill the
+process. When `#450` lands, that same flush moves into the `PreSuspend` hook.
+
+Flags:
+
+| Flag | Default | Purpose |
+|------|---------|---------|
+| `--push-interval` | `60s` | OTLP export interval. Intentionally long, to reproduce #503. |
+| `--idle-flush` | `25s` | Idle duration after which the workload force-flushes. Set to `0` to disable the flush and observe the loss. |
+
+## Prerequisites
+
+- A k8s cluster with Agent Substrate installed (`./hack/install-ate.sh --deploy-ate-system`).
+- `ko` installed for building images.
+- A GCS bucket for storing snapshots (configured via `BUCKET_NAME` env var).
+
+## How to Run on Agent Substrate
+
+### 1. Build and Deploy
+
+> [!NOTE]
+> Do not manually edit `demos/actor-telemetry/actor-telemetry.yaml.tmpl`. The
+> installation script automatically injects your `${BUCKET_NAME}` environment
+> variable during deployment.
+
+```bash
+./hack/install-ate.sh --deploy-demo-actor-telemetry
+```
+
+This command will:
+- Build the demo server image using `ko`.
+- Create the `ate-demo-actor-telemetry` namespace.
+- Create the `WorkerPool` and `ActorTemplate`.
+- Wait until the template is ready.
+
+### 2. Create an Actor
+
+```bash
+# Install the CLI as a kubectl plugin if not already installed.
+go install ./cmd/kubectl-ate
+
+# Create the atespace (required before creating actors).
+kubectl ate create atespace demo
+
+# Create the actor from the actor-telemetry template.
+kubectl ate create actor my-telemetry-1 -a demo --template ate-demo-actor-telemetry/actor-telemetry
+```
+
+### 3. Port-Forward Services
+
+```bash
+# Router, to send requests to the actor.
+kubectl port-forward -n ate-system svc/atenet-router 8000:80
+
+# Jaeger + Prometheus, to see the telemetry arrive (Kind cluster).
+kubectl port-forward -n otel-system svc/jaeger 16686:16686
+kubectl port-forward -n otel-system svc/prometheus 9090:9090
+```
+
+## How to Use
+
+1. Send a few requests through the router:
+   ```bash
+   curl -X POST -H "Host: my-telemetry-1.demo.actors.resources.substrate.ate.dev" http://localhost:8000
+   ```
+   Each response reports the actor it was handled for. The metric
+   `demo.actor.requests` and a `handle-request` span are now buffered.
+
+2. **Stop sending requests.** After `--idle-flush` (25s) elapses, the workload
+   logs `pre-suspend flush complete`. Confirm the buffered telemetry arrived:
+   - Prometheus ([http://localhost:9090](http://localhost:9090)): query
+     `demo_actor_requests_total`.
+   - Jaeger ([http://localhost:16686](http://localhost:16686)): select service
+     `ate-demo-actor-telemetry` and find the `handle-request` traces.
+
+3. Suspend and resume the actor, then send more requests and repeat the checks.
+   In Jaeger, `handle-request` spans continue to carry the actor's own
+   `ate.actor.name` (read fresh from `/run/ate/actor-id`), correlated across the
+   worker pods the actor ran on. The `demo_actor_requests_total` metric is
+   labeled only by the bounded `ate.actor.template.name`, not per actor, to keep
+   metric cardinality bounded at scale (actor identity lives on spans, not
+   series).
+   ```bash
+   kubectl ate suspend actor my-telemetry-1 -a demo
+   ```
+
+4. **Observe the loss (optional).** Redeploy with `--idle-flush=0` (edit the
+   template's container `args`), send requests, then suspend before the 60s push
+   timer fires. With `onPause: Data` the buffered telemetry never reaches the
+   backend, reproducing #503.
+
+5. Clean up:
+   ```bash
+   kubectl ate delete actor my-telemetry-1 -a demo
+   kubectl ate delete atespace demo
+   ```
+
+## How to Uninstall
+
+```bash
+./hack/install-ate.sh --delete-demo-actor-telemetry
+```
+
+## Related
+
+- [`docs/observability.md`](../../docs/observability.md) - how Substrate
+  surfaces logs, metrics, and traces across suspend/resume.
+- [#503](https://github.com/agent-substrate/substrate/issues/503) - the
+  telemetry loss and delivery-delay issue this demo addresses.
+- [#450](https://github.com/agent-substrate/substrate/issues/450) - actor
+  lifecycle hooks (`PostResume` / `PreSuspend`) that will host this flush.
--- a/demos/actor-telemetry/actor-telemetry.yaml.tmpl
+++ b/demos/actor-telemetry/actor-telemetry.yaml.tmpl
@@ -1,0 +1,84 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Reference demo for actor telemetry continuity across suspend/resume (#503).
+#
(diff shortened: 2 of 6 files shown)


demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,309 @@
...
+	)
+	if errors.Is(err, resource.ErrPartialResource) || errors.Is(err, resource.ErrSchemaURLConflict) {
+		slog.WarnContext(ctx, "partial telemetry resource", slog.Any("err", err))
+	} else if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+func actorName() string {
```
git286, line 189: ATE_ACTOR_NAME isn't set anywhere in the repo, and os.Hostname() returns "runsc" for every actor — it's hardcoded in the OCI spec (cmd/atelet/oci.go:271) and applied  via the UTS namespace. So this always returns runsc.

The documented way is /run/ate/actor-id (docs/api-guide.md:108), with a working example in internal/e2e/fixtures/probe/main.go:32. api-guide.md:111 also warns specifically against the env-var approach, since env is frozen in the golden snapshot.

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,309 @@
...
+	otel.SetTextMapPropagator(propagation.TraceContext{})
+	return tp, nil
+}
+
+func main() {
+	pflag.Parse()
+	ctx := context.Background()
+	slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stdout, nil)))
+
+	res, err := newResource(ctx)
```
git286, line 241: This resource is built in main(), so it's captured by the golden snapshot and identical for every actor of the template: even reading /run/ate/actor-id here would give you the golden actor's name.

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,309 @@
...
+	if err != nil {
+		slog.ErrorContext(ctx, "init tracing", slog.Any("err", err))
+		os.Exit(1)
+	}
+
+	meter := mp.Meter(serviceName)
+	requests, err := meter.Int64Counter(
+		"demo.actor.requests",
+		metric.WithDescription("Requests handled by the actor-telemetry demo, by actor."),
+	)
```
git286, line 262: I just filed #761 : the counter might have the merge and conflate problem when actors run on different worker pods during its lifecycle.

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,309 @@
...
+	// hot path non-blocking.
+	activity := make(chan struct{}, 1)
+	go idleFlusher(ctx, *idleFlush, activity, selfSuspend)
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		ctx, span := tracer.Start(r.Context(), "handle-request")
+		defer span.End()
+
+		requests.Add(ctx, 1, metric.WithAttributes(
```
git286, line 297: ate.dev/actor_name on the metric datapoint is unbounded cardinality. The north star of this project is to have billions of actors.

On the span below it's fine, since a span is an event rather than a series.

For a reference demo I'd keep actor identity on the span and use template name on the metric.

demos/actor-telemetry/main.go
sanskar-singh-2403: Good catch, confirmed. `os.Hostname()` returns `runsc` for every actor (hardcoded at `cmd/atelet/oci.go:268`) and `ATE_ACTOR_NAME` is set nowhere, so this always returned `runsc`. Switched to reading `/run/ate/actor-id` fresh per the guidance in `docs/api-guide.md`, mirroring `internal/e2e/fixtures/probe/main.go`.

demos/actor-telemetry/main.go
sanskar-singh-2403: Right, building it in `main()` freezes it into the golden snapshot. Moved actor identity out of the resource entirely and now read it fresh in the request handler, so it reflects the resumed actor rather than the golden one. Updated the test to cover the fresh-read and the missing-file fallback.

demos/actor-telemetry/main.go
sanskar-singh-2403: got it that is the merge/conflate concern across worker pods and is orthogonal to this demo's fix; happy to track the counter-aggregation semantics there.

demos/actor-telemetry/main.go
sanskar-singh-2403: Agreed, per-actor labels on the metric are unbounded cardinality at billions of actors. Changed the counter to label with the bounded `ate.dev/template` and kept `ate.dev/actor_name` on the span only, since a span is an event rather than a series. README updated to match.

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+	for _, addr := range addrs {
+		if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+			if ipnet.IP.To4() != nil {
+				return ipnet.IP.String()
+			}
+		}
+	}
+	return "y.y.y.y"
+}
```
krisztianfekete: This is the `os.Hostname()` problem again. I'd either drop it, or keep it and relabel it to something like fake-static IP.

demos/actor-telemetry/README.md
```diff
@@ -0,0 +1,140 @@
...
+kubectl port-forward -n otel-system svc/prometheus 9090:9090
+```
+
+## How to Use
+
+1. Send a few requests through the router:
+   ```bash
+   curl -X POST -H "Host: my-telemetry-1.demo.actors.resources.substrate.ate.dev" http://localhost:8000
+   ```
+   Each response reports the worker pod that served it. The metric
```
krisztianfekete: This is not correct, see https://github.com/agent-substrate/substrate/pull/750/changes#r3728711006

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,337 @@
...
+	exp, err := otlptracegrpc.New(ctx, otlptracegrpc.WithInsecure())
+	if err != nil {
+		return nil, fmt.Errorf("create OTLP trace exporter: %w", err)
+	}
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithResource(res),
+		sdktrace.WithBatcher(exp, sdktrace.WithBatchTimeout(*pushInterval)),
+	)
+	otel.SetTracerProvider(tp)
+	otel.SetTextMapPropagator(propagation.TraceContext{})
```
krisztianfekete, line 232: You set the propagator but never extract, so tracer always starts a root span and the actor's traces never join the router's. See https://github.com/agent-substrate/substrate/issues/427

You can also use `otelhttp.NewHandler` to do avoid extraction and setting attrs manually.

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,337 @@
...
+		ctx, span := tracer.Start(r.Context(), "handle-request")
+		defer span.End()
+
+		// Read the actor's identity fresh for this request (see actorName).
+		actor := actorName()
+
+		// Metric: label with the bounded template name only. Per-actor labels
+		// are unbounded cardinality at this project's scale (see templateName).
+		requests.Add(ctx, 1, metric.WithAttributes(
+			attribute.String("ate.dev/template", templateName),
```
krisztianfekete: `ate.dev/` is the k8s label and what we log to sdtout. Span and metric attributes use the `ate.*` namespace. So these should be `ate.actor.template.name` here and `ate.actor.name` below.

Please also update this in README.

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,337 @@
...
+		return "unknown"
+	}
+	return strings.TrimSpace(string(b))
+}
+
+// initMeterProvider builds a MeterProvider whose PeriodicReader exports over
+// OTLP every push-interval. The long interval is the whole point: it is what
+// leaves telemetry buffered when a suspend arrives before the timer fires.
+func initMeterProvider(ctx context.Context, res *resource.Resource) (*sdkmetric.MeterProvider, error) {
+	exp, err := otlpmetricgrpc.New(ctx, otlpmetricgrpc.WithInsecure())
```
krisztianfekete: Please don't hardcode `WithInsecure`, see https://github.com/agent-substrate/substrate/issues/741

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,337 @@
...
+	tp, err := initTracerProvider(ctx, res)
+	if err != nil {
+		slog.ErrorContext(ctx, "init tracing", slog.Any("err", err))
+		os.Exit(1)
+	}
+
+	meter := mp.Meter(serviceName)
+	requests, err := meter.Int64Counter(
+		"demo.actor.requests",
+		metric.WithDescription("Requests handled by the actor-telemetry demo, by actor."),
```
krisztianfekete:
```suggestion
		metric.WithDescription("Requests handled by the actor-telemetry demo, by template."),
```

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,337 @@
...
+		os.Exit(1)
+	}
+	tp, err := initTracerProvider(ctx, res)
+	if err != nil {
+		slog.ErrorContext(ctx, "init tracing", slog.Any("err", err))
+		os.Exit(1)
+	}
+
+	meter := mp.Meter(serviceName)
+	requests, err := meter.Int64Counter(
```
krisztianfekete, line 258: Nit: please add `metric.WithUnit("{request}")`

demos/actor-telemetry/actor-telemetry.yaml.tmpl
```diff
@@ -0,0 +1,76 @@
...
+  containers:
+  - name: actor-telemetry
+    image: ko://github.com/agent-substrate/substrate/demos/actor-telemetry
+    env:
+    # The actor must be told where the OTLP collector is; the SDK default is
+    # localhost:4317, which inside the actor's gVisor netns has nothing
+    # listening, so telemetry would be flushed into the void. This mirrors the
+    # endpoint the ate-system components use on Kind (see
+    # manifests/ate-install/kind/kustomization.yaml).
+    - name: OTEL_EXPORTER_OTLP_ENDPOINT
```
krisztianfekete, line 61: @git286, I think it's a good opportunity to set delta temporarily here as well. See my reasoning in https://github.com/agent-substrate/substrate/issues/761#issuecomment-5204423513

demos/actor-telemetry/main.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// newResource builds the OTel resource shared by the meter and tracer
+// providers, mirroring internal/serverboot. WithFromEnv is last so OTEL_*
+// env vars (e.g. resource attributes injected by the actor runtime) win.
+func newResource(ctx context.Context) (*resource.Resource, error) {
+	res, err := resource.New(ctx,
+		resource.WithTelemetrySDK(),
+		resource.WithSchemaURL(semconv.SchemaURL),
+		resource.WithAttributes(
```
krisztianfekete, line 150: Please add `service.instance.id`

demos/actor-telemetry/README.md
sanskar-singh-2403: Fixed. Reworded to "reports the actor it was handled for" now that the IP is gone.

demos/actor-telemetry/main.go
sanskar-singh-2403: Removed `getCurrentIP` entirely. Per #761 every actor egresses behind one constant interior IP inside the gVisor netns, so it could never identify the worker pod. The response now reports the actor name (read fresh from the identity file) instead.

demos/actor-telemetry/main.go
sanskar-singh-2403: Good catch. Wrapped the mux in `otelhttp.NewHandler(mux, "/")` so the incoming W3C trace context is extracted and the per-request span joins the router's trace instead of starting a detached root. Follows the same pattern as the atenet router and glutton.

demos/actor-telemetry/main.go
sanskar-singh-2403: Done. Switched to the dotted `ate.*` namespace: `ate.actor.template.name` on the metric and `ate.actor.name` on the span, matching `internal/ateattr`. Updated the README to match.

demos/actor-telemetry/main.go
sanskar-singh-2403: Dropped the hardcoded `WithInsecure()` from both exporters. The demo endpoint is plaintext `http://`, which the SDK already infers from the scheme, so nothing regresses while the SDK's env-driven TLS config is now respected.

demos/actor-telemetry/main.go
sanskar-singh-2403: done

demos/actor-telemetry/main.go
sanskar-singh-2403: done

demos/actor-telemetry/main.go
sanskar-singh-2403: Added `semconv.ServiceInstanceID`, generated once and shared by the meter and tracer resources, mirroring internal/serverboot.

demos/actor-telemetry/actor-telemetry.yaml.tmpl
sanskar-singh-2403: Set `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=delta` in the template. Agreed it fits this suspend/resume workload: with cumulative, a restore from the golden snapshot carries the golden actor's accumulated value and reset detection cannot see the restart, so delta avoids the inflate/merge. Tracking the broader identity fix in #761.

PR #751, opened 2026-08-05
ci: harden workflows for zizmor (pin actions, fix template injection)

sanskar-singh-2403:
## What this PR does / why we need it

The zizmor gate has two mandatory checks failing on the workflow files:
`unpinned-uses` and `template-injection`. This PR resolves both.

1. Pin every GitHub Actions `uses:` reference to a full commit SHA, keeping
   the semver in a trailing comment for readability and Dependabot:

   - actions/checkout@v5          -> fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 (v5.1.0)
   - actions/setup-go@v5          -> 40f1582b2485089dde7abd97c1529aa768e1baff (v5.6.0)
   - actions/cache@v4             -> 0057852bfaa89a56745cba8c7296529d2fc39830 (v4.3.0)
   - golang/govulncheck-action@v1 -> 032d45514ae346b1db93c04b0c90b841c370344f (v1.1.0)

2. Fix template injection in pr-workflow.yaml: `${{ matrix.auth-mode }}` and
   `${{ needs.e2e-test-matrix.result }}` were interpolated directly into `run:`
   shell blocks. Move each into an `env:` var and reference it as a quoted
   shell variable so the expansion cannot inject code.

No functional behavior change.

## Testing

- Ran `zizmor --persona=pedantic` on both workflow files: zero findings for any
  mandatory check (unpinned-uses, template-injection, etc.).
- Confirmed both workflows remain valid YAML.

### DIFF
--- a/.github/workflows/govulncheck.yaml
+++ b/.github/workflows/govulncheck.yaml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - id: govulncheck
-        uses: golang/govulncheck-action@v1
+        uses: golang/govulncheck-action@032d45514ae346b1db93c04b0c90b841c370344f # v1.1.0
         with:
           go-version-file: go.mod
           go-package: ./...
--- a/.github/workflows/pr-workflow.yaml
+++ b/.github/workflows/pr-workflow.yaml
@@ -89,11 +89,15 @@ jobs:
     - name: Create cluster
       run: hack/create-kind-cluster.sh
     - name: Install Agent Substrate
-      run: hack/install-ate-kind.sh --deploy-ate-system --ateapi-client-auth=${{ matrix.ateapi-client-auth }}
+      env:
+        CLIENT_AUTH: ${{ matrix.ateapi-client-auth }}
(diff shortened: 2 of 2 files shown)


BenTheElder:
isn't most of this https://github.com/agent-substrate/substrate/pull/657/changes?

sanskar-singh-2403:
> isn't most of this https://github.com/agent-substrate/substrate/pull/657/changes?

I'll rebase

sanskar-singh-2403:
@BenTheElder I've reset onto current main and trimmed this down to just the findings #657
didn't cover:

- govulncheck.yaml: pin golang/govulncheck-action@v1 to a SHA (the only
  remaining unpinned-uses).
- pr-workflow.yaml template-injection: move ${{ matrix.ateapi-client-auth }}
  (Install / Deploy steps) and ${{ needs.e2e-test-matrix.result }} (Check E2E
  matrix step) into env: vars referenced as quoted shell variables.

Verified with zizmor: zero unpinned-uses and zero template-injection findings.
Force-pushed the trimmed version. Thanks for the pointer to #657.

PR #754, opened 2026-08-05
feat(otel): onboard atecontroller to the OTLP path

krisztianfekete:
atecontroller had no OTel at all. Dev-mode zap logger, and its controller-runtime metrics were only available on a :8080 that we don't scrape.

After this PR, logs go through the shared slog handler (plus a `--log-level` flag to match the other binaries), and controller-runtime's Prometheus registry is bridged onto the OTLP reader so the reconcile/workqueue metrics actually reach the collector. Filtering these out is a pipeline responsibility. Also added otelgrpc to the ateapi client, which was untraced.

This unblocks #564 the workperpool metrics, cc @Angelawork, @JeffLuoo: there's a working `MeterProvider` to use for `ate.workerpool.desired_workers`/`ready_workers`.

Couple of things to mention for review:

-`InitMetricsPushOnly`, not `InitMetrics`, even though we do serve :8080. That port is controller-runtime's own private registry, not the global one `serverboot.metricsMux` serves, so a pull reader there would collect into something we never expose.
- Bridge is pinned to v0.68.0 to match otelgrpc. Wanted to go to v0.70.0, but that requires otel/sdk/metric 1.45.0 and pulls the whole SDK up with it (406 vendor files instead of 81). Happy to do that bump separately.
- zap/zapr fall out of go.mod since atecontroller was the last importer.
- I included an OTel collector image bump from the early 2024 (!) one to latest, which was breaking exposing native histograms

- [x] Tests pass


docs/observability.md
```diff
@@ -124,6 +124,14 @@ For `atenet.router.route.duration`:
 
 The `ate.*` control-plane metric labels are either fixed value sets (operation, outcome, state, class, kind) or scoped to the deployment catalog (template and pool names are operator-created, never derived from request payloads), and the label set varies per operation: resume carries the most dimensions, delete only the operation and error type. `ate.sandbox.class` is derived from the template (each template has exactly one class), so it adds no extra series next to the template labels; it exists so dashboards can aggregate by class without enumerating template names. High-cardinality actor identity (name/uid/atespace) stays off metrics entirely and lives on logs and traces instead.
 
+### Bridged controller-runtime metrics (atecontroller)
+
+atecontroller bridges controller-runtime's private Prometheus registry, which the manager serves on an unscraped `:8080`, onto its OTLP reader. So `controller_runtime_*`, `workqueue_*`, `rest_client_*`, `leader_election_*`, `go_*`, and `process_*` reach the collector too, keeping their Prometheus names because they are upstream instruments and renaming them would break existing controller-runtime dashboards.
```
JeffLuoo, line 130: A general question - The controller will expose many metrics. Do we really find it useful to have all of them? If not, we should find a way to filter out them and only keep metrics we need from controller.

krisztianfekete: I am always a bit torn when it comes to this question regarding these specific metrics. In general the OTel recommendations are about exposing everything, and filtering in the pipeline, which makes sense as you might not be in the position to change this at the application level. 

The reason I included (or not filtered out) them in this PR was that fairly recently I had to use them and were lucky that they were available. We could add a filter for them in the pipeline as a follow-up. Wdyt?

JeffLuoo: sg, let's do it in a follow-up to update the pipeline if metrics from controller becomes too noisy.

JeffLuoo:
Please rebase

JeffLuoo (approved):

PR #757, opened 2026-08-05
Configurable JWT authentication to ateapi

### LINKED ISSUE #733: Human JWT Auth: Human user of substrate can authenticate to ate-api-server with a JWT issued by an external IDP
We'll probably start with the Google IDP --- once this task is done, a human user should be able to authenticate to ate-api-server using the JWT printed by `gcloud auth print-identity-token`.

Follow up tasks will need to make JWT auth more configurable, following the example of [Kubernetes' pluggable JWT authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#using-authentication-configuration)

EItanya:
## Summary

  Fixes #733 by adding configurable JWT authentication to ateapi.

  - Supports multiple OIDC/JWT providers through a YAML configuration file.
  - Validates issuer, audience, signature, and token lifetime.
  - Supports custom CAs and authenticated OIDC discovery for Kubernetes
    ServiceAccount JWTs.

  - Restricts MintJWT to the configured actor-identity provider.
  - Replaces the existing --client-jwt-* flags.

  RBAC remains follow-up work.

  ## Testing

  - go test ./...
  - Kubernetes ServiceAccount JWT E2E
  - Real Google-issued JWT E2E
  - golangci-lint and repository verification checks


SHRUTI6991:
I overall really like the idea of your PR and it honestly supersedes mine: https://github.com/agent-substrate/substrate/pull/759 which is a much simpler version and focusses only on Human based authentication + K8 (only OIDC).

Some comments:

1. In terms of developer experience, it is much easier for developers to run `kubectl-ate --token $(gcloud auth print-identity-token)` instead of taking a config file. Can you refactor this?

2. Before routing to the correct provider, may be validate the JWT claims. 

3. Not related to the current PR. What's your thought on making a chain of authenticators? For standard OIDC JWTs, matching directly on the iss claim works great. If Substrate expands to support non-OIDC tokens (like static API keys or tokens without an iss claim), wrapping authenticators in an Authenticator chain interface might be a nice extensibility pattern down the road. 

cc: @ahmedtd

EItanya:
> I overall really like the idea of your PR and it honestly supersedes mine: #759 which is a much simpler version and focusses only on Human based authentication + K8 (only OIDC).

Okie dokie, I'll post this a ready for review.

> In terms of developer experience, it is much easier for developers to run kubectl-ate --token $(gcloud auth print-identity-token) instead of taking a config file. Can you refactor this?

This can leak tokens into the shell so isn't a great idea, what about stdin or something like that?

> Before routing to the correct provider, may be validate the JWT claims.

👍 

> Not related  to the current PR. What's your thought on making a chain of authenticators? For standard OIDC JWTs, matching directly on the iss claim works great. If Substrate expands to support non-OIDC tokens (like static API keys or tokens without an iss claim), wrapping authenticators in an Authenticator chain interface might be a nice extensibility pattern down the road.

Absolutely. Do you think we should do that in a follow-up?

SHRUTI6991:
@EItanya yes, STDIN sounds good to me.

> Absolutely. Do you think we should do that in a follow-up?

I would consider this a P2 since we aren't currently working towards having any other authentication mechanisms except the JWT for now. Let me know @ahmedtd  if you think otherwise.

cmd/ateapi/internal/oidcjwt/oidcjwt.go
thompsonmax, line 110: [nit] Is it worthwhile to logically separate out the k8s-specific claims by placing them in a separate, inner struct, or just indicating their group with a comment?

cmd/ateapi/internal/oidcjwt/oidcjwt.go
```diff
@@ -260,6 +282,53 @@ func Verify(ctx context.Context, httpClient *http.Client, jwt string, expectedIs
...
+	if key != nil && now.Sub(v.lastRefresh) < jwksRefreshInterval {
+		return key.PublicKey, nil
+	}
+	if key == nil && len(v.keys) > 0 && !v.lastUnknownKeyRefresh.IsZero() && now.Sub(v.lastUnknownKeyRefresh) < unknownKeyRefreshInterval {
+		return nil, fmt.Errorf("unknown key ID %q", keyID)
+	}
+	if key == nil && len(v.keys) > 0 {
+		v.lastUnknownKeyRefresh = now
+	}
+	keys, err := discoverKeysForIssuer(ctx, v.httpClient, v.issuer)
```
thompsonmax, line 311: Once jwksRefreshInterval elapses, a token whose key may already be in the cache only verifies if this refresh succeeds. So a transient issuer outage 5+ minutes after the last refresh fails all JWT auth, even for keys we already hold.

I believe k8s' OIDC authenticator serves stale keys when refresh fails, should we do the same here?

cmd/ateapi/internal/oidcjwt/oidcjwt.go
```diff
@@ -260,6 +282,53 @@ func Verify(ctx context.Context, httpClient *http.Client, jwt string, expectedIs
...
+	v.mu.RLock()
+	key := findKey(v.keys, keyID)
+	hasKeys := len(v.keys) > 0
+	lastRefresh := v.lastRefresh
+	lastUnknownKeyRefresh := v.lastUnknownKeyRefresh
+	v.mu.RUnlock()
+	if key != nil && now.Sub(lastRefresh) < jwksRefreshInterval {
+		return key.PublicKey, nil
+	}
+	if key == nil && hasKeys && !lastUnknownKeyRefresh.IsZero() && now.Sub(lastUnknownKeyRefresh) < unknownKeyRefreshInterval {
```
thompsonmax: Since we check `hasKeys` here, I believe this check does not guard against the case where the initial set of keys has not yet been fetched (possibly delayed due to an issuer outage), so each token validation would trigger a request to fetch keys unthrottled until the first set of keys is successfully fetched.

Should we implement a similar throttling mechanism to prevent unthrottled fetches before the initial set of keys is fetched?

internal/ateapiauth/server.go
```diff
@@ -143,22 +155,54 @@ func mtlsPeerIdentity(ctx context.Context) (string, bool) {
...
-	id, err := a.verifyBearerToken(ctx, bearer)
+	issuer, err := unverifiedIssuer(bearer)
 	if err != nil {
-		return nil, status.Errorf(codes.Unauthenticated, "invalid bearer token: %v", err)
+		return nil, status.Error(codes.Unauthenticated, "invalid bearer token")
+	}
+	for _, provider := range a.providers {
+		if provider.Issuer != issuer {
+			continue
+		}
```
thompsonmax, line 178: [nit] Is it worth providing a log if no issuers match to make it easier to debug that case?

cmd/ateapi/internal/oidcjwt/oidcjwt.go
```diff
@@ -125,7 +148,12 @@ var (
 //
 // httpClient is used for OIDC discovery and JWKS fetches; nil uses a default
 // client with a whole-request timeout.
-func Verify(ctx context.Context, httpClient *http.Client, jwt string, expectedIssuer, expectedAudience string, now time.Time) (*KubernetesClaims, error) {
+func Verify(ctx context.Context, httpClient *http.Client, jwt string, expectedIssuer, expectedAudience string, now time.Time) (*Claims, error) {
+	return NewVerifier(expectedIssuer, []string{expectedAudience}, httpClient).Verify(ctx, jwt, now)
+}
```
thompsonmax: IIUC this is dead code now, since we call `verifier.Verify` in `cmd/ateapi/main.go`. Should we remove?

internal/ateapiauth/server.go
```diff
@@ -35,19 +37,29 @@ import (
 )
 
 func ValidateServerConfig(cfg ServerConfig) error {
-	if cfg.VerifyBearerToken == nil {
-		return fmt.Errorf("a bearer token verifier is required")
+	if len(cfg.JWTProviders) == 0 {
+		return fmt.Errorf("at least one JWT provider is required")
+	}
+	for i, provider := range cfg.JWTProviders {
+		if provider.Name == "" || provider.Issuer == "" || provider.Verify == nil {
```
thompsonmax, line 45: Should we also validate uniqueness of Issuer in the list of JWTProviders?

SHRUTI6991:
Hi @EItanya, are you working on this PR?

EItanya:
> Hi Eitan Yarmush (@EItanya), are you working on this PR?

Yup! I was focused on #708, but that’s in so I’ll go full steam on this now.

The GitHub CI situation yesterday really messed me up

cmd/ateapi/internal/oidcjwt/oidcjwt.go
EItanya: Removed the unused package-level helper and updated tests to use the cached verifier.

internal/ateapiauth/server.go
EItanya: Added a debug-level unmatched-issuer log to avoid unauthenticated warning spam.

cmd/ateapi/internal/oidcjwt/oidcjwt.go
EItanya: Failed initial fetches are now rate-limited as well.

EItanya:
Addressed in the latest push: --token-file=- now reads one token from stdin. The unverified issuer only selects a configured verifier and grants no authority; signature, issuer, audience, subject, and time validation happen immediately afterward, since cryptographic validation requires first selecting the issuer keys. Agreed that authenticator chaining should remain a P2 follow-up when we add a non-JWT mechanism.

cmd/ateapi/internal/oidcjwt/oidcjwt.go
thompsonmax, line 577: [nit] not related your change, but in general I believe RSA allows exponents larger than 64 bits (although it's uncommon in practice). I think it's worth fixing here or opening an issue to track.

internal/ateapiauth/server.go
```diff
@@ -143,22 +160,55 @@ func mtlsPeerIdentity(ctx context.Context) (string, bool) {
...
+	if err != nil {
+		return nil, status.Error(codes.Unauthenticated, "invalid bearer token")
+	}
+	for _, provider := range a.providers {
+		if provider.Issuer != issuer {
+			continue
+		}
+		id, err := provider.Verify(ctx, bearer)
+		if err != nil {
+			return nil, status.Errorf(codes.Unauthenticated, "invalid bearer token: %v", err)
```
thompsonmax: [nit] We should probably avoid leaking error details to unauthenticated clients

cmd/ateapi/internal/oidcjwt/oidcjwt.go
thompsonmax: [nit] IIUC we're caching the keys now, update this comment?

cmd/ateapi/internal/oidcjwt/http.go
```diff
@@ -0,0 +1,103 @@
...
+	"net/url"
+	"os"
+	"strings"
+	"time"
+)
+
+// NewHTTPClient returns a client for OIDC discovery and JWKS requests.
+func NewHTTPClient(issuer, certificateAuthorityFile, discoveryTokenFile string) (*http.Client, error) {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	if certificateAuthorityFile != "" {
```
thompsonmax, line 34: Should we require certificateAuthorityFile is always set whenever discoveryTokenFile is set? isKubernetesJWKSURL matches on path alone, so if a configured issuer's discovery document names a jwks_uri on a different host (or redirects there), we'll attach ateapi's KSA token to it as long as the path is /openid/v1/jwks. A compromised issuer could mean leakage of the ateapi's KSA JWT.

cmd/ateapi/internal/oidcjwt/oidcjwt.go
EItanya: Updated in e0b9239f, along with a quick audit of the other source comments changed by this PR.

cmd/ateapi/internal/oidcjwt/http.go
EItanya: Addressed in e0b9239f: configuring a discovery token now requires a certificate authority file, with a regression test covering the guard.

cmd/ateapi/internal/oidcjwt/oidcjwt.go
EItanya: Addressed in e0b9239f: oversized RSA exponents are now rejected instead of being silently truncated to Go's int representation.

internal/ateapiauth/server.go
EItanya: Addressed in e0b9239f: clients now receive a generic invalid bearer token error; verifier details are logged at debug level server-side.

thompsonmax:
/lgtm

Thanks for addressing all comments!

LiorLieberman:
Thanks @EItanya @SHRUTI6991 and @thompsonmax !

PR #758, opened 2026-08-05
 Make UpdateActorSnapshot follow substrate API guidelines

### LINKED ISSUE #732: Make Update<ResourceType> methods to follow API guidelines

laoj2:
Fixes #732  

* `UpdateActorSnapshot` now carries the resource itself + `update_mask`
* `scope` is now applied via the update mask
*  Moved `update_mask` to a separate file, so it can be reused by other update RPCs.
* Added `uid` and `version` as optional guards. 

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/update_actor.go
+++ b/cmd/ateapi/internal/controlapi/update_actor.go
@@ -18,23 +18,18 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"maps"
-	"slices"
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
 	"github.com/agent-substrate/substrate/internal/resources"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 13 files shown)


cmd/ateapi/internal/controlapi/update_actor.go
```diff
@@ -64,7 +59,7 @@ func (s *Service) UpdateActor(ctx context.Context, req *ateapipb.UpdateActorRequ
 		expectedVersion = version
 	}
 
-	applyActorUpdateMask(actor, in, req.GetUpdateMask())
+	applyUpdateMask(actor, in, req.GetUpdateMask(), actorMutableFields)
 
 	updated, err := s.persistence.UpdateActor(ctx, actor, expectedVersion)
```
juli4n, line 64: As discussed, this has the same TOCTOU issues we talked in #742 but that will get fixed by #763 so I don't think we should block here.

juli4n (approved):

PR #759, opened 2026-08-05
Implement configurable OIDC JWT token authentication for ate-api-server.

### LINKED ISSUE #733: Human JWT Auth: Human user of substrate can authenticate to ate-api-server with a JWT issued by an external IDP

SHRUTI6991:
# feat(ateapi): support multiple OIDC JWT issuers, CEL ClaimMappings, and tinyjwt verifier

Fixes #733.

## Summary
Expands `ate-api-server` authentication to support multiple, pluggable OIDC JWT providers evaluated sequentially through an authenticator chain (`oidcauth`). It supports both Kubernetes ServiceAccount JWTs and Human OIDC credentials (e.g., Google Workspace accounts).

Additionally, low-level JWT verification is extracted into `internal/tinyjwt`, username mapping uses CEL-style `ClaimMappings`, and `UserInfo.ExtraInfo` (`map[string][]string`) is aligned 1:1 with `k8s.io/apiserver/pkg/authentication/user.Info`.

## Key Changes
* **`internal/tinyjwt`**: Low-level RFC7519 JWT verification, JWKS key discovery & caching, RSA/ECDSA signature checks, sentinel `ErrIssuerMismatch` error, and security fuzzing (`FuzzVerify`).
* **`oidcauth` Chain & CEL `ClaimMappings`**: Multi-issuer `Authenticator` chain replacing `k8sjwt`. Maps identity via CEL expressions (`'claims.sub'` vs `'google:' + claims.email`).
* **Kubernetes Parity (`UserInfo.ExtraInfo`)**: Aligns `UserInfo` and `principal.PrincipalInfo` with `k8s.io/apiserver/pkg/authentication/user.Info` (`map[string][]string`).
* **`kubectl-ate` DX**: Adds persistent `--token` (`-t`) flag and `ATE_TOKEN` environment variable support.

## Testing Performed
* **Unit Tests & Fuzzing**: Passed `go test ./internal/tinyjwt/... ./cmd/ateapi/internal/oidcauth/... ./internal/ateapiauth/...`.
* **Repository Checks**: Passed `make verify` and `make all`.
* **Live GKE Verification**: Verified end-to-end against live `ate-api-server` using Google OIDC identity tokens and GKE projected ServiceAccount tokens (`kubectl create token`).


internal/tinyjwt/tinyjwt.go
```diff
@@ -0,0 +1,533 @@
...
+	Expiration time.Time
+	IssuedAt   time.Time
+
+	// Optional Standard Claims
+	Email     string
+	NotBefore time.Time
+	JTI       string
+
+	// Kubernetes contains structured bound token metadata when present (nil for non-Kubernetes tokens).
+	Kubernetes *KubernetesBoundClaims
```
ahmedtd, line 98: We need to start thinking about how to handle other issuers that have custom claims.  I believe Kubernetes handles this by supporting CEL for constructing the username and extra info from a dynamic map of all the claims that were in the JWT.

internal/tinyjwt/tinyjwt.go
```diff
@@ -0,0 +1,533 @@
+// Copyright 2026 Google LLC
```
ahmedtd, line 1: Did you change anything in this file?

As a nit, I'd prefer to keep the functionality for verifying the JWT in a separate package from the Substrate logic for constructing a user from the verified JWT.  It doesn't need to be called k8sjwt anymore, since we'll be using it for other stuff.  Maybe simplejwt or tinyjwt.

The reason is that the verifier logic is a self-contained task that is security-critical.  It will need a *bunch* of exhaustive tests and fuzz tests.

The substrate user construction logic is also security-critical, but it can be exhaustively tested in isolation --- start with an already-parsed claims object, and then test that you get the expected user name out of it.

cmd/ateapi/internal/oidcauth/oidcauth.go
```diff
@@ -0,0 +1,128 @@
...
+
+
+// Authenticator evaluates a Bearer token and returns an authenticated principal ID.
+//
+// ok == false indicates that the token was not recognized by this authenticator
+// (e.g., unrecognized issuer), so the chain should try the next authenticator.
+// ok == true indicates that the token was evaluated by this authenticator; if err != nil,
+// verification failed (e.g. expired token or invalid signature) and the chain stops.
+type Authenticator interface {
+	AuthenticateToken(ctx context.Context, token string) (id string, ok bool, err error)
```
ahmedtd: Kubernetes needs more info extracted than just the plain username, and it's a safe bet that we will too.

Can you define a structure like the following and have the authenticators return that?

```
type UserInfo struct {
  ID string
  ExtraInfo map[string]string
}
```

cmd/ateapi/internal/oidcauth/oidcauth.go
```diff
@@ -0,0 +1,128 @@
...
+	if a.cfg.IssuerURL == "" {
+		return "", false, nil
+	}
+
+	claims, err := Verify(ctx, a.httpClient, token, a.cfg.IssuerURL, a.cfg.Audiences, a.now())
+
+
+	if err != nil {
+		// If verification failed because of unexpected issuer, return ok=false so chain can continue.
+		if isIssuerMismatch(err) {
```
ahmedtd: Can you make this a unique error returned from the verifier, rather than doing string matching?

cmd/ateapi/internal/oidcauth/oidcauth.go
```diff
@@ -0,0 +1,128 @@
...
+// OIDCAuthenticatorConfig defines the configuration for an OIDC authenticator,
+// mirroring the structured AuthenticationConfiguration API in Kubernetes.
+type OIDCAuthenticatorConfig struct {
+	// IssuerURL is the OIDC issuer URL (e.g., "https://accounts.google.com").
+	IssuerURL string
+	// Audiences is the list of acceptable audience values in the token.
+	Audiences []string
+	// UsernameClaim specifies which JWT claim to map to the principal ID ("email" or "sub").
+	// If empty or if the claim is absent in the token, it falls back to the "sub" claim.
+	UsernameClaim string
```
ahmedtd: Do you think it might be worthwhile to skip the "select a fixed claim as the username" logic, and skip to only supporting CEL claim mappings?  https://kubernetes.io/docs/reference/access-authn-authz/authentication/#jwt-egress-selector-type:~:text=request%20will%20fail.-,expression%3A%20%27claims.username%20%2B%20%22%3Aexternal%2Duser%22%27,-%23%20groups%20represents%20an

We don't even need to hook up CEL evaluation yet, we could just require that your CEL expression has the literal content of `claims.sub` for now.

internal/ateclient/builder.go
```diff
@@ -229,10 +227,25 @@ func serverTLSConfig(ctx context.Context, clientset kubernetes.Interface) (*tls.
...
 
+// Token allows setting an explicit Bearer token for authentication (e.g. via --token flag or ATE_TOKEN env var).
+var Token string
+
 // bearerTokenDialOption attaches a ServiceAccount token for the ate-client SA
 // as per-RPC credentials.
 func bearerTokenDialOption(ctx context.Context, clientset *kubernetes.Clientset) (grpc.DialOption, error) {
+	tok := Token
+	if tok == "" {
+		tok = os.Getenv("ATE_TOKEN")
```
ahmedtd, line 238: Why two environment variables?

cmd/ateapi/main.go
```diff
@@ -188,16 +187,50 @@ func main() {
 		serverboot.Fatal(ctx, "Failed to start listener", err)
 	}
 
+	// TODO: Support structured authentication configuration from a file (--authentication-config),
```
ahmedtd, line 190: Nit: Make it a TODO to your username or TODO(authn), so we can find open TODOs related to this work.

cmd/ateapi/internal/sessionidentity/sessionidentity.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	authorization := reqMetadata["authorization"]
+	if len(authorization) != 1 {
+		return nil, status.Errorf(codes.Unauthenticated, "Need authorization header")
+	}
+
+	clientJWT := strings.TrimPrefix(authorization[0], "Bearer ")
+
+	clientClaims, err := oidcauth.Verify(ctx, s.httpClient, clientJWT, s.k8sJWTIssuer, []string{s.k8sJWTAudience}, time.Now())
```
ahmedtd: This can be a follow up, but ultimately the session identity endpoints should be using the same AuthChain stuff that you install into ate-api-server.  We can then check the username and extra info to make sure that it's the right atelet (or whatever) calling us.

internal/tinyjwt/tinyjwt.go
SHRUTI6991: > Did you change anything in this file?

No, it's mostly some TODO comments. 

> As a nit, I'd prefer to keep the functionality for verifying the JWT in a separate package from the Substrate logic for constructing a user from the verified JWT.

I refactored this into a new sharable library called `tinyjwt`. This library contains the low level JWT verification logic.

cmd/ateapi/internal/oidcauth/oidcauth.go
SHRUTI6991: Instead of `map[string]string` for `ExtraInfo`, I changed it to `map[string][]string` to conform to kubernetes API: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authentication-strategies.

There was an example in the doc which showed returning list:

```
"extra": {
        "example.com/tenant": ["72f988bf-86f1-41af-91ab-2d7cd011db4a"]
    }
```

cmd/ateapi/internal/oidcauth/oidcauth.go
SHRUTI6991: Yes I started with `ClaimMappings` similar to K8 auth. It would be good for extensibility for other types of claims as well once we integrate the CEL library.

cmd/ateapi/internal/sessionidentity/sessionidentity.go
SHRUTI6991: I added a TODO for this.

internal/ateclient/builder.go
SHRUTI6991: Not needed. I kept one.

cmd/ateapi/internal/oidcauth/oidcauth.go
SHRUTI6991: done.

SHRUTI6991:
Favoring https://github.com/agent-substrate/substrate/pull/757 over this PR.

PR #764, opened 2026-08-05
Handle graceful termination in gvisor ateom

sfunkenhauser:
Part of https://github.com/agent-substrate/substrate/issues/23

Trap SIGTERM in the ateom, and forward SIGTERM to all application containers.
Once all application containers exit, ateom will exit itself.


demos/counter/counter.go
```diff
@@ -115,6 +128,24 @@ func main() {
...
+		if durationStr == "" {
+			http.Error(w, "missing duration parameter", http.StatusBadRequest)
+			return
+		}
+		d, err := strconv.Atoi(durationStr)
+		if err != nil || d < 0 {
+			http.Error(w, "invalid duration parameter", http.StatusBadRequest)
+			return
+		}
+		sigtermSleepDurationSecs = d
```
BenTheElder: 🤖 **should-fix** 🟡 – Data race on `sigtermSleepDurationSecs`. This handler goroutine writes the plain `int` while the SIGTERM goroutine reads it at `demos/counter/counter.go:77-78`, with no synchronization. The rest of this file already guards its shared state: `ready` is an `atomic.Bool` and `requestCount` goes through `atomic.AddUint64`. Make this an `atomic.Int64` and use `Store`/`Load`.

cmd/ateom-gvisor/main.go
```diff
@@ -301,13 +336,134 @@ func NewService(interiorNetNS netns.NsHandle, actorLogger *actorlog.ActorLogger,
...
+
+	err := waitContainerStop(sigTermCtx, done)
+	if err == nil {
+		return nil
+	}
+
+	// If the wait completed because the container actually exited (but with an error),
+	// we shouldn't send SIGKILL.
+	if !errors.Is(err, context.DeadlineExceeded) && !errors.Is(err, context.Canceled) {
+		slog.InfoContext(ctx, "Container exited with error status", slog.String("container", name), slog.Any("err", err))
```
BenTheElder: 🤖 **should-fix** 🟡 – A failed `runsc wait` is indistinguishable from a clean exit here. Any non-context error from `cmdWait` reaches this branch and returns `nil`, so `gracefulShutdown` logs "All application containers stopped" and ateom exits without ever sending SIGKILL.

That is correct for a container that exited with a non-zero status. It is wrong when the wait itself failed and the container is still running: a runsc error, a container name that no longer exists, or the spurious `ECHILD` that `cmdWait` maps to success at `cmd/ateom-gvisor/runsc.go:388`. In those cases the actor loses the remainder of its grace period to the kubelet's SIGKILL, which is the outcome this PR exists to prevent.

Consider having `cmdWait` return a typed error that separates "container exited" from "wait failed", and escalating to SIGKILL on the latter.

cmd/ateom-gvisor/main.go
```diff
@@ -301,13 +336,134 @@ func NewService(interiorNetNS netns.NsHandle, actorLogger *actorlog.ActorLogger,
...
+}
+
+// gracefulShutdown propagates SIGTERM into the sandbox and waits for the application's
+// containers to exit.
+func (s *AteomService) gracefulShutdown(ctx context.Context) {
+	s.lock.Lock()
+	s.shuttingDown = true
+	session := s.activeSession
+	// Release the lock so that AteomService and respond to new RPCs.
+	s.lock.Unlock()
```
BenTheElder, line 444: 🤖 **should-fix** 🟡 – Dropping the lock lets `CheckpointWorkload` run during drain, which is the stated intent, but the two paths are not coordinated. By the time a suspend arrives, `killContainer` has already sent SIGTERM to every application container, and it will send SIGKILL once `workloadGracePeriod` expires — including in the middle of an in-flight `runsc checkpoint`.

`CheckpointWorkload` clears `activeSession` at `cmd/ateom-gvisor/main.go:674`, but `gracefulShutdown` captured the session pointer above, so it goes on to signal containers that were just checkpointed. `TestGracefulWorkerTerminationSuspend` only exercises a suspend that completes well inside the grace period, so neither case is covered.

Consider having the checkpoint path cancel the kill timer, or re-checking `activeSession` in `killContainer` before escalating.

Separately, the comment on the line above reads "so that AteomService and respond to new RPCs" — presumably "can respond".

cmd/ateom-gvisor/main.go
```diff
@@ -77,6 +79,8 @@ var (
 // ingress to.
 const actorHTTPUpstream = "http://" + ateomnet.ActorVethIP + ":80"
 
+const workloadGracePeriod = 1 * time.Minute
```
BenTheElder, line 84: 🤖 **question** 🟢 – Why one minute? Worker pods are created with `terminationGracePeriodSeconds: 3600` (`cmd/atecontroller/internal/controllers/workerpool_apply.go:35`), so this gives the actor 1/60th of the budget the pod actually has and SIGKILLs it with 59 minutes left. A deliberately conservative cap is reasonable, but a comment tying this constant to the pod-level grace period would stop the two from drifting apart.

cmd/ateom-gvisor/runsc.go
```diff
@@ -323,3 +325,70 @@ func (r *runsc) cmdState(ctx context.Context, containerName string) error {
 	}
 	return nil
 }
+
+// killArgs builds the argv for `runsc kill <container> <signal>`. Factored out
+// so the argument construction can be unit-tested without executing runsc.
+func (r *runsc) killArgs(containerName, signal string) []string {
```
BenTheElder, line 329: 🤖 **nit** 🟢 – Both `killArgs` and `waitArgs` are documented as factored out "so the argument construction can be unit-tested without executing runsc", but no test was added. `runsc_test.go` already has `TestNvproxyGlobalArgs` to follow.

internal/e2e/suites/demo/termination_test.go
```diff
@@ -0,0 +1,339 @@
...
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestGracefulWorkerTermination exercises the propagated-SIGTERM eviction flow
+// end to end: an actor is scheduled onto a worker pod, that pod is deleted
+// (simulating a Kubernetes eviction), and the control plane is expected to mark
+// the worker DRAINING, then remove it and detach the actor once the pod is gone.
+//
+// The demo counter actor installs a SIGTERM handler that sleeps for 5 seconds
```
BenTheElder: 🤖 **nit** 🟢 – The comment says the handler sleeps for 5 seconds, but the test sets it to 15 at line 81.

BenTheElder (commented):
review skill pass

[needs rebase as well, sorry, these files are high churn]

cmd/ateom-gvisor/main.go
sfunkenhauser: Re-setting the grace period during a checkpoint defeats the purpose. Even if we are checkpointing the container shouldn't be allowed to run for longer than the grace period.

cmd/ateom-gvisor/main.go
sfunkenhauser: This isn't true. A clean exit would not return an error and be handled above. Any error returned from `runsc wait` is a failure to wait on the process, not a non-zero container status code.

That being said we probably should be returning an error here instead of nil, not that it changes the current behavior, but it's more appropriate if we want to add retry logic later.

demos/counter/counter.go
sfunkenhauser: Done

cmd/ateom-gvisor/main.go
sfunkenhauser: Added a comment with more detail.

cmd/ateom-gvisor/runsc.go
sfunkenhauser: Added some basic tests.

internal/e2e/suites/demo/termination_test.go
sfunkenhauser: Updated the comment.

BenTheElder (approved):

PR #769, opened 2026-08-06
atelet: preserve holes when copying a local checkpoint

BenTheElder:
copyFile used a plain `io.Copy`, which reads a hole as zeroes and writes it back as data. The biggest thing it copies is a guest memory image, which is mostly unallocated: staging a local checkpoint for a restore inflated a 164MiB snapshot into its full 2GiB logical size. (Note: That itself is a bug, see https://github.com/agent-substrate/substrate/pull/679)

That costs disk and I/O on every local checkpoint restore, and it compounds, because cloud-hypervisor then loads the whole image and the next checkpoint it writes is dense in turn. Five pause cycles on one actor left 8.2GB of local checkpoints where a few hundred MiB would do.

Copy only the populated extents, located with `SEEK_DATA` and `SEEK_HOLE`, and fall back to the dense copy when the filesystem cannot report holes.

- [x] Tests pass

### DIFF
--- a/cmd/atelet/copyrange_linux.go
+++ b/cmd/atelet/copyrange_linux.go
@@ -35,10 +35,22 @@ func kernelCopyRange(srcFd, dstFd int, off, length int64) (int64, error) {
 	if length > maxKernelCopy {
 		length = maxKernelCopy
 	}
-	roff, woff := off, off
-	n, err := unix.CopyFileRange(srcFd, &roff, dstFd, &woff, int(length), 0)
-	if err != nil {
+	for {
+		// The offsets are only advanced by bytes the call actually copied, so a
+		// retry after EINTR resumes from the right place.
+		roff, woff := off, off
+		n, err := unix.CopyFileRange(srcFd, &roff, dstFd, &woff, int(length), 0)
 		switch {
+		case err == nil:
+			if n == 0 {
+				// No error and no progress: treat as unsupported rather than spin.
+				return 0, errKernelCopyUnsupported
+			}
+			return int64(n), nil
+		case errors.Is(err, unix.EINTR):
+			// A copy this large is interruptible, and the Go runtime signals
+			// goroutines for preemption, so this is expected rather than fatal.
+			continue
 		case errors.Is(err, unix.ENOSYS), // pre-4.5 kernel, or blocked by seccomp
 			errors.Is(err, unix.EXDEV),      // different filesystems
 			errors.Is(err, unix.EOPNOTSUPP), // filesystem does not implement it
@@ -46,12 +58,8 @@ func kernelCopyRange(srcFd, dstFd int, off, length int64) (int64, error) {
 			errors.Is(err, unix.EINVAL),     // ranges or flags this kernel rejects
 			errors.Is(err, unix.EBADF):      // not both regular files
 			return 0, errKernelCopyUnsupported
+		default:
+			return 0, err
 		}
-		return 0, err
 	}
-	if n == 0 {
-		// No error and no progress: treat as unsupported rather than spin.
-		return 0, errKernelCopyUnsupported
-	}
-	return int64(n), nil
 }
--- a/cmd/atelet/copyrange_linux.go
+++ b/cmd/atelet/copyrange_linux.go
@@ -1,0 +1,57 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"errors"
+
+	"golang.org/x/sys/unix"
+)
+
+// maxKernelCopy caps a single copy_file_range request. The syscall takes an int
+// length, and a bounded request keeps one call from monopolising the thread.
+const maxKernelCopy = 1 << 30
+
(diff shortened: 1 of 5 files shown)


BenTheElder:
@dberkov this would've cost a lot for local snapshots with uVM specifically, not sure what you were testing.

cmd/atelet/main.go
```diff
@@ -785,10 +809,105 @@ func copyFile(src, dst string) (int64, error) {
...
+
+	for off := int64(0); off < size; {
+		dataOff, err := unix.Seek(fd, off, unix.SEEK_DATA)
+		if err != nil {
+			if errors.Is(err, unix.ENXIO) {
+				break // no data past off; the tail is a hole
+			}
+			return fmt.Errorf("seeking to data at %d: %w", off, err)
+		}
+		holeOff, err := unix.Seek(fd, dataOff, unix.SEEK_HOLE)
```
dberkov, line 873: If unix.Seek(unix.SEEK_HOLE) returns holeOff <= off (for instance, if concurrent modifications or a non-standard filesystem implementation returns an offset that fails to advance past off), off = holeOff will fail to progress, causing an infinite loop. Additionally, if dataOff >= size is returned, unix.Seek(dataOff, unix.SEEK_HOLE) would query past EOF.

if dataOff >= size {
  break
}
holeOff, err := unix.Seek(fd, dataOff, unix.SEEK_HOLE)
if err != nil {
  return fmt.Errorf("seeking to hole at %d: %w", dataOff, err)
}
if holeOff <= off {
  return fmt.Errorf("seek to hole at %d returned non-advancing offset %d", dataOff, holeOff)
}

cmd/atelet/copyrange_linux.go
```diff
@@ -0,0 +1,57 @@
...
+const maxKernelCopy = 1 << 30
+
+// kernelCopyRange copies up to length bytes at off from srcFd to dstFd without the
+// data crossing into userspace, and reports how much it copied (short copies are
+// normal, so callers must loop).
+//
+// It reports errKernelCopyUnsupported when the kernel or filesystem cannot do the
+// copy — most commonly EXDEV, when source and destination are on different
+// filesystems — so the caller can fall back to a userspace copy.
+func kernelCopyRange(srcFd, dstFd int, off, length int64) (int64, error) {
```
dberkov, line 34: maxKernelCopy is set to 1 << 30 (1 GiB). Performing a 1 GiB copy_file_range syscall can take hundreds of milliseconds to seconds. In Go 1.14+, the Go runtime sends SIGURG signals for asynchronous goroutine preemption. If a signal is caught while copy_file_range is in progress before any bytes are transferred, unix.CopyFileRange returns unix.EINTR.

Currently, unix.EINTR is not handled in kernelCopyRange nor caught as unsupported in copySparse. As a result, copySparse returns copying ...: interrupted system call, aborting the entire checkpoint restore on a transient signal.

```
func kernelCopyRange(srcFd, dstFd int, off, length int64) (int64, error) {
	if length > maxKernelCopy {
		length = maxKernelCopy
	}
	roff, woff := off, off
	for {
		n, err := unix.CopyFileRange(srcFd, &roff, dstFd, &woff, int(length), 0)
		if errors.Is(err, unix.EINTR) {
			continue
		}
		if err != nil {
			switch {
			case errors.Is(err, unix.ENOSYS),
				errors.Is(err, unix.EXDEV),
				errors.Is(err, unix.EOPNOTSUPP),
				errors.Is(err, unix.EPERM),
				errors.Is(err, unix.EINVAL),
				errors.Is(err, unix.EBADF):
				return 0, errKernelCopyUnsupported
			}
			return 0, err
		}
		if n == 0 {
			return 0, errKernelCopyUnsupported
		}
		return int64(n), nil
	}
}
```

cmd/atelet/main_test.go
```diff
@@ -988,3 +988,172 @@ func TestDrainOnShutdownForceStopsAfterTimeout(t *testing.T) {
 		t.Fatal("readiness should be not-ready after drain")
 	}
 }
+
+// allocatedBytes reports how much disk a file actually occupies, which is less than its
+// size when it has holes.
+func allocatedBytes(t *testing.T, path string) int64 {
```
dberkov, line 994: nit:
copyrange_other.go was introduced so non-Linux platforms (e.g. macOS / Windows) can build cmd/atelet during local development. However, main_test.go lacks build tags, and syscall.Stat / syscall.Stat_t are not available on Windows (GOOS=windows), breaking go test ./cmd/atelet on Windows environments.

Suggested Fix: Move allocatedBytes and sparse assertions into main_linux_test.go (with //go:build linux), or wrap allocatedBytes with OS-specific helpers.

dberkov:
Is new behavior applied for both gVisor & microVMs ? I think gVisor already does all the optimization and the new code might make gVisor work slower.

cmd/atelet/main_test.go
BenTheElder: cmd/atelet doesn't build for windows as-is before this PR, I'm not sure if we care about windows.

BenTheElder:
> Is new behavior applied for both gVisor & microVMs ? I think gVisor already does all the optimization and the new code might make gVisor work slower.

That's a fair concern, but the old routine wasn't actually faster, it was a naive io.Copy loop which is generic around readers/writers. This is an optimized file copy routine. It micro-benchmarks faster even for dense files.

cmd/atelet/main.go
BenTheElder: done.

cmd/atelet/copyrange_linux.go
BenTheElder: fixed

cmd/atelet/main_test.go
BenTheElder: I'd go so far as to say we actively don't care about windows, except perhaps client tools/libraries. In-cluster tools it is useful to support macOS developers where mostly things are the same anyhow, but windows is so different and I doubt we will target windows workloads.

dberkov (approved):

PR #770, opened 2026-08-06
Extend the atelet drain time and tGPS for potential long checkpoint

HavenXia:
Large actor checkpoints may take a long time to finish, but the checkpoint itself runs through atelet, whose shutdown will cancel in-flight RPCs after 5 minutes. 

Increase `--drain-timeout` to 60m, and atelet's own `terminationGracePeriodSeconds` to 3630s so kubelet won't kill atelet  before the drain deadline, and log progress once a minute during long drains. 

Also add a log since the draining become extremely long now.

- [x] Tests pass

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -81,7 +81,7 @@ var (
 	logLevelFlag = pflag.String("log-level", "info", "Minimum log level: debug, info, warn, or error.")
 
 	drainDelay   = pflag.Duration("drain-delay", 0, "How long to keep accepting new RPCs after SIGTERM before starting the gRPC drain.")
-	drainTimeout = pflag.Duration("drain-timeout", 5*time.Minute, "Deadline for the graceful gRPC drain on shutdown. In-flight RPCs still running past it are forcefully cancelled.")
+	drainTimeout = pflag.Duration("drain-timeout", 60*time.Minute, "Deadline for the graceful gRPC drain on shutdown. In-flight RPCs still running past it are forcefully cancelled. Sized to the worker pod's 60m termination grace: a Checkpoint/Restore still in flight when atelet is terminated must not be cancelled before the worker's own grace expires.")
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


cmd/atelet/main.go
```diff
@@ -81,7 +81,7 @@ var (
 	logLevelFlag = pflag.String("log-level", "info", "Minimum log level: debug, info, warn, or error.")
 
 	drainDelay   = pflag.Duration("drain-delay", 0, "How long to keep accepting new RPCs after SIGTERM before starting the gRPC drain.")
-	drainTimeout = pflag.Duration("drain-timeout", 5*time.Minute, "Deadline for the graceful gRPC drain on shutdown. In-flight RPCs still running past it are forcefully cancelled.")
+	drainTimeout = pflag.Duration("drain-timeout", 60*time.Minute, "Deadline for the graceful gRPC drain on shutdown. In-flight RPCs still running past it are forcefully cancelled. Sized to the worker pod's 60m termination grace: a Checkpoint/Restore still in flight when atelet is terminated must not be cancelled before the worker's own grace expires.")
```
sfunkenhauser, line 84: While the ateom gets a long tGPS, initially we are only going to give actors 1 minute to shutdown.
https://github.com/agent-substrate/substrate/pull/764/changes#diff-8d90ca5fe4218dfa9c688e2ecf61f3f2c1a32c55e155f7c7dfb0c32f79b36da1R79

I think this timeout should correlate to the actor tGPS, and not necessarily the worker.

HavenXia: https://github.com/agent-substrate/substrate/blob/c155efd1acfe1ee7bf6ab489596d277801924098/cmd/ateom-gvisor/main.go#L353

hmm it seems the checkpoint will hold the lock, so whatever how long the checkpoint takes, the actor will not see an signal of SIGTERM (with the 1min `workloadGracePeriod`), it need to wait for the checkpoint ends. And if the checkpoint takes longer than 5 min, atelet will cancel it, which is why I raise this PR.

sfunkenhauser: That's good point, but I think the underlying issue is that the ateom doesn't kill the container.

Ultimately whenever the ateom receives a SIGTERM, it should start a timer for the actor tGPS. If the actor hasn't exited by then it should be SIGKILLed, even if it's in the middle of a checkpoint. (So I likely need to make some improvements to my PR).

That's why I think this drain timer should be correlated to the actor tGPS. The container should be killed regardless of what's happening after the actor tGPS is elapsed, so having a larger drain timeout here shouldn't be necessary.

HavenXia: That makes sense, in SIGTERM path it's the actor's 1min grace decides the SIGKILL, so either worker pod's grace or atelet's tGPS is not necessary.

But I just thought of a scenario that what if it's atelet Daemonset got restarted during a normal checkpoint process? A harness is trying to checkpoint a large actor which takes 30mins, and at 20min the atelet get a restart so it will cut the upload RPC? 

But honestly I don't have the data of a P95 or P95 checkpoint time, @maxsmythe do you know that number?

maxsmythe: I can give you it for very small images (and gVisor). For them it is well under 1 minute. P99 < 2.5 seconds, usually much less than that.

HavenXia: Thank you, then we are safe to keep the current 5min

PR #771, opened 2026-08-06
fix(ateapi): keep worker reclaimable when crashActor's release fails

### LINKED ISSUE #608: [P2] crashActor proceeds to UpdateActor even when releaseWorker fails — worker permanently consumed
**Severity:** P2 (worker slot consumed until pod death; no automatic recovery)
**Component:** Control Plane — `cmd/ateapi/internal/controlapi/crash.go`
**Audit ID:** CP-6

---

## Summary

`crashActor` (called whenever a workflow step detects the actor is dead) first calls
`releaseWorker` to clear the worker's `Assignment`, then calls `UpdateActor` to set
`STATUS_CRASHED`. If `releaseWorker.UpdateWorker` fails with a transient error (not
`ErrVersionConflict`), `crashActor` logs the error and continues to `UpdateActor`. The
actor is now `STATUS_CRASHED` (correct) but the worker record still has
`Assignment = <crashed-actor-uid>`. The scheduler skips workers with non-nil
assignments. This worker slot is permanently consumed until the pod is deleted.

Unlike the S1 scenario (ISSUE-011), this worker is assigned to a CRASHED actor — the
`AssignWorkerStep` recovery loop will never free it because CRASHED actors are not
re-resumed.

---

## Root Cause

**File:** `cmd/ateapi/internal/controlapi/crash.go` lines 51–76

```go
func crashActor(ctx context.Context, ...) error {
    if err := releaseWorker(...); err != nil {
        errCollected = append(errCollected, err)  // logged but execution continues
    }
    // Always runs regardless of releaseWorker success:
    if err := store.UpdateActor(..., Status: STATUS_CRASHED); err != nil { ... }
    return errors.Join(errCollected...)
}
```

`ErrVersionConflict` from `releaseWorker` is safe (a concurrent writer cleared the
assignment). But a transient Valkey timeout produces a non-`ErrVersionConflict` error
that causes the leak.

---

## Steps to Reproduce

1. Have a RUNNING actor.
2. Inject a Valkey write timeout that fires on the next `UpdateWorker` call (use a proxy
   or Valkey ACL to drop one write).
3. Trigger `maybeCrashActor` (e.g., by having atelet return `actorCrashed = true` on
   a checkpoint call, or via `SuspendActor` with a broken atelet).
4. Inspect Valkey after the crash:
   ```bash
   kubectl exec -n ate-system valkey-cluster-0 -- \
     valkey-cli HGETALL workers:<worker-uid>
   # Assignment field is still set to the crashed actor UID
   ```
5. Check actor status — correctly CRASHED.
6. Observe the worker is never freed (scheduler skips it); pool capacity permanently -1.

---

## Expected Behavior

If `releaseWorker` fails with a transient error, `crashActor` should:
- Either retry `releaseWorker` (with backoff), or
- Refuse to mark the actor CRASHED until the worker release succeeds, so a retry of
  `crashActor` can complete the full sequence.

---

## Suggested Fix

Change `crashActor` to return early (without calling `UpdateActor`) if `releaseWorker`
fails with a non-`ErrVersionConflict` transient error. The caller should retry the
crash sequence on the next control-plane event. Alternatively, add a startup scan that
detects workers assigned to CRASHED actors and clears them (similar to the fix for
ISSUE-011).

aramase:
Fixes #608

`crashActor` collected the `releaseWorker` error but still marked the actor `CRASHED` and cleared its `WorkerAssignment`, stranding the still-assigned worker with no actor left to reclaim it. On a release failure, return instead so the actor stays intact and the workflow's retry reclaims the worker (`releaseWorker` is idempotent).

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/crash.go
+++ b/cmd/ateapi/internal/controlapi/crash.go
@@ -65,10 +65,17 @@ func crashActor(ctx context.Context, st store.Interface, actorRef resources.Acto
 		reason = ateattr.ReasonUnknown
 	}
 
-	var errCollected []error
+	// Release the worker before moving the actor to the terminal CRASHED state.
+	// If the release fails we must not clear the actor's worker assignment or
+	// mark it CRASHED: doing so would strand the still-assigned worker with no
+	// actor referencing it, so nothing would ever retry the release and the
+	// worker slot would be consumed until its pod dies. Returning the error
+	// instead leaves the actor (and its assignment) intact so the caller retries
+	// crashActor, which re-attempts the release. releaseWorker is idempotent, so
+	// a retry after a release that already succeeded is a no-op.
 	sandboxClass, err := releaseWorker(ctx, st, actor)
 	if err != nil {
-		errCollected = append(errCollected, err)
+		return fmt.Errorf("while releasing worker to crash actor: %w", err)
 	}
 
 	// Snapshot crash attributes before pod and pool pointers are cleared below;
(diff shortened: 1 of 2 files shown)


sfunkenhauser (dismissed):

aramase:
@sfunkenhauser thanks for the approval! I had to rebase the PR, could you review and merge?

sfunkenhauser (approved):

PR #772, opened 2026-08-06
fix(ateapi): retry stored-worker scan in syncer startup

### LINKED ISSUE #609: [P2] Startup orphan reconciliation aborts on ListWorkers pagination error — ghost workers persist
**Severity:** P2 (orphaned workers persist until next ateapi restart)
**Component:** Control Plane — `cmd/ateapi/internal/controlapi/syncer.go`
**Audit ID:** CP-8
**Maps to suspect:** S2

---

## Summary

`reconcileOrphanedWorkers` runs once at ateapi startup and paginates through all
worker records. If any `store.ListWorkers` call returns an error (e.g., Valkey is
intermittently slow during startup), the function returns immediately — skipping all
workers on remaining pages. Those workers' orphaned assignments and stale actors persist
until the next ateapi restart. The function logs the error but does not schedule a retry.

---

## Root Cause

**File:** `cmd/ateapi/internal/controlapi/syncer.go` lines 234–270

```go
func (s *WorkerPoolSyncer) reconcileOrphanedWorkers(ctx context.Context) error {
    // ...
    for {
        workers, err := store.ListWorkers(ctx, cursor)
        if err != nil {
            s.logger.Error("Failed to list workers during startup reconcile", ...)
            return err  // ABORTS — remaining pages are never processed
        }
        for _, w := range workers {
            if err := s.reconcileDeadWorker(ctx, w); err != nil {
                s.logger.Error(...)  // individual error: continues
            }
        }
        if !more {
            break
        }
    }
    return nil
}
```

Individual `reconcileDeadWorker` failures (line 267) continue to the next worker, but
a `ListWorkers` pagination error aborts the entire function.

---

## Steps to Reproduce

1. Have many workers in Valkey (enough to span multiple pages, typically > 100).
2. Delete all worker pods while ateapi is down.
3. Configure Valkey to return a transient error on the second `HSCAN` call (use a proxy
   or temporarily set `maxmemory-policy = noeviction` and fill Valkey to trigger OOM
   errors on writes, or use a packet filter timed to fire on the second scan).
4. Start ateapi; observe in logs:
   ```
   Failed to list workers during startup reconcile: <error>
   ```
5. Verify workers from the first page are cleaned up; workers from the second+ pages
   remain with stale actor assignments:
   ```bash
   kubectl ate list actors -a <atespace>
   # Shows actors in STATUS_RUNNING or STATUS_RESUMING that have no running pods
   ```

---

## Expected Behavior

A `ListWorkers` error during startup reconciliation should either:
1. Retry the entire reconciliation after a backoff delay, or
2. Log a warning and continue to the next cursor position (best-effort reconciliation).

---

## Suggested Fix

Wrap the reconciliation loop in a retry with exponential backoff:

```go
func (s *WorkerPoolSyncer) reconcileOrphanedWorkers(ctx context.Context) {
    backoff := wait.Backoff{...}
    wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool, error) {
        err := s.runOrphanReconcilePass(ctx)
        if err != nil {
            s.logger.Warn("Orphan reconcile pass failed, will retry", "err", err)
            return false, nil
        }
        return true, nil
    })
}
```

aramase:
Fixes #609

The startup stored-worker scan (`enqueueStoredWorkers`) returned on the first `ListWorkers` error, skipping orphan cleanup until the next restart. The `#674` workqueue only retries per-key reconciles, not this initial scan, so retry the scan with capped `wait.Backoff` until it succeeds or the context is cancelled.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/syncer.go
+++ b/cmd/ateapi/internal/controlapi/syncer.go
@@ -340,15 +340,33 @@ func (s *WorkerPoolSyncer) reconcileDeadWorker(ctx context.Context, namespace, p
 	return s.persistence.DeleteWorker(ctx, namespace, pool, podName)
 }
 
+// storedWorkerListBackoff and storedWorkerListCap are the exponential backoff
+// schedule for retrying a failed page of the startup stored-worker scan. They
+// are vars so tests can shrink them.
+var (
+	storedWorkerListBackoff = 500 * time.Millisecond
+	storedWorkerListCap     = 30 * time.Second
+)
+
 // enqueueStoredWorkers enqueues a key for every worker record in the store.
 // Records whose pods are live and unchanged reconcile to a no-op; orphaned
 // records (pod gone, or its name reused by a new pod UID) get cleaned up.
+//
+// Each page's ListWorkers call is retried with capped backoff until it succeeds
+// or ctx is cancelled, so a transient store error does not abandon the scan and
+// leave ghost workers behind until the next restart (the per-key workqueue
+// retries reconciles, but nothing retries this initial enqueue scan). Pages are
+// enqueued as they are read, so the whole worker set is never held in memory at
+// once and a late failure does not re-scan the pages already enqueued.
 func (s *WorkerPoolSyncer) enqueueStoredWorkers(ctx context.Context) {
 	var pageToken string
 	for {
-		page, err := s.persistence.ListWorkers(ctx, store.ListOptions{PageSize: 1000, PageToken: pageToken})
+		page, err := s.listWorkersPageWithRetry(ctx, pageToken)
 		if err != nil {
-			slog.ErrorContext(ctx, "Syncer: failed to list workers for orphan reconcile", slog.Any("err", err))
+			// Only ctx cancellation (ate-api-server shutdown) ends the retry
+			// loop. Pages read so far are already enqueued (partial progress);
+			// the rest are recovered by the next startup scan.
+			slog.ErrorContext(ctx, "Syncer: stopped enqueue of stored workers before completing the scan; remaining workers will be retried at the next startup", slog.Any("err", err))
 			return
 		}
 		for _, w := range page.Items {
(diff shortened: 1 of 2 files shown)


cmd/ateapi/internal/controlapi/syncer.go
```diff
@@ -310,19 +310,77 @@ func (s *WorkerPoolSyncer) reconcileDeadWorker(ctx context.Context, namespace, p
...
+	storedWorkerListCap     = 30 * time.Second
+)
+
+// listAllStoredWorkers paginates every worker record, retrying the whole scan
+// with capped exponential backoff until it succeeds or ctx is cancelled. A scan
+// that gave up after a fixed number of attempts could still skip the orphan
+// cleanup until the next restart if the store stayed briefly unavailable; the
+// retry runs in the background startup goroutine (off the serving path) and, on
+// a lasting outage, issues at most one cheap failing list per Cap interval.
+func (s *WorkerPoolSyncer) listAllStoredWorkers(ctx context.Context) ([]*ateapipb.Worker, error) {
```
sfunkenhauser: Can we add the retry logic around the API call, and then just process 1 page of workers at a time so we don't need to load all workers into memory?

sfunkenhauser (changes requested):

cmd/ateapi/internal/controlapi/syncer.go
aramase: Done. It now reads and enqueues one page at a time, so we never hold all workers in memory, and the retry wraps the single `ListWorkers` call with capped backoff and reuses the same page cursor instead of restarting from page 0. Enqueue is incremental now, so a mid-scan cancel keeps the pages already queued and the rest get picked up next startup. Added a multi-page test that fails a later page once and checks every worker still gets enqueued.

sfunkenhauser (approved):

aramase:
@sfunkenhauser the token e2e failure looks like an unrelated resume/readyz flake (`demo_test.go:543`, sandbox restore path). Cert e2e and unit tests passed on the same commit, and this PR only touches `syncer.go`. Could you re-run the token job? I don't have permissions.

sfunkenhauser:
> sfunkenhauser the token e2e failure looks like an unrelated resume/readyz flake (`demo_test.go:543`, sandbox restore path). Cert e2e and unit tests passed on the same commit, and this PR only touches `syncer.go`. Could you re-run the token job? I don't have permissions.

Re-running.

Issue #719, opened 2026-08-03
Graceful shutdown for Atelet


HavenXia:
Atelet needs `SIGTERM` handling:  the same shutdown pattern `ate-api` uses — on `SIGTERM` stop accepting new RPCs, let in-flight ones finish, force-stop after a timeout.

Out of scope: what happens to running actors when eviction (ateom capturing the signal and forward to actors) — that's the graceful worker termination track in #517.

Issue #473, opened 2026-07-20
Substrate Infrastructure Upgrade mechanism


HavenXia:
We need to figure out a best upgrade mechanism for upgrading the substrate system. 

I have in-place upgrade, in-cluster green/blue and across cluster green/blue - 3 mode to evaluate. We need to take cost and performance into consideration.

HavenXia:
Some general ideas




## Option1: in-place rolling upgrade

Upgrade the existing install component by component: apply new images, let each Deployment roll, during the upgrade both old and new components are accepting traffic.

```
        clients
           │   (endpoint never moves)
           ▼
┌─ cluster ───────────────────────────┐
│  one stack, upgraded in place:      │
│  v1 pods and v2 pods run mixed      │
│  while each Deployment rolls        │
│              │                      │
│              ▼                      │
│      Valkey: both versions          │
│      write it at the same time      │
└─────────────────────────────────────┘
```



## Option2: Blue/green inside the same cluster

Install a second, renamed copy of the stack ("green") in the same cluster, plus a new worker node pool. Both stacks can reach the same Valkey. To upgrade: suspend all actors, switch the endpoint to green, resume the actors on green, then delete the blue stack.

```
        clients
           │   (endpoint flips: blue → green)
           ▼
┌─ cluster ─────────────────────────────────┐
│  blue stack (v1)      green stack (v2)    │
│  old node pool        new node pool       │
│        │                    │             │
│        └─────────┬──────────┘             │
│                  │                        │
│                  ▼                        │
│         Valkey (shared, stays put)        │
└───────────────────────────────────────────┘
```

## Option3: Blue/green with a new cluster

Create a brand-new cluster running the new version, get the data over, point clients and DNS at it, delete the old cluster.

```
        clients
           │   (DNS / address repoints to the new cluster)
           ▼
┌─ old cluster ─────┐     ┌─ new cluster ─────┐
│  stack v1         │     │  stack v2         │
│  workers          │     │  workers          │
└────────┬──────────┘     └────────┬──────────┘
         │                         │
         └────────────┬────────────┘
                      ▼
        store reachable from both clusters:
        external Valkey - (does not exist today),
        or an export / import mechanism we build?
```

HavenXia:
The current proposal is at http://doc/1JduAyGZFyqdNp4UhKiv0EN-is3iW5tf5BwGTWbt2ouI?resourcekey=0-MXG12QCkleIxhY7cOB7g6A.

[...]

PR #774, opened 2026-08-06
feat(atenet): graceful termination of atenet-router

### LINKED ISSUE #721: Graceful shutdown for atenet router
The atenet router needs `SIGTERM` handling — the same shutdown pattern `ate-api` uses: flip readiness so the Service stops sending new connections, let in-flight requests finish, drain the Envoy sidecar instead of letting it die mid-connection, and set `terminationGracePeriodSeconds` long enough to also outlive the parking
budget (`--parked-request-budget`), so parked requests finish normally instead of resetting.

shrutiyam-glitch:
### Which issue(s) this PR is related to:
Fixes #721
Required for System Upgrade flow (#473)

### What this PR does / why we need it:
This PR implements graceful termination and zero-outage rolling updates for `atenet-router` , building on the readiness probe and graceful drain patterns established in #719 (`atelet`).

Draining a two-container networking pod (`atenet-router` Go control-plane + `envoy` C++ proxy dataplane) introduces complex lifecycle interdependencies. This PR resolves the dual-container SIGTERM race, respects Envoy's `failClosed` `ext_proc` filter dependency, preserves parked requests riding out worker pool saturation, and accelerates idle deployments via an event-driven file handshake.

#### 1. Event-Driven Dataplane Synchronization (`emptyDir` Marker File)
Envoy fast-exits by default on SIGTERM. To keep Envoy alive while the Go control-plane orchestrates the drain, we configured an IPC handshake between containers via a pod-shared `emptyDir` volume mounted at `/var/run/atenet`:

*   **Go Router Container:** Removes any stale marker at startup, then writes `/var/run/atenet/drain-complete` when its shutdown sequence finishes.
*   **Envoy Container `preStop` Hook:** Runs `while [ ! -f /var/run/atenet/drain-complete ]; do sleep 0.5; done`.

**Outcome:** Envoy exits as soon as — the drain is done, rather than wasting time in a fixed worst-case sleep. If the router crashes, Kubelet terminates the `preStop` hook at `terminationGracePeriodSeconds: 60`—slower cleanup, never a wedge.

#### 2. The Multi-Container Shutdown Sequence (`drain.go`)
Because Kubernetes issues SIGTERM to both containers at once, a coordination state machine ensures Envoy never drops a connection and `ext_proc` is never stopped prematurely:

| Phase / (best-case ex.) Timeline | atenet-router | envoy | K8s / Service Status |
| :--- | :--- | :--- | :--- |
| **SIGTERM Sent**<br>($t=0\text{s}$) | Catches SIGTERM<br>• Flips `/readyz` $\rightarrow$ `503`.<br>• Starts 13s `drain-delay`. | Enters `lifecycle.preStop` hook:<br>`while [ ! -f .../drain-complete ]; do sleep 0.5; done`<br>• Kubelet holds SIGTERM back. | Deployment will recreate the pod.<br>EndpointSlice controller begins dropping Old Pod IP. |
| **Propagation**<br>($t=0\text{s}$ – $t=13\text{s}$) | Keeps serving normally.<br>• `ext_proc` continues unparking/routing requests. | Runs normally inside `preStop` loop.<br>• Serves active TCP connections. | Service endpoint removal completes.<br>No new connections arrive at Old Pod. |
| **Envoy drain**<br>($t=13\text{s}$) | Issues Envoy admin API calls:<br>• `/healthcheck/fail`<br>• `/drain_listeners?graceful&skip_exit`<br>• Polls `/stats` for active downstream connections. | Begins graceful listener drain:<br>• GOAWAY / `Connection: close` on established connections.<br>• In-flight requests keep running. | In-flight HTTP requests finish executing through Envoy. |
| **ext_proc drain**<br>($t\approx18\text{s}$) | Active Envoy connections hit 0 (the poll exits early).<br>• Calls `extproc.GracefulStop()`.<br>• Parked request streams finish. | All downstream connections closed.<br>• Still waiting in the `preStop` loop. | All client HTTP responses delivered. |
| **Handshake & Exit**<br>($t=18.5\text{s}$)| `ext_proc` drain finishes.<br>• Writes `drain-complete` marker.<br>• Hard-stops xDS (`Stop()`) & exits. | `preStop` loop detects marker file!<br>• `preStop` exits 0.<br>• Kubelet sends SIGTERM to Envoy.<br>• Envoy exits immediately. | Old Pod deleted cleanly at ~18.5s (well under the 60s budget). |

Envoy ref: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/draining 

#### Testing & Verification

1. Unit Tests (`drain_test.go`): Added comprehensive unit tests covering:
- Graceful completion of in-flight `ext_proc` streams within deadline.
- Force-stopping straggler streams past `--drain-timeout`.
- Envoy admin API interaction and connection polling.
- Stale marker cleanup and marker file creation.

2. Local testing on kind cluster (Observations):
Test script: `hack/verify-atenet-drain.sh`
  - Steady state: `/readyz=200`, `/healthz=200`; the instant the pod turned Terminating: `/readyz=503` while `/healthz=200` — `NotReady` but alive, for the whole drain.
  - The parked request survived the shutdown: fired at a busy 1-worker pool → parked → pod deleted while parked → worker freed → HTTP=200, total=1.89s, body hello from: 169.254.17.2 | preserved memory count: 1 |
  preserved file counter: 1 — park → resume → route, served by the Terminating pod during its drain-delay window.
  - Pod terminated 14s after deletion — the idle floor, confirmed across three runs: >13s drain-delay (sequence ran), ≪60s grace (marker handshake released Envoy's preStop; no `SIGKILL`).
  - Log sequence captured verbatim, drain-delay honored to the millisecond (17:34:41.706 → 17:34:54.707):
  >Shutdown signal received; draining
  >(+13.000s) Draining Envoy  {window: 15s}
  >Envoy drained
  >Starting ext_proc drain
  >ext_proc drain completed within deadline
  >Drain-complete marker written  {path: /var/run/atenet/drain-complete}
  >Shutdown complete

- [x] Tests pass


cmd/atenet/internal/router/drain.go
```diff
@@ -0,0 +1,143 @@
...
+		go func() {
+			p.extproc.GracefulStop()
+			close(drainComplete)
+		}()
+		select {
+		case <-drainComplete:
+			slog.InfoContext(ctx, "ext_proc drain completed within deadline")
+		case <-time.After(p.timeout):
+			slog.WarnContext(ctx, "ext_proc drain deadline exceeded; forcing stop")
+			p.extproc.Stop()
```
LiorLieberman, line 156: lets add a cmoment and open an issue to add metrics here. Would be good to have a sense of how many unclean shutdowns we had

cmd/atenet/internal/router/drain.go
```diff
@@ -0,0 +1,143 @@
...
+// window needs ext_proc answering.
+type drainParams struct {
+	readiness *serverboot.Readiness
+	// delay is the route-drain window: after the readiness flip, how long to
+	// keep serving while the Service endpoints drop this pod.
+	delay time.Duration
+	// drainEnvoy gracefully drains the dataplane sidecar; nil when there is
+	// none to drive (agentgateway mode). It is handed a context bounded by
+	// envoyWindow and must return when it expires.
+	drainEnvoy  func(context.Context) error
```
LiorLieberman: what is "none to drive"

suggest rewording to be more clear.

Also a function as in the parmas is not great. can we have something like "withDrainFunc" that decorates it or something?

cmd/atenet/internal/router/drain.go
```diff
@@ -0,0 +1,143 @@
...
+// drainParams wires the shutdown sequence. The order is forced by the
+// dataplane: ext_proc is failClosed for Envoy, so the ext_proc server must
+// outlive Envoy's drain — any request Envoy still accepts during its drain
+// window needs ext_proc answering.
+type drainParams struct {
+	readiness *serverboot.Readiness
+	// delay is the route-drain window: after the readiness flip, how long to
+	// keep serving while the Service endpoints drop this pod.
+	delay time.Duration
+	// drainEnvoy gracefully drains the dataplane sidecar; nil when there is
```
LiorLieberman: Can we remove Envoy and agentgateway from the names here and everywhere in the PR? You can have drainDataplaneFunc (or drainDataplane) or something and just say that when set to nil, it simply means the dataplane doesn't require active graceful draining (or manages its own lifecycle). This keeps the generic router orchestrator cleanly decoupled from the underlying proxy implementation.

LiorLieberman (commented):
Thanks @shrutiyam-glitch !

cmd/atenet/internal/router/drain.go
shrutiyam-glitch: Done. Thanks !

LiorLieberman:
LGTM

PR #775, opened 2026-08-06
Adds hack/dev-branch.sh

bowei:
This script automates creating an isolated dev branch in a git worktree for development. It also creates a separate Kind cluster for e2e development for the environment.

Adds agent skill to use the isolated dev branch for work.

### DIFF
--- a/.agents/skills/dev-branch/SKILL.md
+++ b/.agents/skills/dev-branch/SKILL.md
@@ -1,0 +1,106 @@
+---
+name: dev-branch
+description: Parallel development using hack/dev-branch.sh to isolate worktrees and Kind test clusters per feature.
+---
+
+# Multi-Agent Parallel Development with `hack/dev-branch.sh`
+
+This skill provides workflow guidance for executing concurrent development tasks,
+working on separate features within the `agent-substrate/substrate` repository.
+
+By isolating each feature inside its own **Git worktree** and **Kind Kubernetes
+cluster**, multiple agents can build, test, and debug features in parallel
+without step-on or port/cluster resource contention.
+
+---
+
+## Architecture & Principles
+
+| Resource | Isolation Strategy | Purpose |
+| :--- | :--- | :--- |
+| **Source Code** | Git Worktree (`.worktrees/<branch>`) | Independent code branch & checkout per agent |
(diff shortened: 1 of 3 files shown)


.agents/skills/dev-branch/SKILL.md
```diff
@@ -0,0 +1,106 @@
...
+name: dev-branch
+description: Parallel development using hack/dev-branch.sh to isolate worktrees and Kind test clusters per feature.
+---
+
+# Multi-Agent Parallel Development with `hack/dev-branch.sh`
+
+This skill provides workflow guidance for executing concurrent development tasks,
+working on separate features within the `agent-substrate/substrate` repository.
+
+By isolating each feature inside its own **Git worktree** and **Kind Kubernetes
```
BenTheElder, line 11:
```suggestion
By isolating each feature inside its own **Git worktree** and **kind Kubernetes
```

hack/delete-kind-cluster.sh
```diff
@@ -40,8 +40,13 @@ fi
 if [ "${reg_exists}" == true ]; then
   reg_created_by="$(docker inspect --format '{{index .Config.Labels "created-by"}}' "${reg_name}" 2>/dev/null)"
   if [ "${reg_created_by}" == "agent-substrate" ]; then
-    echo "Deleting registry container '${reg_name}' (created by us)..."
-    docker rm -f "${reg_name}" || true
+    remaining_clusters="$("${ROOT}"/hack/kind.sh get clusters 2>/dev/null | grep -v 'No kind clusters found' | grep -v '^$' || true)"
```
BenTheElder, line 43: this script was intended to be the counter to create-kind-cluster.sh, I think if we're going to delete *all* clusters that should be optional or a mode

currently if you are doing other things on the same host with kind, we won't delete that, just the one you requested. now this PR wipes unrelated clusters

botengyao (approved):

hack/delete-kind-cluster.sh
BenTheElder: This change in particular I think needs to be refined. Why can't an agent just tear down its own cluster?

bowei: Yes, good point... let me update

bowei:
Redoing this ..

bowei:
Reopen

PR #776, opened 2026-08-06
feat(otel): add cold start metrics

krisztianfekete:
This PR implements the last two metrics from #433.

Right now, we can see that a resume was slow but not where. `ate.actor.lifecycle.operation.duration` covers the whole ateapi operation, and `atenet.router.route.duration` covers the edge, but everything between ateapi-atelet-actors is one block that contains fetching the manifest, downloading the snapshot, unpack the OCI image, call to ateom. 

We have `rpc.server.call.duration` that gives us the atelet restore total time, but template, kind, and scope labels are missing, so today we cannot really pinpoint why/where we have a regressions in latency.

In this PR I am adding per-phase histograms, here's an example of what we can know after these changes:

```console
ateom_restore   522 ms   ###############################
download        8.9 ms   #
manifest_fetch  3.2 ms
oci_unpack      2.6 ms
---------------------------------------------------------
total           535 ms
```

It also fixes a gap #683 opened where a `data_on_golden` resume was labeled identically to a plain one on the lifecycle histogram.

Things folks might want to argue with:
- total as a phase value. Partly duplicates `rpc.server.call.duration`, but that one has no domain labels and gRPC-specifc. We can drop it, but it's an inferior operational UX, so I'd rather have it here.
- Phases overlap, they are not a partition of total, because download runs concurrently with the asset fetch and unpack. I called this out in the metric description. Do not sum across phases.
- New `ate.snapshot.scope` key rather than a new `ate.snapshot.kind` value for `data_on_golden`. A new value would collapse local and external into one bucket, which is the biggest latency difference there is. This does add a label to the already shipped lifecycle histogram.

Verified on kind, and all e2e suites pass, and the emitted series cover every kind (golden, latest, local) on both metrics with no unknown values.
 
- [x] Tests pass


krisztianfekete:
Adding the last two metrics from #433, cc. @JeffLuoo, @git286.

cmd/atelet/main.go
```diff
@@ -121,6 +121,11 @@ func main() {
 		serverboot.Fatal(ctx, "Failed to create snapshot size metric", err)
 	}
 
+	instruments, err := NewInstruments(otel.Meter("atelet"))
+	if err != nil {
+		serverboot.Fatal(ctx, "Failed to create cold-start metrics", err)
```
JeffLuoo: nit - maybe not use `cold-start metrics` as the NewInstruments may cover more than just cold-start metrics in the future. Consider `atelet metrics` for metrics created under atelet folder? This is not a strong opinion, up to you : )

cmd/atelet/main.go
```diff
@@ -404,10 +435,14 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
...
 		return nil, ateerrors.CrashIfReason(ctx, err, ateerrors.ReasonInvalidSandboxAsset, ateerrors.ReasonTerminalFileSystemError)
 	}
+	op.sandboxClass = sandboxRec.SandboxClass
+
+	tAssets := time.Now()
 	assetPaths, err := s.ensureSandboxAssets(ctx, sandboxRec)
 	if err != nil {
 		return nil, ateerrors.CrashIfReason(ctx, err, ateerrors.ReasonInvalidSandboxAsset, ateerrors.ReasonTerminalFileSystemError, ateerrors.ReasonFailedGetExternalObject, ateerrors.ReasonInvalidObjectURL)
 	}
+	dAssets = time.Since(tAssets)
```
JeffLuoo: `error.type` is appended to the *base* attribute set, so `recordPhases` stamps it on every phase of the operation. Second, every clock in
  `main.go` is stopped *after* the error check — e.g. `dAteom = time.Since(tAteom)` at `main.go:474` (and `:798` for restore) sits below the `if err != nil { return }`, so
  on failure `dAteom` stays 0 and `recordPhases` skips it.

  Net effect for a restore that dies in `ateom_restore` with `Unavailable`:

  | phase | emitted? | error.type |
  |---|---|---|
  | `manifest_fetch` | yes | `Unavailable` ← succeeded |
  | `download` | yes | `Unavailable` ← succeeded |
  | `oci_unpack` | yes | `Unavailable` ← succeeded |
  | `ateom_restore` | **no** | — ← the one that failed |
  | `total` | yes | `Unavailable` ✔ |

  So neither question this metric exists to answer is answerable: per-phase latency is polluted (unfiltered mixes error series in; `error_type=""` drops the healthy
  `download` observations from every failed restore), and the failing phase is invisible. 

Potential fixes:

  1. Move each `d* = time.Since(...)` above its `if err != nil` return, in both `Checkpoint` and `Restore`.
  2. Drop the `err` param from `snapshotOp.attrs`, pass `err` to `recordPhases` instead, and apply the label only when `p.name == ateattr.SnapshotPhaseTotal`.

Also make sure update the unit test to cover it.

JeffLuoo (changes requested):

cmd/atelet/main.go
krisztianfekete: Thanks for catching these!  Moved all timer stops around, and extended your option 2 a bit, so the label lands on the failing phase and the total, so `{phase="download", ...!=""}` tells us where the failure is, while successful phases stay unlabeled.

Also swapped `error.type` for `ate.failure.reason` since a gRPC code read `Unknown` for nearly every failure here. Verified on kind by killing object storage. Persist and total get `FAILED_SAVE_SNAPSHOT`, the phases that succeeded stay as-is. 

This leaves `ate.actor.lifecycle.operation.duration` as the only instrument still using `error.type`, we can make it consistent though.

cmd/atelet/main.go
```diff
@@ -683,22 +772,30 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
 			return ateerrors.CrashIfReason(ctx, err, ateerrors.ReasonTerminalFileSystemError, ateerrors.ReasonInvalidContainerConfig)
 		}
-		dBundles = time.Since(t)
 		return nil
 	})
 	if err := g.Wait(); err != nil {
+		op.failedPhase = downloadFailedPhase
+		if op.failedPhase == "" {
+			op.failedPhase = prepFailedPhase
+		}
```
JeffLuoo: `failedPhase` and the error can come from different goroutines, so the reason gets pinned to the wrong phase.

`errgroup.WithContext` cancels `gctx` as soon as either goroutine fails. If the asset/OCI-prep goroutine fails first, the in-flight download aborts with a context error and its defer sets `downloadFailedPhase` — even though it was only collateral. Here `download` wins unconditionally, while `g.Wait()` returned the prep error. The datapoint reads:

  `ate_snapshot_phase="download", ate_failure_reason="INVALID_CONTAINER_CONFIG"`

  The error is correct; the phase isn't. On a cold node the download is the long-running one, so it's still in flight almost every time prep fails.

  `Wait` returns the first error verbatim, so identity tells you which goroutine owns it:

  ```go
  var downloadErr, prepErr error
  
  g.Go(func() (err error) {
      t := time.Now()
      defer func() {
          dDownload = time.Since(t)
          downloadErr = err
          if err != nil {
              downloadFailedPhase = ateattr.SnapshotPhaseDownload
          }
      }()
      ...
  })

  g.Go(func() (err error) {
      defer func() { prepErr = err }()
      ...
  })

  // The cancelled goroutine's error is only gctx fallout; it must not claim the
  // phase.
  if err := g.Wait(); err != nil {
      switch err {
      case downloadErr:
          op.failedPhase = downloadFailedPhase
      case prepErr:
          op.failedPhase = prepFailedPhase
      }
      return nil, err
  }
  ```

  `errors.Is(err, context.Canceled)` won't work as a filter here — `CrashIfReason` and the GCS wrapping don't reliably preserve it.

JeffLuoo (changes requested):

cmd/atelet/main.go
JeffLuoo: lgtm

cmd/atelet/main.go
krisztianfekete: Thanks!
Fixed this and while I testing I found the same bug in the durations, which is also fixed now. Also swapped `ate.snapshot.file` for the semconv's `file.name` on `atelet.snapshot.size`.

JeffLuoo (dismissed):

JeffLuoo:
Please rebase, thanks

JeffLuoo (approved):

PR #781, opened 2026-08-06
atenet/router: run the parking tests on a fake clock

bowei:
Use a fake clock and synctext to avoid waiting in real-time in the unit test.

Addresses review comments left over from #221.


haiyanmeng (approved):
/lgtm

PR #784, opened 2026-08-06
Some improvements to the way we allocate and track snapshot ids

### LINKED ISSUE #773: Some improvements to actor snapshots
There are a few things about snapshots which I would like to revisit before we commit to it. 

### Terminology and concepts

When we suspend, we generate a snapshot ID, which encodes the timestamp at which the snapshot is created plus a random suffix. They look like this: `2026-08-05T16:47:28-04:00-N7GBFQ2XKJ4ZLWMR6TYHPVCD3A`.

When the snapshot is uploaded to GCS, we construct a snapshot URL, which consist of the base path on the ActorTemplate and the snapshot ID. It looks like this:

```
gs://my-bucket/snapshots/secret-agent/snapshots/2026-08-05T16:47:28-04:00-N7GBFQ2XKJ4ZLWMR6TYHPVCD3A
```

Note that the ID goes into that URI **raw** — those colons end up verbatim in the GCS object names. The sanitizing happens somewhere else entirely: after the upload, when we finalize the suspend, we parse the ID back out of the URI and rewrite it (`:`→`-`, `+`→`-`, plus lowercasing) to get a valid Substrate resource name, because `ObjectRef.name` is validated as a DNS-1123 label.

So one snapshot ends up with two IDs:

| | |
|---|---|
| in GCS | `2026-08-05T16:47:28-04:00-N7GBFQ2XKJ4ZLWMR6TYHPVCD3A` |
| in the API (`ActorSnapshot.metadata.name`) | `2026-08-05t16-47-28-04-00-n7gbfq2xkj4zlwmr6typvcd3a` |

Note that the rewrite is one-way: `:` and `+` both become `-`, so you can't get from the API name back to the path (`t16-47-28-04-00` — was that `-04:00` or `+04:00`?). 

I think we need to simplify the concepts and use consistent terminology throughout the codebase. 

* There is no snapshot ID. When we create a new snapshot, we allocate a new snapshot name, which is used for both the final snapshot URI and the ActorSnapshot name.
* We don't encode semantic information into the name. Today we try to encode a timestamp, but it's not really needed. The `ActorSnapshot` resource has a `create_time` timestamp already.

* We have three concepts:
   * Snapshot location: This is the base path which is dictated by `ActorTemplate.spec.snapshotsConfig.Location`.
   * Snapshot name: The identifier of the snapshot. A single UUID. Opaque. 
   * Snapshot URI: The address of a given snapshot. This is `<location>/snapshots/<atespace>/<snapshot-name>`.

This removes the confusion about snapshot name, snapshot id (sanitized vs not sanitized), etc. 

### ActorSnapshot has an internal version in storage

The ActorSnapshot representation in ValKey today is a json wrapper of the proto, which includes the (not sanitized!) snapshot ID.
We should just add a `snapshot_uri` to the `ActorSnapshot` proto.

### Bookkeeping during operations

Today the suspend and pause workflows are reusing the `in_progress_snapshot` field in  the `Actor` proto for storing two different concepts. We should split them.

juli4n:
Fixes #773 

Some improvements to the way we create, track and propagate snapshot ids.

* Consolidate snapshot id and name into a single concept.
* Use consistent terminology across the whole stack (i.e. remove confusion about prefix vs id vs name).
* Get rid of internal only storage to hide the snapshot URI.


internal/resources/snapshot.go
```diff
@@ -0,0 +1,100 @@
...
+//
+//	gs://bucket/root                              location
+//	gs://bucket/root/snapshots/team-a/<name>      this URI
+//	gs://bucket/root/snapshots/team-a/<name>/...  an object in the snapshot
+type SnapshotURI struct {
+	uri      string
+	location string
+	atespace string
+	name     string
+}
```
EItanya, line 44: I see that this is moving the snapshot _back_ into an atespaced key. I specifically undid this in my PR to ensure that later tags could be made global, or potentially move between atespaces without confusion. However, given that we made tags atespaced I think it should be safe to go back to this. The only thing I'll mention is the situation where we want to copy tags to specific atespaces, if the snapshots are atespaced we'd probably need to copy the data

cmd/ateapi/internal/store/ateredis/ateredis_test.go
```diff
@@ -570,17 +570,20 @@ func TestActorSnapshotLifecycle(t *testing.T) {
...
 	}
-	got, location, err := s.GetActorSnapshot(ctx, testAtespace, "snapshot-1")
+	got, err := s.GetActorSnapshot(ctx, testAtespace, "snapshot-1")
 	if err != nil {
 		t.Fatalf("GetActorSnapshot: %v", err)
 	}
-	if !proto.Equal(created, got) || location != "gs://private/snapshot-1" {
-		t.Fatalf("GetActorSnapshot = (%v, %q), want (%v, private location)", got, location, created)
+	// The store round-trips the whole resource, snapshot_uri included: it is
+	// an ordinary field now, not a value the store keeps beside the record.
```
EItanya, line 584: Just FWIW on this, the reason was to hide the storage internals from the snapshot resource itself, and users who have access to it. But in hindsight that might've been unnecessary

EItanya (dismissed):

cmd/ateapi/internal/store/ateredis/ateredis_test.go
juli4n: Ack.

internal/resources/snapshot.go
juli4n: Ack.

dberkov (approved):

Issue #246, opened 2026-06-14
[Bug] Data plane is hard IPv4


ixcans:
## Expected Behavior
IPv6 only cluster support

## Actual Behavior
Envoy DNS resolver pinned V4_ONLY (atenet/.../xds.go:241), 0.0.0.0 binds, and the actor interior net is hard-coded 169.254.17.x/30 forced through .To4() which rejects non-IPv4.

EItanya:
I agree that ipv6 only support is super important, but I also think we have so many open networking questions we're looking to answer that maybe this should be a fast follow from those.
Some examples:
- #126 
- #240

bowei:
I think that would be good to have as a first step is a way to start the Kind setup with IPv6 single-stack and IPv6 dual stack. Otherwise, if we get patches to fix IPv6 support, we would not be able to test the fixes.

BenTheElder:
> I think that would be good to have as a first step is a way to start the Kind setup with IPv6 single-stack and IPv6 dual stack. Otherwise, if we get patches to fix IPv6 support, we would not be able to test the fixes.

We have to be careful about how much actions time we use up on every PR as long as we're operating with the free default resources. It can add up every quickly covering every dimension.

Also FYI single-stack ipv6 kind has some gaps currently, in general we mostly see ipv4 or dual-stack demand and contributions ...

We're not even running ipv6 kind on every k8s PR currently. We run it periodically post-merge.

bowei:
I'd be ok with starting with dual-stack for local testing but not in the CI.

bowei:
Ok, some concrete steps towards chipping away at this:

* PR to add starting up Kind in dual-stack mode, installing Substrate on the cluster
* Fixing the Envoy bind to include IPv6 addresses
* Test (probably draft PR) to show what else is broken right now.

ixcans:
Hello folks, @bowei, @BenTheElder, @EItanya,

Following up after PR #559 merged: the pod-IP:80 → actor-veth:80 DNAT is gone with no fallback, so atunnel's mTLS listener is now the only actor ingress path. The listen address itself is flag-configurable, but the egress redirect in internal/atunnel/original_dst_linux.go is SO_ORIGINAL_DST-only (IPv6 variant still a TODO), and internal/ateomnet's interior actor network is still hard IPv4 after the #548 refactor moved it into a shared package.

[...]

PR #785, opened 2026-08-06
ateapi: bracket IPv6 atelet addresses when dialing

ygao-g:
Part of #246

ateapi built its atelet dial target by concatenating the pod IP with
":8085", which is wrong for IPv6: the target has to be
`[fd00:10:244::7]:8085`, not `fd00:10:244::7:8085`, and `net.Dial` rejects
the unbracketed form with "too many colons in address". `grpc.NewClient`
connects lazily, so this surfaced only as `Unavailable` on every snapshot
and restore RPC, reading as a storage bug rather than a networking one.
Use `net.JoinHostPort`, which brackets the host only when it is an IPv6
literal.

Also lifts the duplicated 8085 literal into `internal/ateletport.Default`,
shared by atelet's `--port` flag and ateapi's dial target; nothing enforced
agreement between the two binaries before.

This is verifiable without an IPv6 cluster — the test asserts the resolved
`ClientConn.Target()` over a fake pod indexer for v4, v6, and `::1`, so it
does not depend on the dual-stack kind work in #246.

- [x] Tests pass

### DIFF
--- a/internal/atelet/port.go
+++ b/internal/atelet/port.go
@@ -1,0 +1,20 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package atelet
+
+const (
+	// DefaultPort is the port atelet's gRPC server listens on.
+	DefaultPort = 8085
+)
--- a/cmd/ateapi/internal/controlapi/dialer.go
+++ b/cmd/ateapi/internal/controlapi/dialer.go
@@ -19,8 +19,11 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"net"
 	"slices"
+	"strconv"
 
+	"github.com/agent-substrate/substrate/internal/atelet"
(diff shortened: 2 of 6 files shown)


ygao-g:
@bowei @haiyanmeng PTAL

internal/ateletport/ateletport.go
```diff
@@ -0,0 +1,27 @@
...
+
+// Package ateletport defines the port atelet's gRPC server listens on, which
+// ateapi and atelet have to agree on.
+package ateletport
+
+const (
+	// Default is the port atelet's gRPC server listens on: it is the default of
+	// atelet's --port flag (cmd/atelet/main.go) and the port ateapi dials when
+	// it connects to an atelet
+	// (cmd/ateapi/internal/controlapi/dialer.go). It is deliberately untyped so
```
bowei: Please go through your code comments and remove the agent language (extra verbose comments)

internal/atelet/port.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2026 Google LLC
```
bowei, line 1: put this in internal/atelet

doc.go

```
// Package atelet contains shared constants and definitions that are used across
// Substrate components related to atelet.
package atelet
```

Having a package just for port is too narrow.

bowei: @BenTheElder -- opinions?

ygao-g: Done — moved to `internal/atelet` with `doc.go` as suggested; the constant now lives in `port.go`.

One adjustment: renamed `Default` → `DefaultPort`, since `atelet.Default` doesn't say much once the package isn't port-specific. Call sites read `atelet.DefaultPort`.

ygao-g:
Rebased onto main (`9d519ca5`). The earlier red was the `spec.pauseImage` fixture regression from #848 + #816, fixed upstream by #857 — picked up in the rebase, and `go test ./...` is clean locally.

CI is sitting at `action_required`; could someone approve the workflow run?

bowei (approved):

PR #552, opened 2026-07-27
Enable CSI based external volumes for agents

hajiler:
#232

Integrate the with the volume interface created in https://github.com/agent-substrate/substrate/pull/405, to enable CSI managed external volumes throughout the agent lifecycle. 

  CSI driver registration in Substrate is managed dynamically via the CSIDriverConfig CRD, which allows the system to discover and integrate with various CSI drivers on-demand. To support central
  control-plane operations (such as volume provisioning), the driver must expose its Controller service over a network port (gRPC over TCP or DNS) which ateapi dynamically resolves from the CRD. For node-
  level operations (such as volume mounting), the atelet agent on each worker node bypasses the network and connects directly to the driver's Node service using a Unix domain socket mounted from the
  Kubelet plugins directory.

Support for secure connection with the controller will be added in a follow-up PR.

- [ X] Tests pass
- [ ] Appropriate changes to documentation are included in the PR

PR #786, opened 2026-08-06
Add check to skip ExternalVolume e2e test in multi-node environment

msau42:
The mock volume plugin that's used for testing only supports single-node

Part of #232

- [x] Tests pass


msau42:
This is still valid with https://github.com/agent-substrate/substrate/pull/552, as that PR changes the e2e test to use CSI hostpath driver, which still has the same single-node restriction. If we add a test with an nfs csi driver then we could remove this restriction.

msau42:
Thinking about it more, we should look at making the e2e with the oss nfs csi driver so that it could work in all clusters

PR #788, opened 2026-08-07
Allow force deleting an actor from any state.

### LINKED ISSUE #643: Recovery from failed resume, crash scenarios

Sneha-at:
Fixes #643 
                                                                                                                                                                                                                                                                                                                                                 
This feature introduces the ability to delete Substrate actors from any lifecycle state (e.g., RUNNING, PAUSED, PENDING), rather than requiring them to be hibernated into SUSPENDED (or CRASHED) beforehand.  This is crucial for cleaning up stuck or failed actors.
                                                                                                                                                                                                                                                                                                                                                                     
  #### 1. Control Plane & gRPC API (pkg/proto/ateapipb, ateapi)                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                     
  • DeleteActorRequest.any_state: Added a new boolean field any_state to DeleteActorRequest.                                                                                                                                                                                                                                                                         
      • When false (default): enforces the existing behavior where only actors in ACTOR_STATE_SUSPENDED or ACTOR_STATE_CRASHED (or already ACTOR_STATE_DELETING) can be deleted.                                                                                                                                                                                     
      • When true: allows deleting an actor in any state (e.g., RUNNING, PAUSED).                                                                                                                                                                                                                                                                                    
  • Orchestrated Cleanup Workflow (workflow_delete.go):                                                                                                                                                                                                                                                                                                              
         1. Transitions actor state to ACTOR_STATE_DELETING.                                                                                                                                                                                                                                                                                                            
         2. Calls atelet.Terminate to stop live workloads on the node.                                                                                                                                                                                                                                                                                                  
         3. Detaches external volumes from the worker node (with fallback logic using actor status if the ActorTemplate was deleted).                                                                                                                                                                                                                                   
         4. Releases the assigned physical worker Pod back to the WorkerPool.                                                                                                                                                                                                                                                                                           
         5. Deletes external storage volumes through the storage plugins.                                                                                                                                                                                                                                                                                               
         6. Finalizes removal of the actor from the persistence store.                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                     
  #### 2. Node Herder Agent (atelet)                                                                                                                                                                                                                                                                                                                                 
  • Terminate RPC (main.go): Added a new Terminate method to AteomHerder:                                                                                                                                                                                                                                                                                            
      • Calls ateom.TerminateWorkload on the target worker pod.                                                                                                                                                                                                                                                                                                      
      • Unmounts external volumes mounted for the actor on the node.                                                                                                                                                                                                                                                                                                 
      • Cleans up and resets actor runtime host directories (OCI bundles, checkpoints, pid files).                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                     
 #### 3. In-Pod Sandboxes (ateom-gvisor & ateom-microvm)                                                                                                                                                                                                                                                                                                            
  • TerminateWorkload RPC:                                                                                                                                                                                                                                                                                                                                           
      • gVisor (main.go): Stops and deletes active runsc containers, unmounts bundle rootfs overlays, and tears down the actor's interior network namespace.                                                                                                                                                                                                         
      • Micro-VM (checkpoint.go): Shuts down the Cloud Hypervisor VMM, unmounts bundle rootfs overlays, and cleans up the network namespace.                                                                                                                                                                                                                         
      • Emits an "Actor terminated" lifecycle log and clears the active actor attribution so the worker can host new workloads.                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                     
  #### 4. CLI (kubectl-ate)                                                                                                                                                                                                                                                                                                                                          
  • --any-state Flag (delete_actor.go):                                                                                                                                                                                                                                                                                                                              
    kubectl-ate delete actor <actor-name> -a <atespace> --any-state                                                                                                             

  ──────                                                                                                                                                                                                                                                                                                                                                             
  ## Verification 
  [x] Changes tested in local with 
 ```
 • go build ./...: Passed                                                                                                                                                                                                                                                                                                                                       
    go test ./...: Passed (all unit and functional tests pass)                                                                                                                                                                                                                                                                                                   
    make lint: Passed (no lint errors) 
```
Detailed manual tests                                                                                                                                                                                                                                                                                                                                           
 ```                                                                                                                                                                                                                                                                                                                                                
  ### Scenario 1: Force Delete a Running Actor with --any-state                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                     
  1. Create the actor:                                                                                                                                                                                                                                                                                                                                               
    kubectl ate create actor manual-test-1 --template=ate-demo-counter-microvm/counter-microvm -a demo                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                     
  2. Resume the actor:                                                                                                                                                                                                                                                                                                                                               
    kubectl ate resume actor manual-test-1 -a demo                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                     
  3. Force delete the running actor:                                                                                                                                                                                                                                                                                                                                 
    kubectl ate delete actor manual-test-1 -a demo --any-state                                                                                                                                                                                                                                                                                                       
    actor "manual-test-1" deleted                                                                                                                                                                                                                                                                                                                                    
```                                                                                                                                                                                                                                                                                                                                                                   
  ──────                                                                                                                                                                                                                                                                                                                                                             
### Scenario 2: Standard Delete a Suspended Actor                                                                                                                                                                                                                                                                                                                  
```                                                                                                                                                                                                                                                                                                                                  
  1. Create the actor:                                                                                                                                                                                                                                                                                                                                               
    kubectl ate create actor manual-test-2 --template=ate-demo-counter-microvm/counter-microvm -a demo                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                     
  2. Resume the actor:                                                                                                                                                                                                                                                                                                                                               
    kubectl ate resume actor manual-test-2 -a demo                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                     
  3. Suspend the actor:                                                                                                                                                                                                                                                                                                                                              
    kubectl ate suspend actor manual-test-2 -a demo                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                     
  4. Standard delete the suspended actor:                                                                                                                                                                                                                                                                                                                            
    kubectl ate delete actor manual-test-2 -a demo                                                                                                                                                                                                                                                                                                                   
    actor "manual-test-2" deleted                                                                                                                                                                                                                                                                                                                                    
```
  ──────                                                                                                                                                                                                                                                                                                                                                             
 ### Unit & E2E Tests                                                                                                                                                                                                                                                                                                                                               
    # Unit & Functional Tests                                                                                                                                                                                                                                                                                                                                        
    go test ./cmd/ateapi/internal/controlapi -run "TestDeleteActorWorkflow|TestEnsureMarkedDeleting"                                                                                                                                                                                                                                                                 
    # E2E Tests                                                                                                                                                                                                                                                                                                                                                      
    E2E_TEMPLATE_NAMESPACE=ate-demo-counter-microvm E2E_TEMPLATE_NAME=counter-microvm ./hack/run-e2e-kind.sh ./internal/e2e/suites/demo -run TestActorLifecycle                                                                                                                                                                                                      
    ./hack/run-e2e-kind.sh ./internal/e2e/suites/demo -run TestForceDeleteActorWithExternalVolume 
```
- [ x] Appropriate changes to documentation are included in the PR

### DIFF
--- a/cmd/ateapi/internal/controlapi/volumes.go
+++ b/cmd/ateapi/internal/controlapi/volumes.go
@@ -150,6 +150,10 @@ func deleteActorVolumes(ctx context.Context, registry VolumePluginRegistry, acto
 			continue
 		}
 		if err := plugin.DeleteVolume(ctx, volID); err != nil {
+			if status.Code(err) == codes.NotFound {
+				slog.WarnContext(ctx, "Volume not found during delete, assuming already deleted", slog.String("volume_id", volID))
+				continue
+			}
 			errs = append(errs, fmt.Errorf("failed to delete volume %q: %w", volID, err))
 		}
 	}
@@ -211,18 +215,38 @@ func detachActorVolumes(ctx context.Context, st detachActorVolumesStore, registr
 	}
 
 	ref := &ateapipb.ObjectRef{Atespace: actor.GetMetadata().GetAtespace(), Name: actor.GetMetadata().GetName()}
-	for _, vol := range getMountedActorVolumes(ctx, ref, actor.GetStatus().GetActorVolumes(), template) {
+	// If the template is available, only detach volumes that are actively mounted
+	// in the template's containers. If the template is missing/deleted, fall back to
+	// attempting detachment for all external volumes recorded on the actor so we do
+	// not orphan attached disks on the worker node.
+	volumesToDetach := actor.GetStatus().GetActorVolumes()
+	if template != nil {
+		volumesToDetach = getMountedActorVolumes(ctx, ref, actor.GetStatus().GetActorVolumes(), template)
+	}
+	// Collect errors for all volumes to detach, but continue processing so we attempt to detach all volumes.
+	var errs []error
+	for _, vol := range volumesToDetach {
+		// StorageVolumeId is only populated once the volume is provisioned.
+		// Skip volumes that were never created (e.g. failed during PENDING state).
+		if vol.GetStorageVolumeId() == "" {
+			slog.WarnContext(ctx, "Volume has no storage volume ID, skipping detach", slog.String("volume_name", vol.GetVolumeName()), slog.String("actor_id", actor.GetMetadata().GetName()))
+			continue
+		}
 		slog.InfoContext(ctx, "Detaching volume from node", slog.String("volume_id", vol.GetStorageVolumeId()), slog.String("node", node))
 		plugin, err := registry.GetPlugin(ctx, vol.GetVolumeType())
 		if err != nil {
-			return fmt.Errorf("failed to get volume plugin for %q: %w", vol.GetVolumeType(), err)
+			errs = append(errs, fmt.Errorf("failed to get volume plugin for %q: %w", vol.GetVolumeType(), err))
+			continue
 		}
-		err = plugin.DetachVolume(ctx, vol.GetStorageVolumeId(), node)
-		if err != nil {
-			return fmt.Errorf("failed to detach volume %q from node %q: %w", vol.GetStorageVolumeId(), node, err)
+		if err := plugin.DetachVolume(ctx, vol.GetStorageVolumeId(), node); err != nil {
+			if status.Code(err) == codes.NotFound {
+				slog.WarnContext(ctx, "Volume not found during detach, assuming already detached", slog.String("volume_id", vol.GetStorageVolumeId()), slog.String("node", node))
+				continue
+			}
+			errs = append(errs, fmt.Errorf("failed to detach volume %q from node %q: %w", vol.GetStorageVolumeId(), node, err))
 		}
 	}
-	return nil
+	return errors.Join(errs...)
 }
 
 // detachActorVolumesStore enumerates the subset of store methods needed to
--- a/cmd/ateapi/internal/controlapi/workflow_delete.go
+++ b/cmd/ateapi/internal/controlapi/workflow_delete.go
@@ -18,16 +18,20 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
+	"github.com/agent-substrate/substrate/internal/proto/ateletpb"
 	"github.com/agent-substrate/substrate/internal/resources"
+	atev1alpha1 "github.com/agent-substrate/substrate/pkg/api/v1alpha1"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // DeleteActor executes the workflow to delete an actor. Idempotent.
-func (w *ActorWorkflow) DeleteActor(ctx context.Context, actorRef resources.ActorRef) (*ateapipb.Actor, error) {
+func (w *ActorWorkflow) DeleteActor(ctx context.Context, actorRef resources.ActorRef, anyState bool) (*ateapipb.Actor, error) {
 	ctx, lock, err := w.acquireActorLock(ctx, actorRef)
 	if err != nil {
 		return nil, err
@@ -38,11 +42,53 @@ func (w *ActorWorkflow) DeleteActor(ctx context.Context, actorRef resources.Acto
 	if err != nil {
 		return nil, err
 	}
-	if actor, err = w.ensureMarkedDeleting(ctx, actorRef, actor); err != nil {
+
+	if actor, err = w.ensureMarkedDeleting(ctx, actorRef, actor, anyState); err != nil {
 		return nil, err
 	}
+
+	// DeleteActor will attempt best-effort cleanup across all steps, collecting any errors.
+	// If any step fails, errors are returned so the caller can retry, and the actor record
+	// is retained in the store in the DELETING state.
+	// TODO: Ensure GC collects all the remaining resources if the cleanup fails.
+	var errs []error
+	actorTemplate := (*atev1alpha1.ActorTemplate)(nil)
+	if actor.GetActorTemplateNamespace() != "" && actor.GetActorTemplateName() != "" {
+		tmpl, err := w.actorTemplateLister.ActorTemplates(actor.GetActorTemplateNamespace()).Get(actor.GetActorTemplateName())
+		if err != nil && !k8serrors.IsNotFound(err) {
+			errs = append(errs, fmt.Errorf("while fetching actor template: %w", err))
+		}
+		actorTemplate = tmpl
+	}
+
+	var atletTerminatedErr, volumesDetachedErr error
+	if err := w.ensureAteletTerminated(ctx, actorRef, actor, actorTemplate); err != nil {
+		atletTerminatedErr = fmt.Errorf("while terminating atelet: %w", err)
+		errs = append(errs, atletTerminatedErr)
+	}
+	if err := w.ensureVolumesDetachedForDelete(ctx, actor, actorTemplate); err != nil {
+		volumesDetachedErr = fmt.Errorf("while detaching volumes: %w", err)
+		errs = append(errs, volumesDetachedErr)
+	}
+
+	// Release worker if atelet termination and volume detachment did not fail.
+	if atletTerminatedErr == nil && volumesDetachedErr == nil {
+		if actor, err = w.ensureWorkerReleased(ctx, actorRef, actor); err != nil {
+			errs = append(errs, fmt.Errorf("while releasing worker: %w", err))
+		}
+	} else {
+		slog.WarnContext(ctx, "skipping releasing worker due to atelet termination or volume detachment failure",
+			slog.Any("actor", actorRef),
+			slog.Any("atletTerminatedErr", atletTerminatedErr),
+			slog.Any("volumesDetachedErr", volumesDetachedErr))
+	}
+
 	if err := w.ensureVolumesDeleted(ctx, actor); err != nil {
-		return nil, err
+		errs = append(errs, fmt.Errorf("while deleting volumes: %w", err))
+	}
+
+	if len(errs) > 0 {
+		return nil, errors.Join(errs...)
 	}
 	return w.finalizeDeleted(ctx, actorRef)
 }
@@ -62,20 +108,177 @@ func (w *ActorWorkflow) loadActorForDelete(ctx context.Context, actorRef resourc
 	return actor, nil
 }
 
+// ensureAteletTerminated calls atelet to terminate the workload.
+func (w *ActorWorkflow) ensureAteletTerminated(ctx context.Context, actorRef resources.ActorRef, actor *ateapipb.Actor, actorTemplate *atev1alpha1.ActorTemplate) (err error) {
+	ctx, done := stepSpan(ctx, "CallAteletTerminate")
+	defer func() { err = done(err) }()
+
+	assignment := actor.GetStatus().GetWorkerAssignment()
+	if assignment == nil {
+		slog.InfoContext(ctx, "actor has no worker assignment, skipping atlet terminate request", slog.Any("actor", actorRef))
+		return nil
+	}
+
+	if workerName := assignment.GetWorker().GetName(); workerName != "" {
+		worker, err := w.store.GetWorker(ctx, workerName)
+		if err != nil {
+			if errors.Is(err, store.ErrNotFound) {
+				slog.InfoContext(ctx, "worker not found in store, skipping atelet terminate request", slog.String("worker", workerName), slog.Any("actor", actorRef))
+				return nil
+			}
+			return fmt.Errorf("while checking worker assignment: %w", err)
+		}
+		wass := worker.GetStatus().GetAssignment()
+		if wass == nil || wass.GetActorUid() != actor.GetMetadata().GetUid() {
+			slog.InfoContext(ctx, "worker is no longer assigned to this actor, skipping atelet terminate request",
+				slog.String("worker", workerName),
+				slog.Any("actor", actorRef))
+			return nil
+		}
+	}
+
+	workerPodNs := assignment.GetWorkerNamespace()
+	workerPodName := assignment.GetWorkerPod()
+
+	conn, err := w.dialer.DialForWorker(workerPodNs, workerPodName)
+	if err != nil {
+		if errors.Is(err, ErrWorkerPodNotFound) {
+			slog.InfoContext(ctx, "worker pod not found, treating as terminated", slog.String("workerNamespace", workerPodNs), slog.String("workerPod", workerPodName))
+			return nil
+		}
+		return fmt.Errorf("while connecting to worker pod %s/%s: %w", workerPodNs, workerPodName, err)
+	}
+
+	client := ateletpb.NewAteomHerderClient(conn)
+
+	var workloadSpec *ateletpb.WorkloadSpec
+	if actorTemplate != nil {
+		spec, err := workloadSpecFromActorTemplate(actorTemplate, actor)
+		if err != nil {
+			return err
+		}
+		workloadSpec = spec
+	} else {
+		// When the template is missing/deleted, build a fallback workload spec with
+		// all external volumes recorded on the actor so atelet can unmount them on the node.
+		slog.WarnContext(ctx, "actor template not found, constructing fallback workload spec for atelet terminate",
+			slog.String("actor", actorRef.Name),
+			slog.String("templateNamespace", actor.GetActorTemplateNamespace()),
+			slog.String("templateName", actor.GetActorTemplateName()))
+		workloadSpec = &ateletpb.WorkloadSpec{}
+		for _, vol := range actor.GetStatus().GetActorVolumes() {
+			// StorageVolumeId is only populated once the volume is provisioned.
+			// Skip volumes that were never created (e.g. failed during PENDING state).
+			if vol.GetStorageVolumeId() != "" {
+				workloadSpec.Volumes = append(workloadSpec.Volumes, &ateletpb.Volume{
+					Name: vol.GetVolumeName(),
+					Source: &ateletpb.Volume_External{
+						External: &ateletpb.ExternalVolumeSource{
+							StorageVolumeId: vol.GetStorageVolumeId(),
+							VolumeType:      vol.GetVolumeType(),
+							VolumeContext:   vol.GetVolumeContext(),
+						},
+					},
+				})
+			}
+		}
+	}
+
+	req := &ateletpb.TerminateRequest{
+		TargetAteomUid:         assignment.GetWorkerPodUid(),
+		Atespace:               actor.GetMetadata().GetAtespace(),
+		ActorName:              actor.GetMetadata().GetName(),
+		ActorUid:               actor.GetMetadata().GetUid(),
+		ActorTemplateNamespace: actor.GetActorTemplateNamespace(),
+		ActorTemplateName:      actor.GetActorTemplateName(),
+		Spec:                   workloadSpec,
+	}
+
+	if _, err := client.Terminate(ctx, req); err != nil {
+		if status.Code(err) == codes.NotFound {
+			slog.InfoContext(ctx, "workload already terminated on atelet", slog.Any("actor", actorRef))
+			return nil
+		}
+		return fmt.Errorf("while terminating actor on atelet: %w", err)
+	}
+
+	return nil
+}
+
+// ensureVolumesDetachedForDelete detaches external volumes.
+func (w *ActorWorkflow) ensureVolumesDetachedForDelete(ctx context.Context, actor *ateapipb.Actor, actorTemplate *atev1alpha1.ActorTemplate) (err error) {
+	ctx, done := stepSpan(ctx, "DetachVolumesForDelete")
+	defer func() { err = done(err) }()
+
+	return detachActorVolumes(ctx, w.store, w.pluginRegistry, actor, actorTemplate, "delete")
+}
+
+// ensureWorkerReleased releases the worker assigned to the actor.
+func (w *ActorWorkflow) ensureWorkerReleased(ctx context.Context, actorRef resources.ActorRef, actor *ateapipb.Actor) (updated *ateapipb.Actor, err error) {
+	ctx, done := stepSpan(ctx, "ReleaseWorker")
+	defer func() { err = done(err) }()
+
+	if actor.GetStatus().GetWorkerAssignment() == nil {
+		markSkipped(ctx, "worker already released")
+		return actor, nil
+	}
+
+	latestActor, err := w.store.GetActor(ctx, actorRef)
+	if err != nil {
+		return nil, err
+	}
+
+	if latestActor.GetStatus().GetWorkerAssignment() != nil {
+		if _, err := releaseWorker(ctx, w.store, latestActor); err != nil {
+			return nil, err
+		}
+
+		latestActor, err = w.store.GetActor(ctx, actorRef)
+		if err != nil {
+			return nil, err
+		}
+
+		updatedActor, err := w.store.UpdateActor(ctx, actorRef, store.PreconditionFrom(latestActor), func(dbActor *ateapipb.Actor) error {
+			if dbActor.Status != nil {
+				dbActor.Status.LocalSnapshotInfo = nil
+				dbActor.Status.WorkerAssignment = nil
+			}
+			return nil
+		})
+		if err != nil {
+			if errors.Is(err, store.ErrVersionConflict) {
+				return nil, status.Error(codes.Aborted, "concurrent update conflict, please retry")
+			}
+			return nil, err
+		}
+		latestActor = updatedActor
+	}
+	return latestActor, nil
+}
+
 // ensureMarkedDeleting transitions the actor and its volumes to DELETING and
 // persists the change, returning the stored copy. Skips when a previous
 // attempt already marked the actor.
-func (w *ActorWorkflow) ensureMarkedDeleting(ctx context.Context, actorRef resources.ActorRef, actor *ateapipb.Actor) (_ *ateapipb.Actor, err error) {
+func (w *ActorWorkflow) ensureMarkedDeleting(ctx context.Context, actorRef resources.ActorRef, actor *ateapipb.Actor, anyState bool) (updated *ateapipb.Actor, err error) {
 	ctx, done := stepSpan(ctx, "MarkDeleting")
 	defer func() { err = done(err) }()
 
-	if actor.GetStatus().GetState() == ateapipb.ActorState_ACTOR_STATE_DELETING {
+	st := actor.GetStatus().GetState()
+	if st == ateapipb.ActorState_ACTOR_STATE_DELETING {
 		markSkipped(ctx, "actor already DELETING")
 		return actor, nil
 	}
-	if actor.GetStatus().GetState() != ateapipb.ActorState_ACTOR_STATE_SUSPENDED &&
-		actor.GetStatus().GetState() != ateapipb.ActorState_ACTOR_STATE_CRASHED {
-		return nil, status.Errorf(codes.FailedPrecondition, "Actor %s is not in a deletable state (state: %v)", actorRef, actor.GetStatus().GetState())
+	shouldDelete := false
+	switch st {
+	case ateapipb.ActorState_ACTOR_STATE_SUSPENDED,
+		ateapipb.ActorState_ACTOR_STATE_CRASHED:
+		shouldDelete = true
+	default:
+		// This allows deletion for any state
+		shouldDelete = anyState
+	}
+	if !shouldDelete {
+		return nil, status.Errorf(codes.FailedPrecondition, "Actor %s is not in a deletable state (state: %v)", actorRef, st)
 	}
 
 	storedActor, err := w.store.UpdateActor(ctx, actorRef, store.PreconditionFrom(actor), func(toUpdate *ateapipb.Actor) error {
@@ -95,10 +298,15 @@ func (w *ActorWorkflow) ensureMarkedDeleting(ctx context.Context, actorRef resou
 }
 
 // ensureVolumesDeleted removes the actor's external volumes. Volume deletion
-// is idempotent, so a re-entered workflow can safely run it again.
+// is idempotent, so a re-entered workflow can safely run it again. It assumes a no-op success if volume is not found
 func (w *ActorWorkflow) ensureVolumesDeleted(ctx context.Context, actor *ateapipb.Actor) (err error) {
 	ctx, done := stepSpan(ctx, "DeleteVolumes")
 	defer func() { err = done(err) }()
+
+	st := actor.GetStatus().GetState()
+	if st != ateapipb.ActorState_ACTOR_STATE_DELETING {
+		return status.Errorf(codes.FailedPrecondition, "DeleteVolumes prerequisite not met for Actor: %s (got: %v, want %s)", actor.GetMetadata().GetName(), st, ateapipb.ActorState_ACTOR_STATE_DELETING)
+	}
 
 	if err := deleteActorVolumes(ctx, w.pluginRegistry, actor.GetMetadata().GetUid(), actor.GetStatus().GetActorVolumes()); err != nil {
 		return status.Errorf(codes.Internal, "while deleting actor volumes: %v", err)
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -1195,6 +1195,65 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 	return &ateletpb.RestoreResponse{}, nil
 }
 
+// Terminate terminates any running workload on ateom, unmounts external volumes,
+// and resets actor directories on the node.
+func (s *AteomHerder) Terminate(ctx context.Context, req *ateletpb.TerminateRequest) (*ateletpb.TerminateResponse, error) {
+	if err := validateTerminateRequest(req); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
+	}
+
+	actorRef := resources.ActorRef{Atespace: req.GetAtespace(), Name: req.GetActorName()}
+	actorUID := req.GetActorUid()
+
+	var assetPaths map[string]string
+	sandboxRec, err := readSandboxRecord(actorUID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read sandbox record during terminate (actor: %s, actorUID: %s): %w", actorRef, actorUID, err)
+	}
+	paths, err := s.ensureSandboxAssets(ctx, sandboxRec)
+	if err != nil {
+		return nil, fmt.Errorf("failed to ensure sandbox assets during terminate (actor: %s, actorUID: %s): %w", actorRef, actorUID, err)
+	}
+	assetPaths = paths
+
+	client, err := s.dialAteom(ctx, req.GetTargetAteomUid())
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial ateom for terminate (actor: %s, actorUID: %s): %w", actorRef, actorUID, err)
+	}
+
+	spec, err := buildAteomWorkloadSpec(req.GetSpec())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid workload spec: %v", err)
+	}
+	if _, err := client.TerminateWorkload(ctx, &ateompb.TerminateWorkloadRequest{
+		Atespace:               req.GetAtespace(),
+		ActorName:              req.GetActorName(),
+		ActorUid:               req.GetActorUid(),
+		ActorTemplateNamespace: req.GetActorTemplateNamespace(),
+		ActorTemplateName:      req.GetActorTemplateName(),
+		RunscPath:              runscPathFor(assetPaths),
+		Spec:                   spec,
+	}); err != nil {
+		if status.Code(err) == codes.NotFound {
+			slog.InfoContext(ctx, "workload not found on ateom during terminate", slog.Any("actor", actorRef), slog.String("actorUID", actorUID))
+		} else {
+			return nil, fmt.Errorf("failed calling ateom.TerminateWorkload (actor: %s, actorUID: %s): %w", actorRef, actorUID, err)
+		}
+	}
+
+	// Unmount external volumes
+	if err := s.unmountExternalVolumes(ctx, actorUID, req.GetSpec().GetVolumes()); err != nil {
+		return nil, fmt.Errorf("failed to unmount external volumes during terminate (actor: %s, actorUID: %s): %w", actorRef, actorUID, err)
+	}
+
+	// Reset actor directories on the node
+	if err := resetActorDirs(actorUID); err != nil {
+		return nil, fmt.Errorf("failed to reset actor directories during terminate (actor: %s, actorUID: %s): %w", actorRef, actorUID, err)
+	}
+
+	return &ateletpb.TerminateResponse{}, nil
+}
+
 func (s *AteomHerder) copyLocalCheckpoint(ctx context.Context, snapshotName string, srcDir, dstDir string, files []string) error {
 	for _, fileName := range files {
 		if ctx.Err() != nil {
@@ -1862,6 +1921,30 @@ func validateRestoreRequest(req *ateletpb.RestoreRequest) error {
 		return fmt.Errorf("golden_snapshot_uri is only valid with snapshot scope %s", ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA_ON_GOLDEN)
 	}
 	return nil
+}
+
+func validateTerminateRequest(req *ateletpb.TerminateRequest) error {
+	var errs field.ErrorList
+	errs = append(errs, resources.ValidateResourceName(req.GetAtespace(), field.NewPath("atespace"))...)
+	errs = append(errs, resources.ValidateResourceName(req.GetActorName(), field.NewPath("actor_name"))...)
+	errs = append(errs, resources.ValidateResourceName(req.GetActorUid(), field.NewPath("actor_uid"))...)
+	for _, msg := range content.IsDNS1123Label(req.GetActorTemplateNamespace()) {
+		errs = append(errs, field.Invalid(field.NewPath("actor_template_namespace"), req.GetActorTemplateNamespace(), msg))
+	}
+	for _, msg := range content.IsDNS1123Subdomain(req.GetActorTemplateName()) {
+		errs = append(errs, field.Invalid(field.NewPath("actor_template_name"), req.GetActorTemplateName(), msg))
+	}
+	if len(errs) > 0 {
+		return errs.ToAggregate()
+	}
+	if err := resources.ValidateAteomUID(req.GetTargetAteomUid()); err != nil {
+		return err
+	}
+	names := make([]string, 0, len(req.GetSpec().GetContainers()))
+	for _, ctr := range req.GetSpec().GetContainers() {
+		names = append(names, ctr.GetName())
+	}
+	return resources.ValidateContainerNames(names)
 }
 
 func validateSnapshotScope(scope ateletpb.SnapshotScope) error {
--- a/cmd/ateom-gvisor/main.go
+++ b/cmd/ateom-gvisor/main.go
@@ -785,29 +785,13 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 	// reporting its usage is then the honest answer.
 	s.activeActor.Store(nil)
 
-	// After checkpointing the sandbox root, runsc may no longer have a usable
-	// control server for state/delete calls. Keep this as best-effort cleanup:
-	// atelet resets the actor runsc, bundle, pidfile, and checkpoint
-	// directories after uploading the snapshot.
-	if err := rcmd.cleanupContainersAfterCheckpoint(ctx, req.GetSpec().GetContainers()); err != nil {
-		slog.WarnContext(ctx, "Failed to clean up runsc containers after checkpoint",
-			"actor", attribution.Ref,
-			"actorUID", req.GetActorUid(),
-			"err", err)
-	}
-
-	// Detach the overlay rootfs mounts before atelet wipes the bundle dirs
-	// (deleting a bundle out from under a live mount in this namespace would
-	// leave the mount orphaned until the pod restarts). Best-effort, same as
-	// the container cleanup above.
-	if err := imagecache.UnmountAllUnder(ateompath.OCIBundleDir(req.GetActorUid())); err != nil {
-		slog.WarnContext(ctx, "Failed to unmount bundle rootfs overlays after checkpoint",
-			"actorUID", req.GetActorUid(),
-			"err", err)
-	}
-
-	if err := ateomnet.CleanupActorNetwork(ctx, s.interiorNetNS); err != nil {
-		slog.WarnContext(ctx, "Failed to clean up actor network after checkpoint", slog.Any("err", err))
+	// Cleanup the containers after checkpointing.
+	// This is best-effort cleanup for actor containers that may have been left behind after checkpointing.
+	if err := s.terminateWorkload(ctx, attribution.Ref, attribution.UID, req.GetRunscPath(), req.GetSpec().GetContainers()); err != nil {
+		slog.WarnContext(ctx, "failed to terminate workload after checkpoint",
+			slog.String("actor", attribution.Ref.String()),
+			slog.String("actorUID", attribution.UID),
+			slog.Any("err", err))
 	}
 
 	// Report exactly the files runsc wrote so atelet ships precisely this set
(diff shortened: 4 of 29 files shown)

### CODE REVIEW

File: cmd/ateom-microvm/checkpoint.go [hunk 1]
```diff
@@ -154,14 +154,8 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 	// Tear down: the actor returns to "available". Best-effort; the snapshot is
 	// already on disk for atelet to ship.
 	tTeardown := time.Now()
-	s.teardownActor(ctx, actorUID, ra, client)
+	s.terminateWorkload(ctx, actorUID)
```
sandeep-gh-acc: we are not doing error checks here? the method returns err, what if termination fails

cmd/ateapi/internal/controlapi/delete_actor.go
```diff
@@ -45,7 +45,7 @@ func (s *Service) DeleteActor(ctx context.Context, req *ateapipb.DeleteActorRequ
 	actorRef := resources.ActorRefFromObjectRef(req.GetActor())
 	setSpanActorRefAttributes(ctx, actorRef)
 
-	deleted, err = s.actorWorkflow.DeleteActor(ctx, req.GetActor().GetAtespace(), req.GetActor().GetName())
+	deleted, err = s.actorWorkflow.DeleteActor(ctx, req.GetActor().GetAtespace(), req.GetActor().GetName(), req.GetForce())
```
sandeep-gh-acc: Should we squash this to one commit

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -59,11 +67,71 @@ func (s *LoadActorForDeleteStep) Execute(ctx context.Context, input *DeleteInput
 		return fmt.Errorf("while fetching actor: %w", err)
 	}
 	state.Actor = actor
+
+	if s.actorTemplateLister != nil && actor.GetActorTemplateNamespace() != "" && actor.GetActorTemplateName() != "" {
+		tmpl, err := s.actorTemplateLister.ActorTemplates(actor.GetActorTemplateNamespace()).Get(actor.GetActorTemplateName())
+		if err != nil && !k8serrors.IsNotFound(err) {
+			return fmt.Errorf("while fetching actor template: %w", err)
```
sandeep-gh-acc: Do we want to log info, warn in case actor template was not found?

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -59,11 +67,71 @@ func (s *LoadActorForDeleteStep) Execute(ctx context.Context, input *DeleteInput
...
+func (s *MarkTerminatingStep) CheckPrerequisite(ctx context.Context, input *DeleteInput, state *DeleteState) error {
+	if input.Force {
+		switch state.Actor.GetStatus() {
+		case ateapipb.Actor_STATUS_RUNNING,
+			ateapipb.Actor_STATUS_RESUMING,
+			ateapipb.Actor_STATUS_SUSPENDING,
+			ateapipb.Actor_STATUS_PAUSING,
+			ateapipb.Actor_STATUS_PAUSED,
+			ateapipb.Actor_STATUS_CRASHED,
+			ateapipb.Actor_STATUS_TERMINATING,
```
sandeep-gh-acc: Not sure if its elsewhere in the code, do we log info on what the last state of the actor was it will help debug complex lifecycle issues if any

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -59,11 +67,71 @@ func (s *LoadActorForDeleteStep) Execute(ctx context.Context, input *DeleteInput
...
+			ateapipb.Actor_STATUS_DELETING,
+			ateapipb.Actor_STATUS_SUSPENDED:
+			return nil
+		default:
+			return status.Errorf(codes.FailedPrecondition, "Actor %s is not in a deletable status (status: %v)", input.ActorRef, state.Actor.GetStatus())
```
sandeep-gh-acc: In case on non force scenario should this fail if actor is already in TERMINATING step. It could be a no-op with a log?

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -59,11 +67,71 @@ func (s *LoadActorForDeleteStep) Execute(ctx context.Context, input *DeleteInput
...
+		ateapipb.Actor_STATUS_DELETING:
+		return nil
+	default:
+		return status.Errorf(codes.FailedPrecondition, "Actor %s is not in a deletable status (status: %v)", input.ActorRef, state.Actor.GetStatus())
+	}
+}
+
+func (s *MarkTerminatingStep) Execute(ctx context.Context, input *DeleteInput, state *DeleteState) error {
+	state.Actor.Status = ateapipb.Actor_STATUS_TERMINATING
+	updated, err := s.store.UpdateActor(ctx, state.Actor, state.Actor.GetMetadata().GetVersion())
```
sandeep-gh-acc: Not sure if this is pattern else where why do we pass version explicitly if we are passing the Actor object itself in the function here

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -59,11 +67,71 @@ func (s *LoadActorForDeleteStep) Execute(ctx context.Context, input *DeleteInput
...
+		tmpl, err := s.actorTemplateLister.ActorTemplates(actor.GetActorTemplateNamespace()).Get(actor.GetActorTemplateName())
+		if err != nil && !k8serrors.IsNotFound(err) {
+			return fmt.Errorf("while fetching actor template: %w", err)
+		}
+		state.ActorTemplate = tmpl
+	}
 	return nil
 }
 
 func (s *LoadActorForDeleteStep) RetryBackoff() *wait.Backoff { return nil }
```
sandeep-gh-acc: Do we need a backoff here, why is this nil.

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -98,6 +165,151 @@ func (s *MarkDeletingStep) Execute(ctx context.Context, input *DeleteInput, stat
...
 
+type CallAteletTerminateStep struct {
+	store  store.Interface
+	dialer *AteletDialer
+}
+
+func (s *CallAteletTerminateStep) Name() string { return "CallAteletTerminate" }
+
+func (s *CallAteletTerminateStep) IsComplete(ctx context.Context, input *DeleteInput, state *DeleteState) (bool, error) {
+	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_DELETING, nil
```
sandeep-gh-acc: Is there a code path that needs to be still added here, we are not passing any non nil value of err ever, then we should never return err if it can never have non nil error

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -98,6 +165,151 @@ func (s *MarkDeletingStep) Execute(ctx context.Context, input *DeleteInput, stat
...
+func (s *CallAteletTerminateStep) RetryBackoff() *wait.Backoff { return nil }
+
+type DetachVolumesForDeleteStep struct {
+	store store.Interface
+}
+
+func (s *DetachVolumesForDeleteStep) Name() string { return "DetachVolumesForDelete" }
+
+func (s *DetachVolumesForDeleteStep) IsComplete(ctx context.Context, input *DeleteInput, state *DeleteState) (bool, error) {
+	return state.Actor.GetStatus() == ateapipb.Actor_STATUS_DELETING, nil
```
sandeep-gh-acc: same isComplete not using err here.

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -98,6 +165,151 @@ func (s *MarkDeletingStep) Execute(ctx context.Context, input *DeleteInput, stat
...
+		}
+
+		latestActor, err = s.store.GetActor(ctx, input.ActorRef)
+		if err != nil {
+			return err
+		}
+
+		// TODO: can this be done inside releaseWorker()?
+		latestActor.LocalSnapshotInfo = nil
+		updatedActor, err := s.store.UpdateActor(ctx, latestActor, latestActor.GetMetadata().GetVersion())
```
sandeep-gh-acc: so if update actor fails and release worker succeeds. Is release worker op idempotent, in a retry would release worker succeed if its already released?

cmd/ateom-microvm/checkpoint.go
```diff
@@ -303,3 +303,39 @@ func (s *AteomService) teardownActor(ctx context.Context, id string, ra *running
...
+
+	return &ateompb.TerminateWorkloadResponse{}, nil
+}
+
+func (s *AteomService) terminateWorkload(ctx context.Context, actorUID string) {
+	if err := s.deactivateActorNetworking(ctx); err != nil {
+		slog.WarnContext(ctx, "Failed to deactivate actor networking during terminate", slog.Any("err", err))
+	}
+
+	ra := s.running[actorUID]
```
sandeep-gh-acc, line 376: Would it be guaranteed as running when the termination is called? what if it got suspended in interim

cmd/ateapi/internal/controlapi/functional_test.go
```diff
@@ -198,6 +202,23 @@ func (f *FakeAteletServer) Reset() {
 	f.RestoreRequest = nil
 	f.FailRestore = nil
 	f.RestoreDelay = 0
+
+	f.TerminateCalled = false
+	f.TerminateRequest = nil
+	f.FailTerminate = nil
+}
+
+func (f *FakeAteletServer) Terminate(ctx context.Context, req *ateletpb.TerminateRequest) (*ateletpb.TerminateResponse, error) {
```
sandeep-gh-acc: Should we have some negative tests, where actor worker is unresponsive etc. Whether that is still able to force delete in those hang or failure conditions

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -62,23 +94,205 @@ func (w *ActorWorkflow) loadActorForDelete(ctx context.Context, actorRef resourc
...
+
+	if force {
+		switch st {
+		case ateapipb.Actor_STATUS_RUNNING,
+			ateapipb.Actor_STATUS_RESUMING,
+			ateapipb.Actor_STATUS_SUSPENDING,
+			ateapipb.Actor_STATUS_PAUSING,
+			ateapipb.Actor_STATUS_PAUSED,
+			ateapipb.Actor_STATUS_CRASHED,
+			ateapipb.Actor_STATUS_TERMINATING,
```
sandeep-gh-acc: TERMINATING and DELETING state cannot be reached. in the case as earlier check already returns error

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -62,23 +94,205 @@ func (w *ActorWorkflow) loadActorForDelete(ctx context.Context, actorRef resourc
...
+		markSkipped(ctx, "actor already DELETING")
+		return nil
+	}
+
+	st := actor.GetStatus()
+	if st != ateapipb.Actor_STATUS_TERMINATING {
+		return status.Errorf(codes.FailedPrecondition, "CallAteletTerminate prerequisite not met for Actor: %s (got: %v, want %s)", actorRef, st, ateapipb.Actor_STATUS_TERMINATING)
+	}
+	if actorTemplate == nil {
+		return status.Errorf(codes.FailedPrecondition, "actor template %s/%s not found for actor %s", actor.GetActorTemplateNamespace(), actor.GetActorTemplateName(), actorRef)
```
sandeep-gh-acc: What would happen if the actor template goes missing? would this clean up the actor, or atleast not lead to stuckness?

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -62,23 +94,205 @@ func (w *ActorWorkflow) loadActorForDelete(ctx context.Context, actorRef resourc
...
+		slog.InfoContext(ctx, "actor has no worker assignment, skipping CallAteletTerminateStep", slog.Any("actor", actorRef))
+		return nil
+	}
+
+	workerPodNs := assignment.GetWorkerNamespace()
+	workerPodName := assignment.GetWorkerPod()
+
+	conn, err := w.dialer.DialForWorker(workerPodNs, workerPodName)
+	if err != nil {
+		if errors.Is(err, ErrWorkerPodNotFound) {
```
sandeep-gh-acc, line 145: Should we be deleting metadata / clean up the actor even when associated worker was not found?

sandeep-gh-acc: In the force case atleast?

cmd/ateom-microvm/checkpoint.go
Sneha-at: The method does not return any errors, I see `TerminateWorkload` does return error but there is actually no error reported in it's implementation, I think we can update the return value. I also checked other methods called by TerminateWorkload - `teardownActor` and that too doesn't seem to report any error

cmd/ateom-microvm/checkpoint.go
Sneha-at: The `running` variable is a bit confusing, it actually tracks the VMs the actorPod is running on and is not related to the state of the Actor.

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Considering a scenario where the Actor was deleted with force at T0 and then attempted again to be deleted without force at T1 where the gap between T0 and T1 is minimal we can run into a state where the request at T1 finds the Actor in `TERMINATING` state (as mentioned in the comment). If so I agree that it is a no-op. However, if the Actor is already in `DELETING` status then the current logic updates it to `TERMINATING` which is actually taking a step back (For `force` workflow X status -> TERMINATING -> DELETING). I think we should have a different workflow for `DELETING` status.
```
} else {
		switch st {
		case ateapipb.Actor_STATUS_SUSPENDED:
			// allowed
        case ateapipb.Actor_STATUS_DELETING:
             return actor, nil // return the same actor object we received as the function arg
		default:
			return nil, status.Errorf(codes.FailedPrecondition, "Actor %s is not in a deletable status (status: %v)", actorRef, st)
		}
	}
```

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: ActorTemplate is first referred in ensureAteletTerminated where we return err if the template is `nil`. I suppose that should be alright as we do fail soon?
```
	if actorTemplate == nil {
		return status.Errorf(codes.FailedPrecondition, "actor template %s/%s not found for actor %s", actor.GetActorTemplateNamespace(), actor.GetActorTemplateName(), actorRef)
	}
```

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Do you mean we should have backoff for DeleteActor and retry on failure?

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Earlier check as in before the ensureMarkedTerminating function is called?

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Could you please highlight which function are you referring to? I see we return some custom errors with `return status.Errorf(.....)`

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: If the worker is released already from the actor the ensureWorkerReleased is just a no-op
```
	if actor.GetWorkerAssignment() == nil {
		markSkipped(ctx, "worker already released")
		return actor, nil
	}
```

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Ohh I see the comments are on the older version of the code. While rebasing the changes to latest main I realized there were some significant impact of Workflow lifecycle implementation so I had to re-write parts of the code. The code blocks this comments refers to no longer exists.

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Same here,  the code blocks this comments refers to no longer exists.

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: So ignore atletTermination error when force deleting?
```
if err := w.ensureAteletTerminated(ctx, actorRef, actor, actorTemplate); err != nil {
		if force {
			slog.WarnContext(ctx, "ignoring atelet termination failure during force delete", "error", err)
		} else {
			return nil, err
		}
	}

	if err := w.ensureVolumesDetachedForDelete(ctx, actor, actorTemplate); err != nil {
		return nil, err
	}

	if actor, err = w.ensureWorkerReleased(ctx, actorRef, actor); err != nil {
		return nil, err
	}
```

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Please reopen if needed.

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: We need the actor template to fetch the workload spec and detach the volumes from the actor. We can still continue to release the worker if the Actor Template doesn't exist. Though do we think that is a possible as Actor is an instance of actor template

cmd/ateom-microvm/checkpoint.go
```diff
@@ -303,3 +303,39 @@ func (s *AteomService) teardownActor(ctx context.Context, id string, ra *running
...
+
+	s.terminateWorkload(ctx, actorUID)
+
+	s.actorLogger.EmitLifecycleLog("Actor terminated", actorRef, actorUID, req.GetActorTemplateNamespace(), req.GetActorTemplateName())
+
+	return &ateompb.TerminateWorkloadResponse{}, nil
+}
+
+func (s *AteomService) terminateWorkload(ctx context.Context, actorUID string) {
+	if err := s.deactivateActorNetworking(ctx); err != nil {
```
zoez7, line 372: If this fails to clean up, is the worker pod safe to assign to another Actor (I'm assuming no? @LiorLieberman )

cmd/ateom-microvm/checkpoint.go
```diff
@@ -303,3 +303,39 @@ func (s *AteomService) teardownActor(ctx context.Context, id string, ra *running
...
+	chSocket := kata.CLHSocketPath(actorUID)
+	if ra != nil && ra.apiSocket != "" {
+		chSocket = ra.apiSocket
+	}
+	client := ch.NewClient(chSocket)
+
+	s.teardownActor(ctx, actorUID, ra, client)
+	delete(s.running, actorUID)
+
+	if err := ateomnet.CleanupActorNetwork(ctx, s.interiorNetNS); err != nil {
```
zoez7, line 401: Same here, since this doesn't return an error, would we accidentally release an worker with dirty state?

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -62,23 +98,202 @@ func (w *ActorWorkflow) loadActorForDelete(ctx context.Context, actorRef resourc
...
+	if st != ateapipb.Actor_STATUS_TERMINATING {
+		return status.Errorf(codes.FailedPrecondition, "CallAteletTerminate prerequisite not met for Actor: %s (got: %v, want %s)", actorRef, st, ateapipb.Actor_STATUS_TERMINATING)
+	}
+	if actorTemplate == nil {
+		return status.Errorf(codes.FailedPrecondition, "actor template %s/%s not found for actor %s", actor.GetActorTemplateNamespace(), actor.GetActorTemplateName(), actorRef)
+	}
+
+	assignment := actor.GetWorkerAssignment()
+	if assignment == nil {
+		slog.InfoContext(ctx, "actor has no worker assignment, skipping CallAteletTerminateStep", slog.Any("actor", actorRef))
```
zoez7: CallAteletTerminateStep no longer exist?

cmd/ateapi/internal/controlapi/delete_actor.go
```diff
@@ -45,7 +45,7 @@ func (s *Service) DeleteActor(ctx context.Context, req *ateapipb.DeleteActorRequ
 	actorRef := resources.ActorRefFromObjectRef(req.GetActor())
 	setSpanActorRefAttributes(ctx, actorRef)
 
-	deleted, err = s.actorWorkflow.DeleteActor(ctx, actorRef)
+	deleted, err = s.actorWorkflow.DeleteActor(ctx, actorRef, req.GetForce())
```
zoez7: Reading the description of https://github.com/agent-substrate/substrate/issues/643 I think adding a request field is insufficient, with just this alone, the customer will need to repeatedly call `delete actor --force` until it succeeds?

I think we need some sort of "wipeout" mechanism where we
1. the API itself just mark the Actor as "force deleting" and return.
2. have a loop that iterates through all actors that are in the "force deleting" state, and retry clean up until there are no errors.

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -38,12 +42,44 @@ func (w *ActorWorkflow) DeleteActor(ctx context.Context, actorRef resources.Acto
...
+
+	if err := w.ensureVolumesDetachedForDelete(ctx, actor, actorTemplate); err != nil {
+		return nil, err
+	}
+
+	if actor, err = w.ensureWorkerReleased(ctx, actorRef, actor); err != nil {
+		return nil, err
+	}
+
 	if actor, err = w.ensureMarkedDeleting(ctx, actorRef, actor); err != nil {
```
zoez7: This seem to conflict with line 55 ?

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -62,23 +98,202 @@ func (w *ActorWorkflow) loadActorForDelete(ctx context.Context, actorRef resourc
...
+
+	return detachActorVolumes(ctx, w.store, w.pluginRegistry, actor, actorTemplate, "delete")
+}
+
+// ensureWorkerReleased releases the worker assigned to the actor.
+func (w *ActorWorkflow) ensureWorkerReleased(ctx context.Context, actorRef resources.ActorRef, actor *ateapipb.Actor) (updated *ateapipb.Actor, err error) {
+	ctx, done := stepSpan(ctx, "ReleaseWorker")
+	defer func() { err = done(err) }()
+
+	if actor.GetStatus() == ateapipb.Actor_STATUS_DELETING {
```
zoez7: Can an Actor get stuck in Deleting state? I think of  "force deleting" as it can be used to wipe out actors in any state.

cmd/atelet/main.go
```diff
@@ -1111,6 +1111,45 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+		// TODO: handle case where the target ateom is no longer available
+		return nil, fmt.Errorf("failed to dial ateom for terminate (actor: %s, actorUID: %s): %w", actorRef, actorUID, err)
+	}
+	if _, err := client.TerminateWorkload(ctx, &ateompb.TerminateWorkloadRequest{
+		Atespace:               req.GetAtespace(),
+		ActorName:              req.GetActorName(),
+		ActorUid:               req.GetActorUid(),
+		ActorTemplateNamespace: req.GetActorTemplateNamespace(),
+		ActorTemplateName:      req.GetActorTemplateName(),
+		Spec:                   buildAteomWorkloadSpec(req.GetSpec()),
```
zoez7: I don't see the path to runsc binary here? The runsc command iin ateom-gvisor below are silently failing.

zoez7 (changes requested):
Also, we need to update ActorIdentity package to deny actors that are being force deleted.  https://github.com/agent-substrate/substrate/blob/730250dca1d02e50fea36ea96b080474ff625b5f/cmd/ateapi/internal/actoridentity/actoridentity.go

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: We intend to move the actor to `DELETING` status post `TERMINATING` status once we deleted all the resources. A new TERMINATING status is added to indicate that we're starting the termination sequence. The termination sequence will call atelet to terminate the workload. Once atelet termination succeeds, then we will proceed with control plane termination sequence. Finally we will move to the existing DELETING status to delete the actor resources. Let me know if you were referring to something else.

cmd/ateapi/internal/controlapi/delete_actor.go
zoez7: Discussed offline. There are 2 problems here
1. We need a way to call DeleteActor for an Actor that is not in suspended/crashed state. In which case, we would suspend the actor (without uploading any snapshot) first, and delete any remaining resources, this is best effort and may fail.
2. When the above fails, we should have some garbage collection going on in the background to periodically clean things up.

This PR will focus on the 1 aboe - delete actor that is not in supended / crashed state.

zoez7:
I took a stab at removing the "TERMINATING" state all together, I think what this PR wants to achieve is to allow "delete actor for any state" maybe `kubectl ate delete actor --any-state` is a closer description?

https://paste.googleplex.com/5252227961847808

What I did was:
1. using --any-state just for a eligibility check
2. try reaching out to ateom and atelet to delete all resources, and get the error back. Removed the separate "Terminating" state.
3. only delete the actor from Redis when there are no errors (this will allow us to scan all DELETING actors in the future, move to a different key to allow users to recreate a new actor with the same name)

cmd/atelet/main.go
Sneha-at: This is fixed

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Actor can be stuck in deleting state, based on our discussion I have moved this flag to say `--any-state` instead of `--force`. Also I have updated the cleanup to be idempotent so user can call Delete again on a stuck actor.

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: This is no longer relevant, we do not have Terminating status anymore

cmd/ateom-microvm/checkpoint.go
Sneha-at: Updated this to return error on failure

cmd/ateom-microvm/checkpoint.go
Sneha-at: Updated this to return error on failure

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Please review again, the logic has been updated

cmd/ateapi/internal/controlapi/functional_test.go
Sneha-at: Added some more tests. This PR allows us to delete an actor from any state, it doesn't handle the scenario if the actor is stuck in DELETING state.

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Please review again, we no longer have `Terminating` status

cmd/ateom-microvm/checkpoint.go
Sneha-at: Updated this to return error on failure, only for checkpoint we log the error as warning as it is best effort cleanup

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: This is outdated

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Done

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: I have updated the logic to delete an actor even if actor template was deleted

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -771,6 +771,7 @@ message ResumeActorResponse {
 
 message DeleteActorRequest {
   ObjectRef actor = 1;
+  bool force = 2;
```
zoez7: nit: `bool all_states` instead?

internal/proto/ateompb/ateom.proto
```diff
@@ -98,6 +98,25 @@ service Ateom {
...
+
+message TerminateWorkloadRequest {
+  string atespace = 1;
+  string actor_name = 2;
+  string actor_uid = 3;
+
+  string actor_template_namespace = 4;
+  string actor_template_name = 5;
+
+  string runsc_path = 6;
```
zoez7: What about microVM? Is that going to be a followup PR?

cmd/ateom-microvm/checkpoint.go
```diff
@@ -342,23 +308,57 @@ func (s *AteomService) teardownActor(ctx context.Context, id string, ra *running
...
+		if !errors.Is(err, os.ErrNotExist) {
+			errs = append(errs, fmt.Errorf("while unmounting bundle rootfs overlays: %w", err))
+		}
+	}
+	return errors.Join(errs...)
+}
+
+// TerminateWorkload stops the running actor, tears down its VMM, and cleans up
+// networking and overlays.
+func (s *AteomService) TerminateWorkload(ctx context.Context, req *ateompb.TerminateWorkloadRequest) (*ateompb.TerminateWorkloadResponse, error) {
```
zoez7: lgtm. @BenTheElder Could you help take a look ? The goal here is to terminate the microVM  actor without saving any state, to be able to delete a running actor.

cmd/ateom-microvm/checkpoint.go
```diff
@@ -206,20 +176,13 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 	// the two views of "is an actor here" from disagreeing.
 	s.activeActor.Store(nil)
```
zoez7: We can call `s.activeActor.Store(nil)` in terminateWorkload as well?

cmd/ateom-gvisor/main.go
```diff
@@ -830,27 +814,48 @@ func listSnapshotFiles(dir string) ([]string, error) {
 	return files, nil
 }
 
+func isRunscNotFoundError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	return errors.Is(err, os.ErrNotExist) ||
+		strings.Contains(msg, "does not exist") ||
```
zoez7: Nit: matching error strings is an anti-pattern.

cmd/ateom-gvisor/main.go
```diff
@@ -830,27 +814,48 @@ func listSnapshotFiles(dir string) ([]string, error) {
...
-		if err := r.cmdState(ctx, ctr.GetName()); err != nil {
-			return fmt.Errorf("while checking state of %q application container: %w", ctr.GetName(), err)
-		}
+		_ = r.cmdState(ctx, ctr.GetName())
 	}
 
 	for _, ctr := range containers {
 		if err := r.cmdDelete(ctx, ctr.GetName()); err != nil {
-			return fmt.Errorf("while deleting %q application container: %w", ctr.GetName(), err)
+			if !isRunscNotFoundError(err) {
```
zoez7: Could you explain why we skip the `runsc not found` error?

cmd/ateom-gvisor/main.go
```diff
@@ -1051,6 +1056,58 @@ func (s *AteomService) prepareActorEgress(ctx context.Context, actorUID string,
...
+	rcmd := &runsc{
+		path:     runscPath,
+		actorUID: actorUID,
+	}
+
+	// After checkpointing the sandbox root, runsc may no longer have a usable
+	// control server for state/delete calls. Keep this as best-effort cleanup:
+	// atelet resets the actor runsc, bundle, pidfile, and checkpoint
+	// directories after uploading the snapshot.
+	if err := rcmd.cleanupContainersAfterCheckpoint(ctx, containers); err != nil {
```
zoez7: nit: should we rename `cleanupContainersAfterCheckpoint` since we don't checkpoint it here?

cmd/ateom-gvisor/main.go
```diff
@@ -830,27 +814,48 @@ func listSnapshotFiles(dir string) ([]string, error) {
...
+	}
+	msg := err.Error()
+	return errors.Is(err, os.ErrNotExist) ||
+		strings.Contains(msg, "does not exist") ||
+		strings.Contains(msg, "not found") ||
+		strings.Contains(msg, "no such file or directory")
+}
+
 func (r *runsc) cleanupContainersAfterCheckpoint(ctx context.Context, containers []*ateompb.Container) error {
+	if r.path == "" {
```
zoez7: if runsc path is empty, this should return a failed precondition error?

cmd/ateapi/internal/controlapi/volumes.go
```diff
@@ -211,14 +215,30 @@ func detachActorVolumes(ctx context.Context, st detachActorVolumesStore, registr
...
+	// If the template is available, only detach volumes that are actively mounted
+	// in the template's containers. If the template is missing/deleted, fall back to
+	// attempting detachment for all external volumes recorded on the actor so we do
+	// not orphan attached disks on the worker node.
+	volumesToDetach := actor.GetActorVolumes()
+	if template != nil {
+		volumesToDetach = getMountedActorVolumes(ctx, ref, actor.GetActorVolumes(), template)
+	}
+	for _, vol := range volumesToDetach {
+		// Skip volumes that do not have a provisioned storage volume ID.
```
zoez7: is this expected to happen? do we need to log a warning?

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -38,11 +42,73 @@ func (w *ActorWorkflow) DeleteActor(ctx context.Context, actorRef resources.Acto
...
+			actorTemplate = tmpl
+		}
+	} else if !anyState {
+		errs = append(errs, status.Errorf(codes.FailedPrecondition, "actor %s has no actor template configured; use --any-state to force delete", actorRef))
+	}
+
+	var atletTerminatedErr, volumesDetachedErr error
+	if err := w.ensureAteletTerminated(ctx, actorRef, actor, actorTemplate); err != nil {
+		atletTerminatedErr = fmt.Errorf("while terminating atelet: %w", err)
+		if !anyState {
```
zoez7: so when "anyState" is true, the actor resource will be gone, and we won't have a way to clean it up.

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -38,11 +42,73 @@ func (w *ActorWorkflow) DeleteActor(ctx context.Context, actorRef resources.Acto
...
+	}
+
 	if err := w.ensureVolumesDeleted(ctx, actor); err != nil {
-		return nil, err
+		if !anyState {
+			errs = append(errs, fmt.Errorf("while deleting volumes: %w", err))
+		}
+	}
+
+	if len(errs) > 0 {
```
zoez7, line 90: if --any-state is true, we would lose track of leaked resources and will not be able to retry, because the errs is always empty, is that intended? How would we delete the leaked resources later?

cmd/ateom-gvisor/main.go
```diff
@@ -830,27 +814,48 @@ func listSnapshotFiles(dir string) ([]string, error) {
...
+		return nil
+	}
+	for _, ctr := range containers {
+		_ = r.cmdKill(ctx, ctr.GetName(), "SIGKILL")
+		_ = r.cmdWait(ctx, ctr.GetName())
+	}
+	_ = r.cmdKill(ctx, "pause", "SIGKILL")
+	_ = r.cmdWait(ctx, "pause")
```
zoez7, line 833: Do we need to call these for already checkpointed containers?

zoez7: maybe
```
func (r *runsc) stopContainers(ctx context.Context, containers []*ateompb.Container) {
	for _, ctr := range containers {
		_ = r.cmdKill(ctx, ctr.GetName(), "SIGKILL")
		_ = r.cmdWait(ctx, ctr.GetName())
	}
	_ = r.cmdKill(ctx, "pause", "SIGKILL")
	_ = r.cmdWait(ctx, "pause")
}

```

and then in 
```
func (s *AteomService) terminateWorkload(ctx context.Context, actorRef resources.ActorRef, actorUID, runscPath string, containers []*ateompb.Container, sandboxExited bool) error {
  ...

  rcmd.stopContainers(ctx, containers)	
  rcmd.cleanupContainers(...)
}
```

zoez7 (changes requested):

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: I have updated the logic to report errors if anything fails with any-state but continue cleanup. I agree we shouldn't be ignoring the errors there. We don't have any GC implementation for these resources but we can know from the error where did the clean up fail

ensureAteletTerminated and ensureVolumesDetachedForDelete are enhanced to do best-effort cleanup with `--any-state` even if actorTemplate is not available

cmd/ateom-gvisor/main.go
Sneha-at: cleanupContainer is run as a part of `terminateWorkload` workflow, to ensure this is idempotent so that we can retry DELETION we consider `r.path==nil` as a already cleanedup state. Do you foresee any cases where this could be an issue during cleanup?

cmd/ateom-gvisor/main.go
Sneha-at: Makes sense, updating this!

cmd/ateom-gvisor/main.go
Sneha-at: This is to ensure cleanupContainer step in the workflow would be idempotent. While retrying a Delete we can have a state where the container is already cleaned up or we do not have any sandbox assets. We do not want to fail under those conditions as this assumes the asset was already cleanup (hence "NotFound") thus we do not report these as errors. We can probably update this to be restricted to be NOT and issue if `any-state` is set but it is a problem in standard workflow?

internal/proto/ateompb/ateom.proto
```diff
@@ -98,6 +98,25 @@ service Ateom {
...
+  // network and bundle overlays on ateom.
+  rpc TerminateWorkload(TerminateWorkloadRequest) returns (TerminateWorkloadResponse) {}
+}
+
+message TerminateWorkloadRequest {
+  string atespace = 1;
+  string actor_name = 2;
+  string actor_uid = 3;
+
+  string actor_template_namespace = 4;
```
zoez7, line 112: Do we need actor_template_namespace and actor_template_name in TerminateWorkloadRequest? Can Ateom terminate the workload without knowing these parameters?

cmd/ateom-gvisor/main.go
```diff
@@ -840,7 +814,31 @@ func listSnapshotFiles(dir string) ([]string, error) {
 	return files, nil
 }
 
-func (r *runsc) cleanupContainersAfterCheckpoint(ctx context.Context, containers []*ateompb.Container) error {
+func isRunscNotFoundError(err error) bool {
```
zoez7: The contract between ateom and atelet is that atelet should make sure that the expected runsc version is downloaded and stored on disk, ateom should return a FailedPrecondition error if that is not met.

cmd/ateapi/internal/controlapi/workflow_delete.go
Sneha-at: Discussed offline, this method now reports errors so resolving this

internal/proto/ateompb/ateom.proto
Sneha-at: Discussed offline: A recent PR grouped this common metadata for telemetry into [ActorAttribution](https://github.com/agent-substrate/substrate/blob/c618baa593a29ef00a1eccd1ae0d6ae46b1d05e0/cmd/ateom-microvm/run.go#L326-L332) https://github.com/agent-substrate/substrate/commit/c3c68769856d0ccdc6e8cc186b857343b48a0d7e

We decided to follow similar format in Terminate Workload too and create an [ActorAttribution](https://github.com/agent-substrate/substrate/blob/c618baa593a29ef00a1eccd1ae0d6ae46b1d05e0/cmd/ateom-microvm/run.go#L326-L332)
1. [CheckpointWorkloadRequest](https://github.com/agent-substrate/substrate/blob/c618baa593a29ef00a1eccd1ae0d6ae46b1d05e0/internal/proto/ateompb/ateom.proto#L214-L220):Used to create [ActorAttribution](https://github.com/agent-substrate/substrate/blob/c618baa593a29ef00a1eccd1ae0d6ae46b1d05e0/cmd/ateom-gvisor/main.go#L731) record
2. [RunWorkloadRequest](https://github.com/agent-substrate/substrate/blob/c618baa593a29ef00a1eccd1ae0d6ae46b1d05e0/internal/proto/ateompb/ateom.proto#L104-L110): Used to create an [ActorAttribution](https://github.com/agent-substrate/substrate/blob/c618baa593a29ef00a1eccd1ae0d6ae46b1d05e0/cmd/ateom-gvisor/main.go#L608C2-L608C60) record
3. [RestoreWorkloadRequest](https://github.com/agent-substrate/substrate/blob/c618baa593a29ef00a1eccd1ae0d6ae46b1d05e0/internal/proto/ateompb/ateom.proto#L252-L259): --> We don't use this for commands anywhere, it more used as a part of [ActorAttribution](https://github.com/agent-substrate/substrate/blob/c618baa593a29ef00a1eccd1ae0d6ae46b1d05e0/cmd/ateom-gvisor/main.go#L885) which stores the actor details

cmd/ateom-gvisor/main.go
Sneha-at: Discussed offline: this is not a valid scenario so removing runsc not found error checks. Atlet downloads runsc binary on every Terminate call so this is not expected

cmd/ateapi/internal/controlapi/volumes.go
Sneha-at: I have updated the comment, this is expected. If the volume was never provisioned (stuck in deleting) then we have nothing to delete so it is safe to skip these cases

cmd/ateom-gvisor/main.go
Sneha-at: Discussed offline: we should error on such cases so removing the skips the command will fail if this ever happens. It is not actually expected though

cmd/ateom-gvisor/main.go
Sneha-at: Done

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -38,11 +42,51 @@ func (w *ActorWorkflow) DeleteActor(ctx context.Context, actorRef resources.Acto
...
+	if err := w.ensureAteletTerminated(ctx, actorRef, actor, actorTemplate); err != nil {
+		atletTerminatedErr = fmt.Errorf("while terminating atelet: %w", err)
+		errs = append(errs, atletTerminatedErr)
+	}
+	if err := w.ensureVolumesDetachedForDelete(ctx, actor, actorTemplate); err != nil {
+		volumesDetachedErr = fmt.Errorf("while detaching volumes: %w", err)
+		errs = append(errs, volumesDetachedErr)
+	}
+
+	// Release worker if atelet termination and volume detachment did not fail (or if anyState is true).
```
zoez7: stale comment?

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -38,11 +42,51 @@ func (w *ActorWorkflow) DeleteActor(ctx context.Context, actorRef resources.Acto
 	if err != nil {
 		return nil, err
 	}
-	if actor, err = w.ensureMarkedDeleting(ctx, actorRef, actor); err != nil {
+
+	if actor, err = w.ensureMarkedDeleting(ctx, actorRef, actor, anyState); err != nil {
 		return nil, err
 	}
+
+	// DeleteActor will attempt the best effort cleanup, we will log errors if something fails but will continue to delete the actor from the store.
```
zoez7: stale comment?

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -62,20 +106,159 @@ func (w *ActorWorkflow) loadActorForDelete(ctx context.Context, actorRef resourc
 	return actor, nil
 }
 
+// ensureAteletTerminated calls atelet to terminate the workload.
+func (w *ActorWorkflow) ensureAteletTerminated(ctx context.Context, actorRef resources.ActorRef, actor *ateapipb.Actor, actorTemplate *atev1alpha1.ActorTemplate) (err error) {
+	ctx, done := stepSpan(ctx, "CallAteletTerminate")
+	defer func() { err = done(err) }()
+
+	assignment := actor.GetStatus().GetWorkerAssignment()
```
zoez7, line 116: also add a check that the worker is still assigned to the same actor, otherwise we might kill a worker assigned to another actor

cmd/atelet/main.go
```diff
@@ -1195,6 +1195,67 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
+	}
+
+	actorRef := resources.ActorRef{Atespace: req.GetAtespace(), Name: req.GetActorName()}
+	actorUID := req.GetActorUid()
+
+	var assetPaths map[string]string
+	sandboxRec, err := readSandboxRecord(actorUID)
+	if err != nil {
+		slog.WarnContext(ctx, "failed to read sandbox record during terminate", slog.Any("actor", actorRef), slog.String("actorUID", actorUID), slog.Any("err", err))
```
zoez7: return the error here instead of logging?

cmd/atelet/main.go
```diff
@@ -1195,6 +1195,67 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+	} else {
+		paths, err := s.ensureSandboxAssets(ctx, sandboxRec)
+		if err != nil {
+			slog.WarnContext(ctx, "failed to ensure sandbox assets during terminate", slog.Any("actor", actorRef), slog.String("actorUID", actorUID), slog.Any("err", err))
```
zoez7: same here, return the error?

cmd/atelet/main.go
```diff
@@ -1195,6 +1195,67 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+		} else {
+			assetPaths = paths
+		}
+	}
+
+	client, err := s.dialAteom(ctx, req.GetTargetAteomUid())
+	if err != nil {
+		slog.WarnContext(ctx, "failed to dial ateom for terminate, proceeding with volume unmount and dir reset", slog.Any("actor", actorRef), slog.String("actorUID", actorUID), slog.Any("err", err))
```
zoez7: same here, return the error?

zoez7 (dismissed):

PR #792, opened 2026-08-07
Make envtest offline-capable and factor out TestMain boilerplate

### LINKED ISSUE #789: pervasive envtest makes it hard to run our tests on a plane
Having a custom `TestMain` in ./cmd/ateapi/internal/controlapi meant I could not figure out how to run tests without being on-line.

Can we factor this better?

HavenXia:
`setup-envtest use` with no version re-resolves "latest" against the remote release index on every run, so the envtest-backed packages needed network even with fully cached binaries. 

This PR pins the control plane to `1.36.x`, and fold the three per-package `envtest` into `internal/testenv.Start`, which also honors `-short`: `go test -short ./...` skips the envtest-backed packages and is the guaranteed-offline path.

Have Claude to help me fix it

Fixes #789

- [ ] Tests pass

### DIFF
--- a/hack/run-tool.sh
+++ b/hack/run-tool.sh
@@ -30,7 +30,7 @@ fi
 TOOL_NAME="$1"
 shift
 
-ROOT="$(git rev-parse --show-toplevel)"
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 case "${TOOL_NAME}" in
   "client-gen"|"informer-gen"|"lister-gen")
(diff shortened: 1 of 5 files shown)


hack/run-tool.sh
```diff
@@ -30,7 +30,7 @@ fi
 TOOL_NAME="$1"
 shift
 
-ROOT="$(git rev-parse --show-toplevel)"
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
```
thockin, line 33: Why this change?

PR #795, opened 2026-08-07
Allow actors to opt in/out of Linux capabilities

### LINKED ISSUE #744: Allow actors to opt-in/out of Linux capabilities
The set of capabilities we grant to actor sandboxes are hardcoded today. There should be a way for actors to specify which capabilities to add / drop, similar to `securityContext.capabilities` in k8s Pod's.

It would be nice to have a way to define this in the `ActorTemplate` container. We could follow a similar patter as k8s where users can specify capabilities to add / drop, from the default set of capabilities from the sandbox class (i.e gvisor, cloud-hypervisor).

nmn3m:
Fixes #744
- [x] Tests pass


BenTheElder (commented):
can we add an e2e test for this, for both uVM and gvisor? plumbing it is one thing, making sure it is effective is another, atelet doesn't actually spawn the containers

.github/workflows/pr-workflow.yaml
```diff
@@ -110,6 +110,14 @@ jobs:
         E2E_TEMPLATE_NAME: counter-microvm
         E2E_TEMPLATE_READY_TIMEOUT: 600s
       run: hack/run-e2e-kind.sh ./internal/e2e/suites/demo -v -args --no-color
```
BenTheElder: maybe we should expand this to run the whole default suite instead?

cmd/atelet/oci.go
```diff
@@ -235,26 +279,10 @@ func buildActorOCISpec(actorUID string, args []string, env []string, annotations
...
-					"CAP_NET_BIND_SERVICE",
-				},
-				Permitted: []string{
-					"CAP_AUDIT_WRITE",
-					"CAP_KILL",
-					"CAP_NET_BIND_SERVICE",
-				},
+				Bounding:    capabilities,
+				Effective:   capabilities,
+				Inheritable: capabilities,
```
BenTheElder: 🤖 **should-fix** 🟡 – Leave `Inheritable` empty. It was harmless when this was three fixed, innocuous capabilities; now a template can put `SYS_ADMIN`, `DAC_OVERRIDE`, or `SETUID` there.

Inheritable grants the process itself nothing. It only takes effect on `execve`, ANDed with the file's own inheritable set — which is why containerd dropped it in the fix for CVE-2022-24769, and why CRI-O and Docker set it empty too. The exposure is a container that starts as root and then drops to an unprivileged uid (`su`, `gosu`, a `setuid` binary) before exec'ing something with file capabilities.

Removing it costs nothing here. The spec pins `User{UID: 0, GID: 0}`, and for a real-uid-0 process exec'ing a file with no file capabilities the kernel treats `F(permitted)` and `F(inheritable)` as all-ones, so `P'(permitted) = P(bounding)`. Propagation to children rides on `Bounding`, not `Inheritable`.

The capabilities e2e already asserts only bounding, effective, permitted and an empty ambient, so it would not need to change — `TestBuildActorOCISpecCapabilities` in `oci_test.go` would.

docs/api-guide.md
```diff
@@ -175,9 +175,35 @@ Each entry in `containers` describes one process to run in the actor's sandbox.
 | `env` | `[]EnvVar` | Optional. Literal `value` entries. |
 | `readyz` | `ContainerReadyz` | Optional. HTTP readiness probe — see [Container Readiness Probe](#container-readiness-probe-readyz). |
 | `volumeMounts` | `[]VolumeMount` | Optional. Mounts a `spec.volumes` entry (e.g. `durableDir`) into this container. |
+| `securityContext` | `SecurityContext` | Optional. Security settings for the container process — see [Container Capabilities](#container-capabilities-securitycontextcapabilities). |
 
 `command` and `args` resolve against the container image's `ENTRYPOINT`/`CMD` the same way [Kubernetes Pod `command`/`args`](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/) resolve against `ENTRYPOINT`/`CMD`. If the resolved argv is empty — the image sets neither `ENTRYPOINT` nor `CMD`, and the container sets neither `command` nor `args` — `Run`/`Restore` fails.
 
+### Container Capabilities (`securityContext.capabilities`)
+
+Each container runs with a default set of Linux capabilities. `securityContext.capabilities` adjusts that set, mirroring `securityContext.capabilities` on a Kubernetes Pod container.
```
BenTheElder: 🤖 **nit** 🟢 – Worth naming the default set here. A reader writing `add` needs to know what they already have, and one writing `drop` needs it even more — right now the only way to find out is to read `defaultCapabilities` in `cmd/atelet/oci.go`. It is three entries: `AUDIT_WRITE`, `KILL`, `NET_BIND_SERVICE`.

BenTheElder (commented):
Feel free to drop me a ping in the slack for follow-up when these are fixed / PR is rebased and I'll try to come back to this faster, thanks!

BenTheElder:
And sorry for the delay, the incoming PR volume is high 😅

nmn3m:
@BenTheElder TestActorCapabilities passed on gVisor (24.75s, both subtests), but the overall run failed on two tests. I’m not sure how to interpret the failures. Could you take a look and share your feedback?

I also ran `make test` locally, and it passed, so I’m not sure what’s causing the difference.

BenTheElder:
You can search the logs for `--- fail` (`go test` output format)  in github to find the actual failures, also rebasing on main picks up fixes (we might do https://github.com/agent-substrate/substrate/pull/1029)

`TestMakeCert` failed in unit tests and a couple tests failed in gvisor, they look unrelated, I'm rerunning

BenTheElder (approved):

BenTheElder (dismissed):
LGTM, needs a rebase (couldn't merge before due to flaking tests it seems, rebase on main should help with that as well)

BenTheElder (approved):

PR #797, opened 2026-08-07
controlapi: refactor workflows to typed ensure-steps

### LINKED ISSUE #554: Actor's suspend / resume / pause workflows need to be clearly reentrancy safe
The suspend / resume / pause workflow is getting really hard to reason about correctness. With multiple replicas of ate-apiserver, as well as ate-apiserver disruptions during upgrades, it's getting harder to reason about correctness and reentrancy. (i.e. What will happen if ate-apiserver crashed at different stages, and all possible status of worker and actor when ate-apiserver (or the user) retries the operation. 

We need to restructure the workflows so humans can clearly understand and reason about correctness.

troychiu:
Fixes #554 

Each workflow becomes a plain method orchestrating typed, idempotent ensureX steps, e.g.:

```go
func (w *ActorWorkflow) ResumeActor(ctx, ref, boot) (*Actor, bool, error) {
    // acquire actor lock, metrics defer — unchanged
    actor, tmpl, snapSrc, err := w.loadActorForResume(ctx, ref, boot)
    if actor.Status == RUNNING { return actor, false, nil }        // explicit no-op
    actor, err         = w.ensureVolumesCreated(ctx, actor, tmpl)
    actor, worker, err = w.ensureWorkerAssigned(ctx, actor, tmpl)
    err                = w.ensureVolumesAttached(ctx, actor, worker, tmpl)
    tele, err          = w.ensureAteletRestored(ctx, actor, worker, tmpl, snapSrc)
    actor, err         = w.finalizeRunning(ctx, ref)
    return actor, true, nil
}
```
Each `ensureX` 
- early-returns if its postcondition already holds, which is derived from persisted state only, replacing `IsComplete`
- validates the state machine edge internally, replacing `CheckPrerequisite`
- takes explicit typed inputs and returns what it produced.
- persists what it changed before returning, returning the stored copy so the optimistic lock version threads forward. No shared `*ResumeState`.

There is one behavioral change: worker ownership/eligibility validation for a re-entered resume now runs before volumes are attached (it used to run in `CallAteletRestoreStep`'s prerequisite, after attach), so a resume that is about to crash on a reassigned/ineligible worker no longer attaches volumes to it first.

- [X] Tests pass


cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -23,47 +23,103 @@ import (
...
+	var src resumeSnapshotSource
+	actor, actorTemplate, src, err = w.loadActorForResume(lockCtx, actorRef, boot)
+	if err != nil {
+		return nil, false, err
+	}
+	if wasRunning = actor.GetStatus() == ateapipb.Actor_STATUS_RUNNING; wasRunning {
+		return actor, false, nil
+	}
+	var created *ateapipb.Actor
+	if created, err = w.ensureVolumesCreated(lockCtx, actor, actorTemplate); err != nil {
```
juli4n, line 101: Do we need `created`, `assigned`, `running`? Why not just assign to `actor` directly?

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -83,58 +139,48 @@ func validateGoldenSnapshotScope(snapshot *ateapipb.ActorSnapshot) error {
...
-	actor, err := s.store.GetActor(ctx, input.ActorRef)
+	var src resumeSnapshotSource
+	actor, err := w.store.GetActor(ctx, actorRef)
 	if err != nil {
 		if errors.Is(err, store.ErrNotFound) {
-			return status.Errorf(codes.NotFound, "Actor %s not found", input.ActorRef)
+			return nil, nil, src, status.Errorf(codes.NotFound, "Actor %s not found", actorRef)
 		}
-		return fmt.Errorf("while getting actor from DB: %w", err)
+		return nil, nil, src, fmt.Errorf("while getting actor from DB: %w", err)
```
juli4n, line 154: If the actor is already running, can we short-circuit here and avoid fetching the snapshot? Any read we can save on the resume path counts.

cmd/ateapi/internal/controlapi/workflow_delete.go
```diff
@@ -24,138 +24,110 @@ import (
...
 	"google.golang.org/grpc/status"
-	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-// DeleteInput holds the immutable parameters requested by the client.
-type DeleteInput struct {
-	ActorRef resources.ActorRef
-}
+// DeleteActor executes the workflow to delete an actor. Idempotent.
+func (w *ActorWorkflow) DeleteActor(ctx context.Context, atespace, name string) (*ateapipb.Actor, error) {
```
juli4n: Nit: take an ActorRef as parameter here instead of atespace and name? That would be consistent with other workflows (suspend, pause).

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -27,212 +27,192 @@ import (
...
 	}
-	state.ActorTemplate = actorTemplate
-
-	return nil
+	if actor.GetStatus() == ateapipb.Actor_STATUS_PAUSED {
+		// Fully paused already: FinalizePaused commits PAUSED and the cleared
+		// worker assignment in a single update, so there is nothing left to do.
+		return actor, nil
+	}
+	var marked *ateapipb.Actor
```
juli4n, line 63: Same comment here (as in suspend workflow) about `marked` and `finalized`.

cmd/ateapi/internal/controlapi/workflow_resume.go
troychiu: The deferred telemetry handler at the start of the function captures actor to record metrics via `lifecycleOpAttrs(actor, ...)` on failure.

If we assign directly (e.g. `actor, err = w.ensureVolumesCreated(...)`), a failure returning `(nil, err)` would overwrite actor with nil.

cmd/ateapi/internal/controlapi/workflow_delete.go
troychiu: Updated! Thank you.

cmd/ateapi/internal/controlapi/workflow_resume.go
troychiu: Yes, I agree that it is worth it.

juli4n (approved):

Issue #802, opened 2026-08-07
Add a SystemInfo volume source to ActorTemplate for delivering per-actor generated files


thompsonmax:
## Summary

Add a "SystemInfo" volume source that delivers substrate-generated, per-actor files into an actor's containers, similar to [Kubernetes projected volumes](https://kubernetes.io/docs/concepts/storage/projected-volumes/) / downwardAPI. The volume is declared in spec.volumes with a list of data sources, each of which writes one or more files at configurable relative path(s), and is bind-mounted read-only wherever a container mounts it.

This replaces the hardcoded /run/ate/actor-id identity mount with an explicit and extensible API: templates opt in, choose the mount location and file layout, and future identity material (JWTs, certificates) can be added as new data sources without inventing new mount conventions.

## Why a volume, not env vars or baked files

Anything living in checkpointed process memory (env vars) or in the container image is frozen into snapshots. When a snapshot seeds a different actor than the one that created it (the template's shared snapshot, or CreateActorRequest.source_snapshot), every actor restored from it would observe the source actor's values if baked into the snapshot. For rotating material (tokens, certificates), a frozen copy goes stale across suspend/resume cycles even within a single actor. The metadata fields themselves are fixed for an actor's lifetime, only the cross-actor case affects them.

SystemInfo volume contents are instead regenerated by atelet on the host on every Run/Restore, whatever the boot source (cold boot, a shared template snapshot, or the actor's own prior snapshot), so the mounted files always carry the resumed actor's own, current values. Workloads should read them fresh rather than caching at startup.

### Preventing leakage of credentials read into process memory

Note that an actor may read this material into process memory, where it will be captured by snapshots that could later seed a different actor. We rely on these mechanisms to prevent Actor credentials provided from this volume type from leaking into other Actors seeded from a snapshot of a previous Actor:

- Short TTLs. Tokens and certificates expire on the order of their renewal interval. A snapshot is at earliest replayed minutes later and typically much older — by the time a captured credential could be extracted or resumed elsewhere, it has usually already expired. The workload re-reads from the SystemInfo volume at time of use, so expiry never breaks the legitimate holder.
[...]

PR #803, opened 2026-08-07
feat: add systemInfo volume source with actorIdentity data source

thompsonmax:
Part of #802 (first PR: the actorIdentity data source; does not close the issue).

## What changed and why

Adds a systemInfo volume source to ActorTemplate — a read-only volume whose files are generated by atelet on every Run/Restore, analogous to Kubernetes projected volumes. The initial data source, actorIdentity, writes the actor's own name to a configurable relative path:

```
spec:
  volumes:
  - name: system-info
    systemInfo:
      dataSources:
      # Part 1 (this PR): own-metadata projection, downwardAPI-style
      - actorMetadata:
          items:
          - field: name              # enum: name | atespace | uid
            path: actor-name
          - field: atespace
            path: atespace
          - field: uid
            path: actor-uid
  
  containers:
  - name: main
    image: app@sha256:...
    volumeMounts:
    - name: system-info
      mountPath: /run/ate  
```

Because the files are regenerated before the sandbox starts, they carry the resumed actor's own values regardless of what checkpointed state it boots from — the property the old hardcoded /run/ate identity mount provided, now as an explicit, extensible API that future data sources (identity JWTs, certificates — see #802) can slot into.

**Behavior change:** the automatic /run/ate/actor-id mount is removed; actors must opt in by declaring the volume (the e2e identity probe in this PR is the reference example).

### Reviewer notes:

- Over half the diff is vendored + generated code (cmd/atelet/internal/third_party/atomicwriter/, atelet.pb.go, zz_generated.deepcopy.go, the CRD manifest). The hand-written surface is ~700 lines.
- System-info volume roots live under a new ActorPath/system-info/ host dir, deliberately separate from durable-dir/: the micro-VM durable machinery snapshots everything under the durable-dir root, and generated identity files must never be captured into snapshots.
- Supports microVM as well as gVisor.
- The e2e identity suite exercises the new API end-to-end with unchanged probe binary and assertions. It runs in the kind-cluster CI job (not run locally).

## Checklist

- [x] Issue is linked above
- [x] Tests pass locally (go test ./...)
- [x] Root-gated tests pass if applicable (N/A — no root-gated packages touched)
- [x] Documentation updated if behavior changed (docs/api-guide.md: SystemInfo Volumes section with example)


pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -45,12 +45,48 @@ type ExternalVolumeTemplate struct {
 	StorageClassName string `json:"storageClassName"`
 }
 
+// ActorIdentityDataSource is a SystemInfo volume data source that writes the
+// actor's ID to a file.
+type ActorIdentityDataSource struct {
+	// Relative path from the root of the SystemInfo volume that the actor
+	// identity file should be written. Must be a clean relative Unix path:
```
juli4n: IIUC the "identity file" here in this PR is a file with just the actor name? That feels wrong as the name is only unique within the an atespace. Also, actors have UIDs (just like KRM resources). See https://github.com/agent-substrate/substrate/blob/main/docs/api-style-guide.md#21-identity-and-scope and https://github.com/agent-substrate/substrate/blob/main/docs/api-style-guide.md#63-uid

I can imagine two different approaches a) expose an API like k8s downward volumes, or b) a single "identity document" which is a json file and encodes atespace, name and uid.

thompsonmax: Good point. See https://github.com/agent-substrate/substrate/issues/802#issuecomment-5244238046 and updates to the spec in that issue. I updated the PR to support actor_name, atespace, and actor_uid downwardAPI-style like so:

```
volumes:
  - name: system-info
    systemInfo:
      dataSources:
      # Part 1 (this PR): own-metadata projection, downwardAPI-style
      - actorMetadata:
          items:
          - field: name              # enum: name | atespace | uid
            path: actor-name
          - field: atespace
            path: atespace
          - field: uid
            path: actor-uid
```

cmd/atelet/main.go
```diff
@@ -36,6 +36,7 @@ import (
 
 	"cloud.google.com/go/storage"
 	"github.com/agent-substrate/substrate/cmd/atelet/internal/ategcs"
+	"github.com/agent-substrate/substrate/cmd/atelet/internal/third_party/atomicwriter"
```
bowei: How are we structuring these third_party imports?

Do we expect to eventually rebase them with the changes on top or we just assume the code is forked and we never meet with the upstream copy ever again.

If we want to keep it clean, may want to factor it so that the patches to a given third_party are self-contained and clearly delineated.

thompsonmax: I think assuming fork is easiest. Added a README.md explaining where the code is copied from and why we had to fork it like this. Made it clear in the README that we'll try to keep the changes to upstream in our copy mechanical and only introduce behavioral changes in other files to make to easy to rebase if we need to in the future.

internal/proto/ateletpb/atelet.proto
```diff
@@ -141,14 +135,47 @@ message ExternalVolumeSource {
...
+// on every Run/Restore, so they carry the values of the actor actually being
+// started, whatever checkpointed state it boots from.
+message SystemInfoVolume {
+  repeated SystemInfoDataSource data_sources = 1;
+}
+
 message Volume {
   string name = 1;
 
-  VolumeType type = 2;
```
bowei, line 153: Any reason why we used to have this and now we are getting rid of it. Was this a mistake from before?

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -45,12 +45,91 @@ type ExternalVolumeTemplate struct {
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=255
+	// +kubebuilder:validation:XValidation:rule="!self.startsWith('/') && !self.endsWith('/') && !self.contains('//') && !self.contains(':') && !self.matches('[\\x00-\\x1f\\x7f]') && !self.matches('(^|/)[.][.]?(/|$)')",message="path must be a clean relative Unix path: must not start or end with '/', and contain no ':', '..', '.', '//', or control characters"
+	Path string `json:"path"`
+}
+
+// ActorMetadataDataSource is a SystemInfo volume data source that projects the
+// actor's identity fields (name, atespace, uid) to files, one per item —
+// analogous to the Kubernetes downwardAPI volume. Values are written raw with
+// no trailing newline, and are fixed for the actor's lifetime across
```
bowei, line 88: Are certs going to be fixed across life time?

cmd/atelet/internal/third_party/atomicwriter/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Kubelet's atomic writer, copied from the Kubernetes monorepo:
+
+- **Upstream source:** `k8s.io/kubernetes/pkg/volume/util/`
+  (`atomic_writer.go`, `atomic_writer_linux.go`, `atomic_writer_unsupported.go`,
+  `atomic_writer_test.go`)
+- **Delta last verified against:** [`kubernetes/kubernetes@52ba9013`](https://github.com/kubernetes/kubernetes/commit/52ba90138eb40cab0987dac73e05c838149bdd1c) (master, 2026-08-13)
+- **License:** Apache-2.0; the upstream copyright headers are retained in every file.
+
+It is copied, not imported, because upstream lives in the `k8s.io/kubernetes`
```
ahmedtd: This isn't really true --- the upstream atomicwriter is importable from https://pkg.go.dev/k8s.io/kubernetes/pkg/volume/util, but it pulls in a lot of k8s dependencies.

Also, this paragraph is very "Claude".  We don't need an essay to justify this.

cmd/atelet/main.go
```diff
@@ -1455,6 +1451,54 @@ func (s *AteomHerder) prepareOCIBundles(
 	return g.Wait()
 }
 
+// writeSystemInfoVolume populates the root directory of a system-info volume
+// with one file per projected item. It runs on every Run/Restore, before the
+// sandbox starts, so the files carry the values of the actor actually being
+// started, no matter what checkpointed state it boots from. Files are written
+// with the atomic writer so a concurrent reader can never observe a partial
+// write.
+func writeSystemInfoVolume(ctx context.Context, rootPath string, actorRef resources.ActorRef, actorUID string, si *ateletpb.SystemInfoVolume) error {
```
ahmedtd, line 1548: Can you drop a TODO or an issue that we need the content to update on the fly, not just on Run/Restore?

cmd/ateom-microvm/checkpoint.go
```diff
@@ -295,9 +295,10 @@ func (s *AteomService) teardownActor(ctx context.Context, id string, ra *running
 			_ = ra.chCmd.Process.Kill()
 			_, _ = ra.chCmd.Process.Wait()
 		}
-		// Kill the virtiofsds (after CH, their only client): the overlay RO lower's
-		// and, when the actor has durable-dir volumes, the writable share's.
-		for _, cmd := range []*exec.Cmd{ra.vfsdCmd, ra.durableVfsdCmd} {
+		// Kill the virtiofsds (after CH, their only client): the overlay RO
```
ahmedtd: Just "kill the virtiofsds after cloud-hypervisor"

internal/ateompath/ateompath.go
```diff
@@ -189,6 +177,26 @@ func DurableDirVolumeMountPoint(actorUID, volumeName string) string {
 	)
 }
 
+// SystemInfoVolumeRootsDir is the directory containing the per-volume root
+// directories of system-info volumes. It is deliberately separate from
```
ahmedtd: This seems like a non sequitur --- why would keeping the host-side folders in a different folder have any effect on whether or not they end up in snapshots?  It would be fine if we kept *all* of the root folders, for *all* of the volumes, regardless of their type, in the same folder together.

internal/proto/ateletpb/atelet.proto
ahmedtd: Yes, Dima left it in because I was originally going to send this PR a while ago.  But now Max has picked it up.  We don't need an additional field to indicate which fork of the oneof is set.

cmd/ateom-microvm/systeminfo.go
```diff
@@ -0,0 +1,117 @@
...
+// systemInfoMounts returns the OCI mounts that expose a container's
+// system-info volumes at the paths it declared, read-only. Each source is that
+// volume's directory inside the guest's system-info share, which the agent
+// mounts at sandbox creation.
+func systemInfoMounts(mounts []*ateompb.SystemInfoVolumeMount) []specs.Mount {
+	out := make([]specs.Mount, 0, len(mounts))
+	for _, m := range mounts {
+		out = append(out, specs.Mount{
+			Destination: m.GetMountPath(),
+			Source:      kata.GuestSystemInfoVolumeDir(m.GetVolumeName()),
```
ahmedtd, line 77: OK, I'm missing something here...

Earlier, we start up one virtiofsd to handle all SystemInfo volume mounts.  Here we tell CHV which host folder should back the mounted guest folder.

How does CHV know which virtiofsd to talk to to handle this particular mount.  Are we telling it that somewhere?  @BenTheElder

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -45,12 +45,91 @@ type ExternalVolumeTemplate struct {
...
+	// Relative path from the root of the SystemInfo volume at which the
+	// field's value is written. Must be a clean relative Unix path: must not
+	// start or end with '/', and contain no ':', '..', '.', '//', or control
+	// characters.
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=255
+	// +kubebuilder:validation:XValidation:rule="!self.startsWith('/') && !self.endsWith('/') && !self.contains('//') && !self.contains(':') && !self.matches('[\\x00-\\x1f\\x7f]') && !self.matches('(^|/)[.][.]?(/|$)')",message="path must be a clean relative Unix path: must not start or end with '/', and contain no ':', '..', '.', '//', or control characters"
+	Path string `json:"path"`
```
ahmedtd, line 82: High level question: Do we want to put each piece of actor info in a separate file?  Or should we have one JSON file that contains atespace, name, and uid (+ other stuff in the future).

pkg/api/v1alpha1/actortemplate_types.go
ahmedtd: No, but they are going to be handled by a different DataSource.

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -45,12 +45,91 @@ type ExternalVolumeTemplate struct {
...
+// suspend/resume/migration.
+type ActorMetadataDataSource struct {
+	// Items is the list of fields to project and the file path each is
+	// written to.
+	//
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=8
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, y.field == x.field))",message="items must not project the same field twice"
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, y.path == x.path))",message="items must not contain duplicate paths"
```
ahmedtd, line 98: We are going to want this validation to cover *all* the data sources, since they are all part of one mounted volume.  I don't know if we will be able to do that with the built-in CEL validation --- maybe it can be done with object-level CEL validation.

But anyways, this is going to migrate into the Ate API layer, so I guess we will just run validation in Go code there.

ahmedtd (commented):
Overall this looks good.  My main open question is about the virtiofsd handling on kata.

haiyanmeng:
@thompsonmax , who are the intended consumers of  `actorMetadata` introduced in this PR?

pkg/api/v1alpha1/actortemplate_types.go
thompsonmax: I opted for putting each piece of actor info in a separate file to roughly follow the pattern of downward API in k8s: https://kubernetes.io/docs/concepts/workloads/pods/downward-api/. I think this makes it easier to extend in the future with additional metadata and allows each actor to only pull in the fields they need.

cmd/atelet/main.go
thompsonmax: Sure, I avoided implementing it in this PR since actorMetadata doesn't really need it, since all the metadata it serves are fixed across the actor lifetime. Will add a TODO

thompsonmax:
Hi @haiyanmeng, the intended use case is similar to https://kubernetes.io/docs/concepts/workloads/pods/downward-api/ in k8s, basically actors that want to know their own identity for e.g. DNS self address, self-labeling when writing metrics, logs, or other telemetry, etc.

Immediate consumers will be anyone using `/run/ate/actor-id` today to get the actor identity. It is a useful first data source for proving out the SystemInfo volume functionality, since it is relatively simple to implement compared to other data sources we plan to support later (e.g. does not require live updating since these fields are fixed across the actor lifetime, unlike tokens, certs, trust bundles, etc),

internal/ateompath/ateompath.go
thompsonmax: So I think the mechanism of how this keeps the SystemInfo data out of snapshots differs between microVM and gVisor:

For MicroVM, we just tar everything in DurableDirVolumeMountsDir and stores that directly in the snapshot as the included data. So keeping SystemInfo out of that prevents it from being included at pause time.

for gVisor, we declare the volumes registered as durable in mount hints [here](https://github.com/agent-substrate/substrate/blob/c178f5b74ee2a0f8e535009223e0b2cd8b37d001/cmd/atelet/main.go#L1426), so gVisor knows to include them in the runsc checkpointing logic. Since we don't declare the SystemInfo volumes in the same way, they aren't incorporated into the snapshot.

My goal is to keep the SystemInfo data from showing up in snapshots (although it may still leak through process memory, but the SystemInfo volume should be the source of truth), since it represents data specific to the system the actor is running on and will become stale after a SUSPEND + RESUME.

cmd/atelet/internal/third_party/atomicwriter/README.md
thompsonmax: Clarified this is meant to avoid pulling in the full k8s dependency tree into atelet and made this less of an essay :)

pkg/api/v1alpha1/actortemplate_types.go
thompsonmax: Yeah good point. I think we can enumerate type pairs once we have 2 data sources, but that will get really ugly fast as we expand the number of sources.

Is there an issue tracking moving this validating into Ate API? I'd prefer to do that sooner rather than later to avoid having to add/maintain CEL hacks to support this validation across multiple data sources.

cmd/ateom-microvm/systeminfo.go
BenTheElder: Per volume seperation is just subdirs in the shares, so we create one virtiofs device per virtiofsd instance / "volume type", and then each mount is a subdir under that. The virtiofs devices have a tag / dedicated socket, that part we do specify to chv, but that just passes through each instance in bulk.

Setting up the subdir bind mounts guest side from each of the virtiofs share happens in overlay_linux.go, we do that through the kata guest agent currently.

thompsonmax: Tracing this out, how this works is:

1. [stageSystemInfoShare](https://github.com/agent-substrate/substrate/blob/b5f7dfac4a726d9642406d04c30225ec0fa3d79f/cmd/ateom-microvm/systeminfo.go#L100-L117) starts the third virtiofsd with its shared dir pointing at the host folder (ateompath.SystemInfoVolumeRootsDir(actorUID)), listening on a per-share vhost-user socket. [kata.SystemInfoVirtiofsdSocketPath](https://github.com/agent-substrate/substrate/blob/b5f7dfac4a726d9642406d04c30225ec0fa3d79f/cmd/ateom-microvm/internal/kata/restore.go#L41). The host path is only known to that virtiofsd.
2. [buildFsConfigs](https://github.com/agent-substrate/substrate/blob/b5f7dfac4a726d9642406d04c30225ec0fa3d79f/cmd/ateom-microvm/run.go#L679-L697) is where we tell cloud hypervisor which virtiofsd is which. The VM config (built [here](https://github.com/agent-substrate/substrate/blob/b5f7dfac4a726d9642406d04c30225ec0fa3d79f/cmd/ateom-microvm/run.go#L463) gets one FsConfig per share, and the system-info entry is {Tag: [SystemInfoFsTag](https://github.com/agent-substrate/substrate/blob/b5f7dfac4a726d9642406d04c30225ec0fa3d79f/cmd/ateom-microvm/internal/kata/overlay_linux.go#L61) ("ateSystemInfo"), Socket: ...}. cloud hypervisor creates a virtio-fs device bound to that socket, labeled with that tag. So the VM has up to three virtio-fs devices each wired to its own virtiofsd: kataShared , ateDurable, and ateSystemInfo.
3. When we create the sandbox, [CreateSandboxForActor](https://github.com/agent-substrate/substrate/blob/b5f7dfac4a726d9642406d04c30225ec0fa3d79f/cmd/ateom-microvm/internal/kata/overlay_linux.go#L239-L262) tells the kata agent to mount Storage{Source: SystemInfoFsTag, MountPoint: /run/ateom-system-info} [here](https://github.com/agent-substrate/substrate/blob/b5f7dfac4a726d9642406d04c30225ec0fa3d79f/cmd/ateom-microvm/internal/kata/overlay_linux.go#L254-L259). The guest kernel matches that mount to the device by tag. That links the guest folder and the right virtiofsd.
4. By the time we get to [systemInfoMounts](https://github.com/agent-substrate/substrate/blob/b5f7dfac4a726d9642406d04c30225ec0fa3d79f/cmd/ateom-microvm/systeminfo.go#L65-L81), each volume is just a subdirectory of the share mounted in step 3. These are ordinary guest-side bind mounts into the container.

BenTheElder: Also noting: The main reason we have two instances currently is because one of them is for read-only data [the container image rootfs], which we can avoid repeatedly querying the host for, so we configure it differently. 

We might not need a third instance, I haven't had time to read this PR in depth yet, just commenting on this question. https://github.com/agent-substrate/substrate/pull/846 also changes this and needs a fresh review (it was ~completely rewritten yesterday, I haven't caught up yet).

BenTheElder:
keep in mind: virtiofsd resume runs in `--find-paths` mode ... the paths cannot change between suspend/resume, or you'll break state in the guest memory snapshot. I'm not sure the atomic writer usage is safe here, at a quick skim.

BenTheElder:
That's not the right phrasing: I'm reasonably sure it's *not* safe, the timestamping bit ...

ahmedtd:
> That's not the right phrasing: I'm reasonably sure it's _not_ safe, the timestamping bit ...

Can you elaborate?  Are you saying that the content of the host folder needs to be exactly the same across snapshot/restore?

BenTheElder:
So on restore, it will find the guest's inodes again by matching the file paths. Any paths in the guest should be present at the same path again on restore, IIRC restore will just hard-fail otherwise.

AFAIK the contents of the files don't need to be identical, though that's not something we've been doing currently (we're only modifying things while the guest is live).

thompsonmax:
> So on restore, it will find the guest's inodes again by matching the file paths. Any paths in the guest should be present at the same path again on restore, IIRC restore will just hard-fail otherwise.
> 
> AFAIK the contents of the files don't need to be identical, though that's not something we've been doing currently (we're only modifying things while the guest is live).

Got it, thanks @BenTheElder for the heads up on that.

I reworked the approach to not use AtomicWriter and instead use a tmp file + rename to perform the file writes to keep it atomic per-file while maintaining the same file paths on the host once we restore that were present at suspend. I don't think we need the multi-file atomic updates that AtomicWriter provides actually, even with later sources we plan to add , we'll only need to update single files atomically (tokens, certs, trust bundles). Added regression testing in both unit + e2e testing.

ahmedtd:
> > So on restore, it will find the guest's inodes again by matching the file paths. Any paths in the guest should be present at the same path again on restore, IIRC restore will just hard-fail otherwise.
> > AFAIK the contents of the files don't need to be identical, though that's not something we've been doing currently (we're only modifying things while the guest is live).
> 
> Got it, thanks Benjamin Elder (@BenTheElder) for the heads up on that.
> 
> I reworked the approach to not use AtomicWriter and instead use a tmp file + rename to perform the file writes to keep it atomic per-file while maintaining the same file paths on the host once we restore that were present at suspend. I don't think we need the multi-file atomic updates that AtomicWriter provides actually, even with later sources we plan to add , we'll only need to update single files atomically (tokens, certs, trust bundles). Added regression testing in both unit + e2e testing.

Certificates and private keys require cross-file atomic updates if we support writing the private key and certificate to separate files.

ahmedtd:
> So on restore, it will find the guest's inodes again by matching the file paths. Any paths in the guest should be present at the same path again on restore, IIRC restore will just hard-fail otherwise.
> 
> AFAIK the contents of the files don't need to be identical, though that's not something we've been doing currently (we're only modifying things while the guest is live).

Can we test this?  This doesn't match my mental model of how a virtiofs client should work --- it always needs to be prepared for a host-side file delete, and has no mechanism to block it.

thompsonmax:
> Certificates and private keys require cross-file atomic updates if we support writing the private key and certificate to separate files.

Ah good point. I think for now the atomic per-file update will suffice (I'd like to avoid making this PR too much more complex). When we introduce support for certificates, I believe we can write them to a temp directory and then use the `renameat2` syscall with RENAME_EXCHANGE set to atomically swap the directory.

ahmedtd:
> > Certificates and private keys require cross-file atomic updates if we support writing the private key and certificate to separate files.
> 
> Ah good point. I think for now the atomic per-file update will suffice (I'd like to avoid making this PR too much more complex). When we introduce support for certificates, I believe we can write them to a temp directory and then use the `renameat2` syscall with RENAME_EXCHANGE set to atomically swap the directory.

It's fine for now, but atomicwriter grew its complexity by handling all the nasty edge cases.  It didn't start out that complex.

ahmedtd:
> > So on restore, it will find the guest's inodes again by matching the file paths. Any paths in the guest should be present at the same path again on restore, IIRC restore will just hard-fail otherwise.
> > AFAIK the contents of the files don't need to be identical, though that's not something we've been doing currently (we're only modifying things while the guest is live).
> 
> Can we test this? This doesn't match my mental model of how a virtiofs client should work --- it always needs to be prepared for a host-side file delete, and has no mechanism to block it.

Ugh, it sounds like virtiofsd is just very brittle for snapshot/restore: https://gitlab.com/virtio-fs/virtiofsd/-/blob/main/doc/migration.md?ref_type=heads

We should probably run it with `--migration-on-error=guest-error`.  That way, the restore won't fail, but the guest program will get I/O errors if it tries to operate on a file that disappeared during snapshot/restore.

thompsonmax:
> Can we test this? This doesn't match my mental model of how a virtiofs client should work --- it always needs to be prepared for a host-side file delete, and has no mechanism to block it.

> Ugh, it sounds like virtiofsd is just very brittle for snapshot/restore: https://gitlab.com/virtio-fs/virtiofsd/-/blob/main/doc/migration.md?ref_type=heads

I added another e2e test to confirm that suspend + resume with SystemInfo works with microVM and also created a throwaway draft PR https://github.com/agent-substrate/substrate/pull/1020 to run the test with the old implementation to validate it fails in that case.

> We should probably run it with --migration-on-error=guest-error. That way, the restore won't fail, but the guest program will get I/O errors if it tries to operate on a file that disappeared during snapshot/restore.

IIUC, the case we really need to worry about with this is when we need to rotate mounted credentials, which may delete + create files and overlap in time with a suspend + resume. Should we defer setting that option until we need to handle that or is there a case I'm missing where it is helpful for this PR?

thompsonmax:
Just a heads up, it seems the E2E tests are not running in the CI for some reason. But I manually ran the new E2E test I added to validate suspend + resume with a SystemInfo volume on microVM. It fails with the old implementation using k8s atomicwriter, which doesn't preserve file paths across suspend + resume, but passes with the new implementation which does preserve these. https://gist.github.com/thompsonmax/7cbafda5b3df91de13456eb295080fbc

cmd/ateom-microvm/systeminfo.go
thompsonmax: Just a heads up, I ended up putting everything on the unified share introduced in https://github.com/agent-substrate/substrate/pull/846) rather than using a separate instance.

BenTheElder:
> Just a heads up, it seems the E2E tests are not running in the CI for some reason. But I manually ran the new E2E test I added to validate suspend + resume with a SystemInfo volume on microVM. It fails with the old implementation using k8s atomicwriter, which doesn't preserve file paths across suspend + resume, but passes with the new implementation which does preserve these. https://gist.github.com/thompsonmax/7cbafda5b3df91de13456eb295080fbc


Should be fixed now -- #1056 

Also worth noting: 1056 skipped the identity test for uVM (unimplemented) assuming this PR supersedes it anyhow.

cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -67,6 +68,20 @@ func GuestCSIVolumeDir(volumeName string) string {
...
+// volume's contents, i.e. the read-only bind source for that volume's
+// container mount points.
+func GuestSystemInfoVolumeDir(volumeName string) string {
+	return guestSharedDir + "system-info/" + volumeName
+}
+
+// GuestSystemInfoVolumeDir is the in-guest path holding one system-info
+// volume's contents, i.e. the bind source for that volume's container mount
+// points.
+func GuestSystemInfoVolumeDir(volumeName string) string {
```
BenTheElder: 🤖 **blocking** 🔴 – This is a second copy of the function declared six lines above, so the package does not compile — `GuestSystemInfoVolumeDir` redeclared. It also references `guestSystemInfoDir`, which no longer exists: the const block now defines only `guestSharedDir`, so that is a second error.

It reads like a merge resolution that kept both sides. This copy is the retired one — it points at the standalone `/run/ateom-system-info` mount from the third-virtiofsd design. The copy above returns `guestSharedDir + "system-info/" + volumeName`, which is what the const block's own comment documents and what `stageSystemInfoVolumes` binds. Deleting this one is the fix.

Worth a second look at the rest of that merge while you are in there — this is the kind of leftover that comes in pairs.

cmd/ateom-microvm/systeminfo.go
BenTheElder: thanks!

BenTheElder (dismissed):

PR #805, opened 2026-08-08
manifests: stop pinning the control plane and DNS to IPv4

ygao-g:
Two of the three Services on the actor request path were reachable over IPv4 only, for two different reasons.

**`ate-api-server` pinned its listener to the IPv4 wildcard.** The Deployment passed `--grpc-listen-addr=0.0.0.0:443`, overriding the `":443"` that `cmd/ateapi/main.go` already defaults to. The bare form binds every family the host has; the IPv4 wildcard binds one. The override dates to the initial commit and `git log -S` turns up no rationale for it, so this drops the flag rather than rewriting it to `[::]:443`.

**The `dns` and `api` Services carried no `ipFamilyPolicy`,** which the API server defaults to `SingleStack` — an IPv4 ClusterIP and nothing else on a dual-stack cluster. Both now set `PreferDualStack`, not `RequireDualStack`, since `Require` fails Service creation outright on the single-stack clusters CI runs. `api` is headless, so there the policy governs EndpointSlice families and therefore the per-pod DNS records that client-side gRPC load balancing depends on, not a VIP.

The `atenet-router` Service needs the same treatment and is handled separately, alongside the Envoy listener change it is useless without.

## Testing

No unit tests here, deliberately. Neither invariant is visible to a unit test of the affected binary — the binary is fine either way and only the deployed cluster breaks. An earlier revision asserted them by parsing the shipped YAML in a test-only `internal/installmanifests` package; that was dropped at review request in favour of exercising the real thing.

Coverage comes from running the e2e suite against a dual-stack cluster, which is a separate change: it adds an `ip-family` axis to the `e2e-test-matrix` workflow and a `TestDataPathServicesAreDualStack` that reads the cluster'''s families off the node pod CIDRs and **fails**, rather than skips, when a data-path Service has fewer families than the cluster can give it.

Locally on this branch: `go build ./...`, `go vet`, and `hack/verify/{boilerplate,gofmt,go-modules}.sh` pass.

## Compatibility

No behavior change on an IPv4-only cluster: `PreferDualStack` degrades silently, and `":443"` and `"0.0.0.0:443"` bind the same socket when the host has no IPv6.

### DIFF
--- a/manifests/ate-install/ate-api-server.yaml
+++ b/manifests/ate-install/ate-api-server.yaml
@@ -92,7 +92,6 @@ spec:
       - name: ate-api-server
         image: ko://github.com/agent-substrate/substrate/cmd/ateapi
         args:
-          - "--grpc-listen-addr=0.0.0.0:443"
           - "--grpc-server-cred-bundle=/run/servicedns.podcert.ate.dev/credential-bundle.pem"
           - --redis-cluster-address=@env
           - --redis-ca-certs=/etc/valkey-ca/ca.crt
@@ -243,6 +242,8 @@ spec:
   # Headless (i.e. `clusterIP: None`) services make DNS return one A record per
   # ready pod so gRPC clients can load balance client-side.
   clusterIP: None
+  # Prefer, not Require: Require fails Service creation on a single-stack cluster.
+  ipFamilyPolicy: PreferDualStack
   selector:
     app: ate-api-server
   ports:
--- a/manifests/ate-install/atenet-dns.yaml
+++ b/manifests/ate-install/atenet-dns.yaml
@@ -167,6 +167,8 @@ spec:
   selector:
     app: dns
   type: ClusterIP
+  # Prefer, not Require: Require fails Service creation on a single-stack cluster.
+  ipFamilyPolicy: PreferDualStack
   ports:
   - name: dns
     port: 53


manifests/ate-install/ate-api-server.yaml
```diff
@@ -92,7 +92,10 @@ spec:
       - name: ate-api-server
         image: ko://github.com/agent-substrate/substrate/cmd/ateapi
         args:
-          - "--grpc-listen-addr=0.0.0.0:443"
+          # No --grpc-listen-addr: cmd/ateapi/main.go already defaults to ":443",
```
BenTheElder: we can just remove it, instead of adding a 4 line comment here. the commit history will point back to this PR

internal/installmanifests/ipfamily_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	sigsyaml "sigs.k8s.io/yaml"
+)
+
+// installDir is the manifest tree hack/install-ate.sh applies, relative to this
+// package.
+const installDir = "../../manifests/ate-install"
```
BenTheElder: I don't think we need a regression test for the manifests. Better to e2e test with IPv6 Regularly.

manifests/ate-install/ate-api-server.yaml
ygao-g: Done — flag line just removed, no comment.

internal/installmanifests/ipfamily_test.go
ygao-g: Done — dropped `internal/installmanifests` entirely (both files).

ygao-g: Agreed on the e2e point, and that half is already written — it just needs to go up as its own PR, which I'll do next.

It adds an `ip-family` axis to the `e2e-test-matrix` workflow (`main` has no such axis today, so nothing runs against IPv6 regularly right now) plus `TestDataPathServicesAreDualStack` in the networking suite. That test reads the cluster's families off the node pod CIDRs and **fails** rather than skips when a data-path Service has fewer families than the cluster can give it — the rest of the IPv6 suite gates itself on the router Service having two ClusterIPs, so a SingleStack regression would make those tests skip and the suite would still report green.

With `installmanifests` gone that PR is the only thing guarding this one, so I'll prioritise it over the rest of the IPv6 queue.

bowei:
looks ok

bowei:
Commit message could be shorted a little, seems to have agent text in it.

bowei (approved):

ygao-g:
> Commit message could be shorted a little, seems to have agent text in it.

Done, thanks!

PR #806, opened 2026-08-08
atenet-router: raise actor-cluster circuit breakers

chuangw6:
The actor cluster configured no circuit breakers, leaving Envoy's 1,024 default — a low ceiling for the one proxy carrying every actor's ingress. Under overload it surfaces as storms of instant 503s once concurrency crosses 1,024.

- Actor cluster: `max_connections`, `max_pending_requests` and `max_requests` rise to 20,000, kept under the ~28k ephemeral-port budget so overload trips a counted breaker instead of the kernel's opaque `EADDRNOTAVAIL`. `max_retries` keeps Envoy's default — nothing on this cluster retries.
- ext_proc cluster: the `--extproc-max-requests` ceiling now covers `max_pending_requests` as well as `max_requests`.

Tested with `go test ./cmd/atenet/internal/router/`.

### DIFF
--- a/cmd/atenet/internal/router/cmd.go
+++ b/cmd/atenet/internal/router/cmd.go
@@ -79,7 +79,7 @@ func NewRouterCmd() *cobra.Command {
 	cmd.Flags().DurationVar(&cfg.ParkedRequest.RetryInterval, "parked-request-retry-interval", ingress.DefaultParkedRequestRetryInterval, "Delay before a parked request's first resume retry")
 	cmd.Flags().Float64Var(&cfg.ParkedRequest.RetryFactor, "parked-request-retry-factor", ingress.DefaultParkedRequestRetryFactor, "Multiplier applied to the retry delay after each attempt; must be >= 1")
 	cmd.Flags().Float64Var(&cfg.ParkedRequest.RetryJitter, "parked-request-retry-jitter", ingress.DefaultParkedRequestRetryJitter, "Random fraction in [0, 1) added to each retry delay to de-synchronize parked requests")
-	cmd.Flags().IntVar(&cfg.ExtProcMaxRequests, "extproc-max-requests", 0, "Circuit-breaker max_requests for Envoy's ext_proc cluster; 0 (the default) derives it as twice --parked-request-max (minimum 1024). Explicit values must be >= --parked-request-max: every parked request holds one slot for its full wait, and the excess is fast-path headroom")
+	cmd.Flags().IntVar(&cfg.ExtProcMaxRequests, "extproc-max-requests", 0, "Circuit-breaker ceiling for Envoy's ext_proc cluster, applied to max_requests and max_pending_requests alike; 0 (the default) derives it as twice --parked-request-max (minimum 1024). Explicit values must be >= --parked-request-max: every parked request holds one slot for its full wait, and the excess is fast-path headroom")
 	// Graceful shutdown knobs. The router sits behind a Service, so
 	// route-drain window is needed: after SIGTERM the readiness flip
 	// must propagate to the Service endpoints before the drain starts.
--- a/cmd/atenet/internal/router/xds.go
+++ b/cmd/atenet/internal/router/xds.go
@@ -111,10 +111,11 @@ const (
 // otherwise Envoy abandons a parked request (500) long before the router does.
 const defaultExtProcMessageTimeout = 5 * time.Second
 
-// defaultExtProcMaxRequests is the circuit-breaker max_requests set on the
-// ext_proc cluster: ingress.DefaultParkedRequestMax plus equal fast-path headroom, so a
-// full parking lot cannot starve the millisecond-scale header exchanges of
-// requests to already-running actors. See buildCluster.
+// defaultExtProcMaxRequests is the circuit-breaker ceiling set on the ext_proc
+// cluster, applied to both max_requests and max_pending_requests:
+// ingress.DefaultParkedRequestMax plus equal fast-path headroom, so a full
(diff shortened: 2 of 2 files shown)


manifests/ate-install/atenet-router.yaml
```diff
@@ -158,6 +163,15 @@ spec:
...
         # terminationGracePeriodSeconds alongside it.
         # - "--route-timeout=5m"
+        resources:
+          # Small requests so the pod schedules anywhere (including CI kind
+          # nodes); the limits are the real cap.
+          requests:
+            cpu: 250m
+            memory: 256Mi
+          limits:
+            cpu: "8"
```
bowei: You should put a comment why the limit needs to be set:

Otherwise Envoy configures too many event loops on a large machine?
Does this affect scheduling on a small machine?

manifests/ate-install/atenet-router.yaml
```diff
@@ -217,13 +231,22 @@ spec:
         - name: "drain-signal"
           mountPath: "/var/run/atenet"
       - name: envoy
+        # Not v1.39+: it measured 20-38% lower sustained throughput on this
```
bowei: remove this comment

manifests/ate-install/atenet-router.yaml
```diff
@@ -217,13 +231,22 @@ spec:
...
         command:
         - "/usr/local/bin/envoy"
         - "-c"
         - "/etc/envoy/envoy.yaml"
         - "--component-log-level"
         - "upstream:debug,router:debug,ext_proc:debug"
+        # Envoy sizes worker threads from the node's CPU count unless told
+        # otherwise; on a large node that is dozens of event loops.
+        - "--concurrency"
+        - "8"
```
bowei: But what if you install on a smaller machine? Are we ok with setting this to 8 overall or should this be tuned?

manifests/ate-install/atenet-router.yaml
```diff
@@ -217,13 +231,22 @@ spec:
...
+        # Observability: mutex-contention counters on the admin /contention
+        # endpoint. Costs one atomic per contended acquisition.
+        - "--enable-mutex-tracing"
```
bowei: What does this give us in production

bowei (commented):
I see you added new limits and constants -- but how do we expect it to behave for people who install it on a smaller machine than you are giving space for?

bowei:
* Resolve the conflicts
* Are we ok with setting flat 8 core in the specs for now. If users deploy on a smaller machine, is this going to be a problem?

manifests/ate-install/atenet-router.yaml
chuangw6: TL;DR: We probably don't need to set this after envoy v1.37 


This is required on v1.30: without it Envoy runs one worker [per hardware thread](https://www.envoyproxy.io/docs/envoy/v1.30.0/operations/cli#cmdoption-concurrency) — 88 on 88-core nodes — ignoring the CPU limit (cgroup-aware defaults only arrived in Envoy 1.37 ([release note](https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.37/v1.37.0#incompatible-behavior-changes))). Extra workers aren't free: they burn more CPU per request (measured +19% at 8 threads vs 2, +38% at 16), and since connection pools are per-worker, more workers means more pools, so fewer connection reuses and more idle connections held open. 

Matt Klein's [threading-model post](https://medium.com/envoyproxy/envoy-threading-model-a8d44b922310) calls this out directly:
> it is actually quite important to tune the --concurrency option. Having more workers than is needed will waste memory, create more idle connections, and lead to a lower connection pool hit rate.

bowei:
Ok, let's remove the config that won't be needed in version 37 and we should be good with this PR

manifests/ate-install/atenet-router.yaml
chuangw6: It's a diagnostic flag: it feeds the admin /contention endpoint so lock contention can be ruled in or out without a redeploy. It already did its job during the capacity benchmarks — it ruled contention out as the culprit for the per-thread CPU tax (~50 ms of blocked-acquisition wait per 15 s window across all threads, i.e. noise; the tax turned out to be kernel scheduling). Since contention was never a factor on this workload, dropped the flag from the production manifest.

manifests/ate-install/atenet-router.yaml
chuangw6: Done.

manifests/ate-install/atenet-router.yaml
chuangw6: Added a comment. Small machines are unaffected: scheduling considers only requests (250m here) — a limit above node capacity simply never binds. The event-loop concern was Envoy's --concurrency default, which is now dropped along with the flag discussion below.

bowei (dismissed):

bowei (dismissed):

cmd/atenet/internal/router/xds_test.go
```diff
@@ -564,6 +564,59 @@ func TestXdsServer_ExtProcCircuitBreaker(t *testing.T) {
 			t.Errorf("max_requests after SetExtProcMaxRequests(0) = %d, want default %d", got, defaultExtProcMaxRequests)
 		}
 	})
+
```
bowei: remove these test -- this is overkill

chuangw6: Done — removed both added tests; the change ships with no test diff.

bowei:
Get the tests to pass

bowei (approved):

Issue #665, opened 2026-07-31
Measure atenet-router request capacity


chuangw6:
Build a repeatable open-loop benchmark for one atenet-router and find the
request rate where p95 latency crosses a certain latency bound e.g. 500 ms, so we can size replicas.

chuangw6:
closing because it has been fixed by https://github.com/agent-substrate/substrate/pull/998

PR #807, opened 2026-08-08
routercap: capacity benchmark harness for atenet-router

chuangw6:
Part of #665.

A one-command capacity sweep for `atenet-router`: each arm sets one Envoy CPU size and walks a rising load ladder against a dedicated cluster; every run produces a self-contained `report.html`.

- Open-loop pacing with latency timed from *scheduled* send, so coordinated omission is impossible and a struggling generator cannot flatter the router.
- Measures throughput/latency, a per-hop breakdown of the mean request (generator / Envoy / ext_proc sidecar / worker), per-thread Envoy CPU, and container CPU/memory on cAdvisor-aligned windows.
- Six guards distinguish "the rig ran out" from "the router ran out" and mark arms rig-limited.

Methodology and reproduction steps are in `benchmarking/routercap/README.md`. Findings from the first full study land in https://github.com/agent-substrate/substrate/pull/808.

### DIFF
--- a/benchmarking/routercap/README.md
+++ b/benchmarking/routercap/README.md
@@ -1,0 +1,326 @@
+# routercap
+
+Measures how much load one `atenet-router` pod absorbs before it stops
+absorbing, and what the failure looks like when it does. The output is a time
+series per run — `runs/<timestamp>/report.html` — and the findings live in
+[RESULTS.md](RESULTS.md).
+
+## Why this test exists
+
+Every actor request goes through `atenet-router`, and nobody had measured
+what one pod can take. Without that figure there is no basis for the replica
+count, for the CPU limits in
+[atenet-router.yaml](../../manifests/ate-install/atenet-router.yaml), or for
+an alert threshold — and no way to tell a regression from a busier week.
+
+A single "N QPS" number would not settle it either. What matters is the shape
+at the edge: whether throughput plateaus or collapses, whether latency
+degrades or cliffs, and whether the binding constraint is CPU at all. So the
+harness walks a rising load ladder and records the whole curve, at several
+Envoy CPU sizes.
+
+### Why not locust
+
+Locust-style load tests are closed-loop: each simulated user sends a request,
+waits for the reply, then sends the next. The moment the router slows down,
+the users slow down with it — offered load sags exactly when the system is
+most interesting, and the latency samples miss the worst moments because
+fewer requests were in flight during them. The literature calls this
+coordinated omission. A closed-loop test of this router would report a
+flattering curve that bends where the client throttled, not where the router
+failed.
+
+This harness is open-loop: a pacer fires requests on a fixed schedule whether
+or not earlier ones have returned, so offered load stays the independent
+variable all the way through a collapse. The repo's boomer/locust rig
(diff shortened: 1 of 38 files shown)


benchmarking/routercap/README.md
```diff
@@ -0,0 +1,326 @@
...
+
+A single "N QPS" number would not settle it either. What matters is the shape
+at the edge: whether throughput plateaus or collapses, whether latency
+degrades or cliffs, and whether the binding constraint is CPU at all. So the
+harness walks a rising load ladder and records the whole curve, at several
+Envoy CPU sizes.
+
+### Why not locust
+
+Locust-style load tests are closed-loop: each simulated user sends a request,
```
maxsmythe, line 24: I am not sure this is 100% accurate, especially with Boomer.

Is there anything preventing a Locust user from spinning off a goroutine per-iteration? That would ensure any outbound requests are not blocking, removing the "closed-loop" effect.

Not implying this means all benchmarks *must* use Locust/boomer, but we should be accurate in our reasoning.

Reasons for using Locust/Boomer (may not be comprehensive):
* Unified entry point for running benchmark tests and scraping results, simplifying any automation of benchmarking testing.
   * This may or may not have an impact on how easy it is to bootstrap necessary infrastructure (procuring a cluster to run the test on, spinning up Substrate with the necessary config, etc.), since the current work around that is built around Locust but is not actually part of Locust.
* Built-in protocol for aggregating client-level observations across multiple worker pods, allowing scaling traffic beyond one machine and gathering things like percentiles that require non-trivial aggregation.
* Ability to easily customize traffic shape, especially across distributed workers (e.g. the laddering you do)

Reasons for not using Locust (again, may not be comprehensive):
* Needing to work around framework, making the test brittle or overly complex
* Lack of need for any of the above benefits

Is this benchmark something we expect to run regularly?

Also, I have not had a chance to look at this PR as it is very large (12k lines of code).

It appears that large chunks of this are dedicated to data visualization (e.g. charts.py), bootstrapping infrastructure, and scraping data sources.

Data vis -- any chance this can be delegated to an external tool? Generally speaking, benchmarks (especially if they are regularly run), export data in a standardized format that let users choose how they slice, dice and visualize the data. Particularly if these are regularly-running benchmarks.

Bootstrapping infrastructure -- we should try to avoid idiosyncratic bootstrapping where possible. I have not dug too deeply into this PR, so some of this could be justified, but that would be the exception, not the rule.

Scraping data sources -- The ability to scrape different data points and gather statistics on them is super useful. We should commoditize and share these capabilities as much as possible -- example, I would love to sample router stats as part of Locust.

In general, it seems like benchmarking/routercap could use purpose-built sub-packaging. It is hard to tell what class/function is defined where. More modules can help make that clearer via imports. It is also a step towards making the stats scraping shareable.

maxsmythe (changes requested):
Thank you for the PR! Have some generalized thoughts/feedback

bowei:
Can you ping me when you've incorporated Max's suggestions?

benchmarking/routercap/README.md
chuangw6: Thanks Max for sharing your thoughts. I am closing this PR in favor of PR 998 per offline discussions. cc @bowei

PR #808, opened 2026-08-08
routercap: capacity results (Envoy v1.30, 200 workers)

### LINKED ISSUE #665: Measure atenet-router request capacity

chuangw6:
Fixes #665.

Findings from the first routercap study. Companion to #807 (the harness); this PR carries only the write-up and two report snapshots — review #807 for the tool, this one for the numbers. Best merged after #807.

| Topology | Sustained QPS |
|---|---|
| 1 × 4c | 9,000 |
| 1 × 8c | 11,000 |
| 1 × 16c | 13,000 |
| 2 × 4c | 16,000 |

- The wall is per-thread wake-up overhead, not CPU: wide arms die majority-idle, and capacity follows thread count, not the CPU limit.
- One overloaded thread can fail the pod while container CPU reads 29% — monitor per-thread CPU, not the container average.
- Scale out, not up: two 4-core replicas beat one 8-core pod 16k vs 11k on the same cores. Envoy v1.39 regressed 20-38% on every arm and stays off.

`RESULTS.md` carries the analysis; the two `report.html` files are self-contained chart snapshots (raw run data intentionally stays out of the repo).


chuangw6:
closing in favor of PR 998

Issue #778, opened 2026-08-06
Investigate replacing credential broker mTLS with per-worker UDS isolation


EItanya:
## Context

PR #708 authenticates atunnel to the node-local atelet credential broker with mTLS over a Unix domain socket. This is necessary in the current layout because every root-running ateom shares the hostPath containing the broker socket; permissions on one shared socket do not distinguish workers.

Review discussion: https://github.com/agent-substrate/substrate/pull/708#discussion_r3717247245

Investigate whether filesystem isolation can provide the same worker-to-atelet binding with less protocol machinery.

## Questions

- Can atelet create a separate socket directory and listener for each worker?
- Can each worker Pod mount only its own socket directory?
- What ownership and mode should protect the directory and socket when ateom runs as root?
- How are sockets created, removed, and recovered across worker replacement, Pod UID reuse, atelet restart, and failed activation?
- How do we prevent path traversal, symlink, stale-socket, and cross-worker access attacks?
- Can atelet securely derive the worker identity from the selected listener without trusting request metadata?
- If filesystem isolation is sufficient, can TLS be removed, or should it remain as defense in depth?

## Acceptance criteria

- Document the trust boundary and attacker model, including a compromised actor and a compromised sibling worker on the same node.
- Compare the current shared-socket mTLS design with per-worker filesystem-isolated sockets.
- Specify socket naming, mounts, permissions, identity binding, lifecycle, and cleanup.
- Confirm the proposed design fails closed during stale assignment and restart races.
- Recommend whether to keep mTLS, replace it, or combine both.
- Do not remove mTLS until equivalent worker isolation is demonstrated by tests.

Issue #761, opened 2026-08-05
Actor instrumented metrics might merge and inflate


git286:
Any workload that exports OTel metrics loses per-actor identity when it moves to Substrate. Counters come out wrong, and nothing errors.

Two things cause it.

**Everything computed at startup is shared.** The golden snapshot is taken after init, so an OTel `Resource` built in `main()` is identical for every actor of the template. The usual escapes don't work either: `os.Hostname()` is hardcoded to `runsc` for every sandbox (`cmd/atelet/oci.go:271`), and ActorTemplate doesn't support `valueFrom.fieldRef`, so you can't inject a pod name (`docs/api-guide.md:101`).

**Collector enrichment attributes to the worker pod.** Actor egress is masqueraded behind the pod IP in both ateom runtimes (`cmd/ateom-microvm/net.go:22`), so `k8sattributes` stamps the *host worker's* `k8s.pod.name` / `k8s.node.name`. That merges every co-located actor into one series, and changes an actor's series identity every time it migrates. Every actor also shares one constant interior IP (`169.254.17.2`), so no IP-based processor can tell them apart even in principle.

Note that push doesn't get us out of this, it's what gets us into it. Actors can't be scraped, and the pull model is what hands you an `instance` label for free, derived from the target address rather than from anything the workload says. With OTLP push, identity has to come from the payload, and the payload is the part that's frozen.

The result is one series with many writers. OTLP counters default to cumulative, so the merged series goes up and down, `rate()`/`increase()` treat every decrease as a reset, and totals inflate. Gauges show a random actor. The dashboards look plausible, which is the worst part and the pod/node labels make it read like ordinary pod churn, so you debug the wrong thing.

Adding `actor_uid` as a metric label isn't the fix; it's unbounded and churns on every actor create.

Things I think we need to decide:

- what identifies an actor's metrics on the wire vs. what survives into storage
- whether Substrate stamps it out-of-process (the identity dir seems like the natural place, since it's outside the snapshot) or we leave it to actor authors
- interim guidance for actor authors either way

Related: #503 (telemetry continuity), #450 (lifecycle hooks). PR #750's demo hits this: its `actorName()` returns `runsc` for every actor.

git286:
cc @JeffLuoo

krisztianfekete:
Thanks @git286, this is something I've been tracking for a while but never got around writing a proposal.

Couple of notes/questions:
> That merges every co-located actor into one series ...

[...]

PR #809, opened 2026-08-08
ateom: export telemetry through an atelet unix-socket OTLP relay

chw120:
ateom runs inside the worker pod that hosts the actor, and exported OTLP straight to the collector over the pod's network. This adds a node-local relay: ateom pushes OTLP/gRPC over a unix socket that atelet serves and forwards to the collector, so a worker pod needs no network path of its own to export spans and metrics.

Four things motivate it:

  - Blast radius. The pod runs untrusted agent code, so allowing it egress to the collector makes the collector reachable to anything that escapes the sandbox. A unix socket cannot leave the node.
  - Connection count. Worker pods are heavily oversubscribed; N ateoms per node each held their own collector connection. They collapse into atelet's single per-node one.
  - Interference. ateom transparently redirects actor egress to its own atunnel listener, and its own outbound traffic has to stay clear of the rules it installs. A unix socket is not IP traffic.
  - Shutdown loss. Teardown frees the actor's network and then the pod goes away, which is when the spans describing teardown are still queued in the batch processor. atelet outlives the worker pod.

The relay forwards the OTLP request verbatim rather than decoding and re-exporting, so each ateom's own resource (service.name, service.instance.id) survives instead of being absorbed into atelet's.

It is best-effort: an ateom that finds no socket at startup logs it and exports directly to OTEL_EXPORTER_OTLP_ENDPOINT as before, so this is a no-op for a cluster running an older atelet. atelet likewise declines to serve a relay when no collector is configured, since it would accept spans only to drop them. Both halves stay off with --otlp-relay-socket="".

The socket lives in ateompath.BasePath, the hostPath already mounted at the same path into atelet and into every ateom pod, so no new volume or controller change is needed.

Also includes:
  - End-to-end tests covering the full serverboot-to-collector path.
  - Observability documentation updates for Jaeger tracing.

- [x] Tests pass


chw120:
@BenTheElder FYI, this is the PR for collecting ateom traces. Please review it when you have time. Thank you.

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,254 @@
...
+// Serve listens on the relay socket and blocks until the server stops. Designed
+// to be `go`-launched; it returns an error only if the socket cannot be opened
+// or serving fails.
+func (s *Server) Serve(ctx context.Context) error {
+	if err := os.MkdirAll(filepath.Dir(s.sockPath), 0o755); err != nil {
+		return fmt.Errorf("while creating the OTLP relay socket directory: %w", err)
+	}
+	// A socket left behind by a previous atelet would make Listen fail with
+	// EADDRINUSE even though nothing holds it.
+	if err := os.RemoveAll(s.sockPath); err != nil {
```
BenTheElder: 🤖 **should-fix** 🟡 – `RemoveAll` is a lot more than this needs, and the blast radius on a misconfigured path is the node's whole ateom state. The comment explains the goal as clearing a stale socket so `Listen` does not hit `EADDRINUSE` — `os.Remove` does exactly that and fails on a directory, which is the right answer. `RemoveAll` instead succeeds: `--otlp-relay-socket=/var/lib/ateom-gvisor` (the BasePath itself, an easy slip since the socket lives under it) would have atelet silently delete the image cache, every actor bundle, the staged runsc binaries and the local checkpoints, at startup, before anything else runs.

`Stop` already uses `os.Remove` on the same path, so the two ends of the socket's lifecycle disagree about which call is appropriate.

internal/otlprelay/client.go
```diff
@@ -0,0 +1,65 @@
...
+				slog.String("socket", sockPath))
+			return nil, nil
+		}
+		return nil, fmt.Errorf("while checking the OTLP relay socket %q: %w", sockPath, err)
+	}
+
+	// gRPC resolves a "unix://" target to a unix socket dialer natively, so the
+	// OTLP exporters above this connection are unchanged: OTLP is gRPC, and gRPC
+	// needs only a reliable byte stream.
+	conn, err := grpc.NewClient("unix://"+sockPath,
```
BenTheElder, line 61: 🤖 **nit** 🟢 – This only builds a valid target for an absolute `sockPath`. With a relative one, `unix://otlp.sock` parses `otlp.sock` as the *authority* with an empty path, so the dial fails — and it fails lazily, at first export, which is precisely the outcome the `os.Stat` check above exists to avoid. `Stat` would have passed, since a relative path resolves fine against the working directory, so the deterministic startup fallback is bypassed and the telemetry is quietly lost instead.

`grpc.NewClient("unix:"+sockPath, ...)` — the single-colon form — takes everything after the colon as the path and handles both shapes, or the flag could reject a non-absolute path up front.

BenTheElder (commented):
Human vetted, AI first pass.

krisztianfekete (changes requested):
I'd hold off on this. It's the ateom part of the same relay we're designing in #761. That relay has to rewrite telemetry identity, and it needs a network listener regardless since a unix socket won't work with microvm guests. 

Can we design that end first and let ateom telemetry leverage that, I am fine with the socket itself, but let's not codify stuff we are about to undo. cc. @git286

git286:
> I'd hold off on this. It's the ateom part of the same relay we're designing in #761. That relay has to rewrite telemetry identity, and it needs a network listener regardless since a unix socket won't work with microvm guests.
> 
> Can we design that end first and let ateom telemetry leverage that, I am fine with the socket itself, but let's not codify stuff we are about to undo. cc. Da Huang (@git286)

@krisztianfekete I think these are two different hops.
  
**#761 is actor → ateom/atelet**: untrusted client inside the sandbox, identity has to be rewritten, and it needs a network listener because a unix socket is invisible to a micro-VM guest. **This PR is ateom → atelet**: ateom is a normal container in the worker pod with the BasePath hostPath mounted, whatever runtime it drives, and its identity is already correct. The socket will work for ateom -> atelet no matter it's gvisor or microvm, right?

krisztianfekete:
> > I'd hold off on this. It's the ateom part of the same relay we're designing in #761. That relay has to rewrite telemetry identity, and it needs a network listener regardless since a unix socket won't work with microvm guests.
> > Can we design that end first and let ateom telemetry leverage that, I am fine with the socket itself, but let's not codify stuff we are about to undo. cc. Da Huang (Da Huang (@git286))
> 
> Krisztian F (@krisztianfekete) I think these are two different hops.
> 
> **#761 is actor → ateom/atelet**: untrusted client inside the sandbox, identity has to be rewritten, and it needs a network listener because a unix socket is invisible to a micro-VM guest. **This PR is ateom → atelet**: ateom is a normal container in the worker pod with the BasePath hostPath mounted, whatever runtime it drives, and its identity is already correct. The socket will work for ateom -> atelet no matter it's gvisor or microvm, right?

You're right, I take back the microvm concern!

The one thing I'd still ask for is scoping. Forwarding as-is is the correct behaviour for ateom telemetry, but right now that covers relay itself rather than about this particular traffic. Whether #761 rewrites at ateom or at atelet, this relay is on this path too, so it will eventually has traffic that has to be modified alongside traffic that must not be. Could we limit this to ateom source now, so adding the rewriting path later is an extension rather than a redoing it?

internal/otlprelay/relay.go
chw120: Thanks @BenTheElder for review. Updated the code to use os.Remove.

internal/otlprelay/client.go
chw120: Added the validation.

chw120:
> > > I'd hold off on this. It's the ateom part of the same relay we're designing in #761. That relay has to rewrite telemetry identity, and it needs a network listener regardless since a unix socket won't work with microvm guests.
> > > Can we design that end first and let ateom telemetry leverage that, I am fine with the socket itself, but let's not codify stuff we are about to undo. cc. Da Huang (Da Huang (Da Huang (@git286)))
> > 
> > 
> > Krisztian F (Krisztian F (@krisztianfekete)) I think these are two different hops.
> > **#761 is actor → ateom/atelet**: untrusted client inside the sandbox, identity has to be rewritten, and it needs a network listener because a unix socket is invisible to a micro-VM guest. **This PR is ateom → atelet**: ateom is a normal container in the worker pod with the BasePath hostPath mounted, whatever runtime it drives, and its identity is already correct. The socket will work for ateom -> atelet no matter it's gvisor or microvm, right?
> 
> You're right, I take back the microvm concern!
> 
> The one thing I'd still ask for is scoping. Forwarding as-is is the correct behaviour for ateom telemetry, but right now that covers relay itself rather than about this particular traffic. Whether #761 rewrites at ateom or at atelet, this relay is on this path too, so it will eventually has traffic that has to be modified alongside traffic that must not be. Could we limit this to ateom source now, so adding the rewriting path later is an extension rather than a redoing it?

Thanks @krisztianfekete @git286 for review. Updated the code to scope it to only allow ateom sources, "ateom-gvisor" and "ateom-microvm".

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,324 @@
...
+// service.name their resource declares. Mirrors the serviceName constants in
+// cmd/ateom-gvisor and cmd/ateom-microvm, which are package main and cannot be
+// imported; TestAteomServicesMatchTheAteomBinaries guards the duplication.
+//
+// Actor telemetry is what the allowlist excludes: actors share a hostname
+// ("runsc") and an interior IP, so their series merge unless identity is
+// injected from outside the actor (#761) -- a rewrite, not a forward. Nothing
+// but ateom reaches this socket today; naming the contract now makes that path
+// an added branch later rather than a re-argument about pass-through.
+var ateomServices = map[string]bool{
```
baizhenyu, line 136: The allowlist is a good contract, but worth saying explicitly in this comment that it is not a security boundary: `service.name` is written by the client, so a compromised process in a worker pod can claim `ateom-gvisor` and pass the check. It prevents accidental misuse (an actor SDK pointed at the socket) and keeps the pass-through contract explicit for #761 — that's its real value. If we ever need to *authenticate* the peer, that takes per-pod sockets or UDS peer credentials (SO_PEERCRED), which also connects to the open identity question on #741.

internal/otlprelay/client.go
```diff
@@ -0,0 +1,68 @@
...
+func Dial(ctx context.Context, sockPath string) (*grpc.ClientConn, error) {
+	if sockPath == "" {
+		return nil, nil
+	}
+	if err := validateSocketPath(sockPath); err != nil {
+		return nil, err
+	}
+	if _, err := os.Stat(sockPath); err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			slog.WarnContext(ctx, "OTLP relay socket absent, exporting telemetry directly over the pod network",
```
baizhenyu, line 51: The fallback is decided once, at startup, and only surfaces as a per-pod log line. Two consequences worth tracking:

1. A worker that boots while atelet is restarting (e.g. a DaemonSet rollout) exports directly over the pod network for its whole life.
2. While any worker can fall back, network policy cannot deny worker-pod egress to the collector — so the blast-radius motivation in the PR description only becomes true in a phase 2 (periodic re-probe, or a strict relay-only mode, plus removing the controller's endpoint injection).

Could we add a metric (or resource attribute) that marks a worker as on the fallback path, so the migration is observable fleet-wide? And a short note in the docs section that the egress lockdown is future work, so nobody reads this PR as already delivering it.

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,324 @@
...
+	if target == "" {
+		slog.InfoContext(ctx, "OTLP relay disabled: no collector endpoint configured",
+			slog.String("env", endpointEnv))
+		return nil, nil
+	}
+
+	// Lazy by design: grpc.NewClient does not block on the collector being up,
+	// so atelet startup does not depend on the collector's readiness.
+	upstream, err := grpc.NewClient(target,
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
```
baizhenyu: Rejecting `https://` instead of silently downgrading (line 308) is the right call and more honest than serverboot's current behavior. But note this collides with #741: when `WithInsecure()` is removed from serverboot so operators can run TLS collectors, the relay becomes the remaining plaintext-only hop, and every ateom on the node is behind it. Worth a follow-up (or a checkbox on #741) to give the relay's upstream leg TLS support at the same time.

internal/otlprelay/e2e_test.go
```diff
@@ -0,0 +1,110 @@
...
+// quietly kept dialing OTEL_EXPORTER_OTLP_ENDPOINT. This one would not: the
+// endpoint variable points at the fake collector *through* the relay only, and
+// the assertion is that the span arrived carrying ateom's own service.name.
+//
+// Run it on its own to watch the hop happen:
+//
+//	go test ./internal/otlprelay/ -run TestEndToEndThroughServerboot -v
+func TestEndToEndThroughServerboot(t *testing.T) {
+	sink, collector := startFakeCollector(t)
+	sock := startRelay(t, collector)
```
baizhenyu, line 42: The isolation this test's comment claims is weaker than stated. `startRelay` does `t.Setenv(endpointEnv, collector)`, and `collector` is a directly routable loopback `host:port` — so the endpoint variable does not point at the fake collector "*through* the relay only". An exporter that ignored `ExporterConn` would read that env and try to deliver directly.

Today a bypass still fails the test, but only by accident: the value is a bare `host:port`, `url.Parse` rejects it (scheme can't start with a digit), the SDK discards the env var and dials the default `localhost:4317`, and the sink times out. That protection evaporates if the SDK's parsing changes or if this env ever gains an `http://` scheme.

One-line fix: after `startRelay` (the relay resolves its upstream once, in `NewServer`, so this doesn't affect it), re-point the env at a parseable but unroutable address before `InitTracing`:

```go
t.Setenv(endpointEnv, "http://127.0.0.1:1")
```

Then a bypassed relay fails loudly and deterministically. Note the replacement must parse as a URL — an unparseable value gets ignored and falls back to `localhost:4317`, where a dev machine might run a real collector.

baizhenyu (commented):
The overall change looks good to me, left some minor comments

cmd/atelet/main.go
```diff
@@ -150,6 +153,27 @@ func main() {
 		EnableHealthz: true,
 	})
 
+	// The OTLP relay lets the ateom pods on this node export telemetry over a
+	// unix socket instead of their own network (see internal/otlprelay). Started
+	// early: an ateom that finds no socket at startup falls back to exporting
+	// directly for its whole life, so the socket should exist before any worker
+	// pod on this node boots.
+	if relay, err := otlprelay.NewServer(ctx, *otlpRelaySocket); err != nil {
+		serverboot.Fatal(ctx, "Failed to create the OTLP relay", err)
```
krisztianfekete: Are we sure about this?
Everything else here degrades nicely, so Fatal might be too much. What do you think?

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,324 @@
...
+	tracesEndpointEnv  = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"
+	metricsEndpointEnv = "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT"
+
+	// otlpDefaultPort matches atenet's normalizeOtlpCollector.
+	otlpDefaultPort = "4317"
+
+	// socketMode keeps the relay socket reachable by the ateom pods on the node
+	// (which do not necessarily share atelet's uid) while staying off-node by
+	// construction. The socket is inside BasePath, a root-owned host directory.
+	socketMode = 0o666
```
krisztianfekete: We get root from the controller, aren't we?
Would `0600` be enough here?

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,324 @@
...
+//   - Shutdown loss. Teardown frees the actor's network and then the pod goes
+//     away, which is exactly when the spans describing teardown are still queued
+//     in the batch processor. atelet outlives the worker pod.
+//
+// The relay forwards the OTLP request message verbatim rather than decoding it
+// into SDK records and re-exporting. Pass-through keeps each ateom's own
+// resource (service.name, service.instance.id, pod attributes) intact, so its
+// spans stay attributed to ateom instead of being absorbed into atelet's. That
+// holds only for a source whose resource is already right, so the relay carries
+// ateom telemetry only; see ateomServices.
```
krisztianfekete: Could we word the guarantee as being about the ateom source, so the rewriting path is an extension later rather than a contradiction?

docs/observability.md
```diff
@@ -221,6 +223,20 @@ Telemetry is emitted the same way everywhere; only the backend differs between a
 >
 > ateom workers don't read the ConfigMap at all — `ate-controller` copies the value into each worker pod at creation. A new endpoint reaches them only once the controller itself restarts, and that restart then rolls every WorkerPool Deployment, replacing the running workers along with the actors on them.
 
+### The ateom OTLP relay
+
+ateom is the one component that does not talk to the collector itself. It exports over a unix socket at `/var/lib/ateom-gvisor/atelet-otlp.sock`, which `atelet` serves and forwards to the collector on the node's network ([`internal/otlprelay`](../internal/otlprelay)):
```
krisztianfekete: As per this comment: https://github.com/agent-substrate/substrate/issues/761#issuecomment-5208531372, by @git286 the GKE pipelines needs checking as after this change ateom telemetry will be passing trough atelet connections, so k8sattributes might break.

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,324 @@
...
+	generic := strings.TrimSpace(os.Getenv(endpointEnv))
+	traces := strings.TrimSpace(os.Getenv(tracesEndpointEnv))
+	metrics := strings.TrimSpace(os.Getenv(metricsEndpointEnv))
+
+	resolved := generic
+	for _, specific := range []string{traces, metrics} {
+		if specific == "" {
+			continue
+		}
+		if resolved != "" && resolved != generic && specific != resolved {
```
krisztianfekete: resolved != generic is false on the first non-empty pass, so the guard can only ever fire when both signal-specific vars are set.

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,324 @@
...
+//
+// A batch is refused whole rather than having the offending resource dropped: a
+// partial success the sender reads as success loses telemetry silently.
+func (t *traceRelay) Export(ctx context.Context, req *coltracepb.ExportTraceServiceRequest) (*coltracepb.ExportTraceServiceResponse, error) {
+	for _, rs := range req.GetResourceSpans() {
+		if err := checkAteomSource(rs.GetResource()); err != nil {
+			return nil, err
+		}
+	}
+	return t.upstream.Export(ctx, req)
```
krisztianfekete: gRPC won't pass the incoming metadata along to the upstream call, so if an ateom has `OTEL_EXPORTER_OTLP_HEADERS` set the headers will be dropped so we cannot do auth I guess?

krisztianfekete: Same applies for metrics below as well.

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,324 @@
...
+		return nil, err
+	}
+	if target == "" {
+		slog.InfoContext(ctx, "OTLP relay disabled: no collector endpoint configured",
+			slog.String("env", endpointEnv))
+		return nil, nil
+	}
+
+	// Lazy by design: grpc.NewClient does not block on the collector being up,
+	// so atelet startup does not depend on the collector's readiness.
```
krisztianfekete, line 399: 1. What happens when the collector is down? 
2. Can we set compression here?

internal/otlprelay/relay.go
chw120: Updated the comment to cover it.

internal/otlprelay/client.go
chw120: Added the resource attribute, and added a note in observability.md.

internal/otlprelay/relay.go
chw120: Added an inline code comment on grpc.WithTransportCredentials(insecure.NewCredentials()) referencing [#741](https://github.com/agent-substrate/substrate/issues/741) for upstream TLS support.

internal/otlprelay/e2e_test.go
chw120: Added the fix.

cmd/atelet/main.go
chw120: Replaced serverboot.Fatal with an error log.

internal/otlprelay/relay.go
chw120: You're right. Updated.

internal/otlprelay/relay.go
chw120: Update the description.

docs/observability.md
chw120: Checked the GKE pipeline and Downward API config:

`ate-controller` already injects `k8s.namespace.name`, `k8s.pod.name`, `k8s.pod.uid`, and `service.instance.id` directly into `OTEL_RESOURCE_ATTRIBUTES` via the Kubernetes Downward API on every worker pod (see `workerpool_apply.go:31`).

Because the relay forwards OTLP requests and resources verbatim without re-encoding, these `k8s.*` resource attributes arrive at the collector intact directly on the payload, without relying on collector-side connection IP-based `k8sattributes` association. Added a clarifying note about this in `docs/observability.md`.

internal/otlprelay/relay.go
chw120: Added more checks.

internal/otlprelay/relay.go
chw120: Added forwardContext() to extract incoming metadata and propagate it as outgoing metadata on the upstream call.

internal/otlprelay/relay.go
chw120: 1. `grpc.NewClient` is non-blocking. If the collector is down, transport errors (`Unavailable`) propagate over UDS to ateom's OTel SDK batch processor to retry/drop asynchronously without blocking actor execution; gRPC reconnects automatically once it recovers.
2. Added upstream gzip compression support via `OTEL_EXPORTER_OTLP_COMPRESSION` (and signal-specific overrides).

internal/otlprelay/relay.go
krisztianfekete: I've been thinking more on this, maybe it's better the other way round and we should read `OTEL_EXPORTER_OTLP_HEADERS` from atelet's env and attach those, and drop or allow-list the client's. Fixes the case I raised without making the rely forward anything problematic. What do you think?

internal/serverboot/serverboot.go
```diff
@@ -122,7 +134,11 @@ func InitTracing(ctx context.Context, opts TracingOptions) (*sdktrace.TracerProv
 	if opts.Sampling.sampler == nil {
 		return nil, fmt.Errorf("TracingOptions.Sampling is required")
 	}
-	res, err := newResource(ctx, opts.ServiceName)
+	relayStatus := "direct"
+	if opts.ExporterConn != nil {
+		relayStatus = "relay"
+	}
+	res, err := newResource(ctx, opts.ServiceName, attribute.String("substrate.otlp.relay", relayStatus))
```
krisztianfekete: Nice!  Two things on the name though:

1.  our [custom attributes](https://github.com/agent-substrate/substrate/blob/main/internal/ateattr/ateattr.go) live under `ate.*` and nothing uses substrate.*, so use `ate.otlp.relay` please.
2.  since `newResource` is shared, every component gets this, so ateapi, atelet, atecontroller and atenet all set `direct` forever for a relay they'll never use. Worth setting it only on the ateom path if we can.

internal/serverboot/serverboot_test.go
```diff
@@ -67,6 +68,24 @@ func TestNewResourceEnvWins(t *testing.T) {
...
+	}
+
+	resRelay, err := newResource(context.Background(), "ateom-gvisor", attribute.String("substrate.otlp.relay", "relay"))
+	if err != nil {
+		t.Fatalf("newResource: %v", err)
+	}
+	if got := resourceAttrs(resRelay)["substrate.otlp.relay"]; got != "relay" {
+		t.Errorf("substrate.otlp.relay = %q, want relay", got)
+	}
+}
```
krisztianfekete: This isn't really testing `resource.New`, and it would still pass if `InitTracing` stopped setting it, right?
Can we cover the the nil conn vs non-nil conn decision, which only the otlprelay e2e covers today, and only for relay?

cmd/ateom-gvisor/main.go
```diff
@@ -106,16 +106,28 @@ func do(ctx context.Context) error {
...
+	// Export through atelet's node-local relay when it is there, so telemetry
+	// never touches the worker pod's network. A nil conn means it is not, and
+	// both providers fall back to dialing the collector directly.
+	relayConn, err := otlprelay.Dial(ctx, *otlpRelaySocket)
+	if err != nil {
+		serverboot.Fatal(ctx, "Failed to connect to the OTLP relay", err)
+	}
+	if relayConn != nil {
+		defer relayConn.Close()
+	}
```
krisztianfekete, line 126: Can we do the same we have now for atelet to avoid crashes?

cmd/ateom-microvm/main.go
```diff
@@ -103,16 +105,28 @@ func do(ctx context.Context) error {
...
+	// never touches the worker pod's network. A nil conn means it is not, and
+	// both providers fall back to dialing the collector directly.
+	relayConn, err := otlprelay.Dial(ctx, *otlpRelaySocket)
+	if err != nil {
+		serverboot.Fatal(ctx, "Failed to connect to the OTLP relay", err)
+	}
+	if relayConn != nil {
+		defer relayConn.Close()
+	}
+
```
krisztianfekete, line 126: See https://github.com/agent-substrate/substrate/pull/809/changes#r3764908115

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,398 @@
...
+// one: an unidentified source is the one the relay cannot vouch for.
+func checkAteomSource(r *resourcepb.Resource) error {
+	name := resourceServiceName(r)
+	if ateomServices[name] {
+		return nil
+	}
+	return status.Errorf(codes.PermissionDenied,
+		"the OTLP relay carries ateom telemetry only, got service.name %q; a source whose identity has to be rewritten (#761) must not be forwarded verbatim",
+		name)
+}
```
krisztianfekete: Note that OTel env vars can override these, so if someone sets this on an ateom pod every export gets PermissionDenied, which the SDK doesn't retry. Probably fine though?

internal/ateompath/ateompath.go
```diff
@@ -53,6 +53,22 @@ func GVisorReleaseDir(sha256 string) string {
...
+// It sits directly under BasePath, which is the host directory already mounted
+// at the same path into atelet and into every ateom pod, so no new volume is
+// needed for ateom to reach it.
+func AteletOTLPSocketPath() string {
+	return filepath.Join(
+		BasePath,
+		"atelet-otlp.sock",
+	)
+}
+
```
krisztianfekete, line 77: Wondering if a subdir only atelet writes, mounted read-only into the pods, would be better here?

cmd/ateom-gvisor/main.go
chw120: Sure, changed to log.

cmd/ateom-microvm/main.go
chw120: Same.

internal/ateompath/ateompath.go
chw120: It would be better, but I think it would be better to have another PR updating it, since it isn't specific to this socket. Instead, added more comments. WDYT?

internal/serverboot/serverboot_test.go
chw120: Updated the tests.

internal/serverboot/serverboot.go
chw120: Thanks @krisztianfekete for review. 

Updated, and only ateom-gvisor and ateom-microvm will have this attribute.

internal/otlprelay/relay.go
chw120: Good concern. Added a warning.

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,569 @@
...
+		if !found {
+			return nil, fmt.Errorf("OTLP header %d is not in key=value form", i+1)
+		}
+		key = strings.ToLower(strings.TrimSpace(key))
+		if key == "" {
+			return nil, fmt.Errorf("OTLP header %d has an empty name", i+1)
+		}
+		// Percent-decoding is what makes a value containing "," or "=" (a base64
+		// token, say) expressible in this format at all.
+		decoded, err := url.QueryUnescape(strings.TrimSpace(value))
```
krisztianfekete: Can we use `url.PathUnescape` here?

internal/otlprelay/relay.go
```diff
@@ -0,0 +1,569 @@
...
+	name := resourceServiceName(r)
+	if ateomServices[name] {
+		return nil
+	}
+	if _, dup := g.logged.LoadOrStore(name, struct{}{}); !dup {
+		slog.WarnContext(ctx, "OTLP relay rejected telemetry from an unrecognized source; it is being dropped, not retried. If this is an ateom, check whether OTEL_SERVICE_NAME or OTEL_RESOURCE_ATTRIBUTES on the worker pod is overriding its service.name",
+			slog.String("service.name", name),
+			slog.Any("allowed", allowedServices()),
+			slog.String("note", "logged once per distinct service.name"))
+	}
```
krisztianfekete: `service.name` is client-side, so not really bounded. Can we adopt the `lru.New(256)` pattern from atelet?

docs/observability.md
```diff
@@ -221,6 +223,24 @@ Telemetry is emitted the same way everywhere; only the backend differs between a
...
+
+```
+ateom ──OTLP/gRPC over unix socket──► atelet relay ──OTLP/gRPC──► collector
+```
+
+The socket sits in the `BasePath` hostPath already mounted into both, so nothing new is mounted. `atelet` is a DaemonSet, so every ateom on a node shares one relay, and the many per-pod collector connections collapse into one per node. Four things motivate it: the worker pod runs untrusted agent code and will not need egress to the collector once direct fallback is phased out; the connection count drops; ateom's own telemetry stays clear of the transparent egress redirect it installs for the actor; and `atelet` outlives the worker pod, so spans still queued at teardown are not lost with it.
+
+The relay is best-effort. If the socket is absent when ateom starts — `atelet` not up yet, `--otlp-relay-socket=""`, or no collector configured for the relay to forward to — ateom logs it and exports directly to `OTEL_EXPORTER_OTLP_ENDPOINT` as before. That fallback is decided once at startup, not per export, and is stamped on telemetry as the `substrate.otlp.relay` resource attribute (`relay` vs `direct`).
```
krisztianfekete:
```suggestion
The relay is best-effort. If the socket is absent when ateom starts — `atelet` not up yet, `--otlp-relay-socket=""`, or no collector configured for the relay to forward to — ateom logs it and exports directly to `OTEL_EXPORTER_OTLP_ENDPOINT` as before. That fallback is decided once at startup, not per export, and is stamped on telemetry as the `ate.otlp.relay` resource attribute (`relay` vs `direct`).
```

internal/ateompath/ateompath.go
krisztianfekete: Fair, could you please open an issue for it rather than leaving it as a comment though so we can track this?

internal/otlprelay/relay.go
chw120: Updated.

internal/otlprelay/relay.go
chw120: Updated.

docs/observability.md
chw120: Updated.

internal/ateompath/ateompath.go
chw120: Good point! This is part of the broader host socket/filesystem isolation discussion already tracked in #778 (and #853). We can address isolating atelet-owned sockets into dedicated read-only mounts as part of that work.

krisztianfekete (approved):
LGTM!

docs/observability.md
```diff
@@ -198,7 +198,9 @@ To visualize traces locally:
...
+4. **Search and Inspect**: Copy the printed Trace ID from the CLI output and paste it into the Jaeger search box (top right), or select `ateapi`, `atelet`, or `ateom-gvisor` under the **Service** dropdown and click **Find Traces** to inspect detailed call stacks, DB transactions, state updates, and worker pod handoffs.
+
+> ateom carries no manual spans — its only instrumentation is the `otelgrpc` interceptor on the gRPC surface `atelet` calls. So it produces a span for an actor lifecycle operation (`suspend`, `resume`) and nothing at all for a read like `kubectl ate get actor`. Its sampler is parent based, so a lifecycle command is traced end to end into ateom whenever `ateapi` roots a sampled trace, which the kind overlay makes unconditional; the per-component ratio never enters into it. `ateom-gvisor` appearing here at all is also the sign its [OTLP relay](#the-ateom-otlp-relay) is working, since that is the only path its telemetry has out of the worker pod.
```
git286: > `ateom-gvisor` appearing here at all is also the sign its [OTLP relay](#the-ateom-otlp-relay) is working, since that is the only path its telemetry has out of the worker pod.

Is it true? When the socket fails, the trace will still fallback to the network endpoint, right? Why trace show up means the relay is working?

internal/ateompath/ateompath.go
git286: Probably a followup improvement: we need more protection in case actor escape. For example, bounded the memory/message size, limit the max concurrent streams etc.

docs/observability.md
chw120: Good catch. You're right. Updated the doc.

git286 (approved):

Issue #791, opened 2026-08-07
Support suspending a PAUSED actor without waking it


HavenXia:
From #119,

>  If an actor remains paused for an extended period and `--timer` is provided, the system escalates it to a `SUSPENDED` state to ensure data safety and free up local node storage.

We need to support the `SuspendActor` on a `PAUSED` actor, overall it should be one atelet call that uploads the existing local files, then the api records the ActorSnapshot, clears `LocalSnapshotInfo` and the node pin, and sets `SUSPENDED`. 

This is a critical path for both Actor Lifecycle(#119) and System Upgrade(#473)

dberkov:
Proposed implementation plan, split into 5 PRs so work can proceed in parallel (PRs 1–3 are mutually independent; only PR 5 turns the feature on).

## Design summary

Keep the existing 5-step `SuspendActor` workflow and branch inside the steps on a `FromPaused` discriminator computed at load time: `status == PAUSED`, or `status == SUSPENDING && WorkerAssignment == nil && LocalSnapshotInfo != nil` (the nil-assignment conjunct is required because resume never clears `LocalSnapshotInfo`, so a RUNNING actor resumed from pause still carries a stale one).

On the node, a **new atelet RPC `UploadPausedCheckpoint`** (not a new `CheckpointType` — the `Checkpoint` handler drives a live ateom, which a paused actor doesn't have) reads the local pause snapshot's `manifest.json`, uploads the files via the same `ategcs` primitives as `uploadExternalCheckpoint` (manifest last, as commit marker), then prunes the local checkpoint. ateapi dials atelet by node name (`LocalSnapshotInfo.NodeVmsWithLocalSnapshots`) via a new `DialForNode`, since a paused actor has no worker pod.

Scope handling — the upload honors the commit scope (`OnCommit`), like a normal suspend:

| Captured (OnPause) | Desired (OnCommit) | Behavior |
|---|---|---|
| FULL | FULL | plain upload |
| DATA | DATA | plain upload |
| FULL | DATA | microVM: extract durable-disk files; **gVisor: `Unimplemented` until split-checkpoint runsc (#790)** |
| DATA | FULL | rejected (`FailedPrecondition`) — cannot fabricate memory; covers the golden-atespace corner |

[...]

PR #813, opened 2026-08-08
Add DialForNode and always finalize in FinalizeSuspendedStep

dberkov:
Part of #791 (PR 3 of the [implementation plan](https://github.com/agent-substrate/substrate/issues/791#issuecomment-5226674097)).

Two behavior-preserving refactors needed before suspending a PAUSED actor, which has no worker assignment:

**Dialer** — extract `dialAteletOnNode` from `DialForWorker`'s tail (byNode index lookup → single-atelet check → per-pod-UID connection cache → pinned-mTLS dial) and expose `DialForNode(nodeName)` for actors whose state is pinned to a node without a worker pod (a PAUSED actor's local snapshot). A node with no atelet in the informer cache returns the new sentinel `ErrNoAteletOnNode` — documented as retryable (atelet restarting / informer lag), distinct from the crash-worthy `ErrWorkerPodNotFound`. `DialForWorker` behavior is unchanged apart from error wrapping in the zero-atelets case.

**FinalizeSuspendedStep** — hoist finalization (ActorSnapshot record, `LatestSnapshot`, the `SUSPENDED` flip, clearing `InProgressSnapshot`/`WorkerAssignment`/`LocalSnapshotInfo`) out of the `if assignment != nil` worker-freeing branch so it always runs. Previously, a SUSPENDING actor with no assignment was silently skipped — the workflow reported success while the actor stayed SUSPENDING forever. Unreachable today (`CallAteletSuspendStep`'s prerequisite crashes nil-assignment actors first), but load-bearing for the paused-origin suspend, where no worker exists by design.

Tests:
- `TestDialForNode`: no atelet on node → `ErrNoAteletOnNode`; two atelets → error; happy path asserts pod-UID-pinned credentials and connection caching.
- `TestFinalizeSuspendedStep_NoAssignment`: SUSPENDING actor with `LocalSnapshotInfo` and an in-progress URI but no assignment → SUSPENDED, ActorSnapshot created at the URI, `LatestSnapshot` set, in-progress fields and `LocalSnapshotInfo` cleared.

`go test -race ./cmd/ateapi/...` (incl. the functional suite covering the running-path suspend end-to-end), `go vet`, `gofmt` clean. Independent of #810 and #812.

### DIFF
--- a/cmd/ateapi/internal/controlapi/dialer.go
+++ b/cmd/ateapi/internal/controlapi/dialer.go
@@ -36,6 +36,12 @@ import (
 
 var ErrWorkerPodNotFound = errors.New("worker pod not found")
 
+// ErrNoAteletOnNode reports that the informer cache holds no atelet pod for
+// the requested node — e.g. the atelet is restarting, or the node is gone.
+// Distinct from ErrWorkerPodNotFound, which callers treat as crash-worthy;
+// this one is retryable.
+var ErrNoAteletOnNode = errors.New("no atelet pod found on node")
+
 // The SPIFFE identity that atelet serving certs carry, as minted by the
 // podidentity signer (cmd/podcertcontroller/internal/podidentitysigner).
 // The namespace part is ateletNamespace, declared in informer.go.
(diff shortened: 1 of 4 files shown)


cmd/ateapi/internal/controlapi/dialer.go
```diff
@@ -72,6 +78,14 @@ func NewAteletDialer(workerIndexer cache.Indexer, ateletIndexer cache.Indexer, c
 	}
 }
 
+// DialForNode returns a gRPC connection to the atelet running on the given
+// node. Used when an actor has no worker assignment but its state is pinned
+// to a node — e.g. a PAUSED actor whose local snapshot lives there. Returns
+// ErrNoAteletOnNode if the informer cache holds no atelet pod for the node.
+func (d *AteletDialer) DialForNode(nodeName string) (*grpc.ClientConn, error) {
+	return d.dialAteletOnNode(nodeName)
```
HavenXia: nit: This is 0 additional logic wrap, why don't just make it `DialForAteletOnNode`?

HavenXia (dismissed):

cmd/ateapi/internal/controlapi/dialer.go
dberkov: renamed

juli4n:
LGTM. Needs rebase.

juli4n (approved):

cmd/ateapi/internal/controlapi/dialer.go
HavenXia: thanks

PR #814, opened 2026-08-08
docs: fix and fill gaps in the kubectl-ate CLI reference

botengyao:
Follow-up to #561, doc-only — happy to open an issue if preferred.

#561 fixed the `kubectl ate logs` examples that omitted the required `-a/--atespace`. Two more documented commands have the same problem:

- `cmd/kubectl-ate/README.md`, the `--trace` example: `kubectl ate get actor my-counter-1 --trace`
- `demos/parking/README.md`: `kubectl ate get actors`

Both fail with `--atespace is required when getting actors` / `specify --atespace <name> to list one atespace, or -A/--all-atespaces for all`.

While in the CLI README, three gaps that a first-time reader hits:

- The `get actor` STATUS column lists 4 states; the API defines 8. Added `STATUS_PAUSING`, `STATUS_PAUSED`, `STATUS_CRASHED`, `STATUS_DELETING` (all reachable — the control plane sets each of them today).
- `pause actor` was undocumented. Added it alongside `suspend`, with a note on the distinction that matters in practice: `suspend` uploads the snapshot so the actor can resume anywhere, `pause` keeps it on the node VM and biases the next resume back to that node.
- `top` was undocumented, even though the `get workers` section already pointed at "the same flags as `top workers`". Added a section with the flags and the output columns, plus a note that it needs metrics-server.

Plus a typo in `hack/update/proto-fmt.sh` ("wince" → "since").

- [x] Tests pass

### DIFF
--- a/cmd/kubectl-ate/README.md
+++ b/cmd/kubectl-ate/README.md
@@ -54,7 +54,7 @@ The kind overlay installed by `hack/install-ate-kind.sh --deploy-ate-system` alr
 Port-forward the Jaeger UI and invoke any command with `--trace`:
 ```bash
 kubectl port-forward -n otel-system svc/jaeger 16686:16686 &
-kubectl ate get actor my-counter-1 --trace
+kubectl ate get actor my-counter-1 -a demo --trace
 # open http://localhost:16686 and search for the most recent trace
 ```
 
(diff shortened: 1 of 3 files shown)


botengyao:
cc @HavenXia

botengyao:
Hey @HavenXia, would you like to take a look at this PR? Thanks!

PR #816, opened 2026-08-10
Support suspending a PAUSED actor without waking it

dberkov:
Part of #791 — the last planned piece of the [implementation plan](https://github.com/agent-substrate/substrate/issues/791#issuecomment-5226674097) (after #810, #812, #813). #791 stays open until #817 (gVisor Full-capture → Data-commit conversion, blocked on #790) is done; this PR covers micro-VM fully and gVisor for scope-matched suspends. Also part of the actor state machine (#119) and a prerequisite for system upgrade flows (#473).

`SuspendActor` now accepts a PAUSED actor: instead of checkpointing a running workload, ateapi dials the atelet on the node holding the pause snapshot and has it upload the node-local files to object storage, then finalizes as usual — durable `ActorSnapshot`, `SUSPENDED` status, node pinning cleared.

Two commits, reviewable independently:

## Commit 1 — atelet: `UploadPausedCheckpoint` RPC (dead code until commit 2)

- New `AteomHerder` RPC: a pure disk→object-storage copy driven by the snapshot's self-describing manifest — no ateom involved (the sandbox is gone).
- **Scope conversion** dispatches per sandbox class (`narrowFullCaptureToData`): a micro-VM FULL capture narrows to a DATA upload by carving out `durable-dir.tar` (constant hoisted to `ateompath`, shared with ateom-microvm); gVisor returns `Unimplemented` until split checkpoints land (#790); DATA can never widen to FULL; a scope-less manifest (older atelet) is rejected rather than guessed at.
- **Idempotent retry**: local files gone + remote manifest present ⇒ a previous invocation committed, succeed; gone on both sides ⇒ unrecoverable (`LOCAL_SNAPSHOT_GONE`, crashes the actor). Upload failures stay plain retryable errors; the manifest uploads last as the commit marker, never in parallel.
- The golden atespace is rejected at validation (fully on the `field.ErrorList` framework): golden actors are never paused.

## Commit 2 — control plane: enable suspend from PAUSED

- `FromPaused` discriminator: PAUSED status, or SUSPENDING with no worker assignment and a `LocalSnapshotInfo` — the field alone is stale on resumed-from-pause RUNNING actors, so the nil-assignment conjunct is load-bearing.
- `MarkSuspendingStep` accepts PAUSED and rejects a Data-captured pause against a Full commit *before* the actor leaves PAUSED (an upload cannot fabricate memory), using the `content_scope` recorded at pause (#812) with an `onPause` fallback.
- `CallAteletSuspendStep` paused branch dials by node (`DialForAteletOnNode`, #813): missing node record ⇒ crash (the snapshot can never be found); unreachable atelet ⇒ retryable; atelet's `LOCAL_SNAPSHOT_GONE` ⇒ crash via `maybeCrashActor`. `FinalizeSuspendedStep` needed no changes thanks to the #813 hoist.
- Root-cause guard: `MarkPausingStep` rejects pausing golden-atespace actors.
- Docs: pause states + the new `PAUSED → SUSPENDING` edge in the architecture state diagram; glossary Suspend entry covers both origins.

## Tests

- **atelet unit**: 10 upload-helper subtests (conversion matrix, idempotency probe, data-loss crash, retryable upload failure) via a recording object-storage fake; validation table.
- **control-plane unit**: discriminator table, scope-rejection table (incl. onPause fallback), paused preconditions (no node ⇒ CRASHED, no atelet ⇒ `ErrNoAteletOnNode` + still SUSPENDING), golden-pause rejection, prerequisite matrix updated.
- **functional (envtest)**: `TestSuspendActor_FromPaused` (upload called with the pause snapshot name, no Checkpoint RPC, SUSPENDED, pinning cleared, ActorSnapshot at the upload destination) + retry-after-failed-upload (same destination on retry).
- **e2e (demo suite)**: the lifecycle driver gains a suspend-from-PAUSED mode; three durable-dir cases — Full/Full, Data/Data, and the micro-VM Full→Data extraction — assert memory/file counters survive the full pause→suspend→resume journey and the node pinning is gone.

`go test -race ./...` clean (except the pre-existing macOS-only `internal/atunnel` unix-socket-path failures, untouched by this PR), `gofmt`/`go vet` clean, protos regenerated via `hack/protoc.sh`.

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_pause.go
+++ b/cmd/ateapi/internal/controlapi/workflow_pause.go
@@ -114,6 +114,11 @@ func (w *ActorWorkflow) ensureMarkedPausing(ctx context.Context, actorRef resour
 	if actor.GetStatus() != ateapipb.Actor_STATUS_RUNNING {
 		return nil, status.Errorf(codes.FailedPrecondition, "MarkPausing prerequisite not met for Actor: %s (got: %v, want %s)", actorRef, actor.GetStatus(), ateapipb.Actor_STATUS_RUNNING)
 	}
+	// By design a golden actor cannot be paused — it can only be suspended
+	// (committed).
+	if actorRef.Atespace == resources.GoldenActorAtespace {
+		return nil, status.Errorf(codes.FailedPrecondition, "actors in atespace %q are golden actors, which cannot be paused", actorRef.Atespace)
+	}
 
 	snapshotName := resources.NewSnapshotName()
 	updated, err := w.store.UpdateActor(ctx, actorRef, func(dbActor *ateapipb.Actor) error {
--- a/cmd/ateapi/internal/controlapi/workflow_suspend.go
+++ b/cmd/ateapi/internal/controlapi/workflow_suspend.go
@@ -31,9 +31,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-// SuspendActor executes the workflow to suspend a running actor. Idempotent:
-// a re-entered workflow fast-forwards past the steps a previous attempt
-// completed, deriving progress from the persisted actor alone.
+// SuspendActor executes the workflow to suspend a running or paused actor:
+// a running actor is checkpointed on its worker, a paused actor's node-local
+// snapshot is uploaded. Idempotent: a re-entered workflow fast-forwards past
+// the steps a previous attempt completed, deriving progress from the
+// persisted actor alone.
 func (w *ActorWorkflow) SuspendActor(ctx context.Context, actorRef resources.ActorRef) (_ *ateapipb.Actor, err error) {
 	start := time.Now()
 	var actor *ateapipb.Actor
@@ -61,12 +63,20 @@ func (w *ActorWorkflow) SuspendActor(ctx context.Context, actorRef resources.Act
 		// left to do.
 		return actor, nil
 	}
+	// Decided before marking: once SUSPENDING is committed, the loaded status
+	// alone can no longer tell the two origins apart.
+	fromPaused := isPausedOriginSuspend(actor)
 	var marked *ateapipb.Actor
 	if marked, err = w.ensureMarkedSuspending(lockCtx, actorRef, actor, actorTemplate); err != nil {
 		return nil, err
 	}
 	actor = marked
-	if wireSnapshotScope, err = w.ensureAteletSuspended(lockCtx, actorRef, actor, actorTemplate); err != nil {
+	if fromPaused {
+		wireSnapshotScope, err = w.ensurePausedSnapshotUploaded(lockCtx, actorRef, actor, actorTemplate)
+	} else {
+		wireSnapshotScope, err = w.ensureAteletSuspended(lockCtx, actorRef, actor, actorTemplate)
+	}
+	if err != nil {
 		return nil, err
 	}
 	if err = w.ensureVolumesDetached(lockCtx, actor, actorTemplate, "DetachVolumes", ateattr.OperationSuspend); err != nil {
@@ -96,11 +106,11 @@ func (w *ActorWorkflow) loadActorForSuspend(ctx context.Context, actorRef resour
 	return actor, actorTemplate, nil
 }
 
-// ensureMarkedSuspending transitions a RUNNING actor to SUSPENDING, minting
-// the in-progress snapshot location and recording the actor version the
-// snapshot will capture. Skips when a previous attempt already marked the
-// actor; the persisted location and source version then stay authoritative
-// for the rest of the workflow.
+// ensureMarkedSuspending transitions a RUNNING or PAUSED actor to
+// SUSPENDING, minting the in-progress snapshot location and recording the
+// actor version the snapshot will capture. Skips when a previous attempt
+// already marked the actor; the persisted location and source version then
+// stay authoritative for the rest of the workflow.
 func (w *ActorWorkflow) ensureMarkedSuspending(ctx context.Context, actorRef resources.ActorRef, actor *ateapipb.Actor, actorTemplate *atev1alpha1.ActorTemplate) (_ *ateapipb.Actor, err error) {
 	ctx, done := stepSpan(ctx, "MarkSuspending")
 	defer func() { err = done(err) }()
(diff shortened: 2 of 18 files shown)


cmd/atelet/main.go
```diff
@@ -696,6 +698,169 @@ func (s *AteomHerder) uploadExternalCheckpoint(ctx context.Context, req *ateletp
...
+		s.instruments.recordCheckpoint(ctx, op, err,
+			phase{ateattr.SnapshotPhasePersist, dPersist},
+			phase{ateattr.SnapshotPhaseTotal, time.Since(tStart)})
+	}()
+
+	uri, err := resources.ParseSnapshotURI(req.GetDestinationSnapshotUri())
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	localDir := filepath.Join(ateompath.LocalCheckpointsDir(req.GetActorUid()), req.GetLocalSnapshotName())
```
ILLUM1N0X: Joining `ateompath.LocalCheckpointsDir(actorUID)` with `snapshotName` is repeated in two other places [[1]](https://github.com/agent-substrate/substrate/blob/231a9ef189a881c6b7f9ae12fb81db86447537e4/cmd/atelet/main.go#L629) [[2]](https://github.com/agent-substrate/substrate/blob/231a9ef189a881c6b7f9ae12fb81db86447537e4/cmd/atelet/main.go#L778).

WDYT of adding a new helper `LocalSnapshotDir` to `internal/ateompath/ateompath.go`?

cmd/atelet/main.go
```diff
@@ -696,6 +698,169 @@ func (s *AteomHerder) uploadExternalCheckpoint(ctx context.Context, req *ateletp
...
+			return "", nil
+		} else if errors.Is(fetchErr, ateerrors.ReasonFailedGetExternalObject) {
+			return "", ateerrors.NewGRPCError(ctx, codes.DataLoss, ateerrors.ReasonLocalSnapshotGone, ateerrors.ActorCrashedMetadata(),
+				fmt.Errorf("local snapshot %q is gone and no uploaded copy exists: %w", req.GetLocalSnapshotName(), fetchErr))
+		} else {
+			return "", fmt.Errorf("while probing for an already-uploaded snapshot manifest: %w", fetchErr)
+		}
+	} else if err != nil {
+		return "", wrapFileSystemErr("while reading local snapshot manifest", err)
+	}
```
ILLUM1N0X, line 785: Since every branch terminates with an explicit `return`, WDYT of removing redundant `else` statements and using flat `if` statements to improve readability?

cmd/atelet/main.go
```diff
@@ -696,6 +698,169 @@ func (s *AteomHerder) uploadExternalCheckpoint(ctx context.Context, req *ateletp
...
+	if err != nil {
+		return rec.SandboxClass, fmt.Errorf("while marshaling snapshot manifest: %w", err)
+	}
+	// The manifest goes last, never in parallel: its presence is the commit
+	// marker — readers assume every file it lists is already present. A crash
+	// mid-upload thus leaves only orphaned files, never a manifest pointing
+	// at files that never landed.
+	if err := ategcs.SendBytesToGCS(ctx, s.gcsClient, manifestURI, updatedManifest); err != nil {
+		return rec.SandboxClass, fmt.Errorf("while uploading snapshot manifest: %w", err)
+	}
```
ILLUM1N0X: This logic is almost identical to the one inside [`uploadExternalCheckpoint`](
https://github.com/agent-substrate/substrate/blob/231a9ef189a881c6b7f9ae12fb81db86447537e4/cmd/atelet/main.go#L664-L695). WDYT of extracting a shared private helper?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -32,7 +32,8 @@ service Control {
   // Update mutable fields on an existing Actor.
   rpc UpdateActor(UpdateActorRequest) returns (Actor) {}
 
-  // Suspend a given actor to a new snapshot.
+  // Suspend a given actor to a new snapshot. A running actor is checkpointed
+  // on its worker; a paused actor's node-local snapshot is uploaded as-is.
```
ILLUM1N0X: Not necessarily as-is, it could be narrowed down from FULL to DATA

docs/glossary.md
```diff
@@ -74,8 +74,10 @@ because they change too frequently for etcd.
 
 ## Lifecycle
 
-- **Suspend**: hibernate a running Actor by checkpointing it to a snapshot and
-  freeing its Worker. The requested snapshots are uploaded to external storage.
+- **Suspend**: hibernate a running or paused Actor into a durable snapshot in
+  external storage. A running Actor is checkpointed on its Worker (which is
+  then freed); a paused Actor's node-local snapshot is uploaded as-is, ending
```
ILLUM1N0X: Same

internal/proto/ateletpb/atelet.proto
ILLUM1N0X, line 41: Since this now contains an RPC that works on the node level (rather than the worker level), should we consider renaming it?

Alternatively, should we create a new node level service? (It might be too over-kill for now since it's a single RPC, unless we anticipate other node level RPCs in the future)

cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -89,6 +89,12 @@ func (s *MarkPausingStep) CheckPrerequisite(ctx context.Context, input *PauseInp
 	if state.Actor.GetStatus() != ateapipb.Actor_STATUS_RUNNING {
 		return status.Errorf(codes.FailedPrecondition, "MarkPausingStep prerequisite not met for Actor: %s (got: %v, want %s)", input.ActorRef, state.Actor.GetStatus(), ateapipb.Actor_STATUS_RUNNING)
 	}
+	// Golden actors are never paused: their only lifecycle is the golden
+	// flow's create → resume → commit, and a paused golden could later be
+	// promoted to the template's golden snapshot by a suspend.
```
HavenXia: I think you want to say if we allow an golden actor to be paused, it will cause issue that suspending this paused golden actor will overwrite the golden snapshot.

nit: reword

cmd/ateapi/internal/controlapi/workflow_suspend.go
zoez7, line 1: I think going from paused -> suspend should be its own workflow, which will make it easier to review and reason about.

internal/proto/ateletpb/atelet.proto
dberkov: ahh, name of service is confusing. This is the atlet service that runs a seamonset on the VM and controls ateoms. Each ateom has its own service internal/proto/ateompb/ateom.proto

cmd/ateapi/internal/controlapi/workflow_suspend.go
dberkov: It will complicate the system, and everyone will need to check what is the state of the actor and call to different API. semantic supposed to stay the same.  In the future we will extend suspend API by providing additional fields like autoTag the snapshot , so it will require duplicate the behavior with this API too.

cmd/atelet/main.go
dberkov: Done in 462136d2 — added `ateompath.LocalSnapshotDir(actorUID, snapshotName)` and switched all three join sites to it (moveLocalCheckpoint, UploadPausedCheckpoint, Restore's local-manifest read).

cmd/atelet/main.go
dberkov: Done in 462136d2 — the probe block is now flat sequential ifs with returns.

cmd/atelet/main.go
dberkov: Done in 462136d2 — extracted a shared `uploadSnapshot` helper (file fan-out + manifest-last commit marker); both `uploadExternalCheckpoint` and `uploadLocalCheckpointDir` are thin callers now.

pkg/proto/ateapipb/ateapi.proto
dberkov: Good catch — fixed in 462136d2: the comment now says the paused snapshot is narrowed to the template's commit scope where required (Full capture, Data commit).

docs/glossary.md
dberkov: Fixed in 462136d2 alongside the proto comment.

cmd/ateapi/internal/controlapi/workflow_pause.go
dberkov: Reworded in 462136d2: "By design a golden actor cannot be paused — it can only be suspended (committed)."

HavenXia (approved):

Issue #763, opened 2026-08-05
Rework Update<ResourceType> in storage layer


juli4n:
The Update methods in the storage later take the resource and an expected version as parameters. For example:

```
UpdateActor(ctx context.Context, actor *Actor, expectedVersion int64) (*Actor, error)
```

The contract is that the mutable fields are copied over from `actor` iff the `expectedVersion` matches the value in the DB. This is checked within a single tx.

A much simpler (and expressive) API would be one that takes a mutate closure and performs a transactional read-modify-write. For example:

```
UpdateActor(ctx context.Context, actorRef,  update func(*Actor) error) (*Actor, error)
```

This would simplify callers (see discussion https://github.com/agent-substrate/substrate/pull/742#discussion_r3722633272). 

This task is to convert all `Update<Type>` methods in the store Go API to this model.

thockin:
I think we need to let the storage layer be atomic. To my eyes that looks something like:

RPC layer UpdateActor(newActor)
Create a callback
Call storage layer UpdateActor(newActor, callback)
Open txn
Get current value from DB
Call callback(new, old)
6a) Check UID precondition IFF it was specified
6b) Check version IFF it was specified
6c) Apply changes to in-memory state
Commit

laoj2:
Closing this issue. The remaining PR (https://github.com/agent-substrate/substrate/pull/1014) is more closely related to https://github.com/agent-substrate/substrate/issues/954

PR #824, opened 2026-08-10
Add ActorTemplate and ActorTemplateVersion as substrate resources; Add CRUD methods in redis package

zoez7:
Part of #477 .

All ATV fields are immutable. SandboxConifg will be frozen into ATV at creation time.
Adds per-resource store methods and the ateredis package for the AT and ATV as global resources.


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -276,6 +306,264 @@ message ActorSnapshotRef {
...
+message ActorTemplateVersion {
+  // Common resource metadata: name, uid, version, timestamps.
+  ResourceMetadata metadata = 1;
+
+  // actor_template is the parent ActorTemplate. Required at creation and
+  // immutable.
+  ObjectRef actor_template = 2;
+
+  // spec is immutable after creation.
+  ActorTemplateVersionSpec spec = 3;
```
juli4n: As discussed, substrate resources are not split into `spec` / `status`, at least not today, so I would err to the side of consistency and merge them. If we decide to go this route, we can always restructure *all* the resources together.

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -62,6 +62,9 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+// globalAteSpace in Substrate is represented by "".
+const globalAteSpace = ""
```
juli4n: "atespace", one word.

zoez7: Done.

pkg/proto/ateapipb/ateapi.proto
zoez7: Done.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -276,6 +306,257 @@ message ActorSnapshotRef {
...
+
+  repeated Volume volumes = 5;
+
+  SnapshotsConfig snapshots_config = 6;
+
+  // sandbox_config selects the sandbox runtime this version's actors run on.
+  // Required. Resolved and frozen into resolved_sandbox at creation time.
+  SandboxConfig sandbox_config = 7;
+
+  // cpu class will be specified here in the future.
```
juli4n: Add a TODO comment?

cmd/ateapi/internal/store/store.go
```diff
@@ -105,6 +105,50 @@ type Interface interface {
...
+	GetActorTemplate(ctx context.Context, name string) (*ateapipb.ActorTemplate, error)
+
+	// ActorTemplateExists reports whether the ActorTemplate exists.
+	ActorTemplateExists(ctx context.Context, name string) (bool, error)
+
+	// Updates an ActorTemplate with optimistic concurrency check and returns
+	// the stored resource with advanced metadata (version, update_time). The
+	// input is not mutated. Returns ErrNotFound if missing, or
+	// ErrVersionConflict on version mismatch.
+	UpdateActorTemplate(ctx context.Context, template *ateapipb.ActorTemplate, expectedVersion int64) (*ateapipb.ActorTemplate, error)
```
juli4n: Can you sync with @laoj2? We are reworking the way store should expose Update<Type> methods. See #763

zoez7: Updated the UpdateActorTemplate impl in ed51c8dbf0c3da7da6a1df8a745d6efbc5397433 to follow the example in #829 .

pkg/proto/ateapipb/ateapi.proto
zoez7: Done.

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -260,6 +263,312 @@ func (s *Persistence) hasMatching(ctx context.Context, pattern string) (bool, er
...
+		return nil, "", err
+	}
+	return result, nextToken, nil
+}
+
+// DeleteActorTemplate deletes an ActorTemplate with no remaining versions.
+// Returns store.ErrNotFound if the template does not exist, or
+// store.ErrFailedPrecondition while any ActorTemplateVersion still names it
+// as parent.
+func (s *Persistence) DeleteActorTemplate(ctx context.Context, name string) (*ateapipb.ActorTemplate, error) {
```
HavenXia: I think here's a TOCTOU problem - if `DeleteActorTemplate` happens after the ActorTemplate check in but before the actual creation of `CreateActorTemplateVersion`, then we would have a orphaned ATV. 

This is similar to what I realize in `DeleteAtespace` and `CreateActor` in #396, this might need a same pattern to have a terminating marker that can prevent another operator from proceeding.

 Maybe add a todo for that TOCTOU?

zoez7: I was using API level locks, but I think adding deletion markers here makes sense too in case there is a bug in the API level locking. Added in 091cafba.

zoez7: I changed my mind and removed https://github.com/agent-substrate/substrate/commit/091cafbae05c59fefad5f7bc614c007d46caa0a9, after looking at the impl. I think this is worth it's own PR if we decide to use this method. Will rely on global locks instead for now.

Also, since we are moving to postgres, I think we can consider using transactions for ActorTemplate CRUD methods.

HavenXia: Hmm for this case global lock might be enough due to the relative low frequency of creating ATV.

I also thought about api level lock for crud of actor, but I later realized that will hurt performance of actor creation.

zoez7:
Failing tests at head: one of them is being fixed in https://github.com/agent-substrate/substrate/pull/857
Another one is flaky: https://github.com/agent-substrate/substrate/issues/675 ...

cmd/ateapi/internal/store/store.go
zoez7: Sync'd with Luiz offline, he lgtm'd.

zoez7:
I have not yet incorporated the resource limits added in https://github.com/agent-substrate/substrate/pull/679, will incorporate in a separate PR once #679 merges.

HavenXia (dismissed):

EItanya (approved):

internal/resources/actortemplateref.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"log/slog"
+
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+)
+
+// ActorTemplateRef identifies an ActorTemplate by the (atespace, name).
+//
+// ActorTemplateRef is the in-process form of the identity that
```
juli4n, line 25: This looks like a straight copy of ActorRef, no? Something we could consider is having an ResourceRef which implements it and then two type aliases for ActorTemplateRef and ActorTemplateRef (we will have more types in the future as well).

juli4n (approved):

internal/resources/actortemplateref.go
zoez7: Yes, It's mostly a copy of ActorRef. Will address this in the next PR in this chain.

PR #825, opened 2026-08-10
hack: add --help to create-kind-cluster.sh and fix a bashism

ygao-g:
hack/create-kind-cluster.sh is configured entirely through the environment, which was only discoverable by reading it. Add a usage block that names the variables and exits before touching Docker, so --help works even where Docker is not running.

Also switch the registry's running check from == to =, the POSIX test operator the rest of the script already uses.

Split out of a larger IPv6 branch so the address-family change that follows is purely about address families.

## Testing

- `hack/create-kind-cluster.sh --help`
- `hack/verify/boilerplate.sh`

No behaviour change to cluster creation.

### DIFF
--- a/hack/create-kind-cluster.sh
+++ b/hack/create-kind-cluster.sh
@@ -21,11 +21,24 @@ KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
 reg_name="kind-registry"
 reg_port="5001"
 
+if [ "$#" -gt 0 ]; then
+  case "$1" in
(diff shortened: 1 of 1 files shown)


BenTheElder:
> Also switch the registry's running check from == to =, the POSIX test operator the rest of the script already uses.

it's explicitly a bash script, let's use bashism?

ygao-g:
Fair — dropped it, the line is untouched now. Folded into #827, which is one commit and carries the `--help` block; closing this one.

PR #826, opened 2026-08-10
feat: Add control plane telemetry - workerpool desired and ready workers gauges

### LINKED ISSUE #564: Feature request: More system metrics for debuggability (Activation SLI, crash accounting, and capacity signals)

Angelawork:
Part of #564 (Part 4, Fixes #564)

- [x] Tests pass

## Description
This PR implements Part 4 of #564 by adding OpenTelemetry gauge instrumentation for `ate.workerpool.desired_workers` and `ate.workerpool.ready_workers` in `atecontroller`. 

## Key Changes:
* Added `ReadyReplicas int32` to `WorkerPoolStatus` with `+kubebuilder:printcolumn:name="Ready"` annotation in `pkg/api/v1alpha1/workerpool_types.go` and regenerated the CRD manifest.
* Updated `WorkerPoolReconciler.syncStatus` to synchronize `dep.Status.ReadyReplicas` into `wp.Status.ReadyReplicas`.
* Implemented `InitMetrics(meter)` registering `ate.workerpool.desired_workers` and `ate.workerpool.ready_workers` as OpenTelemetry Observable UpDownCounters (`{worker}`) with an asynchronous observer callback that samples controller-runtime's local informer cache (`r.Client.List`).
* Labeled metric datapoints using centralized attributes `ateattr.WorkerPoolNamespaceKey` and `ateattr.WorkerPoolNameKey`.

## Testing
* `make test`
## E2E  Test


cmd/atecontroller/internal/controllers/workerpool_controller_test.go
```diff
@@ -652,3 +652,53 @@ func eventually(t *testing.T, condition func(ctx context.Context) (bool, error))
...
+	if err := k8sClient.Get(t.Context(), types.NamespacedName{Name: wp.Name, Namespace: wp.Namespace}, current); err != nil {
+		t.Fatalf("get WorkerPool failed: %v", err)
+	}
+	if current.Status.ReadyReplicas != 2 {
+		t.Errorf("got ReadyReplicas=%d, want 2", current.Status.ReadyReplicas)
+	}
+	if current.Status.Replicas != 3 {
+		t.Errorf("got Replicas=%d, want 3", current.Status.Replicas)
+	}
+}
```
JeffLuoo: This test races the reconciler that `TestMain` starts, and passes only by winning a ~250 ms window.

  Creating the WorkerPool enqueues the live reconciler, which calls `syncStatus` with the *real* Deployment — all-zero status, since envtest runs no deployment controller. Your manual `syncStatus` writes
  `{3,2}`, then the reconciler overwrites it with `{0,0}`. Re-reading the status at intervals after the `syncStatus` call:

  ```
  after 0s    : Replicas=3 ReadyReplicas=2
  after 250ms : Replicas=0 ReadyReplicas=0
  after 1s    : Replicas=0 ReadyReplicas=0
  ```

  The bare `Get` on line 684 currently lands inside that window. Under CI load it won't.
  
  `TestStatusReplicasPropagation` (line 299) already has the pattern — drive the real Deployment status and assert with `eventually`. That also covers the real reconcile path instead of calling `syncStatus`
  directly, and lets the test run `t.Parallel()` like its neighbours.

  ```suggestion
  // TestSyncStatus_ReadyReplicas verifies that a Deployment reporting 3 replicas
  // with 2 ready propagates to WorkerPool.status.
  func TestSyncStatus_ReadyReplicas(t *testing.T) {
        t.Parallel()
        ctx := t.Context()
        wp := makeWorkerPool("test-sync-ready", "default", 3, "ateom:v1")
        if err := k8sClient.Create(ctx, wp); err != nil {
                t.Fatalf("create WorkerPool: %v", err)
        }
        deleteOnCleanup(t, wp)
  
        eventually(t, func(ctx context.Context) (bool, error) {
                _, err := getDeployment(ctx, wp)
                return err == nil, nil
        })
  
        // Simulate the deployment controller reporting 3 pods, 2 of them ready.
        updateDeploymentStatus(t, ctx, wp, "patch Deployment status", func(dep *appsv1.Deployment) {
                dep.Status.Replicas = 3
                dep.Status.ReadyReplicas = 2
        })
  
        eventually(t, func(ctx context.Context) (bool, error) {
                current := &atev1alpha1.WorkerPool{}
                if err := k8sClient.Get(ctx, types.NamespacedName{Name: wp.Name, Namespace: wp.Namespace}, current); err != nil {
                        return false, nil
                }
                return current.Status.Replicas == 3 && current.Status.ReadyReplicas == 2, nil
        })
  }
  ```

cmd/atecontroller/internal/controllers/workerpool_controller.go
```diff
@@ -139,8 +146,67 @@ func (r *WorkerPoolReconciler) syncStatus(ctx context.Context, wp *atev1alpha1.W
...
+	if err != nil {
+		return fmt.Errorf("create ate.workerpool.ready_workers instrument: %w", err)
+	}
+	r.readyWorkers = readyWorkers
+
+	_, err = meter.RegisterCallback(
+		func(ctx context.Context, obs metric.Observer) error {
+			if r.Client == nil {
+				return nil
+			}
```
JeffLuoo: nit - This guard is dead — every construction path sets `Client`, and if it were nil `r.List` would panic anyway at a more informative point.
  
  ```suggestion
  ```

cmd/atecontroller/internal/controllers/workerpool_controller_test.go
```diff
@@ -652,3 +652,53 @@ func eventually(t *testing.T, condition func(ctx context.Context) (bool, error))
...
+// TestWorkerPoolMetricsInitialization verifies that InitMetrics registers
+// the desired_workers and ready_workers instruments without error.
+func TestWorkerPoolMetricsInitialization(t *testing.T) {
+	r := &WorkerPoolReconciler{
+		Client: k8sClient,
+	}
+	if err := r.InitMetrics(nil); err != nil {
+		t.Fatalf("InitMetrics failed: %v", err)
+	}
+}
```
JeffLuoo: The unit test would not fail and should cover more. Updated unit test verifies metrics registration and metrics value.

  ```suggestion
  // TestWorkerPoolMetrics verifies that the registered callback observes
  // spec.replicas and status.readyReplicas per WorkerPool, labeled with the pool
  // namespace and name. A fake client keeps this off the envtest reconciler,
  // which would otherwise resync status out from under the assertion.
  func TestWorkerPoolMetrics(t *testing.T) {
        t.Parallel()
        scheme := runtime.NewScheme()
        if err := atev1alpha1.AddToScheme(scheme); err != nil {
                t.Fatalf("add scheme: %v", err)
        }
        wp := makeWorkerPool("test-metrics", "default", 4, "ateom:v1")
        wp.Status = atev1alpha1.WorkerPoolStatus{Replicas: 4, ReadyReplicas: 2}

        reader := sdkmetric.NewManualReader()
        r := &WorkerPoolReconciler{
                Client: fake.NewClientBuilder().WithScheme(scheme).WithObjects(wp).Build(),
        }
        if err := r.InitMetrics(sdkmetric.NewMeterProvider(sdkmetric.WithReader(reader)).Meter("test")); err != nil {
                t.Fatalf("InitMetrics: %v", err)
        }

        var rm metricdata.ResourceMetrics
        if err := reader.Collect(t.Context(), &rm); err != nil {
                t.Fatalf("collect metrics: %v", err)
        }

        wantAttrs := attribute.NewSet(
                ateattr.WorkerPoolNamespaceKey.String(wp.Namespace),
                ateattr.WorkerPoolNameKey.String(wp.Name),
        )
        got := map[string]int64{}
        for _, sm := range rm.ScopeMetrics {
                for _, m := range sm.Metrics {
                        sum, ok := m.Data.(metricdata.Sum[int64])
                        if !ok {
                                continue
                        }
                        for _, dp := range sum.DataPoints {
                                if dp.Attributes.Equals(&wantAttrs) {
                                        got[m.Name] = dp.Value
                                }
                        }
                }
        }
        want := map[string]int64{
                "ate.workerpool.desired_workers": 4,
                "ate.workerpool.ready_workers":   2,
        }
        if !reflect.DeepEqual(got, want) {
                t.Errorf("observed %v, want %v", got, want)
        }
  }
  ```

JeffLuoo (changes requested):

cmd/atecontroller/internal/controllers/workerpool_controller_test.go
Angelawork: Thanks for the detailed comment! Suggestion is applied.

cmd/atecontroller/internal/controllers/workerpool_controller.go
```diff
@@ -139,8 +146,64 @@ func (r *WorkerPoolReconciler) syncStatus(ctx context.Context, wp *atev1alpha1.W
...
+		return fmt.Errorf("create ate.workerpool.ready_workers instrument: %w", err)
+	}
+	r.readyWorkers = readyWorkers
+
+	_, err = meter.RegisterCallback(
+		func(ctx context.Context, obs metric.Observer) error {
+			var list atev1alpha1.WorkerPoolList
+			if err := r.List(ctx, &list); err != nil {
+				return nil
+			}
```
JeffLuoo, line 178: Consider log the error here, otherwise the error here will fail silently.

cmd/atecontroller/internal/controllers/workerpool_controller.go
```diff
@@ -139,8 +146,64 @@ func (r *WorkerPoolReconciler) syncStatus(ctx context.Context, wp *atev1alpha1.W
 	return nil
 }
 
+// InitMetrics initializes the OpenTelemetry instruments for ate.workerpool.desired_workers
+// and ate.workerpool.ready_workers and registers the asynchronous callback.
+func (r *WorkerPoolReconciler) InitMetrics(meter metric.Meter) error {
+	if meter == nil {
+		meter = otel.Meter("atecontroller")
+	}
```
JeffLuoo:
```suggestion
  ```

nit - We don't need this block. All `InitMetrics` won't pass a nil metrics meter.

**Review by JeffLuoo — CHANGES_REQUESTED** [review]

File: docs/observability.md [hunk 6]
```diff
@@ -114,6 +114,8 @@ Agent Substrate emits foundational OpenTelemetry system and server metrics to mo
...
 | `ate.scheduler.eligible_workers` | ateapi | histogram | number of eligible unassigned workers available during scheduling given the constraint filters (labels `ate.workerpool.namespace`, `ate.workerpool.name`, `ate.sandbox.class`, `ate.scheduling.constraint`) |
 | `atelet.snapshot.size` | atelet | histogram | uncompressed size in bytes of each gVisor snapshot image written during checkpoint (labels `file.name`, `ate.template.namespace`, `ate.template.name`) |
+| `ate.workerpool.desired_workers` | atecontroller | observable updowncounter | commanded worker pod capacity (`spec.replicas`) per WorkerPool (labels `ate.workerpool.namespace`, `ate.workerpool.name`, unit `{worker}`) |
+| `ate.workerpool.ready_workers` | atecontroller | observable updowncounter | delivered ready worker pod capacity (`status.readyReplicas`) per WorkerPool (labels `ate.workerpool.namespace`, `ate.workerpool.name`, unit `{worker}`) |
```
JeffLuoo: Nit - Use `up/down counter` to align with row below.

  ```suggestion
  | `ate.workerpool.desired_workers` | atecontroller | up/down counter | number of worker pods requested for a WorkerPool, from `spec.replicas` (labels
  `ate.workerpool.namespace`, `ate.workerpool.name`) |
  | `ate.workerpool.ready_workers` | atecontroller | up/down counter | number of worker pods currently ready for a WorkerPool, from `status.readyReplicas` (labels
  `ate.workerpool.namespace`, `ate.workerpool.name`) |
  ```

JeffLuoo (dismissed):

JeffLuoo (approved):

PR #827, opened 2026-08-10
hack: make the kind cluster's IP family configurable

ygao-g:
`KIND_IP_FAMILY` selects `ipv4`, `dual`, or `ipv6` and passes this through to kind. To prevent insidious failures like pods stuck in `ContainerCreating`, the script now drops old Docker networks to force IPv6 recreation if needed, validates the daemon's IPv6 support up-front, and enables `proxy_ndp` (alongside `proxy_arp`) to properly support gVisor loopback pod-to-pod networking across IPv6 nodes.

The networking setup is fully validated using the newly added `hack/verify-kind-networking.sh`, which confirms reliable IPv6 distribution across Node IPs, Service ClusterIPs, Pod IPs, Registry configurations, and CoreDNS lookups.

### DIFF
--- a/hack/create-kind-cluster.sh
+++ b/hack/create-kind-cluster.sh
@@ -18,8 +18,35 @@ set -o errexit -o nounset -o pipefail
 
 ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
+KUBECTL_CONTEXT="kind-${KIND_CLUSTER_NAME}"
 reg_name="kind-registry"
 reg_port="5001"
+
+if [[ $# -gt 0 ]]; then
+  case "$1" in
+    -h|--help)
+      echo "Usage: $0"
+      echo "Creates the kind cluster '${KIND_CLUSTER_NAME}' and a local registry container on port ${reg_port}."
+      echo
+      echo "Configured through the environment:"
+      echo "  KIND_CLUSTER_NAME  Name of the cluster to create (default: kind)."
+      echo "  KIND_IP_FAMILY     Address families for pods and Services: ipv4, ipv6 or dual (default: ipv4)."
+      exit 0
+      ;;
+  esac
+fi
+
+# Only ipFamily is set; kind's per-family podSubnet/serviceSubnet defaults are
+# already what we want.
+KIND_IP_FAMILY="${KIND_IP_FAMILY:-ipv4}"
+case "${KIND_IP_FAMILY}" in
+  ipv4|ipv6|dual)
+    ;;
+  *)
+    echo "error: KIND_IP_FAMILY must be one of ipv4, ipv6, dual (got '${KIND_IP_FAMILY}')" >&2
+    exit 1
+    ;;
+esac
 
 mkdir -p "${ROOT}/bin"
 
@@ -33,6 +60,9 @@ if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true
 fi
 
 if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != "true" ]; then
+  # Published on both loopback families so `ko` reaches localhost:5001 whichever
+  # one its resolver picks. The node side is separate: it goes over the "kind"
+  # network in step 4.
   docker run \
     -d --restart=always \
     --label created-by=agent-substrate \
@@ -57,7 +87,7 @@ else
   echo "/dev/kvm not available: micro-VM support disabled (gVisor still works)."
 fi
 
-echo "Creating kind configuration for cluster '${KIND_CLUSTER_NAME}'..."
+echo "Creating kind configuration for cluster '${KIND_CLUSTER_NAME}' (ipFamily=${KIND_IP_FAMILY})..."
 cat <<EOF > "${ROOT}/bin/kind-config.yaml"
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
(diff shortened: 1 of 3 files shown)


hack/create-kind-cluster.sh
```diff
@@ -70,10 +112,14 @@ else
   echo "/dev/kvm not available: micro-VM support disabled (gVisor still works)."
 fi
 
-echo "Creating kind configuration for cluster '${KIND_CLUSTER_NAME}'..."
+echo "Creating kind configuration for cluster '${KIND_CLUSTER_NAME}' (ipFamily=${KIND_IP_FAMILY})..."
 cat <<EOF > "${ROOT}/bin/kind-config.yaml"
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+networking:
```
ygao-g: Instead of hardcoding KIND ipv4 default subnet, can we just append this section when ipFamily is NOT ipv4

hack/create-kind-cluster.sh
```diff
@@ -21,6 +21,53 @@ KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
...
+      echo
+      echo "Configured through the environment:"
+      echo "  KIND_CLUSTER_NAME  Name of the cluster to create (default: kind)."
+      echo "  KIND_IP_FAMILY     Address families for pods and Services: ipv4, ipv6 or dual (default: ipv4)."
+      exit 0
+      ;;
+  esac
+fi
+
+# Address families the cluster's pods and Services get: ipv4 (the default, and
```
ygao-g: Evaluate whether it is better to move the declaration to right before where they get used

ygao-g: Make the comments more concise.

hack/create-kind-cluster.sh
```diff
@@ -57,10 +97,20 @@ else
...
+EOF
+if [[ "${KIND_IP_FAMILY}" != "ipv4" ]]; then
+  cat <<EOF >> "${ROOT}/bin/kind-config.yaml"
+networking:
+  ipFamily: ${KIND_IP_FAMILY}
+  podSubnet: "${pod_subnet}"
+  serviceSubnet: "${service_subnet}"
+EOF
+fi
+cat <<EOF >> "${ROOT}/bin/kind-config.yaml"
```
ygao-g: Check if order matters here. If not, we can put networking section at the end to reduce one cat command

ygao-g: Order doesn't matter — kind reads the whole file and does a single strict struct decode (`yamlUnmarshalStrict` into `v1alpha4.Cluster`), so top-level keys are position-independent; `KnownFields(true)` only rejects unknown keys. Moved `networking` after `runtimeConfig`, which lets `nodes:` join the opening heredoc — one `cat` fewer.

Checked all three families parse to identical `v1alpha4.Cluster` values before and after, and the ipv4 output is byte-for-byte unchanged.

hack/create-kind-cluster.sh
```diff
@@ -84,16 +124,54 @@ featureGates:
...
+  ipFamily: ${KIND_IP_FAMILY}
+  podSubnet: "${pod_subnet}"
+  serviceSubnet: "${service_subnet}"
+EOF
+fi
 
 echo "Deleting existing kind cluster '${KIND_CLUSTER_NAME}' if it exists..."
 "${ROOT}"/hack/kind.sh delete cluster --name "${KIND_CLUSTER_NAME}" || true
 
+# kind reuses an existing "kind" network as-is, so one left from an ipv4 run
```
ygao-g: Could this happen the other way, e.g., creating a kind ipv4 cluster while the docker network for previous dual-stack or IPv6 KIND clusters weren't cleaned up

ygao-g: In this case will it be easier to always cleanup and recreate?

hack/create-kind-cluster.sh
```diff
@@ -18,9 +18,37 @@ set -o errexit -o nounset -o pipefail
 
 ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
+KUBECTL_CONTEXT="${KUBECTL_CONTEXT:-kind-${KIND_CLUSTER_NAME}}"
```
ygao-g: why allow this to be configurable via env var? Should we just build it on top of cluster name?

hack/create-kind-cluster.sh
ygao-g: No — the asymmetry is real rather than an oversight. `ensureNetwork(name)` (`providers/docker/network.go:50` in kind v0.32.0, the version pinned in `hack/tools/kind/go.mod`) takes only a name, and `provider.go:71-77` calls it with the constant `"kind"`. `ipFamily` never reaches network creation at all — it only feeds node run args and port mappings.

So kind always attempts `--ipv6 --subnet <ULA>` and falls back to v4-only in exactly one case: `isIPv6UnavailableError`, i.e. the daemon has IPv6 disabled. A network left over from a `dual` or `ipv6` run therefore always has IPv6, and reusing it for an `ipv4` cluster is harmless — the nodes pick up an extra v6 address on the bridge that kubeadm, configured v4-only, ignores.

Worth noting that `network.go:58` is literally `// TODO: the network might already exist and not have ipv6 ... :|`, linking the upstream discussion. This guard is the workaround for that TODO, which is why it only needs to fire in the one direction.

ygao-g: I'd lean against it. The network name is fixed to `"kind"` for every kind cluster (`network.go:47`), but `KIND_CLUSTER_NAME` here is configurable — so with a second cluster up, its nodes are attached to that same network and `docker network rm kind` fails. Unconditional teardown would turn a rare, targeted failure into one that fires on every run for anyone keeping two clusters around.

The conditional pays the disconnect/rm/reconnect cost only when it's actually required, and the error path already prints the `docker network inspect` command to find whatever else is attached.

hack/create-kind-cluster.sh
```diff
@@ -84,16 +114,53 @@ featureGates:
 runtimeConfig:
   "certificates.k8s.io/v1beta1": "true"
 EOF
+# Last so the ipv4 path is a single heredoc; kind decodes the whole file into one
+# struct, so the position of the networking section does not matter.
+if [[ "${KIND_IP_FAMILY}" != "ipv4" ]]; then
```
ygao-g: wonder if we can add the networking configurations unconditionally. Does it also works with ipv4 specified as ipFamily?

hack/create-kind-cluster.sh
ygao-g: Done — `podSubnet` and `serviceSubnet` are gone entirely; the block is just `ipFamily` now, emitted only for non-ipv4.

It went further than the ipv4 arm, though: `pkg/apis/config/v1alpha4/default.go:51-72` (v0.32.0) defaults to `fd00:10:244::/56` and `fd00:10:96::/112` for ipv6, and the comma-joined pair for dual — byte-for-byte what the script was writing. All six values were kind's own defaults, so all six came out. The old comment's claim that `/112` was the largest kube-apiserver accepts was also wrong; it's just kind's default.

hack/create-kind-cluster.sh
ygao-g: Moot after the rewrite — there's no declaration left, only validation of `KIND_IP_FAMILY`. Kept at the top so a bad value fails immediately rather than after the `/dev/kvm` probe and registry setup.

ygao-g: Done — 22 lines to 11, one comment instead of three.

hack/create-kind-cluster.sh
ygao-g: Already answered in-thread and the change is in; resolving.

hack/create-kind-cluster.sh
ygao-g: Agreed — now a plain `KUBECTL_CONTEXT="kind-${KIND_CLUSTER_NAME}"`, and the `--help` entry is dropped since it's no longer a knob. `KINDClusterKey` is unconditionally `"kind-" + clusterName` (`kubeconfig/internal/kubeconfig/helpers.go:24-25`), and this script creates the cluster, so it always knows the context; an override could only ever point the `kubectl apply` at a cluster other than the one just created.

hack/create-kind-cluster.sh
ygao-g: Yes — `ipv4` is an explicitly valid `ipFamily` (`internal/apis/config/validate.go:56`), and it's exactly what kind defaults the field to when absent (`v1alpha4/default.go:39-41`), so writing it decodes to the identical `v1alpha4.Cluster`.

Done: `networking` now folds into the trailing heredoc unconditionally, which drops the `if` and the second `cat`. The only thing given up is that the ipv4 config file is no longer byte-for-byte identical to pre-PR — but the decoded config is, which was the property that mattered.

ygao-g:
@BenTheElder @bowei PTAL

ygao-g:
Closing this one — the thread here is all self-review noise from iterating on the script, with no reviewer feedback to carry over. Refiled as #877 with the same commit for a clean read.

PR #829, opened 2026-08-10
Fix toctou race in UpdateActor

laoj2:
UpdateActor now takes an ActorRef, an ActorPrecondition, and a mutate callback. The store reads the stored actor inside the WATCH transaction, checks the precondition against that value, and hands it to mutate, which edits it in place.

The storage layer retries up to 5 times when a concurrent write invalidates it, re-running mutate against the newer state

I still need to migrate the other callers of store.UpdateActor to avoid calling GetActor outside of the closure function.

#763

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/update_actor.go
+++ b/cmd/ateapi/internal/controlapi/update_actor.go
@@ -41,30 +41,22 @@ func (s *Service) UpdateActor(ctx context.Context, req *ateapipb.UpdateActorRequ
 	actorRef := resources.ActorRefFromActor(in)
 	setSpanActorRefAttributes(ctx, actorRef)
 
-	actor, err := s.persistence.GetActor(ctx, actorRef)
-	if err != nil {
-		if errors.Is(err, store.ErrNotFound) {
-			return nil, status.Errorf(codes.NotFound, "Actor %s not found", actorRef)
+	updated, err := s.persistence.UpdateActor(ctx, actorRef, func(dbActor *ateapipb.Actor) error {
+		if err := store.CheckActorPrecondition(dbActor, in.GetMetadata().GetUid(), in.GetMetadata().GetVersion()); err != nil {
+			return err
 		}
-		return nil, fmt.Errorf("while getting actor: %w", err)
-	}
-
-	// UID and version preconditions
-	if uid := in.GetMetadata().GetUid(); uid != "" && uid != actor.GetMetadata().GetUid() {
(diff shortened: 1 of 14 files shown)


cmd/ateapi/internal/store/store.go
```diff
@@ -51,10 +58,21 @@ type Interface interface {
...
+	// mutate receives the stored actor and edits it in place. The mutated actor is
+	// written iff mutate returns nil.
+	//
+	// mutate may run more than once, because the store retries when a concurrent
+	// write invalidates the transaction.
+	//
+	// Returns ErrNotFound if missing, ErrUIDConflict or ErrVersionConflict if the
+	// precondition rejects the stored actor, ErrVersionConflict if the retry
+	// budget is exhausted, or the mutate's error verbatim otherwise.
+	UpdateActor(ctx context.Context, actorRef resources.ActorRef, precondition ActorPrecondition, mutate func(dbActor *ateapipb.Actor) error) (*ateapipb.Actor, error)
```
juli4n: Do we need `ActorPrecondition`? The `mutate` calback can do the check already, and if we want to reuse some preconditions check we can have `WithActorPrecondition` wrap a given `mutate` callback with another one that does the check first and then delegates? WDYT?

cmd/ateapi/internal/store/store.go
```diff
@@ -51,10 +58,21 @@ type Interface interface {
 	// mutated. Returns ErrAlreadyExists if key is taken.
 	CreateActor(ctx context.Context, actor *ateapipb.Actor) (*ateapipb.Actor, error)
 
-	// Updates actor state with optimistic concurrency check and returns the stored
-	// resource with advanced metadata (version, update_time). The input is not
-	// mutated. Returns ErrNotFound if missing, or ErrVersionConflict on version mismatch.
-	UpdateActor(ctx context.Context, actor *ateapipb.Actor, expectedVersion int64) (*ateapipb.Actor, error)
+	// UpdateActor performs a transactional read-modify-write and returns the stored
+	// resource with advanced metadata (version, update_time).
```
juli4n: "... and returns the updated actor".

juli4n (approved):

cmd/ateapi/internal/controlapi/update_actor.go
thockin, line 40: I am still not convinced this is right.  We are doing this GET outside of any transaction, then capture that in a func which is itself called with the result of another GET from inside the txn.

Shouldn't `in` go away and be replaced entirely by dbActor ?

thockin: I misread this, sorry.  My mistake.

PR #834, opened 2026-08-10
imagecache: add ate.imagecache.requests hit and miss telemetry (#831)

### LINKED ISSUE #831: Add `ate.imagecache.requests`: hit and miss telemetry for the node-local image cache
## Summary

The node-local OCI image cache (`internal/imagecache`) emits no telemetry today.
  
The cache is on the Resume path: atelet calls `Store.EnsureImage` before it assembles the OCI bundle for an Actor. A hit costs no network I/O. A miss causes a pull and an unpack, which can take tens of seconds for a large image.

  ## Metric

  | Field | Value |
  |---|---|
  | Name | `ate.imagecache.requests` |
  | Instrument | `Int64Counter` |
  | Unit | `{request}` |
  | Description | Number of image lookups in the node-local image cache, by outcome. |
  | Golden signal | Saturation (leading indicator for latency) |
  | Emitted by | atelet (`internal/imagecache`, `Store.EnsureImage`) |

  ### Attributes
  
  | Key | Values | Notes |
  |---|---|---|
  | `ate.imagecache.outcome` | `hit`, `miss` | New key. A hit is a complete image record with all layer directories present. A miss is each other result. |
  | `error.type` | standardized error values | Set it only when the lookup fails. Its absence means success. |

JeffLuoo:
Add the `ate.imagecache.requests` counter.

  **`ate.imagecache.outcome`** (new key in `internal/ateattr`)

  | Value | Meaning |
  |---|---|
  | `hit` | the node holds a complete image record: each layer directory that the record names is present |
  | `miss` | the lookup must pull |
  | `error` | the lookup failed; the only outcome that carries `error.type` |
  | `cancelled`, `timeout` | the caller gave up, so the cache is not at fault |

  A failed lookup is neither a hit nor a miss, so it gets its own outcome, as
  `no_free_worker` does on `ate.scheduler.outcome`. `cancelled` and `timeout` are
  outcomes for the same reason they are on `ate.router.outcome`. The hit ratio is
  therefore `hit / (hit + miss)`, with failures and abandoned lookups out of the
  denominator.

  **`error.type`** — set only on the `error` outcome.

  | Value | Meaning |
  |---|---|
  | `404`, `401`, `429`, ... | the registry rejected the request; its own HTTP status, reported verbatim |
  | `_OTHER` | the failure carries no status of its own |

Fixes #831

- [x] Tests pass

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -123,6 +123,7 @@ Agent Substrate emits foundational OpenTelemetry system and server metrics to mo
 | `ate.scheduler.assignment.duration` | ateapi | histogram | time it takes for an actor to be assigned to a worker, per attempt (version-conflict retries record only the final attempt), with the outcome (`assigned` / `no_free_worker` / `error`) and sandbox class to catch scheduling latency and capacity starvation problems |
 | `ate.actor.restore.duration` | atelet | histogram | how long each phase of a restore takes on the worker node, which is where cold-start latency actually goes once ateapi hands off (labels `ate.snapshot.phase`, `ate.snapshot.kind`, `ate.snapshot.scope`, `ate.template.namespace`, `ate.template.name`, `ate.sandbox.class`, plus `ate.failure.reason` on failure) |
 | `ate.actor.checkpoint.duration` | atelet | histogram | the same phase breakdown for writing a snapshot, so a slow suspend can be attributed to ateom or to the upload (same labels as the restore histogram) |
+| `ate.imagecache.requests` | atelet | counter | image lookups in the node-local image cache, by outcome (`ate.imagecache.outcome`), with `error.type` on the `error` outcome. A miss pays for the pull and the unpack, so the hit ratio per node is a leading indicator of resume latency |
 
 The table lists the OpenTelemetry instrument names. How a name appears in a query depends on the backend (Cloud Monitoring (GMP) / Kind collector).
 
@@ -136,6 +137,10 @@ For `atenet.router.route.duration`:
 
 For `ate.scheduler.eligible_workers`:
 * `ate.scheduling.constraint` categorizes the scheduling request constraint type: `none` (unconstrained), `selector` (actor or template label selectors specified), or `required_nodes` (pinned to specific node VMs).
+
+For `ate.imagecache.requests`:
+* `ate.imagecache.outcome` is `hit` when the node holds a complete image record — every layer directory the record names is present — and `miss` when the lookup must pull. A failed lookup is neither: `error` is a failed lookup whatever the cause, and `cancelled` or `timeout` is the caller giving up, as on `ate.router.outcome`. So the hit ratio is `hit / (hit + miss)`, with failures and abandoned lookups out of the denominator.
+* `error.type` is present only on the `error` outcome, and carries the registry's own HTTP status for its rejection, from a fixed set: `401`, `403`, `404`, `429`, `500`, `502`, `503`, `504`. The set is an allow-list because the registry client reports whatever the remote returned. Each other status, and each failure that carries no status, reports `_OTHER`.
 
 The three snapshot labels are orthogonal and mean the same thing on every histogram that carries them:
 * `ate.snapshot.kind`: which snapshot the operation reads or writes. `local` (node-local, written by a pause), `latest` (the actor's own durable snapshot), `golden` (the template's image), or `boot` (from scratch, so it never appears on the atelet histograms).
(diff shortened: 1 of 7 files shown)


JeffLuoo:
Also cc  @krisztianfekete to review it as well.

git286 (dismissed):

internal/imagecache/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return ateattr.ImageCacheOutcomeError
+}
+
+// errorType reports the registry's own status for its rejection, the only
+// domain status this path has. Each other failure carries none.
+func errorType(err error) string {
+	var terr *transport.Error
+	if errors.As(err, &terr) && terr.StatusCode != 0 {
+		return strconv.Itoa(terr.StatusCode)
```
krisztianfekete, line 97: Not sure if the Go client is actually bounded, can we use an allow-list and fold all the rest into an `other` value?

docs/observability.md
```diff
@@ -129,6 +130,10 @@ For `atenet.router.route.duration`:
 For `ate.scheduler.eligible_workers`:
 * `ate.scheduling.constraint` categorizes the scheduling request constraint type: `none` (unconstrained), `selector` (actor or template label selectors specified), or `required_nodes` (pinned to specific node VMs).
 
+For `ate.imagecache.requests`:
+* `ate.imagecache.outcome` is `hit` when the node holds a complete image record — every layer directory the record names is present — and `miss` when the lookup must pull. A failed lookup is neither: `error` is a fault of the cache, and `cancelled` or `timeout` is the caller giving up, as on `ate.router.outcome`. So the hit ratio is `hit / (hit + miss)`, with failures and abandoned lookups out of the denominator.
+* `error.type` is present only on the `error` outcome, and carries the registry's own HTTP status for its rejection (`404`, `401`, `429`, ...), the only domain status this path has. A failure with no status reports `_OTHER`; what broke is on `ate.actor.restore.duration{ate.snapshot.phase="oci_unpack"}` with the substrate failure reason beside it.
```
krisztianfekete: Do we actually wrap an ateerrors reason here?

internal/ateattr/ateattr.go
```diff
@@ -103,6 +107,20 @@ const (
 	RouterResumeJoined = "joined"
 )
 
+// Values for ImageCacheOutcomeKey. A hit is a complete image record: every
+// layer directory it names is present. A miss must pull. A failed lookup is
+// neither, so it is its own outcome, as no_free_worker is on the scheduler:
+// only Error is a fault of the cache, and only Error carries an error.type.
```
krisztianfekete: > only Error is a fault of the cache

Nit: technically, not always, right?

internal/imagecache/metrics.go
JeffLuoo: Thanks for catching it. Bound the list.

docs/observability.md
JeffLuoo: updated

internal/ateattr/ateattr.go
JeffLuoo: reworded

docs/observability.md
```diff
@@ -129,6 +130,10 @@ For `atenet.router.route.duration`:
 For `ate.scheduler.eligible_workers`:
 * `ate.scheduling.constraint` categorizes the scheduling request constraint type: `none` (unconstrained), `selector` (actor or template label selectors specified), or `required_nodes` (pinned to specific node VMs).
 
+For `ate.imagecache.requests`:
+* `ate.imagecache.outcome` is `hit` when the node holds a complete image record — every layer directory the record names is present — and `miss` when the lookup must pull. A failed lookup is neither: `error` is a failed lookup whatever the cause, and `cancelled` or `timeout` is the caller giving up, as on `ate.router.outcome`. So the hit ratio is `hit / (hit + miss)`, with failures and abandoned lookups out of the denominator.
+* `error.type` is present only on the `error` outcome, and carries the registry's own HTTP status for its rejection (`404`, `401`, `429`, ...), the only domain status this path has. A status outside that set, and each failure that carries no status, reports `_OTHER`.
```
krisztianfekete: Can you please make this reflect `reportedStatuses`?

JeffLuoo: Sure, updated!

krisztianfekete (approved):
Think this just needs a rebase and it's good to go!

krisztianfekete (approved):

git286 (approved):

PR #835, opened 2026-08-10
Remove the secretKeyRef env sources feature

zoez7:
This reverts the ActorTemplate valueFrom.secretKeyRef support added in #20 (issue #15).

We don't want actors to have any access to secrets, they will be injected on the egress route instead. Removing this now so we don't need to copy secrets into substrate resources.


EItanya:
Can we hold off on this until egress policy merges? We still need some way to get secrets for testing agents which use LLMs

zoez7:
@EItanya Do you have ETA for egress policy merge? Once we cut over to substrate resources instead of CRDs, the workflow that rely on the secretKeyRef will break.

EItanya:
> Eitan Yarmush (@EItanya) Do you have ETA for egress policy merge? Once we cut over to substrate resources instead of CRDs, the workflow that rely on the secretKeyRef will break.

Ok that’s fair, we can just remove it for now, I will review tomorrow

EItanya (dismissed):

EItanya (approved):

PR #836, opened 2026-08-10
atelet: watermark-driven image cache GC loop

igooch:
Phase 2 of #463, continuing from #735 (the eviction engine): the watermark-driven loop in atelet that turns eviction on, plus the flags that govern it. After this: observability (metrics/spans), the ops-tool swap, and the e2e suite.

## What

- A serialized, panic-recovered pass every `--image-cache-gc-period` (default 5m; `0` disables the loop entirely). Each tick: `statfs` the cache volume + sum the pool's recorded sizes, compute a byte target — down to `--image-cache-low-percent` when usage crossed `--image-cache-high-percent`, and/or down to `--image-cache-max-bytes` — and hand it to `Store.EvictUnused`.
- **The target is capped at the pool's own size.** The cache is one tenant of a shared volume; uncapped kubelet-style watermark math on a pressured volume asks a tiny cache to fix pressure it didn't cause, evicting everything every tick.
- **Gated passes are detected by contract, not inference**: the engine exports `ErrIncompleteEnumeration` (wrapped into every gate return), and `classifyGCPass` — a pure, table-tested function — maps each pass to skipped/shortfall/complete/quiet. A gated pass logs ERROR "pass skipped" and never feeds the shortfall backoff.
- **Genuine shortfall warns with backoff** then decays to a periodic reminder: on a volume under foreign pressure, shortfall is the steady state and must not ERROR every tick.
- atelet passes `WithActorsDir(ateompath.ActorsDir)` so the root set sees placed actors (the wiring #735's notes required); `ateompath` gains the `ActorsDir` constant.
- The first pass runs immediately at boot (a node starting under disk pressure shouldn't wait a full period), and a `CacheSize` failure skips the pass explicitly rather than silently zeroing the target.
- `--image-cache-gc-dry-run` computes and logs every decision while deleting nothing — the production soak mechanism. A cache dir outside `BasePath` logs a warning (its watermarks would measure a different volume than actor state).
- README: the GC section's "loop lands next" intro replaced with the loop/flag documentation.

## For reviewers: the enabled-by-default question

Defaults ship **enabled** (5m / 85% / 80%, mirroring kubelet so operator intuition transfers), with dry-run as the opt-out soak path. The alternative is dry-run-by-default for a release. The target cap removes the known pathological case; happy to flip the default if you'd rather soak first.

## Testing

- `imagegc_test.go`: watermark/cap target-math table (including the capped-target cases), flag validation.
- Loop + engine together, Kind: full gVisor e2e suite green under aggressive flags (15s period, 1-byte cap); planted orphan ignored by periodic passes and reclaimed at startup; shortfall backoff observed against an impossible cap; no enumeration-gate false positives.
- GKE (6 nodes): idle pool evicted under aggressive watermarks, subsequent resumes cold-pulled cleanly with snapshot memory continuity; flags restored to defaults after.

### DIFF
--- a/cmd/atelet/imagegc.go
+++ b/cmd/atelet/imagegc.go
@@ -26,6 +26,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
@@ -64,13 +65,10 @@ func validateImageCacheGCFlags() error {
 	if *imageCacheLowPct < 0 || *imageCacheLowPct >= *imageCacheHighPct {
 		return fmt.Errorf("--image-cache-low-percent %d must be in [0,%d)", *imageCacheLowPct, *imageCacheHighPct)
 	}
-	// The watermark is measured on the cache dir's filesystem while the root
-	// set is read from the actors dir. If an operator points the cache at a
-	// different volume than BasePath, those are different filesystems: the
-	// pass would evict based on disk pressure the cache doesn't contribute
-	// to. Warn rather than fail — a separate cache volume is a legitimate
-	// (and recommended, for IOPS) configuration, it just wants its own
-	// watermarks.
+	// Outside BasePath the watermarks measure a different volume than
+	// actor state, so the pass would chase pressure the cache doesn't
+	// contribute to. Warn, don't fail: a separate cache volume is
+	// legitimate (recommended for IOPS) — it just wants its own numbers.
 	if !strings.HasPrefix(*imageCacheDir, ateompath.BasePath+string(os.PathSeparator)) {
 		slog.Warn("Image cache dir is outside the ateom base path; its volume watermarks are measured separately from actor state",
 			slog.String("image_cache_dir", *imageCacheDir),
@@ -79,24 +77,18 @@ func validateImageCacheGCFlags() error {
 	return nil
 }
 
-// imageCacheGCTarget computes the bytes an eviction pass should free.
+// imageCacheGCTarget computes the bytes a pass should free: the larger
+// of the watermark shortfall (kubelet's formula — usage at highPct frees
+// down to lowPct) and the pool's overage past maxBytes, but never more
+// than the cache actually holds.
 //
-// Watermark half (kubelet's formula): when volume usage is at or above the
-// high watermark, free down to the low one. Cap half: when the pool's
-// recorded size exceeds maxBytes, free the difference. The pass pursues the
-// larger — but never more than the cache actually holds.
-//
-// That cache-size ceiling is the important difference from kubelet, which
-// owns its imagefs and can therefore assume the whole shortfall is its to
-// free. Our cache is one tenant of a volume it shares with containerd's
-// image store, kubelet, logs, actor uppers and local snapshots, so the raw
-// watermark target asks the cache to free far more than it holds (measured:
-// a 105 GiB volume at 98% yields an 18.9 GiB target against an 11 MiB
-// cache). The pass would then evict every unrooted image on every tick —
-// a permanent 0% hit rate, turning every actor start back into a full
-// re-pull — while barely moving disk usage. Capped at its own size, the
-// cache gives back everything it can and no more; the residual shortfall
-// is reported (it is someone else's disk), not chased.
+// The cache-size ceiling is the difference from kubelet, which owns its
+// imagefs. This cache is one tenant of a shared volume, so the raw
+// watermark target can dwarf it (measured: a 105 GiB volume at 98% asks
+// an 11 MiB cache for 18.9 GiB), and an uncapped pass would evict
+// everything every tick for a permanent 0% hit rate. Capped, the cache
+// gives back all it can; the residual shortfall is someone else's disk —
+// reported, not chased.
 func imageCacheGCTarget(capacity, available uint64, cacheSize, maxBytes int64, highPct, lowPct int) int64 {
 	var target int64
 	if capacity > 0 {
@@ -125,10 +117,13 @@ func imageCacheGCTarget(capacity, available uint64, cacheSize, maxBytes int64, h
 // is done. Passes are strictly serialized: a slow pass delays the next
 // tick rather than overlapping it.
 func runImageCacheGC(ctx context.Context, store *imagecache.Store, cacheDir string) {
+	consecutiveShortfalls := 0
+	// First pass immediately: a node booting under disk pressure must not
+	// wait a full period (startup recovery reclaims debris, not pressure).
+	runImageCacheGCPass(ctx, store, cacheDir, &consecutiveShortfalls)
+
 	ticker := time.NewTicker(*imageCacheGCPeriod)
 	defer ticker.Stop()
-
-	consecutiveShortfalls := 0
 	for {
 		select {
 		case <-ctx.Done():
@@ -160,13 +155,13 @@ func runImageCacheGCPass(ctx context.Context, store *imagecache.Store, cacheDir 
 	capacity := st.Blocks * uint64(st.Bsize)
 	available := st.Bavail * uint64(st.Bsize)
 
-	// A sizing error is not fatal to the pass: the watermark half of the
-	// target needs only statfs.
+	// Same failure class as the enumeration gates (ReadDir of the layer
+	// pool): fail toward retention, retry next tick.
 	cacheSize, err := store.CacheSize()
 	if err != nil {
-		slog.WarnContext(ctx, "Image cache GC: sizing the pool failed; watermark target only this pass",
+		slog.WarnContext(ctx, "Image cache GC: sizing the pool failed; skipping this pass",
 			slog.Any("err", err))
-		cacheSize = 0
+		return
 	}
 
 	target := imageCacheGCTarget(capacity, available, cacheSize, *imageCacheMaxBytes, *imageCacheHighPct, *imageCacheLowPct)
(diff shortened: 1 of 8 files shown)


cmd/atelet/imagegc.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/spf13/pflag"
+	"golang.org/x/sys/unix"
+)
+
+var (
+	imageCacheGCPeriod = pflag.Duration("image-cache-gc-period", 5*time.Minute, "How often to run the image cache eviction pass. 0 disables eviction entirely.")
+	imageCacheHighPct  = pflag.Int("image-cache-high-percent", 85, "Cache-volume usage percentage above which eviction starts.")
+	imageCacheLowPct   = pflag.Int("image-cache-low-percent", 80, "Cache-volume usage percentage eviction frees down to. Must be lower than --image-cache-high-percent.")
+	imageCacheMaxBytes = pflag.Int64("image-cache-max-bytes", 0, "Absolute cap on the summed size of cached layers, evicted down to independently of the volume watermarks. 0 means no cap.")
+	imageCacheMinAge   = pflag.Duration("image-cache-min-age", 2*time.Minute, "Layers and image records younger than this are never evicted (protects images pulled but not yet mounted).")
```
dberkov, line 49: A negative `--image-cache-min-age` is accepted but inverts the protection: `cutoff := time.Now().Add(-s.minAge)` lands in the future, making everything — including a layer pulled milliseconds ago — evictable. That's exactly the window min-age exists to close. Worth rejecting `< 0` in `validateImageCacheGCFlags`.

cmd/atelet/imagegc.go
```diff
@@ -0,0 +1,241 @@
...
+		return fmt.Errorf("--image-cache-high-percent %d out of range [1,100]", *imageCacheHighPct)
+	}
+	if *imageCacheLowPct < 0 || *imageCacheLowPct >= *imageCacheHighPct {
+		return fmt.Errorf("--image-cache-low-percent %d must be in [0,%d)", *imageCacheLowPct, *imageCacheHighPct)
+	}
+	// Outside BasePath the watermarks measure a different volume than
+	// actor state, so the pass would chase pressure the cache doesn't
+	// contribute to. Warn, don't fail: a separate cache volume is
+	// legitimate (recommended for IOPS) — it just wants its own numbers.
+	if !strings.HasPrefix(*imageCacheDir, ateompath.BasePath+string(os.PathSeparator)) {
```
dberkov: Unclean paths defeat this check: `strings.HasPrefix` on the raw flag value misfires for `/var/lib/ateom-gvisor//image-cache` or `.../foo/../image-cache`. `filepath.Clean` the flag first. Warning-only, so cosmetic.

cmd/atelet/imagegc.go
```diff
@@ -0,0 +1,241 @@
...
+// imagefs. This cache is one tenant of a shared volume, so the raw
+// watermark target can dwarf it (measured: a 105 GiB volume at 98% asks
+// an 11 MiB cache for 18.9 GiB), and an uncapped pass would evict
+// everything every tick for a permanent 0% hit rate. Capped, the cache
+// gives back all it can; the residual shortfall is someone else's disk —
+// reported, not chased.
+func imageCacheGCTarget(capacity, available uint64, cacheSize, maxBytes int64, highPct, lowPct int) int64 {
+	var target int64
+	if capacity > 0 {
+		usedPct := 100 - int(available*100/capacity)
```
dberkov, line 103: Nit: this floors the available percentage, so eviction triggers up to ~1% before the nominal watermark (84.01% used reads as 85%). Fine in practice — a comment would prevent a future "off by one" report.

cmd/atelet/imagegc.go
```diff
@@ -0,0 +1,241 @@
...
+		slog.Int("orphan_layers", stats.OrphanLayers),
+		slog.Int("skipped_rooted", stats.SkippedRooted),
+		slog.Int("skipped_fresh", stats.SkippedFresh),
+		slog.Int64("cache_size_bytes", cacheSize),
+		slog.Bool("dry_run", *imageCacheGCDryRun),
+		slog.Duration("took", time.Since(tStart)),
+	}
+	outcome := classifyGCPass(err, target, stats.FreedBytes)
+	if outcome == gcPassSkipped {
+		slog.ErrorContext(ctx, "Image cache GC pass skipped", append(attrs, slog.Any("err", err))...)
```
dberkov, line 226: Double ERROR per gated tick: `EvictUnused` already logs ERROR internally ("eviction pass skipped: bundle specs could not be fully enumerated", gc.go) and this line logs ERROR again — a persistent bad record produces two ERROR lines every 5 minutes. Keep one; I'd drop the engine-side logs now that the sentinel exists, since this line carries the full attrs plus the wrapped error.

Relatedly: even with zero target, every tick pays `CacheSize` plus `EvictUnused`'s full root-set scan and record enumeration before the target check. The "gated wins even with zero target" test suggests that's intentional (corruption surfaces before pressure does) — if so, a one-line comment here would keep someone from "optimizing" it away.

cmd/atelet/imagegc_test.go
```diff
@@ -0,0 +1,166 @@
...
+		})
+	}
+}
+
+func TestValidateImageCacheGCFlags(t *testing.T) {
+	setFlags := func(high, low int) {
+		*imageCacheHighPct = high
+		*imageCacheLowPct = low
+	}
+	defer setFlags(85, 80)
```
dberkov: Nit: `t.Cleanup` is the house style for this elsewhere in the repo. Also, boundary cases `high=100` / `low=0` are legal-but-untested.

cmd/atelet/imagegc.go
```diff
@@ -0,0 +1,241 @@
...
+
+		runImageCacheGCPass(ctx, store, cacheDir, &consecutiveShortfalls)
+	}
+}
+
+// runImageCacheGCPass performs one pass. It recovers from panics: this is a
+// background janitor, and a bug here (or a malformed directory an operator
+// dropped into the pool) must not take atelet down with it and strand every
+// actor on the node.
+func runImageCacheGCPass(ctx context.Context, store *imagecache.Store, cacheDir string, consecutiveShortfalls *int) {
```
dberkov: `consecutiveShortfalls *int` is mutable loop state threaded as an out-param, and the idiomatic home for that is a small receiver:

```go
type imageCacheGC struct {
    store    *imagecache.Store
    cacheDir string
    consecutiveShortfalls int
}

func (g *imageCacheGC) Run(ctx context.Context)     // ticker loop
func (g *imageCacheGC) runPass(ctx context.Context) // one pass
```

Today it is only one int, so this is marginal on its own — but the observability phase will add exactly this kind of state (pass counters, last-pass duration, metric handles), and each becomes another pointer parameter without a struct. If the constructor also snapshots the flag values into fields, the pass logic stops reading package-level flag globals, which would make `runPass` unit-testable without the setFlags/restore dance.

To be clear on scope: the pure decision functions (`imageCacheGCTarget`, `classifyGCPass`) are the right split and should stay free functions — this is only about the loop state.

cmd/atelet/imagegc.go
igooch: Fixed. Rejected in validateImageCacheGCFlags with the inversion noted; covered in the flag table test.

cmd/atelet/imagegc.go
igooch: Fixed. filepath.Clean before the prefix check.

cmd/atelet/imagegc.go
igooch: Comment added.

cmd/atelet/imagegc.go
igooch: Took your suggestion of dropping the engine-side logs — all four gates (RecoverOrphans included) now only return the sentinel, and each caller logs once: this line for the periodic pass, New at ERROR for the startup scan. The zero-target intent is documented at the call, with the cost stated accurately (full root-set scan plus a read per record — you were right that "two ReadDirs" undersold it).

cmd/atelet/imagegc_test.go
igooch: Fixed. Table with t.Cleanup, plus the high=100/low=0 boundary and a negative min-age case.

cmd/atelet/imagegc.go
igooch: Done. imageCacheGC holds the loop state, flag values snapshotted at construction so runPass reads no globals; imageCacheGCTarget/classifyGCPass stay free functions per your scoping.

internal/imagecache/imagecache.go
```diff
@@ -224,10 +224,11 @@ func New(root string, opts ...Option) (*Store, error) {
...
 	}
 	// Startup-only orphan recovery (see RecoverOrphans for why it must not
-	// run during normal operation). Failure is logged inside, never fatal:
-	// a corrupt record must not keep atelet from serving actors.
+	// run during normal operation). Never fatal: a corrupt record must not
+	// keep atelet from serving actors. Orphans (if any) persist until the
+	// named file is repaired and the next restart.
 	if _, err := s.RecoverOrphans(context.Background()); err != nil {
-		slog.Warn("Image cache startup orphan recovery incomplete", slog.Any("err", err))
+		slog.Error("Image cache startup orphan scan skipped", slog.Any("err", err))
```
dberkov: This mislabels and escalates per-item recovery errors. `RecoverOrphans` returns two distinct shapes: the gates (wrapping `ErrIncompleteEnumeration` — the scan did nothing) and per-item errors from `sweepOrphanLayers`/`removeRetiredDirs` via `errors.Join(errs...)` — the scan ran and may have reclaimed most orphans. For the per-item case, ERROR "scan skipped" is wrong twice: the scan wasn't skipped, and a mostly-successful sweep now logs at ERROR (the old code had it at WARN). Same gate-vs-per-item split the loop already does:

```go
if _, err := s.RecoverOrphans(context.Background()); err != nil {
	if errors.Is(err, ErrIncompleteEnumeration) {
		slog.Error("Image cache startup orphan scan skipped", slog.Any("err", err))
	} else {
		slog.Warn("Image cache startup orphan recovery incomplete", slog.Any("err", err))
	}
}
```

igooch: Fixed. Same split as the loop: errors.Is on the sentinel → ERROR "scan skipped", otherwise WARN "startup orphan recovery incomplete" — per-item errors no longer escalate.

PR #837, opened 2026-08-10
validate-image-cache: drive eviction through Store.EvictUnused + GC loop review follow-ups

igooch:
Phase 2 of #463, following #735 and #836: `validate-image-cache` drives eviction through `Store.EvictUnused`, deleting its pre-engine prototype. Also carries #836's post-merge review follow-ups.

## Tool
- **Prototype evictor deleted** (mtime-sorted tree removal, then dropping every manifest record — no refcounts, no two-phase rename, no restore). The tool now asks the engine to reclaim the shortfall below `--min-free-gb`, so corpus runs exercise production semantics. `--evict-idle` maps to `WithMinAge`.
- **`--evict-all`**: one-shot flush, no refs file or auth. Unlike `rm -rf`, running actors' images survive via bundle-spec rooting (`WithActorsDir`). It is not synchronized with a running atelet (the engine's locks are per-process): the pool can't be corrupted, but a layer reused mid-pass can be evicted from under an actor — a not-yet-mounted start fails once and heals; an already-mounted actor can take EIO. Documented, warned at runtime; the intended end state is an atelet-owned flush (RPC or trigger), not a second process in the pool.
- **`--evict-idle` floors at 1m when an actors dir exists**: min-age is the only protection that applies across processes, so on a live node it can't be tuned away. Validation hosts (no actors dir) keep full freedom.
- Gate-aware errors in both eviction paths (`ErrIncompleteEnumeration` = "did nothing, repair the named path" vs per-item errors = stats still print, exit 1); a 30s cooldown after fruitless passes so workers don't serialize full engine passes when nothing is evictable.

## Carried #836 follow-ups
- Negative `--image-cache-gc-period` rejected (was silently disabling the loop); `noteOutcome` extracted with the backoff cadence unit-tested; engine per-pass line demoted to DEBUG (no-target ticks are now silent); single-line sentinel wraps; `filepath.Abs` + table test for the outside-BasePath warning; `runPass`/`Run` covered behind a `gcStore` seam; flag help and README made truthful (`period=0` still runs startup recovery; watermark reads ~5 pts above `df`; the total-size cap still evicts everything evictable under sustained foreign pressure — retention floor named as the extension).

## Testing
- Unit: `-race` clean; new tests for backoff cadence, flag validation, path check, `runPass` skip/panic paths, and `Run` (immediate first pass pinned deterministically; ticking covered separately).
- Kind: full e2e green; gated-pass arc verified live (one ERROR per tick naming the corrupt record; startup scan gates and recovers; planted orphan reclaimed at restart); `--evict-all` on a live node evicted 4 unrooted images while all rooted images survived.
- GKE (6 nodes): zero image-cache log lines fleet-wide at default log level across ticks.
- **Corpus sweep** (7.9 GB volume, always-low-water, `--evict-idle=10s` on a host with no actors dir, parallel pulls): 64/64 images validated with two eviction episodes mid-run (62 img / 103 layers / 5.2 GB, then 46 img / 65 layers / 2.7 GB) — continuous eviction under concurrent real pulls, zero races.

### DIFF
--- a/tools/validate-image-cache/main.go
+++ b/tools/validate-image-cache/main.go
@@ -33,10 +33,14 @@
 // placed actors' mounted images. The engine's locks are per-process, so a
 // run beside a live atelet is unsynchronized with its GC and pulls — the
 // pool cannot be corrupted (record-first pulls, two-phase retirement),
-// but a layer idle past --evict-idle can be retired just as a new pull
-// cache-hits it: that actor start fails once and heals on re-pull. Run
-// as the user that owns the cache and actors dirs — an unreadable actors
-// dir gates every pass.
+// but a layer idle past --evict-idle can be retired just as atelet
+// reuses it: a not-yet-mounted actor start fails once and heals on
+// re-pull; an already-mounted actor can take EIO from a removed
+// lowerdir. min-age is the only cross-process guard, hence the floor on
+// --evict-idle when an actors dir exists. (The end state is an
+// atelet-owned flush — RPC or trigger file — not a second process in
+// the pool.) Run as the user that owns the cache and actors dirs — an
+// unreadable actors dir gates every pass.
 package main
 
 import (
@@ -63,7 +67,7 @@ import (
 )
 
 var (
-	refsFile  = flag.String("refs-file", "", "File with one image ref per line (required)")
+	refsFile  = flag.String("refs-file", "", "File with one image ref per line (required unless --evict-all)")
 	sample    = flag.Int("sample", 0, "Validate a random sample of N refs (0 = all)")
 	seed      = flag.Int64("seed", 1, "Seed for reproducible sampling")
 	cacheDir  = flag.String("cache-dir", "", "Cache root (required); reused across runs")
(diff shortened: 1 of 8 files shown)


tools/validate-image-cache/main.go
```diff
@@ -76,12 +100,52 @@ type result struct {
...
+		}
+	}
 	ctx := context.Background()
 
+	if *evictAll {
+		// Flush mode: no refs, no registry auth. New also reclaims any
+		// crash-debris orphans before the pass.
+		if _, err := os.Stat(ateompath.ActorsDir); err == nil {
+			// An actors dir is a decent proxy for a live node.
+			log.Printf("WARNING: %s exists — this looks like a live node, and this run is not synchronized with other processes using the cache", ateompath.ActorsDir)
```
dberkov: Refs mode doesn't print this live-node warning, but the unsynchronized-pool caveat applies equally when `evictIfLow` fires during a validation run beside a live atelet — only the 1m `--evict-idle` floor guards that case. Consider emitting the same WARNING in refs mode when the actors dir exists, since the eviction path is identical.

tools/validate-image-cache/main.go
```diff
@@ -76,12 +100,52 @@ type result struct {
...
+		fmt.Fprintln(os.Stderr, "need --cache-dir plus exactly one of --refs-file or --evict-all")
 		flag.Usage()
 		os.Exit(2)
 	}
+	// min-age is the only protection that works across processes (the
+	// engine's locks are per-process), so on a live node it must not be
+	// tuned away. Validation hosts (no actors dir) keep full freedom.
+	const liveNodeIdleFloor = time.Minute
+	if *evictIdle < liveNodeIdleFloor {
+		if _, err := os.Stat(ateompath.ActorsDir); err == nil {
```
dberkov: This floor check is silently bypassed when `Stat` fails with something other than ENOENT (e.g. EACCES on the actors dir). It fails safe in practice — an unreadable actors dir gates every pass via `InUse` — but `if err == nil || !errors.Is(err, os.ErrNotExist)` would make the guard airtight and surface the permission problem at startup instead of as gated-pass log lines.

tools/validate-image-cache/main.go
```diff
@@ -76,12 +100,52 @@ type result struct {
 
 func main() {
 	flag.Parse()
-	if *refsFile == "" || *cacheDir == "" {
+	if *cacheDir == "" || (*refsFile == "" && !*evictAll) || (*refsFile != "" && *evictAll) {
```
dberkov: Nit: `--evict-all` silently ignores `--min-free-gb`, `--sample`, `--seed`, `--out`, and `--parallel`. Harmless, but rejecting (or logging) non-default values would prevent "why didn't my min-free apply" confusion.

tools/validate-image-cache/main.go
```diff
@@ -203,57 +267,54 @@ func shortRef(ref string) string {
...
 	}
-	var candidates []aged
-	for _, e := range entries {
-		info, err := e.Info()
-		if err != nil || time.Since(info.ModTime()) < *evictIdle {
-			continue
-		}
-		candidates = append(candidates, aged{filepath.Join(layersDir, e.Name()), info.ModTime()})
+	if stats.EvictedImages > 0 || stats.EvictedLayers > 0 {
+		log.Printf("evicted %d images / %d layers, %.1f GB credited (free now %.0f GB)",
```
dberkov: Cosmetic: if `freeBytes` fails mid-run, its `^uint64(0)` sentinel prints here (and in the evict-all summary) as ~18 billion GB. Not worth fixing unless you touch the file again.

dberkov (approved):

PR #840, opened 2026-08-10
Enable CSI support for micro-vm

### LINKED ISSUE #232: [Feature] External volume mounts

hajiler:
Add support for external volumes for micro vm environments.

This PR adds a new virtiofs specifically for CSI volume mounts. 

Fixes #232 

- [X ] Tests pass

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -466,6 +466,11 @@ func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (resp *
 		return nil, err
 	}
 
+	spec, err := buildAteomWorkloadSpec(req.GetSpec())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid workload spec: %v", err)
+	}
+
 	// Tell ateom to start the workload. gVisor uses RunscPath; the micro-VM
 	// runtime uses the full RuntimeAssetPaths set.
 	if _, err := client.RunWorkload(ctx, &ateompb.RunWorkloadRequest{
@@ -475,7 +480,7 @@ func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (resp *
 		ActorTemplateName:      req.GetActorTemplateName(),
 		RunscPath:              runscPathFor(assetPaths),
 		RuntimeAssetPaths:      assetPaths,
-		Spec:                   buildAteomWorkloadSpec(req.GetSpec()),
+		Spec:                   spec,
 		ActorUid:               actorUID,
 		EgressGateway:          toAteomEgressGateway(req.GetEgressGateway()),
 		CpuMilli:               req.GetCpuMilli(),
@@ -580,6 +585,11 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	// Tell ateom to take the checkpoint and delete containers. ateom reports the
 	// exact files it wrote so we ship precisely that set (gVisor's image files,
 	// cloud-hypervisor's snapshot set, ...) rather than a hardcoded list.
+	spec, err := buildAteomWorkloadSpec(req.GetSpec())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid workload spec: %v", err)
+	}
+
 	tAteom := time.Now()
 	resp, err := client.CheckpointWorkload(ctx, &ateompb.CheckpointWorkloadRequest{
 		Atespace:               actorRef.Atespace,
@@ -588,7 +598,7 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 		ActorTemplateName:      req.GetActorTemplateName(),
 		RunscPath:              runscPathFor(assetPaths),
 		RuntimeAssetPaths:      assetPaths,
-		Spec:                   buildAteomWorkloadSpec(req.GetSpec()),
+		Spec:                   spec,
 		Scope:                  toAteomSnapshotScope(req.GetScope()),
 		ActorUid:               actorUID,
 	})
@@ -601,7 +611,7 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	}
 
 	sandboxRec.SnapshotFiles = resp.GetSnapshotFiles()
-	if len(sandboxRec.SnapshotFiles) == 0 {
+	if len(sandboxRec.SnapshotFiles) == 0 && shouldHaveSnapshots(req) {
 		return nil, ateerrors.NewGRPCError(ctx, codes.DataLoss, ateerrors.ReasonInvalidCheckpointResult, ateerrors.ActorCrashedMetadata(), errors.New("ateom reported no snapshot files for checkpoint"))
 	}
 	sandboxRec.Atespace = req.GetAtespace()
@@ -691,6 +701,20 @@ func (s *AteomHerder) moveLocalCheckpoint(ctx context.Context, req *ateletpb.Che
 	}
 
 	return nil
+}
+
+// shouldHaveSnapshots returns true if the checkpoint request is expected to produce snapshot files.
+func shouldHaveSnapshots(req *ateletpb.CheckpointRequest) bool {
+	if req.GetScope() != ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA {
+		return true
+	}
+
+	for _, vol := range req.GetSpec().GetVolumes() {
+		if vol.GetType() == ateletpb.VolumeType_VOLUME_TYPE_DURABLE_DIR {
+			return true
+		}
+	}
+	return false
 }
 
 func (s *AteomHerder) uploadExternalCheckpoint(ctx context.Context, req *ateletpb.CheckpointRequest, checkpointDir string, rec *sandboxAssetsRecord) error {
@@ -1105,6 +1129,11 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 
 	// Tell ateom to do runsc create + runsc restore for pause container and
 	// all application containers.
+	spec, err := buildAteomWorkloadSpec(req.GetSpec())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid workload spec: %v", err)
+	}
+
 	tAteom := time.Now()
 	_, err = client.RestoreWorkload(ctx, &ateompb.RestoreWorkloadRequest{
 		Atespace:               actorRef.Atespace,
@@ -1113,7 +1142,7 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 		ActorTemplateName:      req.GetActorTemplateName(),
 		RunscPath:              runscPathFor(assetPaths),
 		RuntimeAssetPaths:      assetPaths,
-		Spec:                   buildAteomWorkloadSpec(req.GetSpec()),
+		Spec:                   spec,
 		Scope:                  toAteomSnapshotScope(req.GetScope()),
 		ActorUid:               req.GetActorUid(),
 		EgressGateway:          toAteomEgressGateway(req.GetEgressGateway()),
@@ -1495,32 +1524,50 @@ func (s *AteomHerder) dialAteom(ctx context.Context, targetAteomUid string) (ate
 
 // buildAteomWorkloadSpec projects the atelet-facing workload spec onto
 // the ateom-facing one.
-func buildAteomWorkloadSpec(spec *ateletpb.WorkloadSpec) *ateompb.WorkloadSpec {
-	ddVolumes := make(map[string]bool)
+func buildAteomWorkloadSpec(spec *ateletpb.WorkloadSpec) (*ateompb.WorkloadSpec, error) {
+	volumes := make(map[string]ateletpb.VolumeType)
 	for _, vol := range spec.GetVolumes() {
-		if vol.GetType() == ateletpb.VolumeType_VOLUME_TYPE_DURABLE_DIR {
-			ddVolumes[vol.GetName()] = true
+		name := vol.GetName()
+		if _, duplicate := volumes[name]; duplicate {
+			return nil, fmt.Errorf("duplicate volume name %q in workload spec", name)
 		}
+		volumes[name] = vol.GetType()
 	}
 
 	out := &ateompb.WorkloadSpec{}
 	for _, ctr := range spec.GetContainers() {
 		var ddMounts []*ateompb.DurableDirVolumeMount
+		var csiMounts []*ateompb.VolumeMount
 		for _, vm := range ctr.GetVolumeMounts() {
-			if ddVolumes[vm.GetName()] {
+			volName := vm.GetName()
+			volType, ok := volumes[volName]
+			if !ok {
+				return nil, fmt.Errorf("container %q mounts volume %q which is not defined in workload volumes", ctr.GetName(), volName)
+			}
+
+			switch volType {
+			case ateletpb.VolumeType_VOLUME_TYPE_DURABLE_DIR:
 				ddMounts = append(ddMounts, &ateompb.DurableDirVolumeMount{
-					VolumeName: vm.GetName(),
+					VolumeName: volName,
 					MountPath:  vm.GetMountPath(),
 				})
+			case ateletpb.VolumeType_VOLUME_TYPE_EXTERNAL:
+				csiMounts = append(csiMounts, &ateompb.VolumeMount{
+					VolumeName: volName,
+					MountPath:  vm.GetMountPath(),
+				})
+			default:
+				return nil, fmt.Errorf("container %q mounts volume %q with unsupported type %v", ctr.GetName(), volName, volType)
 			}
 		}
 		out.Containers = append(out.Containers, &ateompb.Container{
 			Name:                   ctr.GetName(),
 			DurableDirVolumeMounts: ddMounts,
+			CsiVolumeMounts:        csiMounts,
 			Readyz:                 toAteomReadyz(ctr.GetReadyz()),
 		})
 	}
-	return out
+	return out, nil
 }
 
 func toAteomEgressGateway(gateway *ateletpb.EgressGateway) *ateompb.EgressGateway {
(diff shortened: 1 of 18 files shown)


demos/counter/counter-microvm-csi-test.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: v1
```
BenTheElder, line 15: missing required boilerplate header here

cmd/atelet/main.go
```diff
@@ -564,7 +564,22 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
...
+		for _, vol := range req.GetSpec().GetVolumes() {
+			switch vol.GetType() {
+			case ateletpb.VolumeType_VOLUME_TYPE_DURABLE_DIR:
+				hasDurable = true
+			case ateletpb.VolumeType_VOLUME_TYPE_EXTERNAL:
+				hasCsi = true
+			}
+		}
+		isDataScope := req.GetScope() == ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
+		if isDataScope && hasCsi && !hasDurable {
```
hajiler: Currently, we do not snapshot csi volumes since they are external and persisted beyond pause/suspend. So we are basically quitely ignoring Data snapshot config when only CSI volumes are present.

Instead should we either:
1) Disallow Data scope snapshot when only CSI volumes are present?
2) Take snapshots of CSI volumes?
3) Add some additional scope specific to CSI volumes?

IMO maybe we just add a note to the the Data snapshot scope that it is particular to durable dir, and external volumes will not be snapshotted.

cmd/atelet/main.go
```diff
@@ -1316,27 +1331,33 @@ func (s *AteomHerder) dialAteom(ctx context.Context, targetAteomUid string) (ate
...
+		volumes[vol.GetName()] = vol.GetType()
 	}
 
 	out := &ateompb.WorkloadSpec{}
 	for _, ctr := range spec.GetContainers() {
 		var ddMounts []*ateompb.DurableDirVolumeMount
+		var csiMounts []*ateompb.VolumeMount
 		for _, vm := range ctr.GetVolumeMounts() {
-			if ddVolumes[vm.GetName()] {
+			switch volumes[vm.GetName()] {
```
hajiler: Do we have any validation logic that ensures all volume mounts are in volumes and are unique? Should we add error to this function in case the spec turns out to be invalid?

cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -193,7 +207,10 @@ func ReconstructSharedDirFromImage(ctx context.Context, bundleRootfs, restoreID,
 //
 // withDurableShare additionally mounts the writable durable-dir share, whose
 // per-volume subdirectories the containers bind-mount at their declared paths.
-func (a *AgentClient) CreateSandboxForActor(ctx context.Context, sandboxID, hostname string, withDurableShare bool) error {
+//
+// withCsiShare additionally mounts the writable CSI volumes share, whose
+// per-volume subdirectories the containers bind-mount at their declared paths.
+func (a *AgentClient) CreateSandboxForActor(ctx context.Context, sandboxID, hostname string, withDurableShare bool, withCsiShare bool) error {
```
hajiler: nit: maybe lets create an opts for this to avoid an increasingly long signature

cmd/ateom-microvm/checkpoint.go
```diff
@@ -81,13 +81,15 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
...
 	durable := hasDurableVolumes(req.GetSpec().GetContainers())
+	csi := hasCsiVolumes(req.GetSpec().GetContainers())
 	scope := req.GetScope()
 	switch scope {
 	case ateompb.SnapshotScope_SNAPSHOT_SCOPE_FULL:
 	case ateompb.SnapshotScope_SNAPSHOT_SCOPE_DATA:
-		if !durable {
+		// TODO: Revisit handling for CSI volumes since snapshots are currently quietly ignored.
+		if !durable && !csi {
 			return nil, status.Error(codes.FailedPrecondition,
```
hajiler, line 99: Similar to the commend above, should we continue to quietly ignore snapshots for CSI volumes?

cmd/ateom-microvm/csi.go
```diff
@@ -0,0 +1,80 @@
...
+	"path/filepath"
+
+	"github.com/agent-substrate/substrate/cmd/ateom-microvm/internal/kata"
+	"github.com/agent-substrate/substrate/internal/ateompath"
+	"github.com/agent-substrate/substrate/internal/proto/ateompb"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// hasCsiVolumes reports whether any container mounts a CSI volume.
+func hasCsiVolumes(containers []*ateompb.Container) bool {
```
hajiler, line 35: Add unit test

cmd/ateom-microvm/csi.go
```diff
@@ -0,0 +1,80 @@
...
+	out := make([]specs.Mount, 0, len(mounts))
+	for _, m := range mounts {
+		out = append(out, specs.Mount{
+			Destination: m.GetMountPath(),
+			Source:      kata.GuestCsiVolumeDir(m.GetVolumeName()),
+			Type:        "bind",
+			Options:     []string{"rbind", "rw"},
+		})
+	}
+	return out
```
hajiler, line 57: Add unit test

cmd/ateom-microvm/csi.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func csiVirtiofsdLogPath(id string) string {
+	return filepath.Join(kata.VMDir(id), "virtiofsd-csi.log")
+}
```
hajiler: Add unit test

cmd/ateom-microvm/run.go
```diff
@@ -66,6 +66,10 @@ type runningActor struct {
 	// durable-dir volumes. nil when the actor declares none. Owned and torn down
 	// exactly like vfsdCmd.
 	durableVfsdCmd *exec.Cmd
+	// csiVfsdCmd is the third virtiofsd, serving the actor's writable
```
hajiler: nit: replace 'the third' with 'a'

pkg/api/v1alpha1/actortemplate_types.go
```diff
@@ -355,7 +355,7 @@ type SnapshotsConfig struct {
 // ActorTemplateSpec defined desired spec of an actor.
 //
 // +kubebuilder:validation:XValidation:rule="!has(self.volumes) || self.volumes.all(v, has(self.containers) && self.containers.exists(c, has(c.volumeMounts) && c.volumeMounts.exists(vm, vm.name == v.name)))",message="All volumes defined in spec.volumes must be mounted by at least one container"
-// +kubebuilder:validation:XValidation:rule="!has(self.sandboxClass) || self.sandboxClass != 'microvm' || !has(self.volumes) || !self.volumes.exists(v, has(v.externalVolumeTemplate))",message="ExternalVolumes are not supported when sandboxClass is 'microvm'"
+	// Bypassed validation for microVM CSI test
```
hajiler: Remove.

docs/dev/microvm-csi-shortcomings.md
hajiler, line 1: Remove this file

cmd/atelet/main.go
```diff
@@ -1457,32 +1487,46 @@ func (s *AteomHerder) dialAteom(ctx context.Context, targetAteomUid string) (ate
...
 	out := &ateompb.WorkloadSpec{}
 	for _, ctr := range spec.GetContainers() {
 		var ddMounts []*ateompb.DurableDirVolumeMount
+		var csiMounts []*ateompb.VolumeMount
 		for _, vm := range ctr.GetVolumeMounts() {
-			if ddVolumes[vm.GetName()] {
+			volName := vm.GetName()
+			volType, ok := volumes[volName]
+			if !ok {
+				return nil, fmt.Errorf("container %q mounts volume %q which is not defined in workload volumes", ctr.GetName(), volName)
```
BenTheElder, line 1545: 🤖 **should-fix** 🟡 – This answers your own question from the last round about validating mounts against volumes, but it only covers one half of it and only at runtime.

The uniqueness half is still open: `volumes` is a map keyed by name, so two entries in `spec.volumes` sharing a name silently collapse to the last one. If they differ in type, the mount is classified by whichever came last in the slice.

For the dangling-reference half, the error is right but late. The ActorTemplate has a CEL rule requiring every volume to be mounted by some container, but not the converse, so a template with a `volumeMounts` entry naming no volume passes admission and then fails at `Run` with `InvalidArgument` — a template that can never start an actor. The mirror-image CEL rule alongside the existing one would reject it at create time.

cmd/atelet/main.go
```diff
@@ -566,7 +576,22 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
...
+			switch vol.GetType() {
+			case ateletpb.VolumeType_VOLUME_TYPE_DURABLE_DIR:
+				hasDurable = true
+			case ateletpb.VolumeType_VOLUME_TYPE_EXTERNAL:
+				hasCsi = true
+			}
+		}
+		isDataScope := req.GetScope() == ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
+		if isDataScope && hasCsi && !hasDurable {
+			// OK: CSI volumes don't produce snapshot files in DATA scope.
```
BenTheElder: 🤖 **question** 🟢 – Minor: the empty `if` body with the real path in `else` reads backwards. Inverting to `if !(isDataScope && hasCsi && !hasDurable) { return ... }` — or a small helper — would say it plainly.

cmd/ateom-microvm/internal/kata/restore.go
```diff
@@ -35,3 +35,9 @@ func VsockSocketPath(id string) string { return filepath.Join(VMDir(id), "clh.so
 func DurableVirtiofsdSocketPath(id string) string {
 	return filepath.Join(VMDir(id), "virtiofsd-durable.sock")
 }
+
+// CsiVirtiofsdSocketPath is the vhost-user-fs socket for the actor's writable
+// CSI volumes share, served by a third virtiofsd (or second if no durableDir is present).
```
BenTheElder: 🤖 **nit** 🟢 – The "third virtiofsd (or second if no durableDir is present)" phrasing is the one you asked to simplify last round. It was fixed in `run.go` but not here.

cmd/ateom-microvm/run.go
```diff
@@ -403,6 +410,23 @@ func (s *AteomService) coldBootActor(ctx context.Context, p actorBootParams) (re
 		}()
 	}
 
+	// CSI volumes (if any) share one writable virtio-fs share, served by
+	// a virtiofsd from the host directory atelet; each volume is a
```
BenTheElder: 🤖 **nit** 🟢 – This sentence lost some words: "served by a virtiofsd from the host directory atelet; each volume is a subdirectory of it." Probably meant something like "from the host directory atelet stages them into".

cmd/atelet/main.go
BenTheElder: I think that's correct for now. @dberkov FYI.

cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -54,6 +54,14 @@ const (
 	// volume's contents live at <guestDurableDir>/<volumeName> and are bind-mounted
 	// from there into the containers that declare the volume.
 	guestDurableDir = "/run/ateom-durable"
+
+	// CsiFsTag is the virtio-fs tag for the actor's WRITABLE CSI volumes share,
+	// served by a virtiofsd.
+	CsiFsTag = "ateCsi"
```
BenTheElder: nit:
```suggestion
	csiFSTag = "ateCSI"
```

Do we really need to export this?

Also, more broadly: Acronyms should be same-case, https://go.dev/talks/2014/names.slide#5:~:text=Acronyms%20should%20be%20all%20capitals%2C%20as%20in%20ServeHTTP%20and%20IDProcessor.

cmd/atelet/main.go
hajiler: Added both another runtime validation for duplicate volume keys (as a fail-safe), while also adding some kubebuilder validation as well, so this case should never actually be hit.

cmd/atelet/main.go
hajiler: Added a helper for better readability, removing the empty if statement.

cmd/ateom-microvm/internal/kata/overlay_linux.go
hajiler: Fixed, and yes it's referenced by the main package, and DurableDir tag is also exported.

cmd/ateom-microvm/checkpoint.go
hajiler: For now yes.

BenTheElder:
[needs rebase]

cmd/ateom-microvm/restore.go
```diff
@@ -506,8 +483,12 @@ func rewriteSnapshotSocketPaths(snapshotDir, id string) error {
 			switch tag, _ := fm["tag"].(string); tag {
 			case kata.FsTag:
 				fm["socket"] = kata.VirtiofsdSocketPath(id)
-			case kata.DurableFsTag:
+			case "ateDurable":
+				// Legacy multi-virtiofs snapshot backward compatibility.
```
BenTheElder: 🤖 **should-fix** 🟡 – These two cases can't deliver the compatibility they claim. This PR collapses to one virtiofsd — `stageMergedRootfs` holds the only `StartVirtiofsd` call, and `buildFsConfigs` emits a single `kataShared` device — so nothing ever creates `virtiofsd-durable.sock` or `virtiofsd-csi.sock` again.

Rewriting a legacy snapshot's device onto one of those paths therefore hands CH a vhost-user-fs device pointing at a socket that will never appear. The restore fails late, at connect time, with a socket error instead of a clear "this snapshot predates the single-share layout".

The sibling subtest already takes the opposite position on the same situation: `ateUpper`, also a retired share, is asserted to fail loudly because "one showing up must fail loudly". `ateDurable` and `ateCSI` are now in exactly that category.

Dropping both cases and letting `default:` reject them is the honest version. `DurableVirtiofsdSocketPath` and `CsiVirtiofsdSocketPath` in `internal/kata/restore.go` then go with them — along with the stale "served by a second virtiofsd" comment on the first one — as does the `ateDurable` assertion in `restore_test.go`, which currently makes this path look verified.

BenTheElder, line 487: 🤖 **should-fix** 🟡 – Drop these two cases; they are dead code, and we do not need the compatibility they are reaching for.

This PR collapses to one virtiofsd — `stageMergedRootfs` holds the only `StartVirtiofsd` call, and `buildFsConfigs` emits a single `kataShared` device — so `virtiofsd-durable.sock` and `virtiofsd-csi.sock` are never created again. Rewriting a legacy snapshot's device onto one of those paths hands CH a vhost-user-fs device pointing at a socket that will never appear, so the "backward compatibility" is a comment rather than a behavior.

The neighbouring subtest already takes the opposite position on the same situation: `ateUpper`, also a retired share, is asserted to fail loudly because "one showing up must fail loudly". `ateDurable` and `ateCSI` are now in that same category, and letting `default:` reject them says so.

That also retires `DurableVirtiofsdSocketPath` and `CsiVirtiofsdSocketPath` in `internal/kata/restore.go` — along with the now-wrong "served by a second virtiofsd" comment on the first — and the `ateDurable` assertion in `restore_test.go`, which currently makes this path look verified.

BenTheElder: We can drop the dead code in a follow-up, this PR has been stuck for a while.

BenTheElder (approved):

PR #841, opened 2026-08-10
fix(actorlogger): add actor UID to JSON logs and handle null input

### LINKED ISSUE #838: actorlog: safely and consistently enrich structured container logs
## Expected Behavior

`internal/actorlog` should safely normalize every non-empty container log line and consistently attach authoritative actor metadata.

* For a valid structured JSON object, `ate.dev/actor_uid` should be injected just like it is for plain-text logs and lifecycle logs
* A top-level JSON value that is not an object, including `null`, should be treated as plain text

## Actual Behavior

* The valid-object branch adds actor atespace, name, template, and container labels, but omits `ate.dev/actor_uid`

* A line containing `null` decodes successfully into a nil `map[string]any`; the subsequent assignment of `time` panics the forwarding goroutine.

## Specifications

- Version: 596cfb8ab02e01bce6178538dfb7160de96acc10
- Platform: N/A

yelshall:
Fixes #838 

Noticed that `WrapContainerLogs` doesn't add UID to the logs, and then also fixed handling null messages.

- [x] Tests pass

### DIFF
--- a/internal/actorlog/logger.go
+++ b/internal/actorlog/logger.go
@@ -123,6 +123,9 @@ func (al *ActorLogger) WrapContainerLogs(r io.Reader, actorRef resources.ActorRe
 			dec.UseNumber()
 
 			unmarshalErr := dec.Decode(&m)
+			if unmarshalErr == nil && m == nil {
+				unmarshalErr = errors.New("JSON value is not an object")
+			}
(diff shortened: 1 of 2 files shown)


maxsmythe:
@git286 For triage/review

git286 (approved):

PR #846, opened 2026-08-11
microvm: write container rootfs to host disk instead of guest memory

Oneimu:
### Problem

Each container's rootfs upper previously lived on a tmpfs inside the guest: every written byte consumed guest RAM 1:1, and writes failed with `ENOSPC` at the tmpfs cap (20% of guest RAM nominally; **~264 MiB effective** on the standard 2 GiB guest once `/run`'s other occupants are counted). Write-heavy actors crash; the rest degrade worker density.

### Change

The rootfs overlay is now **assembled on the host** — the conventional arrangement for VM-isolated containers: lower = the read-only OCI image bundle, upper/work = per-actor host directories, merged by the host kernel and served to the guest over the **single existing virtio-fs share** (`cache=auto`). The guest runs each container directly on the merged directory in one step; it never mounts an overlay, needs no special mount options, and no additional virtiofsd exists. The mount pins `metacopy=off,index=off` so every copy-up is a full data copy — the upper never contains file-handle references to lower inodes, which would go stale when restore rebuilds the lower from the image (review finding).

**Snapshot/restore integration:**

- **FULL checkpoints** archive the upper as `rootfs-upper.tar`, captured while the guest is paused (the share is write-through, so the tar is coherent) and **concurrently** with the cloud-hypervisor memory snapshot — the paused window costs the slowest artifact, not the sum.
- **Restores** re-materialize the upper in the background (overlapped with bundle preparation), re-mount the merged trees at the frozen find-paths locations, then start the share.
- **Restore is self-describing:** the tar's presence routes it. Snapshots from the current memory implementation restore unchanged (bare-image bind at `cache=always`; their upper rides inside the restored guest memory), and their re-checkpoints stay correct.
- **DATA scope unchanged:** rootfs state discarded, as today.

**Enabling change — `tarutil` now round-trips overlay deletion metadata:**  
Whiteout device nodes (`mknodat`, parent-fd contained) and `user.*` / `trusted.overlay.*` xattrs (PAX `SCHILY.xattr` records, restored through the extraction root so a crafted archive cannot write outside it). Without these, deleted files and directories silently reappeared after resume.

### How it was tested

- **Unit** (privileged tests run as root, no skips): `tarutil` round-trips including a literal `0:0` whiteout device and `trusted.overlay.opaque`; extraction-escape rejections; overlay layout and snapshot-config rewrite regression tests.
- **End-to-end** (live cluster, counter actor with a durable volume — so every run also covers durable-dir coexistence):
  - In-guest writes visible in the host upper.
  - In-guest `rm` producing a genuine kernel whiteout (`char 0:0, nlink 2`) and directory replacement producing `trusted.overlay.opaque`, both surviving `FULL` suspend → restore verified *from inside the guest*.
  - Multi-cycle suspend/resume with a byte-identical 64 MiB payload (`md5`).
  - `DATA`-scope commits emit no rootfs tar.
  - `DATA_ON_GOLDEN` stitches the golden's rootfs tar with the actor's durable data.
  - Legacy (tmpfs-era) snapshots restore and re-checkpoint correctly.
- **Benchmark**: Compared head-to-head against the memory baseline on GKE with real GCS snapshot storage ($N=25$ full-suspend cycles per side, same node/bucket) — suspend and restore latency at parity, no regression.

### DIFF
--- a/cmd/ateom-microvm/checkpoint.go
+++ b/cmd/ateom-microvm/checkpoint.go
@@ -33,6 +33,7 @@ import (
 	"github.com/agent-substrate/substrate/internal/imagecache"
 	"github.com/agent-substrate/substrate/internal/proto/ateompb"
 	"github.com/agent-substrate/substrate/internal/resources"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
@@ -47,11 +48,10 @@ import (
 //   - FULL: the whole guest. ateom drives the CH REST api-socket: pause -> snapshot
 //     file://<CheckpointStateDir> (config.json + state.json + sparse memory-ranges)
 //     -> tear the VMM down. Each container's rootfs is overlay(virtio-fs RO lower +
-//     guest-tmpfs upper), so the writable upper lives in guest RAM and is captured by
-//     the memory snapshot — process memory and rootfs writes both persist across
-//     suspend/resume. The RO lower is reconstructed from the OCI image at restore, so
-//     nothing rootfs-related ships. Durable-dir volumes are host-backed rather than in
-//     guest RAM, so they ship alongside as a tar.
+//     disk-backed upper): the upper is host-backed like the durable-dir volumes and
+//     ships alongside as its own tar (see rootfsupper.go); process memory persists
+//     via the memory snapshot. The RO lower is reconstructed from the OCI image at
+//     restore, so it never ships. Durable-dir volumes ship alongside as a tar.
 //   - DATA: the durable-dir volumes only, as that same tar. The guest is discarded, so
 //     the actor cold-starts on restore with its volumes re-materialized.
 //
@@ -120,27 +120,56 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 		return nil, fmt.Errorf("while creating checkpoint dir %q: %w", checkpointDir, err)
 	}
 
-	// Only a Full snapshot captures the guest. A Data snapshot deliberately
-	// captures no VM state — no memory image, and no base-id, since nothing will
-	// reattach to the frozen virtio-fs lower: at restore the actor cold-boots
-	// from the OCI image (or, under an OnGolden data resume policy, is combined
-	// with the golden snapshot's guest state) and gets its durable-dir volumes
-	// back from the tar below.
-	var dSnapshot time.Duration
+	// Capture the snapshot's pieces CONCURRENTLY: the CH snapshot, the
+	// durable-dir tar, and the rootfs upper tar read independent data from a
+	// quiesced guest and write distinct files into checkpointDir, so the paused
+	// window costs the slowest of them rather than their sum (the tars scale
+	// with the actor's data; suspend latency is the metric that matters).
+	//
+	//   - CH snapshot (Full only): the guest memory + VM state. A Data snapshot
+	//     deliberately captures no VM state — no memory image, and no base-id,
+	//     since nothing will reattach to the frozen virtio-fs lower: at restore
+	//     the actor cold-boots from the OCI image (or, under an OnGolden data
+	//     resume policy, is combined with the golden snapshot's guest state).
+	//   - Durable-dir tar (any scope, when declared): host-backed, so pausing
+	//     the write-through share makes the tar coherent.
+	//   - Rootfs upper tar (Full only): host-backed like the durable volumes —
+	//     the memory snapshot does not carry rootfs writes. Under Data the
+	//     workload cold-starts on restore, discarding rootfs state. Gated on
+	//     the host dir a disk-upper boot creates (actorHasDiskUpper) so a
+	//     legacy actor restored from a tmpfs-upper snapshot checkpoints
+	//     correctly (its upper is inside the memory image).
+	var dSnapshot, dDurable, dUpper time.Duration
+	g, gctx := errgroup.WithContext(ctx)
 	if scope == ateompb.SnapshotScope_SNAPSHOT_SCOPE_FULL {
-		var err error
-		if dSnapshot, err = s.snapshotVMState(ctx, client, ra, actorUID, checkpointDir); err != nil {
-			return nil, err
-		}
+		g.Go(func() error {
+			var err error
+			dSnapshot, err = s.snapshotVMState(gctx, client, ra, actorUID, checkpointDir)
+			return err
+		})
 	}
-
-	var dDurable time.Duration
 	if durable {
-		tDurable := time.Now()
-		if err := tarDurableVolumes(ctx, ateompath.DurableDirVolumeMountsDir(actorUID), checkpointDir); err != nil {
-			return nil, err
-		}
-		dDurable = time.Since(tDurable)
+		g.Go(func() error {
+			t := time.Now()
+			if err := tarDurableVolumes(gctx, ateompath.DurableDirVolumeMountsDir(actorUID), checkpointDir); err != nil {
+				return err
+			}
+			dDurable = time.Since(t)
+			return nil
+		})
+	}
+	if scope == ateompb.SnapshotScope_SNAPSHOT_SCOPE_FULL && actorHasDiskUpper(actorUID) {
+		g.Go(func() error {
+			t := time.Now()
+			if err := tarRootfsUpper(gctx, ateompath.RootfsUpperDir(actorUID), checkpointDir); err != nil {
+				return err
+			}
+			dUpper = time.Since(t)
+			return nil
+		})
+	}
+	if err := g.Wait(); err != nil {
+		return nil, err
 	}
 
 	// Report exactly the files we wrote so atelet ships precisely this snapshot: for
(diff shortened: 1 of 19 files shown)


cmd/ateom-microvm/restore.go
```diff
@@ -165,14 +166,39 @@ func (s *AteomService) restoreFullScope(ctx context.Context, p actorBootParams,
...
+	// below, so it hides behind the lower staging, network, and tap setup and is
+	// joined right before the share's virtiofsd starts (the guest must never
+	// observe the directory mid-restore). Legacy tmpfs-upper snapshots have no
+	// tar and skip all of this — their upper rides inside the restored guest
+	// memory.
+	//
+	// An error return between here and the join MUST drain the goroutine (the
+	// deferred receive below): returning with the untar still writing would let
+	// a retried restore's own untar race it inside the same directory.
+	hasUpper := snapshotHasRootfsUpper(restoreDir)
```
BenTheElder, line 204: 🤖 **nit** 🟢 – The legacy branch is the one entry path that leaves the upper directory as it found it. Cold boot calls `resetRootfsUpperDir` and a disk-upper restore's `untarRootfsUpper` starts with `RemoveAll`, but a no-tar restore touches nothing. Teardown's removal is warn-and-continue, and the directory sits on the BasePath hostPath, so it also survives an ateom kill before teardown runs.

When a stale directory does survive into a legacy restore, `actorHasDiskUpper` reports true for an actor whose upper is really in guest memory. The next FULL checkpoint then tars that directory and emits a `rootfs-upper.tar` the guest never used, and the restore after it stages a third virtiofsd for a share the snapshot's `config.json` does not reference. The real upper still comes back from memory, so this is a mislabelled snapshot rather than a broken one — but the mislabel is sticky.

Calling `resetRootfsUpperDir` on this branch too would make "directory absent" mean the same thing on all three entry paths.

internal/ateompath/ateompath.go
```diff
@@ -172,6 +172,19 @@ func DurableDirVolumeMountPoint(actorUID, volumeName string) string {
 	)
 }
 
+// RootfsUpperDir is the host directory backing the actor's rootfs overlay
+// uppers in ateom-microvm: one subdirectory per container, served into the
+// guest over virtio-fs so rootfs writes land on host disk instead of guest
+// RAM. Unlike the durable-dir volumes it is owned entirely by ateom (created
+// at cold boot, archived at checkpoint, removed at teardown); atelet never
+// touches it.
```
BenTheElder: This package is used for paths that need to be shared between ateom and atelet, or at least multiple ateom implementations, is there a reason to expose this?

BenTheElder (commented):
Is it possible to do the read-only image rootfs + overlay upper on the host side behind the virtiofsd mount, and then tar up the upper if a snapshot is requested (and move the virtiofs to auto instead of always), or does that not work? That might be somewhat less complex and less expensive than the additional virtiofsd if it works.

internal/ateompath/ateompath.go
Oneimu: Fixed — it was misplaced symmetry with the durable-dir path. Unlike those (atelet-owned host side), this directory is managed entirely by ateom-microvm, so it's now a private helper inside the binary.

cmd/ateom-microvm/restore.go
Oneimu: Fixed — with one deviation from the literal suggestion: `resetRootfsUpperDir` ends by recreating the directory (RemoveAll + MkdirAll), and since actorHasDiskUpper keys on bare  existence, calling it on the legacy branch would flip the gate back to true and reintroduce the same mislabel from the other direction (an empty tar emitted, a share staged that the  snapshot's config never references). The invariant we actually want is "directory present ⇔ host-merged `rootfs`," so the legacy restore path now does a plain RemoveAll. With that,  "directory absent" means the same thing on all three entry paths: cold boot resets it, a `merged-rootfs` restore rebuilds it from the tar, and a legacy restore removes anything a crash orphaned.

cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -172,11 +184,64 @@ func waitForSocket(ctx context.Context, path string, timeout time.Duration) erro
...
+	// Drop any stale mount first (lazy if busy), then ensure clean mountpoints.
+	if err := reaper.Run(exec.Command("umount", dst)); err != nil {
+		_ = reaper.Run(exec.Command("umount", "-l", dst))
+	}
+	for _, d := range []string{dst, upper, work} {
+		if err := os.MkdirAll(d, 0o755); err != nil {
+			return fmt.Errorf("creating %q: %w", d, err)
+		}
+	}
+	opts := "lowerdir=" + bundleRootfs + ",upperdir=" + upper + ",workdir=" + work
```
BenTheElder: 🤖 **should-fix** 🟡 – This mount takes the kernel's defaults for `metacopy` and `index`, and the snapshot preserves every `trusted.overlay.*` xattr (`readOverlayXattrs` matches the whole prefix). Those two combine badly.

With `metacopy=on` — a module parameter, `/sys/module/overlay/parameters/metacopy`, that some distros ship enabled — a chmod or chown of a lower file produces a metadata-only copy-up: an upper entry with no data, carrying `trusted.overlay.metacopy` and a `trusted.overlay.origin` file handle pointing at the lower inode. The tar round-trips both faithfully. But restore re-unpacks the lower from the OCI bundle, so those inodes are new and the handle is stale, and overlayfs can no longer find the file's data. That is a silently unreadable file after resume, and it doesn't need a second node to happen — the lower is rebuilt on every restore.

Pinning `metacopy=off,index=off` in `opts` makes every copy-up a full data copy, which keeps the upper self-contained and portable. That's the property the doc comment above already claims ("the merged path is identical on every node"), so it's worth making the mount enforce it rather than inheriting whatever the host is configured for.

`redirect_dir` is fine to leave alone: `trusted.overlay.redirect` holds a path, not a handle, and travels correctly.

cmd/ateom-microvm/checkpoint.go
```diff
@@ -120,27 +120,56 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
...
+			t := time.Now()
+			if err := tarRootfsUpper(gctx, rootfsUpperDir(actorUID), checkpointDir); err != nil {
+				return err
+			}
+			dUpper = time.Since(t)
+			return nil
+		})
+	}
+	if err := g.Wait(); err != nil {
+		return nil, err
```
BenTheElder, line 172: 🤖 **nit** 🟢 – Anchoring here because the lines themselves fall outside the diff: the comment just below, "for Full, the CH snapshot ... plus any durable-dir tar; for Data, that tar alone" (lines 175-177), is now short one file. A Full snapshot also ships `rootfs-upper.tar`.

`listFiles` picks it up regardless, so nothing is broken — but that comment is the description of what a snapshot contains, and it's exactly what this PR changes.

BenTheElder:
I think the PR description is substantially out of date now? It seems to still describe the initial version versus the refactor to host-side overlay. Please update the PR body.

BenTheElder (commented):
exciting that we can drop the carrier container workaround, and fix memory-bound writes to boot. that should actually have measurable impact on boot time as well, the createcarrier call was wasteful.

at least one outstanding issue.

cmd/ateom-microvm/checkpoint.go
Oneimu: ack.

cmd/ateom-microvm/internal/kata/overlay_linux.go
Oneimu: Fixed — pinned metacopy=off,index=off in the mount options with a comment capturing exactly this failure mode (stale trusted.overlay.origin handles after the lower is rebuilt on restore).

cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -172,11 +184,75 @@ func waitForSocket(ctx context.Context, path string, timeout time.Duration) erro
...
+	// LOWER inodes in the upper (a metacopy'd file is a data-less upper entry
+	// whose trusted.overlay.origin handle is where the data lives), and the
+	// snapshot tar preserves trusted.overlay.* verbatim — but restore rebuilds
+	// the lower from the OCI bundle with fresh inodes, so a preserved handle
+	// goes stale and the file turns silently unreadable after resume. With both
+	// off, every copy-up is a full data copy and the upper is self-contained —
+	// the portability the find-paths comment above promises. (redirect_dir is
+	// path-based and travels fine, so it stays at the kernel default.)
+	opts := "lowerdir=" + bundleRootfs + ",upperdir=" + upper + ",workdir=" + work +
+		",metacopy=off,index=off"
```
BenTheElder, line 223: 🤖 **nit** 🟢 – Worth a regression test, since this string is now load-bearing and its failure mode is silent — a file that reads fine in every test that mounts the overlay, and turns unreadable only after a resume rebuilds the lower.

`virtiofsdArgs` next door is the pattern: a pure function returning the args, with a test asserting both what must be present and what must never appear (`--xattr`). Lifting the option string out of `StageMergedRootfs` the same way would make it assertable, and `UpperWorkDirs` already sets the precedent of pinning the load-bearing bits with a test that explains why.

BenTheElder: Human: Let's do this as a follow-up, this overall change is load bearing and it's in pretty good shape.

PR #850, opened 2026-08-11
ateomnet: create the actor veth peer inside the actor netns

BenTheElder:
`SetupActorNetwork` built the veth pair in the worker pod netns with the peer under a temporary name, moved the peer into the actor netns with LinkSetNsFd, and renamed it to eth0 on the far side. The temporary name existed only because the pod netns has an eth0 of its own for the peer to collide with.

Those two operations were 13.8ms of the 15.0ms of veth work: each drives a netdev unregister/register with an RCU grace period taken under the global RTNL lock. Neither parallelizes, so the fix is to not do them. Creating the pair with PeerName plus PeerNamespace has the peer born in the actor netns already called eth0, which costs neither: the name is resolved in the peer's namespace, so it never sees the pod's eth0.

Measured on the real function, 200 activations on a 6.8 kernel:

  before:  `SetupActorNetwork  mean=18.07ms  p50=17.83ms  p99=26.81ms`
  after:   `SetupActorNetwork  mean=2.96ms   p50=2.69ms   p99=5.72ms`

That is ~15ms off every actor resume, against a 100ms p95 activation target, since RestoreWorkload calls this before runsc restore. CleanupActorNetwork is unchanged at ~16ms; that cost is the veth delete and it lands on suspend.

The resulting namespace is identical, which is what the new tests assert: they pin the end state gVisor and the micro-VM guest read -- which links exist, in which namespace, with which addresses and routes -- rather than the sequence of netlink calls that produced it. They also cover repeated activation on a reused worker, cleanup idempotency, and the two micro-VM options (the fixed host veth MAC that keeps a restored guest's frozen ARP entry valid, and the interior link sweep). They run in throwaway anonymous namespaces, so they leave nothing behind on the machine, and are root-gated via internal/roottest like the other privileged tests.

`ActorVethTempName` goes away with its last user.

- [x] Tests pass

### DIFF
--- a/internal/ateomnet/net.go
+++ b/internal/ateomnet/net.go
@@ -37,7 +37,6 @@ import (
 const (
 	HostVethName      = "ateom0"
 	ActorVethName     = "eth0"
-	ActorVethTempName = "ateom1"
 	HostVethCIDR      = "169.254.17.1/30"
 	ActorVethCIDR     = "169.254.17.2/30"
 	ActorVethGateway  = "169.254.17.1"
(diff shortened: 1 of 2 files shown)


internal/ateomnet/net_linux_test.go
```diff
@@ -0,0 +1,311 @@
...
+// linkByName returns the link, or nil when it does not exist.
+func linkByName(t *testing.T, name string) netlink.Link {
+	t.Helper()
+	link, err := netlink.LinkByName(name)
+	if err == nil {
+		return link
+	}
+	if _, ok := errors.AsType[netlink.LinkNotFoundError](err); ok {
+		return nil
+	}
```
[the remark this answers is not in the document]
BenTheElder: https://pkg.go.dev/errors#AsType 🙄

internal/ateomnet/net.go
```diff
@@ -538,11 +527,30 @@ func SetupActorNetwork(ctx context.Context, cfg NetworkConfig) (retErr error) {
 		}
 	}
 
+	// Create the pair with its peer already in the interior netns, already called
```
bowei: Probably should summarize this and make it easier to understand...

bowei:
Do we have any issues with the fact that the test may require privileged to run?

BenTheElder:
> Do we have any issues with the fact that the test may require privileged to run?

I e2e tested it with both uVM and gvisor ateoms, gvisor ateom is not full `privileged`, but it does have many capabilities. uVM ateom still runs with privileged.

I don't think so?

internal/ateomnet/net.go
BenTheElder: Summarize: Can do.

BenTheElder: should be a bit more straightforward now.

PR #858, opened 2026-08-11
Make UpdateWorker follow substrate API guidelines

laoj2:
UpdateWorker was the last Update method in the storage layer still taking the whole Worker as both identity and payload, the version as a positional argument, and returning only an error.

It now takes (namespace, pool, pod) plus a mutate
closure, matching GetWorker and its two Update (matching the other UpdateXYZ) and returns the worker that was stored

https://github.com/agent-substrate/substrate/issues/732
https://github.com/agent-substrate/substrate/issues/763

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/store/ateredis/ateredis.go
+++ b/cmd/ateapi/internal/store/ateredis/ateredis.go
@@ -65,6 +65,10 @@ import (
 // globalAtespace in Substrate is represented by "".
 const globalAtespace = ""
 
+// updateMaxAttempts bounds how many times an updater re-runs its
+// read-modify-write after a concurrent writer invalidates the transaction.
+const updateMaxAttempts = 5
+
 type workerPubSubMsg struct {
 	Type   int    `json:"t"`
 	Worker string `json:"w"` // protojson-encoded Worker
@@ -883,14 +887,9 @@ func validateUpdateActorSnapshotTagMutation(storedTag, mutatedTag *ateapipb.Acto
 	return nil
 }
 
-// updateActorSnapshotTagMaxAttempts bounds how many times UpdateActorSnapshotTag
-// re-runs its read-modify-write after a concurrent writer invalidates the
-// transaction.
-const updateActorSnapshotTagMaxAttempts = 5
-
 func (s *Persistence) UpdateActorSnapshotTag(ctx context.Context, atespace, name string, mutate func(*ateapipb.ActorSnapshotTag) error) (*ateapipb.ActorSnapshotTag, error) {
 	tagKey := actorSnapshotTagDBKey(atespace, name)
-	for range updateActorSnapshotTagMaxAttempts {
+	for range updateMaxAttempts {
 		var dbTag *ateapipb.ActorSnapshotTag
 		var abortErr error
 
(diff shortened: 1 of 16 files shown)


cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -637,67 +637,96 @@ func (s *Persistence) GetWorker(ctx context.Context, namespace, pool, pod string
...
+const updateWorkerMaxAttempts = 5
 
-	err := s.rdb.Watch(ctx, func(tx *redis.Tx) error {
-		currentVal, err := tx.Get(ctx, dbKey).Bytes()
-		if err != nil {
-			if errors.Is(err, redis.Nil) {
-				return store.ErrNotFound
+func (s *Persistence) UpdateWorker(ctx context.Context, namespace, pool, pod string, mutate func(*ateapipb.Worker) error) (*ateapipb.Worker, error) {
+	dbKey := workerDBKey(namespace, pool, pod)
+	for range updateWorkerMaxAttempts {
```
sfunkenhauser: Do we want to retry updates? What are the failure cases here, is it just if there is another conflicting update? If that's the case will having a retry change the outcome? Ultimately one will update last and clobber the other. Without a retry, first update wins. With a retry, second update wins. Do we care? If we don't I'd recommend we remove the retry to make this logic a bit cleaner.

sfunkenhauser (commented):
Tests are failing due to #860, rebasing should fix them.

cmd/ateapi/internal/store/ateredis/ateredis.go
laoj2: > What are the failure cases here, is it just if there is another conflicting update? 

That's right, we get TxFailedErr error when the optimistic lock is lost: https://redis.io/docs/latest/develop/clients/go/transpipe/#watch-keys-for-changes -> something dirtied the watched key between watch and exec. 

The redis docs also suggest to retry in that case. This is especially useful for updates to disjoint fields (with retry, both updates win)

Dropping the retry in that case would mean that the second caller changes would be discarded due to an unrelated update.

sfunkenhauser: That's a fair point about disjoint fields.

Another issue around retires is WithPrecondition. If the version has changed (due to a conflict) we return with ErrVersionConflict and don't actually retry. In the case of using a precondition, the caller needs to handle the retry as well. Can we loosen this restriction?

We likely need to have retry logic at the call sites regardless (especially when we create a worker API and these call sites move out of process). A broader question is whether we want retry logic in both places (call site and data layer, probably not)? But maybe we can defer that until we move this logic out of the api-server.

sfunkenhauser: Given that we will be moving to postgres soon, let's not worry about the retry semantics now. We can revisit this discussion if it's still relevant with postgres.

sfunkenhauser (dismissed):

cmd/ateapi/internal/store/ateredis/ateredis.go
laoj2: With postgres, we don't need to retry anymore at the storage layer: `SELECT ... FOR UPDATE` locks that row, differently from Redis' opportunistic writes. 

Also, we're discussing making version a required precondition in https://github.com/agent-substrate/substrate/issues/954

So, we'll revisit this logic

laoj2:
@sfunkenhauser I rebased it, and added the postgres changes. Can you ptal?

PR #862, opened 2026-08-11
Add library to apply a field mask during updates.

juli4n:
Instead of requiring custom code for each field, let the callers (i.e RPC handlers) declare which fields are mutable and implement a generic `Apply` function that copies fields over, using proto reflection.


laoj2 (dismissed):

zoez7 (approved):

internal/fieldmask/fieldmask.go
```diff
@@ -0,0 +1,102 @@
...
+	dstMsg, srcMsg := dst.ProtoReflect(), proto.Clone(src).ProtoReflect()
+	for _, path := range mask.GetPaths() {
+		copyMaskedField(dstMsg, srcMsg, strings.Split(path, "."))
+	}
+}
+
+func copyMaskedField(dst, src protoreflect.Message, segments []string) {
+	fd := dst.Descriptor().Fields().ByName(protoreflect.Name(segments[0]))
+	if fd == nil {
+		return
```
jpbetz, line 82: If the field path somehow gets past validation and we late-detect that it doesn't match a field here, why silently do nothing?  I would expect an internal error or panic here.

internal/fieldmask/fieldmask.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	if len(segments) == 1 {
+		// TODO: when fd is message-kind, this copies it whole, including any
+		// output-only or immutable subfields it may have. Skip those once
+		// fields carry that annotation.
```
jpbetz, line 87: Consider a tracking issue instead of a TODO here.

There are two problems (not just the annotations part):

1. The issue mentioned in the TODO (output-only subfields getting clobbered). I don't know AIP so read up briefly. I think "AIP-203: ignore output-only in requests" is what the issue should reference for this part.
1. The client version skew case where a client masking worker_selector from a stale schema silently wipes subfields it doesn't know about. I'm uncomfortable with this aspect since it means that exposing a container path is a one-way door (allowing API authors to paint themselves into a corner) so the issue should include guidance on when authors expose container paths at all. Strong warnings to API authors about the risks here and when it's OK vs. when it's a bad idea, feels warranted.. (I'm glad this didn't get implemented yet TBH)

PR #869, opened 2026-08-11
Add Image Pull Secrets

### LINKED ISSUE #868: ImagePullSecrets in actor templates
Right now, you can only auth to google artifact registry. We should add support for pulling from arbitrary private registry with `ImagePullSecrets` on the ActorTemplate. I think the main question is who should resolve the secret? Atelet or ate-api. I'm leaning towards `atelet` much like how kubelet resolve's secret. This however does put a kube apiserver call in the golden image creation path, which is probably fine. Also, we can add a secret cache since I would guess that there aren't that many image pull secrets in a cluster.

The one thing to watch out for is that the atelet kubernetes service account is going to need configuration to access  the referenced secrets.

Stevenjin8:
Fixes #868

* Add an `ImagePullSecret` field to ActorTemplate to pull from private registries
* Atelet, not ate-apiserver resolves the secret on golden image creation.

- [x] Tests pass


BenTheElder (changes requested):
Making every atelet a k8s client is something of a regression for scaling substrate.

Would it be possible for your use case to instead supply credentials to workerpools in advance, via the pod mounting a secret (which allows us to reuse kubelet for this)?

Anyhow, I think this needs more discussion, this is a big change. Commenting further ont he issue.

BenTheElder:
we recently discussed secrets in this week's community call @bowei has a doc shared with the mailinglist

BenTheElder:
https://docs.google.com/document/d/1-zQKvu-oYJlva_kwL3xbAOaFoJJaqO0qEqmqRlt1r0c/edit?resourcekey=0-TDrWohfRuYolcXYO-QPVTQ&tab=t.0

BenTheElder:
FYI: As part of egress there's a pattern emerging to have an indirect secret provider concept. I think image pull secrets would fit as a second form of that.

PR #870, opened 2026-08-11
post-draft-review skill for AI *assisted* PR reviews

BenTheElder:
This is a skill I've been iterating on, first with a few coworkers' PRs only, and then progressively more broadly. It gives pretty decent results with frontier models, but I think we need to continue refining it.

The basic idea is that this creates or updates draft PR comments to add AI findings, categorized by severity, and *clearly denoted* as coming from an agent. It is instructed to bias towards brevity as these models tend to write massive comments. That still needs more tweaking.

Reviewers are expected to leverage this skill to suggest draft comments which they then review and delete as needed, and finally submit an overall review with their own top-level comments.

It does NOT do things like "summarize what the PR does" ... PR authors already do this, and adding that information to the PR thread is not helpful. Anyone interested can ask their agent to summarize the PR for them locally.

What it does do, is report possible bugs, for a reviewer to vet and then post as they see fit.

NOTE: **this should be squash merged**

- [x] Tests pass

### DIFF
--- a/.agents/skills/post-draft-review/README.md
+++ b/.agents/skills/post-draft-review/README.md
@@ -15,6 +15,11 @@ A pending review is visible **only to you**. No one else can see it, the PR auth
 notified, and nothing is published until you submit it. That is the whole point — it
 gives you a checkpoint between an agent's findings and the author's inbox.
 
+**If you had already started your own draft on that PR, the agent added its findings to
+it.** GitHub allows you only one pending review per PR, and the agent posts as you, so
+there is no separate agent review to look at — your comments and its findings are now in
+the same draft, and yours are untouched. The 🤖 prefix is what tells them apart.
+
 ## What you have to do
 
 The pending review is yours, not the agent's. **Every comment publishes under your name.**
@@ -52,6 +57,9 @@ avoids the failure mode entirely.
 ```bash
 gh api repos/<owner>/<repo>/pulls/<num>/reviews/<review_id> --method DELETE
 ```
+
+This deletes **every** comment in the draft, including any you wrote yourself. To drop
+only the agent's findings and keep your own, delete them individually in the UI instead.
 
 ## For agents
 
--- a/.agents/skills/post-draft-review/README.md
+++ b/.agents/skills/post-draft-review/README.md
@@ -1,0 +1,58 @@
+# post-draft-review
+
+How agents post pull request review findings in this repo — and what you have to do
+before any of those findings become public.
+
+If an agent just told you it "posted a review" on your PR, read this first. Nothing has
+been published yet, and finishing the review is your job.
+
+## What the agent did
+
+It reviewed the PR and created a **draft (pending)** GitHub review: inline comments
+anchored to specific lines, with an empty summary body.
+
+A pending review is visible **only to you**. No one else can see it, the PR author is not
+notified, and nothing is published until you submit it. That is the whole point — it
+gives you a checkpoint between an agent's findings and the author's inbox.
+
+## What you have to do
+
+The pending review is yours, not the agent's. **Every comment publishes under your name.**
+Each finding leads with 🤖 to disclose that an agent found it, but that marker does not
+tell readers you skipped checking it — they will reasonably assume you did.
+
+1. Open the PR's **Files changed** tab. Pending comments carry a **Pending** badge, and a
+   **Finish your review** button sits at the top.
+2. **Read each finding against the code it points at.** Agents produce confident,
+   well-worded findings that are wrong often enough that this step is the entire reason
+   the review is a draft.
+3. **Edit or delete** anything wrong, redundant, or not worth the author's time. Deleting
+   a pending comment costs nothing — it was never published.
+4. **Write your own summary** in the "Finish your review" box.
+5. Submit.
+
+## Two things that catch people out
+
+- **Submitting publishes every remaining pending comment at once.** There is no partial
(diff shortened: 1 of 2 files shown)


juli4n (dismissed):

.agents/skills/post-draft-review/SKILL.md
```diff
@@ -0,0 +1,218 @@
...
+Append instead, one mutation per finding, using the review's `node_id` (the `PRR_…`
+value, not the numeric id):
+
+```bash
+gh api graphql -f query='
+mutation($review: ID!, $path: String!, $line: Int!, $body: String!) {
+  addPullRequestReviewThread(input: {
+    pullRequestReviewId: $review,
+    path: $path, line: $line, side: RIGHT, body: $body
+  }) { thread { id } }
```
aditya-shantanu, line 143: 🤖 **should-fix** 🟡 – Shell expansion happens before `gh api` sees the value. If a finding body contains `$varname`, backtick sequences, or backslashes, the shell mangles the content. The batch REST path avoids this with `jq --rawfile`; the GraphQL path should do the same. One approach: write a variables JSON file with jq and pipe it in:

```bash
jq -n --rawfile body c1.md '{review: "PRR_kwDO...", path: "...", line: 214, body: $body}' \
  | gh api graphql -f query='mutation($review: ID!, $path: String!, $line: Int!, $body: String!) {
      addPullRequestReviewThread(input: {pullRequestReviewId: $review, path: $path, line: $line, side: RIGHT, body: $body}) { thread { id } }
    }' --input -
```

.agents/skills/post-draft-review/SKILL.md
```diff
@@ -0,0 +1,218 @@
...
+
+```bash
+jq -n --rawfile c1 c1.md --rawfile c2 c2.md '{
+  body: "",
+  comments: [
+    {path:"cmd/ateapi/internal/store/ateredis/ateredis.go", line:849, side:"RIGHT", body:$c1},
+    {path:"cmd/ateapi/main.go", line:57, side:"RIGHT", body:$c2}
+  ]
+}' > /tmp/review.json
+
```
aditya-shantanu, line 98: 🤖 **nit** 🟢 – If any single comment's `line` falls outside the diff hunk, the whole POST fails and no comments are created — GitHub rejects the batch, not just the bad entry. Worth mentioning a fallback: when creation fails, post comments one at a time (loop over `comments[]` building individual single-entry payloads) to isolate which anchor is invalid.

.agents/skills/post-draft-review/README.md
```diff
@@ -0,0 +1,66 @@
...
+- **The "checked and found clean" notes in the chat hand-off are agent claims too.** They
+  read as reassurance and are the easiest thing to paste into a summary unverified. An
+  agent asserting it verified something is not the same as it having verified it.
+
+## Why the agent doesn't write the summary
+
+The overall verdict and framing are yours. Beyond that, GitHub's "Finish your review" box
+does not prefill a body set through the API, so an agent-written summary would be silently
+dropped when you submit — and it cannot be restored afterwards. Leaving the body empty
+avoids the failure mode entirely.
```
aditya-shantanu, line 53: 🤖 **nit** 🟢 – `<review_id>` here is opaque to a human who wasn't watching the agent's hand-off. Add a one-liner so they can find it without digging:

```bash
gh api repos/<owner>/<repo>/pulls/<num>/reviews \
  --jq '.[] | select(.state == "PENDING") | .id'
```

.agents/skills/post-draft-review/SKILL.md
```diff
@@ -0,0 +1,218 @@
...
+keeps it in draft; including `event` submits immediately.
+
+Inline-comment notes:
+
+- `line` is the line number **in the file at the PR head**. `side: "RIGHT"` is the new
+  (post-change) version, `"LEFT"` the base side.
+- Multi-line: use `start_line` and `start_side` together with `line`/`side`.
+- The line **must fall within the PR's diff hunk**, or the API rejects the whole review.
+  A file that isn't in the diff at all (for example a caller that wasn't touched) can't
+  take an inline comment. Anchor the point on a *changed* line nearby, such as the struct
```
aditya-shantanu, line 115: 🤖 **nit** 🟢 – The skill correctly says to pin line numbers from a checkout of the PR head, but the review payload omits `commit_id`. If the author pushes a fixup commit between when you fetch the diff and when you POST the review, GitHub will attach the anchors to the new head — where the line numbers are now wrong. Capture the head SHA when you checkout and include it:

```bash
commit_id=$(gh pr view <num> --repo <owner>/<repo> --json headRefOid --jq '.headRefOid')
# then add "commit_id": "$commit_id" to the review JSON
```

.agents/skills/post-draft-review/SKILL.md
BenTheElder: Command substitution results aren't re-expanded by the shell.

.agents/skills/post-draft-review/SKILL.md
BenTheElder: updated

.agents/skills/post-draft-review/SKILL.md
BenTheElder: updated

.agents/skills/post-draft-review/README.md
BenTheElder: added

aditya-shantanu (approved):

juli4n (approved):

PR #878, opened 2026-08-12
Fix benchmarking auth bitrot; add microvm

maxsmythe:
Benchmarking needed updates to fix auth bitrot and to enable readint endpointslices. Also streamlined the ability to manually benchmark microvm and removed Python Locust workers by default to simplify self-serve benchmarking.

### DIFF
--- a/benchmarking/locust/manifests/locust.yaml
+++ b/benchmarking/locust/manifests/locust.yaml
@@ -55,7 +55,6 @@ spec:
         - "/app/tests/glutton.py"
         - "--master"
         - "--master-bind-host=0.0.0.0"
-        - "--class-picker"
         env:
         - name: OTEL_EXPORTER_OTLP_ENDPOINT
           value: http://opentelemetry-collector.gke-managed-otel.svc.cluster.local:4317
--- a/benchmarking/locust/manifests/locust.yaml
+++ b/benchmarking/locust/manifests/locust.yaml
@@ -52,7 +52,7 @@ spec:
         - "-m"
         - "locust"
         - "-f"
-        - "/app/tests"
+        - "/app/tests/glutton.py"
         - "--master"
         - "--master-bind-host=0.0.0.0"
         - "--class-picker"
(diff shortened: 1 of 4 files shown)


benchmarking/locust/manifests/locust.yaml
```diff
@@ -193,3 +209,17 @@ spec:
     protocol: TCP
     port: 8001
     targetPort: 8001
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: benchmarking-permissions
```
dberkov, line 215: This RoleBinding has the same name (`benchmarking-permissions`) and namespace (`ate-system`) as the one added in `benchmarking/automation/manifests/runner-job.yaml.tmpl`, but a different subject (`default` SA here vs `benchmark-runner` there). Since a RoleBinding's `subjects` list is replaced on `kubectl apply`, whichever flow deploys second silently strips the other SA's EndpointSlice access — and `deploy_locust.sh --delete` (`kubectl delete -f` on this manifest) deletes the binding outright, breaking a live automation runner. The failure mode is the boomer client's `k8s:///` resolver quietly losing EndpointSlice reads, presenting as confusing ateapi dial failures.

Suggest distinct names (e.g. `benchmarking-locust-endpointslices` / `benchmarking-runner-endpointslices`), or one shared manifest listing both subjects.

benchmarking/locust/manifests/locust.yaml
```diff
@@ -52,7 +52,7 @@ spec:
         - "-m"
         - "locust"
         - "-f"
-        - "/app/tests"
+        - "/app/tests/glutton.py"
         - "--master"
         - "--master-bind-host=0.0.0.0"
         - "--class-picker"
```
dberkov: nit: with the master now loading only `glutton.py` and boomer as the sole worker, there's a single selectable user class — `--class-picker` is vestigial and could be dropped for clarity. (The other test files — `counter_demo.py`, `kernelmem.py`, `sleep.py`, `usermem.py` — become unreachable, which I take to be intentional per the PR description.)

maxsmythe: Yep, intentional for sure. `--class-picker` is probably forever-vistigial (we will likely need a different way of managing multi-class load tests in the future). Removing.

benchmarking/locust/manifests/locust.yaml
maxsmythe: good catch

PR #879, opened 2026-08-12
atecontroller: reclaim golden actor on ActorTemplate delete

AlexBulankou:
The ActorTemplate controller declares the `actortemplates/finalizers` RBAC verb but installs no finalizer, so deleting an ActorTemplate lets the CR be GC'd immediately while the backing golden actor + snapshot blobs live on in the store — unreclaimable once `status.GoldenActorID` is gone with the CR.

This adds an `ate.dev/golden-actor-cleanup` finalizer:
- **Installed on first reconcile** (before the golden actor is created), so a delete that races template setup still triggers reclamation.
- On delete, `cleanupGoldenActor` runs while the finalizer is held, then releases it:
  - **Idempotent** — `NotFound` (already reclaimed) or empty `GoldenActorID` (actor never created) → success.
  - **SUSPENDED-precondition aware** — `DeleteActor` requires `SUSPENDED`/`CRASHED`; a template deleted mid-lifecycle may have a `RUNNING` golden actor (`FailedPrecondition`), so we `SuspendActor` + requeue (5s) rather than deadlock the CR's deletion.

Tests: mock extended with `DeleteActor`/`SuspendActor`; new cases for finalizer-install, delete-when-suspended, delete-when-running (suspend+requeue), delete-when-NotFound, delete-when-no-GoldenActorID. `go build ./...`, `go vet`, and the controller test suite pass locally.

### DIFF
--- a/cmd/atecontroller/internal/controllers/actortemplate_controller.go
+++ b/cmd/atecontroller/internal/controllers/actortemplate_controller.go
@@ -30,10 +30,23 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
 	GoldenSnapshotCreationReason = "GoldenSnapshotCreation"
+
+	// goldenActorFinalizer gates ActorTemplate deletion on reclaiming the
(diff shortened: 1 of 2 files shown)


juli4n:
The `ActorTemplate` resource is being moved to substrate API so I don't see much value in fixing this.

zoez7:
+1 I don't think we need to worry about this right now, because we decided to have ActorTemplate as a substrate resources so will remove the current CRD and controllers for it.

PR #883, opened 2026-08-12
benchmarking: add SLO-knee throughput post-processor

AlexBulankou:
Pure-stdlib post-processor over `runner.py`'s `stats.jsonl`. For each gRPC op in a build tag's concurrency sweep it computes the **SLO-knee throughput**: the max sustained RPS among sweep points whose chosen latency percentile stays under a ceiling AND whose failure ratio is within tolerance. Past the knee, offered load keeps climbing but latency has blown past the SLO, so that extra throughput isn't usable capacity.

Gives the sweep a single honest "throughput @ pXX<SLO" number per op (e.g. `--ceiling-ms 1000` / `5000`) instead of eyeballing `stats_history.csv`.

Offline-only: `benchmarking/analysis/` ships tests + 3 synthetic `stats.jsonl` fixtures — `python3 -m pytest analysis/ -q` runs 18/18 with no cluster.

Note: the fixtures are the repo's first `.jsonl` files, so this also adds `.jsonl` to the data-extension exemption in `hack/util/verify-boilerplate.py` (alongside `.json`) — `verify-all.sh` runs clean with it.


AlexBulankou:
Design question from exercising this against live sweep output: `slo_knee.py` requires an external name→users map (`--users-by-name` / `--tests-yaml`) because stats.jsonl doesn't carry offered concurrency. An alternative: an in-band per-entry `users` field stamped by `runner.py` at write time, with `record_to_point` preferring it and falling back to the external map — drops the analysis-time tests.yaml dependency and makes each stats line self-describing. If you'd prefer in-band, I can send that as a small follow-up (runner.py stamp + the fallback here); if the external-map design is deliberate to keep the stats.jsonl schema stable, this stands as-is.

PR #884, opened 2026-08-12
benchmarking: add cold-start (TTFE) Locust load shape

AlexBulankou:
The existing `ate_api.py` load shape creates its actor in `on_start` and Resume/Suspend-cycles it per task — that measures **warm** activation. This adds `locust/tests/coldstart.py`, a load shape for true **time-to-first-execution**: a fresh actor per iteration, measuring `boot=False` (restore from golden snapshot) and `boot=True` (full boot) as two separate named metrics, so cold-start p50/p95 land in the same `stats.jsonl` pipeline as everything else.

Follows the channel/tracing/atespace helper idioms of the existing `ate_api.py`.

### DIFF
--- a/benchmarking/locust/tests/coldstart.py
+++ b/benchmarking/locust/tests/coldstart.py
@@ -1,0 +1,156 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


AlexBulankou:
Three candidate extensions surfaced while exercising this class against a live ladder config — flagging pre-review so scope can be weighed here rather than post-merge:

1. A composite create→resume→first-read TTFE task alongside the resume-only axis (keeps the `boot=True/False` split intact — the resume-vs-full-boot comparison is the point of this class; the composite adds the scale-from-zero usability span).
2. Per-leg tracing (`CreateActor` / `GetActor`) inside the composite span for regression localization.
3. Provision-per-sample (StopUser-style) so N users = N genuinely concurrent cold activations — clean ladder rungs.

If these fit the intent, I can push them onto this branch as a revision; equally fine as a follow-up after this merges as-is.

PR #893, opened 2026-08-12
Handle graceful termination in microvm ateom

sfunkenhauser:
Part of https://github.com/agent-substrate/substrate/issues/23

Trap SIGTERM in the ateom, and forward SIGTERM to all application containers.
Once all application containers exit, ateom will exit itself.

### DIFF
--- a/cmd/ateom-microvm/run.go
+++ b/cmd/ateom-microvm/run.go
@@ -94,6 +94,11 @@ type runningActor struct {
 	// post-restore dial), which loses both log forwarding and guest stats for
 	// this activation.
 	guestAgent *kata.AgentClient
+
+	// workloadIDs are the kata overlay-workload container ids (overlayWorkloadID of
+	// each container name) running in the guest. The SIGTERM handler signals and
+	// waits on these to gracefully stop the actor before the VM is torn down.
+	workloadIDs []string
 }
 
 // baseIDFile is a tiny snapshot file (under the checkpoint/restore dir) holding
@@ -147,6 +152,17 @@ const maxActorContainers = 25
 // id (names are unique within an actor) — even for containers named "x" and "x-ovl". A
 // "-ovl" suffix would let "x"'s workload id collide with the "x-ovl" carrier id.
 func overlayWorkloadID(name string) string { return name + "_ovl" }
+
+// overlayWorkloadIDs returns the overlay-workload container ids for the actor's
+// containers, in order. Recorded on runningActor so the SIGTERM handler knows
+// which guest workloads to signal and wait on.
+func overlayWorkloadIDs(ctrs []actorContainer) []string {
+	ids := make([]string, 0, len(ctrs))
+	for _, c := range ctrs {
+		ids = append(ids, overlayWorkloadID(c.name))
+	}
+	return ids
+}
 
 // actorContainer is one of the actor's containers prepared for the shared micro-VM:
 // its name (also the kata containerID + the overlay lower's find-paths subdir), the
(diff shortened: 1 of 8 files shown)


sfunkenhauser:
@BenTheElder PTAL

I've patched the e2e test fixes from #904 just to get a passing run, but will submit and rebase once that lands.

BenTheElder:
that's in, taking a look at the other commit

cmd/ateom-microvm/shutdown.go
```diff
@@ -0,0 +1,240 @@
...
+// out the grace period, escalating to SIGKILL.
+func gracefullyStopActor(ctx context.Context, id string, ra *runningActor) {
+	if ra == nil {
+		return
+	}
+
+	// Obtain a kata-agent client to signal the guest: reuse the log-forwarding
+	// connection if it's open, else dial a fresh one (best-effort). A dial we open
+	// here is closed here; ra.guestAgent belongs to the log forwarder.
+	agent := ra.guestAgent
```
BenTheElder: 🤖 **should-fix** 🟡 – This reads `ra.guestAgent` with `lock` released, and a concurrent teardown writes it. `gracefulShutdown` drops the lock before calling here so a suspend can still land, and `CheckpointWorkload` is deliberately not gated by `rejectIfDraining`. Its `teardownActor` does `ra.guestAgent.Close()` followed by `ra.guestAgent = nil`, so the field is written while this goroutine reads it — an unsynchronised read/write that `-race` will flag.

The observable damage is mild: a stale pointer means signalling a closed client and logging the error, and a nil one means fifteen seconds of `dialAgentRetry` against a VM that is being destroyed. Both are bounded and already tolerated.

The comment above `maps.Clone` has the right instinct about the map; the same reasoning applies one level down to the struct it hands out. Capturing what the drain needs — the agent and `workloadIDs` — into a value while the lock is still held would close it, and `workloadIDs` is immutable after construction so only the agent really needs copying.

sfunkenhauser: save the pointer to the ra.guestAgent while we hold the lock above to avoid the race. If the actor has already been torn down and the connection closed, the calls on ra.guestAgent will immediately fail. This is fine as the actor is already gone and there isn't anything else we need to clean up.

sfunkenhauser:
@BenTheElder PTAL

cmd/ateom-microvm/run.go
```diff
@@ -215,14 +231,24 @@ func (s *AteomService) RunWorkload(ctx context.Context, req *ateompb.RunWorkload
...
 	}
 
-	p := actorBootParams{
-		actorRef:     resources.ActorRef{Atespace: req.GetAtespace(), Name: req.GetActorName()},
-		actorUID:     req.GetActorUid(),
-		templateNS:   req.GetActorTemplateNamespace(),
-		templateName: req.GetActorTemplateName(),
-		containers:   req.GetSpec().GetContainers(),
-		assetPaths:   req.GetRuntimeAssetPaths(),
+	if err := s.rejectIfDraining(); err != nil {
```
BenTheElder, line 246: 🤖 **nit** 🟢 – The drain check lands after a side effect here, and the two boot RPCs disagree about the order. `RestoreWorkload` checks `rejectIfDraining` immediately after taking the lock (restore.go:85); this one runs `deactivateActorNetworking` first and only then decides to turn the request away.

Harmless in practice — a draining worker with no actor has nothing much to deactivate — but the reject is meant to be the cheap early-out, and having the two paths differ invites the next person to copy whichever one they read first. Moving it above the `deactivateActorNetworking` call makes them match.

BenTheElder (commented):
generally LGTM, will need a rebase

cmd/ateom-microvm/run.go
sfunkenhauser: Done

sfunkenhauser:
@BenTheElder  PTAL

BenTheElder (approved):

PR #898, opened 2026-08-12
Add UNSPECIFIED zero value for ActorSnapshotTagScope

### LINKED ISSUE #897: Add UNSPECIFIED zero value for ActorSnapshotTagScope
See https://github.com/agent-substrate/substrate/blob/main/docs/api-style-guide.md#51-enum-naming

laoj2:
Fixes #897

- [x] Tests pass


EItanya:
I don't fully understand why this is necessary? If the default is `ATESPACE`, why can't that be `0`?

laoj2:
> I don't fully understand why this is necessary? If the default is `ATESPACE`, why can't that be `0`?

substrate's API style guide (https://github.com/agent-substrate/substrate/blob/main/docs/api-style-guide.md#51-enum-naming) made an explicit decision of requiring `_UNSPECIFIED` values for the "not set" case so we can avoid issues like this: 

1. We add a new ACTOR_SNAPSHOT_TAG_SCOPE_XXX to the enum
2. A client gets released before the server. 
3. The stale server receives a request with the new `ACTOR_SNAPSHOT_TAG_SCOPE_XXX`
4. The proto field accessor in the server returns the default value: the server silently gets `_ATESPACE`, potentially making a wrong decision instead of failing

juli4n (approved):

Issue #896, opened 2026-08-12
Actor resource telemetry Phase 1: atelet reader and OTel emission


baizhenyu:
## What

Phase 1 of #550: atelet polls the `GetWorkloadStats` RPC that Phase 0 (#594) put on both ateom runtimes, and turns the samples into telemetry — template-level metrics for the TSDB, actor-level events for the log stream. After this phase, "how much CPU/memory is this actor using" has an answer you can query, not just an RPC you can call.

Phase 0 delivered the server side: #667 (RPC + attribution retention), #739 (gVisor cgroup read), #832 (micro-VM guest-agent read, in review). Nothing calls it yet.

## Emission model

Two channels, split exactly along #174's cardinality decisions:

* **Metrics (TSDB):** aggregated per ActorTemplate on each node before export. Labels: `ate.template.namespace`, `ate.template.name`, `ate.sandbox.class`, stats source — all bounded sets. Actor and atespace identity never appear as metric labels. (`internal/ateattr`'s metric-label comment should be reconciled with #174 decision 2 while here: template keys are sanctioned labels but currently documented in the identity block.)
* **Events (logs):** per-actor samples as structured log events carrying the full identity echoed by the RPC (`ate.actor.uid`, `ate.actor.name`, `ate.atespace`, template, class, source, the four measurements, `observed_at`) — same stream and label vocabulary as `actorlog`'s lifecycle events. Queryable per actor and per atespace in the log store; never a time series.

Transport, to be explicit: metrics ride the existing OTLP push (`serverboot.InitMetricsPushOnly` → the collector in `ate-otel-config.yaml`) — new instruments on a pipeline atelet already runs. Events do NOT go to the collector in this phase: like every event in this system they are structured JSON on stdout, picked up by the node logging agent, landing in the same store as the lifecycle events they join with. Nothing here uses the OTel Logs SDK today and the collector config carries no logs pipeline; switching events to OTLP logs later is additive (#503 is where that conversation lives).

Samples from the two runtimes are not comparable with each other (cgroup measures the sandbox including sentry overhead; guest-agent measures only the workload's containers) — the `source` label is what keeps them apart, and rollups must group by it.

## Emission cadence

[...]

Issue #44, opened 2026-05-21
High Concurrent Active Actors per Worker


brandonrjacobs:
Reading the architecture and overview, as I understand it only one actor per worker can be running / active at once. Is the goal to support multiple active actors per worker in the future?

thockin:
We spent some time on this topic in the past.  My fairly strong feeling (but always convinceable) is that it's WAY more complicated to support 1:N than 1:1.  Simplicity is really important here, especially in the first 6 months -- we have a LOT of work to tackle, this seems like a distraction.

1 worker to N actors means:
* intra-pod performance isolation (e.g. cgroups)
* more complicated pod lifecycle (externally e.g. termination, draining)
* more complicated intra-pod lifecycle management (approximating a full kubelet)
* more complicated autoscaling
* more complicated resource management (e.g. k8s in-place updates)
* more complex scheduling
* more complex APIs and CLIs

So I would flip it around - why would we want to do 1:N instead of 1:1 ?

bittermandel:
I wasn't aware this was the case until I read this issue. Just so I understand; does this mean one worker per node? It doesn't feel like it would take advantage of running really big nodes to scale better horizontally.

For background: our agent is centralized outside of the sandbox (or actor in this case)  and requires a long-running running connection, which would be tricky to achieve here. On top of that, our users interact with the sandboxis through a Vite HMR connection directly to the process. This experience would be jittery, prone to failure and require a proper control plane + decent resumption support across our application if we need to cycle through actors at all times.

In terms of resources, at our scale I believe we'd spend significant resources suspending and resuming workloads, though definitely less than it would take to just keep them running. Probably that'd be many gbit/s across all nodes for the snapshotting & resuming, which also isn't free.
Could we run one worker per core perhaps with session pinning and disk caching of snapshots etc? There's more ideas but I'm struggling to see how to scale this type of infra.

BenTheElder:
> wasn't aware this was the case until I read this issue. Just so I understand; does this mean one worker per node? It doesn't feel like it would take advantage of running really big nodes to scale better horizontally

No. One per pod, which runs an interior sandbox that mutates to run different actors as needed. But each of those "ateom" pods runs one workload at a time. When the workload is suspended the same pod can host another workload.

A node has one "atelet" to coordinate amongst the "ateom" pods hosting workloads, which are many.

It already supports multiple workers per node this way, as you can see in the demo in the README.

thockin:
@bittermandel 

[...]

PR #899, opened 2026-08-12
ateom: add GetActiveWorkloadStats RPC

baizhenyu:
First piece of #896 (Phase 1 of #550): `GetActiveWorkloadStats`, a parameterless sibling to `GetWorkloadStats` for a scraper that enumerates ateoms and holds no worker-to-actor mapping.

## Contract

* An **available** ateom answers an empty `stats` list, not an error — an idle worker is a normal thing for a scraper to find.
* An **executing** ateom answers the same sample the keyed read would give, wrapped in a one-element list.
* `FAILED_PRECONDITION` keeps the meaning it has on `GetWorkloadStats`: executing, but no numbers yet (mid-boot, guest not answering) — skip this sample, take the next one.
* Consumers MUST attribute each sample solely from the identity echoed inside it, never from a mapping they hold: with no asserted uid, the response is the only statement of who was measured. This rule is on the RPC's doc comment.

## Implementation

Both runtimes share the measurement half of their existing `GetWorkloadStats`, extracted as `sampleSandbox` (gVisor) / `sampleGuest` (micro-VM) — the discovery handler is attribution-load, empty-if-nil, same helper, re-check. No change to what is measured or how.

The one deliberate semantic split: a lifecycle transition underneath the lock-free read is `NOT_FOUND` on the keyed read (the caller asserted an actor that is now gone; its mapping wants re-resolving) but `FAILED_PRECONDITION` on the discovery read (there is no requested actor to disown — the numbers just cannot be attributed to any single actor this tick).

Naming note: #896's sketch called this `GetCurrentActiveWorkloadStats`; "current" was redundant with "active", so it landed shorter.

## Testing

* Both runtimes: available → empty list; executing mid-boot (no sandbox cgroup / no guest target) → `FAILED_PRECONDITION`; and a cross-check pinning that the discovery sample is identical to the keyed sample against the same fixture — one measurement, two addressing modes.
* `make verify` clean through the proto checks (`go-generate.sh` confirms the regenerated `ateom.pb.go` / `ateom_grpc.pb.go` are canonical).

Part of #896, toward #550.

### DIFF
--- a/cmd/ateom-gvisor/stats.go
+++ b/cmd/ateom-gvisor/stats.go
@@ -113,9 +113,20 @@ func (s *AteomService) GetWorkloadStats(ctx context.Context, req *ateompb.GetWor
 		return nil, status.Errorf(codes.NotFound, "ateom is executing actor %q, not the requested %q", active.UID, req.GetActorUid())
 	}
 
-	resp, err := s.sampleSandbox(active)
+	sample, err := s.sampleSandbox(active)
 	if err != nil {
-		return nil, err
+		// The requested actor is the active one but its cgroup is not there.
+		// Most often that is a poll landing in the boot: the ateom retains the
+		// attribution from the moment it accepts the actor, before runsc has
+		// created the leaf. The other way in is a sandbox that went away
+		// underneath the read, which the next CheckpointWorkload turns into the
+		// NOT_FOUND above. Either way it is "no numbers right now" and the
+		// caller should take the next sample, so FAILED_PRECONDITION. Anything
+		// else is a real read failure.
+		if errors.Is(err, fs.ErrNotExist) {
+			return nil, status.Error(codes.FailedPrecondition, "no sandbox cgroup to measure yet")
+		}
+		return nil, status.Errorf(codes.Internal, "reading sandbox cgroup: %v", err)
 	}
 
 	// Re-check that the same workload is still the active one. The read above
@@ -133,7 +144,7 @@ func (s *AteomService) GetWorkloadStats(ctx context.Context, req *ateompb.GetWor
 		return nil, status.Errorf(codes.NotFound, "ateom stopped executing actor %q while the sample was being taken", req.GetActorUid())
 	}
 
-	return resp, nil
+	return &ateompb.GetWorkloadStatsResponse{Sample: sample}, nil
 }
 
 // GetActiveWorkloadStats implements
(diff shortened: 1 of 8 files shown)


internal/proto/ateompb/ateom.proto
```diff
@@ -334,3 +352,15 @@ message GetWorkloadStatsResponse {
 
   int64 observed_at_unix_nano = 12;
 }
+
+message GetActiveWorkloadStatsRequest {
+}
+
+message GetActiveWorkloadStatsResponse {
```
git286, line 379: Nesting one RPC's response message inside another's is a proto anti-pattern (I actually expect AI to know that :))

I think something like following is a better design:

message WorkloadStatsSample

message GetWorkloadStatsResponse       { WorkloadStatsSample sample = 1; }
message GetActiveWorkloadStatsResponse { repeated WorkloadStatsSample samples = 1; }

However, I have another question: why GetActiveWorkloadStatsResponse needs to return a list at all? One worker pod can only have 1 workload (actor), right?

cmd/ateom-gvisor/stats.go
```diff
@@ -144,6 +133,62 @@ func (s *AteomService) GetWorkloadStats(ctx context.Context, req *ateompb.GetWor
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Same re-check as GetWorkloadStats, different code: with no uid asserted
+	// there is no "requested actor" for NOT_FOUND to disown. A transition
+	// underneath the read just means these numbers cannot be attributed to any
+	// single actor, so the answer is the discovery read's usual "no numbers
+	// this tick, take the next sample".
+	if s.activeActor.Load() != active {
```
git286: I am a little worried if we treat everything in-active as error. Can we instead return a state in the response, and the sample is nil when the state is not active to distinguish true error vs expected non results (when status is in-active), especially given that the caller is expected to call this RPC without knowing the state of the workload.

internal/proto/ateompb/ateom.proto
baizhenyu: I refactored the proto as you described. The list was chosen just in case if a worker ever hosts more than one actor (#44) without a wire break. But I reverted that to return a single response.

cmd/ateom-gvisor/stats.go
baizhenyu: I added `WorkloadState` enum to distinguish between real error and other states.

internal/proto/ateompb/ateom.proto
git286: After think it again, I think following might be a better design:

oneof result {
    WorkloadStatsSample sample = 1; 
    NoSampleReason      reason = 2;   // NO_WORKLOAD | NOT_MEASURABLE_YET etc
}

WDYT?

cmd/ateom-microvm/stats.go
```diff
@@ -150,7 +131,86 @@ func (s *AteomService) GetWorkloadStats(ctx context.Context, req *ateompb.GetWor
...
+// sampleGuest reads the guest's container cgroups through the agent and builds
+// the sample attributed to active. Every error it returns means "no numbers
+// right now" rather than a bug -- there is deliberately no Internal class on
+// this runtime, since a guest that has stopped answering is routine here --
+// and it comes back raw because the two RPCs express that differently: an
+// error code for the keyed read, a normal EXECUTING answer for the discovery
+// read. Callers re-check s.activeActor against the pointer they loaded after
+// this returns; the read holds no lock.
+func (s *AteomService) sampleGuest(ctx context.Context, active *ateomstats.ActorAttribution) (*ateompb.WorkloadStatsSample, error) {
+	// The actor is the one here, but there is no guest to ask yet. Usually that
```
git286, line 205: See my other comment, I think it would be better to return sample, nosample_reason, and error to distinguish expected no sample cases vs unexpected errors

cmd/ateom-microvm/stats.go
```diff
@@ -150,7 +131,86 @@ func (s *AteomService) GetWorkloadStats(ctx context.Context, req *ateompb.GetWor
...
+			State: ateompb.WorkloadState_WORKLOAD_STATE_EXECUTING,
+		}, nil
+	}
+
+	// Same re-check as GetWorkloadStats, different answer: with no uid asserted
+	// there is no "requested actor" for NOT_FOUND to disown, and a transition
+	// underneath the read just means these numbers cannot be attributed to any
+	// single actor. Report the state as of now, with no sample -- the next tick
+	// resolves it either way.
+	if latest := s.activeActor.Load(); latest != active {
```
git286, line 167: AI finding:

cmd/ateom-microvm/stats.go:165 (and the mirror at cmd/ateom-gvisor/stats.go:181) — the transition re-check has no test on either runtime.

  This is the branch the PR description calls out as its one deliberate semantic split, and the latest commit grew it from one outcome to two: a transition to a different
  actor reports EXECUTING with no sample, a transition to idle reports AVAILABLE. Neither is exercised. Every activeActor.Store in both test files is setup before the
  call, so nothing ever changes the attribution during a sample — the latest != active condition is always false under test.

internal/proto/ateompb/ateom.proto
baizhenyu: Yeah, I think this is a better design. Uploaded a new commit to reflect this change.

cmd/ateom-microvm/stats.go
baizhenyu: Done

cmd/ateom-microvm/stats.go
baizhenyu: Updated the test case.

cmd/ateom-microvm/stats.go
```diff
@@ -150,7 +131,89 @@ func (s *AteomService) GetWorkloadStats(ctx context.Context, req *ateompb.GetWor
...
+// whatever is executing with no identity asserted. Same lock discipline as
+// GetWorkloadStats above, for the same reasons.
+func (s *AteomService) GetActiveWorkloadStats(ctx context.Context, req *ateompb.GetActiveWorkloadStatsRequest) (*ateompb.GetActiveWorkloadStatsResponse, error) {
+	active := s.activeActor.Load()
+	if active == nil {
+		return noSample(ateompb.NoSampleReason_NO_SAMPLE_REASON_NO_WORKLOAD), nil
+	}
+
+	sample, err := s.sampleGuest(ctx, active)
+	if err != nil {
```
git286, line 151: Should we check the error type here, similar to the gvisor implementation?

baizhenyu: Mostly no — vsock failures have no type structure to check (closed/timeout/refused are all routine "guest not answering", unlike local-fs ENOENT-vs-else). But the question flushed out one real case: the stale-target check is an invariant violation, not a routine state, and it was hiding in the routine bucket. 0ef14de2 maps it to Internal on both reads via a sentinel, mirroring the gVisor split.

git286 (approved):

PR #902, opened 2026-08-12
skill: add triage-issues for labeling open issues

aditya-shantanu:
## What changed and why

Adds `.agents/skills/triage-issues/SKILL.md` — a skill agents invoke (via "triage issues") to classify open GitHub issues with the correct `kind/`, `area/`, and `prio/` labels.

Of 202 open issues, 13 are currently unlabeled. Several labeled issues also lack a `kind/` label or use `kind/feature` for what are really design discussions.

### What the skill does

- Documents the full label taxonomy with clear when-to-apply rules for every existing label
- Processes unlabeled issues first, then reviews labeled issues for obvious gaps
- Uses `--add-label` (not `--label`) so existing human-applied labels are never overwritten
- Outputs a report table of every change made, plus flagged uncertainties for maintainer review
- Never removes labels or adds `wontfix`/`duplicate`/`invalid` — those stay with maintainers

### Two new labels proposed

The current taxonomy has two gaps I found while analyzing the backlog:

**`kind/design`** — Several issues are design discussions or investigations that must be resolved before any code is written (e.g. _"Design discussion: WorkerPool vs. node-level Actor runtime"_, _"Identify what metrics to use for HPA autoscaling"_, _"Measure atenet-router request capacity"_). These are currently unlabeled or mislabeled as `kind/feature`. The skill creates this label if it doesn't exist.

**`area/benchmarking`** — The repo has a full `benchmarking/` directory but no corresponding area label. Performance measurement issues have nowhere to go. The skill creates this label if it doesn't exist.

## Checklist

- [x] Issue is linked above
- [x] Tests pass locally (`go test ./...`)
- [x] Root-gated tests pass if applicable (`hack/run-root-tests.sh`)
- [x] Documentation updated if behavior changed (N/A)

### DIFF
--- a/.agents/skills/triage-issues/SKILL.md
+++ b/.agents/skills/triage-issues/SKILL.md
@@ -110,6 +110,13 @@ capacity, or researching unknowns before any code is written, it is `kind/design
 Keywords: "Design discussion", "Investigate", "Explore options", "Identify what",
 "Measure", "Research".
 
+The clearest way to distinguish `kind/design` from `kind/feature`:
+- **`kind/feature`**: you could open a PR today — the *what* and *how* are known.
+- **`kind/design`**: a decision or investigation must happen first; a PR cannot be written yet.
+
+When in doubt, ask: *"Could someone reasonably start coding this issue right now?"*
+If yes → `kind/feature`. If no → `kind/design`.
+
 **4. Is it a code quality improvement with no behavior change?**
 Refactoring, renaming, consolidating duplicated code, removing dead code → `kind/cleanup`.
 
--- a/.agents/skills/triage-issues/SKILL.md
+++ b/.agents/skills/triage-issues/SKILL.md
@@ -55,13 +55,18 @@ Apply all areas that are touched. Most issues need one or two; a few need more.
 | `area/cli` | `kubectl-ate` CLI plugin |
 | `area/demos` | Demo applications in `demos/` |
 | `area/benchmarking` | Performance benchmarks and capacity measurements (`benchmarking/`) |
+| `area/reliability` | Reliability, stability, error recovery, and uptime concerns |
 
-`area/benchmarking` does not exist yet — create it first:
+`area/benchmarking` and `area/reliability` do not exist yet — create them first:
 ```bash
 gh label create "area/benchmarking" \
   --repo agent-substrate/substrate \
   --description "Performance benchmarks and capacity measurement" \
   --color "f9d0c4"
+gh label create "area/reliability" \
+  --repo agent-substrate/substrate \
+  --description "Reliability, stability, error recovery, and uptime" \
+  --color "e4e669"
 ```
 
 ### `prio/` — priority
@@ -121,11 +126,11 @@ necessarily land. A bug in actor scheduling reported via the API is `area/api` +
 
 ### Step 1 — Create missing labels (once)
 
-Check whether `kind/design` and `area/benchmarking` exist before creating them:
+Check whether `kind/design`, `area/benchmarking`, and `area/reliability` exist before creating them:
 
 ```bash
 gh label list --repo agent-substrate/substrate --json name \
-  --jq '[.[].name] | contains(["kind/design"])'
+  --jq '[.[].name] | contains(["kind/design", "area/benchmarking", "area/reliability"])'
 ```
 
 Create each if not present using the `gh label create` commands above.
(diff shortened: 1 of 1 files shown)


BenTheElder:
> kind/design — Several issues are design discussions or investigations that must be resolved before any code is written (e.g. "Design discussion: WorkerPool vs. node-level Actor runtime", "Identify what metrics to use for HPA autoscaling", "Measure atenet-router request capacity"). These are currently unlabeled or mislabeled as kind/feature. The skill creates this label if it doesn't exist.

the distinction between this and a feature seems fuzzy ...

aditya-shantanu:
Good point. Added a concrete decision rule: if someone could open a PR today the *what* and *how* are settled → `kind/feature`; if a decision or investigation must happen first → `kind/design`. Also added the quick test: *'Could someone reasonably start coding this right now?'* (44a296a6)

.agents/skills/triage-issues/SKILL.md
```diff
@@ -0,0 +1,218 @@
...
+| `area/identity` | Actor identity, token issuance, JWT auth, ActorIdentity extension |
+| `area/dev-infra` | CI/CD, GitHub Actions, presubmit checks, tooling, proto generation |
+| `area/tests` | Test coverage, flaky tests, test infrastructure (not a bug in production code) |
+| `area/cli` | `kubectl-ate` CLI plugin |
+| `area/demos` | Demo applications in `demos/` |
+| `area/benchmarking` | Performance benchmarks and capacity measurements (`benchmarking/`) |
+| `area/reliability` | Reliability, stability, error recovery, and uptime concerns |
+
+`area/benchmarking` and `area/reliability` do not exist yet — create them first:
+```bash
```
haiyanmeng: These two labels already exist.

aditya-shantanu: Partially — `area/benchmarking` and `area/reliability` do exist now, so I dropped their create-block in fd990bcb. `kind/design` still does not exist (`gh api .../labels/kind%2Fdesign` → 404), so its create command stays.

haiyanmeng (dismissed):

.agents/skills/triage-issues/SKILL.md
```diff
@@ -0,0 +1,206 @@
+---
```
bowei, line 1: So I don't think we should have the bot muck directly with the labels just yet.

Instead what we should do is the following:

1) There needs to be a label that the bot script can use to cause it to skip an issue or PR. This is to enable people to skip issues that cause the bot problems until the problem is fixed.

2) I would prefer to start with the bot in read-only mode that runs a more deterministic way:

bot looks at gh and does issue triage. BTW -- gh can become rate-limited pretty easily with these batch operations so people should know that this can happen.

bot outputs a yaml (or json) formatted output of the triage results. user can edit the yaml to remove entries that were done badly etc.

A script that can take the formatted output and actually apply it.

We can wrap the whole thing in an agent-generate TUI for ease of use.

aditya-shantanu: How are existing labels looking like ?
I, for example and FWIW, did the read only mode for a few days before sending out this skill.

aditya-shantanu: What I'm saying is that once this is checked in, we can have the community use it more and directly edit the skill if something that happens doesn't meet someone's expectations.

.agents/skills/triage-issues/SKILL.md
```diff
@@ -0,0 +1,284 @@
...
+| `area/security` | Auth, mTLS, RBAC, threat detection, credential management, PodCertificate |
+| `area/identity` | Actor identity, token issuance, JWT auth, ActorIdentity extension |
+| `area/dev-infra` | CI/CD, GitHub Actions, presubmit checks, tooling, proto generation |
+| `area/tests` | Test coverage, flaky tests, test infrastructure (not a bug in production code) |
+| `area/cli` | `kubectl-ate` CLI plugin |
+| `area/demos` | Demo applications in `demos/` |
+| `area/benchmarking` | Performance benchmarks and capacity measurements (`benchmarking/`) |
+| `area/reliability` | Reliability, stability, error recovery, and uptime concerns |
+
+`area/benchmarking` does not exist yet — include its creation in the proposal, and
```
bowei: It should say:

If any labels don't exist, the follow command can be used to create, vs just talking about a specific label. This may yield weird results

bowei (commented):
one comment

.agents/skills/triage-issues/SKILL.md
aditya-shantanu: Done in b5baa5fd — dropped both hardcoded "X does not exist yet" blocks (they had already gone stale) and replaced them with a generic rule: Step 1 lists the repo's labels, compares against the taxonomy, and the proposal includes a parameterized `gh label create` for whichever are missing.

PR #904, opened 2026-08-12
Delete namespaces on test completion

sfunkenhauser:
E2E test WorkerPools (and thus pods) weren't cleaned up until all tests were finished running. This resulted in resource exhaustion and pods not being scheduled. If a test completes successfully delete the namespace immediately.

- [ x ] Tests pass

### DIFF
--- a/internal/e2e/namespace.go
+++ b/internal/e2e/namespace.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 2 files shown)


internal/e2e/namespace.go
```diff
@@ -100,6 +102,22 @@ func CreateNamespace(t *testing.T) *Namespace {
 		t.Fatalf("Failed to create namespace %s: %v", nsName, err)
 	}
 
+	// Release the namespace as soon as this test is done rather than at the end of
+	// the run. A suite's namespaces each carry a WorkerPool, and its worker pods
```
BenTheElder, line 106: We also might need to revisit that, or seriously downsize them or ...

BenTheElder (approved):
thanks! this should help with flakiness, cc @EItanya

Issue #853, opened 2026-08-11
[Design] Actor Otel Telemetry Relay


git286:
# Design: per-actor otel telemetry relay

**Status:** draft / for discussion
**Related:** [#761](https://github.com/agent-substrate/substrate/issues/761) (actor instrumented metrics issue), [#503](https://github.com/agent-substrate/substrate/issues/503) (telemetry continuity), [#450](https://github.com/agent-substrate/substrate/issues/450) (lifecycle hooks)

---

## 1. Summary

Actors cannot correctly identify their own telemetry, so all telemetry from an ActorTemplate collapses into a single time series with many writers.

This proposes that **ateom** expose a fixed OTLP endpoint inside the worker pod, stamp trusted `ate.dev/*` identity onto everything that arrives.

The principle is one Substrate already applies to logs:

> **Attribution belongs to the channel, not the payload.**

`internal/actorlog/logger.go:132-158` already does this for stdout — atelet injects `ate.dev/actor_atespace`, `actor_name`, `actor_uid`,
`actor_template_namespace`, `actor_template_name`, `container_name` into every log line, out of band. The actor never touches them, so they can't be frozen, forged, or lost. Metrics and traces are the signals that haven't caught up.

It's also the same division of labour Kubernetes already uses. Under Prometheus, identity comes from the scrape target, not the workload — `instance` and `pod`are assigned by the scraper, and emitting them yourself is an anti-pattern.

Actors can't be scraped, so the relay is that job relocated to somewhere it can still happen. This is not a novel mechanism; it's the conventional one adapted to push.

## 2. Goals

1. Every actor's telemetry is attributed to **that actor**, with no cooperation from actor code beyond pointing at an endpoint.
2. Attribution is **trusted** — derived by the platform, not asserted by the workload — so it can back per-tenant views, quota, and billing later.
3. Attribution **survives suspend/resume/migration** with nothing re-derived.
4. Storage cardinality is **bounded and independent of actor count**.
5. Works on **both runtimes**, gVisor and micro-VM.

## 3. Non-goals

- Buffered-telemetry loss on `onPause: Data`. That is #503 and needs #450's `PreSuspend`. Complementary, not a substitute.
- Control-plane component telemetry, which works today.
- Per-actor metric series in long-term storage — explicitly rejected

## 4. Architecture

```
┌─ worker pod ────────────────────────────────────────┐
│                                                     │
│  ┌─ actor sandbox ─────────┐      ┌─ ateom ───────┐ │
[...]

PR #905, opened 2026-08-13
Rethinking the ActorTemplate and ActorTemplateVersion relationship, inspired by ActorSnapshotTag

zoez7:
Previously, the `ActorTemplate` object expressed both a group and a pointer.

* It's a grouping mechanism of `ActorTemplateVersion` resources.
* It has a pointer to one of the ATVs, i.e. `default_version_on_create` points to an ATV.

In this modified API, we separate out the "pointer" concept into `ActorTemplateVersionTag`, similar to `ActorSnapshotTag`.
After this, the `ActorTemplate` resource itself doesn't have any field anymore, therefore the `ActorTemplate` as a grouping concept that only needs a name and a ID.

Advantages are:
* Users can have multiple named tags and is not restricted to one (`default_version_on_create`) only. Users can make their own decision of which Tag is considered "default". 
* Users can decide _when_ to use a tag, and not restricted to a "tag" being available during CreateActor only.

below is an illustration of the relationships. This PR doesn't contain rollout policy tag, but we can add that in the future.

```
// +-- ActorTemplate ---------------------------------------+
// |                           +------------+               |
// |   +------------+          | RolloutTag |               |
// |   | DefaultTag |          +------------+               |
// |   +------------+        10% /          \ 90%           |
// |         |                  /            \              |
// |         v                 v              v             |
// |     +-------+         +-------+      +-------+         |
// |     | ATV-1 |         | ATV-2 |      | ATV-3 |         |
// |     +-------+         +-------+      +-------+         |
// +--------------------------------------------------------+
```

The API to create an Actor then becomes:
```
1. at := CreateActorTemplate(...)
2. atv := CreateActorTemplateVersion(at, spec)
4. tag := CreateActorTemplateTag(tag{"atespace", "my-default"}, atv) // replacement to UpdateActorTemplate.default_version_on_create
5. actor := CreateActor(tag)
```

@juli4n Let me know what you think.

I'm happy to see the amount of code I was able to delete in this PR :)


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -313,50 +312,49 @@ message ActorSnapshotRef {
...
+// ActorTemplateVersion and is retargeted via UpdateActorTemplateVersionTag.
+message ActorTemplateVersionTag {
   // Common resource metadata: atespace, name, uid, version, timestamps.
   ResourceMetadata metadata = 1;
 
-  // default_version_on_create names the ActorTemplateVersion used by
-  // CreateActor calls that do not pin a version. If unset, CreateActor
-  // without an explicit version fails with FailedPrecondition.
-  ObjectRef default_version_on_create = 3;
+  // Output only. Set at creation to the tagged version's actor_template_id;
```
HavenXia: Why we need the actor_template_id here if you can extract it from actor_template_version? Is it to save one read?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -313,50 +312,49 @@ message ActorSnapshotRef {
...
+  // the tag can only ever point at versions of this actor template.
+  string actor_template_id = 2;
```
HavenXia: Who will generate the id? There's no message ActorTemplate anymore

pkg/proto/ateapipb/ateapi.proto
zoez7: This is useful to support APIs like `ListActorTemplateVersionTags(actor_template_id)`, so we don't have to first list all ATVs by actor_template_version

pkg/proto/ateapipb/ateapi.proto
zoez7: The user will decide this. Perhaps calling it application_name would be more appropriate.

zoez7: That's a good point actually, the server would need to keep record of mapping between (name <-> id)

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -671,12 +624,10 @@ message GetActorTemplateVersionRequest {
 }
 
 message ListActorTemplateVersionsRequest {
-  // The parent ActorTemplate whose versions to list. An unset ref lists
-  // versions across all templates.
-  ObjectRef actor_template = 1;
+  // The actor template whose versions to list. Required.
+  string actor_template_id = 1;
```
krisztianfekete: This field is where `ate.template.name` comes from, and it's a metric label today. We are also planning to leverage it in #853. Can we ensure this is globally unique?

mini-cookie:
Good mr！I want to add support for the E2B protocol. Building on the ActorTemplateVersion design, we should be able to implement the fork and checkpoint endpoints in ate-api-server fairly quickly～

pkg/proto/ateapipb/ateapi.proto
zoez7: Added back ActorTemplate that just have ResourceMetadata with UID.

pkg/proto/ateapipb/ateapi.proto
zoez7: Added back ActorTemplate that just have ResourceMetadata with UID.

zoez7:
@mini-cookie the slack discussion thread for this is https://cloud-native.slack.com/archives/C0B6M3E2J3D/p1786656116150749 feel free to add your thoughts.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -248,6 +257,9 @@ message Actor {
 
   string in_progress_local_snapshot_name = 12;
 
+
+  ActorTemplateVersion actor_template_version = 13;
```
krisztianfekete: Can this mirror `ActorSnapshotSource` right below?
I am thinking of two things:

- Our lifecycle histogram reads the template from the request on `create` but from the stored Actor on `resume/suspend/pause/delete`. So if you resolve the tag away, only `create` can label with it and the rest can't, which means I can't use it at all. If it's on the Actor it's actually a better label than the ATV name. Same reason OTel sets `container.image.tags` next to the digest.

- We need an ATV uid too, for the same reason we have [here](https://github.com/agent-substrate/substrate/pull/905/changes#diff-f29aa971d0e861ade55215208c2d5e2eaa7d9f743969517826128207b1e4095eR337-R340) for snapshot_uid, so names come back after a delete. We already pair `ate.actor.uid` with `ate.actor.name` for this in spans/logs.
 
Unrelated, just curious: field 13 was an ObjectRef on main and is now the whole message, so every Actor read carries a copy of the ATV's phase and golden_snapshot.
Is it intentional?

zoez7:
Closing this pull request, we agreed on:

1. move the AT CRD to substrate as-is, immutable. No ATV resource on its own.
2. ActorTemplateTag addresses a usability concern, is a P1 since users can work around the limitation by having a "pointer to a ActorTemplate" that they maintain outside of substrate.
3. There will be no "ActorTemplateGroup" concept for now, until a concrete use case emerges.

PR #907, opened 2026-08-13
benchmarks: durdir benchmark that evaluates the efficacy and performance of DurDir

### LINKED ISSUE #673: Benchmark DurDir
We should have a benchmark that evaluates the efficacy and performance of DurDir. I'm thinking roughly:

* Write an X byte file of random bytes to DurDir
* Serve file
* [LOOP ENTRY POINT]
* Suspend
* Serve file (zero changes expected to validate DurDir is Dur(able)
* re-serve file (were there lazy loading bottlenecks? Caching performance improvements?)
* Overwrite file with new X byte file (do infinite overwrites cause infinite snapshot size growth?)
* [RESUME LOOP]

Interesting signals:
* Write speed
* Serving speed
   * Relative timings of first post-write request, first post-suspend request and second post-suspend request may reveal interesting properties of underlying infrastructure.
* Resume/Suspend speed over time (Locust gathers percentiles, but we will at least be able to see if numbers increase over time by checking periodic statistics emitted by Locust).
* Whether served bytes match expected bytes (consider request failed if not)
* Image size over time (not gathered by-default by Locust, can defer solving this)

We can consider running this same test with data stored in RAM, but RAM is expected to be lost on occasion (e.g. upgrades), so DurDir is most salient.

sairajp-rewind:
# Benchmark DurDir

Fixes #673

Adds a `DurDirUser` load-generation workload that exercises the DurableDir
suspend/resume loop end to end, verifies every served byte against a SHA-256
digest, and emits separable latency percentiles for each step.

## The loop

    per VU, first iteration:  create -> resume -> WriteDisk -> ReadDisk+verify
    steady state:             suspend -> resume -> ReadDisk+verify (cold)
                                      -> ReadDisk+verify (warm)
                                      -> WriteDisk (overwrite, TRUNCATE)

Under `onCommit: Data` the container cold-boots from the OCI image and process
memory is discarded, so a matching digest after resume can only have come from
the restored DurableDir. That is the durability assertion.

## Results

All six scenarios ran on GKE/gvisor at 1 VU for 1m each: `Data` vs `Full`
snapshot scope, explicit vs implicit resume, and a 5/10/64 MiB size sweep.
**Zero failures across every run**, so every served byte matched its digest in
all six.

**Snapshot growth over repeated overwrites:** `SuspendActor` latency stayed flat
across consecutive overwrite-and-suspend cycles on the same DurableDir volume.
The loop overwrites with `WRITE_MODE_TRUNCATE`, so the file is exactly X bytes
after every write and the captured directory contents are the same size every
cycle. Nothing accumulates across suspends. That is the growth question the
issue asks about.

Latency percentiles per step are in the run artifacts.

## Change surface

- **glutton:** `WriteDisk` returns size + sha256; new `ReadDisk` with a
  `READ_MODE_DIGEST_ONLY` mode for measuring restore cost without paying wire
  transfer; disk RPCs exposed over HTTP mode.
- **manifests:** two new ActorTemplates, `glutton-durdir-{data,full}`, with a
  `durableDir` volume and `onPause: Full` / `onCommit: {Data,Full}`.
- **boomer:** shared actor-lifecycle plumbing extracted from the ping task, then
  a `DurDirUser` task on top of it; a general `resume_mode` knob.
- **harnesses:** `--workload` selects the task at deploy time; `durdir.py` stub,
  typed dynconfig flags, six nightly scenarios.

The two boomer commits above are incremental extractions made as the second
workload landed. The final package layout for `internal/benchmarking/boomer`
lands as a follow-up PR.

## Testing

- `go build ./...`, `go test -race ./cmd/benchmarking/... ./internal/benchmarking/...`
  clean, no race warnings.
- `hack/verify-all.sh`: all nine checks pass. Python protos regenerated, tree
  clean.
- Both ActorTemplates reach `Ready`; golden snapshots confirmed in the bucket.
- **Manual durability proof:** 200 MiB payload written, suspended, resumed, and
  re-read with matching sha256 across every read. Process memory discarded and
  container cold-booted in between.
- **Scale validation:** DurableDir persistence verified up to 1 GiB with 0
  failures.
- **Regression gate:** `glutton_baseline_5_users` ran 919 requests with 0
  failures and 7 ms ping latency post-rebase. No regression on the existing
  benchmark.

## Deliberately out of scope

- **Image size over time:** `ActorSnapshot` has no size field, so there is
  nothing for a client to read. The issue permits deferring this. `SuspendActor`
  latency is the available proxy; `atelet.snapshot.size` is the server-side one.
- **boomer package restructure:** landing as a follow-up so this PR's files stay
  reviewable in place.
- **RAM-backed variant:** glutton already has `WriteRAM`; small follow-up.

### DIFF
--- a/benchmarking/README.md
+++ b/benchmarking/README.md
@@ -62,6 +62,53 @@ You can also configure things like the number of users, how quickly those users
 are spawned, the frequency with which requests are made and whether or not tracing is
 enabled.
 
+User classes implemented in boomer rather than Python are selected at deploy
+time — the stack runs one per deployment:
+
+```bash
+./benchmarking/locust/deploy.sh --deploy --user-class durdir
+```
+
+### Headless (automation only)
+
+`runner.py` runs a test without the web UI, writing CSVs, logs and traces to
+`--dest`. The nightly automation submits it as a Job on the test cluster; it is
+not a local entry point. See [automation/README.md](automation/README.md).
+
+```bash
+python3 runner.py -f tests/<user-class>.py -t 1m -u 1 --name <run-name> --dest /tmp/bench
+```
+
+Test-specific flags are appended to the same command; see the sections below.
+
+### DurDir Benchmark
+
+The DurDir benchmark evaluates actor suspend/resume performance, disk persistence overhead,
+and state restoration latency when a durable directory is attached to the actor.
+
+#### DurDir Configuration Knobs
+
+* `--durdir-file-size-bytes`: Size in bytes of the data file (default `8388608` = 8 MiB).
+* `--resume-mode`: Resume trigger mode:
+  * `explicit` (default): Client invokes the `ResumeActor` RPC before sending traffic.
+  * `implicit`: Client sends traffic through the router without an explicit wake RPC, testing traffic-triggered resume.
+* `--durdir-read-mode`: Verification read mode:
+  * `data` (default): Server returns full payload bytes for client-side SHA-256 verification.
+  * `digest`: Server hashes the file and returns size and digest, reducing network transfer.
+* `--durdir-template`: ActorTemplate name:
+  * `glutton-durdir-data` (default): Attaches a durable data directory without memory snapshot restore.
+  * `glutton-durdir-full`: Attaches a durable data directory and performs a full memory snapshot restore.
+
+#### DurDir Reported Metrics
+
+* `DurDirWrite`: Initial truncate-write creating the data file.
+* `DurDirServeInitial`: First read immediately following file creation.
+* `SuspendActor`: Actor suspend latency (snapshot creation + persistence upload).
+* `ResumeActor`: Actor resume latency.
+* `DurDirServeAfterResume`: First read after resume (measures page faults / lazy load overhead on restored volume).
+* `DurDirServeWarm`: Subsequent read within the same active cycle (cached state baseline).
+* `DurDirOverwrite`: In-place file overwrite with checksum verification.
+
 ### Viewing Traces
 You must have enabled otel tracing for your cluster to view traces.
 
--- a/benchmarking/locust/common/boomer_config.py
+++ b/benchmarking/locust/common/boomer_config.py
@@ -17,6 +17,8 @@
 Flag registration lives in the modules that own each flag:
   * --trace-probability             → common.trace.init_tracing
   * --min-wait-time / --max-wait-time → common.wait_time.init_wait_time
+  * --resume-mode                   → common.resume_mode.add_resume_mode_arguments
+  * --durdir-*                      → common.durdir_config.add_durdir_arguments
 
 This module ties them together so boomer-Go workers can pick up the values
 the operator set in the web UI form:
@@ -34,17 +36,19 @@ import json
 import logging
 from collections.abc import Iterable
 
-from locust import events
-from locust.env import Environment
-
-from common.trace import init_tracing
-from common.wait_time import init_wait_time
-
 logger = logging.getLogger(__name__)
 
-# Boomer-tunable flags. CLI form ("--foo-bar") is converted to the
-# attribute / JSON-key form ("foo_bar") by _attr().
-_FLAGS = ("--trace-probability", "--min-wait-time", "--max-wait-time")
+# Boomer-tunable flags and their types. CLI form ("--foo-bar") is converted
+# to the attribute / JSON-key form ("foo_bar") by _attr().
+_FLAGS = {
+    "--trace-probability": float,
+    "--min-wait-time": float,
+    "--max-wait-time": float,
+    "--durdir-file-size-bytes": int,
+    "--resume-mode": str,
+    "--durdir-read-mode": str,
+    "--durdir-template": str,
+}
 
 
 def _attr(flag: str) -> str:
@@ -56,8 +60,8 @@ def build_config_json(argv: Iterable[str]) -> str:
     --config-json flag. Unknown args are ignored; unset flags are omitted so
     boomer falls back to its own defaults."""
     p = argparse.ArgumentParser(add_help=False)
-    for flag in _FLAGS:
-        p.add_argument(flag, type=float)
+    for flag, type_func in _FLAGS.items():
+        p.add_argument(flag, type=type_func)
     parsed, _ = p.parse_known_args(argv)
     cfg = {
         _attr(f): getattr(parsed, _attr(f))
@@ -71,6 +75,14 @@ def init_boomer_config() -> None:
     """Ensure the owning modules have registered the boomer-tunable flags,
     then expose their current values at /boomer-config so boomer-Go workers
     can fetch them at runtime."""
+    from locust import events
+    from locust.env import Environment
+
+    from common.durdir_config import add_durdir_arguments
+    from common.resume_mode import add_resume_mode_arguments
+    from common.trace import init_tracing
+    from common.wait_time import init_wait_time
+
     init_tracing()
     init_wait_time()
 
@@ -82,7 +94,7 @@ def init_boomer_config() -> None:
             return
 
         @environment.web_ui.app.route("/boomer-config")
-        def boomer_config() -> dict[str, float | None]:
+        def boomer_config() -> dict[str, float | int | str | None]:
             opts = environment.parsed_options
             return {_attr(f): getattr(opts, _attr(f), None) for f in _FLAGS}
 
--- a/benchmarking/locust/common/durdir_config.py
+++ b/benchmarking/locust/common/durdir_config.py
@@ -1,0 +1,43 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""DurDir benchmark runtime flags."""
+
+from locust import events
+from locust.argument_parser import LocustArgumentParser
+
+
+@events.init_command_line_parser.add_listener
+def add_durdir_arguments(parser: LocustArgumentParser) -> None:
+    group = parser.add_argument_group("DurDir Benchmark")
+    group.add_argument(
+        "--durdir-file-size-bytes",
+        type=int,
+        default=8388608,
+        help="Size of the test file written and read during the DurDir benchmark (default: 8388608 = 8 MiB)",
+    )
+    group.add_argument(
+        "--durdir-read-mode",
+        type=str,
+        default="data",
+        choices=["data", "digest"],
+        help="Read mode for DurDir serves: 'data' (default) returns and client-verifies the full payload; "
+        "'digest' returns only size+sha256 for reduced network overhead",
+    )
+    group.add_argument(
+        "--durdir-template",
+        type=str,
+        default="glutton-durdir-data",
+        help="ActorTemplate name to benchmark (default: glutton-durdir-data)",
+    )
--- a/benchmarking/locust/deploy.sh
+++ b/benchmarking/locust/deploy.sh
@@ -29,15 +29,20 @@ if [ -z "${PROJECT_ID:-}" ]; then
   exit 1
 fi
 
-MANIFEST="benchmarking/locust/manifests/locust.yaml"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+MANIFEST="${SCRIPT_DIR}/manifests/locust.yaml"
+
+# Substituted into the boomer container's --user-class argument and the master's -f.
+BENCHMARK_USER_CLASS=glutton
 
 usage() {
   echo "Usage: $0 [options]"
   echo ""
   echo "Options:"
-  echo "  --deploy   Deploy the locust workers"
-  echo "  --delete   Delete the locust workers"
-  echo "  -h|--help  Show this help message"
+  echo "  --deploy           Deploy the locust workers"
+  echo "  --delete           Delete the locust workers"
+  echo "  --user-class NAME  Locust user class, lowercase; runs tests/NAME.py (default: glutton)"
+  echo "  -h|--help          Show this help message"
 }
 
 deploy() {
@@ -46,7 +51,7 @@ deploy() {
   # benchmarking/monitoring.yaml is otherwise optional.
   echo "Ensuring benchmarking namespace exists..."
   kubectl create namespace benchmarking --dry-run=client -o yaml | kubectl apply -f -
-  echo "Deploying Locust load (PROJECT_ID=${PROJECT_ID})..."
+  echo "Deploying Locust load (PROJECT_ID=${PROJECT_ID}, user_class=${BENCHMARK_USER_CLASS})..."
   envsubst < "${MANIFEST}" | kubectl apply -f -
 }
 
@@ -65,6 +70,8 @@ while [[ "$#" -gt 0 ]]; do
   case "$1" in
     --deploy) action="deploy" ;;
     --delete) action="delete" ;;
+    --user-class) shift; BENCHMARK_USER_CLASS="$(printf '%s' "$1" | tr '[:upper:]' '[:lower:]')" ;;
+    --user-class=*) BENCHMARK_USER_CLASS="$(printf '%s' "${1#*=}" | tr '[:upper:]' '[:lower:]')" ;;
     -h|--help) usage; exit 0 ;;
     *)
       echo "Error: Unknown option: $1" >&2
@@ -74,6 +81,12 @@ while [[ "$#" -gt 0 ]]; do
   esac
   shift
 done
+
+if [[ ! -f "${SCRIPT_DIR}/tests/${BENCHMARK_USER_CLASS}.py" ]]; then
+  echo "Error: no tests/${BENCHMARK_USER_CLASS}.py; --user-class must name a test file" >&2
+  exit 1
+fi
+export BENCHMARK_USER_CLASS
 
 if [[ "${action}" == "deploy" ]]; then
   deploy
--- a/benchmarking/locust/manifests/locust.yaml
+++ b/benchmarking/locust/manifests/locust.yaml
@@ -20,7 +20,7 @@
 #     LOCUST_NO_GLUTTON_USER=1 prevents tests/glutton.py from declaring the
 #     stub User on this worker so GluttonUser spawns are owned exclusively
 #     by the boomer container.
-#   * boomer-glutton: Go re-implementation of GluttonUser. Connects to the
+#   * boomer-glutton: Go re-implementation of GluttonUser and DurdirUser. Connects to the
 #     master at localhost:5557 via the locust worker ZMQ protocol; exposes
 #     its own per-worker diagnostics at /metrics on :8001 (aggregate stats
 #     flow through the master via boomer.RecordSuccess).
@@ -52,7 +52,7 @@ spec:
         - "-m"
         - "locust"
         - "-f"
-        - "/app/tests/glutton.py"
+        - "/app/tests/${BENCHMARK_USER_CLASS}.py"
         - "--master"
         - "--master-bind-host=0.0.0.0"
         env:
@@ -134,6 +134,7 @@ spec:
         # are honored. The endpoint is served by
         # benchmarking/locust/common/boomer_config.py on the master.
         - "--master-web-port=8089"
+        - "--user-class=${BENCHMARK_USER_CLASS}"
         env:
         - name: OTEL_EXPORTER_OTLP_ENDPOINT
           value: opentelemetry-collector.gke-managed-otel.svc.cluster.local:4317
--- a/benchmarking/locust/runner.py
+++ b/benchmarking/locust/runner.py
@@ -80,10 +80,20 @@ def parse_args() -> argparse.Namespace:
     return args
 
 
+# Tests whose User class is implemented in Python. Closed set: new load
+# generators are written in boomer, so anything else runs on boomer. Remove
+# entries as these are retired; when empty, drop needs_boomer entirely.
+PYTHON_TESTS = frozenset({
+    "ate_api.py",
+    "counter_demo.py",
+    "sleep.py",
+    "usermem.py",
+    "kernelmem.py",
+})
+
+
 def needs_boomer(test_file: str) -> bool:
-    """Return True if the test file is the glutton stub; the real GluttonUser
-    implementation lives in the boomer-glutton binary."""
-    return os.path.basename(test_file) == "glutton.py"
+    return os.path.basename(test_file) not in PYTHON_TESTS
 
 
 def tee(logs: TextIO, msg: str) -> None:
@@ -208,7 +218,7 @@ def run_test(args: argparse.Namespace, csv_prefix: Path, logs: TextIO, traces: T
 
     boomer_proc = None
     if with_boomer:
-        boomer_cmd = [BOOMER_BINARY]
+        boomer_cmd = [BOOMER_BINARY, "--user-class", Path(args.file).stem]
         cfg_json = build_config_json(args.locust_extra)
         if cfg_json:
             boomer_cmd += ["--config-json", cfg_json]
--- a/benchmarking/locust/tests/durdir.py
+++ b/benchmarking/locust/tests/durdir.py
@@ -1,0 +1,46 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Stub DurdirUser declaration.
+
+The real load implementation lives in the boomer-Go worker at
+cmd/benchmarking/boomer-glutton/; this Python class is declared only so the
+master recognizes the name and attributes boomer's stats rows to it. The
+master loads this stub file (or glutton.py), selected by
+${BENCHMARK_USER_CLASS} in locust/manifests/locust.yaml. The Python worker
+container sets LOCUST_NO_DURDIR_USER=1 to skip loading this file, leaving
+boomer as the sole owner of DurdirUser load.
+"""
+
+import os
+
+if os.environ.get("LOCUST_NO_DURDIR_USER") != "1":
+    from locust import User, task
+    from common.boomer_config import init_boomer_config
+
+    # Master serves /boomer-config so the boomer-glutton workers can fetch
+    # runtime flag values (trace probability, wait times, durdir config) the
+    # operator set in the web UI form. No-op on workers without a web UI.
+    init_boomer_config()
+
+    class DurdirUser(User):
+        host = "api.ate-system.svc.cluster.local:443"
+
+        @task
+        def noop(self) -> None:
+            # Unreached under normal operation: the Python worker container
+            # does not load this file (LOCUST_NO_DURDIR_USER=1). Body is
+            # required because locust validates that every User has at least
+            # one @task method.
+            pass
--- a/cmd/benchmarking/boomer-glutton/main.go
+++ b/cmd/benchmarking/boomer-glutton/main.go
@@ -32,6 +32,7 @@ import (
 	"github.com/agent-substrate/substrate/internal/benchmarking/boomer/glutton"
 	bmetrics "github.com/agent-substrate/substrate/internal/benchmarking/boomer/metrics"
 	btrace "github.com/agent-substrate/substrate/internal/benchmarking/boomer/trace"
+	"github.com/agent-substrate/substrate/internal/benchmarking/boomer/userclass"
 	"github.com/myzhan/boomer"
 )
 
@@ -44,7 +45,7 @@ func main() {
 		configJSON    = flag.String("config-json", "", "Initial dynconfig as a JSON object (keys: trace_probability, min_wait_time, max_wait_time in seconds, durdir_file_size_bytes, resume_mode, durdir_read_mode, durdir_template). Unset fields keep their built-in defaults.")
 		masterWebPort = flag.Int("master-web-port", 0, "If non-zero, fetch dynconfig from http://{master-host}:{master-web-port}/boomer-config on each spawn message and fail fatally on error. {master-host} comes from boomer's existing --master-host flag.")
 		useTokenAuth  = flag.Bool("use-token-auth", false, "Use Kubernetes ServiceAccount token for ateapi auth instead of client certificate.")
-		userClass     = flag.String("user-class", "glutton", "Locust user class to run, lowercase; maps to tests/<value>.py.")
+		userClass     = flag.String("user-class", "glutton", fmt.Sprintf("Locust user class to run, lowercase; one of %s.", strings.Join(userclass.Names(), "|")))
 	)
 	// boomer.Run will call flag.Parse() if we haven't yet; calling here so
 	// our flag-derived values are usable before that.
@@ -99,7 +100,7 @@ func main() {
 		slog.Info("dynconfig fetch enabled", slog.String("url", configURL))
 	}
 
-	cfg := &glutton.Config{
+	cfg := &userclass.Config{
 		APIStub:    apiStub,
 		HTTPClient: httpClient,
 		RouterURL:  *routerURL,
@@ -107,23 +108,17 @@ func main() {
 		Dyn:        dyn,
 	}
 
-	var (
-		taskFn     func()
-		shutdownFn func(context.Context)
-	)
-
-	switch class {
-	case "glutton":
-		taskFn, shutdownFn = glutton.Register(cfg)
-	case "durdir":
-		taskFn, shutdownFn = glutton.RegisterDurDir(cfg)
-	default:
-		slog.Error("fatal: unknown --user-class value", slog.String("user_class", *userClass))
+	entry, ok := userclass.Lookup(class)
+	if !ok {
+		slog.Error("fatal: unknown --user-class value",
+			slog.String("user_class", *userClass),
+			slog.String("known", strings.Join(userclass.Names(), ",")))
 		os.Exit(1)
 	}
+	taskFn, shutdownFn := entry.Init(cfg)
 
 	slog.Info("registered boomer task",
-		slog.String("user_class", userClassName(class)),
+		slog.String("user_class", entry.UserClass),
 		slog.String("user_class_flag", class),
 	)
 
@@ -138,7 +133,7 @@ func main() {
 	// Blocks until SIGINT/SIGTERM or master quit. Boomer registers its own
 	// signal handlers; we do cleanup after it returns.
 	boomer.Run(&boomer.Task{
-		Name:   userClassName(class),
+		Name:   entry.UserClass,
 		Weight: 1,
 		Fn:     taskFn,
 	})
@@ -147,13 +142,4 @@ func main() {
 	shutdownCtx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
 	defer cancel()
 	shutdownFn(shutdownCtx)
-}
-
-// userClassName maps a --user-class value to its Locust class name, per the
-// tests/<value>.py convention: glutton -> GluttonUser.
-func userClassName(v string) string {
-	if v == "" {
-		return ""
-	}
-	return strings.ToUpper(v[:1]) + v[1:] + "User"
 }
--- a/cmd/benchmarking/boomer-glutton/main.go
+++ b/cmd/benchmarking/boomer-glutton/main.go
@@ -25,6 +25,7 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/agent-substrate/substrate/internal/benchmarking/boomer/dynconfig"
@@ -40,13 +41,16 @@ func main() {
 		routerURL     = flag.String("router-url", "http://atenet-router.ate-system.svc.cluster.local", "atenet HTTP router base URL (no trailing slash).")
 		atespace      = flag.String("atespace", "benchmark", "Atespace every actor this worker creates lives in. Ensured (CreateAtespace, AlreadyExists is ok) at startup.")
 		promAddr      = flag.String("prometheus-addr", ":8001", "Address for the Prometheus /metrics endpoint.")
-		configJSON    = flag.String("config-json", "", "Initial dynconfig as a JSON object (keys: trace_probability, min_wait_time, max_wait_time in seconds). Unset fields keep their built-in defaults.")
+		configJSON    = flag.String("config-json", "", "Initial dynconfig as a JSON object (keys: trace_probability, min_wait_time, max_wait_time in seconds, durdir_file_size_bytes, resume_mode, durdir_read_mode, durdir_template). Unset fields keep their built-in defaults.")
 		masterWebPort = flag.Int("master-web-port", 0, "If non-zero, fetch dynconfig from http://{master-host}:{master-web-port}/boomer-config on each spawn message and fail fatally on error. {master-host} comes from boomer's existing --master-host flag.")
 		useTokenAuth  = flag.Bool("use-token-auth", false, "Use Kubernetes ServiceAccount token for ateapi auth instead of client certificate.")
+		userClass     = flag.String("user-class", "glutton", "Locust user class to run, lowercase; maps to tests/<value>.py.")
 	)
 	// boomer.Run will call flag.Parse() if we haven't yet; calling here so
 	// our flag-derived values are usable before that.
 	flag.Parse()
+
+	class := strings.ToLower(*userClass)
 
 	slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stdout, nil)))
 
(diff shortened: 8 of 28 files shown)


benchmarking/locust/common/durdir_config.py
```diff
@@ -0,0 +1,66 @@
...
+    def on_init_parser(parser: LocustArgumentParser) -> None:
+        parser.add_argument(
+            "--durdir-file-size-bytes",
+            type=int,
+            default=8388608,
+            env_var="LOCUST_DURDIR_FILE_SIZE_BYTES",
+            help="Size in bytes written to /var/lib/glutton/bench-data (default 8 MiB)",
+            include_in_web_ui=True,
+        )
+        parser.add_argument(
```
maxsmythe: If this applies to any workload, it should probably live in its own Python file

benchmarking/locust/manifests/locust.yaml
```diff
@@ -24,6 +24,15 @@
...
 #     flow through the master via boomer.RecordSuccess).
+#
+# ${BENCHMARK_WORKLOAD} is substituted by benchmarking/locust/deploy.sh
+# (--workload, default "glutton"). It drives the master's -f and boomer's
+# --workload from one value on purpose: the master's test file decides which
+# User class the stack offers, --workload decides which task boomer runs, and
+# if those disagree the stats rows describe traffic that was never sent.
+# The value must name a stub in locust/tests/, a --workload case in
+# cmd/benchmarking/boomer-glutton/main.go, and an entry in runner.py's
+# BOOMER_WORKLOADS dict.
```
maxsmythe: TODO: look to reduce # of sources of truth

benchmarking/locust/runner.py
```diff
@@ -80,10 +85,15 @@ def parse_args() -> argparse.Namespace:
     return args
 
 
+def boomer_workload(test_file: str) -> str | None:
```
maxsmythe: It would be better to test for Python workloads... I suspect we will no longer be adding those, so it will be a static list. A cleanup work item will be to remove those tests.

This will also remove one place workload lists need to be kept in alignment

benchmarking/README.md
```diff
@@ -56,6 +57,55 @@ You can also configure things like the number of users, how quickly those users
...
+and state restoration latency when durable actor directories (`/var/lib/glutton`) are attached.
+
+#### Running DurDir
+
+* **Headless / Nightly Automation**: Run via `runner.py` with `-f tests/durdir.py`:
+  ```bash
+  python3 runner.py -f tests/durdir.py -t 1m -u 1 --name durdir_run --dest /tmp/bench \
+    --durdir-template glutton-durdir-data --resume-mode explicit --durdir-file-size-bytes 8388608
+  ```
+* **Interactive Web UI**: The stack runs one workload per deployment, chosen at deploy time:
```
maxsmythe: I don't think "workload" is correct... that is the name given to the ActorTemplate and workers.

One "test type"? "Virtual user type"?

benchmarking/README.md
```diff
@@ -56,6 +57,55 @@ You can also configure things like the number of users, how quickly those users
...
+* `--durdir-read-mode`: Verification read mode:
+  * `data` (default): Server returns the full payload bytes and client verifies size, server digest, and client-computed digest.
+  * `digest`: Server streams the file through SHA-256 at constant memory and returns only size + digest (no payload over the wire). The actor still reads the whole file within Envoy's route timeout (--route-timeout, 10s by default), so very large files remain bounded by disk throughput rather than by transfer size.
+* `--durdir-template`: ActorTemplate name:
+  * `glutton-durdir-data` (default): Attaches a durable data directory under `/var/lib/glutton` without memory snapshot restore.
+  * `glutton-durdir-full`: Attaches a durable data directory and performs a full memory snapshot restore.
+
+#### File-Size Sweeps
```
maxsmythe: I don't think we need to document that we can run multiple tests by adding multiple tests. Also, I think setting the read mode to digest would be orthogonal to "sweep".

I'd remove this section.

cmd/benchmarking/boomer-glutton/main.go
```diff
@@ -40,9 +40,10 @@ func main() {
...
 		promAddr      = flag.String("prometheus-addr", ":8001", "Address for the Prometheus /metrics endpoint.")
-		configJSON    = flag.String("config-json", "", "Initial dynconfig as a JSON object (keys: trace_probability, min_wait_time, max_wait_time in seconds). Unset fields keep their built-in defaults.")
+		configJSON    = flag.String("config-json", "", "Initial dynconfig as a JSON object (keys: trace_probability, min_wait_time, max_wait_time in seconds, durdir_file_size_bytes, resume_mode, durdir_read_mode, durdir_template). Unset fields keep their built-in defaults.")
 		masterWebPort = flag.Int("master-web-port", 0, "If non-zero, fetch dynconfig from http://{master-host}:{master-web-port}/boomer-config on each spawn message and fail fatally on error. {master-host} comes from boomer's existing --master-host flag.")
 		useTokenAuth  = flag.Bool("use-token-auth", false, "Use Kubernetes ServiceAccount token for ateapi auth instead of client certificate.")
+		workload      = flag.String("workload", "glutton", "Workload to run: 'glutton' (Ping) or 'durdir' (DurableDir benchmark).")
```
maxsmythe: We should rename workload. user-class would be Locust-y (duplicate of above comment)

cmd/benchmarking/glutton/main.go
```diff
@@ -176,11 +178,28 @@ func splitGRPC(grpcSrv, rest http.Handler) http.Handler {
...
+func newMux(svc *gluttonService) *http.ServeMux {
+	mux := readyzMux()
+	mux.HandleFunc("/ping", protoRoute("Ping", svc.Ping))
+	mux.HandleFunc("/writedisk", protoRoute("WriteDisk", svc.WriteDisk))
+	mux.HandleFunc("/readdisk", protoRoute("ReadDisk", svc.ReadDisk))
+	return mux
+}
+
+// protoRoute creates an HTTP handler for a gRPC-style protobuf RPC.
+// It checks POST method (returning 405), unmarshals the request protobuf (400),
```
maxsmythe: Comments should not describe the literal logic of the code, but the intent of the function or the structural purpose, etc. -- they are a TL;DR for someone reading the code base or a "heads up" in case there is something non-obvious.

TL;DR - shrink the comment :)

cmd/benchmarking/glutton/main.go
```diff
@@ -176,11 +178,28 @@ func splitGRPC(grpcSrv, rest http.Handler) http.Handler {
 	})
 }
 
-// httpPingHandler accepts a POST whose body is proto.Marshal(PingRequest) and
-// returns proto.Marshal(PingResponse) (same Ping handler the gRPC server
-// uses, so the per-call stats stay comparable across protocols).
-func httpPingHandler(svc *gluttonService) http.HandlerFunc {
+// newMux builds the HTTP-mode route table on top of the readiness probe.
+// Shared with the tests so a route renamed here cannot silently pass
```
maxsmythe: "route renamed here"?

cmd/benchmarking/glutton/main.go
```diff
@@ -176,11 +178,28 @@ func splitGRPC(grpcSrv, rest http.Handler) http.Handler {
...
+	mux.HandleFunc("/readdisk", protoRoute("ReadDisk", svc.ReadDisk))
+	return mux
+}
+
+// protoRoute creates an HTTP handler for a gRPC-style protobuf RPC.
+// It checks POST method (returning 405), unmarshals the request protobuf (400),
+// starts an OTel span with spanName, invokes handler, maps gRPC status codes to
+// HTTP status codes, sets Content-Type to application/x-protobuf, and writes
+// marshaled response bytes.
+func protoRoute[Req any, Resp proto.Message, PtrReq interface {
```
maxsmythe, line 193: generics? Fancy. Like the code re-use.

cmd/benchmarking/glutton/main_test.go
```diff
@@ -111,3 +120,436 @@ func TestSplitGRPCRoutesOnContentType(t *testing.T) {
...
+	readResp, err := svc.ReadDisk(ctx, &glutton.ReadDiskRequest{Key: key})
+	if err != nil {
+		t.Fatalf("ReadDisk failed: %v", err)
+	}
+
+	if readResp.GetSize() != 4096 {
+		t.Errorf("expected ReadDisk size 4096, got %d", readResp.GetSize())
+	}
+	if bytes.Equal(readResp.GetSha256(), overwriteResp.GetSha256()) {
+		t.Errorf("expected ReadDisk whole-file digest to differ from WriteDisk prefix digest under OVERWRITE mode")
```
maxsmythe: I don't think this should be true. If we are only returning the prefix SHA, then that SHA has no further use... there is no ability to re-calculate it to ensure persistence is correct.

May be at some point we need to be fancier with SHA calculation (let user specify how much data to use when calculating SHA), but that can be later.

internal/benchmarking/boomer/dynconfig/dynconfig.go
```diff
@@ -122,6 +139,28 @@ func (p payload) merge(current Config) Config {
...
 	}
+	if p.DurDirFileSize != nil {
+		out.DurDirFileSize = int64(*p.DurDirFileSize)
+	}
+	if p.ResumeMode != nil {
+		mode := *p.ResumeMode
+		if mode == ResumeModeExplicit || mode == ResumeModeImplicit {
+			out.ResumeMode = mode
+		} else {
+			slog.Warn("dynconfig ignoring invalid resume_mode", slog.String("mode", mode))
```
maxsmythe: We might want to panic or force errors for every request on a bad config... something that fails loudly to help identify misconfigurations.

internal/benchmarking/boomer/glutton/actor.go
```diff
@@ -79,100 +61,21 @@ type Config struct {
...
 	cfg          *Config
 	actorName    string
 	hostHeader   string
+	userClass    string
+	templateName string
 	firstResume  bool
-	actorRunning bool
+	// bootOnFirstResume asks ResumeActor to skip the template's golden
+	// snapshot and boot the workload from the ActorTemplate spec instead.
+	// Only meaningful on the first resume: after that the actor has its own
```
maxsmythe: It could be meaningful with other resumes. Not sure I like the "boot on first resume" paradigm, but can be addressed later.

internal/benchmarking/boomer/glutton/actor.go
```diff
@@ -79,100 +61,21 @@ type Config struct {
...
+	// snapshot to restore from and the flag is ignored. Ping measures that
+	// cold-boot path deliberately; DurDir does not, because a cold boot
```
maxsmythe: I am leery of coupling so many internal implementation to an external docstring. Also, not strictly correct for dur-dir... one of the actortemplates resumes from data (cold boot) after snapshot taken

internal/benchmarking/boomer/glutton/actor.go
```diff
@@ -79,100 +61,21 @@ type Config struct {
...
-	}
-	jitter := cfg.MaxWait - cfg.MinWait
-	return cfg.MinWait + time.Duration(rand.Float64()*float64(jitter))
-}
-
 type gluttonUser struct {
 	cfg          *Config
 	actorName    string
 	hostHeader   string
+	userClass    string
```
maxsmythe: I like this idea, though we will probably need some design/refactoring to make this truly generic across all users. We'll tackle that more on a later test.

internal/benchmarking/boomer/glutton/actor.go
```diff
@@ -208,22 +111,49 @@ func (u *gluttonUser) create(ctx context.Context) error {
...
+				ActorTemplateName:      u.templateName,
 			},
 		}, grpc.Trailer(tr))
 		return err
 	})
 }
 
+// resume brings the actor up before the next request.
+//
+// The first resume is always explicit, whatever the mode. It is the only
```
maxsmythe: This is far too much context for a comment. Documents the code and likely to rot quickly.

internal/benchmarking/boomer/glutton/durdir.go
```diff
@@ -0,0 +1,240 @@
...
+		newUser: newDurDirUser,
+	}
+	return rt.iterate, rt.shutdown
+}
+
+type durDirUser struct {
+	user           *gluttonUser
+	expectedDigest string
+	fileSize       int64
+	readMode       gluttonpb.ReadMode
```
maxsmythe: Things that are stored as dynamic config should not be persisted to the class, they should be read from the dynamic config, so behavior can change dynamically (might need to have some read atomicity to avoid breaking behavior due to time aliasing)

internal/benchmarking/boomer/glutton/durdir.go
```diff
@@ -0,0 +1,240 @@
...
+
+	readMode := gluttonpb.ReadMode_READ_MODE_DATA
+	if dyn.DurDirReadMode == "digest" {
+		readMode = gluttonpb.ReadMode_READ_MODE_DIGEST_ONLY
+	}
+
+	slog.Info("configured durdir file size", slog.Int64("bytes", fileSize), slog.String("template", tmpl))
+
+	if readMode == gluttonpb.ReadMode_READ_MODE_DATA && fileSize > 64*1024*1024 {
+		slog.Warn("durdir_file_size_bytes > 64MB with durdir_read_mode=data may hit router or Envoy timeouts; set durdir_read_mode=digest",
```
maxsmythe: No need to warn about this... if we are hitting timeouts that would be the point of the benchmark, to find out.

benchmarking/locust/manifests/locust.yaml
```diff
@@ -24,6 +24,15 @@
 #     master at localhost:5557 via the locust worker ZMQ protocol; exposes
 #     its own per-worker diagnostics at /metrics on :8001 (aggregate stats
 #     flow through the master via boomer.RecordSuccess).
+#
+# ${BENCHMARK_WORKLOAD} is substituted by benchmarking/locust/deploy.sh
```
maxsmythe: Comment describes too many internal details

maxsmythe (changes requested):
Thank you for this.

Left some comments.

Also, we should avoid sending 4k line PRs where possible -- they take multiple hours to review closely.

Can we avoid refactoring anything under `internal/benchmarking/boomer` for now? I don't think the abstractions are quite right and the refactoring is a large portion of the delta. We can implement ~3-4 tests and then figure out what the best refactor is.

benchmarking/README.md
maxsmythe: UserClass?

BenTheElder:
let's squash this on merge (and more PRs, generally, unless there's really useful commit history)

benchmarking/locust/common/durdir_config.py
sairajp-rewind: Split out. --resume-mode now lives in common/resume_mode.py

benchmarking/locust/manifests/locust.yaml
sairajp-rewind: Addressed

benchmarking/locust/runner.py
sairajp-rewind: Went with a single static map from stub filename to user class:

BOOMER_USER_CLASSES = {"glutton.py": "GluttonUser", "durdir.py": "DurDirUser"}

needs_boomer is membership in it and the same entry supplies --user-class, so the filename and the class name stay in one place rather than two.
Kept the check on the boomer side rather than the Python side because runner.py needs the class name string either way

benchmarking/README.md
sairajp-rewind: Fixed.

benchmarking/README.md
sairajp-rewind: Removed.

cmd/benchmarking/boomer-glutton/main.go
sairajp-rewind: Renamed to --user-class across boomer, runner.py, deploy.sh, the manifest and the README. Values are the Locust class names, and boomer rejects anything that is not GluttonUser or DurDirUser at startup.

cmd/benchmarking/glutton/main.go
sairajp-rewind: Addressed

cmd/benchmarking/glutton/main.go
sairajp-rewind: kept as is

cmd/benchmarking/glutton/main_test.go
sairajp-rewind: WriteDisk now seeks back to 0 and hashes the whole file, and returns the real on-disk size rather than the requested size, so it is directly comparable with ReadDisk. TestOverwriteModeDigestCoversWrittenPrefixOnly is deleted and replaced by TestWriteDiskOverwriteDigestMatchesReadDisk, which pins the new behaviour. Configurable hash length parked for later.

internal/benchmarking/boomer/dynconfig/dynconfig.go
sairajp-rewind: Added Config.Validate(), called from both Parse and Fetch before the new value is stored, so a bad value never reaches a running task.

internal/benchmarking/boomer/glutton/actor.go
sairajp-rewind: The paradigm is gone rather than deferred. Reverting the extraction means lifecycle.go is byte-identical to main, so ping keeps its existing first-resume boot behaviour and there is no shared knob for a future user class to inherit. DurDir never sets Boot, so it always restores from whatever snapshot exists, which is what it is trying to measure.

internal/benchmarking/boomer/glutton/actor.go
sairajp-rewind: Addressed

internal/benchmarking/boomer/glutton/actor.go
sairajp-rewind: Addressed

internal/benchmarking/boomer/glutton/durdir.go
sairajp-rewind: Fixed, including the aliasing half.

internal/benchmarking/boomer/glutton/durdir.go
sairajp-rewind: Addressed

benchmarking/locust/manifests/locust.yaml
sairajp-rewind: Addressed

benchmarking/locust/common/boomer_config.py
```diff
@@ -34,17 +36,19 @@
...
-from common.wait_time import init_wait_time
-
 logger = logging.getLogger(__name__)
 
-# Boomer-tunable flags. CLI form ("--foo-bar") is converted to the
-# attribute / JSON-key form ("foo_bar") by _attr().
-_FLAGS = ("--trace-probability", "--min-wait-time", "--max-wait-time")
+# Boomer-tunable flags and their types. CLI form ("--foo-bar") is converted
+# to the attribute / JSON-key form ("foo_bar") by _attr().
+_FLAGS = {
```
maxsmythe, line 43: Not a thing for this review, but probably the correct way to do this is to add each flag to a unified registry upon import.

Again, not a thing for this PR but a cleanup for later to help remove duplicate sources of truth.

benchmarking/locust/deploy.sh
```diff
@@ -31,13 +31,17 @@ fi
...
 usage() {
   echo "Usage: $0 [options]"
   echo ""
   echo "Options:"
-  echo "  --deploy   Deploy the locust workers"
-  echo "  --delete   Delete the locust workers"
-  echo "  -h|--help  Show this help message"
+  echo "  --deploy           Deploy the locust workers"
+  echo "  --delete           Delete the locust workers"
+  echo "  --user-class NAME  User class to deploy: GluttonUser | DurDirUser (default: GluttonUser)"
```
maxsmythe: let's not use case-sensitive arguments (just do lowercase). Also, we can name the class after the file.

Let's assume a convention of /tests/{user_class}.py (e.g. GluttonUser becomes glutton).

That also removes the need to do mapping and argument correctness check is just looking for the existence of the correct file.

benchmarking/locust/runner.py
```diff
@@ -80,10 +80,20 @@ def parse_args() -> argparse.Namespace:
     return args
 
 
+# Maps locust test filename to boomer's --user-class flag.
+BOOMER_USER_CLASSES: dict[str, str] = {
```
maxsmythe: Adopting the filename == user + .py convention also removes the need for this code.

benchmarking/locust/runner.py
```diff
@@ -80,10 +80,20 @@ def parse_args() -> argparse.Namespace:
...
+    "glutton.py": "GluttonUser",
+    "durdir.py": "DurDirUser",
+}
+
+
 def needs_boomer(test_file: str) -> bool:
-    """Return True if the test file is the glutton stub; the real GluttonUser
-    implementation lives in the boomer-glutton binary."""
-    return os.path.basename(test_file) == "glutton.py"
+    """Return True if the test file's User class runs on boomer."""
```
maxsmythe: Per previous review, let's flip this to test whether the user class needs Python. membership of that set should be static, reducing brittleness.

benchmarking/README.md
```diff
@@ -56,6 +56,48 @@ You can also configure things like the number of users, how quickly those users
 are spawned, the frequency with which requests are made and whether or not tracing is
 enabled.
 
+### DurDir Benchmark
+
+The DurDir benchmark evaluates actor suspend/resume performance, disk persistence overhead,
+and state restoration latency when durable actor directories (`/var/lib/glutton`) are attached.
```
maxsmythe: nit: remove reference to "var/lib/glutton" -- too much implementation detail.

benchmarking/README.md
```diff
@@ -56,6 +56,48 @@ You can also configure things like the number of users, how quickly those users
...
+
+#### Running DurDir
```
maxsmythe: Including this section as something dur-dir specific feels wrong.

If we want a "how to run a test" section, that should live as its own thing because it applies to all tests.

Documenting DurDir configuration knobs makes sense.

benchmarking/README.md
```diff
@@ -56,6 +56,48 @@ You can also configure things like the number of users, how quickly those users
...
+* `--durdir-read-mode`: Verification read mode:
+  * `data` (default): Server returns full payload bytes for client-side SHA-256 verification.
+  * `digest`: Server hashes the file and returns size and digest, reducing network transfer.
+* `--durdir-template`: ActorTemplate name:
+  * `glutton-durdir-data` (default): Attaches a durable data directory under `/var/lib/glutton` without memory snapshot restore.
+  * `glutton-durdir-full`: Attaches a durable data directory and performs a full memory snapshot restore.
+
+#### DurDir Reported Metrics
+
+* `DurDirWrite`: Initial truncate-write creating `/var/lib/glutton/bench-data`.
```
maxsmythe: removing specific filename and just saying "data file" or similar actually improves readability -- reader does not need to infer what `bench-data`'s purpose is.

cmd/benchmarking/boomer-glutton/main.go
```diff
@@ -40,9 +40,10 @@ func main() {
 		routerURL     = flag.String("router-url", "http://atenet-router.ate-system.svc.cluster.local", "atenet HTTP router base URL (no trailing slash).")
 		atespace      = flag.String("atespace", "benchmark", "Atespace every actor this worker creates lives in. Ensured (CreateAtespace, AlreadyExists is ok) at startup.")
 		promAddr      = flag.String("prometheus-addr", ":8001", "Address for the Prometheus /metrics endpoint.")
-		configJSON    = flag.String("config-json", "", "Initial dynconfig as a JSON object (keys: trace_probability, min_wait_time, max_wait_time in seconds). Unset fields keep their built-in defaults.")
+		configJSON    = flag.String("config-json", "", "Initial dynconfig as a JSON object (keys: trace_probability, min_wait_time, max_wait_time in seconds, durdir_file_size_bytes, resume_mode, durdir_read_mode, durdir_template). Unset fields keep their built-in defaults.")
```
maxsmythe, line 45: Side note:

this doc string is going to get unweildy and the assumption that all configs share the same attributes seems slightly broken. Eventually this will need to get overhauled, but not in this PR.

cmd/benchmarking/boomer-glutton/main.go
```diff
@@ -40,9 +40,10 @@ func main() {
...
 		masterWebPort = flag.Int("master-web-port", 0, "If non-zero, fetch dynconfig from http://{master-host}:{master-web-port}/boomer-config on each spawn message and fail fatally on error. {master-host} comes from boomer's existing --master-host flag.")
 		useTokenAuth  = flag.Bool("use-token-auth", false, "Use Kubernetes ServiceAccount token for ateapi auth instead of client certificate.")
+		userClass     = flag.String("user-class", "GluttonUser", "Locust User class to run: 'GluttonUser' or 'DurDirUser'.")
```
maxsmythe: Same comment about case-sensitive flag values and the filename <-> user class name convention.

benchmarking/locust/tests/durdir.py
```diff
@@ -0,0 +1,46 @@
...
+master recognizes the name and attributes boomer's stats rows to it. The
+master loads this stub file (or glutton.py), selected by
+${BENCHMARK_USER_CLASS} in locust/manifests/locust.yaml. The Python worker
+container sets LOCUST_NO_DURDIR_USER=1 to skip loading this file, leaving
+boomer as the sole owner of DurDirUser load.
+"""
+
+import os
+
+if os.environ.get("LOCUST_NO_DURDIR_USER") != "1":
```
maxsmythe, line 28: I have a feeling we will get to a point where these files are auto-genned

cmd/benchmarking/glutton/main.go
```diff
@@ -137,15 +137,10 @@ func main() {
...
-		// HTTP/1.1 mode: a single /ping route that consumes
-		// proto.Marshal(PingRequest) and returns proto.Marshal(PingResponse).
-		// Only Ping is exposed in HTTP mode; the other RPCs remain gRPC-only
-		// (re-exposable as additional routes if/when needed).
-		mux.HandleFunc("/ping", httpPingHandler(svc))
 		// otelhttp at the mux level + per-handler span follows
 		// docs/dev/best-practices/tracing.md: extract incoming context,
 		// then name the span after the operation in each handler.
-		handler = otelhttp.NewHandler(mux, "/")
+		handler = otelhttp.NewHandler(newMux(svc), "glutton-http")
```
maxsmythe: Why are we renaming "/" "glutton-http"?

cmd/benchmarking/glutton/main.go
```diff
@@ -409,8 +450,60 @@ func (s *gluttonService) WriteDisk(ctx context.Context, req *glutton.WriteDiskRe
 		return nil, status.Errorf(codes.Internal, "write %s: %v", path, err)
 	}
 
+	if _, err := f.Seek(0, io.SeekStart); err != nil {
+		return nil, status.Errorf(codes.Internal, "seek %s: %v", path, err)
+	}
+	h := sha256.New()
```
maxsmythe: I'm not sure we want to always pair reads with writes.

It is necessary if we want the full SHA for a file that was only partially overwritten, otherwise, can we calculate SHA as data is written to disk?

that way people can benchmark pure writes.

internal/benchmarking/boomer/glutton/fixture_test.go
```diff
@@ -0,0 +1,261 @@
...
+)
+
+type fakeControlClient struct {
+	ateapipb.ControlClient
+	mu          sync.Mutex
+	calls       []string
+	resumeBoots []bool
+}
+
+func (f *fakeControlClient) CreateAtespace(ctx context.Context, in *ateapipb.CreateAtespaceRequest, opts ...grpc.CallOption) (*ateapipb.Atespace, error) {
```
maxsmythe, line 37: Is there no other place where API server has been faked, where we can use that? Can we centralize this so only one fake needs to be created?

internal/benchmarking/boomer/glutton/fixture_test.go
```diff
@@ -0,0 +1,261 @@
...
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return append([]bool(nil), f.resumeBoots...)
+}
+
+// diskServer models an actor holding a file for HTTP tests.
+// The data slice is the source of truth: both routes report len(data) and
+// sha256(data), and /readdisk serves data as payload.
+// Each override field makes the actor lie about exactly one property.
+type diskServer struct {
```
maxsmythe: We should just create a mock glutton server as part of glutton, vs. relying on each test to make its own mock.

bonus... can probably do this by just creating shims in actual server code for anything that has a side effect (like writing to disk), then use dependency injection to differentiate between a real and fake server.

internal/proto/glutton/glutton.proto
```diff
@@ -52,6 +55,22 @@ enum WriteMode {
   WRITE_MODE_OVERWRITE = 1;
 }
 
+// ReadMode selects how much of the file ReadDisk sends back. Digest-only
```
maxsmythe: Shorten this doc string, assume a user knows why they would want functionality, they just need to know what a thing does.

maxsmythe (changes requested):
Thanks! Couple of comments

benchmarking/locust/deploy.sh
sairajp-rewind: Adopted. `--user-class` is lowercase and case-insensitive, the file is
`tests/<value>.py`, and the class is `<Value>User` — so `glutton` ->
`GluttonUser`, `durdir` -> `DurdirUser`. Renamed the Python class to match the
convention rather than renaming the file.

`deploy.sh` validation is now just a file-existence check, `BENCHMARK_USER_FILE`
is gone from the manifest, and `runner.py`'s filename->class map is gone
entirely — the flag value is the test file's stem.

benchmarking/locust/runner.py
sairajp-rewind: addressed

benchmarking/locust/runner.py
sairajp-rewind: Flipped. `PYTHON_TESTS` is the static set and anything not in it routes to
boomer.

benchmarking/README.md
sairajp-rewind: Done

benchmarking/README.md
sairajp-rewind: Moved. There's a generic `### Headless` section under "Running Tests" now, plus
a note in the web UI section about `deploy.sh --user-class`.
`runner.py` wasn't documented anywhere else in this README, so it needed a home
regardless. The DurDir section is knobs and metrics only.

benchmarking/README.md
sairajp-rewind: Done

benchmarking/locust/tests/durdir.py
sairajp-rewind: Agreed

cmd/benchmarking/glutton/main.go
sairajp-rewind: Not intentional, leftover from when the mux had a single route. Reverted to
`"/"`, which is what `docs/dev/best-practices/tracing.md` prescribes: the root
mux gets `"/"` and the operation name comes from the per-handler span. That's
already what `protoRoute` does, so `Ping`/`WriteDisk`/`ReadDisk` show up as
children.

cmd/benchmarking/glutton/main.go
sairajp-rewind: Good Catch. Fixed by branching on the write mode. Under `TRUNCATE` the file is exactly N
bytes by construction, so the digest streams through an `io.MultiWriter` during
the write and there is no read-back at all. `OVERWRITE` is the only mode that
can leave bytes past N from a larger earlier write, so that branch keeps the
re-hash. Both existing tests pin the two paths, unchanged.

internal/benchmarking/boomer/glutton/fixture_test.go
sairajp-rewind: Went looking — five packages already do this, and every one is test-local, so
none of them is importable:

- `cmd/atecontroller/internal/controllers/actortemplate_controller_test.go:98`
- `cmd/atenet/internal/router/health_test.go:42`
- `cmd/atenet/internal/router/ingress/ingress_test.go:40`
- `cmd/atenet/internal/router/ingress/resumer_test.go:33`
- `cmd/atenet/internal/router/egress/egress_test.go:198`

(`internal/testenv` is envtest running a real Kubernetes apiserver, so it's a
different thing.)

So the duplication is real, but centralizing means a new non-test package plus
migrating five call sites across atecontroller and atenet. Happy to take it as a
follow-up.

internal/benchmarking/boomer/glutton/fixture_test.go
sairajp-rewind: `main.go:18` is `package main`, so the service isn't importable, a shared mock
means moving those 698 lines into a real package first.

The in-memory store would also have to model on-disk residue (`main.go:457`, the
`OVERWRITE` branch) or it goes green while the server is wrong, the bug from
the thread above. glutton's own tests use `t.TempDir()` for that reason.

Narrower point: the override fields at `fixture_test.go:99-106` exist to make the
actor lie so the client's verification path can be proven to catch it, so some
lying variant stays here regardless.

Follow-up alongside the apiserver fake, one PR or two?

internal/proto/glutton/glutton.proto
sairajp-rewind: Done

cmd/benchmarking/boomer-glutton/main.go
sairajp-rewind: Addressed

benchmarking/README.md
```diff
@@ -56,6 +56,52 @@ You can also configure things like the number of users, how quickly those users
...
+./benchmarking/locust/deploy.sh --deploy --user-class durdir
+```
+
+### Headless
+
+`runner.py` runs a test without the web UI, writing CSVs, logs and traces to
+`--dest`. This is what the nightly automation uses.
+
+```bash
+python3 runner.py -f tests/<name>.py -t 1m -u 1 --name <run-name> --dest /tmp/bench
```
maxsmythe: name => user class

cmd/benchmarking/boomer-glutton/main.go
```diff
@@ -102,7 +106,26 @@ func main() {
...
 	}
-	taskFn, shutdownFn := glutton.Register(cfg)
+
+	var (
+		taskFn     func()
+		shutdownFn func(context.Context)
+	)
+
+	switch class {
+	case "glutton":
```
maxsmythe: General pattern for things like this (unknown set of extensible dependencies) is to declare the linkage next to the code that implements it and add it to a registry. See the type code for KRM objects:

https://github.com/agent-substrate/substrate/blob/8541d6d9571c9ee120f69338821305216fec3307/pkg/api/v1alpha1/actortemplate_types.go#L433-L435

in our case, something like:

registry.go
```golang

type Entry struct {
   name string
   locustFile string
   userClass string
   initFunc func(Config) (func(), func())
}

var reg map[string]entry

func Add(entry Entry) {
   reg[name] := entry{locustFile: locustFile, userClass: userClass, initFunc: initFunc}
}

func Init(name string, cfg Config) (func(), Func()) {
   return reg[name].initFunc(cfg)
}

// etc...
```

userclass.go
```golang
import registry // if not in same package (which it normally wouldn't be)

entry := registry.Entry{
    name: "something",
   locustFile: "something.py",
   userClass: "SomethingUser",
   initFunc: initFunc,
}

registry.Add(entry)
```

etc.

Registry class should live inside internal/benchmarking directory tree.

This will make it so newer entries do not need to update main.go

I realize that this will require a (hopefully minor) update to the glutton user class file.

Also, the term "register" is now overloaded... original usage of the term should probably be updated to "initBoomer" etc. anyway... "register" is a misnomer.

cmd/benchmarking/glutton/main.go
```diff
@@ -122,10 +125,7 @@ func main() {
 	// ateom blocks RestoreWorkload until /readyz returns 200, so ResumeActor
 	// cannot report success before this listener is reachable. The probe is
 	// an HTTP GET, so both modes must serve it.
-	mux := http.NewServeMux()
-	mux.HandleFunc("/readyz", func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(http.StatusOK)
-	})
+	mux := readyzMux()
```
maxsmythe: this mux is only used in the gRPC case and should only be defined there.

cmd/benchmarking/glutton/main.go
```diff
@@ -405,12 +446,69 @@ func (s *gluttonService) WriteDisk(ctx context.Context, req *glutton.WriteDiskRe
...
+	h := sha256.New()
+	if err := streamRandomBytes(io.MultiWriter(f, h), int64(req.GetSize())); err != nil {
 		return nil, status.Errorf(codes.Internal, "write %s: %v", path, err)
 	}
+	size := int64(req.GetSize())
+
+	// OVERWRITE leaves any bytes past size from a larger earlier write, so the
+	// streamed digest covers only a prefix. Re-hash from disk to get the real one.
+	if req.GetWriteMode() == glutton.WriteMode_WRITE_MODE_OVERWRITE {
+		if _, err := f.Seek(0, io.SeekStart); err != nil {
```
maxsmythe: Is it not possible to continue reading from the pre-existing cursor location until the end of the file, continuing to compute the SHA?

We should avoid double-reading. I believe I had suggested this in a previous comment.

internal/benchmarking/boomer/glutton/durdir.go
```diff
@@ -0,0 +1,466 @@
...
+	// 5. Overwrite file with fresh random bytes
+	if err := u.writeDisk(ctx, "DurDirOverwrite", fileSize, gluttonpb.WriteMode_WRITE_MODE_TRUNCATE); err != nil {
+		return
+	}
+}
+
+func (u *durDirUser) bootstrap(ctx context.Context, dynCfg dynconfig.Config) error {
+	fileSize, readMode := u.params(dynCfg)
+
+	// First resume: always explicit to warm up actor from golden snapshot
```
maxsmythe: I'm not sure explicit resume is necessary on first boot as long as actor exists?

maxsmythe (commented):
Getting close, a few more comments to sort out, then the tests are left for reviewing.

cmd/benchmarking/boomer-glutton/main.go
sairajp-rewind: Done, as internal/benchmarking/boomer/userclass/. Entry{Name, LocustFile, UserClass, Init}, with Add/Lookup/Names; Add panics on a duplicate name. 
Both classes register from func init() next to their implementation - lifecycle.go:62 and durdir.go:62 - matching the SchemeBuilder.Register pattern you linked. main.go no longer has a per-class branch; adding a class touches only its own file.
Also renamed the overloaded term: RegisterPing is now initPing.

cmd/benchmarking/glutton/main.go
sairajp-rewind: Split into two constructors, each called at its use site. readyzMux() is just the probe, and gRPC mode passes it to splitGRPC as the non-gRPC handler (main.go:135). HTTP mode calls newMux(svc) (:184), which layers the three proto routes on top of the same probe. Nothing is built for a mode that doesn't use it.

cmd/benchmarking/glutton/main.go
sairajp-rewind: Added, no re-read of the prefix in either mode. The digest streams through io.MultiWriter(f, h) as the bytes are generated (main.go:451). TRUNCATE is complete at that point. OVERWRITE is the only mode that can leave a tail from a larger earlier write, and the cursor is already sitting at size, so it just continues io.Copy(h, f) to EOF and folds that into the same digest. The response also returns the real on-disk size rather than the requested one, so it lines up with ReadDisk.

internal/benchmarking/boomer/glutton/durdir.go
sairajp-rewind: Bootstrap now uses the configured ResumeMode like every other iteration, so implicit mode issues no ResumeActor at all; the forced-explicit special case and its test are deleted.

benchmarking/README.md
sairajp-rewind: Done

internal/benchmarking/boomer/glutton/fixture_test.go
maxsmythe: Let's file it as an issue.

internal/benchmarking/boomer/glutton/fixture_test.go
maxsmythe: Lets move the fake to `internal/benchmarking/glutton/fake`

We should factor glutton's core logic from main.go into `internal/benchmarking/glutton/...`, but that can be done later.

Override fields on fakes can be public, no need to localize that code (indeed, the point of centralized fakes is to make it easy to override behavior in multiple places).

Also, to be sure, there is no need to write tests for the test fixture, so this should minimally impact line count.

sairajp-rewind: Moved to `internal/benchmarking/glutton/fake`; `diskServer` → `fake.Server` with the
  override fields exported. No tests for the fixture.
  
  One deliberate duplication: the fake declares its own `WriteDiskRoute`/`ReadDiskRoute`
  instead of importing the client's (`durdir.go:54-55`), so it depends on nothing. Both
  collapse onto one definition when glutton's core moves into `internal/benchmarking/glutton/...`.

internal/benchmarking/boomer/glutton/fixture_test.go
sairajp-rewind: Added: https://github.com/agent-substrate/substrate/issues/1044

cmd/benchmarking/glutton/main.go
```diff
@@ -405,12 +446,68 @@ func (s *gluttonService) WriteDisk(ctx context.Context, req *glutton.WriteDiskRe
...
 	defer f.Close()
 
-	if err := streamRandomBytes(f, int64(req.GetSize())); err != nil {
+	h := sha256.New()
+	size := int64(req.GetSize())
+	if err := streamRandomBytes(io.MultiWriter(f, h), size); err != nil {
 		return nil, status.Errorf(codes.Internal, "write %s: %v", path, err)
 	}
 
+	// OVERWRITE has no O_TRUNC, so bytes past size can survive a larger earlier
```
maxsmythe: nit:

"bytes past size can survive a larger earlier write" -> "bytes from a larger, earlier write will persist"

benchmarking/README.md
```diff
@@ -56,6 +56,52 @@ You can also configure things like the number of users, how quickly those users
...
+User classes implemented in boomer rather than Python are selected at deploy
+time — the stack runs one per deployment:
+
+```bash
+./benchmarking/locust/deploy.sh --deploy --user-class durdir
+```
+
+### Headless
+
+`runner.py` runs a test without the web UI, writing CSVs, logs and traces to
```
maxsmythe, line 74: nit: clarify that this is a command that is invoked by automation as a job on workload cluster. It is not meant to be run on a local machine to run tests in a headless mode.

Parallel documentation with deploy.sh makes it seem like the two approaches have equivalent UX.

maxsmythe (dismissed):
LGTM after two nits.

Thank you for the PR!

cmd/benchmarking/glutton/main.go
sairajp-rewind: Addressed

benchmarking/README.md
sairajp-rewind: Addressed

maxsmythe:
LGTM, thank you for all the work!

PR #909, opened 2026-08-13
atenet: add agentgateway egress support

EItanya:
## Summary

  Add agentgateway support to the actor egress gateway. `--atenet-router=agentgateway` now selects
  agentgateway for both ingress and egress, while Envoy remains the default.

  ## What changed

  - Configure agentgateway to:
    - terminate downstream mTLS using the actor-identity CA
    - accept HTTP CONNECT
    - authorize each CONNECT through the existing egress `ext_proc` handler
    - dynamically connect to the requested `IP:port`
  - Pass agentgateway’s verified `source.certificate` to `ext_proc`.
  - Preserve the existing certificate re-verification and actor identity/status checks.
  - Add agentgateway egress Kustomize and targeted deployment support.
  - Make the egress demo verifier dataplane-neutral.
  - Keep the existing Envoy path unchanged.


LiorLieberman:
looking at this today!

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -48,6 +48,9 @@ import (
 )
 
 const (
+	// ClientCertificateAttribute is the PEM peer certificate agentgateway
```
LiorLieberman: is this only relevant for agentgateway?

If yes, maybe you call that agentgatewayClientCertificateAttribute?

LiorLieberman: also is this dynamic metadata? just curios how you compare that vs the forwarded Client Cert Header ?

cmd/atenet/internal/router/extproc/dispatch.go
```diff
@@ -31,6 +31,9 @@ const (
 )
 
 const (
+	// DirectionAttribute is set from a dataplane expression, not a client
+	// header, by dataplanes without Envoy filter chains.
```
LiorLieberman, line 35: >by dataplanes without Envoy filter chains.

who sets this request attribute, agw? Also I think we probably want some more specific namespace like ate.calllout(maybe a better key name?).direction

hack/install-ate.sh
```diff
@@ -67,7 +67,7 @@ function usage() {
   echo "  --delete-ate-system                    Delete core system"
   echo "  --delete-all                           Delete core system and all registered demos"
   echo "  --ateapi-client-auth=cert|token        Select how in-cluster clients authenticate to ateapi for --deploy-ate-system (default: cert; the server always accepts both)"
-  echo "  --atenet-router=envoy|agentgateway     Select the atenet router dataplane (default: envoy)"
+  echo "  --atenet-router=envoy|agentgateway     Select the ingress and egress dataplane (default: envoy)"
```
LiorLieberman, line 70: hmmm dont we want separate flags for that..? `atenet-egress|atenet-ingress`?

LiorLieberman (commented):
LGTM overall. left a few comments.

cmd/atenet/internal/router/egress/egress.go
EItanya: Yes, this attribute is agentgateway-specific. I renamed it to `agentgatewayClientCertificateAttribute` and made it unexported. It is an ext_proc request attribute, not dynamic metadata: agentgateway evaluates `source.certificate` from the verified downstream TLS connection into `ProcessingRequest.attributes`. Envoy continues to use the sanitized XFCC header. The handler independently re-verifies the certificate in both cases.

cmd/atenet/internal/router/extproc/dispatch.go
EItanya: Yes, agentgateway sets this through the extProc `requestAttributes` config. I renamed the key to the more specific `ate.extproc.direction`.

hack/install-ate.sh
EItanya: I would prefer to keep one flag initially. We do not currently have a use case for mixing Envoy ingress with agentgateway egress or vice versa, and separate flags add configuration combinations we would need to support and test. We can split it later if that need appears.

LiorLieberman:
ok, lgtm. But lets wait until #715 is merged since it touches a few parts that your are touching here

LiorLieberman:
(will get merged today, pending on a few last touchups from keith)

cmd/atenet/internal/router/egress/egress.go
LiorLieberman: I cant see the rename, did you commit it?

cmd/atenet/internal/router/extproc/dispatch.go
LiorLieberman: I cant see the rename, did you commit it?

PR #911, opened 2026-08-13
atenet: bind the Envoy sockets and Services dual-stack

### LINKED ISSUE #910: atenet Envoy sockets and Services lack IPv6 support (hardcoded to 0.0.0.0)
## Overview
Currently, all Envoy listener and admin sockets in `atenet` are hardcoded to bind to the IPv4 wildcard (`0.0.0.0`), and the gateway Kubernetes Services lack explicit dual-stack configuration. This prevents `atenet` from functioning correctly on IPv6-only or Dual-Stack clusters.

## Hardcoded Bindings
The following components are currently restricted to IPv4:

| Socket | Location | Current Binding |
| :--- | :--- | :--- |
| **Router Ingress** (`:8080` / `:8443`) | `router/xds.go` (`buildListener`, `buildHttpsListener`) | `0.0.0.0` |
| **Router Envoy Admin** (`:9901`) | `manifests/ate-install/atenet-router.yaml` | `0.0.0.0` |
| **Egress Envoy Admin** (`:15000`) | `manifests/ate-install/atenet-egress.yaml` | `0.0.0.0` |
| **Egress CONNECT Listener** (`:443`) | `manifests/ate-install/atenet-egress.yaml` | `0.0.0.0` |
| **Gateway Services** (`atenet-router`/`-egress`) | Both manifests | `SingleStack` (Default) |

### 1. IPv6-Only Clusters (CrashLoopBackOff)
On an IPv6-only cluster, the pods never start successfully. The `kubelet` attempts to probe the pod using its assigned IPv6 address, but Envoy is only listening on `0.0.0.0`. This results in connection refusal and causes the pods (like `atenet-egress`) to crashloop.

> **Example Error:**
> ```
> Startup probe failed: Get "http://[fd00:10:244::18]:15000/ready": connect: connection refused
> ```

### 2. Dual-Stack Clusters (Dark Data Path)
On a dual-stack cluster, the pods may report as healthy, but half of the data path is dark. The router only answers on its Service's IPv4 `ClusterIP` and ignores IPv6 traffic entirely. Furthermore, because the Kubernetes Services do not define an `ipFamilyPolicy`, they only hand out a single IPv4 address, keeping the IPv6 path unreachable.

---

### Not in Scope (Related Outbound IPv4 Restrictions)
The following are distinct IPv4 limitations that leave gateways IPv4-only for *outbound* resolution and are not covered by this bug:
* `dns_lookup_family: V4_ONLY` on the egress dynamic forward proxy.
* The OTLP collector cluster defaults to Envoy's V4-only mode.
* `cmd/atenet/internal/dns/corefile.go` emits `template IN A` only (no AAAA path).
* `atunnel` original-destination lookup lacks IPv6 support.

ygao-g:
Fixes #910
Part of #246

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR (none needed — no flags, APIs, or user-facing behaviour change)

---

The router and egress manifests bind the IPv4 wildcard on every Envoy socket, and neither gateway's Service asks for a second IP family. On a dual-stack cluster the router answers on its IPv4 ClusterIP and on nothing at all for IPv6; on an IPv6-primary cluster the kubelet cannot probe the pod on its only address, so `atenet-egress` crashloops while Envoy itself starts fine.

Both gateways now bind `::` as well, and both Services ask for `PreferDualStack`. A new `hack/verify` check guards the `ipv4_compat` flag that the in-pod loopback callers depend on, since nothing else in the tree reads these manifests. The experimental `atenet-egress-with-sdsmint.yaml` variant is deliberately out of scope.

This makes both gateways *accept* IPv6; it does not make them reach IPv6 destinations. The egress `dns_lookup_family`, the DNS AAAA path, and atunnel's original-destination lookup stay IPv4, and are tracked in #246 and #686.

## Testing

CI has no IP-family matrix, so the green e2e run here is the IPv4 regression check and nothing more. The other two families were verified by hand on kind, using `IP_FAMILY` from #877.

- **Dual-stack.** The networking suite passes with zero skips: an actor is reached over each of the router's ClusterIPs, and an IPv4 actor transits the egress `[::]:443` socket. Reverting only the two bind lines turns those assertions red, so they are not passing vacuously. `ipv4_compat` was read from `/config_dump` — `/listeners` never emits it.
- **IPv6-only**, with the kind CoreDNS fix from #958 underneath. `atenet-egress` goes from `1/2 CrashLoopBackOff` (64 restarts, startup probe refused) to `2/2 Running`, and `TestActorDirectAccess` passes, including through the router's ingress. `TestActorEgress` still fails inside the actor's netns, which is #979's scope.

### DIFF
--- a/cmd/atenet/internal/router/dataplane.go
+++ b/cmd/atenet/internal/router/dataplane.go
@@ -39,7 +39,9 @@ type dataplaneHealthCheck struct {
 func (r atenetRouter) healthCheck() dataplaneHealthCheck {
 	switch r {
 	case atenetRouterEnvoy:
-		return dataplaneHealthCheck{url: "http://127.0.0.1:9901/ready", expectedBody: "LIVE"}
+		// localhost, not 127.0.0.1: the admin socket binds `::`, so the dial
+		// has to be able to fall through to the IPv6 loopback.
+		return dataplaneHealthCheck{url: "http://localhost:9901/ready", expectedBody: "LIVE"}
 	case atenetRouterAgentgateway:
 		return dataplaneHealthCheck{url: "http://127.0.0.1:15021/healthz/ready", expectedBody: "ready"}
 	default:
--- a/cmd/atenet/internal/router/xds.go
+++ b/cmd/atenet/internal/router/xds.go
@@ -1108,6 +1108,27 @@ func (x *XdsServer) buildTracing() *hcmv3.HttpConnectionManager_Tracing {
 	}
 }
 
+// dualStackAdditionalAddresses returns the IPv6 half of a dual-stack ingress
+// listener, to pair with a primary 0.0.0.0 socket on the same port. Ipv4Compat
+// stays false: clearing IPV6_V6ONLY would collide with that primary.
+func dualStackAdditionalAddresses(port int) []*listenerv3.AdditionalAddress {
+	return []*listenerv3.AdditionalAddress{
+		{
+			Address: &corev3.Address{
+				Address: &corev3.Address_SocketAddress{
+					SocketAddress: &corev3.SocketAddress{
+						Address:    "::",
+						Ipv4Compat: false,
+						PortSpecifier: &corev3.SocketAddress_PortValue{
+							PortValue: uint32(port),
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
 func (x *XdsServer) buildListener() *listenerv3.Listener {
 	hcm := x.buildHcm("ingress_http", true)
 
@@ -1123,6 +1144,7 @@ func (x *XdsServer) buildListener() *listenerv3.Listener {
 				},
 			},
 		},
+		AdditionalAddresses: dualStackAdditionalAddresses(x.ingressPort),
 		FilterChains: []*listenerv3.FilterChain{
 			{
 				Filters: []*listenerv3.Filter{
@@ -1182,6 +1204,7 @@ func (x *XdsServer) buildHttpsListener() *listenerv3.Listener {
 				},
 			},
 		},
+		AdditionalAddresses: dualStackAdditionalAddresses(x.httpsPort),
 		FilterChains: []*listenerv3.FilterChain{
 			{
 				Filters: []*listenerv3.Filter{
@@ -1213,6 +1236,7 @@ func (x *XdsServer) buildConnectTerminateListener() *listenerv3.Listener {
 				},
 			},
 		},
+		AdditionalAddresses: dualStackAdditionalAddresses(x.connectPlainTextPort),
 		FilterChains: []*listenerv3.FilterChain{
 			{
 				Filters: []*listenerv3.Filter{
@@ -1246,6 +1270,7 @@ func (x *XdsServer) buildConnectTerminateTLSListener() *listenerv3.Listener {
 				},
 			},
 		},
+		AdditionalAddresses: dualStackAdditionalAddresses(x.connectTLSPort),
 		FilterChains: []*listenerv3.FilterChain{
 			{
 				Filters: []*listenerv3.Filter{
--- a/manifests/ate-install/atenet-egress.yaml
+++ b/manifests/ate-install/atenet-egress.yaml
@@ -37,7 +37,8 @@ data:
   envoy.yaml: |
     admin:
       address:
-        # ipv4_compat is load-bearing: see --envoy-admin-address below.
+        # ipv4_compat is load-bearing: the probes below are the kubelet
+        # dialling the pod IP, which is IPv4 on an IPv4 cluster.
         socket_address: { address: "::", ipv4_compat: true, port_value: 15000 }
     static_resources:
       listeners:
@@ -307,7 +308,7 @@ spec:
         # defaults to (that is the ingress gateway's port). Without this the
         # drain sequence dials a closed port, reads the connection refusal as
         # "Envoy already exited", and reports a drain it never performed.
-        - --envoy-admin-address=127.0.0.1:15000
+        - --envoy-admin-address=localhost:15000
         env:
         - name: POD_NAME
           valueFrom:
--- a/manifests/ate-install/atenet-egress.yaml
+++ b/manifests/ate-install/atenet-egress.yaml
@@ -37,12 +37,15 @@ data:
   envoy.yaml: |
     admin:
       address:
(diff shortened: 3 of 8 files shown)


ygao-g:
@bowei mind taking a look?

cmd/atenet/internal/router/xds.go
```diff
@@ -797,6 +797,32 @@ func (x *XdsServer) buildTracing() *hcmv3.HttpConnectionManager_Tracing {
...
+// manifests/ate-install/atenet-router.yaml is a single socket, so it sets the
+// opposite.
+func dualStackAdditionalAddresses(port uint32) []*listenerv3.AdditionalAddress {
+	return []*listenerv3.AdditionalAddress{
+		{
+			Address: &corev3.Address{
+				Address: &corev3.Address_SocketAddress{
+					SocketAddress: &corev3.SocketAddress{
+						Address:    "::",
+						Ipv4Compat: false,
```
bowei, line 1121: Do we want to sync up the configuration in the YAML with the configuration here?

bowei: Actually, the comment here might have confused me. This is talking about two different sockets? If it is, can you remove the comment above.

ygao-g: Yes, two different sockets. And I have removed the confusing comments.

 This is the router's ingress data-plane listener; the `ipv4_compat: true` you saw in the YAML is the Envoy admin socket on 9901. I've cut the comment down to the one line explaining why `Ipv4Compat` is false here — the cross-reference to the admin socket was the confusing part and it's gone.

On the original question: the admin sockets can't take this shape anyway, since `bootstrap.v3.Admin` has a single `address` field and no `additional_addresses` — one `::` socket with `ipv4_compat` is the only way to serve both families there. The egress `:443` listener is a real Listener, so that one could match the ingress shape if you'd rather have a single idiom across listeners. Happy to do it; it's a few lines plus the test that asserts the current shape.

cmd/atenet/internal/router/dataplane.go
```diff
@@ -36,6 +36,8 @@ type dataplaneHealthCheck struct {
 // untouched, so atunnel always authorizes by the actor's own DNS name --
 // ingress.New needs no per-dataplane routing mode.
 
+// healthCheck dials IPv4 loopback, which is why the admin socket in
```
ygao-g: Better to move to the yaml file, where ipv4_compat is specified.

cmd/atenet/internal/router/xds.go
```diff
@@ -1123,6 +1144,7 @@ func (x *XdsServer) buildListener() *listenerv3.Listener {
 				},
 			},
 		},
+		AdditionalAddresses: dualStackAdditionalAddresses(uint32(x.ingressPort)),
```
ygao-g: Nit. Instead of always using uint32(...), change the helper function signature to take the type `x.ingressPort`, and convert it to a uint32 inside the helper function

cmd/atenet/internal/router/dataplane.go
ygao-g: Just remove these added new comments then.

manifests/ate-install/atenet-egress.yaml
```diff
@@ -379,6 +382,8 @@ metadata:
   namespace: ate-system
 spec:
   type: ClusterIP
+  # Prefer, not Require: Require fails Service creation on a single-stack cluster.
```
ygao-g: Nit. Remove the explanation. Apply to other occurrences as well.

cmd/atenet/internal/router/xds.go
ygao-g: Happy to switch if you'd rather have it — but I think the two shapes are equivalent here, so it's a style call rather than a correctness one.

The reason to keep an IPv4 primary is to stop v4 peers arriving as `::ffff:`-mapped addresses. Nothing on either path reads the downstream peer: ingress ext_proc requests one attribute, the authority filter-state (`xds.go:1016`), and egress requests one, `xds.filter_chain_name` (`atenet-egress.yaml:134-135`). There's no `use_remote_address`, no XFF config and no `DOWNSTREAM_REMOTE_ADDRESS` in `cmd/atenet` or either manifest, and the `original_dst` on ingress is the destination ext_proc supplies in dynamic metadata, not the source.

I kept egress as one socket to match the admin listener in the same file, which is `::` + `ipv4_compat: true` and where the flag is load-bearing — the drain sequence dials `127.0.0.1:15000` (`atenet-egress.yaml:310`), as the comment at line 40 already notes. The symmetric move doesn't exist: ingress can't take `ipv4_compat` while keeping its `0.0.0.0` primary, because both would bind IPv4:port — that's what "collide with that primary" means at `xds.go:1113`.

So it's one idiom within the file against one across the component, and I don't have a strong preference. Say which you'd rather see and I'll match it — a few lines plus the test.

hack/verify/atenet-admin-bind.sh
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
```
bowei: what is this script used for

I think we should just rely on e2e tests to tell us that IPv6 is working.

bowei:
Changes look ok, but can we remove the extra script?

hack/verify/atenet-admin-bind.sh
ygao-g: Removed. CI does catch it on the egress side — the kubelet [probes that admin port](https://github.com/agent-substrate/substrate/blob/20951c06a5e4f7720bd956e2b4c926f46cfa65c1/manifests/ate-install/atenet-egress.yaml#L254-L263), so dropping `ipv4_compat` means the pod never goes Ready and every run turns red. The router's admin port has no such probe, and the drain [reads a refused dial as "Envoy already exited"](https://github.com/agent-substrate/substrate/blob/fc02127b/cmd/atenet/internal/router/envoydrain.go#L88-L93), so I dropped the dependency instead of guarding it: the in-pod callers dial `localhost` now, with a unit test against an IPv6-only loopback listener.

ygao-g:
Done. The in-pod callers now dial `localhost` so nothing quietly depends on `ipv4_compat`; reasoning in the thread on the file.

bowei (approved):

PR #914, opened 2026-08-13
A few changes to the RPC methods for `ActorSnapshot`/`ActorSnapshotTag`.

### LINKED ISSUE #912: Revisit ActorSnapshot/ActorSnapshotTag gRPC methods
There are a few things in the API methods for `ActorSnapshot` and `ActorSnapshotTag` that look wrong to me. I'll try to go over them one by one.

The `ActorSnapshotRef` message, as defined today, it's not really consistent with what we call a `<Type>Ref` (which is an optional atespace, and a name). Instead, it's defined as:

```
message ActorSnapshotRef {
  oneof reference {
    ObjectRef snapshot = 1;
    ObjectRef tag = 2;
  }
}
```

This caught my attention, so I started trying to unravel it, why we need it? It's been used in two places:

- It's used in `CreateActor`, to specify which snapshot should be use as a starting point when creating a new actor.

```
message CreateActorRequest {
  // The actor to create.
  Actor actor = 1;

  // Optional durable snapshot used to initialize the Actor.
  ActorSnapshotRef source_snapshot = 2;
}
```

There are a few issues with this: a) the field is a top level field, and not part of the 'actor' resource (this deviates from the API convention we have and will very likely bite us in the future with things liked declarative clients). Top level fields should be reserved for "control" fields, such as a dry run option, and b) it's not really needed. The implementation returns an error if a `snapshot` is specified. And c) the actor doesn't keep any audit trail of the intent / base snapshot used when the actor was created.

- It's also used In `GetActorSnapshot`.

```
message GetActorSnapshotRequest {
  ActorSnapshotRef snapshot = 1;
}
```

This also deviates from the API convention. The idea is that a single method can return `ActorSnapshots`, resolving tags if necessary. I find this an unnecessary optimization, which tries to save a method (GetActorSnapshotTag) at the cost of less orthogonal APIs.

### Proposal

I think we can untangle this and solve the mentioned issues by:

* Removing the `ActorSnapshotRef` message.
* Adding a `GetActorSnapshotTag` method for the `ActorSnapshotTag` resource. The request should (as all Get<Type> requests) have a single `ObjectRef tag` field.
* Change `GetActorSnapshotRequest` to have a single `ObjectReg snapshot` field.
* Add a new field to `Actor`, called `source_snapshot`, which the following schema:

```
message ActorSnapshotSource {
  // If specified, the actor will fork from this snapshot on the first resume.
  // Immutable.
  ObjectRef tag = 1;

  // A reference to the snapshot the tag resolved to at creation time.
  // Output only.
  ObjectRef snapshot = 2;

  // UID of the snapshot the tag resolved to at creation time.
  // Output only.
  string snapshot_uid = 3;
}
```

juli4n:
* Remove the ActorSnapshotRef message.
* Add a `GetActorSnapshotTag` method for the `ActorSnapshotTag` resource. The request should (as all Get requests) have a single `ObjectRef` tag field.
* Change `GetActorSnapshotRequest` to have a single `ObjectRef` snapshot field.
* Add a new field to `Actor,` called `source_snapshot`, which the following schema:

```
message ActorSnapshotSource {
  // If specified, the actor will fork from this snapshot on the first resume.
  // Immutable.
  ObjectRef tag = 1;

  // A reference to the snapshot the tag resolved to at creation time.
  // Output only.
  ObjectRef snapshot = 2;

  // UID of the snapshot the tag resolved to at creation time.
  // Output only.
  string snapshot_uid = 3;
}
```

See #912 for a more detailed description of the rationale for this change. Fixes #912

### DIFF
--- a/cmd/ateapi/internal/controlapi/actor_snapshot.go
+++ b/cmd/ateapi/internal/controlapi/actor_snapshot.go
@@ -45,10 +45,12 @@ var actorSnapshotTagScopeNames = func() []string {
 }()
 
 func (s *Service) GetActorSnapshot(ctx context.Context, req *ateapipb.GetActorSnapshotRequest) (*ateapipb.ActorSnapshot, error) {
-	if err := validateActorSnapshotRef(req.GetSnapshot(), "snapshot"); err != nil {
-		return nil, err
+	// TODO: extract into validateGetActorSnapshotRequest(req) field.ErrorList,
+	// as done for other RPC methods, and return via toGRPCStatusError.
+	if errs := resources.ValidateObjectRef(req.GetSnapshot(), field.NewPath("snapshot")); len(errs) > 0 {
+		return nil, status.Error(codes.InvalidArgument, errs.ToAggregate().Error())
 	}
-	snapshot, _, _, err := s.getActorSnapshot(ctx, req.GetSnapshot())
+	snapshot, err := s.persistence.GetActorSnapshot(ctx, req.GetSnapshot().GetAtespace(), req.GetSnapshot().GetName())
 	if errors.Is(err, store.ErrNotFound) {
 		return nil, status.Error(codes.NotFound, "ActorSnapshot not found")
 	}
@@ -58,7 +60,25 @@ func (s *Service) GetActorSnapshot(ctx context.Context, req *ateapipb.GetActorSn
 	return snapshot, nil
 }
 
+func (s *Service) GetActorSnapshotTag(ctx context.Context, req *ateapipb.GetActorSnapshotTagRequest) (*ateapipb.ActorSnapshotTag, error) {
+	// TODO: extract into validateGetActorSnapshotTagRequest(req) field.ErrorList,
+	// as done for other RPC methods, and return via toGRPCStatusError.
+	if errs := resources.ValidateObjectRef(req.GetTag(), field.NewPath("tag")); len(errs) > 0 {
+		return nil, status.Error(codes.InvalidArgument, errs.ToAggregate().Error())
+	}
+	_, tag, err := s.persistence.GetActorSnapshotByTag(ctx, req.GetTag().GetAtespace(), req.GetTag().GetName())
+	if errors.Is(err, store.ErrNotFound) {
+		return nil, status.Error(codes.NotFound, "ActorSnapshot tag not found")
+	}
+	if err != nil {
+		return nil, fmt.Errorf("while getting actor snapshot tag: %w", err)
+	}
+	return tag, nil
+}
+
 func (s *Service) ListActorSnapshots(ctx context.Context, req *ateapipb.ListActorSnapshotsRequest) (*ateapipb.ListActorSnapshotsResponse, error) {
+	// TODO: extract into validateListActorSnapshotsRequest(req) field.ErrorList,
+	// as done for other RPC methods, and return via toGRPCStatusError.
 	var fldPath *field.Path
 	var errs field.ErrorList
 	if req.GetAtespace() != "" {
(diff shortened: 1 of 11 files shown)


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -710,9 +725,6 @@ message GetActorRequest {
 message CreateActorRequest {
   // The actor to create.
   Actor actor = 1;
-
-  // Optional durable snapshot used to initialize the Actor.
-  ActorSnapshotRef source_snapshot = 2;
```
zoez7, line 715: I see, we removed it here because the field is now in Actor itself. 👍🏼

cmd/kubectl-ate/internal/cmd/actor_snapshots.go
```diff
@@ -114,7 +117,7 @@ var createActorSnapshotTagCmd = &cobra.Command{
 		defer client.Close()
 
 		resp, err := client.TagActorSnapshot(ctx, &ateapipb.TagActorSnapshotRequest{
-			Snapshot: &ateapipb.ActorSnapshotRef{Reference: &ateapipb.ActorSnapshotRef_Snapshot{Snapshot: &ateapipb.ObjectRef{Atespace: createTagAtespaceFlag, Name: createTagSnapshotFlag}}},
+			Snapshot: &ateapipb.ObjectRef{Atespace: createTagAtespaceFlag, Name: createTagSnapshotFlag},
```
zoez7, line 120: Do we want the request to specify a atespace, or should the server always assign the tag the same atespace as the Snapshot itself?

cmd/ateapi/internal/controlapi/create_actor.go
```diff
@@ -46,19 +46,20 @@ func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequ
...
 	var sourceSnapshot *ateapipb.ActorSnapshot
-	var sourceSnapshotRef *ateapipb.ObjectRef
-	if ref := req.GetSourceSnapshot(); ref != nil {
-		if _, ok := ref.GetReference().(*ateapipb.ActorSnapshotRef_Tag); !ok {
-			return nil, status.Error(codes.FailedPrecondition, "source ActorSnapshot must be referenced by tag")
-		}
-		lock, snapshot, canonical, tag, err := s.lockActorSnapshot(ctx, ref)
+	var sourceSnapshotInfo *ateapipb.ActorSnapshotSource
+	if src := in.GetSourceSnapshot(); src != nil {
+		lock, snapshot, canonical, tag, err := s.lockActorSnapshot(ctx, nil, src.GetTag())
```
zoez7: Nit: do we want to separate [lockActorSnapshot](https://github.com/agent-substrate/substrate/blob/febeb7afe7e720e00f4755c71be1e9b1b43d52cf/cmd/ateapi/internal/controlapi/actor_snapshot.go#L214) to both a GetActorSnapshot and a LockActorSnapshot?

cmd/ateapi/internal/controlapi/actor_snapshot.go
```diff
@@ -188,43 +212,41 @@ func (s *Service) DeleteActorSnapshotTag(ctx context.Context, req *ateapipb.Dele
...
-	canonical := &ateapipb.ObjectRef{Atespace: snapshot.GetMetadata().GetAtespace(), Name: snapshot.GetMetadata().GetName()}
-	return snapshot, canonical, tag, nil
+	resolvedCanonical := &ateapipb.ObjectRef{Atespace: snapshot.GetMetadata().GetAtespace(), Name: snapshot.GetMetadata().GetName()}
+	return snapshot, resolvedCanonical, resolvedTag, nil
 }
 
-func (s *Service) lockActorSnapshot(ctx context.Context, ref *ateapipb.ActorSnapshotRef) (*store.Lock, *ateapipb.ActorSnapshot, *ateapipb.ObjectRef, *ateapipb.ActorSnapshotTag, error) {
-	_, canonical, _, err := s.getActorSnapshot(ctx, ref)
+// lockActorSnapshot resolves and locks a snapshot by its canonical identity
+// or by tag. Exactly one of canonical or tag must be set.
```
zoez7, line 235: > Exactly one of canonical or tag must be set.

Add a check for this to return error if both are set?

zoez7 (dismissed):

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -304,13 +312,20 @@ message ObjectRef {
...
-    ObjectRef snapshot = 1;
-    ObjectRef tag = 2;
-  }
+// ActorSnapshotSource records how an Actor was seeded from an ActorSnapshot.
+// Set once at CreateActor; every field is immutable afterward.
+message ActorSnapshotSource {
+  // The tag specified by the caller at CreateActor.
+  ObjectRef tag = 1;
+
+  // Output only. Canonical (atespace, name) of the snapshot the tag
```
zoez7, line 321: Also probably out of scope for this PR:  I'm curious why this is output only? Why a snapshot must be tagged before we can use it in CreateActor()?

cmd/ateapi/internal/controlapi/actor_snapshot.go
juli4n: Done.

cmd/ateapi/internal/controlapi/create_actor.go
juli4n: Yes. I thought about that but I would prefer to keep the scope of this PR small. FWIW, I think we need to revisit the whole locking mechanism here.

cmd/kubectl-ate/internal/cmd/actor_snapshots.go
juli4n: I think it should require the tag, just like any other Create request.

pkg/proto/ateapipb/ateapi.proto
juli4n: This is the current semantics. We could turn this into an immutable field if we ever want that.

zoez7 (dismissed):

PR #917, opened 2026-08-13
atelet: pull image credentials from the kubelet credential provider

BenTheElder:
atelet authenticated image pulls with Google application default credentials, applied to gcr.io and pkg.dev only. That compiles one cloud's registry story into the binary and leaves every other cloud unserved.

Every managed node already answers "which credentials does this registry want": the kubelet ships a credential provider executable and a CredentialProviderConfig naming the registries it handles. Reuse that contract instead. internal/credentialprovider is an authn.Keychain that speaks the kubelet's exec-plugin protocol -- config parsing and validation, the kubelet's image match rules, the request/response exchange over stdio, and per-cacheKeyType response caching. atelet resolves credentials through it when --image-credential-provider-config and --image-credential-provider-bin-dir are set (the same flags the kubelet takes), and the DaemonSet mounts the node's plugin dir and config read-only.

The result is registry-agnostic: the same atelet authenticates to Artifact Registry on GKE and to ECR on EKS, given each node's own plugin. The GCP application default credentials path stays for installs that do not set the new flags.

POC for proposal at https://docs.google.com/document/d/1rWkmIWTAo_MUGleZ48k3py9jmL7Yfjdf1qffdyiLJCw/edit?resourcekey=0-z9N8Ee108biqI4CsTTjQUg&tab=t.0#heading=h.6vr9xodr524z (shared with the mailinglist from the README.md for access)


Stevenjin8 (commented):
I have two thoughts:

1. In various discussions, I've seen the credential-provider-bin-dir be described as a host mount. I have nothing against it being a host mount, I just don't want to make it a requirement. For example, AKS' credential provider is dynamically linked:

```
# ldd acr-credential-provider
	linux-vdso.so.1 (0x00007fffef38b000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fe670585000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fe67079b000)
```

and the last thing any of us wants to deal with is libc compatibility shenanigans.

2. In the `CredentialProviderRequest` we pass into the credential provider, we are missing `serviceAccountToken` and `serviceAccountAnnotations`. `serviceAccountToken` isn't too bad. It's usually a file mounted on the pod, so we can just pass in the filename as an arg.

`serviceAccountAnnotations` is much more annoying but still important. Ideally, we look it up on the kubeapiserver and maybe have some caching in place. It puts the api server in the path, but I think its fine because it only happens on golden image creation (and we can cache/have watchers)?

BenTheElder:
> In various discussions, I've seen the credential-provider-bin-dir be described as a host mount. I have nothing against it being a host mount, I just don't want to make it a requirement. For example, AKS' credential provider is dynamically linked:

Yeah, I expected this to happen with some providers, the most obvious workaround I have to suggest is a custom atelet image with the appropriate binary + libraries added directly to the image. That will add some friction to deploying to that environment but it doesn't seem terrible.

We can make the mount customizable as well, and we can consider using a minimal libc base image.

Hopefully more vendors can leverage GOFIPS140 for static binaries in the near future ...

> serviceAccountAnnotations is much more annoying but still important. Ideally, we look it up on the kubeapiserver and maybe have some caching in place. It puts the api server in the path, but I think its fine because it only happens on golden image creation (and we can cache/have watchers)?

We _really_ don't want apiserver in the path, or the need to have that many objects, like, we want O(100ms) end to end spawning lightweight actors, churning them multiple times per second. We're also targeting ~1B actors with ~1M active (which will require some other changes ...). 

We absolutely do not have that today, e.g. the sandbox setup is too slow, but we're treating every performance gap as a bug to burn down. We're basically using Kubernetes for compute. Actors do not have KSA.

This also needs to be not just on golden image creation, we have to pull the image again to another node if the actor resumes there.

The current inbuilt GCP support is atelet wide, not per-actor, so this is not a regression on that front, it just opens it up to more vendors without compiling them in (and we can look at dropping the built-in GCP support).

Stevenjin8:
@BenTheElder 

> Hopefully more vendors can leverage GOFIPS140 for static binaries in the near future ...

Some vendors are really against static linking. But I don't think that should concern the project too much.

> We absolutely do not have that today, e.g. the sandbox setup is too slow, but we're treating every performance gap as a bug to burn down. We're basically using Kubernetes for compute. Actors do not have KSA.

I was imagining that we use atelet's KSA to pull

If we really don't want to be truly kubernetes agnostic, that's fine, but let's make it super clear that we only fill in the `image` field of `CredentialProvideRequest` and vendors need to find a way to make that work. That could be by using a different credential provide binary than that on the node

cc @aramase

ahmedtd:
> I was imagining that we use atelet's KSA to pull

To make an analogy to Kubernetes-land, that is the same as using the "node" service account to pull the image, which corresponds to no service account token input in the helper binary call.

The token input is for fetching the image using the identity of the workload.  Eventually, we could plumb the Substrate actor JWT there.

BenTheElder:
I think we should also drop the existing GCP bits, will update. 

Just discussing how we handle the caching topic, @ahmedtd suggested we ignore cacheDuration below one minute for now.

https://pkg.go.dev/k8s.io/kubernetes/pkg/kubelet/apis/config#CredentialProviderConfig:~:text=//%20The%20pod%20metadata,the%20caching%20of

I think this makes sense, at least for now. If we decide we don't want that generally, we can instead tune GCP via these env on the installed atelet pods:
https://github.com/kubernetes/cloud-provider-gcp/blob/4e24531247d7f37b936881c0a5c186ffc8cd72d1/cmd/auth-provider-gcp/provider/provider.go#L35-L36

But I suspect it's a really bad idea not to cache in general, so "just ignore ultra low cache duration" seems OK, and I'd rather trend towards dropping the GCP bits everywhere.

tools/validate-image-cache/main.go
```diff
@@ -225,13 +225,12 @@ func main() {
 		log.Fatalf("invalid --platform %q", *platform)
 	}
 
-	auth, err := googlecontainerauth.NewEnvAuthenticator(ctx)
-	if err != nil {
-		log.Fatalf("creating GCP authenticator (need application-default credentials): %v", err)
-	}
-
+	// This tool runs on a developer's machine, not a node, so it has no kubelet
```
BenTheElder, line 228: TODO: I'm not sure we even need to keep this tool. There's a broader cleanup of tools/ to do though so I suggest we defer this for now, this isn't used in production.

BenTheElder:
Rebased, made the proposed caching handling change, and added a commit to drop the GCP specific option / code, GCP now only uses the kubelet credential plugin.

Tested on GKE.

aoxn:
Hi, do we have a basic method to config image pull secrets, because not all cloud support the credential provider method.  @BenTheElder

tdg5:
Any motion here? I created a [competing PR](https://github.com/agent-substrate/substrate/pull/1790) before finding this one. That said, I'm also not sure this PR covers my use case.

My setup is k3s on bare metal (no managed node image), actor images in a private AWS ECR repo. Substrate 0.0.30 via the kagent-dev Helm chart, whose default is `atelet.gcpAuthForImagePulls: false` — so atelet has no credential source at all and every actor image pull goes out anonymous and fails:

```
in imageCache.EnsureImage: while resolving tag "<redacted>.dkr.ecr.us-east-1.amazonaws.com/<repo>:latest" to a digest: HEAD https://<redacted>.dkr.ecr.us-east-1.amazonaws.com/v2/<repo>/manifests/latest: unexpected status code 401 Unauthorized
```

It seems like with this fix there is no credential path that doesn't involve a plugin binary on the node, which may be right for managed nodes, but doesn't cover some other cases.

Would it be possible to also fold `authn.DefaultKeychain` into the keychain #917 builds, so a Docker config file works for any registry no plugin covers, and pulls stay anonymous when there's no config? There's precedent in-tree — `cmd/ate-setup/internal/images/digest.go` already pulls with `authn.NewMultiKeychain(authn.DefaultKeychain, google.Keychain)`; it seems like `atelet` is the outlier.

BenTheElder:
> Any motion here?

I'm going to follow-up soon with the other maintainers .. I've just been pretty busy with #1266 and hadn't gotten much follow-up yet. There are a lot of competing priorities mostly to solve difficult problems that require changing the APIs. We know this one is tractable. That said I want it fixed.

> That said, I'm also not sure this PR covers my use case.

Yeah, this avoids compiling in cloud provider SDKs. But we'll also want a solution for static creds. Last night I left a comment on an older PR related to that with a pointer to some recent changes we can build on.

BenTheElder:
> Would it be possible to also fold authn.DefaultKeychain into the keychain https://github.com/agent-substrate/substrate/pull/917 builds, so a Docker config file works for any registry no plugin covers, and pulls stay anonymous when there's no config? There's precedent in-tree — cmd/ate-setup/internal/images/digest.go already pulls with authn.NewMultiKeychain(authn.DefaultKeychain, google.Keychain); it seems like atelet is the outlier.

I'll follow up on this later, I wouldn't really take ate-setup as meaningful as it's not deployed, there's some other bootstrapping utils that are not meant to be long-lived here (setup-gcp ...). Atlet is the only "real" component we ship pulling images.

The current state is not acceptable, but this is an obviously tractable problem and there are bigger architecture changes in flight in the short term, lots to juggle.

PR #918, opened 2026-08-13
split controlapi functional tests into their own package

### LINKED ISSUE #891: Reorganize RPC handlers under cmd/ateapi/internal/controlapi
As we add more resources to the Substrate API (`ActorTemplate`/`ActorTemplateVersion`, `EgressPolicy`, etc), the cmd/ateapi/internal/controlapi directory is getting large, with a lot of go files.

Something we should consider is merging all RPC handlers for each resource type in a single file, named after the resource. We are already doing this for the `ActorSnapshot` resource, which has all the CRUDL RPC handler in a single actor_snapshot.go file.

We should probably also consider breaking functional_test.go into a dedicated test for each resource (e.g. actor_snapshot_functional_test.go) as that test is also getting way too large.

laoj2:
functional_test.go was also split into one file per resource, to match what we're doing for the RPC handlers too. See #891

I also needed to make some changes to move the tests to a separate package:

* `NewAteletDialer` now takes `options`, and `WithDialCredentials` lets a test build its own transport credentials. The fake atelet is reached over insecure transport. Didn't change existing callers. This is the only non-test change.
* Moved a few unit tests that are actually functional tests to the respective file under functionaltest/

Fixes #891

- [x] Tests pass


cmd/ateapi/internal/controlapi/functionaltest/actor_test.go
```diff
@@ -0,0 +1,1138 @@
+// Copyright 2026 Google LLC
```
juli4n, line 1: Now that you have a dedicated package (named `functionaltest`), do we need to call ti `actor_functional_test.go`? Maybe just `actor_test.go` is enough?

juli4n:
Needs rebase

cmd/ateapi/internal/controlapi/functionaltest/actor_test.go
laoj2: ah good point, done

laoj2:
CI was failing because tests were broken at HEAD. Rebased, so this should be good to be reviewed again now.

juli4n (approved):

PR #923, opened 2026-08-13
Mount an OCI image read-only into an actor

### LINKED ISSUE #783: Add an image volume source to ActorTemplate
Follow-up to [#553](https://github.com/agent-substrate/substrate/issues/553).

## Summary
Add a volume source that mounts an OCI image's filesystem read-only into an actor, the same way Kubernetes does with [image volumes](https://kubernetes.io/docs/concepts/storage/volumes/#image).

This lets content be delivered into an actor without rebuilding the container image it runs.

## API sketch
A third variant in the existing `VolumeSource`:

```yaml
volumes:
  - name: platform-bin
    image:
      reference: registry.example/shim@sha256:...
      pullPolicy: IfNotPresent

containers:
  - name: workload
    image: user-image@sha256:...
    command: ["/opt/platform/bin/shim"]
    volumeMounts:
      - name: platform-bin
        mountPath: /opt/platform/bin
```

Semantics follow the Kubernetes image volume: the referenced image's filesystem is mounted read-only at a single directory, resolved at actor start/resume.

ronlv10:
Fixes #783
Adds an `image` source to `ActorTemplate`'s `VolumeSource`: a container can mount the contents of an OCI image it does not run. This is how tooling gets into images built by third parties without rebuilding them.

```yaml
spec:
  containers:
  - name: sandbox
    image: docker.io/example/benchmark@sha256:...
    command: ["/ate/agent"]
    volumeMounts:
    - name: agent
      mountPath: /ate
  volumes:
  - name: agent
    image:
      reference: registry.example.com/agent@sha256:...
```
## How it works

atelet pulls the image through the existing layer cache and records the volume's layers in the bundle's overlay spec, next to the rootfs layers. ateom composes the volume inside the bundle — the cached layers with no writable layer on top, so the mount is read-only — and the container binds it at the declared path. The volume is composed per container: containers of one actor may mount the same volume, and each gets its own mount point inside its own bundle, all backed by the same shared layers. On resume the volume is re-composed the same way.

References must be digest-pinned, the same rule as container images: a snapshot is only valid against the exact bytes it was taken with.

On micro-VMs the volume rides the same read-only virtio-fs share as the container rootfs: ateom stages each composed volume beside the rootfs on the host, and the guest binds it into the container at the declared path.


internal/imagecache/spec.go
```diff
@@ -53,6 +53,23 @@ type OverlaySpec struct {
...
+}
+
+// ImageVolumeOverlay is one image volume's contents.
+type ImageVolumeOverlay struct {
+	// Name is the ActorTemplate's name for the volume.
+	Name string `json:"name"`
+	// ImageDigest is the manifest digest the volume's ref resolved to, in the
+	// same form and for the same reason as OverlaySpec.ImageDigest: the GC's
+	// root-set scan protects an image by digest.
+	ImageDigest string `json:"imageDigest,omitempty"`
```
BenTheElder, line 70: 🤖 **blocking** 🔴 – Nothing roots these, so the GC can evict an image volume's layers while an actor is running on them. `addSpecRoots` in `internal/imagecache/gc.go` reads `spec.ImageDigest` and `spec.Layers` and never looks at `spec.ImageVolumes`, so a volume's digest never reaches `RootSet.ImageDigests` and its layer hexes never reach `LayerHexes` or `LayerSets`.

Both eviction guards then miss. The volume's image record fails the digest check and is retired, which drops its layers' refcount to zero, and the bundle-spec fallback that would otherwise save them is keyed on `LayerHexes`. The layers go while they are bind-mounted at `<bundle>/volumes/<name>`. A resume is the visible failure: `setupImageVolumes` calls `FinalizeLayer` on a directory that no longer exists.

The watermark makes it likelier rather than rarer — the GC fires under cache pressure, which is the state this feature encourages by adding images per actor.

`addSpecRoots` walking `spec.ImageVolumes` covers it. Worth giving each volume its own `LayerSets` signature rather than folding its layers into the rootfs one, since the signature is meant to match a record's exact layer list.

BenTheElder (commented):
huge +1 in concept, it's a fair bit of code understandably, I need to read back through it in depth. attaching one plausible looking agent finding now.

internal/imagecache/spec.go
ronlv10: Good catch- added !

internal/proto/ateompb/ateom.proto
```diff
@@ -145,12 +145,24 @@ message Container {
...
+
+  // image_volume_mounts are the image volumes this container mounts, if any.
+  repeated ImageVolumeMount image_volume_mounts = 5;
 }
 
 // DurableDirVolumeMount is one durable-dir volume mounted into a container.
 message DurableDirVolumeMount {
-  // volume_name is the name the ActorTemplate gave the volume. It selects the
-  // per-volume directory atelet prepared for the actor on the host.
+  // volume_name is the name the ActorTemplate gave the volume.
```
BenTheElder: 🤖 **nit** 🟢 – This lost a sentence. It used to read "It selects the per-volume directory atelet prepared for the actor on the host", which is the only place that said what `volume_name` actually resolves to for a durable volume. The new `ImageVolumeMount` below doesn't need the same sentence — its source is composed in the bundle, not prepared by atelet — so this looks like collateral from writing the two messages side by side rather than an intended trim.

internal/ateompath/ateompath.go
```diff
@@ -137,6 +137,16 @@ func OCIBundlePath(actorUID, containerName string) string {
 	)
 }
 
+// The path is per-container: containers of one actor may mount the same
```
BenTheElder: 🤖 **nit** 🟢 – Two new exported helpers, and neither has a conventional doc comment: this one starts mid-thought ("The path is per-container") instead of with the identifier, and `ImageVolumeMountPathInBundle` below has none at all. The content is worth keeping — it explains why the path is per-container — it just needs "ImageVolumeMountPath returns ..." in front, and a line on the in-bundle variant saying when to reach for it (the caller already has a bundle path, e.g. `setupImageVolumes`).

cmd/ateom-microvm/run.go
```diff
@@ -579,6 +582,13 @@ func (s *AteomService) stageOverlayLowers(ctx context.Context, rr resolvedRuntim
 		if err := kata.ReconstructSharedDirFromImage(ctx, c.bundleRootfs, id, c.name); err != nil {
 			return nil, fmt.Errorf("while staging overlay lower for %q: %w", c.name, err)
 		}
+		// Image volumes ride the same read-only virtiofsd share.
```
BenTheElder: We're moving towards using a single share for host <> guest and overlay otherwise.

cc @Oneimu

BenTheElder: (One share for everything, instead of per volume type with the carrier container back, for performance reasons)

internal/proto/ateompb/ateom.proto
ronlv10: Done - returned it back as in main

internal/ateompath/ateompath.go
ronlv10: Done.

cmd/ateom-microvm/run.go
ronlv10: As far as I understand, this is aligned with the current implementation, right? 
Did you mean this as an FYI, should I drop the comment as obsolete, or something else?

ronlv10 (commented):
Thank you @BenTheElder !

cmd/ateom-microvm/run.go
ronlv10: removed the comment

internal/e2e/probe.go
```diff
@@ -0,0 +1,84 @@
...
+
+	root, err := FindRepoRoot()
+	if err != nil {
+		t.Fatalf("FindRepoRoot: %v", err)
+	}
+
+	// E2E_SANDBOX_CLASS selects the probe manifest variant; suites copy the
+	// probe's runtime, so this is what runs a suite on gVisor or micro-VM.
+	tmplName := "probe.yaml.tmpl"
+	if os.Getenv("E2E_SANDBOX_CLASS") == "microvm" {
```
BenTheElder: 🤖 **should-fix** 🟡 – Nothing sets `E2E_SANDBOX_CLASS`, so this branch never runs and `probe-microvm.yaml.tmpl` is dead in CI.

The micro-VM E2E step in `.github/workflows/pr-workflow.yaml` sets only `E2E_TEMPLATE_NAMESPACE`, `E2E_TEMPLATE_NAME`, and `E2E_TEMPLATE_READY_TIMEOUT`, and it still scopes the run to `./internal/e2e/suites/demo`. So the imagevolume suite runs only in the gVisor job, and even there it copies the gVisor probe's class.

That leaves the whole micro-VM half of this PR unexercised: `StageImageVolume`, the `<cid>/volumes/<name>` bind under `SharedDir`, `GuestSharedVolumeDir`, virtiofsd announcing it as a submount, and the re-staging that a restore depends on. I read that path and it looks right, so this is a coverage gap rather than a suspected bug — but it is the half where the interesting failure modes live, and it is the reason this fixture exists.

Adding `E2E_SANDBOX_CLASS: microvm` to that step's `env:` is most of it; the suite also has to be in scope for the run, which #795 separately widens from `./internal/e2e/suites/demo` to the full suite.

BenTheElder: nothing sets this, so we're not testing microvm?

we probably need to refactor e2e a bit so we don't depend on CI plumbing a bunch of options, but as a stopgap it seems OK to add this to the actions config.

cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -168,6 +180,35 @@ func waitForSocket(ctx context.Context, path string, timeout time.Duration) erro
 	}
 }
 
+// StageImageVolume bind-mounts one composed image volume read-only at
+// <cid>/volumes/<name> under SharedDir(id), so virtiofsd exposes it to the
+// guest.
+func StageImageVolume(ctx context.Context, src, id, cid, volumeName string) error {
```
Oneimu, line 194: Suggestion on ordering: `StageImageVolume` and #1034's `kata.BindIntoShare` are the same stale-umount/mkdir/rbind skeleton (this one adds the ro remount). Rather than landing two parallel bind-into-share helpers, could we let #1034 merge first and rebase this over it? Then this function becomes `BindIntoShare` + the ro remount (or a small ro option on it), and the tree has one copy of the mount logic.

Oneimu: If your pr merge first I can to do the unification on my side instead.

ronlv10: I'm fine with both ways let's see if it is merged today :)

internal/e2e/probe.go
ronlv10: Ahh I didn't realize that - added

cmd/ateom-microvm/internal/kata/overlay_linux.go
BenTheElder: I chatted with lucky, let's land your PR first and we'll have his PR more generally consolidate virtiofsd

.github/workflows/pr-workflow.yaml
```diff
@@ -103,6 +103,12 @@ jobs:
         E2E_TEMPLATE_NAME: counter-microvm
         E2E_TEMPLATE_READY_TIMEOUT: 600s
       run: hack/run-e2e-kind.sh ./internal/e2e/suites/demo -v -args --no-color
+    - name: Run E2E tests (micro-VM image volumes)
+      # Same image volume e2e suite on the micro-VM runtime.
+      env:
```
BenTheElder, line 107: TODO: factor this out so it's easy to test locally (not exclusive to this PR, it's the established pattern for now)

internal/e2e/probe.go
```diff
@@ -0,0 +1,84 @@
...
+	// from its working directory, which is the test's package dir rather than
+	// the repo root; without it the build silently loses defaultPlatforms and
+	// produces images that cannot run on the cluster's nodes.
+	applyArgs := []string{"ko", "apply", "-f", manifest}
+	if KubeContext != "" {
+		applyArgs = append(applyArgs, "--", "--context="+KubeContext)
+	}
+	RunCmdWithEnv(t, []string{"KO_CONFIG_PATH=" + root}, filepath.Join(root, "hack/run-tool.sh"), applyArgs...)
+
+	t.Cleanup(func() {
```
BenTheElder, line 56: Human comment: I think we should probably stop hardcoding `ate-e2e-probe` as the preferable fix.

-----

🤖 **should-fix** 🟡 – This deletes a fixture two suites now share, and nothing serializes them.

`hack/run-e2e.sh` ends in `go test -v ./internal/e2e/suites/...` with no `-p 1`, so Go runs the suite packages concurrently as separate binaries — the mutexes in `namespace.go` and `testmain.go` only coordinate within one process. Before this PR only `identity` used the probe; now `imagevolume` calls `DeployProbe` too, and both get a cleanup that runs `kubectl delete -f` over the same manifest, namespace `ate-e2e-probe` included.

The likely ordering is the bad one. `imagevolume` only reads the probe's WorkerPool and ActorTemplate to clone them into its own namespace, so it finishes early; `identity` runs its actors *inside* `ate-e2e-probe` for the whole test (`CreateActor`, `SuspendActor`, `ResumeActor`, `/whoami` all against `probeNamespace`). So `imagevolume`'s cleanup deletes the namespace out from under `identity`'s running actors. The reverse order fails too: a re-apply into a Terminating namespace is rejected.

The kind cluster is throwaway, so simply not deleting the shared fixture would do it. Giving `DeployProbe` a namespace argument so each suite gets its own copy is the other direction, and would also let `identity` stop hardcoding `"ate-e2e-probe"` alongside `imagevolume`'s use of `e2e.ProbeNamespace`.

internal/e2e/probe.go
```diff
@@ -0,0 +1,67 @@
...
```
BenTheElder: 🤖 **should-fix** 🟡 – Half of this is fixed: the suites no longer hardcode the literal, and the name is now scoped per sandbox class. But `FixtureName` only appends the class, so within one run `identity` and `imagevolume` still resolve to the same namespace, both `DeployProbe` it, and both register the cleanup above — which deletes the Namespace out from under whichever is still running.

`hack/run-e2e.sh` still ends in `go test -v ./internal/e2e/suites/...` with no `-p 1`, so those two packages run as concurrent binaries; the mutexes in `namespace.go` and `testmain.go` only coordinate inside one process. `imagevolume` just reads the pool and template before cloning them elsewhere, so it finishes first, while `identity` runs `CreateActor`/`SuspendActor`/`ResumeActor` inside the shared namespace for its whole test.

#1056 widened the micro-VM step to the full suite as well, so this now applies to both CI steps rather than only the gVisor one.

Making the returned namespace unique per caller would finish it — `identity` would then have to use the return value instead of recomputing `FixtureName("ate-e2e-probe")` itself, which is the remaining bit of the hardcoding. Not deleting the shared fixture at all is the cheaper alternative, since the kind cluster is thrown away either way.

internal/e2e/probe.go
BenTheElder: #1056 partially resolved this

BenTheElder: we solved between classes, but we still need to deconflict between competing tests in the same sandboxclass, the mutex doesn't cut it since these run as different go test processes.

ronlv10: Done- added a "name" parameter that would be added to suffix env - hopefully it makes sense

BenTheElder (dismissed):

BenTheElder (approved):

PR #931, opened 2026-08-13
atelet: measure what a snapshot image costs, not how long it is

chw120:
atelet_snapshot_size_bytes read fi.Size() for every image it recorded. For the gVisor images that is right, but the micro-VM memory image is sparse: cloud-hypervisor creates memory-ranges at the guest's full RAM size and writes only the pages the guest ever touched. Its length is therefore a constant — the VM's configured memory — so the metric reported the same number for a guest that had just booted and one that had filled its RAM. On a live counter actor it read 2147483648 for a file that occupies 163577856 bytes: the configured 2048 MiB, every time, on every actor.

Record st_blocks*512 instead, which is what the filesystem actually gave the file. Sparse or not, that is the answer to "what did this checkpoint cost", and it is the figure the snapshot's transfer time and storage bill follow.

The count rounds up to a block, so the small JSON images beside the memory one now over-report by under 4 KiB. That is noise on a histogram whose first bucket boundary is 1 MB, and the gVisor images are dense, so for them the two readings agree to within that rounding.

docs/observability.md described this metric as the uncompressed size of a gVisor image, which it was back when gVisor was the only sandbox. It now says what is recorded, and why the two sandboxes read differently.

The existing test helper of the same name is now diskUsage, so the hole-preservation tests keep statting by path rather than going through the code they would otherwise be validating.

- [x] Tests pass


chw120:
The previous snapshot size metric did not consider the micro-VM case, so this update ensures both scenarios are supported. Please let me know if this approach is correct. Thanks. 

cc: @BenTheElder @maxsmythe @HavenXia

cmd/atelet/metrics_test.go
krisztianfekete, line 1: Can we test with something less compressible?

cmd/atelet/main.go
```diff
@@ -468,7 +467,7 @@ func initSnapshotSizeMetric() error {
 	snapshotSizeBytes, err = otel.Meter("atelet").Int64Histogram(
 		"atelet.snapshot.size",
 		metric.WithUnit("By"),
-		metric.WithDescription("Uncompressed size in bytes of each gVisor snapshot image written during checkpoint."),
+		metric.WithDescription("On-disk size in bytes of each snapshot image written during checkpoint: the blocks the filesystem allocated, not the length the file reports. The two differ for the micro-VM memory image, which is sparse — see allocatedBytes."),
```
krisztianfekete: This string here will become the HELP text of the metric and will show up at lots of places, where users cannot look up the Go function. Can you please shorten it to one sentence and move the explanation to docs/observability.md?

krisztianfekete (commented):
Thanks, added a few comments, but looks good overall!

We should probably also wire `atelet.snapshot.size` to `PlatformMetricPrefixes` e2e tests, it can also be a follow-up, though.

cmd/atelet/metrics_test.go
chw120: Changed to use patternBytes().

cmd/atelet/main.go
chw120: Yep shorten it and the docs/observability.md also has the explanation.

chw120:
> Thanks, added a few comments, but looks good overall!
> 
> We should probably also wire `atelet.snapshot.size` to `PlatformMetricPrefixes` e2e tests, it can also be a follow-up, though.

Added "atelet_snapshot_size" into e2e PlatformMetricPrefixes.

chw120:
Thanks @krisztianfekete for review. I have resolved comments. Please review again.

cmd/atelet/main.go
```diff
@@ -561,13 +561,45 @@ func recordSnapshotSize(ctx context.Context, file, path, atNamespace, atName str
...
+//
+// Falls back to the length when the stat carries no Unix shape. No platform
+// atelet runs on is in that case; the fallback exists so a future one degrades
+// to the old reading rather than to zero.
+func allocatedBytes(fi os.FileInfo) int64 {
+	st, ok := fi.Sys().(*syscall.Stat_t)
+	if !ok {
+		return fi.Size()
+	}
+	return int64(st.Blocks) * 512
```
BenTheElder, line 600: 🤖 **should-fix** 🟡 – `st_blocks` counts blocks the filesystem has *allocated*, and on the one image this change exists to measure those blocks are deliberately still unallocated when the stat happens.

Both call sites stat immediately after ateom hands the files back — before the rename in `moveLocalCheckpoint`, before the upload in `uploadSnapshot` — with no fsync anywhere on that path. `MergeDeltaIntoBase` says so outright ("No fsync: atelet ships the merged image to GCS"), and #921 went further, replacing the rename-over-existing with unlink-then-rename precisely so ext4's `auto_da_alloc` writeback *doesn't* fire and the dirty pages stay in page cache for atelet to ship. Under delayed allocation `i_blocks` doesn't count a reserved-but-unwritten extent, so the reading is however much writeback happens to have flushed at that instant — a function of dirty ratio, file size and host load, not of the snapshot.

Your 163577856 against a 2 GiB length says writeback had largely caught up on that run, so this is variance and under-reporting rather than a number that reads zero. Still worth not trading a constant-wrong reading for a load-dependent one.

Walking `SEEK_DATA`/`SEEK_HOLE` and summing the data extents gives the same answer and is immune to it — ext4 reports delalloc extents as data — and `copySparseRegions` in `cmd/ateom-microvm/internal/ch/merge.go` already does exactly that walk.

cmd/atelet/main.go
```diff
@@ -561,13 +561,45 @@ func recordSnapshotSize(ctx context.Context, file, path, atNamespace, atName str
...
 	}
-	snapshotSizeBytes.Record(ctx, fi.Size(), metric.WithAttributes(
+	snapshotSizeBytes.Record(ctx, allocatedBytes(fi), metric.WithAttributes(
 		semconv.FileNameKey.String(file),
 		ateattr.TemplateNamespaceKey.String(atNamespace),
 		ateattr.TemplateNameKey.String(atName),
 	))
 }
 
+// allocatedBytes is what a snapshot image actually costs: the blocks the
```
BenTheElder, line 571: prefer less excessively verbose comments, this is a lot to read and most of it is redundant.

if you reload your session or ask it to apply the AGENTS.md rules from upstream/main directly this should shrink quite a bit.

BenTheElder (commented):
for upload we can just do the compressed size, for pause ... hmm, this is complicated. I don't think the current approach is quite correct either.

thanks for working on this

PR #934, opened 2026-08-13
atepg: deliver worker watch events via a transactional change feed

### LINKED ISSUE #920: db optimization: worker update latency degrades sharply under load due to per-write pg_notify
Worker writes in atepg (`CreateWorker`/`UpdateWorker`/`DeleteWorker`) carry a `pg_notify` inside the write transaction for `WatchWorkers`([code](https://github.com/agent-substrate/substrate/blob/febeb7afe7e720e00f4755c71be1e9b1b43d52cf/cmd/ateapi/internal/store/atepg/atepg.go#L1290)). Postgres serializes all notifying commits through a global lock, which caps worker writes at ~600/s on any instance size — latency is fine at low rates but collapses past the ceiling:

Latency results for workerUpdates when 1000 workers preloaded. 
| worker updates/s | p50 | p99 |
|---|---|---|
| 250 | 4.7 ms | 7.0 ms |
| 500 | 4.7 ms | 16 ms |
| 1,000 | 610 ms | 1,163 ms |

The same update path without the notify (`UpdateActor`) stays flat at 2,000/s (p99 = 6.5 ms), so the write itself is not the problem. This is a known Postgres limitation ([Recall.ai](https://www.recall.ai/blog/postgres-listen-notify-does-not-scale), [DBOS](https://www.dbos.dev/blog/postgres-listen-notify-scalability);. The requirements target O(10K) worker updates/s. NOTIFY's 8 KB payload limit also fails writes outright.

**Proposed fix**

Replace `pg_notify` with a `worker_changes` outbox table written in the same transaction (delivery-iff-commit preserved), and have `WatchWorkers` poll it every 100 ms instead of LISTEN (well within the ≤1 s watch target).
Reference articles: https://microservices.io/patterns/data/transaction-log-tailing.html 
https://oltionzefi.com/en/blog/scaling-outbox-postgres-part-3/

shrutiyam-glitch:
Fixes #920 partially. 

Replaces per-write pg_notify with a worker_changes outbox table written in the same transaction, and LISTEN with a 100ms polling watcher.

#### Motivation
* **Scalability Bottleneck**: `pg_notify` serializes the commits of all notifying transactions through a global lock held across the commit (including `fsync`). This artificially caps worker writes at ~600/s on Cloud SQL regardless of instance size, whereas our target is O(10K) worker updates/s. 
* **Payload Limits**: Bypasses the 8KB `NOTIFY` payload limit that previously caused writes to fail.
* **Reliability**: A cursor-based polling watcher survives reconnects and failovers without missing events, which was a known flaw with the ephemeral `LISTEN` approach.

*(Known Postgres pathology prior art: [Recall.ai](https://www.recall.ai/blog/postgres-listen-notify-does-not-scale), [DBOS](https://www.dbos.dev/blog/postgres-listen-notify-scalability)).*

### Performance Improvement
WorkerUpdate @ 1,000 QPS , 1M workers (preloaded) — before vs after the change feed:
  
| | p50 | p90 | p95 | p99 |
|---|---|---|---|---|
| Before (per-update pg_notify) | 40.3s |55.6s | 61.2s | 63.8s |
| After (change-feed table) | 7.08 ms | 8.04 ms | 8.52 ms | 27.5 ms |

#### Changes Made
* **Schema**: Added transactional outbox table `worker_outbox`.
* **Write Path**: Worker writes now append to the `worker_outbox` feed inside the same transaction instead of calling `pg_notify()`.
* **Watch Path**: Replaced `LISTEN` in `WatchWorkers` with a polling watcher that queries the feed every 50ms.
* **Cleanup**: Implemented a janitor process during polling to periodically delete old feed rows.
* **Tests**: Updated atomicity tests to verify feed inserts instead of `pg_notify` payloads.

For full architecture: https://docs.google.com/document/d/10K0wB6aTeFkJCL4HN3NbLJCdFGoLYdhIcqkFnqFHkKc/edit?tab=t.txqcjwvmhp3v

- [x] Tests pass


jpbetz:
It's good to see the WorkerUpdate numbers come down into the millisecond range. I'm curious how much lower we can get.

Some questions:

1. For the100ms poll interval: each watcher runs its own two-statement batch per tick, so database load scales with watchers * replicas * tick rate even when the feed is idle. Do we have a ballpark for how low the interval can go (or how many concurrent watchers we can carry at 100ms) before polling load itself becomes the problem? No need for an exact number.. just ballpark number (OoM)

2. What does this change cost on the write path itself (create/update)? The before/after table shows the win at 1K qps, but at that load it's mostly measuring the NOTIFY commit-serialization cost. Could we also get a low contention comparisong of per-write latency with the extra feed INSERT vs. the old pg_notify call?

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1462,47 +1466,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+	SELECT EXISTS(
+		SELECT 1 FROM worker_changes_trim
+		WHERE xid > $1::xid8 AND xid > $2::xid8),
+	pg_postmaster_start_time()::text`
+
+// maintainWorkerChangesPartitions is one maintenance pass. Partition
+// creation runs on every replica, unelected — it is idempotent.
+func (p *Persistence) maintainWorkerChangesPartitions(ctx context.Context) error {
+	now := time.Now().UTC()
+	if err := p.createWorkerChangesPartitions(ctx, changeFeedPartitionLeadTimes(now)...); err != nil {
```
jpbetz: I think there's an unrecoverable failure mode in this ordering. If partition creation ever stalls past the lead window (wedged advisory-lock holder, DDL timeouts) while writes continue, strays land in `worker_changes_default` with `created_at` in the current interval. From then on, `CREATE ... PARTITION OF` covering that range fails outright — Postgres refuses to create a range partition when the DEFAULT partition holds rows in that range ("updated partition constraint for default partition ... would be violated"). Since creation runs first here and returns on error, the strays check/truncate below is never reached, so every subsequent pass fails the same way. And `NewPersistence` runs the same creation at startup, so a replica restarted in this state fails to boot — under sustained write traffic this doesn't self-heal and needs manual SQL to clear.

Unless I'm misreading the ordering, checking/truncating the default partition *before* (or on failure of) partition creation would let the system dig itself out. This is the one I'd want resolved before merge.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1234,68 +1255,51 @@ func (p *Persistence) DeleteActorSnapshotTag(ctx context.Context, atespace, name
...
-	maxNotifyPayloadBytes = 8000
-)
-
-type workerEventEnvelope struct {
-	Type   int    `json:"t"`
-	Worker string `json:"w"` // protojson-encoded Worker
-}
-
+// Feed payload format: one event-type byte followed by the binary Worker proto.
+// The tag byte is read by other replicas during rolling deploys, so
```
jpbetz: This covers payload-encoding compatibility across rolling deploys, but I think the *transport* has a deploy-window gap: old binaries publish only via `pg_notify` and watch via `LISTEN`; new binaries write only feed rows and poll. During a rolling update (ate-api-server runs 2 replicas, one at a time), neither side sees the other's events — worker caches in both directions degrade to the 5-minute relist, and the cache feeds scheduling decisions.

A one-release dual-write (keep the `pg_notify` alongside the feed INSERT, drop it next release) would close this cheaply. Also fine as an explicit release-note decision if the window is considered acceptable — I mainly don't want it to be an accident.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1462,47 +1466,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+				if closeErr := br.Close(); err == nil {
+					err = closeErr
+				}
+				if err != nil {
+					if watchCtx.Err() != nil {
+						return
+					}
+					// Transient poll failure: keep the cursor, try again on the next tick. 
+					// If the outage was a restart, the next successful safety check catches it.
+					slog.WarnContext(watchCtx, "worker change feed poll failed", slog.Any("err", err))
```
jpbetz: I think this quietly changes the watch contract: the old LISTEN watcher closed the Events channel when its connection died, which made workercache flip not-ready so callers failed fast. Here a *persistent* poll failure (revoked creds, DB unreachable) warns and retries forever — the channel never closes, workercache stays ready, and consumers serve arbitrarily stale worker state indefinitely (the periodic relist also only warns on failure).

Retrying transient blips is clearly the right call; would it be reasonable to close the watch after N consecutive failures or a failure deadline, so persistent loss still surfaces as not-ready? store.Interface's docs promise close when "the underlying notification system is lost", and I don't think this path ever delivers that anymore.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1462,47 +1466,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+func (p *Persistence) truncateWorkerChangesDefault(ctx context.Context, q querier) error {
+	// Highest xid is recorded as a trim mark in the same transaction so lagging watchers detect the loss and resync.
+	if _, err := q.Exec(ctx, `
+		INSERT INTO worker_changes_trim (xid)
+		SELECT xid FROM worker_changes_default ORDER BY xid DESC LIMIT 1
+		ON CONFLICT (id) DO UPDATE SET xid = EXCLUDED.xid
+		WHERE EXCLUDED.xid > worker_changes_trim.xid`); err != nil {
+		return fmt.Errorf("recording trim mark for feed default partition: %w", err)
+	}
+	if _, err := q.Exec(ctx, `TRUNCATE worker_changes_default`); err != nil {
```
jpbetz: nit (fine as a follow-up issue): I think the trim mark can under-record here. The mark is read at this statement's snapshot, but `TRUNCATE` then waits for ACCESS EXCLUSIVE behind in-flight feed inserts — a row committing in that window is destroyed with an xid *above* the recorded mark, so a caught-up watcher's fell-behind check won't fire and that event is silently lost until the consumer's relist. Only reachable while partition creation is already stalled (which is when this path runs), but that's exactly when loss detection matters most. An explicit `LOCK TABLE worker_changes_default IN ACCESS EXCLUSIVE MODE` before reading the mark would close the window.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1462,47 +1466,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+// If the watcher detects missed events—either by lagging behind retention drops
+// or if a database restart truncates the UNLOGGED partitions—it closes the channel
+// to force the consumer to resync from the primary tables.
+func (p *Persistence) WatchWorkers(ctx context.Context) (*store.WorkerWatch, error) {
+	watchCtx, cancel := context.WithCancel(ctx)
+
+	// cursorXid starts at xmin - 1. The minus 1 is required because the poll
+	// query is exclusive (xid > cursor); starting exactly at xmin on an idle
+	// system would skip the very next event.
+	// baselineXid records the highest xid or trim mark at subscribe time so
```
jpbetz: nit (follow-up ok): I *think* the baseline can mask a real loss in one corner. `baselineXid` is a committed-visibility max, but a transaction that acquired a lower xid before this watch subscribed and commits after it is still a legitimate post-subscribe event. If that row is later trimmed while fenced behind a long transaction, the dropped partition's greatest xid (the recorded mark) can be <= baseline, so the `xid > baseline` clause suppresses the fell-behind close and the event is skipped silently. Deriving the baseline from the same xmin snapshot as the cursor, rather than the max visible feed xid, would avoid the case — unless the current form is deliberately looser to avoid spurious resyncs on fresh watches?

jpbetz:
One thing I noticed when testing this out is that the polling watchers are competing for the same connection pool as normal write traffic. We should consider splitting those out into dedicated pools to avoid one starving the other.

shrutiyam-glitch:
> 1. For the100ms poll interval: each watcher runs its own two-statement batch per tick, so database load scales with watchers * replicas * tick rate even when the feed is idle. Do we have a ballpark for how low the interval can go (or how many concurrent watchers we can carry at 100ms) before polling load itself becomes the problem? No need for an exact number.. just ballpark number (OoM)

So, I think there is only one watcher per replica — `workercache` is the sole `WatchWorkers` consumer; everything else reads the in-memory cache. Assuming 10 replicas, the polling load on the database would be: 
`polls/s = replicas ÷ poll interval`(today: 10 ÷ 100ms = `~100 polls/s`)

Assuming an idle feed, each poll is one batched round trip carrying two statements: the cursor query (`xid > $cursor` probe against the 2–3 live partitions' xid indexes — an empty result touches ~3–4 buffers, no disk I/O) and the safety query (a single-row table read plus `pg_postmaster_start_time()`). _Estimated_ cost is roughly `0.1–0.3ms` of backend CPU per poll, no I/O. Now, polls/s × per-poll cost against the instance's CPU:
  - ~100 polls/s (today: 10 replicas @ 100ms) → ~10–30ms of CPU per second, i.e. ~1–3% of one core. Polling is noise.
  - ~1,000 polls/s (100 replicas @ 100ms, or 10 replicas @ 10ms) → ~20% of one core. Visible in profiles, still fine.
  - ~10,000 polls/s (a 10–100× move on either axis) → ~1–2 whole cores doing nothing but empty probes. At this point polling is a real tenant on the instance, competing with the actual write workload. 
 
So, the estimates in both directions: 
  - How many watchers at 100ms: O(100) replicas maybe is comfortably fine; 
  - How low can the interval go: fleet-wide at today's replica count (10), probably ~10ms before polling load becomes comparable to our real write load. 
 Wdyt ? 


> 2. What does this change cost on the write path itself (create/update)? The before/after table shows the win at 1K qps, but at that load it's mostly measuring the NOTIFY commit-serialization cost. Could we also get a low contention comparisong of per-write latency with the extra feed INSERT vs. the old pg_notify call?

Fair point — at 1,000 QPS was tested in an open-loop harness calling the interface methods directly, not pause/suspend/resume cost. Will run the locust benchmarks by mocking the `atelet`.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1275,47 +1324,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+	SELECT EXISTS(
+		SELECT 1 FROM worker_changes_trim
+		WHERE xid > $1::xid8 AND xid > $2::xid8),
+	pg_postmaster_start_time()::text`
+
+// maintainWorkerChangesPartitions is one maintenance pass. Partition
+// creation runs on every replica, unelected — it is idempotent.
+func (p *Persistence) maintainWorkerChangesPartitions(ctx context.Context) error {
+	now := time.Now().UTC()
+	if err := p.createWorkerChangesPartitions(ctx, changeFeedPartitionLeadTimes(now)...); err != nil {
```
jpbetz: AI review comment that looks valid and important to address. This would be hard to recover from it it were to happen:

"I think there's a self-wedging order here: if partition creation ever stalls past the lead window while writes continue, strays land in the DEFAULT partition with current-interval timestamps — and from then on `CREATE ... PARTITION OF` for that range fails outright (Postgres won't carve a range the DEFAULT already holds rows in). Creation runs first and returns on error, so the truncate-strays rescue below is never reached; and `NewPersistence` runs the same creation at boot, so a restarted replica won't come up either. I don't think this self-heals without manual SQL. Would checking/truncating the DEFAULT *before* (or on failure of) creation let it dig itself out? This is the one I'd want sorted before merge — we're reproducing it in a fault drill this week to confirm."

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1275,47 +1324,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+				if closeErr := br.Close(); err == nil {
+					err = closeErr
+				}
+				if err != nil {
+					if watchCtx.Err() != nil {
+						return
+					}
+					// Transient poll failure: keep the cursor, try again on the next tick.
+					// If the outage was a restart, the next successful safety check catches it.
+					slog.WarnContext(watchCtx, "worker change feed poll failed", slog.Any("err", err))
```
jpbetz: AI review comment that looks valid here:

"I think this quietly changes the watch contract: the LISTEN watcher closed `Events` when its connection died, so workercache flipped not-ready and callers failed fast. Here a *persistent* failure (revoked creds, DB unreachable) warns and retries forever — the channel never closes and consumers serve stale state indefinitely. Retrying blips is clearly right; would closing after N consecutive failures (or a deadline) be reasonable so persistent loss still surfaces? store.Interface's docs still promise a close when the notification system is lost. (FWIW, the close→relist cost this guards is ~2.1s of not-ready at 1M workers, ~30ms at 10k — measured.)"

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1275,47 +1324,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+			}
+			// Drain until a batch is partial. Sleeping between full batches would
+			// cap throughput and cause unrecoverable lag during bursts.
+			for {
+				// Safety checks share the batch round trip but must remain separate
+				// queries so we can detect gaps and restarts even when no rows match.
+				b := &pgx.Batch{}
+				b.Queue(pollWorkerChangesSQL, cursorXid, changeFeedBatch)
+				b.Queue(pollSafetySQL, cursorXid, baselineXid)
+				br := p.watchPool.SendBatch(watchCtx, b)
```
jpbetz: The dedicated watch pool is a great addition — on the previous revision we measured the shared-pool cost at ~0.94ms per write op at 12k writes/s (~30% of service time), and this should erase it; we'll re-verify on this head. Two curiosities though: (1) the retention transaction and partition DDL now also run on this 2-conn pool — can a slow maintenance pass (e.g. the DROP waiting on ACCESS EXCLUSIVE) starve the poller for its tick, and is that intentional coupling? (2) the *writer* pool is still bare `pgxpool.New`, so production MaxConns silently defaults to the client's core count — worth pinning or documenting?

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1275,47 +1324,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+	ident := pgx.Identifier{name}.Sanitize()
+	// The mark is the partition's greatest xid.
+	if _, err := q.Exec(ctx, fmt.Sprintf(`
+		INSERT INTO worker_changes_trim (xid)
+		SELECT xid FROM %s ORDER BY xid DESC LIMIT 1
+		ON CONFLICT (id) DO UPDATE SET xid = EXCLUDED.xid
+		WHERE EXCLUDED.xid > worker_changes_trim.xid`, ident)); err != nil {
+		return fmt.Errorf("recording trim mark for feed partition %s: %w", name, err)
+	}
+	if _, err := q.Exec(ctx, `DROP TABLE `+ident); err != nil {
```
jpbetz: An AI review found this and it looks right to me:

"This DROP takes ACCESS EXCLUSIVE on the parent, briefly blocking every write's feed append, every 15 min. Would `ALTER TABLE ... DETACH PARTITION CONCURRENTLY` + dropping the detached table avoid that? (We're measuring the spike under load — it may turn out negligible, in which case ignore me.)"

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1275,47 +1324,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+// WatchWorkers subscribes by polling the worker_changes feed using an xid cursor.
+// It fences reads behind pg_snapshot_xmin (the oldest in-flight transaction)
+// to guarantee gap-free delivery. Note that a long-running transaction anywhere
+// in the database will stall this feed.
+//
+// Events are delivered in xid order, so consumers must reconcile worker versions.
+// If the watcher detects missed events—either by lagging behind retention drops
+// or if a database restart truncates the UNLOGGED partitions—it closes the channel
+// to force the consumer to resync from the primary tables.
+func (p *Persistence) WatchWorkers(ctx context.Context) (*store.WorkerWatch, error) {
```
jpbetz: Follow-up sized, not blocking: feed health is invisible from outside — no cursor-lag/backlog gauge, no counter for closes by cause (trim/restart), and an xmin-fence stall (any long-running tx anywhere in the DB) just looks like silence until consumers notice. A cursor-age metric plus close counters would make operating this much saner.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1275,47 +1324,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+	}
+	return nil
+}
+
+// truncateWorkerChangesDefault discards the DEFAULT partition wholesale to
+// un-stall partition creation.
+func (p *Persistence) truncateWorkerChangesDefault(ctx context.Context, q querier) error {
+	// Highest xid is recorded as a trim mark in the same transaction so lagging watchers detect the loss and resync.
+	if _, err := q.Exec(ctx, `
+		INSERT INTO worker_changes_trim (xid)
```
jpbetz: Add an explicit `LOCK TABLE worker_changes_default IN ACCESS EXCLUSIVE MODE` before reading here?

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1275,47 +1324,352 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
...
+	// cursorXid starts at xmin - 1. The minus 1 is required because the poll
+	// query is exclusive (xid > cursor); starting exactly at xmin on an idle
+	// system would skip the very next event.
+	// baselineXid records the highest xid or trim mark at subscribe time so
+	// past garbage collection isn't mistaken for events lost during this watch.
+	// Xids are passed as decimal strings end-to-end; all ordering happens in SQL.
+	var cursorXid, baselineXid, baselineStart string
+	if err := p.watchPool.QueryRow(watchCtx, `
+		SELECT (pg_snapshot_xmin(pg_current_snapshot())::text::numeric - 1)::text,
+		       GREATEST(
```
jpbetz: One last AI comment that I think we should address:

"nit: I *think* there's a corner where a tx that took its xid before this watch subscribed but commits after it is a legit post-subscribe event — yet if it's later trimmed while fenced, the dropped partition's max xid can be ≤ this baseline and the fell-behind close never fires. Deriving the baseline from the same xmin snapshot as the cursor would avoid it — unless the looser form is deliberate to dodge spurious resyncs on fresh watches?"

shrutiyam-glitch: done. 
Updated the `baselineXid` query at subscribe time to strictly query `WHERE` `xid < pg_snapshot_xmin()`. By deriving the baseline from the exact same `xmin` fence as the cursor, we guarantee it never artificially inflates the baseline with an already-committed `xid` that is > `xmin`.

cmd/ateapi/internal/store/atepg/atepg.go
shrutiyam-glitch: Done. Thanks !

cmd/ateapi/internal/store/atepg/atepg.go
shrutiyam-glitch: ` DETACH PARTITION CONCURRENTLY` - added significant orchestration complexity for a marginal gain. Instead, optimized the maintenance loop by splitting partition drops into their own dedicated, isolated transaction.

cmd/ateapi/internal/store/atepg/atepg.go
shrutiyam-glitch: done. Added a `failingSince` tracker with a 30-second deadline (`outboxPollFailureCloseAfter`). Transient failures still retry gracefully, but if the poll fails persistently past the deadline, we now close the watch channel. This properly flips `workercache` to not-ready so callers fail fast.

cmd/ateapi/internal/store/atepg/atepg.go
shrutiyam-glitch: Fixed. When `CREATE PARTITION` fails when `DEFAULT` holds rows in the target range, we now catch the error, `TRUNCATE` the `DEFAULT` partition strays (which updates the trim mark to trigger watcher resyncs), and re-run `CREATE PARTITION` inside the same transaction.

jpbetz:
This LGTM

Of the available change capture mechanisms, this approach appears to make the best trade-off of complexity vs. performance for substrate's needs, as I understand them.  This can achieve as little as 5ms median latency for events at modest compute overhead, scales well to over 10k writes/s with negligible impact on write latency. There is complexity in the management of the outbox (deleting partitions, ...) and there are limitations because the change capture writes are not written to the WAL. But this limitation is allowed by the design, so this appears OK.

PR #935, opened 2026-08-13
atelet: stage local checkpoints by hard link instead of copying

Jefftree:
Resuming from a node-local pause checkpoint copied the whole snapshot into the restore dir, so the node held two copies of the same guest memory image and paid a full write to make the second. Both dirs sit under the same actor dir, so link instead, falling back to a copy across filesystems.

Linking makes the staged image share an inode with the cached snapshot, which `MergeDeltaIntoBase` would otherwise overlay in place and corrupt. Two changes keep that safe. Earlier pause snapshots are now pruned before the checkpoint rather than after, which releases the second link while the staged image still holds the inode, so the merge keeps its cheap in-place path. And the merge refuses that path outright when base still carries a second link, copying instead, as a backstop for any ordering it cannot see.

Counter demo on kind, 2 GiB guest with ~133 MiB populated:

| | before | after |
|---|---|---|
| resume, staging the image | 105 ms | under 1 ms |
| suspend, merge | 16 ms | 14 ms |

The merge figure is from the `Merged OnDemand delta into base` log line on a live cluster across a pause, resume, pause cycle, confirming the prune lands before the merge reads the link count. Without the reordering the same merge costs ~130 ms.

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -577,20 +577,6 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 		return nil, err
 	}
 
-	// Drop earlier pause snapshots BEFORE checkpointing, not after. No earlier one
-	// can be restored again: the control plane tracks a single local snapshot, which
-	// this checkpoint either overwrites (pause) or clears (suspend). Ordering it here
-	// matters because restore staged this actor's memory image by linking to that
-	// snapshot, and MergeDeltaIntoBase can only take its cheap in-place path while
-	// the staged image is the sole name for those bytes. Releasing the other name
-	// first leaves the image untouched (the staged link still holds the inode) and
-	// keeps the merge off the copying path.
-	//
-	// The cost is that a checkpoint failing from here on leaves no earlier snapshot
-	// to fall back to. That is survivable: the guest stays paused when
-	// CheckpointWorkload fails, so the checkpoint can simply be retried.
-	pruneLocalCheckpoints(ctx, actorUID)
-
 	// Tell ateom to take the checkpoint and delete containers. ateom reports the
 	// exact files it wrote so we ship precisely that set (gVisor's image files,
 	// cloud-hypervisor's snapshot set, ...) rather than a hardcoded list.
@@ -624,6 +610,22 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	sandboxRec.ActorTemplateNamespace = req.GetActorTemplateNamespace()
 	sandboxRec.ActorTemplateName = req.GetActorTemplateName()
 	sandboxRec.Scope = ateattr.SnapshotScopeValue(req.GetScope())
+
+	// No earlier pause snapshot can ever be restored again, so remove them
+	// all: the actor's current state was just captured by CheckpointWorkload,
+	// and the control plane tracks only a single local snapshot, which this
+	// checkpoint either overwrites (pause) or clears (suspend).
+	//
+	// Do not move this above CheckpointWorkload to keep MergeDeltaIntoBase on its
+	// in-place path: that leaves the whole checkpoint window with no local snapshot
+	// while LocalSnapshotInfo still names the pruned one, and a crash there strands
+	// the actor for good (resume never falls back to object storage, RequiredNodes
+	// pins it to this node, nothing clears the field).
+	//
+	// Pruning stays outside the persist window: it collects superseded
+	// snapshots on both paths, so timing it as part of an external upload would
+	// mix local disk deletion into the object-storage measurement.
+	pruneLocalCheckpoints(ctx, actorUID)
 
 	tPersist := time.Now()
 	switch req.GetType() {
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -1163,14 +1163,24 @@ func (s *AteomHerder) copyLocalCheckpoint(ctx context.Context, snapshotName stri
 		dst := filepath.Join(dstDir, fileName)
 		// Link rather than copy. The local checkpoint lives under the same actor dir
 		// as the restore staging area, so this stages the memory image in constant
-		// time instead of re-writing its whole working set. Nothing mutates the
-		// staged file: CH demand-pages from it read-only, and MergeDeltaIntoBase
-		// refuses its in-place overlay once the image carries a second link, so the
-		// cached snapshot cannot be rewritten through the staged name. Falls back to
-		// copying if the two ever land on different filesystems.
-		if err := os.Link(src, dst); err == nil {
+		// time instead of re-writing its whole working set. Nothing rewrites the
+		// shared inode: CH demand-pages from the staged image read-only,
+		// rewriteSnapshotSocketPaths renames its rewritten config.json into place
+		// rather than truncating, and MergeDeltaIntoBase refuses its in-place overlay
+		// once the image carries a second link.
+		//
+		// EXDEV alone falls back to copying, so an unexpected link failure surfaces
+		// instead of silently reverting to the full copy this exists to remove. It
+		// also keeps copyFile off a dst that is already a link to src, where its
+		// O_TRUNC would empty both and report a successful copy of the old size.
+		switch err := linkFile(src, dst); {
+		case err == nil:
 			continue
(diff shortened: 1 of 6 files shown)


cmd/atelet/main.go
```diff
@@ -1156,6 +1161,16 @@ func (s *AteomHerder) copyLocalCheckpoint(ctx context.Context, snapshotName stri
 		}
 		src := filepath.Join(srcDir, snapshotName, fileName)
 		dst := filepath.Join(dstDir, fileName)
+		// Link rather than copy. The local checkpoint lives under the same actor dir
+		// as the restore staging area, so this stages the memory image in constant
+		// time instead of re-writing its whole working set. Nothing mutates the
+		// staged file: CH demand-pages from it read-only, and MergeDeltaIntoBase
```
BenTheElder: 🤖 **should-fix** 🟡 – True for `memory-ranges`, not for the rest of the set. `copyLocalCheckpoint` links every file in `SnapshotFiles`, and `rewriteSnapshotSocketPaths` (`cmd/ateom-microvm/restore.go:180`) rewrites `restore-state/config.json` with `os.WriteFile`, which truncates and writes through the link into the cached snapshot. The `Nlink` guard added in this PR covers the memory image; nothing covers this one.

It is benign today only because a local snapshot is always restored by the actor that took it, so the rewritten socket paths come out equal. A crash mid-write still leaves the actor's only local restore point holding a truncated `config.json`.

Writing that config atomically — temp file in the same dir, then rename — breaks the link instead of writing through it, and closes the crash window as well.

cmd/atelet/main.go
```diff
@@ -1156,6 +1161,16 @@ func (s *AteomHerder) copyLocalCheckpoint(ctx context.Context, snapshotName stri
...
+		// refuses its in-place overlay once the image carries a second link, so the
+		// cached snapshot cannot be rewritten through the staged name. Falls back to
+		// copying if the two ever land on different filesystems.
+		if err := os.Link(src, dst); err == nil {
```
BenTheElder: 🤖 **should-fix** 🟡 – Any `os.Link` failure falls through to a full copy with nothing logged, so the optimization this PR exists for can stop working silently. `MergeDeltaIntoBase` already has the narrower idiom for its own cross-filesystem fallback: `errors.Is(err, unix.EXDEV)`.

Narrowing it also removes a footgun. If `dst` ever exists as a link to `src`, `copyFile`'s `os.Create` truncates the shared inode and then copies from the now-empty source, returning the pre-truncate size and a nil error. `resetActorDirs` is the only thing making that unreachable today.

cmd/ateom-microvm/internal/ch/merge.go
```diff
@@ -128,6 +130,16 @@ func MergeDeltaIntoBase(ctx context.Context, baseFile, deltaFile string) error {
 		return fmt.Errorf("MergeDeltaIntoBase: size mismatch base=%d delta=%d", bi.Size(), di.Size())
 	}
 
+	// The fast path below MUTATES baseFile's inode in place. That is only safe while
+	// baseFile is the sole name for it. A second link means the bytes are shared with
+	// something outside this actor (atelet stages restore-state by linking from its
+	// local snapshot cache rather than copying), and overlaying would corrupt every
+	// other name for that inode, including the golden snapshot other actors restore
```
BenTheElder: 🤖 **nit** 🟢 – No cross-actor sharing exists to protect here. `os.Link` appears once outside `vendor/` — the call this PR adds — and it links from the actor's own `LocalCheckpointsDir(actorUID)`, so the only second name for those bytes is that same actor's cached pause snapshot. Golden images arrive by download, on a path that never links. Naming the actual sharer would point a reader at the right place.

Jefftree: Reworded to name the actor's own cached pause snapshot.

cmd/atelet/main.go
Jefftree: Narrowed it to EXDEV, so any other link failure is now an error rather than a silent copy. You're right that resetActorDirs is the only thing keeping the truncate unreachable, so I put a test on that too.

cmd/atelet/main.go
Jefftree: Confirmed, the test fails against the old os.WriteFile with the cached snapshot's config.json coming back pointed at the restoring actor. Switched to temp file plus rename, and fixed the claim in this comment.

cmd/atelet/main.go
```diff
@@ -577,6 +577,20 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
...
+	// matters because restore staged this actor's memory image by linking to that
+	// snapshot, and MergeDeltaIntoBase can only take its cheap in-place path while
+	// the staged image is the sole name for those bytes. Releasing the other name
+	// first leaves the image untouched (the staged link still holds the inode) and
+	// keeps the merge off the copying path.
+	//
+	// The cost is that a checkpoint failing from here on leaves no earlier snapshot
+	// to fall back to. That is survivable: the guest stays paused when
+	// CheckpointWorkload fails, so the checkpoint can simply be retried.
+	pruneLocalCheckpoints(ctx, actorUID)
```
BenTheElder: 🤖 **question** 🟢 – The stated mitigation covers an RPC failure but not a crash, and the reordering widens the window where neither snapshot exists.

A gap between the prune and `moveLocalCheckpoint` was already there, but it used to be short. It now spans the whole `CheckpointWorkload` call — pause, write the memory image, tear down — which is the expensive part. Through all of it the node holds no local snapshot while the actor's `LocalSnapshotInfo` still names the pruned one.

"The checkpoint can simply be retried" holds when the RPC returns an error, because the guest is still paused. It doesn't hold if atelet or the node dies mid-snapshot: nothing retries, and the actor is left pinned by `RequiredNodes` to a node whose local snapshot has been deleted. Before this change the same crash left the earlier snapshot intact and the resume worked.

Keeping the old snapshot isn't free — the merge would take the copying path at roughly 130ms instead of 14ms — so trading that for a wider crash window may well be the right call. Worth saying so in the comment, though, since it currently reads as if a retry always covers the cost.

BenTheElder: Human: We would _probably_ prefer to keep the disk fuller than lose reliability.

Jefftree: Yeah this is a tradeoff losing reliability for improved performance. It's also a suspend only optimization and resume seems more important. Removed the early pruning.

cmd/ateom-microvm/internal/ch/merge.go
```diff
@@ -128,6 +130,18 @@ func MergeDeltaIntoBase(ctx context.Context, baseFile, deltaFile string) error {
 		return fmt.Errorf("MergeDeltaIntoBase: size mismatch base=%d delta=%d", bi.Size(), di.Size())
 	}
 
+	// The fast path below MUTATES baseFile's inode in place, which is only safe while
+	// baseFile is the sole name for it. atelet stages restore-state by linking from
+	// this actor's local pause snapshot rather than copying it, so a second link means
+	// the overlay would rewrite that cached snapshot too, corrupting the actor's only
+	// local restore point. Copy in that case, which is what MergeSparseOverlay does.
+	// atelet holds that snapshot for the whole checkpoint, so this is the normal path
+	// for a locally staged restore; the in-place path below is for one staged from
```
BenTheElder: 🤖 **should-fix** 🟡 – This comment is now the accurate one, and the PR description contradicts it.

Dropping the early prune means the cached snapshot holds its link for the whole checkpoint, so `Nlink > 1` is true on every locally staged restore and the copying merge is the path that ships. The description still says the opposite — "Earlier pause snapshots are now pruned before the checkpoint rather than after ... so the merge keeps its cheap in-place path" — and its table reports the merge at 14 ms, the figure measured with the reordering. By the description's own numbers the shipped path is the ~130 ms one.

That matters beyond tidiness, because the description is what lands as the merge commit and it currently hides the trade the PR now makes: resume gains ~104 ms on staging, suspend loses ~115 ms on the merge. Your call that resume is the latency that counts is a reasonable one, and Ben's "keep the disk fuller" was about disk rather than about a slower suspend — so it reads better as a stated decision than as a side effect of the revert.

Worth rewriting the description around what the code does now, and either re-measuring the merge row or dropping it.

BenTheElder:
cc @chw120 @msau42

PR #938, opened 2026-08-13
[WIP] atenet/dns: publish the router's IPv6 ClusterIP as an AAAA

ygao-g:
Depends on #874, whose commit appears in this diff until it merges.

**Actor names have no IPv6 address.** Every actor name resolves to the same
place — the `atenet-router` ClusterIP, with per-actor demux at Envoy on the
Host header — so the zone now publishes an address record per family the
router has a ClusterIP in. That also fixes a hard failure on IPv6-only
clusters, where the router's only ClusterIP is a v6 address published as an
`IN A`: every `A` query for an actor name SERVFAILs there today, for every
client, not just musl.

**Ordering against #911.** Prefer landing after it, but the dependency is
soft — without #911 the router Service is SingleStack, so no dual-stack
cluster has a v6 ClusterIP to publish and the rendered zone is unchanged. The
case to avoid is #911 landing split: a dual-stack Service without the `::`
listener publishes an AAAA nothing is bound to, and Happy Eyeballs takes the
working IPv4 path down with it.

Part of #246.

### DIFF
--- a/internal/ipfamily/ipfamily.go
+++ b/internal/ipfamily/ipfamily.go
@@ -1,0 +1,60 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


internal/ipfamily/ipfamily.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2026 Google LLC
```
bowei: we don't need this as a at the root library -- can you put this in the cmd/atenet/internal/dns package

bowei:
DNS was removed

PR #939, opened 2026-08-13
[WIP] ci: add an IPv6-only kind e2e job for both sandbox classes

ygao-g:
**Do not merge until #958 and #1275 land.** The micro-VM leg also needs #1122.

Part of #246.

No CI job exercises IPv6-only, so nothing catches a change that breaks it, and
the IPv6 work is a set of changes that do little apart. This job builds a
single-stack IPv6 kind cluster, installs the full system, and runs the
networking suite — where every test that can tell one address family from
another lives — on both sandbox classes off the one cluster, asserting the
cluster really is single-stack before anything is installed so a green run
cannot be vacuous. It carries no copy of the code it exercises: dispatch it
with a list of pull requests and it merges them in that order, and the stack
lives only for the length of the run.

Both classes, because that is where address-family support diverges: runsc
adopts the interior netns wholesale and picks the actor's addresses up for
free, while a micro-VM's `eth0` is a tap cross-connected to that netns at L2,
so the guest has to be told over the kata-agent channel. A lane that ran gVisor
alone could not tell a working micro-VM guest from one that never got an
address. The second commit adds that leg, under `always()` so a gVisor failure
cannot hide the micro-VM result. It stays red until the guest is given an IPv6
address, which is the defect it exists to catch.

A fresh IPv6-only cluster resolves nothing from a pod, and a runner has no IPv6
egress of its own, so both the cluster's DNS (#958) and the translator it needs
(#1275) come from `hack/`; this job asks for them and reads back what it got.
Setting that up here instead would have meant the job supplying the very thing
it exists to gate. The cost is that against main alone it stops at the NAT64
step, including on the `pull_request` path that lets this file test itself,
until those land. The job stays out of the `e2e-test` gate either way, so it
never blocks a pull request.

Demonstrated on a run of this exact file with the IPv6 stack and a not-yet-filed
micro-VM guest fix merged in — 20 passed, 0 skipped on each class
([run](https://github.com/ygao-g/substrate/actions/runs/33575103224)). Its
control is the same lane without that fix: gVisor 19/0, micro-VM 15/0 with the
four egress tests red
([run](https://github.com/ygao-g/substrate/actions/runs/33449844845)).

### DIFF
--- a/.github/workflows/e2e-ipv6.yaml
+++ b/.github/workflows/e2e-ipv6.yaml
@@ -13,13 +13,21 @@
 # limitations under the License.
 
 name: e2e-ipv6
-# Separate from pr-workflow.yaml so this can be gated independently -- and so a
-# 40-minute IPv6 run never delays that workflow's merge-gating jobs.
+# Separate from pr-workflow.yaml so this can be gated independently -- and so an
+# hour-long IPv6 run never delays that workflow's merge-gating jobs.
 #
 # Only the networking suite runs. Every test that can tell one address family
 # from another lives in it, and the demo suite is a large share of the wall
 # clock for coverage that is not family-specific. The demos still deploy,
 # because networking builds its actors from their templates.
+#
+# Both sandbox classes run, gVisor then micro-VM, off the one cluster and the
+# one E2E_SANDBOX_CLASS knob. The class is where address-family support
+# diverges: gVisor's runsc adopts the interior netns wholesale and picks the
+# actor's IPv6 up for free, while a micro-VM's eth0 is a tap cross-connected to
+# that netns at L2, so the guest has to be told over the kata-agent channel. A
+# lane that ran gVisor alone could not tell a working micro-VM guest from one
+# that never got an address.
 #
 # Two ways in. A pull request that edits this file runs it: the paths filter is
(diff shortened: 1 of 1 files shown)


ygao-g:
Dispatch record for this lane. The runs are on the fork; a fork pull request cannot dispatch a workflow upstream.

| `prs` | outcome |
|---|---|
| `958 753 1080 979 1060 1116 1057 1275` | 11 passed, 0 skipped — [33132033499](https://github.com/ygao-g/substrate/actions/runs/33132033499), and again on a later main in [33187917892](https://github.com/ygao-g/substrate/actions/runs/33187917892) |
| `1275` | 7 passed; `TestActorEgress`, `TestActorEgressHTTPS`, `TestActorEgressGRPC` and `TestActorEgressNonStandardPort` fail — [33129489452](https://github.com/ygao-g/substrate/actions/runs/33129489452) |

#1275 on its own brings the cluster up and leaves exactly those four failing, so everything else in the stack is what those four measure. Nothing bisects the eight, so the minimal green set is unmeasured.

The two green runs used a byte-identical `e2e-ipv6.yaml` on different bases, so the pass is not an artifact of one main.

Issue #901, opened 2026-08-12
Define postresql schema update strategy


juli4n:
In #731 we added a PostgreSQL based backend to replace ValKey. We need to define how Substrate will keep track of DDL commands to evolve the SQL schema, and the machinery to run / apply the schema changes during upgrades in a reliable way.

EItanya:
@iplay88keys assigning you.

It won't let me officially do it...

iplay88keys:
Edit: The proposal doc is now at: https://docs.google.com/document/d/13ixDKRoAIFXeLxS-_1nikNcAy76ca8m0eVgobfib93E/edit?usp=sharing (need to be part of the ate-dev@googlegroups.com google group to access)

maxsmythe:
I'm in ate-dev, but no access?

iplay88keys:
>I'm in ate-dev, but no access?

@maxsmythe  I'm looking into this. It seems like an org-level sharing policy that I don't have visibility into. I'm working with my IT team to resolve it. Can you request access in the meantime?

iplay88keys:
I have a PR up for this here: https://github.com/agent-substrate/substrate/pull/1196

iplay88keys:
@juli4n @maxsmythe In order to resolve the share permissions, I had to move the doc. The new link is [here](https://docs.google.com/document/d/13ixDKRoAIFXeLxS-_1nikNcAy76ca8m0eVgobfib93E/edit?usp=sharing).

PR #940, opened 2026-08-13
Replace ateredis with atepg

supreme-gg-gg:
## Summary

A follow up to #640 where we introduced PostgreSQL as an alternative storage backend, selected conditionally in ateapi.

- Deleted ateredis, its tests, and its dependencies
- Removed Redis backend selection and configuration so ateapi always connects to Postgres
- Replaced Valkey resources with Postgres in the standard and Kind deployment paths and simplified install script
- Replaced miniredis fixtures with isolated Postgres testcontainers and added centralized helpers for seeding resources
- Renamed Redis-specific debug flush command to backend-neutral `debug-clear-store` in CLI
- Updated comments and docs where applicable

## Benchmarking

Extensive benchmarking have been performed to evaluate Redis vs Postgres, and results can be found in these two documents:

- https://docs.google.com/document/d/10K0wB6aTeFkJCL4HN3NbLJCdFGoLYdhIcqkFnqFHkKc/edit?usp=sharing
- https://docs.google.com/document/d/12-ko_BFHcBo_nJkx9f4B7zMbiiWKC2saGhMhZG3aQ-s/edit?usp=sharing


cmd/ateapi/main.go
```diff
@@ -65,15 +62,9 @@ var (
 	metricsListenAddr    = pflag.String("metrics-listen-addr", ":9090", "Address and port the prometheus metrics server should listen on.")
 	grpcServerCredBundle = pflag.String("grpc-server-cred-bundle", "", "File with the server TLS credential bundle.")
 
-	redisClusterAddress = pflag.String("redis-cluster-address", "", "The address of the redis cluster.")
-	redisCACerts        = pflag.String("redis-ca-certs", "", "The file that contains the CA certificate for Redis cluster.")
-	redisUseIAMAuth     = pflag.String("redis-use-iam-auth", "true", "Whether to use Google IAM authentication for Redis/Valkey.")
-	redisTLSServerName  = pflag.String("redis-tls-server-name", "", "The ServerName to use for Redis TLS hostname verification.")
-	redisClientCert     = pflag.String("redis-client-cert", "", "The file containing client TLS certificate/key credential bundle for Redis/Valkey.")
-
```
EItanya, line 74: We need to backfill authentication to postgres. Can we do that now?

EItanya (commented):
Very large PR, but so far this is my main comment

cmd/ateapi/main.go
supreme-gg-gg: Authentication and mTLS for Postgres has already been added in #640 see 

https://github.com/agent-substrate/substrate/blob/73aa9824c2e0020bcf07dbe20379855d875a401f/hack/install-ate.sh#L187

https://github.com/agent-substrate/substrate/blob/73aa9824c2e0020bcf07dbe20379855d875a401f/manifests/ate-install/postgres.yaml#L21-L27

docs/roadmap.md
```diff
@@ -79,7 +78,7 @@ Below is a collection of finer-grained efforts which we believe align with the a
 * Provisioning load test and benchmarking compute/infrastructure
 * Storage and visualization for benchmark results
 * Integrate debugging into load tests
-* State Store Scale: Horizontal sharding support (via Redis Hash Tags) to enable management of 1M+ concurrent actors.
+* State Store Scale: PostgreSQL scaling and partitioning support to enable management of 1M+ concurrent actors.
```
jpbetz, line 81: The startup DDL is CREATE TABLE IF NOT EXISTS and is not going to be sufficient long term.

Reliable detection of the DDL state to ensure compatibility with the binary will be needed early, and a upgrade mechanism will need to be introduced to allow for future schema updates.  Fix now or track with an issue?

cmd/ateapi/main.go
```diff
@@ -316,114 +297,17 @@ func logFlagValues(ctx context.Context) {
...
+	if *postgresConnectionString == "" {
+		return nil, fmt.Errorf("--postgres-connection-string is required")
 	}
-}
-
-// connectRedis builds the Redis/Valkey TLS config, plumbs IAM auth if
-// requested, opens the cluster client, and pings with retries.
-func connectRedis(ctx context.Context) (*redis.ClusterClient, error) {
-	tlsConfig, err := buildRedisTLSConfig(ctx)
+	persistence, err := atepg.Connect(ctx, *postgresConnectionString)
```
jpbetz: Is the switch to requiring the DB be available on the first request on startup intentional here?  The redis path was more tolerant of blips.

hack/install-ate.sh
```diff
@@ -170,19 +163,6 @@ atenet_router() {
...
-      ;;
-    *)
-      echo "Error: store backend must be redis or postgres, got '${backend}'" >&2
-      exit 1
-      ;;
-  esac
-}
-
 default_postgres_connection_string() {
   echo "postgresql://postgres@postgres.ate-system.svc:5432/atepg?sslmode=verify-full&sslrootcert=/run/servicedns.podcert.ate.dev/trust-bundle.pem&sslcert=/run/podidentity.podcert.ate.dev/credential-bundle.pem&sslkey=/run/podidentity.podcert.ate.dev/credential-bundle.pem"
```
jpbetz, line 201: This connects as the postgres superuser with no password, and the server side is trust auth, so ateapi runs all its DDL and DML with full privileges. Recommend sorting out the app roles now (or very soon) so that you build on least-privilege instead of having to figure it out later.

cmd/kubectl-ate/internal/cmd/admin_debug_clear_store.go
```diff
@@ -22,27 +22,24 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var debugRedisFlushCmd = &cobra.Command{
-	Use:   "debug-flush-redis",
-	Short: "DANGEROUS: Flush all data from Redis",
+var debugClearStoreCmd = &cobra.Command{
+	Use:   "debug-clear-store",
+	Short: "DANGEROUS: Clear all control-plane state",
```
jpbetz: nit: now that this wipes the only store, a confirmation prompt or a --yes flag seems like cheap insurance. As far as I can tell the DebugClear RPC has no admin-only check server side, so the CLI is the only place a guard could live today. Fine as a follow-up.

jpbetz, line 27: I can see the value having this had for ease-of-testing for redis.
But I checked and there is are no server side role restrictions. And this is now deleting persistent state.

Is there an appropriately prioritized tracking issue to figure out how to remove or replace this before v1.0?

cmd/ateapi/internal/store/storecontract/contract.go
```diff
@@ -61,9 +61,7 @@ func newTestActorTemplateVersion(atespace, name, template string) *ateapipb.Acto
 }
 
 // mustCreateAtespace creates the atespace an actor test is about to populate.
-// Backends that enforce the actor->atespace foreign key (atepg) reject
-// CreateActor for a nonexistent atespace, so every actor test needs a real
-// parent atespace even though ateredis doesn't check.
+// The PostgreSQL store enforces the actor->atespace foreign key.
```
jpbetz, line 96: Is there a reason why creating a actor_template_versions without a parent is allowed? Is it to support out-of-order creation? The difference between how the referential integrity of this reference and object->namespace (which is FK projected) is what made me wonder what the rationale is here.

cmd/ateapi/internal/workercache/workercache.go
```diff
@@ -37,7 +37,7 @@ const relistPageSize = 1000
 // Cache maintains an in-memory snapshot of all workers.
 //
 // TODO: add metrics — at minimum a gauge for worker count, a counter for
-// resync events, and a counter for failed PUBLISH operations (in ateredis).
+// resync events, and a counter for failed worker-watch notifications.
```
jpbetz, line 40: I'm concerned that the proto payload can go over the 8000 byte NOTIFY limit with not well defined handling of the resulting behavior.

This appears to be a real problem because a larg-ish label set would make every worker write for that pool fail (scheduling included).

docs/architecture.md
```diff
@@ -226,7 +226,7 @@ environment definitions.
 
 These resources represent the high-frequency, ephemeral state of individual
 actors and workers. They are stored in a high-performance, low-latency state
-store (currently ValKey/Redis) to support real-time operations.
+store (PostgreSQL) to support real-time operations.
```
jpbetz, line 229: One thing to keep an eye on now that Postgres backs the hot path. ResumeActor takes the actor lease before the already-running check, so every routed cold check does a lease upsert plus delete on a tiny table. 

The leases have potential to cause a lot of churn. The leases can also become permanently orphaned. Since this is persistent state, cleanup is needed (open issue or address now?)

cmd/ateapi/internal/controlapi/workflow_resume_test.go
```diff
@@ -240,13 +237,10 @@ func TestAssignWorkerAttempt_RetryAfterConflictPicksFreshWorker(t *testing.T) {
 		t.Fatalf("UpdateWorker (concurrent claim): %v", err)
 	}
 
-	actor, err := persistence.CreateActor(ctx, &ateapipb.Actor{
+	actor := storetest.MustCreateActor(t, ctx, persistence, &ateapipb.Actor{
```
jpbetz: Two error paths in the claim flow end up with the wrong gRPC code, and neither is pinned by a test in this file.

First, if the syncer deletes the worker row between the cache read and the claim CAS, UpdateWorker returns ErrNotFound for the worker, the backoff loop only retries version conflicts ([workflow_resume.go#L323](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/controlapi/workflow_resume.go#L323)), and the handler's bare errors.Is turns it into NotFound 'Actor X not found' ([actor.go#L410-L418](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/controlapi/actor.go#L410-L418)). The router treats NotFound as terminal, so a transient scheduling race becomes a 404 about the wrong object. The real actor-missing case on this path is already a status error, so that branch fires almost only on the wrong-object case.

Second, when all 5 backoff attempts hit ErrVersionConflict, wait.ExponentialBackoff returns its bare timeout error, which matches neither errors.Is check and comes out as codes.Internal. The router doesn't retry Internal, so pure contention becomes a terminal 500. ensureVolumesCreated surfaces the conflict properly, this loop should too.

cmd/ateapi/internal/controlapi/actor_test.go
```diff
@@ -183,16 +183,13 @@ func TestCreateActor_RejectsDifferentTemplateForDataSnapshot(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Get source ActorTemplate: %v", err)
 	}
-	snapshot, err := tc.persistence.CreateActorSnapshot(context.Background(), &ateapipb.ActorSnapshot{
+	snapshot := storetest.MustCreateActorSnapshot(t, context.Background(), tc.persistence, &ateapipb.ActorSnapshot{
```
jpbetz: Related coverage gap for CreateActor. The store deliberately maps the atespace FK violation to ErrFailedPrecondition to close the race with the handler's pre-check ([atepg.go#L663-L667](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/store/atepg/atepg.go#L663-L667)), but the CreateActor handler only checks ErrAlreadyExists, so losing that race surfaces as codes.Internal instead of FailedPrecondition ([actor.go#L103-L109](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/controlapi/actor.go#L103-L109)). CreateActorSnapshotTag maps the same store error correctly. Worth a test that deletes the atespace concurrently with the create.

cmd/ateapi/internal/controlapi/actor_snapshot_test.go
```diff
@@ -353,20 +353,17 @@ func TestCreateActorSnapshotTag_RejectsUnsetScope(t *testing.T) {
...
-// in a miniredis-backed store, and returns a Service over it.
+// in a PostgreSQL-backed store, and returns a Service over it.
 func serviceWithActorSnapshotTag(t *testing.T, tag *ateapipb.ActorSnapshotTag) (*Service, *ateapipb.ActorSnapshotTag) {
 	t.Helper()
 	persistence, cleanup := storetest.SetupTestStore(t)
 	t.Cleanup(cleanup)
 
 	atespace, name := tag.GetMetadata().GetAtespace(), tag.GetMetadata().GetName()
-	snapshot, err := persistence.CreateActorSnapshot(context.Background(), &ateapipb.ActorSnapshot{
+	snapshot := storetest.MustCreateActorSnapshot(t, context.Background(), persistence, &ateapipb.ActorSnapshot{
```
jpbetz: One for a follow-up on tag creation. The contract pins CreateActorSnapshotTag with a missing snapshot to ErrNotFound, but that only holds when the pre-check catches it. If the snapshot is deleted between the pre-check and the INSERT, the FK violation surfaces as ErrFailedPrecondition instead ([atepg.go#L1103-L1122](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/store/atepg/atepg.go#L1103-L1122)). So the same condition produces two different errors depending on timing. Golden snapshot deletion makes the window real, since RESTRICT doesn't protect a tag that isn't inserted yet.

cmd/ateapi/internal/controlapi/functional_test.go
```diff
@@ -1646,11 +1621,11 @@ func TestListActors_Pagination(t *testing.T) {
 	}
 }
 
-// TestListWorkers tests that workers mirrored to Redis are listed.
+// TestListWorkers tests that workers mirrored to the store are listed.
```
jpbetz: The pagination tests above only cover well formed tokens. A malformed or wrong-method page token comes out of atepg as a plain error ([pagetoken.go#L62-L81](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/store/atepg/pagetoken.go#L62-L81)), no List handler special-cases it, and the interceptor converts it to codes.Internal. So a client that sends a bad argument gets a 13 and pages our server-error alerting. EItanya flagged this on #640 (it was Unknown back then) and it's still open. It should be InvalidArgument, and a test here would pin it.

cmd/ateapi/internal/store/storecontract/contract.go
```diff
@@ -97,9 +95,8 @@ func receiveEvent(t *testing.T, ch <-chan store.WorkerEvent) store.WorkerEvent {
 // against a fresh store.Interface built by setup for each subtest. setup is
 // responsible for its own cleanup (e.g. via t.Cleanup).
 //
-// Backend-specific behavior (e.g. ateredis's multi-shard pagination, atepg's
-// foreign-key races and transactional notifications) is NOT covered here; see
-// each backend's own test file for that.
+// PostgreSQL-specific behavior such as foreign-key races and transactional
```
jpbetz: Another gap worth pinning in the contract. Nothing here says what PageSize <= 0 means, and atepg panics on it. With PageSize 0 (the zero value of ListOptions) and a non-empty table, the truncation branch indexes names[-1] and takes down the whole process ([atepg.go#L862-L864](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/store/atepg/atepg.go#L862-L864), same shape in the other List methods). It's unreachable through gRPC today because controlapi defaults the page size first, but any internal caller passing a zero value ListOptions crashes ateapi. Either the contract should define PageSize <= 0 or the store should reject it.

cmd/ateapi/internal/controlapi/workflow_resume_test.go
jpbetz, line 339: An AI scan found these cases and they both look legitimate: 

**Case 1: the worker vanishes mid-claim, client is told the actor doesn't exist.**
1. Router asks to resume actor X. Scheduler picks worker W from the in-memory cache.
2. W's pod dies. The syncer notices and deletes W's row from the store.
3. The resume tries to claim W. The CAS UpdateWorker finds no row and returns ErrNotFound. That error is about the *worker*.
4. The retry loop only retries version conflicts, so it gives up and passes the error up as is ([workflow_resume.go#L323](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/controlapi/workflow_resume.go#L323)).
5. The ResumeActor handler sees ErrNotFound and assumes it means the *actor* ([actor.go#L410-L418](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/controlapi/actor.go#L410-L418)). Client gets NotFound 'Actor X not found'.
6. The router treats NotFound as final, so the request dies with a 404 for an actor that exists. The right outcome was to retry with another worker.

**Case 2: heavy contention becomes a permanent 500.**
1. Resume tries to claim a worker. Another resume gets there first, the CAS loses with ErrVersionConflict.
2. The loop retries with a fresh worker, 5 attempts total.
3. Pool is busy, all 5 lose the same way.
4. The backoff helper then returns its own generic 'timed out waiting for the condition' error, not the last conflict.
5. Nothing recognizes that error, so it surfaces as Internal, which the router won't retry.

cmd/ateapi/internal/controlapi/actor_test.go
jpbetz: This handler only checks `ErrAlreadyExists` ([actor.go#L103-L109](https://github.com/agent-substrate/substrate/blob/82f9821e5366c4e72a7de10b0588f83e994ae685/cmd/ateapi/internal/controlapi/actor.go#L103-L109)), so the `ErrFailedPrecondition` error falls through and the client gets Internal instead of 'Atespace not found'. 

Recommend a test that deletes the atespace concurrently with the create.

cmd/ateapi/internal/controlapi/actor_snapshot_test.go
jpbetz, line 349: Same problem as mentioned for CreateActor. If you create a tag for a snapshot that doesn't exist and you get ErrNotFound from the pre-check. But if the snapshot is deleted after the pre-check and before the insert, the FK violation comes back as ErrFailedPrecondition instead.

Generally we need to not rely on pre-checks to guard an operation if it can race with other operations.

cmd/ateapi/internal/controlapi/functional_test.go
jpbetz: Please add tests for invalid page tokens. These should not result in internal errors.

cmd/ateapi/internal/store/storecontract/contract.go
jpbetz, line 130: Looks like it's possible to panic the server by setting PageSize to 0 or negative. Please fix and add tests.

jpbetz (commented):
Thanks for the careful work on the schema design.

At a high level, the main concerns I have with the PostgreSQL implementation are below. Please consider either addressing in this PR or opening issues to ensure this all gets resolved:

1. Superuser for everything. Let's move to a least-privilege app role before this gets real exposure.
2. No schema evolution or mismatch detection. The schema is applied as CREATE TABLE IF NOT EXISTS on every startup. Later, if the schema were to be updated, there is nothing to detect drift for a fresh install vs. upgraded clusters.
3. Pessimistic locking. The read-modify-write paths hold SELECT ... FOR UPDATE row locks while app code runs between statements means that rows are locked for however long it takes for the server to run the code between statements. I'm concerned the scalability implications of having the lock duration paired with the go code execution.
4. Found worrying number of concurrency issues around pre-condition checks that can race with other DB updates.  I pointed out quite a few of them and did a AI run to search for them as well, but they are difficult to find.  Recommend adding significantly more tests that interleave operations that can interact.
5. Not enough unexpected input test cases. PageSize cases I found are a good example. Recommend going through all user inputs on all operations and checking for bad inputs.
6. LISTEN/NOTIFY robustness. Allowing undecodable events to be silently dropped instead of closing the channel, so consumers can't tell they have a gap, is unsafe. Recommend hardening the channel code for bad event shapes. Better to fail with a clear error than to silently drop data on a channel.
7. Object lifecycles. Orphaned leases is a good example of where transitioning to a persistent store surfaces issues that are not safe to ignore. In a persistent store like PostgreSQL, the orphaned data needs to get cleaned up.

cmd/ateapi/internal/controlapi/actor_snapshot_test.go
```diff
@@ -383,12 +380,10 @@ func TestUpdateActorSnapshotTag_DeleteRecreateRace(t *testing.T) {
 	t.Cleanup(cleanup)
 
 	for _, name := range []string{"snapshot-1", "snapshot-2"} {
-		if _, err := persistence.CreateActorSnapshot(ctx, &ateapipb.ActorSnapshot{
+		storetest.MustCreateActorSnapshot(t, ctx, persistence, &ateapipb.ActorSnapshot{
```
jpbetz, line 368: I decided to also do a quick scan for unindexed queries.

Postgres automatically indexes the referenced side of a foreign key but not the referencing side, and there's no index on actor_snapshot_tags (snapshot_atespace, snapshot_name) so the queries against are full scans.

docs/roadmap.md
HavenXia: +1 for the upgrade mechanism,  ideally it should keep the schema compatible during a pod rolling window, where old and new binaries are hitting the same database at the same time, so migrations follow an expand and contract pattern.

EItanya:
> Thanks for the careful work on the schema design.
> 
> At a high level, the main concerns I have with the PostgreSQL implementation are below. Please consider either addressing in this PR or opening issues to ensure this all gets resolved:
> 
> 1. Superuser for everything. Let's move to a least-privilege app role before this gets real exposure.
> 2. No schema evolution or mismatch detection. The schema is applied as CREATE TABLE IF NOT EXISTS on every startup. Later, if the schema were to be updated, there is nothing to detect drift for a fresh install vs. upgraded clusters.
> 3. Pessimistic locking. The read-modify-write paths hold SELECT ... FOR UPDATE row locks while app code runs between statements means that rows are locked for however long it takes for the server to run the code between statements. I'm concerned the scalability implications of having the lock duration paired with the go code execution.
> 4. Found worrying number of concurrency issues around pre-condition checks that can race with other DB updates.  I pointed out quite a few of them and did a AI run to search for them as well, but they are difficult to find.  Recommend adding significantly more tests that interleave operations that can interact.
> 5. Not enough unexpected input test cases. PageSize cases I found are a good example. Recommend going through all user inputs on all operations and checking for bad inputs.
> 6. LISTEN/NOTIFY robustness. Allowing undecodable events to be silently dropped instead of closing the channel, so consumers can't tell they have a gap, is unsafe. Recommend hardening the channel code for bad event shapes. Better to fail with a clear error than to silently drop data on a channel.
> 7. Object lifecycles. Orphaned leases is a good example of where transitioning to a persistent store surfaces issues that are not safe to ignore. In a persistent store like PostgreSQL, the orphaned data needs to get cleaned up.

First off, thanks so much for the detailed review! I'll run through these one at a time so we can try and see if we're on the same page.

1. 10000%. I think the idea was to merge something quickly, but this needs to be locked down before release
2. 10000%. We are aware on this and someone from my team should be able to pick this up soon. https://github.com/agent-substrate/substrate/issues/901
3. Totally agree with this takeaway. I also think this falls a bit under schema design, from what I can tell we could avoid the client side code by moving the types directly into the schema, but that has downsides as well. `FOR UPDATE NO KEY` however seems like an obvious win here.
4. Again totally agree. I think a big part of this is the awkward split between the service layer and the DB layer. This feels like an artifact of redis code and something we should absolutely address.
5. Yup!
6. We're actually moving away from `LISTEN/NOTIFY` anyway, see https://github.com/agent-substrate/substrate/pull/934. But we should definitely make sure to handle the cases you mentioned.
7. Absolutely

docs/roadmap.md
juli4n: +1 See #901

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -71,7 +71,7 @@ service Control {
   // List Actors.
   rpc ListActors(ListActorsRequest) returns (ListActorsResponse) {}
 
-  // Create a new Atespace. Substrate-native, stored in Redis.
+  // Create a new Atespace. Substrate-native, stored in PostgreSQL.
```
juli4n: Nit: We shouldn't talk about PostgreSQL here as it's not really relevant information for the API client.

jpbetz:
@juli4n @BenTheElder For scale, a couple things to consider:

- Don't use pessimistic concurrency at all. Switch to optimistic concurrency to shift the concurrency cost off of the DB.
- Drop the DB level UNIQUE constraint on UID and leave it to the server to set this correctly.  This lets you avoid paying the latency cost and DB load of a unique check for each write. If you ensure that the server mints UIDs, you still have the guarantee.
- Drop the cascading deletes and consider dropping the FKs entirely. Also, don't rely on the FKs to provide fast lookup. FK's are not indexed automatically, So the RESTRICT check on delete results in a full scan, as written.

hack/install-ate.sh
supreme-gg-gg: I think it's best to define this alongside migration design in #901, opened an issue to track it in #997

cmd/kubectl-ate/internal/cmd/admin_debug_clear_store.go
supreme-gg-gg: opened #999 to track this

cmd/ateapi/main.go
supreme-gg-gg: Not an intentional design, will add retry like redis

cmd/ateapi/internal/controlapi/actor_snapshot_test.go
supreme-gg-gg: Good catch for this, thanks! I missed the usage of this index in the FK trigger when deleting snapshots

cmd/ateapi/internal/workercache/workercache.go
supreme-gg-gg: I was aware of this size limit, but we will also be moving away from the `pg_notify` mechanism #934 which will bypass this issue

supreme-gg-gg:
@jpbetz, thank you for the detailed review! I addressed the review comments in a separate PR #1023, to keep those changes separate from the larger changes in this PR, which do not modify the PostgreSQL implementation itself.

I also switched the PostgreSQL store to optimistic concurrency, removed the UNIQUE constraints on server-generated UIDs, and added the missing indexes on referencing columns, as you've correctly pointed out.

> Drop the cascading deletes and consider dropping the FKs entirely.

I don’t believe there are currently any cascading deletes; the foreign keys use ON DELETE RESTRICT.

Dropping the foreign keys would be a significant semantic tradeoff. I’m not aware of an alternative that preserves the current race-free referential-integrity guarantees without introducing additional locking or transactional coordination. As you noted in several comments, application-level existence checks alone cannot provide the same guarantees because they can race with concurrent deletion.

jpbetz:
Thanks @supreme-gg-gg!  I'll follow along on the subsequent PR.

jpbetz:
@supreme-gg-gg Ordering question:

Should https://github.com/agent-substrate/substrate/pull/1023 merge before the switch over to atepg?  Without schema migration, switching to atepg before merging #1023 means that we almost immediately are updating the schema without any migration support.

BenTheElder:
FYI: https://github.com/agent-substrate/substrate/pull/1033/changes/809a4ae900d619a01b00f988ec33958f342eeae9

BenTheElder:
> Should https://github.com/agent-substrate/substrate/pull/1023 merge before the switch over to atepg? Without schema migration, switching to atepg before merging https://github.com/agent-substrate/substrate/pull/1023 means that we almost immediately are updating the schema without any migration support.

That's ~OK, substrate often has changes which require a ground up rebuild and full reinstall from main without backwards compat at the moment. We're going to have to exit that, but we haven't for other changes yet.

I do think we're going to need this, but with redis we were not really in a better place.

EItanya:
> > Should #1023 merge before the switch over to atepg? Without schema migration, switching to atepg before merging #1023 means that we almost immediately are updating the schema without any migration support.
> 
> That's ~OK, substrate often has changes which require a ground up rebuild and full reinstall from main without backwards compat at the moment. We're going to have to exit that, but we haven't for other changes yet.
> 
> I do think we're going to need this, but with redis we were not really in a better place.

Agreed, and honestly while we’re still quickly developing the schema and making a lot of changes I’d rather not waste time on safe movement between versions

jpbetz:
> > > Should #1023 merge before the switch over to atepg? Without schema migration, switching to atepg before merging #1023 means that we almost immediately are updating the schema without any migration support.
> > 
> > 
> > That's ~OK, substrate often has changes which require a ground up rebuild and full reinstall from main without backwards compat at the moment. We're going to have to exit that, but we haven't for other changes yet.
> > I do think we're going to need this, but with redis we were not really in a better place.
> 
> Agreed, and honestly while we’re still quickly developing the schema and making a lot of changes I’d rather not waste time on safe movement between versions

OK. Sounds good.

BenTheElder:
I have been doing some testing of my own, even with https://github.com/agent-substrate/substrate/pull/934, writing updates to worker is quite a bit more expensive on pg than redis.

I still don't think redis is going to be a great plan, but ... I'm not sure we have sufficient performance gap closure yet.

I'm also really wondering if we should still be exploring other options.

EDIT: debugged some of the issues to be in ate-api internals (resolvable), the notify change goes a long way, questions above remain though. Even though I'd rather have a durable backend.

EItanya:
> I have been doing some testing of my own, even with #934, writing updates to worker is quite a bit more expensive on pg than redis.
> 
> I still don't think redis is going to be a great plan, but ... I'm not sure we have sufficient performance gap closure yet.
> 
> I'm also really wondering if we should still be exploring other options.
> 
> EDIT: debugged some of the issues to be in ate-api internals (resolvable), the notify change goes a long way, questions above remain though. Even though I'd rather have a durable backend.

Do we have specific measurements as to where exactly the bottleneck is? Let's compile those specific spots and see if we can come up with solutions before talking about moving on from PG.

BenTheElder:
> Do we have specific measurements as to where exactly the bottleneck is? Let's compile those specific spots and see if we can come up with solutions before talking about moving on from PG.

I held off following up further because we clearly have in-flight performance fixes. But I think we should check some common operations more closely when we're otherwise actually ready.

We also have the same robustness issue with long lived installs as we did initially with valkey, filed a fix: https://github.com/agent-substrate/substrate/pull/1170

pkg/proto/ateapipb/ateapi.proto
supreme-gg-gg: Updated the comment

EItanya (approved):

EItanya:
The controversial bits of this PR were pulled out into #1023, which has now merged, so going to push this through

Issue #932, opened 2026-08-13
Add trustBundle as a SystemInfo volume data source


thompsonmax:
## Summary

Add a `trustBundle` data source to SystemInfo volumes (#802) that projects the trust anchors of a named trust bundle to a PEM file in the volume — inspired by the [Kubernetes clusterTrustBundle projected volume source](https://kubernetes.io/docs/concepts/storage/projected-volumes/#clustertrustbundle), but deliberately source-neutral: the substrate API names a bundle, and how that bundle is fetched is a deployment concern behind ateapi. Initially the only supported bundle is the well-known `egress-mitm.ate.dev`, resolved from a Kubernetes ClusterTrustBundle.

## Motivation

#871 lands MITM'd egress TLS: Envoy terminates actor egress with short-lived per-SNI leaves minted under the egress-mitm-ca-pool CA. It explicitly leaves one problem open: how to make actors trust that CA. The plan (per the egress design on #823) is a trust bundle maintained by atecontroller from the CA pool, published as a Kubernetes ClusterTrustBundle. This issue introduces the mechanism for getting those trust anchors onto each actor's filesystem, correct across suspend/resume and snapshot seeding.

This is precisely what SystemInfo volumes exist for (#802): per-actor files generated by the platform on every Run/Restore, never captured into checkpointed state.

Note the k8s ClusterTrustBundle is a distribution channel here, not the source of truth (the CA pool is), and substrate APIs should not take a hard dependency on Kubernetes APIs. The data source is therefore named `trustBundle`, not `clusterTrustBundle`: templates reference a bundle by name, and resolving that name to a backing store is substrate's job. Initially we will just support the well known  `egress-mitm.ate.dev`, fetched from the ClusterTrustBundle of the same name. Eventually we will introduce a configurable backend registry so users can specify the trustBundles that are available to their actors and where they can be fetched from.

## Proposed API

```yaml
spec:
  volumes:
  - name: trust
    systemInfo:
      dataSources:
      - trustBundle:
          name: egress-mitm.ate.dev
          path: egress-ca.pem
  containers:
  - name: main
    volumeMounts:
    - name: trust
      mountPath: /run/substrate/certs  # e.g. the actor reads /run/substrate/certs/egress-ca.pem
```

## Semantics

[...]

Issue #823, opened 2026-08-10
Add an SDS server that mints a short-lived leaf per SNI on demand at handshake time


haiyanmeng:
The traffic from an actor is sent to the egress gateway, `Envoy`, through `atunnel egress`, an HTTP CONNECT tunnel. To allow Envoy to inspect the HTTP headers of egress traffic, we plan to introduce an SDS server that mints a short-lived leaf per SNI on demand at handshake time, `sdsmintd`.

Stevenjin8:
I'll just say that this would require some application level changes, notably, nodejs uses a bundled set of CAs, rather than discovering them from /etc/certs or whatever

EItanya:
> I'll just say that this would require some application level changes, notably, nodejs uses a bundled set of CAs, rather than discovering them from /etc/certs or whatever

It will depend on the application. Most codebases/applications follow some wellknown folders and/or env vars, but there are of course some that won't.

PR #941, opened 2026-08-13
Add trustBundle as a SystemInfo volume data source

thompsonmax:
Part of #932 (PR 1 of 3). Adds the user-declarable trustBundle data source for SystemInfo volumes (#802) and the end-to-end proof that the projected anchors work against the MITM egress gateway. Live refresh for running actors (PR 2) and auto-injection (PR 3) come separately.

What this adds

A SystemInfo volume data source that projects the trust anchors of a named trust bundle to a PEM file:

volumes:
- name: trust
  systemInfo:
    dataSources:
    - trustBundle:
        name: egress-mitm.ate.dev
        path: egress-ca.pem

Inspired by the Kubernetes clusterTrustBundle projected volume source, but source-neutral: the template names a bundle; where it's fetched from is a deployment concern, not part of the API.

Design points

- Resolution lives on the node. The wire carries only {name, path}; atelet resolves the name at write time through an informer-backed lister on ClusterTrustBundles and writes the sanitized PEM with the temp+rename discipline from #803 (find-paths safe). Contents refresh on every Run/Restore. ateapi is not involved, per review discussion — the same informer is what live refresh (PR 2) will hang off.
- Allowlist in atelet, not the CRD schema. Today only egress-mitm.ate.dev (the egress gateway CA bundle, #823), mapped to the ClusterTrustBundle that atecontroller's EgressMITMTrustReconciler (#946) derives from the egress-mitm-ca-pool Secret. The signer-linked object name stays a backend detail; the future backend registry (#932) widens the allowlist without an API change.
- The watch is scoped to the one backing object via a metadata.name field selector — this informer runs on every node, so an unfiltered watch would fan every ClusterTrustBundle in the cluster out to every atelet. RBAC can't express this (resourceNames doesn't apply to list/watch), so the field selector is the enforcement point. get/list/watch on clustertrustbundles moves to the atelet ClusterRole.
- No availability probe. The informer registers unconditionally; a cluster that doesn't serve the feature-gated certificates.k8s.io/v1beta1 blocks atelet startup at cache sync, with the reflector errors naming the missing API (hack/create-kind-cluster.sh enables the gate).
- Fail-closed. Unknown names, missing bundles, and unusable bundles fail actor start naming the bundle — an actor that declared a trust bundle must not start without one.
- Kubelet-parity sanitization (internal/pemutil): CERTIFICATE blocks only, deduplicated, headers stripped, and anchors deliberately shuffled so consumers can't grow a dependence on order.
- Schema note: dataSources MaxItems tightened 32→8 while adding the trustBundle member. Vacuous in practice (the old schema couldn't admit more than one entry), but flagged since it's ratchet-shaped.

E2E — delivery and consumption

Delivery (identity suite, both sandbox classes): provisions the egress-mitm-ca-pool Secret and drives the real #946 reconciler (writing the bundle directly isn't possible — the reconciler reverts hand-edits), asserts the projected file byte-exact, then rotates the pool across a suspend/resume to prove refresh-on-restore. Since the probe fixture is shared and fail-closed, e2e.DeployProbe itself ensures the bundle exists for whatever suite deploys it.

Consumption (new egressmitm suite, both sandbox classes): deploys the sdsmint (MITM) egress gateway and proves an actor completes a TLS handshake with the gateway's per-SNI minted leaf using ONLY the projected anchors — plus a system-roots negative control that must fail. The pair is unambiguous in both directions: the positive can't pass under passthrough (the bundle holds no public CAs), and the negative can't fail under passthrough.

CI: two steps appended to the existing e2e job after the standard lanes (the gateway swap is cluster-wide and breaks passthrough assumptions): --deploy-atenet --experimental-use-sdsmint redeploys only the atenet components, then the egressmitm suite runs once per sandbox class.

Flake mitigation: the probe fixture pool drops from 3 workers to 2. Each suite deploys its own copy and drives one actor at a time, so the third worker per copy was idle memory multiplied across suites on the one-node CI cluster — pressure that has been killing sandboxes mid-test (runsc: signal: killed, a vanished ateom socket) on this PR and on main's identity suite. This reduces the pressure; right-sizing e2e concurrency or worker-pod QoS cluster-wide is follow-up material.

Not in this PR

- Live refresh for running actors (#932 PR 2) — until then, a running actor's file is the bundle as of its last Run/Restore, and correctness rests on overlap rotation by the bundle publisher.
- Auto-injection of the egress trust volume (#932 PR 3).
- Configurable backend registry (#932) — the allowlist is the seam it will replace.


ahmedtd:
My main question is: what purpose does having ate-api-server involved serve?  It seems much simpler for atelet to read the ClusterTrustBundle with an informer (and it will need to anyways, in order to support dynamic refresh).

internal/pemutil/pemutil.go
```diff
@@ -0,0 +1,54 @@
...
+			continue
+		}
+		key := sha256.Sum256(block.Bytes)
+		if seen[key] {
+			continue
+		}
+		seen[key] = true
+		out = append(out, pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: block.Bytes})...)
+	}
+	if len(out) == 0 {
```
ahmedtd: We should also apply a (stable) shuffle: https://github.com/kubernetes/kubernetes/blob/e81f39c0e03ce8ed8e2660c9147b391edd9e262b/pkg/kubelet/clustertrustbundle/clustertrustbundle_manager.go#L321

cmd/atelet/main.go
```diff
@@ -1604,13 +1545,27 @@ func (s *AteomHerder) prepareOCIBundles(
...
 	}
 
 	for _, dataSourceAny := range si.GetDataSources() {
 		switch dataSource := dataSourceAny.GetDataSource().(type) {
+		case *ateletpb.SystemInfoDataSource_TrustBundle:
+			tb := dataSource.TrustBundle
+			// ateapi resolves the bundle before sending the spec; empty bytes
+			// mean it did not, and an empty trust file would fail the workload
+			// in far more confusing ways than failing the start does.
+			if len(tb.GetPemBundle()) == 0 {
```
ahmedtd: Let's cut out all of the ate-api-server involvement.  Atelet should have a ClusterTrustBundle informer for this one bundle, and use that.

.github/workflows/pr-workflow.yaml
```diff
@@ -107,6 +107,30 @@ jobs:
       env:
         E2E_SANDBOX_CLASS: microvm
       run: hack/run-e2e-kind.sh -v -args --no-color
+    - name: Deploy MITM egress (sdsmint)
+      # Swap the passthrough egress gateway for the sdsmint variant, which
+      # mints per-SNI leaves from the egress-mitm-ca-pool (created here if
+      # missing). Cluster-wide, so it must come AFTER the standard lanes:
+      # once egress TLS is intercepted, their passthrough assumptions
+      # (TestActorEgressHTTPS's end-to-end TLS with the origin) no longer hold.
+      run: hack/install-ate-kind.sh --deploy-ate-system --experimental-use-sdsmint
```
haiyanmeng: Adding a new workflow for sdsmint is expensive. 

We are exploring how to run sdsmint e2e tests in presubmit github workflow in
https://github.com/agent-substrate/substrate/issues/1004. One way is to add a new step into our existing e2e github workflow to test sdsmint.

docs/api-guide.md
```diff
@@ -226,6 +226,30 @@ spec:
...
 
+#### trustBundle
+The trustBundle data source projects the trust anchors of a named trust bundle to a single PEM file — inspired by the [Kubernetes clusterTrustBundle projected volume source](https://kubernetes.io/docs/concepts/storage/projected-volumes/#clustertrustbundle), but source-neutral: the name selects a bundle substrate knows how to fetch, and where it is fetched from is a deployment concern, not part of the API.
+
+Supported names are allowlisted. Today the only supported bundle is `egress-mitm.ate.dev` — the egress gateway CA bundle — resolved from the [ClusterTrustBundle](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#cluster-trust-bundles) (`certificates.k8s.io/v1beta1`) that atecontroller's reconciler derives from the `egress-mitm-ca-pool` Secret. A configurable backend registry may widen the allowlist later.
```
haiyanmeng: Should we call it out that the Secret is from the namespace `ate-system`?

.github/workflows/pr-workflow.yaml
haiyanmeng: Running the sdsmint tests on both gvisor and microvm is expensive. Can we get away with just running on one of them?

cc @BenTheElder  @bowei

internal/e2e/fixtures/probe/probe.yaml.tmpl
haiyanmeng: `# the probe reads /run/ate/actor-id` -> `# the probe reads actor metadata and trust bundles under /run/ate`

.github/workflows/pr-workflow.yaml
BenTheElder: We should be running tests on both, end to end. What makes them expensive?

haiyanmeng: Adding a new workflow (the may need more testing resource?). Or adding new steps to an existing workflow (which could make it longer to run the workflow).

haiyanmeng: If this is not a concern, I am happy to have a dedicated workflow.

internal/pemutil/pemutil.go
thompsonmax: Good point, added

cmd/atelet/main.go
thompsonmax: Yeah good point, that simplifies things a bit. Done

thompsonmax:
> My main question is: what purpose does having ate-api-server involved serve? It seems much simpler for atelet to read the ClusterTrustBundle with an informer (and it will need to anyways, in order to support dynamic refresh).

@ahmedtd Good point. I was unsure of how much coupling we wanted on k8s in atelet, but it already has some informers set up on it, so one more doesn't seem too bad. Updated to informer on atelet approach.

When we extend support to actor identity tokens and certs in SystemInfo volumes, I think some added code in ateapi is unavoidable, but we can keep things simple for CTBs for now.

.github/workflows/pr-workflow.yaml
thompsonmax: My understanding is that this isn't a new workflow per-se, it's just adding new steps to the `e2e-test` workflow to apply the sdsmint gateway and run these tests against an already existing KIND cluster. That being said, I believe the added time is from:

1. `install-ate-kind.sh --deploy-ate-system --experimental-use-sdsmint` does a fully control plane re-apply
2. Then the MITM test run for both gVisor and μVM. μVM seems likely to take the most time since the golden is a cold cloud-hypervisor boot + checkpoint on nested KVM.

Not sure how we can optimize this without reducing the test coverage. @haiyanmeng seems you have some ideas in https://github.com/agent-substrate/substrate/issues/1004.

Another thought I had was maybe to create some kind of "nightly" test run for substrate where we can put these tests that are a bit more expensive.

haiyanmeng: > install-ate-kind.sh --deploy-ate-system --experimental-use-sdsmint

@thompsonmax , we can do `install-ate-kind.sh --deploy-atenet --experimental-use-sdsmint` to save time

haiyanmeng: > My understanding is that this isn't a new workflow per-se, it's just adding new steps to the e2e-test workflow

You are right. My bad. 
Let us try `install-ate-kind.sh --deploy-atenet --experimental-use-sdsmint` to see how much longer it makes the workflow. If it is not too bad, we can merge this as-is, since we expect turn on sdsmint within the next couple weeks. If things get really bad, we will skip the new e2e test.

thompsonmax: Good idea, I switched to that command. I will run the E2E tests with the new tests timed to see how long it would be expected to extend the CI.

BenTheElder: yeah, just redeploying the strictly necessary components is probably good enough, let's not add a dedicated workflow yet.

cmd/ateapi/internal/controlapi/workload_spec.go
```diff
@@ -56,6 +56,18 @@ func workloadSpecFromActorTemplate(actorTemplate *atev1alpha1.ActorTemplate, act
 							ActorMetadata: actorMetadata,
 						},
 					})
+				case dataSource.TrustBundle != nil:
+					// The NAME crosses the wire: atelet resolves it against
```
ahmedtd: Nit: agent-y comment.

cmd/atelet/main.go
```diff
@@ -276,10 +280,45 @@ func main() {
 	ateFactory := externalversions.NewSharedInformerFactory(ateClient, 0)
 	csiDriverConfigLister := ateFactory.Api().V1alpha1().CSIDriverConfigs().Lister()
 
+	// trustBundle SystemInfo data sources resolve on the node through an
```
ahmedtd: "Start an informer on the ClusterTrustBundle we care about"

cmd/atelet/main.go
```diff
@@ -411,6 +451,10 @@ type AteomHerder struct {
 	mu                    sync.RWMutex
 	volumePlugins         map[string]volume.VolumePluginWorkerPlane
 	csiDriverConfigLister listersv1alpha1.CSIDriverConfigLister
+	// clusterTrustBundleLister backs trustBundle SystemInfo data sources
```
ahmedtd: Nit: agent-y comment

cmd/atelet/trustbundle.go
```diff
@@ -0,0 +1,143 @@
...
+//     enabling the gate, a 404 can be transient, so NotFound consumes the
+//     same retry budget as any other error and only a consistent final
+//     NotFound degrades.
+//   - Forbidden: the API exists but RBAC is stale — fail startup naming the
+//     missing rule rather than hang on cache sync or silently degrade.
+//   - anything else (apiserver busy during a rollout, transient network):
+//     retried briefly, then fails startup; a restart heals it, whereas a
+//     permanently degraded lister would blame the cluster's API support in
+//     every actor-start error until someone thought to bounce atelet.
+func clusterTrustBundleAPIAvailable(ctx context.Context, clientset kubernetes.Interface) (bool, error) {
```
ahmedtd: I don't think we need this --- just let informer creation fail if ClusterTrustBundles aren't available.

manifests/ate-install/generated/ate.dev_actortemplates.yaml
```diff
@@ -498,9 +498,9 @@ spec:
                             DataSources is the list of data sources to place within the SystemInfo
                             volume.
 
-                            At most one actorMetadata entry may appear; this is what keeps file
```
ahmedtd, line 501: @HavenXia @zoez7 Have ActorTemplates as Substrate resources landed?

cmd/ateapi/internal/controlapi/workload_spec.go
thompsonmax: de-clauded this one a bit

cmd/atelet/main.go
thompsonmax: I just removed this one, probably too much detail here.

cmd/atelet/trustbundle.go
thompsonmax: Sure, removed this

ahmedtd (dismissed):

manifests/ate-install/generated/ate.dev_actortemplates.yaml
thompsonmax: AFAICT the ActorTemplate protos landed in https://github.com/agent-substrate/substrate/pull/1026 but Actors are still being created from the CRD?

I'll plan to add the SystemInfo fields to ActorTemplate in a follow-up PR.

.github/workflows/pr-workflow.yaml
thompsonmax: Sounds good. From a local run on my linux machine, it seems like the added tests didn't extend the overall time too much (generated a quick script to quantify the difference):

```
──────────────── results ────────────────
baseline: create cluster             0m 14s
baseline: install ate-system         1m 26s
baseline: microvm demo               0m 38s
baseline: gvisor demo                0m 26s
baseline: egress demo                0m 07s
baseline: egress demo (mvm)          0m 07s
baseline: wait mvm golden            0m 00s
baseline: gVisor lane                2m 19s
baseline: micro-VM lane              2m 45s
ADDED: sdsmint gateway swap          0m 20s
ADDED: egressmitm gVisor             0m 14s
ADDED: egressmitm micro-VM           0m 17s
──────────────────────────────────────────
baseline total (pre-existing job)    8m 02s
ADDED total (this PR)                0m 51s
overhead: +10%
```

thompsonmax: But also, worth noting my machine has a few more cores than the CI runner :^) still, probably reasonable to expect a 10% increase in CI runtime.

ahmedtd (approved):

.github/workflows/pr-workflow.yaml
haiyanmeng: Sounds good. Thanks for measuring it.

haiyanmeng (approved):

manifests/ate-install/generated/ate.dev_actortemplates.yaml
BenTheElder: Yes. We're still using the CR, it's not cutover yet.

PR #942, opened 2026-08-14
fix(hack): let teardown delete crashed actors

### LINKED ISSUE #794: install-ate.sh: a crashed golden actor makes --delete-all keep the old golden snapshot
## Expected Behavior

`--delete-all` removes every actor, including crashed ones, so `--deploy-demo-counter` installs a fresh ActorTemplate with a newly generated golden snapshot.

## Actual Behavior

When the skipped actor is a golden actor, its ActorTemplate survives the
teardown. The redeploy is then rejected, because an ActorTemplate spec is
immutable:
```
The ActorTemplate "counter" is invalid: spec: Invalid value: Spec is immutable
Error: error executing 'kubectl apply': exit status 1
```

The script continues and exits with 0. The cluster looks ready but still serves the previous golden snapshot. In my case it had been built by an older gVisor release, so every resume of the counter template failed with
`while running "runsc restore": exit status 128`, with nothing in the install output pointing back at the cause.

As a workaround, you can run `--delete-demo-counter` right before redeploying.
 

1. `hack/install-ate-kind.sh --deploy-ate-system && hack/install-ate-kind.sh --deploy-demo-counter`
2. Get the template's golden actor into `STATUS_CRASHED` (deleting its worker pod does it; running the e2e suites does it incidentally)
3. `hack/install-ate-kind.sh --delete-all`, and note it reports `cannot delete actor ... STATUS_CRASHED` but exits 0
4. `hack/install-ate-kind.sh --deploy-demo-counter`, and note the `Spec is immutable` error, also exit 0
5. `kubectl get actortemplate -n ate-demo-counter counter -o jsonpath='{.status.goldenSnapshot}'` still shows the pre-teardown snapshot ID

## Proposal
Let teardown delete STATUS_CRASHED actors directly, since a crashed actor has nothing to suspend, and make the ActorTemplate apply failure fail the step.

## Specifications

- Version: main
- Platform: kind, Kubernetes v1.36.1, gvisor sandbox class

orangeCatDeveloper:
Fixes #794

A crashed actor makes `--delete-all` abort mid-teardown, leaving the ActorTemplate, atespace and `ate-system` behind; the next `--deploy-demo-counter` keeps the existing template, so the cluster goes on serving the stale golden snapshot.

`prepare_actor_for_delete` treats `STATUS_CRASHED` as unexpected and returns 1, which errexit turns into a hard abort, even though DeleteActor accepts CRASHED and is idempotent for an actor already marked DELETING (`workflow_delete.go`). Teardown now treats crashed and already-deleting actors as deletable as-is.

The demo flag dispatch had the mirror-image defect: handlers ran from an `if` condition, which suppresses errexit for their whole call tree, so this same teardown and failed deploys like the issue's immutable-spec apply exited 0. Handlers now run as plain commands and report an unclaimed flag through `ate_demo_flag_unhandled`; the Jupyter handler follows the same contract.

Repro: deploy the counter demo, resume an actor, delete its worker pod, wait for the syncer to mark it `STATUS_CRASHED`, then run `--delete-all`.

### DIFF
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -1110,7 +1110,7 @@ get_actor_state() {
 }
 
 # prepare_actor_for_delete suspends (or resumes then suspends) until DeleteActor
-# accepts the actor: ACTOR_STATE_SUSPENDED or ACTOR_STATE_CRASHED.
+# accepts the actor: ACTOR_STATE_SUSPENDED, ACTOR_STATE_CRASHED, or ACTOR_STATE_DELETING.
 prepare_actor_for_delete() {
   local actor_name="$1"
   local atespace="$2"
@@ -1124,7 +1124,7 @@ prepare_actor_for_delete() {
     fi
 
     case "${state}" in
-      ACTOR_STATE_SUSPENDED | ACTOR_STATE_CRASHED)
+      ACTOR_STATE_SUSPENDED | ACTOR_STATE_CRASHED | ACTOR_STATE_DELETING)
         return 0
         ;;
       ACTOR_STATE_PAUSED)
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -38,6 +38,11 @@ fi
 # ATE_DEMOS is an array that registers the prefix name of the demo functions.
 ATE_DEMOS=()
 
+# Called by a ${demo}_cmdline handler for an argument it does not own.
+ate_demo_flag_unhandled() {
+  ATE_DEMO_FLAG_HANDLED=false
+}
+
 # Include demos.
 source "${ROOT}"/hack/install-demo-counter.sh
 source "${ROOT}"/hack/install-demo-egress.sh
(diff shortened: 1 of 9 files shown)


hack/install-ate.sh
```diff
@@ -641,7 +646,7 @@ prepare_actor_for_delete() {
     fi
 
     case "${status}" in
-      STATUS_SUSPENDED)
+      STATUS_SUSPENDED | STATUS_CRASHED)
```
krisztianfekete: Should we include `DELETING` here?

krisztianfekete (commented):
Added one comment, otherwise looks good to me!

hack/install-ate.sh
orangeCatDeveloper: Good catch. Updated in the latest commit.

krisztianfekete (dismissed):
Thanks!

orangeCatDeveloper:
> Thanks!

Could you help trigger the CI workflow? Thank you!

krisztianfekete:
Hey, I don't have those kind of permission, but now I see that someone triggered them!

krisztianfekete (approved):

PR #949, opened 2026-08-14
fix(ateapi): keep worker watches draining

### LINKED ISSUE #624: [P2] Worker cache pub/sub channel (cap 128) silently drops events under high pod churn — forces unscheduled full resync
**Severity:** P2 (scheduling unavailability during large rolling restarts)
**Component:** Control Plane — `cmd/ateapi/internal/store/ateredis/ateredis.go`
**Confirmed:** Yes — code-level analysis

---

## Summary

`WatchWorkers` creates a Go channel with capacity 128 to buffer worker change events
from Valkey pub/sub. Under high worker churn (large rolling restarts, pool scale-ups),
more than 128 events can arrive faster than the consumer processes them. When the buffer
fills, the goroutine forwarding events from Valkey to the channel blocks. The underlying
go-redis socket backs up, causing Valkey to detect a stalled consumer and close the
connection. This triggers a full resync — during which `workercache.Cache.ready` is
set to `false` and all `Workers()` calls return an error, making every concurrent
`ResumeActor` call fail with "worker cache not ready."

---

## Impact

- During a rolling restart of a large worker pool (200+ pods), all resume calls fail
  for the duration of the resync (up to 30s per retry attempt, capped by exponential
  backoff).
- No event-drop metric or counter exists; operators have no signal before the outage
  begins.
- The only log is "worker cache: watch channel closed, resyncing" which appears AFTER
  the damage is done.
- This is a correlated failure: large deployments (which are most likely to trigger
  the overflow) are also most likely to have many concurrent actors being resumed.

---

## Root Cause

**File:** `cmd/ateapi/internal/store/ateredis/ateredis.go` lines 261–303

```go
func (s *Persistence) WatchWorkers(ctx context.Context) (*store.WorkerWatch, error) {
    ...
    ch := make(chan store.WorkerEvent, 128)   // fixed 128-event buffer, line 276
    ...
    go func() {
        for msg := range msgCh {             // msgCh is go-redis pubsub channel
            ...
            select {
            case ch <- event:                // blocks when ch is full
            case <-watchCtx.Done():
                return
            }
        }
        close(ch)
    }()
    return &store.WorkerWatch{Events: ch, Close: closeFn}, nil
}
```

When `ch <- event` blocks (buffer full), the goroutine stops reading from `msgCh`.
The go-redis library's default pubsub channel buffer is 100; when that also fills,
the underlying TCP socket stops being drained. Valkey detects the stalled client and
closes the connection. The goroutine sees `msgCh` closed, closes `ch`, and the
`workercache` triggers `resync()`.

**File:** `cmd/ateapi/internal/workercache/workercache.go` lines 138–145:
```go
case event, ok := <-watch.Events:
    if !ok {
        c.ready.Store(false)    // Scheduling immediately unavailable
        ...
        watch = c.resync(ctx)  // Can take 1-30s+ with exponential backoff
        c.ready.Store(true)
    }
```

The backoff cap is 30 seconds per attempt (`workercache.go:167`); under sustained
Valkey pressure, the resync could be delayed significantly.

---

## Steps to Reproduce

```bash
# Deploy a large worker pool (200+ pods)
kubectl patch workerpool counter -n ate-demo-counter \
  --type=merge -p '{"spec":{"replicas":200}}'

# Wait for all pods to be running
kubectl wait --for=condition=available deployment/counter -n ate-demo-counter --timeout=300s

# Trigger a rolling restart (generates 200+ UPDATED events in rapid succession)
kubectl rollout restart deployment/counter -n ate-demo-counter &

# Simultaneously send ResumeActor requests
for i in $(seq 1 20); do
  kubectl ate resume actor actor-$i -a stress-test &
done

# Observe: some resumes fail with "worker cache not ready"
# Check ateapi logs:
kubectl logs -n ate-system -l app=ate-api-server | grep "watch channel closed\|not ready"
```

---

## Expected Behavior

Worker cache event overflow should be handled gracefully:
1. Emit a `dropped_events_total` counter metric when the channel would block.
2. Apply the dropped event as a cache invalidation (force a relist) rather than
   losing the event silently.
3. Alternatively, use an unbounded queue (with a semaphore to cap memory) so individual
   events are never lost.

---

## Suggested Fix

Replace the blocking `ch <- event` with a non-blocking send that triggers a relist
on overflow:

```go
case ch <- event:
    // normal fast path

default:
    // channel full — drop event and trigger corrective relist
    droppedEvents.Add(1)  // increment a metric counter
    select {
    case triggerRelist <- struct{}{}:
    default:
    }
```

And in `watchEvents`, handle `triggerRelist` by calling `c.relist(ctx)` immediately
rather than waiting for the ticker. This ensures cache accuracy is restored within
one relist call (milliseconds) rather than after a full TCP disconnect + reconnect cycle.

orangeCatDeveloper:
Fixes #624

Under heavy worker churn the worker cache silently serves a stale worker list
for up to a full relist interval — long enough to schedule actors onto workers
that no longer exist, with nothing in the logs or metrics saying so.

## Root cause

`WatchWorkers` forwards pub/sub messages into a 128-event channel with a
blocking send. A slow consumer fills it, the forwarding goroutine parks on
`ch <- event` and stops reading the socket, and Valkey evicts the subscriber for
overrunning its pub/sub output buffer:

```
Client id=7 ... flags=P ... omem=41008 lib-name=go-redis(,go1.26.3)
  scheduled to be closed ASAP for overcoming of output buffer limits.
```

The eviction never reaches us: go-redis reconnects and resubscribes underneath,
closing its message channel only on `pool.ErrClosed`. So there is no closed watch
and no resync — queued and in-flight events are simply gone while `ready` stays
`true`. Quieter and worse than #624 assumed: not a scheduling outage, but
silently wrong scheduling input.

## The fix

Overflow drops the event, counts it, and raises a coalesced
`WorkerWatch.Invalidated` signal instead of blocking, so the socket keeps
draining and Valkey has no cause to evict. `workercache` reads that as "relist
now", first draining events that predate the drop (they would roll the fresh
snapshot back). A failed relist retries on a 100ms→5s backoff instead of waiting
for the periodic ticker; a successful resync or periodic relist cancels a pending
retry. A signal arriving while a retry is armed folds into it rather than
relisting again, so the backoff still holds under the churn that caused the
drops. The old snapshot keeps serving throughout — no `ready=false` window.

The PostgreSQL store passes no invalidation channel: its outbox poller never
drops events — on a truncated outbox, retention gap, or corrupt row it closes
the watch to force a resync instead.

## Evidence

Real Valkey 8, `client-output-buffer-limit pubsub 32kb 8kb 3`, stalled consumer,
50000 events (~6MB), fresh container per run:

| | before | after |
|---|---|---|
| `client_output_buffer_limit_disconnections` | 1 | 0 |
| invalidation signal | none | SIGNALED |
| `ate.store.worker_watch.dropped_events` | no metric | 49872 |

Cache recovery after a lost DELETE, at the production 5-minute relist interval:

```
before  corrected_after=NEVER (>10s; waits for the 5m relist)
after   corrected_after=6ms
```

### DIFF
--- a/cmd/ateapi/internal/actoridentity/actoridentity_test.go
+++ b/cmd/ateapi/internal/actoridentity/actoridentity_test.go
@@ -172,7 +172,7 @@ func newTestServer(t *testing.T, st store.Interface) *Server {
 type staleWatchStore struct{ store.Interface }
 
 func (s staleWatchStore) WatchWorkers(context.Context) (*store.WorkerWatch, error) {
-	return store.NewWorkerWatch(make(chan store.WorkerEvent), func() {}), nil
+	return store.NewWorkerWatch(make(chan store.WorkerEvent), nil, func() {}), nil
 }
 
 // TestMintCertReadsThroughStaleWorkerCache pins the authorization
--- a/cmd/ateapi/internal/controlapi/functionaltest/common_test.go
+++ b/cmd/ateapi/internal/controlapi/functionaltest/common_test.go
@@ -116,7 +116,7 @@ func setupTestWithVolumePlugins(t *testing.T, ns string, plugins map[string]volu
 	rdb := redis.NewClusterClient(&redis.ClusterOptions{
 		Addrs: []string{mr.Addr()},
 	})
-	persistence := ateredis.NewPersistence(rdb)
+	persistence := ateredis.NewPersistence(rdb, nil)
 
 	// 2. Initialize Clientsets using global cfg
 	k8sClient, err := kubernetes.NewForConfig(cfg)
--- a/cmd/ateapi/internal/controlapi/workflow_suspend_test.go
+++ b/cmd/ateapi/internal/controlapi/workflow_suspend_test.go
@@ -237,7 +237,7 @@ func newTestPersistence(t *testing.T) store.Interface {
 	t.Cleanup(mr.Close)
 	rdb := redis.NewClusterClient(&redis.ClusterOptions{Addrs: []string{mr.Addr()}})
 	t.Cleanup(func() { rdb.Close() }) //nolint:errcheck // test cleanup
-	return ateredis.NewPersistence(rdb)
+	return ateredis.NewPersistence(rdb, nil)
 }
 
 // newDanglingDialer returns a dialer whose informer cache has no pods, so
--- a/cmd/ateapi/internal/store/atepg/outbox.go
+++ b/cmd/ateapi/internal/store/atepg/outbox.go
@@ -567,5 +567,6 @@ func (p *Persistence) WatchWorkers(ctx context.Context) (*store.WorkerWatch, err
 			}
 		}
 	}()
-	return store.NewWorkerWatch(ch, cancel), nil
+	// nil invalidation: this watch closes rather than dropping events.
+	return store.NewWorkerWatch(ch, nil, cancel), nil
 }
--- a/cmd/ateapi/internal/store/ateredis/ateredis.go
+++ b/cmd/ateapi/internal/store/ateredis/ateredis.go
@@ -57,6 +57,7 @@ import (
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
 	"github.com/google/uuid"
 	"github.com/redis/go-redis/v9"
+	"go.opentelemetry.io/otel/metric"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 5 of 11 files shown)


orangeCatDeveloper:
Obsolete after #940 replaced the Redis store with atepg.

The drop-and-invalidate path this PR added lived in `ateredis.WatchWorkers`, which no longer exists. `atepg.WatchWorkers` never drops: it blocks on a bounded channel and closes the watch outright when events could be lost (postmaster restart, retention gap, unmarshal failure), and the consumer already resyncs on a closed channel. Rebasing this onto main leaves `WorkerWatch.Invalidated` with no producer.

Closing rather than porting: the failure mode it guards against does not exist in the current store.

Issue #277, opened 2026-06-18
Add time metadata to actors


juli4n:
In many cases it's useful to have some time metadata for actors (e.g. created / updated time). Today there are no fields to track this.

- `create_time`: Timestamp when actor is created. Immutable.
- `update_time`: Last time the actor was modified in any way (suspended, resumed, updated, etcCanc).
- `last_resume_time`: Timestamp of the most recent transition into RUNNING.
- `total_running_duration`: Cumulative time the actor spent running, since creation. Never resets. (this could be tricky to get right with high resolution if ate-apiservers have some clock skew)

The https://google.aip.dev/142 has some guidelines on how to deal with times in gRPC APIs. We should probably use the `google.protobuf.Timestamp` and `google.protobuf.Duration` well-known protos.

thockin:
Should create/update just be standard metadata on all "resource" messages?

juli4n:
I agree. Create and update should be across all resources.

EItanya:
I added this as a sub issue to the general substrate gRPC style guide

PR #950, opened 2026-08-14
feat(ateapi): record last_resume_time on actors

orangeCatDeveloper:
Part of #277

Actors carry no record of when they last started running, so there is no way to
answer "how long has this actor been up" from the API.

## What this adds

A `last_resume_time` timestamp on `Actor`, stamped at the single place an actor
transitions into RUNNING:

```go
// workflow_resume.go, finalizeRunning
toUpdate.Status = ateapipb.Actor_STATUS_RUNNING
toUpdate.LastResumeTime = timestamppb.Now()
```

It is not cleared on suspend or pause — the field answers "when did it last come
up", and stays queryable while the actor is down.

## Scope

#277 lists four fields. `create_time` and `update_time` already exist on
`ResourceMetadata`, leaving two; this covers `last_resume_time`.

`total_running_duration` is deliberately left out. Unlike an absolute timestamp
written by one replica, a cumulative duration is a subtraction across replicas —
resume stamps on replica A, suspend computes `now(B) - last_resume_time(A)` —
so clock skew can produce a negative delta that accumulates permanently into a
field documented as never resetting. It also needs accounting at all six exits
from RUNNING (crash, syncer crash, suspending, pausing, pause-failure crash,
deleting), and the cut-off for an actor killed without a clean transition is
undefined. Those are design questions, not implementation detail.

Rollout: ateredis unmarshals with strict protojson, so roll all ateapi replicas
forward together, and clear `lastResumeTime` from stored actors before any
rollback.

## Tests

`TestFinalizeRunning_SetsLastResumeTime` covers the stamp.
`TestEnsureSuspendedFinalized_KeepsLastResumeTime` pins the preservation
behaviour — it fails with `LastResumeTime = 1970-01-01 ..., want 2026-01-01 ...`
if the suspend path drops the field.

### DIFF
--- a/cmd/ateapi/internal/controlapi/functionaltest/actor_test.go
+++ b/cmd/ateapi/internal/controlapi/functionaltest/actor_test.go
@@ -1679,7 +1679,7 @@ func TestResumeActor(t *testing.T) {
 			},
 		},
 	}
-	if diff := cmp.Diff(want, getResp, protocmp.Transform(), ignoreUID, ignoreVersion, ignoreTimestamps); diff != "" {
+	if diff := cmp.Diff(want, getResp, protocmp.Transform(), ignoreUID, ignoreVersion, ignoreTimestamps, ignoreResumeTime); diff != "" {
 		t.Errorf("GetActor response mismatch (-want +got):\n%s", diff)
 	}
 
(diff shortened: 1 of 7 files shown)


orangeCatDeveloper:
This change is not safe during a rolling ateapi deployment. A new replica writes last_resume_time into the Redis proto JSON, while an old replica rejects that unknown field and can no longer read the actor. Solving that cleanly requires either a compatibility change deployed before this feature or a different storage design, so I am closing this PR rather than expanding it into a much larger change.

PR #947, opened 2026-08-14
Align ActorSnapshot/ActorSnapshotTag store methods to other CRUD methods.

juli4n:
* Align ActorSnapshot/ActorSnapshotTag store methods to other CRUD methods.
* Add common struct types for List<Type> methods to carry parameters and returned pages.
* Sort methods in store interface to group methods for each resource.
* Remove pessimistic locking from ActorSnapshot/ActorSnapshotTag RPC handlers.
* Align CreateActorSnapshotTag method to API guideline. It should take just the resource, no extra top level field.

PR #952, opened 2026-08-14
Extract validations for actor snapshot rpc handlers into dedicated fns

laoj2:
So, the resource RPC methods are consistent

- [x] Tests pass


laoj2:
already being done in https://github.com/agent-substrate/substrate/pull/947

PR #955, opened 2026-08-14
Split store contract tests into a function per resource

laoj2:
RunContractTests is getting large and resource tests are not always grouped together. So, let's start with splitting them into separate functions. As they grow, we may consider moving them to their own files, like we did for functional tests in https://github.com/agent-substrate/substrate/pull/918

- [x] Tests pass


juli4n:
Needs to resolve conflicts.

juli4n (approved):

PR #956, opened 2026-08-14
identity: Secure CSI Driver TCP Connections via Pod Identity

hajiler:
Secures network-based gRPC connections between the ate-api control plane and CSI driver controller services using mutual TLS (mTLS) backed by Substrate's Pod Identity. 

  
Requests from ate-api to CSI driver controllers occur over the network, as configured via the CSIDriverConfig custom resource. This PR extends the CSIDriverConfig CRD to include a tls configuration section, which leverages
Substrate's Pod Identity certificate infrastructure (managed by podcertcontroller) to secure these
TCP connections using mTLS. Because standard CSI drivers typically communicate over Unix domain sockets, a proxy sidecar (e.g., envoy) is added to the CSI controller deployment to terminate TLS and forward requests to the local CSI socket.

#232 

- [X ] Tests pass


mattcary:
/lgtm

Seems to do the right thing from what I understand.

hack/third_party/csi-driver-host-path/deploy/csi-hostpath-testing.yaml
ahmedtd: Seems like it is no longer done with socat?

hack/third_party/csi-driver-host-path/deploy/csi-hostpath-testing.yaml
```diff
@@ -7,6 +7,65 @@
...
+                - certificate_chain: { filename: /run/servicedns/credential-bundle.pem }
+                  private_key: { filename: /run/servicedns/credential-bundle.pem }
+                  watched_directory: { path: /run/servicedns }
+                validation_context:
+                  trusted_ca: { filename: /run/podidentity-ca/trust-bundle.pem }
+                  watched_directory: { path: /run/podidentity-ca }
+                  match_typed_subject_alt_names:
+                  - san_type: URI
+                    matcher:
+                      exact: "spiffe://cluster.local/ns/ate-system/sa/ate-api-server"
```
ahmedtd, line 49: In the near future we are going to have to make the cluster.local part configurable.  What sort of matches does Envoy support in its config?

internal/volume/csi/plugin.go
```diff
@@ -288,3 +323,41 @@ func NewCSIPlugin(ctx context.Context, lister listersv1alpha1.CSIDriverConfigLis
...
+
+	tlsCfg := cfg.Spec.TLS
+
+	if !tlsCfg.UsePodIdentity {
+		// TODO: Support manual certificates loaded from Secrets specified in the config.
+		return nil, fmt.Errorf("only pod identity TLS is supported in this configuration")
+	}
+
+	// Load CA Pool
+	caPoolBytes, err := os.ReadFile(paths.caCert)
```
ahmedtd: The CA file may update on disk if the CA goes through a rotation.

We may need to copy the CA handling code from https://github.com/ahmedtd/tinycert/blob/main/lib/spiffefsd/spiffefsd.go (I wrote this under the terms of my IP agreement with Google, and it's Apache-licensed, so it's fine to copy in and attribute to Google)

By default tls.Config only allows specifying a static pool of root certificates.  We have to use the combination of InsecureSkipVerify and VerifyConnection to enable automatic reloading.

hack/third_party/csi-driver-host-path/deploy/csi-hostpath-testing.yaml
hajiler: Envoy supports both an exact suffix and regex match, so I think we can accomodate future configuration changes once they are decided.

internal/volume/csi/plugin.go
hajiler: Thank you for the reference, I worked https://github.com/ahmedtd/tinycert/blob/main/lib/spiffefsd/spiffefsd.go  into resolveTLSConfig, to verify dynamically reload CA bundle using verifyConnection.

hack/third_party/csi-driver-host-path/deploy/csi-hostpath-testing.yaml
hajiler: Fixed to envoy.

internal/volume/csi/plugin.go
```diff
@@ -288,3 +323,82 @@ func NewCSIPlugin(ctx context.Context, lister listersv1alpha1.CSIDriverConfigLis
...
+		// we set InsecureSkipVerify=true and verify the server certificate chain dynamically
+		// against the latest CA bundle read from disk in VerifyConnection.
+		InsecureSkipVerify: true,
+		VerifyConnection: func(state tls.ConnectionState) error {
+			if len(state.PeerCertificates) == 0 {
+				return fmt.Errorf("server did not present certificates")
+			}
+
+			// Read CA trust bundle on each TLS connection handshake.
+			roots, err := getCertPool(paths.caCert)
```
ahmedtd, line 369: This is fine for now, but we will need to add caching pre-GA.

ahmedtd (approved):

PR #958, opened 2026-08-14
hack: fix kubeconfig and pod DNS on IPv6-only kind clusters

### LINKED ISSUE #1099: hack: the ipv6 kubeconfig repoint breaks kubectl on a native-Linux Docker host
`hack/create-kind-cluster.sh:148-160` rewrites kind's `https://[::1]:PORT` kubeconfig entry to
`https://localhost:PORT` whenever `IP_FAMILY=ipv6`, on the reasoning in the comment that `localhost`
"lets the client pick a family that works from either side". It cannot: on Debian-family hosts
`/etc/hosts` gives `::1` the names `ip6-localhost` and `ip6-loopback`, not `localhost`, so
`localhost` resolves v4-only while the apiserver is published on v6 only.

Measured on Ubuntu 25.10 with a native Docker daemon: kind publishes `6443/tcp -> [::1]:42311`,
`https://[::1]:42311/healthz` answers `ok`, and `https://localhost:42311/healthz` is refused. The
repoint replaces a working address with a broken one and every later `kubectl` in the script dies at
connect. It is only needed when the daemon lives in a VM (Lima on macOS), which is the topology
every IPv6-only run in this project has used so far — hence it went unnoticed. Introduced in #877;
#958 makes the repoint conditional on `[::1]` failing to answer, and this issue tracks landing that
on main if #958 does not.

ygao-g:
Fixes #1099.
Part of #246.

An IP_FAMILY=ipv6 cluster created by hack/create-kind-cluster.sh fails in
two independent ways:

1. The script rewrites kind's https://[::1]:PORT kubeconfig entry to
   https://localhost:PORT, which breaks Debian-family hosts: there,
   localhost resolves IPv4-only while the apiserver is published on v6
   only (#1099). 

2. Nothing resolves from inside a pod. CoreDNS inherits the node's IPv4
   resolver, which a v6-only pod cannot reach. Forward to an overridable
   IPv6 upstream instead, and add a kind-registry:53 server block so atelet
   can pull images from its own network namespace.

The CoreDNS patch works around kubernetes-sigs/kind#4152 — kind's node
entrypoint picks the IPv4 default gateway as the node's nameserver — and
moby/moby#41651 behind it. Neither has a fix committed, so the patch stays
until kind runs a resolver proxy on the node.

Manually verified in a Lima guest on macOS: IP_FAMILY=ipv6 comes up clean, kind's
untouched [::1] kubeconfig answers, and from a v6-only pod www.google.com,
kubernetes.default.svc.cluster.local and kind-registry. all resolve.

CI cannot verify
this by itself: GitHub runners have no IPv6 egress, so an IPv6-only lane
additionally needs the DNS64/NAT64 that #1275 stacks on this PR.

### DIFF
--- a/hack/create-kind-cluster.sh
+++ b/hack/create-kind-cluster.sh
@@ -156,21 +156,6 @@ if [[ "${IP_FAMILY}" != "ipv4" &&
   exit 1
 fi
 
-# For ipv6 kind writes a kubeconfig pointing at [::1], the address it published
-# the apiserver on, which only works for a client on the Docker host itself: a
-# VM-hosted daemon (Lima on macOS) forwards the port to the *v4* loopback, so
-# every kubectl below fails at connect. localhost is a SAN on the apiserver
-# cert and lets the client pick a family that works from either side.
-if [[ "${IP_FAMILY}" == "ipv6" ]]; then
-  server="$(kubectl config view \
-    -o jsonpath="{.clusters[?(@.name==\"${KUBECTL_CONTEXT}\")].cluster.server}")"
-  if [[ "${server}" == "https://[::1]:"* ]]; then
-    echo "Repointing the kubeconfig for '${KUBECTL_CONTEXT}' at localhost..."
-    kubectl config set-cluster "${KUBECTL_CONTEXT}" \
-      --server="https://localhost:${server##*:}" >/dev/null
-  fi
-fi
-
 # 2.5 Enable Proxy ARP/NDP on kind nodes for gVisor loopback pod-to-pod networking
 echo "Enabling Proxy ARP/NDP on kind nodes..."
 for node in $("${ROOT}"/hack/kind.sh get nodes --name "${KIND_CLUSTER_NAME}"); do
--- a/hack/create-kind-cluster.sh
+++ b/hack/create-kind-cluster.sh
@@ -21,6 +21,7 @@ KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
 KUBECTL_CONTEXT="kind-${KIND_CLUSTER_NAME}"
 reg_name="kind-registry"
 reg_port="${KIND_REGISTRY_PORT:-5001}"
+IPV6_DNS_UPSTREAM="${IPV6_DNS_UPSTREAM:-2001:4860:4860::8888 2001:4860:4860::8844}"
 
 if [[ $# -gt 0 ]]; then
   case "$1" in
@@ -31,6 +32,9 @@ if [[ $# -gt 0 ]]; then
       echo "Configured through the environment:"
       echo "  KIND_CLUSTER_NAME  Name of the cluster to create (default: kind)."
       echo "  IP_FAMILY          Address families for pods and Services: ipv4, ipv6 or dual (default: ipv4)."
+      echo "  IPV6_DNS_UPSTREAM  Space-separated IPv6 resolvers CoreDNS forwards to when IP_FAMILY=ipv6"
+      echo "                     (default: Google Public DNS). These replace the host's resolver, so any"
+      echo "                     split-horizon names it served stop resolving from pods."
       exit 0
       ;;
   esac
@@ -146,6 +150,23 @@ fi
 echo "Creating kind cluster '${KIND_CLUSTER_NAME}'..."
 "${ROOT}"/hack/kind.sh create cluster --name "${KIND_CLUSTER_NAME}" --config "${ROOT}/bin/kind-config.yaml"
 
+# kind create returns before the apiserver answers, and every kubectl below races
+# it. Poll from inside the node, where the answer does not depend on how the
+# daemon published the port.
+echo "Waiting for the control plane to answer..."
+for attempt in $(seq 60); do
+  if docker exec "${KIND_CLUSTER_NAME}-control-plane" \
+    kubectl --kubeconfig=/etc/kubernetes/admin.conf get --raw /healthz >/dev/null 2>&1; then
+    break
+  fi
+  if [[ "${attempt}" == 60 ]]; then
+    echo "error: the control plane did not answer /healthz within 2m of create:" >&2
+    echo "         docker logs ${KIND_CLUSTER_NAME}-control-plane" >&2
+    exit 1
+  fi
+  sleep 2
+done
+
 # A daemon with IPv6 off hands kind a v4-only network whatever it asked for.
 if [[ "${IP_FAMILY}" != "ipv4" &&
       "$(docker network inspect kind --format '{{.EnableIPv6}}')" != "true" ]]; then
@@ -190,6 +211,50 @@ done
 echo "Connecting local registry to cluster network..."
 if [ "$(docker inspect -f='{{json .NetworkSettings.Networks.kind}}' "${reg_name}")" = "null" ]; then
   docker network connect "kind" "${reg_name}"
+fi
+
+# 4.5. Point CoreDNS at an IPv6 resolver and teach it the registry's name
+if [[ "${IP_FAMILY}" == "ipv6" ]]; then
+  echo "Repointing CoreDNS at an IPv6 resolver and teaching it '${reg_name}'..."
+  reg_v6="$(docker inspect "${reg_name}" \
+    --format '{{.NetworkSettings.Networks.kind.GlobalIPv6Address}}' 2>/dev/null || true)"
+  if [[ -z "${reg_v6}" ]]; then
+    echo "error: '${reg_name}' has no IPv6 address on the 'kind' network" >&2
+    exit 1
+  fi
+
+  # CoreDNS runs dnsPolicy: Default and inherits the node's IPv4 resolver, which
+  # no pod here can reach.
+  corefile="$(kubectl --context="${KUBECTL_CONTEXT}" -n kube-system get cm coredns \
+    -o jsonpath='{.data.Corefile}')"
+  search="forward . /etc/resolv.conf"
+  # $search unquoted: bash 3.2 splices the quotes in literally.
+  patched="${corefile/$search/forward . ${IPV6_DNS_UPSTREAM}}"
+  if [[ "${patched}" == "${corefile}" ]]; then
+    echo "error: '${search}' not found in the CoreDNS Corefile" >&2
+    echo "       the Corefile layout changed upstream; update this block" >&2
+    exit 1
+  fi
+
+  # Step 3's registry wiring is node-side, while atelet pulls from its own netns,
+  # where "kind-registry" does not resolve. Own zone, so no fallthrough is needed:
+  # only this name reaches the hosts stanza.
+  patched="${patched}
+${reg_name}:53 {
+    errors
+    hosts {
+        ${reg_v6} ${reg_name}
+    }
+}"
+
+  # A YAML patch file avoids escaping the Corefile's newlines into JSON.
+  { printf 'data:\n  Corefile: |\n'; printf '%s\n' "${patched}" | sed 's/^/    /'; } \
+    > "${ROOT}/bin/coredns-patch.yaml"
+  kubectl --context="${KUBECTL_CONTEXT}" -n kube-system patch cm coredns \
+    --type=merge --patch-file "${ROOT}/bin/coredns-patch.yaml"
+  kubectl --context="${KUBECTL_CONTEXT}" -n kube-system rollout restart deploy/coredns
+  kubectl --context="${KUBECTL_CONTEXT}" -n kube-system rollout status deploy/coredns \
+    --timeout=120s
 fi
 
 # 5. Document the local registry in kube-public ConfigMap


ygao-g:
## Testing

Manual Testing: `IP_FAMILY=ipv6 hack/create-kind-cluster.sh` → `--deploy-ate-system` → `--deploy-demo-counter`: atelet pulls the gVisor tarball from `storage.googleapis.com` over IPv6 from its own pod netns, the actor image pulls through `kind-registry`, and a golden actor runs and checkpoints. 

CI has no IPv6 job and this PR does not add one — [#1065 ](https://github.com/agent-substrate/substrate/pull/1065) is a signal-only draft that produces that coverage on demand.

ygao-g:
Ping — this one is a two-file `hack/` change, and it unblocks the IPv6-only CI job in #939, which carries it verbatim and can't merge first.

Worth knowing that the branch has moved since it was last looked at. Two commits went on, both fixes for defects I hit while trying to run it rather than responses to review: the DNS gate was reporting green with one of kind's two CoreDNS replicas unable to resolve, and the IPv6-only kubeconfig repoint was replacing a working apiserver address with a refused one on a native-Linux Docker host, which killed every `kubectl` after it.

hack/create-kind-cluster.sh
```diff
@@ -146,14 +149,18 @@ if [[ "${IP_FAMILY}" != "ipv4" &&
...
-# VM-hosted daemon (Lima on macOS) forwards the port to the *v4* loopback, so
-# every kubectl below fails at connect. localhost is a SAN on the apiserver
-# cert and lets the client pick a family that works from either side.
+# the apiserver on. That works for a client on the Docker host itself but not
+# behind a VM-hosted daemon (Lima on macOS), which forwards the port to the *v4*
+# loopback. localhost is a SAN on the apiserver cert and covers that case -- but
+# only where it resolves to ::1, and Debian-family hosts give ::1 the names
+# ip6-localhost and ip6-loopback, not localhost. Repointing unconditionally
+# therefore trades a working address for a broken one on a Linux host. Ask the
+# apiserver instead of guessing from the platform.
```
ygao-g: Nit. be more  concise about the issue it tries to solve, and what the solution is

hack/create-kind-cluster.sh
```diff
@@ -196,6 +203,61 @@ if [ "$(docker inspect -f='{{json .NetworkSettings.Networks.kind}}' "${reg_name}
...
+  if [[ -z "${reg_v6}" ]]; then
+    echo "error: '${reg_name}' has no IPv6 address on the 'kind' network" >&2
+    exit 1
+  fi
+
+  corefile="$(kubectl --context="${KUBECTL_CONTEXT}" -n kube-system get cm coredns \
+    -o jsonpath='{.data.Corefile}')"
+  search="forward . /etc/resolv.conf"
+  replace="forward . ${IPV6_DNS_UPSTREAM}"
+  # $search unquoted: bash 3.2 splices the quotes in literally. Replacing just
```
ygao-g: Nit. consider removing it

hack/create-kind-cluster.sh
```diff
@@ -196,6 +203,61 @@ if [ "$(docker inspect -f='{{json .NetworkSettings.Networks.kind}}' "${reg_name}
...
+  # the target leaves kind's trailing "{ max_concurrent 1000 }" in place.
+  patched="${corefile/$search/$replace}"
+  if [[ "${patched}" == "${corefile}" ]]; then
+    echo "error: '${search}' not found in the CoreDNS Corefile" >&2
+    echo "       the Corefile layout changed upstream; update this block" >&2
+    exit 1
+  fi
+
+  # Its own server block, not a hosts entry in .:53. A query is served by the
```
ygao-g: Nit. consider removing this

hack/create-kind-cluster.sh
```diff
@@ -196,6 +203,61 @@ if [ "$(docker inspect -f='{{json .NetworkSettings.Networks.kind}}' "${reg_name}
   docker network connect "kind" "${reg_name}"
 fi
 
+# 4.5. Give CoreDNS an IPv6 forwarder and a registry entry
+#
+# CoreDNS runs dnsPolicy: Default, inheriting the node's IPv4 resolver, which
```
ygao-g: nit. make comments more concise, briefly mention the issue and the fix. IIUC, what this logic did is to replace "forward . /etc/resolv.conf" with "forward . ${IPV6_DNS_UPSTREAM}", and add registry.

hack/verify-ipv6-dns.sh
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env bash
```
ygao-g: Nit. remove this file, we can rely on e2e tests to provide signals

hack/create-kind-cluster.sh
ygao-g: Cut to four lines: what breaks, and why probing beats guessing. The Debian `ip6-localhost` detail lives in the commit message now.

The logic changed with it — it probes `[::1]` first and only falls back to `localhost` if that fails, so a native Linux daemon (where `[::1]` works and `localhost` is refused) is no longer broken by the rewrite.

hack/create-kind-cluster.sh
ygao-g: Yes, that is exactly what it does. Header is down to three lines naming the two causes.

hack/create-kind-cluster.sh
ygao-g: Kept, as one line. Quoting `$search` is the natural-looking tidy-up and it stops the substitution matching, which then trips the guard below with a message blaming an upstream Corefile change — a misdiagnosis rather than a clean failure. Happy to drop it if you would still rather.

hack/create-kind-cluster.sh
ygao-g: Kept, shortened to three lines. Without the rationale the obvious simplification is to fold the `hosts` into `.:53`, which then needs `fallthrough` or every other name NXDOMAINs. Same offer as above — say the word and it goes.

hack/verify-ipv6-dns.sh
ygao-g: File is gone; the check is ~25 lines inlined at the end of step 4.5, and the per-replica and want/have legs are dropped with it.

e2e cannot replace it though. On IPv6-only this fix is a precondition of `hack/install-ate-kind.sh --deploy-ate-system`: atelet fetches the gVisor bundle from GCS and pulls through `kind-registry`, both from its own pod netns, so the install fails before any suite runs. And the only IPv6-only lane is #939, which is gated on this PR. I will add an e2e guard for the external-name half on top of #939, where there is somewhere to run it.

hack/create-kind-cluster.sh
```diff
@@ -145,18 +149,38 @@ if [[ "${IP_FAMILY}" != "ipv4" &&
...
     -o jsonpath="{.clusters[?(@.name==\"${KUBECTL_CONTEXT}\")].cluster.server}")"
-  if [[ "${server}" == "https://[::1]:"* ]]; then
+  if [[ "${server}" == "https://[::1]:"* ]] && ! apiserver_answers; then
+    alt_server="https://localhost:${server##*:}"
+    if ! apiserver_answers --server="${alt_server}"; then
+      echo "error: the apiserver answered /healthz at neither ${server} nor" >&2
+      echo "       ${alt_server}. Check where the daemon published it:" >&2
+      echo "         docker port ${KIND_CLUSTER_NAME}-control-plane 6443" >&2
+      exit 1
+    fi
```
haiyanmeng: The probe exits 1 when neither endpoint answers, but it cannot distinguish a wrong address from an apiserver that has not finished starting.

hack/create-kind-cluster.sh
```diff
@@ -145,18 +149,38 @@ if [[ "${IP_FAMILY}" != "ipv4" &&
...
-# For ipv6 kind writes a kubeconfig pointing at [::1], the address it published
-# the apiserver on, which only works for a client on the Docker host itself: a
-# VM-hosted daemon (Lima on macOS) forwards the port to the *v4* loopback, so
-# every kubectl below fails at connect. localhost is a SAN on the apiserver
-# cert and lets the client pick a family that works from either side.
+# For ipv6 kind publishes the apiserver on [::1]. Under a VM-hosted daemon (Lima
+# on macOS) that publish lands inside the VM, and limactl re-forwards it to the
+# host's *v4* loopback only, so kind's kubeconfig is unusable and has to be
+# repointed at localhost.
+# https://github.com/lima-vm/lima/issues/1540
```
haiyanmeng: What is the github issue's relationship with this PR?

hack/create-kind-cluster.sh
```diff
@@ -21,6 +21,7 @@ KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
 KUBECTL_CONTEXT="kind-${KIND_CLUSTER_NAME}"
 reg_name="kind-registry"
 reg_port="${KIND_REGISTRY_PORT:-5001}"
+IPV6_DNS_UPSTREAM="${IPV6_DNS_UPSTREAM:-2001:4860:4860::8888 2001:4860:4860::8844}"
```
haiyanmeng, line 24: IPV6_DNS_UPSTREAM defaults to Google Public DNS over IPv6 but is never probed, so an unreachable upstream leaves DNS broken while the script exits 0.

BenTheElder (commented):
@bowei we're going to have to rewrite this in go too, sigh

this is meant to represent a portable cluster, if we customize it heavily then we're really not proving substrate works generically

can we please file upstream issues with the relevant projects for each thing we're working around, so we can actually fix them?

hack/create-kind-cluster.sh
ygao-g: Do you think CI test coverage would be sufficient for this? For example, [TestActorEgress](https://github.com/agent-substrate/substrate/blob/main/internal/e2e/suites/networking/networking_test.go#L98) fetches http://example.com/

Also worth mentioning that a bad upstream only cause external lookup to fail unambiguously: `connection timed out; no servers could be reached`. In-cluster names and kind-registry. still resolve — CoreDNS serves both locally and never reaches the upstream.

hack/create-kind-cluster.sh
ygao-g: Correct, and it's moot now — the whole repoint block, probe included, is gone as of `fa366071`. The script no longer rewrites the kubeconfig on any family: kind's `https://[::1]:PORT` entry is already right for a client in the same network namespace as the Docker daemon, which is where the loop runs.

What replaced it is a bounded `/healthz` poll run inside the control-plane container, needed because `kind create` returns before the apiserver answers and step 4.5 patches the Corefile straight after. It tests one thing — readiness — so there's no address-vs-startup ambiguity left.

hack/create-kind-cluster.sh
ygao-g: That link went with the repoint block in `fa366071`, but the relationship, for the record:

We develop on macOS with Lima running the Linux host that Docker — and therefore the kind nodes — live in. limactl re-forwards a published port to the macOS **v4** loopback only, so on an IPv6-only cluster the `https://[::1]:PORT` kubeconfig kind writes is unreachable from macOS. lima-vm/lima#1540 is that upstream bug, and the repoint was the workaround for it.

This PR stops working around it instead: run the loop inside the Lima guest and `[::1]` is local and correct. A macOS client reaching kind inside the VM is the only path that ever needed the rewrite, and it isn't a path we need.

hack/verify-ipv6-dns.sh
ygao-g: Correcting this for the record: the inlined check is not at head — `b540c38e` removed it, and step 4.5 now ends at `kubectl rollout status deploy/coredns` with no pod-level DNS probe in the script.

The e2e guard did land, though, on #939: it resolves `kubernetes.default.svc.cluster.local`, an external name, and `kind-registry.` from a pod on the IPv6-only lane. That's where the signal lives now, and it's the basis for not gating `IPV6_DNS_UPSTREAM` in the script — see the line-24 thread.

hack/create-kind-cluster.sh
ygao-g: The pod-level probe is already in the CI lane (#939): it resolves an in-cluster Service, an external name, and `kind-registry.` from a pod before the install. But the GitHub runner (`ubuntu-latest`) has no IPv6 egress at all, so the lane can only reach an upstream through NAT64 — `hack/setup-nat64.sh` from #1275, with `IPV6_DNS_UPSTREAM` overridden to `64:ff9b::808:808`.

So CI covers the mechanism, not this default — `2001:4860:4860::8888` stays unprobed.

haiyanmeng: I see. Thanks for the explanation.

ygao-g:
> Bowei Du (@bowei) we're going to have to rewrite this in go too, sigh
> 
> this is meant to represent a portable cluster, if we customize it heavily then we're really not proving substrate works generically
> 
> can we please file upstream issues with the relevant projects for each thing we're working around, so we can actually fix them?

For the workaround in this PR, found existing kind issue github.com/kubernetes-sigs/kind/issues/4152 and added a comment there.

ygao-g:
@bowei This PR is ready for you to review.

PR #961, opened 2026-08-14
atelet: poll ateom workload stats into template-level metrics

baizhenyu:
Second piece of #896 (Phase 1 of #550): atelet polls every local ateom's `GetActiveWorkloadStats` and turns the samples into template-level metrics — the TSDB half of #174's cardinality split. After this PR, "how much CPU and memory is this template using" is a Cloud Monitoring query.

## The reader

A `statsPoller` in atelet, driven by `--actor-stats-poll-interval` (default 1m; `0` disables; nonzero values are clamped to a 50s floor, the worst-case duration of sampling one micro-VM ateom — 25 containers × 2s per guest-agent call).

- **Stateless discovery**: each tick lists `ateoms/*` (entry names are worker pod UIDs — the same sockets the lifecycle RPCs dial) and probes each with the parameterless discovery read. No worker-to-actor mapping, no control-plane dependency, nothing to recover after an atelet restart. Attribution comes solely from the identity echoed in each sample, per the RPC's contract.
- **One tolerance rule**: any dial or call failure means "not a target this tick" — covering stale directories of deleted workers, ateoms that have not started listening, and teardowns mid-poll. `NO_WORKLOAD` / `NOT_MEASURABLE_YET` are skips by the RPC's own contract.
- **Bounded concurrency**: distinct ateoms are probed 8 at a time (one probe per guest, so nothing the interval floor defends against is reintroduced); a node of stuck-but-accepting sockets degrades to ceil(n/8) timeouts instead of n sequential ones.
- **WorkerPool enrichment**: one field+label-selected pod LIST per tick maps pod UID → owning pool (`ate.dev/worker-pool`, the label the pool controller stamps); unresolved pods group without pool labels rather than vanish. Chosen over an informer deliberately — negligible apiserver cost at this cadence, no cache-sync ordering, and the resolver sits behind a function seam if that trade ever changes. Needs one new Downward API env (`NODE_NAME`); the pods RBAC already existed.

## The metrics

Labels on every series: `ate.template.namespace/name`, `ate.sandbox.class`, `ate.stats.source`, `ate.workerpool.namespace/name` — all bounded sets; actor and atespace identity never reach a metric label (they belong to the events channel, the next PR).

- `ate.actor.stats.sampled_actors`, `…memory_current_bytes`, `…memory_working_set_bytes` — **observable gauges** over the latest tick's snapshot: each collection observes exactly the groups that currently exist, so a template whose actors leave a node disappears from the export. (Synchronous gauges would re-export their last value until process exit — stale memory for actors long gone.)
- `ate.actor.stats.cpu_usage` — **Float64Counter in seconds** (cAdvisor / OTel `*.cpu.time` convention; the wire stays µs). The raw `cpu_usage_usec` is cumulative per-epoch per actor, so the poller accumulates per-sweep *increases* against per-actor baselines: first sight establishes a baseline and charges nothing (atelet cannot tell a new actor from its own restart — re-charging epochs the previous process counted would spike `rate()`), a decrease is an epoch reset charged from the new value, and baselines are pruned to the actors seen. Bounded imprecision (≤1 interval per actor across restarts; the pre-checkpoint tail) is documented on the instrument; per-actor precision arrives with the lifecycle events.

## Validated live on ate-dev

Both source families, simultaneously, with one actor per class:

image

The same counter application reads 5-6× larger through the cgroup source (whole sandbox: sentry heap, netstack, gofers) than through the guest agent (workload containers only) — the concrete case for the `ate.stats.source` label and its group-don't-sum rule. Also exercised live: pool labels resolved on every series, restart-without-spike on the CPU counter across a DaemonSet rollout, and OTLP delivery to the gke-managed-otel collector with zero export errors.

## Out of scope

- Per-actor events + lifecycle first/final samples — next PR per #896.
- `k8s.node.name` resource attribute (one manifest line, any time).
- e2e coverage — the metrics e2e suite is the natural home once the events channel lands.

Part of #896, toward #550.

### DIFF
--- a/cmd/atelet/statspoller.go
+++ b/cmd/atelet/statspoller.go
@@ -1,0 +1,540 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/agent-substrate/substrate/internal/ateattr"
+	"github.com/agent-substrate/substrate/internal/ateompath"
+	"github.com/agent-substrate/substrate/internal/proto/ateompb"
+)
+
+// workerPoolLabel is the label the pool controller stamps on every worker pod
+// it creates; the value is the WorkerPool's name and the pod's namespace is
+// the pool's. Must agree with cmd/atecontroller/internal/controllers/
+// workerpool_apply.go.
+const workerPoolLabel = "ate.dev/worker-pool"
+
+// minActorStatsPollInterval is the floor a configured poll interval is clamped
+// to. It is the worst-case duration of one ateom's sweep on the micro-VM
+// runtime -- maxActorContainers containers at statsCallTimeout each, constants
+// that live with that runtime -- so a shorter interval could start a new poll
+// into a guest agent still serving the previous one.
+const minActorStatsPollInterval = 50 * time.Second
+
+// statsRPCTimeout bounds one ateom's GetActiveWorkloadStats call. It has to
+// cover the ateom's own worst-case sweep (see minActorStatsPollInterval);
+// anything still unanswered past that is a stuck socket, not a slow guest.
+const statsRPCTimeout = 55 * time.Second
+
+// statsSweepConcurrency bounds how many ateoms one sweep probes at once. The
+// interval floor protects a single guest from overlapping polls; probing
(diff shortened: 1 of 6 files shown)


git286 (changes requested):
I just found kubelet is leaking the gRPC connections: https://github.com/agent-substrate/substrate/pull/1010 while reviewing this PR.

It makes think that for polling, we should not use the lru cache because it will scan all the socket files even those from the inactive pod. For a node that run a long time, there could be many old worker pod socket files not cleaned up (which is probably a separate issue). At least for polling, I think a fresh connection per probe, closed right after is probably a better design. (LRU cache should be left for lifecycle RPCs and poller gains nothing from caching anyway)

baizhenyu:
Adopted in 38977fde, and agreed it's stronger than "gains nothing": sweeping the node's stale sockets through the shared LRU would let telemetry evict connections the lifecycle RPCs are using — same poller-must-not-hurt-the-node principle as the earlier Fatal removal. The poller now dials a fresh connection per probe and closes it before the probe returns; it no longer references AteomDialer at all, so the isolation is structural. Tests pin the close contract including the RPC-failed path (the #1010 leak shape), and measured live on ate-dev: atelet's open-fd count is flat (12→13→13) across three sweep cycles.

Agreed stale ateom dirs are a separate issue — happy to file it.

cmd/atelet/statspoller.go
```diff
@@ -0,0 +1,534 @@
...
+				namespace: pod.Namespace,
+				name:      pod.Labels[workerPoolLabel],
+			}
+		}
+		return pools
+	}
+}
+
+const (
+	sampledActorsMetric = "ate.actor.stats.sampled_actors"
```
git286, line 402: Let's make sure these names follow the otel convention, or example, for containers, the otel convention is

container.cpu.time 
container.memory.usage
container.cpu.usage

so, maybe

ate.actor.stats.memory.usage
ate.actor.stats.memory.working_set
ate.actor.stats.cpu.time
etc

baizhenyu: Adopted in the updated commit: memory.usage, memory.working_set, cpu.time — the _bytes suffixes were also violating the units-not-in-names rule, so this fixes two things at once. Note cpu.time rather than cpu.usage, since ours is the cumulative counter (semconv cpu.usage is the instantaneous gauge). Prometheus rendering barely moves because exporters re-attach units: memory.working_set renders character-identical to before, and cpu.time becomes ate_actor_stats_cpu_time_seconds_total. Kept sampled_actors: no unit in the name, underscore leaf matches working_set, and the plural has in-repo precedent (ate.workerpool.workers).

git286 (approved):

PR #965, opened 2026-08-14
ateapi: authorize MintCert against the store, not the worker cache

### LINKED ISSUE #964: ateapi: MintCert flakes on stale worker-cache assignments
## Summary

Resume intermittently fails at its first `MintCert` call even though the worker→actor assignment has already been committed to the authoritative store.

The failure is flaky: the authorization decision reads a replica-local `workercache.Cache` updated by an asynchronous watch, while `ResumeActor` writes the assignment directly to the store. A normal propagation delay can therefore make a valid assignment appear absent.

## CI failures

Representative error:

```
failed to resume Actor again: rpc error: code = PermissionDenied desc = mint actor certificate:
rpc error: code = PermissionDenied desc = caller is not permitted to mint credentials for this actor
```

Recent examples across unrelated PRs:

- [run 31628573958, attempt 1](https://github.com/agent-substrate/substrate/actions/runs/31628573958/attempts/1), `e2e-test (cert)` — failed, then a rerun of the same workflow succeeded
- [run 31836887339](https://github.com/agent-substrate/substrate/actions/runs/31836887339), `e2e-test (token)`
- [run 31833601999](https://github.com/agent-substrate/substrate/actions/runs/31833601999), `e2e-test (token)`
- [run 31832037757](https://github.com/agent-substrate/substrate/actions/runs/31832037757), `e2e-test (cert)`
- [run 31830638113](https://github.com/agent-substrate/substrate/actions/runs/31830638113), `e2e-test (cert)`

A local serial Full/Full experiment also reproduced the same `PermissionDenied` twice in 50 runs, so E2E test parallelism is not required.

## Representative timeline

From run 31628573958, using each RPC's logged elapsed time to derive its start time:

```
18:50:06.844197  ateapi begins ResumeActor
18:50:06.844788  selects the worker and commits UpdateWorker
18:50:06.847159  atelet Restore begins
18:50:06.852643  ateom RestoreWorkload begins
18:50:06.856462  MintCert begins
18:50:06.856480  ActorIdentity denies: worker has no actor assignment
18:50:06.858441  ResumeActor fails at CallAteletRestore
```

The authoritative write precedes the denied read by only **11.7ms**. No event loss is required.

## Root cause

`ResumeActor` commits the worker assignment to the store before asking atelet to restore. Restore immediately mints actor credentials because the workload must have its identity before startup.

`MintCert`, however, authorizes using the worker cache. The store write and cache watch delivery are not synchronized, and requests can reach either ateapi replica. If the selected replica has not applied the worker update yet—or has not observed the worker entry at all—it denies a valid mint. There is no retry on this path, so the transient cache state fails the entire resume.

Reading the store instead is not a small substitution. A worker's key there is `(namespace, pool, pod)`, while a mint request carries only `(namespace, pod, podUID)`, so the cache is the only thing that can supply the missing pool — which is what puts an asynchronously updated copy on the authorization path in the first place. The pool earns nothing as part of that key: Kubernetes already guarantees Pod names are unique within a namespace, and no query, prefix scan or bulk operation depends on it.

## Deterministic reproduction

1. Start a worker cache whose initial relist sees an unassigned worker.
2. Prevent its worker watch from delivering subsequent updates.
3. Write an assignment for that worker to the authoritative store.
4. Confirm `store.GetWorker` returns the assignment.
5. Call `MintCert`.

The mint is denied because authorization reads the stale cache, even though the assignment is committed.

## Expected behavior

Once the assignment is committed, `MintCert` must authorize from authoritative worker state without waiting for asynchronous cache propagation. The authorization must continue to verify the caller node, Pod UID, actor UID, and the reciprocal actor↔worker placement.

orangeCatDeveloper:
Fixes #964

Resume can deny `MintCert` after the worker assignment is committed because authorization reads an asynchronously updated worker cache. The issue has representative CI failures and a reproduction.

## Design decisions

- **Authorize from the store, not the worker cache.** Credential minting must use committed assignment state. Cache retries only reduce the stale window; a version barrier would add coordination with the replica's watch. A direct store read avoids both. Authorization already reads the actor from the store, so this adds one worker point read rather than a new service dependency. Store failures return `Internal`; an authoritative mismatch returns `PermissionDenied`.

- **Identify a worker by `(namespace, pod)`.** Kubernetes makes a Pod name unique within its namespace. Pool identifies ownership and configuration, not the Pod itself. Keeping pool in the key would require carrying that redundant attribute into `MintCert`; using `(namespace, pod)` lets the request locate the worker and then validates the authenticated atelet node and requested Pod UID against the stored record. Both storage backends now use the same logical identity.

- **Use Pod UID as an incarnation check, not as part of the key.** `(namespace, pod)` names the current logical worker, while Pod UID distinguishes delete-and-recreate incarnations. Keeping UID out of the key prevents multiple incarnations from coexisting; comparing it during authorization and mutation rejects stale requests and snapshots.

- **Keep `worker_pool` as an immutable attribute.** Pool still drives scheduling, metrics, WorkerPool lookup, sandbox configuration, and CLI output. It remains on the worker record, and both PostgreSQL and Redis reject changing it in place.

- **Preserve the existing relabel-as-replacement behavior.** A pool relabel runs the existing worker cleanup path, deletes the old record, and recreates it under the current pool. This PR preserves that behavior while changing the key; it does not decide whether relabel should become nondisruptive.

- **Key the workqueue by `(namespace, pod)`.** Including pool or UID would give old and new events different keys, allowing them to run concurrently on the two syncer workers. One key serializes and coalesces events for the same Pod name, and reconcile reads the latest Pod state from the informer.

- **Make deletion conditional on the record inspected.** Both ateapi replicas run the syncer, so a worker can change between read and delete. `DeleteWorker` atomically checks namespace, pool, pod, Pod UID, and version. Version alone cannot distinguish an ABA delete-and-recreate because a new record starts again at version 1.

- **Release actors only from the worker placement inspected.** Worker deletion and actor update are separate operations. Before changing actor state, the syncer compares actor UID plus namespace, pool, pod, and Pod UID. A stale snapshot therefore cannot affect an actor that has moved or a new Pod incarnation that reused the name.

The worker cache remains in scheduling and worker-count paths, where eventual consistency is acceptable; only credential authorization stops depending on it.

## Evidence

Before, in [run 31628573958, attempt 1](https://github.com/agent-substrate/substrate/actions/runs/31628573958/attempts/1):

```
18:50:06.844788  UpdateWorker commits the assignment
18:50:06.856480  ActorIdentity denies: worker has no actor assignment
18:50:06.858441  ResumeActor fails at CallAteletRestore
```

The denied read occurred 11.7ms after the authoritative write.

The regression suite covers these counterexamples:

```
MintCert reads the cache          PermissionDenied after a committed assignment
precondition checks version only  replacement incarnation or another pool is deleted
release compares ns/pod only      a stale snapshot crashes the reassigned actor
queue key includes pool           a relabel leaves two independently runnable items
PostgreSQL permits pool mutation  stored key fields and worker proto diverge
```

After, local `go test ./cmd/ateapi/... -race -v` passes the corresponding cases on this branch:

```
--- PASS: TestMintCertUsesAnAssignmentAsSoonAsItIsWritten
--- PASS: TestSyncer_StaleWorkerSnapshotDoesNotCrashReassignedActor
--- PASS: TestSyncer_PoolRelabelEnqueuesOneKey
--- PASS: TestSyncer_PoolRelabel_RebuildsRecord
  atepg    --- PASS: TestContractSuite/WorkerIdentityIsNamespaceAndPod
  atepg    --- PASS: TestContractSuite/DeleteWorker_VersionConflict
  atepg    --- PASS: TestContractSuite/DeleteWorker_RejectsReplacedIncarnation
  atepg    --- PASS: TestContractSuite/DeleteWorker_RejectsOtherPool
  ateredis --- PASS: TestContractSuite/WorkerIdentityIsNamespaceAndPod
  ateredis --- PASS: TestContractSuite/DeleteWorker_VersionConflict
  ateredis --- PASS: TestContractSuite/DeleteWorker_RejectsReplacedIncarnation
  ateredis --- PASS: TestContractSuite/DeleteWorker_RejectsOtherPool
```

## Boundaries and rollout

The project has other known flaky tests. This PR fixes only the stale-cache `MintCert` failure described in #964; the rest are not addressed here.

This PR contains no schema or data migration and no compatibility layer for the old worker identity:

- An existing PostgreSQL table keeps its `(namespace, pool, pod)` primary key. Before deploying the new code, migration must remove duplicate `(namespace, pod)` rows and replace that primary key.

- Existing Redis worker keys include pool and are not read by the new code. Their records must be rewritten. Resetting them is safe only after all actors have no live worker placement; the syncer can rebuild worker metadata from Pods, but it cannot reconstruct actor assignments.

- Old and new ateapi versions must not write the same store concurrently. Existing PostgreSQL worker page tokens also become invalid after migration.

- [x] Tests pass

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -466,6 +466,11 @@ func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (resp *
 		return nil, err
 	}
 
+	spec, err := buildAteomWorkloadSpec(req.GetSpec())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid workload spec: %v", err)
+	}
+
 	// Tell ateom to start the workload. gVisor uses RunscPath; the micro-VM
 	// runtime uses the full RuntimeAssetPaths set.
 	if _, err := client.RunWorkload(ctx, &ateompb.RunWorkloadRequest{
@@ -475,7 +480,7 @@ func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (resp *
 		ActorTemplateName:      req.GetActorTemplateName(),
 		RunscPath:              runscPathFor(assetPaths),
 		RuntimeAssetPaths:      assetPaths,
-		Spec:                   buildAteomWorkloadSpec(req.GetSpec()),
+		Spec:                   spec,
 		ActorUid:               actorUID,
 		EgressGateway:          toAteomEgressGateway(req.GetEgressGateway()),
 		CpuMilli:               req.GetCpuMilli(),
@@ -580,6 +585,11 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	// Tell ateom to take the checkpoint and delete containers. ateom reports the
 	// exact files it wrote so we ship precisely that set (gVisor's image files,
 	// cloud-hypervisor's snapshot set, ...) rather than a hardcoded list.
+	spec, err := buildAteomWorkloadSpec(req.GetSpec())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid workload spec: %v", err)
+	}
+
 	tAteom := time.Now()
 	resp, err := client.CheckpointWorkload(ctx, &ateompb.CheckpointWorkloadRequest{
 		Atespace:               actorRef.Atespace,
@@ -588,7 +598,7 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 		ActorTemplateName:      req.GetActorTemplateName(),
 		RunscPath:              runscPathFor(assetPaths),
 		RuntimeAssetPaths:      assetPaths,
-		Spec:                   buildAteomWorkloadSpec(req.GetSpec()),
+		Spec:                   spec,
 		Scope:                  toAteomSnapshotScope(req.GetScope()),
 		ActorUid:               actorUID,
 	})
@@ -601,7 +611,7 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	}
 
 	sandboxRec.SnapshotFiles = resp.GetSnapshotFiles()
-	if len(sandboxRec.SnapshotFiles) == 0 {
+	if len(sandboxRec.SnapshotFiles) == 0 && shouldHaveSnapshots(req) {
 		return nil, ateerrors.NewGRPCError(ctx, codes.DataLoss, ateerrors.ReasonInvalidCheckpointResult, ateerrors.ActorCrashedMetadata(), errors.New("ateom reported no snapshot files for checkpoint"))
 	}
 	sandboxRec.Atespace = req.GetAtespace()
@@ -691,6 +701,20 @@ func (s *AteomHerder) moveLocalCheckpoint(ctx context.Context, req *ateletpb.Che
 	}
 
 	return nil
+}
+
+// shouldHaveSnapshots returns true if the checkpoint request is expected to produce snapshot files.
+func shouldHaveSnapshots(req *ateletpb.CheckpointRequest) bool {
+	if req.GetScope() != ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA {
+		return true
+	}
+
+	for _, vol := range req.GetSpec().GetVolumes() {
+		if vol.GetType() == ateletpb.VolumeType_VOLUME_TYPE_DURABLE_DIR {
+			return true
+		}
+	}
+	return false
 }
 
 func (s *AteomHerder) uploadExternalCheckpoint(ctx context.Context, req *ateletpb.CheckpointRequest, checkpointDir string, rec *sandboxAssetsRecord) error {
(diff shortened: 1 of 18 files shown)


ahmedtd:
Hmmm, I'm not sure we want to remove caching on reads here.

In the analogous situation in Kubernetes, we return an error code indicating that the denial could be due to watch/cache lag, and the caller must retry.

@juli4n WDYT?

orangeCatDeveloper:
Thanks — could you point me to the Kubernetes mechanism you have in mind? I want to make sure I'm comparing against the same consistency model.

I have two concerns with retrying a cache-based denial here.

First, the worker cache doesn't expose a freshness watermark/resource version today, so after a stale read the caller has no way to know when the cache has observed the assignment it is waiting for. A dropped watch event (see #949) can leave the cache stale until the next successful relist, so retrying is effectively polling an eventually consistent replica without a freshness condition. And unlike the controller-style case, there isn't an independent reconciliation loop here today: MintCert is on the synchronous ResumeActor path, so retry-on-lag would introduce a new retry protocol with its own bounds and failure semantics.

More importantly, retry-on-denial only handles one direction of staleness. Suspend clears the worker's assignment in the store before clearing the actor's back-pointer. During that window, the cache can still contain the old worker→actor assignment while the actor in the store still points back to that worker, so the reciprocal check can still succeed after the authoritative worker assignment has already been cleared. There is no denial to retry in that case.

That's the main reason I was leaning toward reading the worker's authorization state from the authoritative store. There is also Kubernetes precedent for bypassing cached state when making a stale decision would be costly — e.g. GC does a live lookup before acting on deletion decisions, and NamespaceLifecycle falls back to / forces live lookup when its local namespace state may be stale. Cost-wise, minting here is lifecycle-rate rather than request-rate, and this handler already reads the actor from the store, so this is one additional point read per activation.

If we want to preserve the cache on this path, I think we'd need something stronger than a retryable denial: a freshness/version mechanism that lets the serving replica prove its cache has observed the relevant assignment state. That seems like substantially more consistency machinery than a point read here. WDYT?

orangeCatDeveloper:
My coding agent closed this PR by mistake

orangeCatDeveloper:
Can't reopen the PR due to github deadlock. Create a new one https://github.com/agent-substrate/substrate/pull/1052

PR #969, opened 2026-08-15
ateom-microvm: start the actor's two virtiofsds concurrently

BenTheElder:
~no impact if no durabledirs are used, but saves some time if they are

- [x] Tests pass


BenTheElder:
Actually, let's wind this down to one, following the rework of https://github.com/agent-substrate/substrate/pull/846

PR #971, opened 2026-08-15
localca: Rotation support part 1

ahmedtd:
This commit fleshes out localca's rotation support by changing the interface so that signing operations should always go through a pool. All existing in-tree callers are converted over to this pattern.  This ensures that signing can properly continue as a CA pool is rotated, without requiring any process restarts.

The pool is periodically reloaded from disk in the course of signing operations, with the loaded data being cached for up to a minute.

Some code for storing the CA state using PEM files was removed because it seemed to be dead.

Follow-on changes will give localjwt a similar treatment, and add administrative commands for running rotations on pools.


internal/localca/localca.go
```diff
@@ -22,38 +36,188 @@ import (
...
+
+	return nil
+}
+
+func (p *RefreshingPool) CreateCertificate(template *x509.Certificate, subjectPublicKey crypto.PublicKey) ([][]byte, error) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	if err := p.refreshIfNecessary(); err != nil {
+		return nil, fmt.Errorf("while refreshing pool: %w", err)
+	}
```
thompsonmax, line 132: If we fail to refresh the pool from disk, should we still allow certs to be created with the pool that's currently in memory?

internal/localca/localca.go
```diff
@@ -22,38 +36,188 @@ import (
...
+
+	// Pick which cert should be used for signing the leaf --- if there are
+	// intermediates, pick the intermediate closest to the leaf.  Otherwise, use
+	// the root certificate.
+	signingCert := selectedCA.RootCertificate
+	if len(selectedCA.IntermediateCertificates) != 0 {
+		signingCert = selectedCA.IntermediateCertificates[0]
+	}
+
+	subjectCertDER, err := x509.CreateCertificate(rand.Reader, template, signingCert, subjectPublicKey, selectedCA.SigningKey)
```
thompsonmax: Do we need to use the private key of the intermediate here, if the signing cert is an intermediate? Currently we're always passing in the private key of the CA IIUC (`selectedCA.SigningKey`).

internal/localca/localca.go
```diff
@@ -22,38 +36,188 @@ import (
...
+	// TrustAnchors returns the root certificates for all of the pool's CAs, including
+	TrustAnchors() ([]*x509.Certificate, error)
+}
+
+// RefreshingPool is a wrapper around Pool that periodically reloads the CA
+// state from disk.  This allows our various pieces that sign certificates
+// (Actor identity broker, egress gateway) to properly continue signing even as
+// an administrator rotates one of the CA pools, without requiring any
+// components to restart.
+type RefreshingPool struct {
```
thompsonmax, line 82: Should we add tests for `RefreshingPool`?

internal/localca/localca.go
```diff
@@ -14,6 +14,20 @@
 
 // Package localca implements a CA whose state can be stored in a local file or
 // Kubernetes secret.
+//
+// In substrate's default setup, the CA pool state is kept in a Kubernetes
+// secret, and administered with admin CLI commands.
+//
+// If you are writing an online signing component, us a projected volume to put
```
thompsonmax: [nit] "us a projected ..." -> "use a projected ..."

cmd/kubectl-ate/internal/cmd/admin_make_ca_pool.go
```diff
@@ -50,7 +50,7 @@ var makeCaPoolCmd = &cobra.Command{
 			return fmt.Errorf("while generating CA: %w", err)
 		}
 
-		pool := &localca.Pool{
+		pool := &localca.ConcretePool{
 			CAs: []*localca.CA{ca},
```
thompsonmax: Should we set this ca as `ActiveForSigning` to avoid hard reliance on the `selectedCA := p.CAs[0]` fallback in `RefreshingPool.CreateCertificate`?

internal/localca/localca.go
```diff
@@ -22,38 +36,188 @@ import (
...
+
+func (p *ConcretePool) CreateCertificate(template *x509.Certificate, subjectPublicKey crypto.PublicKey) ([][]byte, error) {
+	if len(p.CAs) == 0 {
+		return nil, fmt.Errorf("pool has no CAs")
+	}
+
+	// For backwards compatibility, pick the first CA if none is designated.
+	selectedCA := p.CAs[0]
+	for _, ca := range p.CAs {
+		if ca.ID == p.ActiveForSigning {
```
thompsonmax: Falling back to `p.CAs[0]` if `ActiveForSigning` is actually set but no CAs match seems dangerous. Should we return an error in that case?

internal/localca/localca.go
ahmedtd: The private key of the CA *is* the private key of the intermediate, if intermediates are in use.  I should make this clearer.

internal/localca/localca.go
```diff
@@ -22,38 +36,188 @@ import (
...
+//
+// Normally, we let callers define their own compatibility interfaces.  But in
+// most cases you'll want to either use a RefreshingPool (for controllers and
+// servers), or a ConcretePool (for CLIs and tests).
+type Pool interface {
+	// CreateCertificate signs the given template certificate using one of the
+	// Pool's currently-active CAs.
+	CreateCertificate(template *x509.Certificate, subjectPublicKey crypto.PublicKey) ([][]byte, error)
+
+	// TrustAnchors returns the root certificates for all of the pool's CAs, including
```
haiyanmeng: the comment seems unfinished

internal/localca/localca.go
```diff
@@ -22,38 +36,188 @@ import (
...
+	TrustAnchors() ([]*x509.Certificate, error)
+}
+
+// RefreshingPool is a wrapper around Pool that periodically reloads the CA
+// state from disk.  This allows our various pieces that sign certificates
```
haiyanmeng, line 78: Remove the extra space between `disk.` and `This`

internal/localca/localca.go
```diff
@@ -22,38 +36,188 @@ import (
...
+	return nil
+}
+
+func (p *RefreshingPool) CreateCertificate(template *x509.Certificate, subjectPublicKey crypto.PublicKey) ([][]byte, error) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	if err := p.refreshIfNecessary(); err != nil {
+		return nil, fmt.Errorf("while refreshing pool: %w", err)
+	}
+	return p.pool.CreateCertificate(template, subjectPublicKey)
```
haiyanmeng, line 133: Does `p.pool.CreateCertificate` need to be protected by `p.lock`?

internal/localca/localca.go
```diff
@@ -22,38 +36,188 @@ import (
...
+}
+
+func (p *RefreshingPool) TrustAnchors() ([]*x509.Certificate, error) {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	if err := p.refreshIfNecessary(); err != nil {
+		return nil, fmt.Errorf("while refreshing pool: %w", err)
+	}
+	return p.pool.TrustAnchors()
```
haiyanmeng, line 142: Does `p.pool.TrustAnchors` need to be protected by `p.lock`?

internal/localca/localca_test.go
haiyanmeng, line 1: Let us add some test coverage for the new logic added in localca.go

cmd/kubectl-ate/internal/cmd/admin_make_ca_pool.go
ahmedtd: Oh, good point, we need to do this.

However, I was actually intending the fallback behavior to handle all the existing generated CAs that people have created for their dev installs of substrate, so it needs to remain.

ahmedtd: Done.

internal/localca/localca.go
ahmedtd: Done.

internal/localca/localca.go
ahmedtd: Ooops, fixed.

internal/localca/localca.go
ahmedtd: This isn't a mistake --- I was taught to always put two spaces after a period when typing.  Looking online, this seems to be a highly contentious war.

internal/localca/localca.go
ahmedtd: It is protected currently.

Technically, only the p.pool access needs to be protected, the actual function call could be made unlocked.  But I'll let profiling show that to be a problem before complicating the code for that.

internal/localca/localca.go
ahmedtd: See the reply about CreateCertificate.

internal/localca/localca.go
ahmedtd: Good point.  Done.

cmd/kubectl-ate/internal/cmd/admin_make_ca_pool.go
thompsonmax: Sure, I saw your intent for that fallback was backwards compatibility. But I think we should avoid introducing new places where we rely on it as much as possible.

internal/localca/localca.go
haiyanmeng, line 1: @ahmedtd , https://github.com/agent-substrate/substrate/pull/871/ was merged. So we will need to apply this change to https://github.com/agent-substrate/substrate/tree/main/cmd/atenet/internal/sdsmint.

internal/localca/localca_test.go
ahmedtd: Done

internal/localca/localca.go
ahmedtd: It's covered now.  I reorganized the package a bit since certauth wasn't doing anything except acting as a wrapper over localca.Pool.

internal/localca/localca.go
ahmedtd: Done.

internal/localca/localca.go
ahmedtd: I ended up removing intermediate support since it's a bit confusing and not actually used.

internal/localca/localca.go
ahmedtd: I don't think that can actually happen --- secret projections will leave the latest content if something goes wrong with reading the secret.

cmd/kubectl-ate/internal/cmd/admin_make_ca_pool.go
```diff
@@ -47,17 +47,26 @@ var makeCaPoolCmd = &cobra.Command{
...
-			CommonName: caCommonName,
-			KeyType:    localca.KeyType(caKeyType),
-		})
+		var keyType localca.KeyType
+		switch caKeyType {
+		case "ED25519":
+			keyType = localca.KeyTypeED25519
+		case "ECDSAP256":
+			keyType = localca.KeyTypeECDSAP256
+		}
```
thompsonmax, line 58: Since we have no default case here, I believe this falls back to the zero value of the enum (`KeyTypeED25519`) if nothing matches. Should we instead return an error if no key types match to avoid confusing behavior?

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
```diff
@@ -40,39 +40,14 @@ import (
...
-// them rather than their identity.
-const workerPoolLabel = "ate.dev/worker-pool"
-
-func extKeyUsages(pod *corev1.Pod, namespace, serviceAccount string) []x509.ExtKeyUsage {
-	usages := []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth}
-	_, isWorker := pod.ObjectMeta.Labels[workerPoolLabel]
-	isAtelet := namespace == ateletNamespace && serviceAccount == ateletServiceAccount
-	if isAtelet || isWorker {
-		usages = append(usages, x509.ExtKeyUsageServerAuth)
-	}
```
thompsonmax, line 64: [Question] IIUC, this PR now allows any pod identity cert to act as a TLS server, rather than just those generated for atelet/workers. Is that intentional? Maybe worth calling that out in the PR description as a behavior change?

cmd/kubectl-ate/internal/cmd/admin_make_ca_pool.go
```diff
@@ -90,9 +99,6 @@ func init() {
 	makeCaPoolCmd.Flags().StringVar(&caID, "ca-id", "", "The ID of the initial CA in the Pool")
```
thompsonmax, line 101: [nit] Should we add a TODO here as well to make it clear we need to make this field required once we remove the fallback to the first CA if this isn't set?

cmd/atenet/internal/sdsmint/minter.go
```diff
@@ -15,80 +15,142 @@
...
 		// checkHostSyntax already quotes the host, so this does not repeat it.
 		return nil, fmt.Errorf("%w: %w", errHostNotAllowed, err)
 	}
 
-	cert, err := m.signer.Sign(host, m.ttl)
+	now := time.Now()
+
+	tmpl := &x509.Certificate{
+		NotBefore:             now.Add(-5 * time.Minute),
+		NotAfter:              now.Add(m.ttl),
```
thompsonmax, line 107: Should we clamp this to the NotAfter of the signing root (or track this with a TODO)?

internal/localca/localca_test.go
```diff
@@ -183,303 +97,79 @@ func TestMarshalUnmarshalWithIntermediates(t *testing.T) {
...
-	if err := restoredInterm.CheckSignatureFrom(restored.CAs[0].RootCertificate); err != nil {
-		t.Errorf("intermediate cert signature verification against root failed: %v", err)
-	}
-
-	// Verify the intermediate's public key matches the generated key pair.
-	intermPubFromCert := restoredInterm.PublicKey.(ed25519.PublicKey)
-	if !bytes.Equal(intermPubFromCert, intermPub) {
-		t.Error("intermediate cert public key does not match generated key pair")
+	if diff := cmp.Diff(restored, pool); diff != "" {
+		t.Fatalf("Pool didn't round-trip; diff (-got +want)")
```
thompsonmax, line 101: [nit] missing args for `Fatalf`

internal/localca/localca_test.go
```diff
@@ -183,303 +97,79 @@ func TestMarshalUnmarshalWithIntermediates(t *testing.T) {
...
-	}
-	if len(pool.CAs) != 1 {
-		t.Fatalf("CAs length = %d, want 1", len(pool.CAs))
+		t.Fatalf("Unexpected error generating CA 2: %v", err)
 	}
-	if _, ok := pool.CAs[0].SigningKey.(*rsa.PrivateKey); !ok {
-		t.Fatalf("SigningKey type = %T, want *rsa.PrivateKey", pool.CAs[0].SigningKey)
+	pool2 := &ConcretePool{
+		CAs:              []*CA{ca2},
+		ActiveForSigning: "2",
```
thompsonmax, line 125: [nit] we set ActiveForSigning: '2' here but the pool only has a CA with id='1', is that intentional?

cmd/ateapi/internal/actoridentity/actoridentity_test.go
thompsonmax, line 897: [nit] stale comment? I believe the pool we create now is from a valid source.

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
```diff
@@ -148,20 +128,26 @@ func (h *Impl) MakeCert(ctx context.Context, pcr *certsv1beta1.PodCertificateReq
...
 		KeyUsage:              x509.KeyUsageDigitalSignature,
-		ExtKeyUsage:           extKeyUsages(pod, pcr.ObjectMeta.Namespace, pcr.Spec.ServiceAccountName),
-		// Link the leaf to its issuing CA by key id so verifiers can disambiguate
-		// a multi-CA trust bundle when a service trusts both the servicedns and
-		// podidentity CAs).
-		// https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.1
-		AuthorityKeyId: parent.SubjectKeyId,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		// AuthorityKeyID is automatically set to the SubjectKeyID of the parent
+		// certificate.
```
thompsonmax, line 150: [nit] I believe this is only true outside of the self-signing case (which we're seeding a random subject to avoid). Maybe worth calling this out in this comment?

thompsonmax (dismissed):

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
ahmedtd: That's fine --- SPIFFE certs should generally allow both usages.  X.509 cert usages are the wrong place to try to lock down what actors can and can't do.

cmd/atenet/internal/sdsmint/minter.go
ahmedtd: No, I investigated the behavior here and it is legitimate to issue a leaf that outlives the root (and expected in some scenarios).  The trust is established in the root *key*, not the root *certificate*, so a CA could rotate from root certificate *a* to root certificate *b* while maintaining the same key.  The leaf would then be continuously valid --- in the time period before *a* expires, it chains to *a*, then afterwards it chains to *b*.

cmd/kubectl-ate/internal/cmd/admin_make_ca_pool.go
ahmedtd: Done.

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
ahmedtd: Added the commit description

cmd/kubectl-ate/internal/cmd/admin_make_ca_pool.go
ahmedtd: I've added validation that this is non-empty.

internal/localca/localca_test.go
ahmedtd: Added

internal/localca/localca_test.go
ahmedtd: Whoops, that's a mistake.

cmd/ateapi/internal/actoridentity/actoridentity_test.go
ahmedtd: Updated.

cmd/podcertcontroller/internal/podidentitysigner/podidentitysigner.go
ahmedtd: Updated

PR #974, opened 2026-08-15
Add update/verify for Locust Python proto stubs

### LINKED ISSUE #738: Presubmit doesn't catch drift in the Locust Python proto stubs
Nothing prevents the generated Python code for protos from drifting. For Go generated code we have presubmit checks in hack/verify-all.sh, but we are missing Python.

It looks like this happened a couple of times in the past (e.g. https://github.com/agent-substrate/substrate/commit/7c6d9862).

This task is to add this missing presubmit check.

YQ-Wang:
Fixes #738

Adds update and verify scripts for the Locust Python proto stubs so hack/verify-all.sh catches drift from the .proto sources. Updates the benchmarking documentation with the regeneration command.

- [x] Tests pass

### DIFF
--- a/benchmarking/README.md
+++ b/benchmarking/README.md
@@ -81,9 +81,10 @@ Once installed:
 
 ### Rebuilding gRPC Python clients
 
-Make sure you have a virtual environment created (`python3 -m venv venv`)
-and activated (`source venv/bin/activate`).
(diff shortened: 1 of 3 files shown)


YQ-Wang:
@juli4n PTAL

PR #979, opened 2026-08-15
ateomnet: enable IPv6 forwarding in worker pod netns

### LINKED ISSUE #945: ateomnet: worker pod netns enables IPv4 forwarding only, dropping actor IPv6 traffic

`EnableIPv4Forwarding` writes `/proc/sys/net/ipv4/ip_forward` and has no IPv6 counterpart (`internal/ateomnet/net.go:206`). Actor packets enter the worker pod on a veth and are forwarded to `eth0`, so with `net.ipv6.conf.all.forwarding` left at 0 every IPv6 packet an actor sends is dropped by `ip6_forward()` — including the actor's own DNS queries.

On an IPv6-only cluster this is a hard failure: `TestActorEgress` returns 504, reproduced 4/4. The netns setup also has no IPv6 veth addressing and no v6 nftables rules, per the existing TODO at `net.go:243`.

krsnaSuraj:
## ateomnet: enable IPv6 forwarding in the worker pod netns

Fixes: https://github.com/agent-substrate/substrate/issues/945

### Problem

`EnableForwarding` (formerly `EnableIPv4Forwarding`) only wrote `/proc/sys/net/ipv4/ip_forward`. It had no
IPv6 counterpart, so `net.ipv6.conf.all.forwarding` stayed at 0 in the worker
pod network namespace. On dual-stack / IPv6-only clusters, every IPv6 packet
an actor sends — including its own DNS queries — is dropped by the kernel's
`ip6_forward()`, because the packet enters on the actor veth and must be
forwarded out through `eth0`.

Reproduced as `TestActorEgress` returning 504 on IPv6-only clusters (4/4 runs).

### Fix

The helper is renamed to `EnableForwarding` since it now covers both address
families, and it writes `/proc/sys/net/ipv6/conf/all/forwarding`. `conf.all.forwarding=1` sets the
forwarding state for all existing interfaces (and is the default for new
ones), so a single write covers both the veth and `eth0`.

The write is factored into a `writeSysctlIfUnset` helper that preserves the
original behavior:
- if the sysctl already reads `1`, return early (no remount, no write);
- otherwise clear the read-only bind-mount on `/proc/sys` (the worker holds
  CAP_SYS_ADMIN and uses no user namespace), write `1\n`, and restore `ro`.

### Test

`internal/ateomnet/write_sysctl_test.go` covers the helper's fast paths
(already-set is untouched, unset/zero is written) with a temp file, no root
required. The privileged remount path is exercised by the existing netns
integration tests (`withTestNetNS`).

### Compatibility

- IPv4-only clusters are unaffected (IPv4 path unchanged; writing the IPv6
  sysctl is a no-op where IPv6 is disabled or the file is absent — the helper
  returns nil on missing path).
- No change to the nftables rules; the existing
  `InstallActorNftablesRules` IPv4-only TODO remains accurate — this PR only
  restores the kernel forwarding path so IPv6 packets reach the pod's eth0.

### Verification

- `gofmt -l` clean.
- `go build ./internal/ateomnet/` passes.
- `go test ./internal/ateomnet/ -run TestWriteSysctlIfUnset` passes.
- Full `go test ./internal/ateomnet/` passes.

### Merge order

#1057 gives the actor veth an IPv6 address and a default route; without this
PR's forwarding write that route is a black hole. **This PR must merge before
#1057.** No current CI lane catches the ordering (default e2e clusters are
IPv4-only); tracked in #1094 (opt-in dual-stack kind e2e).

### DIFF
--- a/internal/ateomnet/net.go
+++ b/internal/ateomnet/net.go
@@ -192,31 +192,44 @@ func PodIPv4() (net.IP, error) {
 	return nil, fmt.Errorf("pod eth0 has no IPv4 address")
 }
 
-// EnableIPv4Forwarding enables IPv4 forwarding in the current network namespace.
-func EnableIPv4Forwarding() error {
+// EnableForwarding enables IPv4 forwarding in the current network namespace.
+func EnableForwarding() error {
 	// Forwarding is required because actor packets now enter the worker pod via
 	// the host-side veth and then leave through the pod's eth0. Without this, the
 	// kernel would not route traffic between those interfaces even though both
 	// live in the worker pod network namespace.
-	//
-	// Without privileged, the container runtime bind-mounts /proc/sys read-only.
-	// The worker holds CAP_SYS_ADMIN and uses no user namespace, so the ro flag
-	// is not locked: clear it, write the sysctl, restore ro.
 	const path = "/proc/sys/net/ipv4/ip_forward"
+	if err := writeSysctlIfUnset(path); err != nil {
+		return fmt.Errorf("while enabling IPv4 forwarding in worker pod netns: %w", err)
+	}
+	return nil
+}
+
+// writeSysctlIfUnset writes "1\n" to a sysctl path unless it already reads "1".
+// If the path does not exist, it returns nil — the sysctl is simply unavailable,
+// not an error.
+func writeSysctlIfUnset(path string) error {
 	if b, err := os.ReadFile(path); err == nil && len(b) > 0 && b[0] == '1' {
 		return nil
 	}
 	if err := os.WriteFile(path, []byte("1\n"), 0o644); err == nil {
 		return nil
 	}
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		// Path absent (e.g. IPv6 disabled in kernel): nothing to enable.
+		return nil
+	}
+	// Without privileged, the container runtime bind-mounts /proc/sys read-only.
+	// The worker holds CAP_SYS_ADMIN and uses no user namespace, so the ro flag
+	// is not locked: clear it, write the sysctl, restore ro.
 	if err := unix.Mount("none", "/proc/sys", "", unix.MS_BIND|unix.MS_REMOUNT, ""); err != nil {
-		return fmt.Errorf("while remounting /proc/sys read-write to enable IPv4 forwarding: %w", err)
+		return fmt.Errorf("while remounting /proc/sys read-write to enable forwarding: %w", err)
 	}
 	defer func() {
 		_ = unix.Mount("none", "/proc/sys", "", unix.MS_BIND|unix.MS_REMOUNT|unix.MS_RDONLY, "")
 	}()
 	if err := os.WriteFile(path, []byte("1\n"), 0o644); err != nil {
-		return fmt.Errorf("while enabling IPv4 forwarding in worker pod netns: %w", err)
+		return fmt.Errorf("while writing %s: %w", path, err)
 	}
 	return nil
 }
@@ -565,7 +578,7 @@ func SetupActorNetwork(ctx context.Context, cfg NetworkConfig) (retErr error) {
 		return fmt.Errorf("while configuring actor veth in interior netns: %w", err)
 	}
 
-	if err := EnableIPv4Forwarding(); err != nil {
+	if err := EnableForwarding(); err != nil {
 		return err
 	}
 	if err := InstallActorNftablesRules(cfg.EgressRedirectPort); err != nil {
--- a/internal/ateomnet/net.go
+++ b/internal/ateomnet/net.go
@@ -192,7 +192,10 @@ func PodIPv4() (net.IP, error) {
 	return nil, fmt.Errorf("pod eth0 has no IPv4 address")
 }
 
-// EnableForwarding enables IPv4 forwarding in the current network namespace.
+// EnableForwarding enables IPv4 and IPv6 forwarding in the current network
+// namespace, so actor traffic (including DNS queries on IPv6-capable clusters)
+// is routed between the veth and eth0 instead of being dropped by ip_forward()
+// or ip6_forward().
 func EnableForwarding() error {
 	// Forwarding is required because actor packets now enter the worker pod via
 	// the host-side veth and then leave through the pod's eth0. Without this, the
@@ -202,12 +205,22 @@ func EnableForwarding() error {
 	if err := writeSysctlIfUnset(path); err != nil {
 		return fmt.Errorf("while enabling IPv4 forwarding in worker pod netns: %w", err)
 	}
+	// IPv6 forwarding: actor packets that arrive on the veth and leave via eth0
+	// are IPv6 on dual-stack / IPv6-only clusters. Without
+	// net.ipv6.conf.all.forwarding the kernel drops every IPv6 packet in
+	// ip6_forward(), including the actor's DNS queries. conf.all.forwarding=1
+	// also implies the per-interface default, so a single write covers the veth
+	// and eth0.
+	const v6path = "/proc/sys/net/ipv6/conf/all/forwarding"
+	if err := writeSysctlIfUnset(v6path); err != nil {
+		return fmt.Errorf("while enabling IPv6 forwarding in worker pod netns: %w", err)
+	}
 	return nil
 }
 
 // writeSysctlIfUnset writes "1\n" to a sysctl path unless it already reads "1".
-// If the path does not exist, it returns nil — the sysctl is simply unavailable,
-// not an error.
+// If the path does not exist (e.g. IPv6 sysctls on a kernel with IPv6 disabled),
+// it returns nil — IPv6 forwarding is simply unavailable, not an error.
 func writeSysctlIfUnset(path string) error {
 	if b, err := os.ReadFile(path); err == nil && len(b) > 0 && b[0] == '1' {
 		return nil
--- a/internal/ateomnet/net.go
+++ b/internal/ateomnet/net.go
@@ -24,7 +24,10 @@ import (
 	"log/slog"
 	"net"
 	"os"
+	"path/filepath"
 	"runtime"
+	"strings"
+	"syscall"
 
 	"github.com/google/nftables"
 	"github.com/google/nftables/binaryutil"
@@ -192,6 +195,17 @@ func PodIPv4() (net.IP, error) {
 	return nil, fmt.Errorf("pod eth0 has no IPv4 address")
 }
 
+// procSysDir is the mount point of the sysctl tree. ensureProcSysOn refuses any
+// path that does not resolve inside it.
+const procSysDir = "/proc/sys"
+
+// Forwarding sysctls. conf.all.forwarding also supplies the per-interface
+// default, so one IPv6 write covers both the veth and eth0.
+const (
+	procSysIPv4Forwarding = "/proc/sys/net/ipv4/ip_forward"
+	procSysIPv6Forwarding = "/proc/sys/net/ipv6/conf/all/forwarding"
+)
+
 // EnableForwarding enables IPv4 and IPv6 forwarding in the current network
 // namespace, so actor traffic (including DNS queries on IPv6-capable clusters)
 // is routed between the veth and eth0 instead of being dropped by ip_forward()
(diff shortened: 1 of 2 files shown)


krsnaSuraj:
Any chance of a review on this? It's now part of the IPv6 e2e chain — #1059 (the signal-only draft) rides on the worker-pod forwarding + `writeSysctlIfUnset` fix from this branch, and the `hack/create-kind-cluster.sh` head it depends on. Landing #979 lets #1059 drop it from its branch instead of carrying it.

CI is green (`check-changes` + CLA pass) and the branch is mergeable. Happy to adjust anything from review.

ygao-g:
Two asks before merge. The premise checks out first: writing `net.ipv6.conf.all.forwarding` does propagate to interfaces that already exist and to `conf.default`, and it does not purge a statically installed default route — so the single write covers the veth and eth0 whatever the ordering.

**1. Commit 2's branch is never executed.** A coverage run puts the `os.Stat`/`IsNotExist` lines at 0. `unset_written` looks like it covers the case, but its temp path *can* be created, so it returns at the `os.WriteFile` fast path. Procfs cannot create a missing node, so production always reaches `os.Stat` — the opposite of what the test asserts. Pointing a subtest at a path under a directory that does not exist closes it; with the one below, both lines go covered.

**2. `EnableIPv4Forwarding` now enables both families.** One call site, so the rename is contained.

<details>
<summary>Illustrative diff — builds, vet clean, passes</summary>

```diff
diff --git a/internal/ateomnet/net.go b/internal/ateomnet/net.go
index a2a95017..1f0fbc92 100644
--- a/internal/ateomnet/net.go
+++ b/internal/ateomnet/net.go
@@ -192,11 +192,11 @@ func PodIPv4() (net.IP, error) {
 	return nil, fmt.Errorf("pod eth0 has no IPv4 address")
 }
 
-// EnableIPv4Forwarding enables IPv4 forwarding in the current network namespace.
-// It also enables IPv6 forwarding so actor IPv6 traffic (including DNS queries
-// on IPv6-capable clusters) is routed between the veth and eth0 instead of
-// being dropped by ip6_forward().
-func EnableIPv4Forwarding() error {
+// EnableForwarding enables IPv4 and IPv6 forwarding in the current network
+// namespace, so actor traffic (including DNS queries on IPv6-capable clusters)
+// is routed between the veth and eth0 instead of being dropped by ip_forward()
+// or ip6_forward().
+func EnableForwarding() error {
 	// Forwarding is required because actor packets now enter the worker pod via
 	// the host-side veth and then leave through the pod's eth0. Without this, the
 	// kernel would not route traffic between those interfaces even though both
@@ -595,7 +595,7 @@ func SetupActorNetwork(ctx context.Context, cfg NetworkConfig) (retErr error) {
 		return fmt.Errorf("while configuring actor veth in interior netns: %w", err)
 	}
 
-	if err := EnableIPv4Forwarding(); err != nil {
+	if err := EnableForwarding(); err != nil {
 		return err
 	}
 	if err := InstallActorNftablesRules(cfg.EgressRedirectPort); err != nil {
diff --git a/internal/ateomnet/write_sysctl_test.go b/internal/ateomnet/write_sysctl_test.go
index b4cb1db1..d687932d 100644
--- a/internal/ateomnet/write_sysctl_test.go
+++ b/internal/ateomnet/write_sysctl_test.go
@@ -65,6 +65,21 @@ func TestWriteSysctlIfUnset(t *testing.T) {
 		}
 	})
 
+	t.Run("missing_path_is_noop", func(t *testing.T) {
+		// A node under a directory that does not exist stands in for
+		// /proc/sys/net/ipv6/... on a kernel with IPv6 disabled. The other
+		// subtests' paths can be created, so they return at the os.WriteFile
+		// fast path; this is the only one that reaches the os.Stat branch,
+		// which is what procfs always does in production.
+		p := filepath.Join(dir, "no-such-dir", "forwarding")
+		if err := writeSysctlIfUnset(p); err != nil {
+			t.Fatalf("writeSysctlIfUnset on a missing path: %v", err)
+		}
+		if _, err := os.Stat(p); !os.IsNotExist(err) {
+			t.Fatalf("expected %s to stay absent, stat err = %v", p, err)
+		}
+	})
+
 	t.Run("zero_is_rewritten", func(t *testing.T) {
 		p := filepath.Join(dir, "zero")
 		if err := os.WriteFile(p, []byte("0\n"), 0o644); err != nil {
```

</details>

Rebased on current main here: `go test ./...` and the root-gated suite both pass.

ygao-g (approved):
LGTM.

Confirmed by measurement rather than by reading: in a netns replica of the worker topology, actor IPv6 egress across the veth is 100% packet loss with `net.ipv6.conf.all.forwarding=0` and 0% with it set. On a running dual-stack kind cluster the worker pod netns reads `ipv6 all.forwarding=0` and `ipv4 ip_forward=1`, so this is the write that actually changes behaviour — the existing IPv4 one is already a no-op in a pod.

One sequencing note for whoever merges: #1057 gives the actor veth an IPv6 address and a default route, and without this PR that route is a black hole. Neither PR declares the dependency, and no current CI lane would catch it — the e2e clusters are IPv4-only by default, so the IPv6 path never runs.

krsnaSuraj:
Both asks are addressed; the delta since your approval is exactly two commits:

- 63af4fe9 — `EnableIPv4Forwarding` renamed to `EnableForwarding`, doc comment updated, single call site in `SetupActorNetwork` adjusted. Matches your illustrative diff.
- c1809e4a — new `missing_path_is_noop` subtest pointing at a node under a nonexistent directory, so the `os.Stat`/`IsNotExist` branch is the one that executes, as procfs always does in production.

Local verification: `go test -count=1 ./internal/ateomnet/` passes, and the coverage profile now shows the `IsNotExist` return (net.go:235-238) executed — previously 0.

On the sequencing note: agreed, the dependency should be declared. GitHub has no native PR-to-PR dependency, so the practical options are a merge-order note in both PR bodies — I'm adding one here stating #979 must merge before #1057; worth mirroring in #1057 — or a `- [ ]` task-list item if you prefer something merge automation can read. Longer term, the CI gap you flagged looks like what #1094 (opt-in dual-stack kind e2e) is aimed at.

CI is green on the updated head (check-changes + CLA). Happy to re-request review if you want a fresh look at the two commits; otherwise I'll take your approval as covering them.

ygao-g:
Suggestion on the history before merge: it's one commit plus three review fixups. Resplit as refactor-then-behaviour, the extraction reads as a no-op and the second commit is only the IPv6 change — https://github.com/ygao-g/substrate/pull/20, same tree as yours on the same base (`a0bb3382`).

Separately, in `net.go`: the "Without privileged…" block is now duplicated verbatim at lines 205-207 and 239-241. Drop it from `EnableForwarding` — that function no longer remounts anything, and the copy in `writeSysctlIfUnset` is where the remount lives.

krsnaSuraj:
Both asks addressed on the updated head (`9925d0fc`).

**1. History resplit as refactor-then-behaviour** (2 commits on current main):
- `a203ccb6` — `ateomnet: extract the forwarding sysctl write into a helper` — pure refactor: the inline sysctl read/write/remount sequence moves into `writeSysctlIfUnset`, `EnableIPv4Forwarding` → `EnableForwarding` (single call site updated), plus unit tests for the helper's fast paths.
- `9925d0fc` — `ateomnet: enable IPv6 forwarding in worker pod netns` — the actual behaviour change: the IPv6 write via the helper, plus the `missing_path_is_noop` coverage test.

The first commit reads as a no-op (same sysctl, same write, same remount fallback — just moved), and the second is only the IPv6 change, matching the suggested split.

**2. Duplicate "Without privileged…" block dropped** from `EnableForwarding` — that function no longer remounts anything; the only remaining copy is in `writeSysctlIfUnset` (lines 239-241) where the remount actually lives.

**3. Coverage gap closed** — `missing_path_is_noop` points at a node under a nonexistent directory (`<tmpdir>/no-such-dir/forwarding`), so the `os.Stat`/`IsNotExist` return executes instead of returning at the `os.WriteFile` fast path. Verified with a coverage run: `writeSysctlIfUnset` now reports the `IsNotExist` branch executed (previously 0).

Verification: `gofmt -l` clean, `go vet ./internal/ateomnet/` exit 0, `go test -count=1 ./internal/ateomnet/` PASS (all 4 subtests including `missing_path_is_noop`).

internal/ateomnet/net.go
```diff
@@ -192,31 +192,57 @@ func PodIPv4() (net.IP, error) {
...
+// If the path does not exist (e.g. IPv6 sysctls on a kernel with IPv6 disabled),
+// it returns nil — IPv6 forwarding is simply unavailable, not an error.
+func writeSysctlIfUnset(path string) error {
 	if b, err := os.ReadFile(path); err == nil && len(b) > 0 && b[0] == '1' {
 		return nil
 	}
 	if err := os.WriteFile(path, []byte("1\n"), 0o644); err == nil {
 		return nil
 	}
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
bowei: I'm confused about the ordering. Don't we need to do stat(2) first before trying to read and write the file?

internal/ateomnet/net.go
```diff
@@ -192,31 +192,57 @@ func PodIPv4() (net.IP, error) {
...
+		return fmt.Errorf("while enabling IPv6 forwarding in worker pod netns: %w", err)
+	}
+	return nil
+}
+
+// writeSysctlIfUnset writes "1\n" to a sysctl path unless it already reads "1".
+// If the path does not exist (e.g. IPv6 sysctls on a kernel with IPv6 disabled),
+// it returns nil — IPv6 forwarding is simply unavailable, not an error.
+func writeSysctlIfUnset(path string) error {
 	if b, err := os.ReadFile(path); err == nil && len(b) > 0 && b[0] == '1' {
```
bowei: Suggest a different split:

```golang
func EnableForwarding() {
  const (
    v4path = "/proc/sys/net/ipv4/ip_forward"
    v6path = "/proc/sys/net/ipv6/conf/all/forwarding"
  )

  ensure := func() error {
    err := ensureProcSysOn(v4path)
    if err != nil {
      return err
    }
    // When IPv6 is disabled, this path will not be found.
    // Treat this as benign, but log.
    if _, err := os.Stat(v6path); errors.Is(err, os.ErrNotExist) {
      slog.Info("IPv6 is disabled", "path", v6path)
    } else if err != nil {
      return err
    } else {
      err := ensureProcSysOn(v6path)
      if err != nil {
        return err
      }
    }
  }
  
  err := ensure()
  if errors.Is(err, syscall.EROFS) {
    // Read-only proc/sys, temporarily mount as writable.
    return withWritableProcSys(ensure)
  }

  return err
}

func withWritableProcSys(do func() error) error {
    if err := unix.Mount("none", "/proc/sys", "", unix.MS_BIND|unix.MS_REMOUNT, ""); err != nil {
		  return fmt.Errorf("EnableForwarding: remounting /proc/sys read-write: %w", err)
	  }
	  defer func() {
      // Restore read-only mount.
		  err := unix.Mount("none", "/proc/sys", "", unix.MS_BIND|unix.MS_REMOUNT|unix.MS_RDONLY, "")
      if err != nil { /* log error */ }
	 }()

   return do()
}

func ensureProcSysOn(path string) error {
  // Sanity check the path first to make sure it is actually in /proc/sys.
  if ! filepath.IsAbs(path) || !strings.HasPrefix(path, "/proc/sys") {
    return fmt.Errorf("ensureProcSysOn: invalid path %q", path)
  }
  if b, err := os.ReadFile(path); err == nil && len(b) > 0 && b[0] == '1' {
    return nil
  }
	err = os.WriteFile(path, []byte("1\n"), 0o644)
  if err != nil {
    return fmt.Errorf("ensureProcSysOn: %w", err)
  }
  return nil
}
```

internal/ateomnet/net.go
bowei: See my suggestion above.

krsnaSuraj: Fixed. The missing-node check now runs *before* the read/write, and it moved out of the helper into `EnableForwarding` — the IPv6 node is the only one that can legitimately be absent, so the family-specific knowledge belongs there rather than in a family-agnostic helper:

```go
switch _, err := os.Stat(procSysIPv6Forwarding); {
case errors.Is(err, os.ErrNotExist):
	slog.Info("IPv6 forwarding sysctl is absent, skipping", "path", procSysIPv6Forwarding)
	return nil
case err != nil:
	return fmt.Errorf("while checking %s: %w", procSysIPv6Forwarding, err)
}
```

The helper is now a straight read → write with no ordering question left in it:

```go
func ensureProcSysOn(path string) error {
	if err := validateProcSysPath(path); err != nil {
		return err
	}
	if procSysIsSet(path) {
		return nil
	}
	if err := os.WriteFile(path, []byte("1\n"), 0o644); err != nil {
		return fmt.Errorf("while writing %s: %w", path, err)
	}
	return nil
}
```

The returned error still wraps the errno, which is what lets the caller react to EROFS.

internal/ateomnet/net.go
krsnaSuraj: Done, close to your sketch. Three deliberate differences and one thing your sketch surfaced:

1. `stat` stays in `EnableForwarding` instead of the helper. Only the IPv6 node can be absent (IPv6 compiled out / disabled), so keeping the check at the call site leaves `ensureProcSysOn` family-agnostic and keeps the function at one level of wrapping. The helper checks `validateProcSysPath(path)` instead — same intent as your sanity check, but with `filepath.Clean` and a `procSysDir + "/"` prefix so `/proc/sysfoo/...` and `/proc/sys/../etc/passwd` don't slip through.

2. The remount is driven by the error, not attempted after any failure: `ensure()` runs with the mount as found, and only an `errors.Is(err, syscall.EROFS)` sends it through `withWritableProcSys`. A writable `/proc/sys` (the norm outside the container case) no longer pays for two extra mounts, which also means the EROFS path only runs when it is actually the problem.

3. The restore in `withWritableProcSys` logs on failure rather than being dropped — if it can't put the read-only mount back, that should be visible.

Your sketch also caught a real bug in the old tests: they claimed "the privileged bind-remount path is covered by the netns integration tests", but no test called `EnableForwarding`. There is now one (`TestEnableForwardingInNetNS`, root-gated via `roottest.Require`) that pins both nodes to `0` and asserts the call turns them on — first end-to-end coverage of the write path. Worth noting for the review: a fresh netns *inherits* the parent's IPv4 forwarding value (host with forwarding on hands its namespaces `1`), so the test pins `0` first and asserts on the transition instead of on the inherited value.

The temp-file tests are replaced by tests for the two pieces that carry the logic (`validateProcSysPath`, `procSysIsSet`), plus a check that a refused path is not created.

Verified locally: `go build ./...`, `go vet`, `gofmt`, `hack/verify/boilerplate.sh`, and `go test -race ./internal/ateomnet/ ./internal/atunnel/` (with the netns test running as root on Linux).

internal/ateomnet/net.go
krsnaSuraj: Same change as the reply on the suggestion above — the helper no longer stats, and the missing-node check moved into the IPv6 branch of EnableForwarding.

krsnaSuraj:
Implemented the review feedback in fdf79359:

- `ensureProcSysOn` (renamed from `writeSysctlIfUnset`) now guards its path, and the missing-node check moved into the IPv6 branch of `EnableForwarding`, ahead of the read/write instead of after a failed one.
- The read-only `/proc/sys` remount is driven by `EROFS` from the write rather than attempted after any failure, split out into `withWritableProcSys`.
- The temp-file tests are replaced by tests for the logic itself (`validateProcSysPath`, `procSysIsSet`) plus a root-gated netns test that pins both nodes to `0` and asserts `EnableForwarding` turns them on — the previous test file claimed the netns tests covered the write path, and no test actually called `EnableForwarding`.

Individual replies are in the comment threads. The `pr-workflow` run on the new SHA is sitting in `action_required` (outside-collaborator gate) with no jobs started, so it needs an approval from someone with write access before CI can run — the previous run was started by @bowei.

Also worth flagging: the commit before this one is a rebase onto current `main`. The old base had `cmd/ate-setup/internal/steps/publish.go` calling `e.Ko()` while no `Ko` method existed anywhere in the package, so `run-tests` failed in `cmd/ate-setup` for reasons unrelated to this change; current `main` has `koRunner` in its place and the rebase clears it.

PR #981, opened 2026-08-16
DNM: WIP: minimal example of DV for create

thockin:
This does not handle all fields - just enough to make an example.
This does not add the "middle" layer to make workflows safe.
This does not handle updates.

What it does do it show how we might wield DV to effect solid tests and invariants.

It requires patches on top of the 1.37 k8s code, which is ugly but (IMO) manageable.

Most of the first commits are just setting up the code.  Everything from "Main commits" on should be reviewable

@juli4n if this aligns with how you were thinking, I can proceed a bit more.
@laoj2 for your input

@jpbetz @lalitc375 @yongruilin to see how badly I abused DV


thockin:
superceded

Issue #968, opened 2026-08-15
Start sandbox creation before the image pull / checkpoint download completes


milantracy:
ateom doesn't hear anything until all of atelet's prep is done, so sentry startup is serialized after the image pull (cold start) or the checkpoint download (restore).
  
- Restore: download and bundle prep run concurrently, but [g.Wait()](https://github.com/agent-substrate/substrate/blob/89b83a5c89db8b189401645a645b01c144b7d728/cmd/atelet/main.go#L1090) gates the `RestoreWorkload` RPC on both.
- Cold start: `prepareOCIBundles` completes before `RunWorkload` is sent

Nothing about the sandbox itself depends on the actor image or the checkpoint. The sandbox could be booting while those are still downloading.
  
## Proposal
  
Split sandbox creation out of `RunWorkload` / `RestoreWorkload` into its own RPC that atelet can send as soon as the sandbox-level config is ready, then have the existing RPC start the containers.

It helps cold start and DATA-scope restore. Does not help FULL / DATA_ON_GOLDEN restore, where the sentry comes from the memory image.

BenTheElder:
Keep in mind that we do not only support gVisor.

For uVM, I'm not sure this is safe in FULL mode either, since we need to find-paths for the rootfs during booting.

EtiennePerot:
@BenTheElder The idea here is an informational-only RPC that allows parallelizing non-workload-specific runtime initialization to run simultaneously with non-runtime initialization (image pull, checkpoint download).

For runtimes that do not have any non-workload-specific initialization tasks to do, this RPC can be implemented trivially as a no-op.

PR #984, opened 2026-08-16
ateom: prepare sandboxes during image pull and DATA restore

dubin555:
Related to #968.

This implements early sandbox preparation for both cold starts and DATA-scope restores on gVisor and microVM. FULL and DATA_ON_GOLDEN restore paths remain unchanged because they resume the captured sandbox state.

## Summary

- split atelet OCI preparation into sandbox prerequisites and application-bundle preparation
- add internal PrepareSandbox and DiscardPreparedSandbox RPCs with request matching, idempotency, and retryable cleanup
- carry from_checkpoint on PrepareSandbox so runtimes can distinguish RunWorkload from RestoreWorkload consumers
- overlap DATA checkpoint download, application OCI preparation, and runtime initialization
- prepare gVisor networking + pause rootfs early; cold Run starts the root container immediately, while DATA restore defers runsc create until the filesystem checkpoint is available
- prepare microVM host networking + virtiofsd + Cloud Hypervisor + kata-agent early; RunWorkload or DATA RestoreWorkload late-mounts the completed rootfs/volume tree before guest CreateSandbox
- preserve CPU/memory limits, egress shape, runtime assets, and request origin across Prepare and its consumer
- retain the original inline RunWorkload / RestoreWorkload behavior when a runtime returns Unimplemented

## Synchronization

RunWorkload and RestoreWorkload are the barriers: atelet calls them only after both the checkpoint/image leg and sandbox preparation have completed. If either parallel leg fails, atelet discards the prepared sandbox with a detached, bounded cleanup context.

Only DATA restore uses from_checkpoint=true. FULL and DATA_ON_GOLDEN do not call PrepareSandbox, because their sandbox state comes from the memory image.

## Testing

- [x] make verify, including full-repository race tests, generation checks, lint, licenses, proto formatting, and shellcheck
- [x] real gVisor cold-start lifecycle in throwaway user/mount/network namespaces
- [x] real virtiofsd late-mount propagation test
- [x] real local KVM + Cloud Hypervisor + kata-agent PrepareSandbox -> RunWorkload through application-container startup
- [x] real local KVM + Cloud Hypervisor + kata-agent PrepareSandbox(from_checkpoint=true) -> RestoreWorkload(DATA) through application-container startup
- [x] compatibility test for runtimes returning Unimplemented
- [x] request-origin matching and partial-cleanup state tests

### DIFF
--- a/internal/proto/ateompb/ateom.proto
+++ b/internal/proto/ateompb/ateom.proto
@@ -21,7 +21,8 @@ option go_package = "github.com/agent-substrate/substrate/internal/proto/ateompb
 // Ateom is the interface to control a single gVisor (or, in the future microVM)
 // guest inside a worker pod.
 //
-// Each ateom server has two main states, "available" and "executing".
+// Each ateom server may be available, have a prepared root sandbox, or be
+// executing a workload.
 //
 // When the ateom is "available", the substrate control plane is free to either
 // boot a new workload (using RunWorkload), or restore an existing workload from
@@ -32,6 +33,14 @@ option go_package = "github.com/agent-substrate/substrate/internal/proto/ateompb
 // running workload (with CheckpointWorkload).  This moves the ateom back to
 // "free" state.
 service Ateom {
+  // PrepareSandbox starts the root sandbox before application images finish
+  // downloading. RunWorkload later attaches and starts the applications.
+  rpc PrepareSandbox(PrepareSandboxRequest) returns (PrepareSandboxResponse) {}
+
+  // DiscardPreparedSandbox tears down a sandbox that cannot be used because
+  // another concurrent preparation step failed.
+  rpc DiscardPreparedSandbox(DiscardPreparedSandboxRequest) returns (DiscardPreparedSandboxResponse) {}
+
   // RunWorkload tells ateom to begin running a new workload (one or more
   // containers, potentially with shared filesystems).
   rpc RunWorkload(RunWorkloadRequest) returns (RunWorkloadResponse) {}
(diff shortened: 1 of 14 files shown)


BenTheElder (commented):
https://github.com/agent-substrate/substrate/issues/968#issuecomment-5299741913

dubin555:
Thanks — agreed. The issue's statement that "nothing about the sandbox itself depends on the actor image or checkpoint" is too broad for microVM: its boot/restore requires the application rootfs to be staged for virtiofsd/find-paths.

This PR intentionally implements only the gVisor cold-Run optimization. microVM returns `Unimplemented` for `PrepareSandbox`; atelet treats that as "optional fast path unsupported", waits for the runtime assets and all OCI bundles, and then invokes the unchanged microVM `RunWorkload`, which stages the rootfs and performs the complete VM boot.

All Restore paths, including FULL and DATA_ON_GOLDEN, are unchanged. They still wait for checkpoint download and OCI preparation before calling `RestoreWorkload`.

Therefore this PR is related to #968, but does not implement early microVM boot or close the microVM/Restore portions of the issue.

BenTheElder:
We're doing uVM-first currently, so I'm not super keen to see new complications that are only implemented for gvisor. Earlier in the project we only had gvisor but we've closed the feature gap (not performance yet, in progress).

Can we please look at how this would make sense for uVM *before* changing the APIs? Thank you.

dubin555:
Thanks — I reworked this around the uVM path before asking for another review. The updated API now carries the runtime assets and workload shape that microVM preparation actually needs.

For a cold microVM Run, `PrepareSandbox` now configures host networking, starts virtiofsd on an empty shared tree, launches Cloud Hypervisor, boots the guest, and waits for kata-agent while atelet prepares the application OCI bundles. Once those bundles are ready, `RunWorkload` mounts their rootfs overlays beneath the already-running virtiofsd and only then calls `CreateSandbox` (the guest’s first access to the shared tree), configures guest networking, and starts the application containers. Restore keeps its existing ordering. The `Unimplemented` fallback for older/other runtimes is also unchanged.

I validated the load-bearing late-mount behavior with real virtiofsd, and also ran the complete `PrepareSandbox -> RunWorkload` path locally with KVM + Cloud Hypervisor + kata-agent through application-container startup. Full `make verify` passes. The implementation is in d2e2fcfd.

internal/proto/ateompb/ateom.proto
```diff
@@ -401,3 +410,34 @@ message GetActiveWorkloadStatsResponse {
     NoSampleReason no_sample_reason = 2;
   }
 }
+
+message PrepareSandboxRequest {
```
EtiennePerot, line 468: Can this message carry a `from_checkpoint` bool?

This is useful if the runtime's initialization steps differ from the cold-start case and the from-checkpoint case.

dubin555: Done in 40d6888e.

PrepareSandboxRequest now carries from_checkpoint. Cold Run sends false; only DATA-scope restore sends true. FULL and DATA_ON_GOLDEN do not call PrepareSandbox because their sandbox state comes from the memory image.

For DATA restore, atelet now overlaps checkpoint download, application OCI preparation, and PrepareSandbox. gVisor uses the flag to prepare networking/rootfs but defer runsc create until the filesystem checkpoint is available. microVM boots through a responsive kata-agent and RestoreWorkload consumes that matching prepared VM before staging volumes/rootfs and calling CreateSandbox.

If PrepareSandbox returns Unimplemented, atelet still waits for checkpoint + OCI preparation and calls the unchanged inline RestoreWorkload path, so the compatibility fallback remains serial.

I also added origin/cleanup-state tests, ran the affected packages with the race detector, ran full make verify, and exercised PrepareSandbox(from_checkpoint=true) -> RestoreWorkload(DATA) on local KVM through application-container startup.

milantracy:
@dubin555 is the PR still alive?

dubin555:
Yes, it is still active. I have rebased the branch onto the latest `main` and resolved the conflicts. The PR is mergeable again; the refreshed workflow is waiting for maintainer approval to run.

PR #987, opened 2026-08-16
DNM: Example of "middle layer" for create.

thockin:
This includes #986 and #981 and adds commits to sketch out a middle layer between RPC and storage.

It's just an example, enough to keep pushing with the idea but not mergable.

@juli4n @laoj2


thockin:
Superceded

PR #990, opened 2026-08-17
benchmarking/orchestrator: add --junit-output flag and record test durations

Qqkyu:
This PR adds the `--junit-output` flag to `benchmarking/automation/orchestrator.py` to record test execution durations and export structured JUnit XML reports.


maxsmythe (approved):
LGTM

PR #993, opened 2026-08-17
atunnel: ingress CONNECT support

keithmattix:
Finishes up the vision from #715 to have atunnel serve CONNECT on the ingress path. This will give us the option to hit actors on other ports besides 80. I haven't wired up atenet router yet because it's nontrivial; we should do that in a second step so we can have a baseline for performance

### DIFF
--- a/cmd/ateapi/internal/controlapi/syncer_test.go
+++ b/cmd/ateapi/internal/controlapi/syncer_test.go
@@ -968,11 +968,12 @@ func TestReleaseActorOnDeadWorker_StatusTransitions(t *testing.T) {
 type conflictStore struct {
 	store.Interface
 	conflictTriggered atomic.Bool
+	shouldInject      func(worker *ateapipb.Worker) bool
 	onUpdate          func(ctx context.Context, worker *ateapipb.Worker)
 }
 
 func (c *conflictStore) UpdateWorker(ctx context.Context, worker *ateapipb.Worker, expectedVersion int64) error {
-	if c.onUpdate != nil && c.conflictTriggered.CompareAndSwap(false, true) {
+	if c.shouldInject != nil && c.shouldInject(worker) && c.conflictTriggered.CompareAndSwap(false, true) {
 		c.onUpdate(ctx, worker)
 	}
 	return c.Interface.UpdateWorker(ctx, worker, expectedVersion)
@@ -998,7 +999,20 @@ func TestSyncer_UpdateWorker_RetryOnVersionConflict(t *testing.T) {
 
 	var cs *conflictStore
 	persistence, fakeK8s, fakeAte, syncer, cleanup := setupSyncerTestWithStore(t, ctx, func(s store.Interface) store.Interface {
-		cs = &conflictStore{Interface: s}
+		// Configure the injector before the syncer starts. It only fires for the
+		// update under test, so the initial worker creation cannot consume it.
+		cs = &conflictStore{
+			Interface: s,
+			shouldInject: func(w *ateapipb.Worker) bool {
+				return w.GetSandboxClass() == "microvm"
+			},
+			onUpdate: func(c context.Context, w *ateapipb.Worker) {
+				if cw, err := s.GetWorker(c, ns, poolName, podName); err == nil {
+					cw.NodeName = "node2"
+					_ = s.UpdateWorker(c, cw, cw.Version)
+				}
+			},
+		}
 		return cs
 	}, pool)
 	defer func() {
@@ -1066,14 +1080,6 @@ func TestSyncer_UpdateWorker_RetryOnVersionConflict(t *testing.T) {
 	})
 	if err != nil {
 		t.Fatalf("pool informer cache failed to update: %v", err)
-	}
-
-	// Configure conflictStore to inject a concurrent version bump in Redis when the syncer calls UpdateWorker.
-	cs.onUpdate = func(c context.Context, w *ateapipb.Worker) {
-		if cw, err := cs.Interface.GetWorker(c, ns, poolName, podName); err == nil {
-			cw.NodeName = "node2"
-			_ = cs.Interface.UpdateWorker(c, cw, cw.Version)
-		}
 	}
 
 	// Touch the pod ONCE in K8s so the syncer reconciles it. The first reconcile's
--- a/cmd/atecontroller/internal/controllers/workerpool_apply.go
+++ b/cmd/atecontroller/internal/controllers/workerpool_apply.go
@@ -78,6 +78,7 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool, otel ateomOTelSettin
 		WithArgs(
 			"--pod-uid=$(POD_UID)",
 			"--atunnel-listen-address=0.0.0.0:443",
+			"--atunnel-connect-listen-address=0.0.0.0:444",
 			"--atunnel-credential-bundle="+atunnelIdentityMountPath+"/credential-bundle.pem",
 			"--atunnel-trust-bundle="+atunnelIdentityMountPath+"/trust-bundle.pem",
 			"--atunnel-egress-listen-address=0.0.0.0:15001",
@@ -86,7 +87,11 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool, otel ateomOTelSettin
 		WithPorts(corev1ac.ContainerPort().
 			WithName("https").
 			WithContainerPort(443).
-			WithProtocol(corev1.ProtocolTCP)).
+			WithProtocol(corev1.ProtocolTCP),
+			corev1ac.ContainerPort().
+				WithName("connect").
+				WithContainerPort(444).
+				WithProtocol(corev1.ProtocolTCP)).
 		WithSecurityContext(ateomSecurityContext(wp.Spec.SandboxClass)).
 		WithEnv(ateomContainerEnv(otel)...).
 		WithVolumeMounts(
--- a/demos/counter/counter.go
+++ b/demos/counter/counter.go
@@ -69,6 +69,7 @@ func main() {
 	secondFileCounterDirectory := pflag.String("second-file-counter-directory", "", "Directory for a second file counter; empty disables it. Used to exercise an Actor with more than one durable volume")
 	validateExistingFilePath := pflag.String("validate-existing-file-path", "", "Path to existing file to validate reading")
 	extraPort := pflag.Int("extra-port", 0, "Additional port to listen on, for exercising atenet-router's arbitrary-port ingress support; 0 disables it")
+	tcpPort := pflag.Int("tcp-port", 0, "Plain TCP echo port for exercising atunnel CONNECT ingress; 0 disables it")
 	pflag.Parse()
 	ctx := context.Background()
 
@@ -173,6 +174,28 @@ func main() {
 			if err := http.ListenAndServe(addr, extraMux); err != nil {
 				slog.ErrorContext(ctx, "Error starting extra-port server", slog.Any("err", err))
 				os.Exit(1)
+			}
+		}()
+	}
+
+	if *tcpPort > 0 {
+		go func() {
+			listener, err := net.Listen("tcp", fmt.Sprintf(":%d", *tcpPort))
+			if err != nil {
+				slog.ErrorContext(ctx, "Error starting counter TCP echo server", slog.Any("err", err))
+				os.Exit(1)
+			}
+			slog.InfoContext(ctx, "Starting counter TCP echo server", slog.Int("port", *tcpPort))
+			for {
+				conn, err := listener.Accept()
+				if err != nil {
+					slog.ErrorContext(ctx, "Counter TCP echo accept failed", slog.Any("err", err))
+					return
+				}
+				go func() {
+					defer conn.Close()
+					_, _ = io.Copy(conn, conn)
+				}()
 			}
 		}()
 	}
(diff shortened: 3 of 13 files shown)


cmd/atecontroller/internal/controllers/workerpool_apply.go
```diff
@@ -78,6 +78,7 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool, otel ateomOTelSettin
 		WithArgs(
 			"--pod-uid=$(POD_UID)",
 			"--atunnel-listen-address=0.0.0.0:443",
```
EItanya, line 80: Do we even need another listen address? Can't we serve everything here?

keithmattix: The logic for doing CONNECT from atenet router - > atunnel is non-trivial and it has some connection pooling implications, so I wanted to start with a separate port to keep the scope of this down. Doing it this way also helps us keep a baseline of current state around for a bit so we can make sure we don't regress by accident

internal/atunnel/egress.go
```diff
@@ -280,22 +280,35 @@ func (e *Egress) handle(downstream net.Conn) {
...
 	}()
-	<-done
-	<-done
+	for range 2 {
+		select {
+		case <-ctx.Done():
+			return
+		case <-done:
+		}
+	}
```
EItanya: Not a huge fan of this. Can we use an errgroup.WithContext so we don't need to range like this at the end?

internal/atunnel/egress.go
```diff
@@ -280,22 +280,35 @@ func (e *Egress) handle(downstream net.Conn) {
...
 }
 
-func closeWrite(conn net.Conn) {
+func closeWrite(conn any) {
```
EItanya: Why is this `any` and then it checks the interface below?

internal/atunnel/egress.go
```diff
@@ -280,22 +280,35 @@ func (e *Egress) handle(downstream net.Conn) {
 }
 
 func copyBothWays(a, b net.Conn) {
+	relayWithHalfClose(context.Background(), a, b, b)
```
EItanya: Why is this background?

internal/atunnel/egress.go
keithmattix: Ah this whole file was Codex getting overzealous, Going to revert all of this

internal/atunnel/egress.go
keithmattix: Reverted

internal/atunnel/egress.go
keithmattix: Reverted

cmd/ateapi/internal/controlapi/syncer_test.go
```diff
@@ -968,11 +968,12 @@ func TestReleaseActorOnDeadWorker_StatusTransitions(t *testing.T) {
 type conflictStore struct {
 	store.Interface
 	conflictTriggered atomic.Bool
+	shouldInject      func(worker *ateapipb.Worker) bool
```
keithmattix, line 971: This was added to try to reduce some flakiness with OnUpdate being written to/read across tests

ygao-g:
Heads up: `--atunnel-connect-listen-address` defaults to `0.0.0.0:444`, which adds a third instance of #943. It is a declared containerPort that the router dials for CONNECT, so on an IPv6-only cluster nothing is listening on the worker's v6 PodIP — the same failure as 443. `:444` binds all address families and still accepts IPv4.

It also collides textually with #978 and #1042: all three edit the same flag var blocks in `cmd/ateom-gvisor/main.go` and `cmd/ateom-microvm/main.go`, and the same arg list in `workerpool_apply.go`.

keithmattix:
@ygao-g I'm not sure I understand; if 0.0.0.0:443 binds dualstack by default, then an ipv6 cluster should still work with it right? The main difference is that #943 was on the egress path and required IP6T_SO_ORIGINAL_DST. The status quo should work with ipv6 only clusters

ygao-g:
You're right, and my heads-up was wrong.

`net.Listen("tcp", "0.0.0.0:444")` doesn't bind IPv4-only. Go treats an unspecified address as a wildcard on a listen, so `favoriteAddrFamily` picks `AF_INET6` with `IPV6_V6ONLY=0` and the socket is `[::]:444`, accepting both families — the router reaching the worker's v6 PodIP on 444 works as written. Your read of #943 is the right one: there the dual-stack bind is the bug rather than the fix, because the accepted v6 connection has no IPv4 `SO_ORIGINAL_DST` to recover. CONNECT takes its target from the request authority and never needs that lookup.

One thing still applies: #1042 changes `atunnel-listen-address` to `:443` in the same var block in both `main.go`s, so whichever lands second gets a conflict. I'll take the rebase on my side.

howardjohn (approved):

EItanya (approved):

bowei:
I see this was merged, but looking at some of the commits, they look like intermediary commits. In the future, it would be useful to squash this right before the merge to keep the commits meaningful for the overall history...

BenTheElder:
Alternatively: We have squash merge enabled as the default, if the PR doesn't have clean commit history then you can just squash it on merge.

EItanya:
> I see this was merged, but looking at some of the commits, they look like intermediary commits. In the future, it would be useful to squash this right before the merge to keep the commits meaningful for the overall history...

According to the commit history this was squashed and merged already?

https://github.com/agent-substrate/substrate/commits/main/
https://github.com/agent-substrate/substrate/commit/b9e42db348a4fdc690f89fd9635a18a992fcbd31

bowei:
Ok, sorry, was looking at this PR itself. No worries. We should make the guidance more clear -- I've been looking at some of the other changes that have been going in.

BenTheElder:
> According to the commit history this was squashed and merged already?

Thanks @EItanya, and the recent the log looks clean to me ... even passed tests after merge this time 🤞 

> We should make the guidance more clear -- I've been looking at some of the other changes that have been going in.

Yeah, it should be relatively uncommon that we have a PR large enough to warrant multiple commits at this point, wondering if we even want any other mode ...

demos/counter/counter.go
```diff
@@ -69,6 +69,7 @@ func main() {
 	secondFileCounterDirectory := pflag.String("second-file-counter-directory", "", "Directory for a second file counter; empty disables it. Used to exercise an Actor with more than one durable volume")
 	validateExistingFilePath := pflag.String("validate-existing-file-path", "", "Path to existing file to validate reading")
 	extraPort := pflag.Int("extra-port", 0, "Additional port to listen on, for exercising atenet-router's arbitrary-port ingress support; 0 disables it")
+	tcpPort := pflag.Int("tcp-port", 0, "Plain TCP echo port for exercising atunnel CONNECT ingress; 0 disables it")
```
LiorLieberman, line 72: not sure I follow what is it for?

internal/atunnel/ingress.go
```diff
@@ -36,6 +37,10 @@ import (
 )
 
 const (
+	// DefaultConnectPort is the worker port on which atunnel accepts inbound
+	// mTLS CONNECT tunnels from the ingress router.
+	DefaultConnectPort = 444
```
LiorLieberman, line 42: nit: any reason 444 is used? e.g why not 8443? I understand this is from ingress --> atunnel but IIRC what what we used in #715 was 8443 for client --> ingress for CONNECT?

Is the plan to migrate this to be served on the same 443 long term or leave it as two distinct listeners?

internal/atunnel/ingress.go
```diff
@@ -209,6 +232,123 @@ func (s *Server) Serve(ctx context.Context, lis net.Listener) error {
 	return err
 }
 
+// ServeConnectHTTP accepts a router-authenticated CONNECT request and relays
```
LiorLieberman, line 235: what does router-authenticated mean?

internal/atunnel/ingress.go
```diff
@@ -209,6 +232,123 @@ func (s *Server) Serve(ctx context.Context, lis net.Listener) error {
...
+	ref, ctx, release, ok := s.authorize(r)
+	if !ok {
+		s.reject(w)
+		return
+	}
+	defer release()
+
+	_, port, err := net.SplitHostPort(r.Host)
+	if err != nil {
+		http.Error(w, "CONNECT authority must include a port", http.StatusBadRequest)
```
LiorLieberman, line 251: so we are taking the port from authority, and we are not using any connect metadata?

manifests/ate-install/components/agentgateway/kustomization.yaml
```diff
@@ -55,7 +55,7 @@ patches:
         path: /spec/template/spec/containers/1
         value:
           name: agentgateway
-          image: cr.agentgateway.dev/agentgateway:v1.4.1
+          image: cr.agentgateway.dev/agentgateway:v0.0.0-alpha.a6c0e366
```
LiorLieberman, line 58: Whats is the expectation here wrt to pre-released versions? (given that its in the main project). Ideally we avoid putting non stable versions here. We have been trying to avoid doing this with envoy.

internal/atunnel/ingress.go
EItanya: The plan long term will be to migrate to a single port and only support connect, but AFAIU envoy will need some changes to make that happen.

internal/atunnel/ingress.go
EItanya: The router accepts the request and forwards it securely. We don't have authn/authz in the router today so maybe a tad misleading

manifests/ate-install/components/agentgateway/kustomization.yaml
EItanya: Honestly I don't think that's a realistic expectation right now. This project is so unstable itself, and is moving so quickly that I think it's ok. Especially given that we have promised an immediate turnaround on all agentgateway fixes

internal/atunnel/ingress.go
keithmattix: Not sure what you mean by CONNECT metadata? This is a new TCP connection/HTTP request to atunnel on a worker, so we have no envoy metadata. We could add extra headers, but :authority is built to solve this problem, so I used that

internal/atunnel/ingress.go
keithmattix: Ah sorry, I missed this in a rename. There was a function called authorize() but all it did was reserve/allocate a worker. Must've added this comment before the rename

demos/counter/counter.go
keithmattix: Oh sorry I was testing a branch for adding TCP support and this snuck into the final version. I can remove if we'd like, but it's probably nice to keep around

keithmattix:
@bowei - good call on the squashed commits.  I typically try to be good about doing another push after and approval, but I'll be honest in saying that flaky CI made me hesitant to push again 😅 I'll do better about this

manifests/ate-install/components/agentgateway/kustomization.yaml
LiorLieberman: I am open for thoughts. Also fine saying this is something that will be changed moving forward when substrate is more stable but we need to think about that. 

We avoided doing it with envoy - we probably need to be opinionated whether we should embrace un-released versions of the proxies (and other dependencies) or no

internal/atunnel/ingress.go
LiorLieberman: yeah I meant CONNECT headers. the requests goes to one port on atunnel, (I assumed it is through authority port) but then needs to go to a different port on the actor (which is taken from the originally dialed dns)? how does this actor port persisted?

internal/atunnel/ingress.go
LiorLieberman: its still a nit but why 444 and not same connect protocol as we are doing for client-->atunnel? just simpler to remember

internal/atunnel/ingress.go
keithmattix: Might be helpful to walk through the full flow:
1. Request gets to atenet router
2. atenet router inspects the request to figure out what actor to use (currently via Host header) as well as what port to go to (also via host header)
3. atenet router sends CONNECT to atunnel with the port communicated via :authority
4. atunnel forwards traffic to the actor based on the port in that authority

The key thing about getting that port from the atenet router is that it gives us a future where, after we stop using Host header for actor identification, we can still know exactly what port on the actor we intend to talk to without having to parse the original request (assuming that's possible; it might be TCP). In that scenario (i.e. non-HTTP) CONNECT to atenet router tells us what the port should be and we propagate that to the CONNECT to atunnel.

/cc @EItanya

internal/atunnel/ingress.go
bowei: Do we want to reuse a port from the lower 0-1023 range?

https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports

This is currently occupied by an actual protocol assignment.

keithmattix: Pagers on Kubernetes!

Seriously, good find though; I can change this to the router connect port

manifests/ate-install/components/agentgateway/kustomization.yaml
keithmattix: I think it's fair game for everything; if we're going to iterate quickly before we're stable, prerelease seems expected IMO

BenTheElder:
The flakes are really bad, standing offer to prioritize reviewing flake fixes...

PR #996, opened 2026-08-17
feat: Support Cloud SQL via Auth Proxy for PostgreSQL backend

shrutiyam-glitch:
### Description
This PR introduces native, secure support for using Cloud SQL as the PostgreSQL store backend for `ate-api-server`. 

To ensure the highest level of security and ease of use in GCP environments, this integration leverages the Cloud SQL Auth Proxy sidecar with automatic IAM database authentication. This means transport security (TLS 1.3 tunnel) is handled automatically, and database sessions are authenticated using Workload Identity via short-lived OAuth tokens, completely eliminating the need for database passwords.

### Key Changes
* **Cloud SQL Auth Proxy Sidecar:** Added `manifests/ate-install/cloudsql-proxy-patch.yaml` to patch the sidecar into the `ate-api-server` deployment when a Cloud SQL instance is configured.
* **Automated Provisioning:** Extended `tools/setup-gcp` with a new `cloudsql` command. This handles the idempotent creation of the Cloud SQL instance, Google Service Accounts (GSA), IAM bindings, and Workload Identity bindings.
* **Installation Script Updates:** Updated `hack/install-ate.sh` to parse new environment variables (e.g., `ATE_API_POSTGRES_CLOUDSQL_INSTANCE`, `ATE_API_POSTGRES_CLOUDSQL_GSA`) and correctly synthesize the passwordless DSN and ConfigMaps for the proxy.
* **Security & Documentation:** 
  * Added extensive documentation in `tools/setup-gcp/cloud-sql.md` covering provisioning, schema privileges, deployment, and database scaling.
  * Updated `docs/threat-model.md` to reflect the new Cloud SQL egress flows and Auth Proxy tunnel mechanics.
* **Dependencies:** Vendored required Google API clients (`sqladmin/v1`, `servicenetworking/v1`, `iam/v1`) for the GCP setup tool.

- [X] Tests pass


Stevenjin8:
I'm a bit -1 on this. I think we want substrate to be as cloud neutral where possible (https://docs.google.com/document/d/1rWkmIWTAo_MUGleZ48k3py9jmL7Yfjdf1qffdyiLJCw/edit?resourcekey=0-z9N8Ee108biqI4CsTTjQUg&tab=t.0#heading=h.6vr9xodr524z). For example, folks using Cloud SQL could use a cloud sql proxy sidecar (https://github.com/GoogleCloudPlatform/cloud-sql-proxy) either as a sidecar and substrate workloads can just connect to localhost.

You get all the benefits of mtls/gke workload identity without any Cloud SQL specifics in the codebase.

shrutiyam-glitch:
@Stevenjin8 Agreed ! 
The installation script just supports the use of CloudSQL using Cloud SQL Auth Proxy as a side car or using a direct connection under a flag `ATE_API_POSTGRES_CLOUDSQL_INSTANCE`. The `atepg` module is still cloud neutral and supports any Postgresql connection.

Stevenjin8:
@shrutiyam-glitch I see. I misread the PR

tools/setup-gcp/cmd/cloudsql.go
```diff
@@ -0,0 +1,448 @@
...
+		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(5 * time.Second):
+		}
+		var err error
+		op, err = svc.Operations.Get(cfg.ProjectID, op.Name).Context(ctx).Do()
+		if err != nil {
+			return fmt.Errorf("poll operation %s: %w", op.Name, err)
```
jpbetz:
```suggestion
		name := op.Name
		op, err = svc.Operations.Get(cfg.ProjectID, name).Context(ctx).Do()
		if err != nil {
			return fmt.Errorf("poll operation %s: %w", name, err)
```

To avoid the panic

hack/install-ate.sh
```diff
@@ -684,9 +769,39 @@ deploy_ate_apiserver() {
...
+    # SQL IAM database user.
+    if [[ -n "${ATE_API_POSTGRES_CLOUDSQL_GSA:-}" ]]; then
+      run_kubectl annotate serviceaccount ate-api-server -n ate-system \
+        "iam.gke.io/gcp-service-account=${ATE_API_POSTGRES_CLOUDSQL_GSA}" --overwrite
+    fi
+    run_kubectl patch deployment ate-api-server -n ate-system \
+      --patch-file manifests/ate-install/cloudsql-proxy-patch.yaml
+  elif run_kubectl get deployment ate-api-server -n ate-system \
+      -o jsonpath='{.spec.template.spec.initContainers[*].name}' 2>/dev/null \
+      | grep -qw cloud-sql-proxy; then
```
jpbetz, line 1023: I agree with this catch too:

"I think this removal trigger is too easy to hit by accident. It fires when the variable is merely absent from the caller's shell. A routine `--deploy-ate-apiserver` from a fresh terminal, or a CI job that never exported `ATE_API_POSTGRES_CLOUDSQL_INSTANCE`, deletes the sidecar, strips the Workload Identity annotation, and, via `create_api_server_env_vars`, overwrites the Secret DSN with the default in-cluster one. That DSN can point at a StatefulSet that was never deployed. The deployment itself already records whether Cloud SQL is in use, so intent doesn't need to be inferred from an unset variable."

hack/install-ate.sh
```diff
@@ -507,15 +539,62 @@ create_api_server_env_vars() {
...
-    --from-literal=ATE_API_STORE_BACKEND="${backend}" \
+    "${cm_args[@]}" \
+    --dry-run=client -o yaml \
+    | run_kubectl apply -f -
+
+  # The Postgres DSN may carry a password (external databases without IAM
+  # auth), so it lives in a Secret, not the ConfigMap above. The deployment
+  # lists the secretRef after the configMapRef, so this value wins if both
+  # define the key.
+  run_kubectl create secret generic -n ate-system ate-api-server-secret-envvars \
```
jpbetz, line 754: Found by AI:


"Unless I'm misreading the upgrade path, this migration has a bad window for existing installs. A pre-PR install has the DSN in the ConfigMap and a running Deployment with no secretRef. Running `--create-api-server-env-vars` on its own prunes the DSN key from the ConfigMap (client-side apply, the key is in last-applied) and writes it only to this Secret, which the live Deployment does not reference yet. The next pod restart comes up with no DSN. Probably worth making the standalone flag apply the manifest first, or documenting that it is unsafe on its own across this upgrade."

hack/install-ate.sh
```diff
@@ -616,14 +695,18 @@ deploy_ate_system() {
   # selected so backend configuration and deployed resources cannot diverge.
   # Store-specific overlay composition can remove the unused Valkey resources
   # in a separate change.
-  if [[ "$(store_backend)" == "postgres" ]]; then
+  # An externally provided database — a DSN or a Cloud SQL instance —
+  # replaces the in-cluster PostgreSQL, so skip deploying it in that case.
+  if [[ "$(store_backend)" == "postgres" && -z "${ATE_API_POSTGRES_CONNECTION_STRING:-}" && -z "${ATE_API_POSTGRES_CLOUDSQL_INSTANCE:-}" ]]; then
     apply_postgres
```
jpbetz: Also caught by AI. Looks like a blocker:

"This changes behavior for existing setups: pre-PR, `--store-backend=postgres` applied the StatefulSet unconditionally. Anyone whose `ATE_API_POSTGRES_CONNECTION_STRING` intentionally targets the in-cluster `postgres.ate-system.svc` (custom pool params, say) now gets an install with no database, and the only symptom is the ateapi rollout timeout 120 seconds later. The new semantics seem right to me, but a log line saying the StatefulSet was skipped and why would make that failure diagnosable."

hack/install-ate.sh
```diff
@@ -491,9 +505,27 @@ create_api_server_env_vars() {
...
+      # Cloud SQL via the Auth Proxy sidecar: ateapi talks plaintext to the
+      # proxy on pod-local loopback; the proxy owns TLS and IAM database
+      # authentication. The database username is the GSA email with the
+      # .gserviceaccount.com suffix trimmed.
+      if [[ -z "${ATE_API_POSTGRES_CLOUDSQL_GSA:-}" ]]; then
+        echo "Error: ATE_API_POSTGRES_CLOUDSQL_INSTANCE requires ATE_API_POSTGRES_CLOUDSQL_GSA" \
+          "(or an explicit ATE_API_POSTGRES_CONNECTION_STRING)" >&2
+        exit 1
+      fi
+      postgres_connection_string="user=${ATE_API_POSTGRES_CLOUDSQL_GSA%.gserviceaccount.com} host=127.0.0.1 port=5432 dbname=atepg sslmode=disable"
```
jpbetz: The docs say `ATE_API_POSTGRES_CLOUDSQL_IAM_AUTH=false` requires an explicit `ATE_API_POSTGRES_CONNECTION_STRING`, but nothing enforces that.  Let's add validation.

hack/install-ate.sh
```diff
@@ -684,9 +769,39 @@ deploy_ate_apiserver() {
...
   run_kubectl rollout status deployment/ate-api-server -n ate-system --timeout=120s
 }
 
+# Reconciles the Cloud SQL Auth Proxy sidecar and Workload Identity
+# annotation on ate-api-server. Runs after the deployment manifest is
+# applied: patches the sidecar in when ATE_API_POSTGRES_CLOUDSQL_INSTANCE is
+# set, and removes it (and the annotation) when it is not. Kept out of the
+# base manifest so non-GCP installations carry no Cloud SQL configuration.
+reconcile_cloudsql_proxy_sidecar() {
+  if [[ -n "${ATE_API_POSTGRES_CLOUDSQL_INSTANCE:-}" ]]; then
```
jpbetz: Maybe less critical but suprising:

"This keys only off the variable, not the selected backend, and the GSA-required check runs only inside the postgres branch of `create_api_server_env_vars`. With the default redis backend and a stale `ATE_API_POSTGRES_CLOUDSQL_INSTANCE` in the environment, the sidecar still gets patched in, possibly with no Workload Identity annotation. Its startup probe then gates ateapi, and an otherwise healthy redis deployment times out. I think both this function and the ConfigMap block want a `store_backend == postgres` guard."

hack/install-ate.sh
```diff
@@ -491,9 +505,27 @@ create_api_server_env_vars() {
...
+      # .gserviceaccount.com suffix trimmed.
+      if [[ -z "${ATE_API_POSTGRES_CLOUDSQL_GSA:-}" ]]; then
+        echo "Error: ATE_API_POSTGRES_CLOUDSQL_INSTANCE requires ATE_API_POSTGRES_CLOUDSQL_GSA" \
+          "(or an explicit ATE_API_POSTGRES_CONNECTION_STRING)" >&2
+        exit 1
+      fi
+      postgres_connection_string="user=${ATE_API_POSTGRES_CLOUDSQL_GSA%.gserviceaccount.com} host=127.0.0.1 port=5432 dbname=atepg sslmode=disable"
+      # pgxpool sizing (per ateapi replica); defaults to max(4, NumCPU).
+      if [[ -n "${ATE_API_POSTGRES_POOL_MAX_CONNS:-}" ]]; then
+        postgres_connection_string+=" pool_max_conns=${ATE_API_POSTGRES_POOL_MAX_CONNS}"
```
jpbetz: The way the if/else statements are structured means this environment variable is not always honored. It will be hard to trace down if not as there is no obvious error, just client-side queuing..

hack/install-ate.sh
```diff
@@ -507,15 +539,62 @@ create_api_server_env_vars() {
...
+      --from-literal=CSQL_PROXY_AUTO_IAM_AUTHN="${ATE_API_POSTGRES_CLOUDSQL_IAM_AUTH:-true}"
+      --from-literal=CSQL_PROXY_PORT="5432"
+      --from-literal=CSQL_PROXY_HEALTH_CHECK="true"
+      --from-literal=CSQL_PROXY_HTTP_ADDRESS="0.0.0.0"
+      --from-literal=CSQL_PROXY_HTTP_PORT="9801"
+      --from-literal=CSQL_PROXY_STRUCTURED_LOGS="true"
+    )
+    case "${ATE_API_POSTGRES_CLOUDSQL_IP_TYPE:-private}" in
+      private) cm_args+=(--from-literal=CSQL_PROXY_PRIVATE_IP="true") ;;
+      psc) cm_args+=(--from-literal=CSQL_PROXY_PSC="true") ;;
```
jpbetz, line 737: The tool only creates private-services-access instances (`Ipv4Enabled=false` + `PrivateNetwork`; no PSC config in `cloudSQLInstanceSpec`). But cloud-sql.md lists `psc` as supported...

tools/setup-gcp/cloud-sql.md
```diff
@@ -0,0 +1,174 @@
...
+
+The cluster must have Workload Identity enabled (clusters created by
+`setup-gcp create cluster` do), and the VPC needs [private services
+access](https://cloud.google.com/sql/docs/postgres/configure-private-services-access)
+(one-time per VPC; the tool prints the two `gcloud` commands if it is
+missing). Then:
+
+```sh
+export PROJECT_ID=<project>
+go run ./tools/setup-gcp create cloudsql   # flags: --instance, --tier, --edition, --storage-size, --gsa-name, --network
```
jpbetz: This needs `--region` else you get it defaulted to us-central1 and could end up with a cross-region instance.

tools/setup-gcp/cmd/cloudsql.go
```diff
@@ -0,0 +1,448 @@
...
+
+	"cloud.google.com/go/iam/apiv1/iampb"
+	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
+	serviceusage "cloud.google.com/go/serviceusage/apiv1"
+	"cloud.google.com/go/serviceusage/apiv1/serviceusagepb"
+	"github.com/spf13/cobra"
+	"google.golang.org/api/googleapi"
+	iam "google.golang.org/api/iam/v1"
+	servicenetworking "google.golang.org/api/servicenetworking/v1"
+	sqladmin "google.golang.org/api/sqladmin/v1"
```
jpbetz, line 34: Not a problem created by this PR, but it looks like this is GCP specific code but it's not in it's own go.mod so it's putting these imports into the root module's vendor tree. Should we make this a separate module?

jpbetz (commented):
A couple issues, mostly AI detected, but good overall.

Biggest high level gaps I noticed:
- No backup plan
- No deletion protection
- Configuration only applies on create, and silently noops against already running pods

The sidecar approach is a nice way to  keep the GCE could provider specific code is organized in a way that would extend to other cloud provides in a sane way. The credentials look good.  The configuration options look good.  I appreciate all the documentation and tests.

docs/dev/cloud-sql-scaling-guide.md
```diff
@@ -0,0 +1,93 @@
...
+
+- Persistent-disk IOPS and throughput scale with provisioned size — the disk
+  is also the I/O knob. Pre-size to ~2× the expected dataset (records +
+  indexes + WAL + bloat) rather than relying on auto-resize, which grows in
+  small steps and stalls under bulk loads.
+
+## Connection sizing for a target throughput
+
+Set with `ATE_API_POSTGRES_POOL_MAX_CONNS` at deploy time. Connections are
+concurrency, so `needed ≈ QPS × mean latency`. At 10k ops/s and ~6ms per
```
maxsmythe: QPS x 1/ (mean latency in seconds) ?

hack/install-ate.sh
shrutiyam-glitch: Done — intent now comes from the cluster rather than the shell: an unset variable adopts the recorded configuration, and only an explicit `ATE_API_POSTGRES_CLOUDSQL_INSTANCE=""` removes the sidecar.

hack/install-ate.sh
shrutiyam-glitch: Added a guard - the standalone flag now refuses to run when the live Deployment doesn't reference the Secret yet, and points at `--deploy-ate-apiserver`, which updates both.

hack/install-ate.sh
shrutiyam-glitch: Done — the skip now logs which external database it deferred to. While adding it I found the skip wasn't actually skipping on GKE:` ko resolve -f manifests/ate-install` picks up top-level `postgres.yaml,` so the bundle applied the `StatefulSet` regardless. so, moved it into a subdirectory.

hack/install-ate.sh
shrutiyam-glitch: Since redis backend is gone from main - postgreSQL is the only store backend now, so there's no backend left to guard on.

hack/install-ate.sh
shrutiyam-glitch: Updated, thanks !

tools/setup-gcp/cmd/cloudsql.go
shrutiyam-glitch: Yeah, can we handle that in a follow up PR ?

cmd/ate-setup/internal/steps/postgres.go
```diff
@@ -50,7 +50,7 @@ func (e *Env) DeployPostgres(ctx context.Context) error {
 		return err
 	}
 
-	if err := e.Kube.ApplyPath(ctx, e.Cfg.Manifest("postgres.yaml")); err != nil {
+	if err := e.Kube.ApplyPath(ctx, e.Cfg.Manifest("postgres", "postgres.yaml")); err != nil {
```
jpbetz, line 69: I think this might break `ate-setup deploy ate-system` on GKE. For a plain GKE install `renderSystemManifests` hands the `manifests/ate-install` directory to ko, which does not descend into subdirectories, so the rendered bundle no longer contains the StatefulSet. `DeployAteSystem` then waits on `StatefulSet/postgres` unconditionally and fails with `StatefulSet/postgres not found in namespace ate-system`. CI can't see it: the workflow only runs the shell script on kind, and the kind overlay never carried postgres.

Suggest mirroring `use_bundled_postgres` from the shell script: have `DeployAteSystem` apply `postgres/postgres.yaml` when `ATE_API_POSTGRES_CONNECTION_STRING` is unset and gate the StatefulSet wait on the same condition. The `deploy postgres` help text ("already brings PostgreSQL up as part of the rendered bundle") needs the matching update.

tools/setup-gcp/cmd/cloudsql.go
```diff
@@ -0,0 +1,457 @@
...
+		// only to connect to public/PSC instances provisioned out-of-band.
+		IpConfiguration: &sqladmin.IpConfiguration{
+			Ipv4Enabled:     false,
+			PrivateNetwork:  privateNetworkURL(cfg),
+			ForceSendFields: []string{"Ipv4Enabled"},
+		},
+		DatabaseFlags: []*sqladmin.DatabaseFlags{
+			{Name: "cloudsql.iam_authentication", Value: "on"},
+		},
+	}
```
jpbetz, line 217: The instance is created with no backups and no deletion protection: `Settings` sets neither `BackupConfiguration` nor `DeletionProtectionEnabled`, and with the Admin API an omitted `backupConfiguration` leaves automated backups and point-in-time recovery off. This store is the control plane's entire state, so I'd default both on and let operators opt out. Deletion protection means teardown needs `gcloud sql instances patch --no-deletion-protection` first, which is the point.

```suggestion
		DatabaseFlags: []*sqladmin.DatabaseFlags{
			{Name: "cloudsql.iam_authentication", Value: "on"},
		},
		// Daily automated backups with point-in-time recovery, and a guard
		// against accidental deletion; clear it with `gcloud sql instances
		// patch --no-deletion-protection` before tearing the instance down.
		BackupConfiguration: &sqladmin.BackupConfiguration{
			Enabled:                    true,
			PointInTimeRecoveryEnabled: true,
		},
		DeletionProtectionEnabled: true,
	}
```

Builds and `go test ./tools/setup-gcp/...` passes with this applied.

hack/install-ate.sh
```diff
@@ -542,24 +556,217 @@ wait_for_podcertificate_trust_bundles() {
...
+    elif [[ "${postgres_connection_string}" == *"://"*"?"* ]]; then
+      postgres_connection_string+="&pool_max_conns=${ATE_API_POSTGRES_POOL_MAX_CONNS}"
+    elif [[ "${postgres_connection_string}" == *"://"* ]]; then
+      postgres_connection_string+="?pool_max_conns=${ATE_API_POSTGRES_POOL_MAX_CONNS}"
+    else
+      postgres_connection_string+=" pool_max_conns=${ATE_API_POSTGRES_POOL_MAX_CONNS}"
+    fi
   fi
 
   echo "POSTGRES_CONNECTION_STRING: ${postgres_connection_string}"
```
jpbetz: nit: this line stale now

```suggestion
  # Redact any password before logging (URI user:pw@host and keyword password=).
  echo "POSTGRES_CONNECTION_STRING: $(printf '%s' "${postgres_connection_string}" \
    | sed -E 's#(://[^:/@]*):[^@]*@#\1:***@#; s/(password=)[^ &]*/\1***/g')"
```

cmd/ate-setup/internal/steps/postgres.go
shrutiyam-glitch: `ate-setup` introduced recently (#1190) and yeah it was breaking, `DeployAteSystem` was relying on the sweep and waiting unconditionally. 
Updated it now to apply `postgres/postgres.yaml` and gates the rollout wait only when no external DB is configured (with skip logging and updated help text). Verified both bundled (1/1) and external paths on GKE, and added unit tests guarding against manifest placement drift.

`ate-setup` doesn't yet support `ATE_API_POSTGRES_CLOUDSQL_*`, which I've documented in `differences.md`. I can file a follow-up issue for the port, or fold it in here if you'd prefer.

PR #1002, opened 2026-08-17
test(k8sresolver): assert on convergence, not the first update

### LINKED ISSUE #1001: E2E test flake: TestK8sResolverEndpointSliceUpdates asserts on the first update instead of convergence
## Symptom

`TestK8sResolverEndpointSliceUpdates` intermittently fails at `resolver_test.go:215`:

```
updated state.Addresses = [{Addr: "10.0.0.1:443", ServerName: "", }],
                    want [{Addr: "10.0.0.1:443", ServerName: "", } {Addr: "10.0.0.2:443", ServerName: "", }]
```

Hit rate is high: **8 of the 30 most recent failed `pr-workflow` runs (~27%)**. All eight verified individually — each fails on this test at `resolver_test.go:215`:

- https://github.com/agent-substrate/substrate/actions/runs/32045558292
- https://github.com/agent-substrate/substrate/actions/runs/32045171791
- https://github.com/agent-substrate/substrate/actions/runs/31987832602
- https://github.com/agent-substrate/substrate/actions/runs/31978191442
- https://github.com/agent-substrate/substrate/actions/runs/31937546256
- https://github.com/agent-substrate/substrate/actions/runs/31916613395
- https://github.com/agent-substrate/substrate/actions/runs/31851248168
- https://github.com/agent-substrate/substrate/actions/runs/31849526035

## Cause

`Build` registers the informer event handlers and also starts a goroutine that reports once the cache has synced:

```go
go func() {
    if cache.WaitForCacheSync(ctx.Done(), inf.Informer().HasSynced) {
        r.updateState()
    }
}()
```

That extra report is deliberate and correct: without it, a service with no EndpointSlices at all would never fire `AddFunc`, and the resolver would never call `UpdateState` — leaving gRPC waiting instead of learning the answer is an empty set.

The catch is how it waits. `WaitForCacheSync` polls with `syncedPollPeriod = 100 * time.Millisecond` (`vendor/k8s.io/client-go/tools/cache/shared_informer.go:369`) via `PollImmediateUntil`. The immediate check normally finds the cache not yet synced, so the goroutine fires at **roughly 100ms after Build** — exactly once, then exits.

So there is a race between that 100ms timer and how fast the test gets from its first `select` to creating the second EndpointSlice:

| test reaches the Create | the ~100ms report contains | result |
|---|---|---|
| within 100ms | both addresses | passes |
| **after 100ms** | **only `10.0.0.1`** | **fails** |

The second `select` takes whatever is next in the channel, so on a slow runner it takes that stale report and asserts against it.

Measured timeline, with a 150ms sleep inserted between the first `select` and the Create to stand in for CI scheduling delay:

```
[  0.5ms]  update #1: [10.0.0.1]              <- AddFunc for slice1
           test's first select takes it
           ... 150ms stall ...
[101.1ms]  update #2: [10.0.0.1]              <- the WaitForCacheSync goroutine
           test creates slice2
[151.4ms]  update #3: [10.0.0.1, 10.0.0.2]
```

Without the stall, Build-to-Create is **0.6ms** locally — far ahead of the timer — which is why this never reproduces on a developer machine and only surfaces on loaded runners.

Note the resolver is not at fault. It never promised the first update after a change is the final one; it promises eventual convergence, and a duplicate report costs gRPC nothing. The test asserted on the wrong contract.

## Verification

`-count=N` proves nothing here — the unfixed test passes locally at any count because the window is never hit. A/B with the delay as the only variable:

| Build→Create delay | old assertion | new assertion |
|---|---|---|
| 0ms | 5/5 pass | 5/5 pass |
| 150ms | **0/5 pass** | **5/5 pass** |

orangeCatDeveloper:
Fixes #1001

`TestK8sResolverEndpointSliceUpdates` intermittently fails at `resolver_test.go:215` — 8 of the 30 most recent failed `pr-workflow` runs (~27%):

```
updated state.Addresses = [{Addr: "10.0.0.1:443", ServerName: "", }],
                    want [{Addr: "10.0.0.1:443", ServerName: "", } {Addr: "10.0.0.2:443", ServerName: "", }]
```

`Build` starts a goroutine that calls `updateState` once `WaitForCacheSync` returns. That report is deliberate: a service with no EndpointSlices never fires `AddFunc`, so without it the resolver would stay silent instead of telling gRPC the answer is an empty set. But `WaitForCacheSync` polls at `syncedPollPeriod = 100ms`, so it fires roughly 100ms after Build — and if the test has not yet created the second slice by then, that update still carries only `10.0.0.1` and sits in the channel ahead of the real one. The second `select` took whatever came next, so it asserted against the stale update.

Locally Build-to-Create is 0.6ms, well ahead of the timer, which is why this only shows up on loaded runners.

The resolver is not at fault — it promises eventual convergence, not that the first update after a change is final, and a duplicate update costs gRPC nothing. So both waits now go through one `waitForAddrs` helper that consumes updates until the set matches, with a timeout so a genuinely broken resolver still fails and reports the last set it saw. Note this changes the first wait as well: it asserted the *first* update equals `[10.0.0.1]`, and now waits for that set instead. The same argument applies there — nothing promises the first update is final.

## Verification

`-count=N` proves nothing here: the unfixed test passes locally at any count because the window is never hit. A/B with the Build→Create delay as the only variable:

| delay | old assertion | new assertion |
|---|---|---|
| 0ms | 5/5 pass | 5/5 pass |
| 150ms | **0/5 pass** | **5/5 pass** |

Measured timeline with the 150ms stall in place:

```
[  0.5ms]  update #1: [10.0.0.1]              <- AddFunc for slice1
           first select takes it
           ... 150ms stall ...
[101.1ms]  update #2: [10.0.0.1]              <- the WaitForCacheSync goroutine
           test creates slice2
[151.4ms]  update #3: [10.0.0.1, 10.0.0.2]
```

The probe tests used for this are not included.

Rebased over #1013. That fixes a different bug — concurrent `updateState` calls letting an older address set win — and does not close this one: the two updates here are ~100ms apart, so the queue has nothing to coalesce. Re-measured on top of it, unchanged: at a 150ms delay the old assertion is 0/5 and the new one 5/5.

### DIFF
--- a/internal/k8sresolver/resolver_test.go
+++ b/internal/k8sresolver/resolver_test.go
@@ -54,6 +54,25 @@ func (c *testClientConn) ParseServiceConfig(serviceConfigJSON string) *serviceco
 	return nil
 }
 
+// waitForAddrs consumes updates until the set matches want: the resolver promises
+// eventual convergence, not that the first update after a change is final.
+func waitForAddrs(t *testing.T, cc *testClientConn, want []resolver.Address) {
+	t.Helper()
+	deadline := time.After(2 * time.Second)
+	var last []resolver.Address
+	for {
+		select {
+		case state := <-cc.stateChan:
+			last = state.Addresses
+			if reflect.DeepEqual(last, want) {
+				return
+			}
+		case <-deadline:
+			t.Fatalf("state.Addresses never converged: last = %v, want %v", last, want)
+		}
+	}
+}
+
 func TestParseTarget(t *testing.T) {
 	tests := []struct {
 		name          string
@@ -168,15 +187,7 @@ func TestK8sResolverEndpointSliceUpdates(t *testing.T) {
 	}
 	defer res.Close()
 
-	select {
-	case state := <-cc.stateChan:
-		wantAddrs := []resolver.Address{{Addr: "10.0.0.1:443"}}
-		if !reflect.DeepEqual(state.Addresses, wantAddrs) {
-			t.Errorf("initial state.Addresses = %v, want %v", state.Addresses, wantAddrs)
-		}
-	case <-time.After(2 * time.Second):
-		t.Fatal("timed out waiting for initial state update")
-	}
+	waitForAddrs(t, cc, []resolver.Address{{Addr: "10.0.0.1:443"}})
 
 	// Add a new EndpointSlice
 	slice2 := &discoveryv1.EndpointSlice{
@@ -205,18 +216,10 @@ func TestK8sResolverEndpointSliceUpdates(t *testing.T) {
 		t.Fatalf("failed to create slice2: %v", err)
 	}
 
-	select {
-	case state := <-cc.stateChan:
-		wantAddrs := []resolver.Address{
-			{Addr: "10.0.0.1:443"},
-			{Addr: "10.0.0.2:443"},
-		}
-		if !reflect.DeepEqual(state.Addresses, wantAddrs) {
-			t.Errorf("updated state.Addresses = %v, want %v", state.Addresses, wantAddrs)
-		}
-	case <-time.After(2 * time.Second):
-		t.Fatal("timed out waiting for state update after slice addition")
-	}
+	waitForAddrs(t, cc, []resolver.Address{
+		{Addr: "10.0.0.1:443"},
+		{Addr: "10.0.0.2:443"},
+	})
 }
 
 func TestK8sResolverClose(t *testing.T) {


internal/k8sresolver/resolver_test.go
```diff
@@ -205,17 +205,24 @@ func TestK8sResolverEndpointSliceUpdates(t *testing.T) {
...
-		}
-		if !reflect.DeepEqual(state.Addresses, wantAddrs) {
-			t.Errorf("updated state.Addresses = %v, want %v", state.Addresses, wantAddrs)
+	// Build's WaitForCacheSync goroutine polls at 100ms and can queue a stale
+	// one-address update ahead of this one, so consume until the set converges.
+	wantAddrs := []resolver.Address{
+		{Addr: "10.0.0.1:443"},
+		{Addr: "10.0.0.2:443"},
+	}
+	deadline := time.After(5 * time.Second)
```
michaelasp: The resolver shouldn't take more than a second or 2 to converge, should we bump this timeout? I guess it helps us in the case that we are on a heavily loaded CI node and the usual time the test will take is less than a second because we return the second we see that the addresses are equal.

internal/k8sresolver/resolver_test.go
michaelasp: Since this is being used here as well can we make this a single helper function so we don't diverge any place we check for this convergence?

internal/k8sresolver/resolver_test.go
```diff
@@ -205,17 +205,24 @@ func TestK8sResolverEndpointSliceUpdates(t *testing.T) {
...
+	var lastAddrs []resolver.Address
+	for {
+		select {
+		case state := <-cc.stateChan:
```
michaelasp: ~~Actually thinking about this more, might this be a race condition as well? What if both the update func and the `WaitForCacheSync` run at the same time. We may get the older information second unless we wrap the critical section in a mutex. It's very unlikely since the cache sync should effectively happen immediately but there is the case for that or if some worker calls `ResolveNow`.~~

That is a separate issue. Created #1012 for that and #1013 to fix.

internal/k8sresolver/resolver_test.go
orangeCatDeveloper: I created a helper function in the latest commit

internal/k8sresolver/resolver_test.go
orangeCatDeveloper: Changed to 2s instead, matching what the first wait already used.

BenTheElder (approved):

PR #1007, opened 2026-08-17
ateom-microvm: smaller guests

BenTheElder:
Shrink unnecessary memory floor, right-size the counter demo.

I measured actual needed memory and relaxed the floor.

This will really help with performance (snapshot size, density, memory snapshot + restore timing).

Next up after https://github.com/agent-substrate/substrate/pull/972

### DIFF
--- a/cmd/ateom-microvm/run.go
+++ b/cmd/ateom-microvm/run.go
@@ -134,11 +134,15 @@ const vmmMemReserveMiB = 256
 // minGuestMemMiB is the floor for guest RAM (the declared limit minus the VMM
 // reserve); a declared memory limit that leaves less is rejected at cold boot with a
 // clear error instead of being silently honored (see resolveGuestMemMiB), since too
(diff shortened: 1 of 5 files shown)


BenTheElder:
Noting: The failures so far are all flakes occurring on `main`, we have some other PRs in flight for that (and one is new, from #893 it appears)

BenTheElder:
https://github.com/agent-substrate/substrate/pull/1021 addresses the most persistent failure (again, unrelated to this PR, but causing test failures on it)

HavenXia (approved):

BenTheElder:
FYI @maxsmythe, glutton is currently not setting limits => getting 2GB memory arbitrarily for actors on uVM

Probably we should require setting limits and drop that default (cc @Oneimu @zoez7), but in the meantime we should explicitly configure glutton for benchmarking with how much memory it needs. I'm not sure how much it needs though ... this PR should give an idea for counter demo for example.

PR #1008, opened 2026-08-17
ci/cd: Enable external volume e2e test by default

hajiler:
https://github.com/agent-substrate/substrate/issues/232

Install CSI NFS driver by default when installing ate, and default to the NFS storage class for external volumes test. 

- [X ] Tests pass

### DIFF
--- a/cmd/ate-setup/internal/cmd/deploy.go
+++ b/cmd/ate-setup/internal/cmd/deploy.go
@@ -95,6 +95,7 @@ func init() {
 		deployPostgresCmd,
 	)
 
-	deployAteSystemCmd.Flags().BoolVar(&deployOpts.SetupCSI, "setup-csi", false,
-		"Also install the hostpath and NFS CSI drivers (Kind only)")
+	deployAteSystemCmd.Flags().StringVar(&deployOpts.SetupCSI, "setup-csi", "none",
+		"Also install CSI driver (nfs, hostpath, both, none; default: none)")
+	deployAteSystemCmd.Flags().Lookup("setup-csi").NoOptDefVal = "none"
 }
--- a/cmd/ate-setup/internal/demos/counter/counter.go
+++ b/cmd/ate-setup/internal/demos/counter/counter.go
@@ -27,16 +27,17 @@ import (
 	"github.com/agent-substrate/substrate/internal/resources"
 )
 
-// namespace is the pool's k8s namespace; it doubles as the atespace holding
-// the demo's ActorTemplate.
-const namespace = "ate-demo-counter"
+const (
+	namespace = "ate-demo-counter"
+	template  = "demos/counter/counter-template.yaml.tmpl"
 
-// externalVolumeValues fills the optional external-volume placeholder lines
-// of the counter template manifest, which a plain deploy drops.
-func externalVolumeValues(e *steps.Env) map[string]string {
-	storageClass := "standard"
-	if e.Cfg.Kind {
-		storageClass = "csi-hostpath-sc"
+	defaultStorageClass = "standard"
+)
+
+func (d *demo) externalVolumeValues(e *steps.Env) map[string]string {
+	sc := defaultStorageClass
+	if d.storageClass != "" {
+		sc = d.storageClass
 	}
 	return map[string]string{
 		"VALIDATE_EXISTING_FILE_PATH_ARG": "  - --validate-existing-file-path=/external-data/test.txt",
@@ -46,7 +47,7 @@ func externalVolumeValues(e *steps.Env) map[string]string {
 			"  type: ExternalVolumeTemplate\n" +
 			"  externalVolumeTemplate:\n" +
 			"    capacity: 1Gi\n" +
-			"    storageClassName: " + storageClass,
+			"    storageClassName: " + sc,
 	}
 }
 
@@ -56,6 +57,7 @@ type demo struct {
 	demos.Substrate
 
 	withExternalVolume bool
+	storageClass       string
 }
 
 func init() {
@@ -77,6 +79,8 @@ func init() {
 func (d *demo) Flags(fs *pflag.FlagSet) {
 	fs.BoolVar(&d.withExternalVolume, "with-external-volume", false,
 		"Attach an external volume and validate a pre-seeded file on it (run \"setup csi\" first)")
+	fs.StringVar(&d.storageClass, "storage-class", defaultStorageClass,
+		"StorageClass backing the external volume, e.g. csi-nfs-sc or csi-hostpath-sc. Must be used --with-external-volume=true.")
 }
 
 // renderValues opts the template's external-volume lines in when the flag is
@@ -85,5 +89,5 @@ func (d *demo) renderValues(_ context.Context, e *steps.Env) (map[string]string,
 	if !d.withExternalVolume {
 		return nil, nil
 	}
-	return externalVolumeValues(e), nil
+	return d.externalVolumeValues(e), nil
 }
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -68,7 +68,8 @@ function usage() {
   echo "Overall infrastructure (all infrastructure components):"
   echo ""
   echo "  --deploy-ate-system                    Deploy core system (CRDs, atelet, apiserver)"
-  echo "  --setup-csi                            Setup CSI hostpath and NFS drivers (Kind only)"
+  echo "  --setup-csi[=DRIVER]                   Setup CSI driver: nfs, hostpath, both, none (default: none;"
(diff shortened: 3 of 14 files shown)


hack/create-kind-cluster.sh
```diff
@@ -84,6 +84,17 @@ else
   echo "/dev/kvm not available: micro-VM support disabled (gVisor still works)."
 fi
 
+# Attempt to load NFS kernel modules on the host if not already active
```
msau42: This may be a surprise if users are runing this on their local workstation. Can you put this under the csi flag and also maybe add a user prompt to proceed?

msau42: Also this is already run in setup-csi-nfs.sh. Do we also need to do it here?

hack/install-ate.sh
```diff
@@ -67,13 +67,14 @@ function usage() {
   echo "Overall infrastructure (all infrastructure components):"
   echo ""
   echo "  --deploy-ate-system                    Deploy core system (CRDs, atelet, apiserver)"
-  echo "  --setup-csi                            Setup CSI hostpath and NFS drivers (Kind only)"
+  echo "  --setup-csi                            Setup CSI NFS driver"
```
msau42: Maybe we can make this flag with "hostpath" or "nfs" arguments to allow users to install none, either or both drivers.

For basic testing, maybe hostpath is good enough. NFS is mainly for multi-node testing.

hack/create-kind-cluster.sh
hajiler: Removed, to solely rely on setup-csi-nfs..

hack/install-ate.sh
hajiler: Updated to specify none, either, or both. NFS by default.

hack/install-ate.sh
```diff
@@ -779,6 +799,14 @@ delete_ate_system() {
   if [[ "${ATE_INSTALL_KIND:-false}" == "true" ]]; then
     kubectl kustomize manifests/ate-install/kind --load-restrictor LoadRestrictionsNone \
       | run_kubectl delete --ignore-not-found -f -
+    run_kubectl delete --ignore-not-found csidriverconfig nfs.csi.k8s.io || true
```
msau42: I believe kubectl has an option where you can give it a directory with manifests and it will delete all resources with those names in the manifest.

hack/setup-csi-nfs.sh
```diff
@@ -18,10 +18,32 @@ set -o errexit -o nounset -o pipefail
 
 # Check if NFS server module is loaded on the host
 if [ -f /proc/filesystems ] && ! grep -q "nfsd" /proc/filesystems; then
-  echo "ERROR: NFS server support (nfsd) is not active in the host kernel."
-  echo "Please load the nfsd module on your host by running:"
-  echo "  sudo modprobe nfsd"
-  exit 1
+  if [ "$(id -u)" -eq 0 ]; then
+    modprobe nfs 2>/dev/null || true
```
msau42: I think in general we don't want to automatically enable the kernel modules, and just return error if it's non-interactive. We have it already enabled in the github workslow so that should be ok for CI.

msau42 (commented):
https://github.com/agent-substrate/substrate/pull/1190 is converting the hack scripts into golang.  We should rebase on that.

cmd/ate-setup/internal/demos/counter/counter.go
```diff
@@ -44,7 +45,7 @@ func (d *demo) externalVolumeValues(e *steps.Env) map[string]string {
 		"EXTERNAL_VOLUMES": "  - name: external-data\n" +
 			"    externalVolumeTemplate:\n" +
 			"      capacity: 1Gi\n" +
-			"      storageClassName: " + storageClass,
+			"      storageClassName: " + sc,
```
msau42: we probably shouldn't assume every storage class ends in "-sc"

hack/install-ate-kind.sh
```diff
@@ -32,6 +32,8 @@ export KO_DEFAULTPLATFORMS="linux/${goarch}"
 export ATE_INSTALL_KIND="true"
 # use default bucket name for local deployment
 export BUCKET_NAME="ate-snapshots"
+# install CSI driver by default on Kind
+export SETUP_CSI="${SETUP_CSI:-nfs}"
```
msau42: i'm actually thinking we may want to install hostpath by default for kind because it doesn't require the nfs modules

cmd/ate-setup/internal/cmd/deploy.go
```diff
@@ -95,6 +95,7 @@ func init() {
 		deployPostgresCmd,
 	)
 
-	deployAteSystemCmd.Flags().BoolVar(&deployOpts.SetupCSI, "setup-csi", false,
-		"Also install the hostpath and NFS CSI drivers (Kind only)")
+	deployAteSystemCmd.Flags().StringVar(&deployOpts.SetupCSI, "setup-csi", "nfs",
```
msau42: I think by default we may not want to install a CSI driver.

hack/install-ate.sh
```diff
@@ -69,7 +69,7 @@ function usage() {
   echo "Overall infrastructure (all infrastructure components):"
   echo ""
   echo "  --deploy-ate-system                    Deploy core system (CRDs, atelet, apiserver)"
-  echo "  --setup-csi                            Setup CSI hostpath and NFS drivers (Kind only)"
+  echo "  --setup-csi[=DRIVER]                   Setup CSI driver: nfs (default), hostpath, both, none"
```
msau42: @bowei are we still maintaining this script now that we have the tool?

hack/install-ate.sh
```diff
@@ -965,7 +986,8 @@ done
 # flag they configure (e.g. --benchmark-worker-count before/after
 # --deploy-benchmarks). The dispatch loop below also accepts these flags but
 # treats them as no-ops since the value is already captured here.
-SETUP_CSI="${SETUP_CSI:-false}"
+# Valid values for SETUP_CSI: nfs, hostpath, both, none (default: nfs).
+SETUP_CSI="${SETUP_CSI:-nfs}"
```
msau42: same thing I don't think we want to install a csi driver by default

cmd/ate-setup/internal/cmd/deploy.go
hajiler: Done.

hack/install-ate.sh
hajiler: I had run into some failures after #1190  was merged because the script was missing the defaulting changes. I think we should update it for now.

hack/install-ate.sh
hajiler: Done.

cmd/ate-setup/internal/demos/counter/counter.go
hajiler: sc is a variable with the actual name of the storage class being used.

hack/setup-csi-nfs.sh
hajiler: Done.

hack/install-ate.sh
hajiler: Done.

hack/install-ate-kind.sh
hajiler: I think that makes the install complexity a bit high, if NFS install NFS works. But can make that change as well

cmd/ate-setup/internal/demos/counter/counter.go
```diff
@@ -77,6 +79,8 @@ func init() {
 func (d *demo) Flags(fs *pflag.FlagSet) {
 	fs.BoolVar(&d.withExternalVolume, "with-external-volume", false,
 		"Attach an external volume and validate a pre-seeded file on it (run \"setup csi\" first)")
+	fs.StringVar(&d.storageClass, "storage-class", defaultStorageClass,
```
msau42, line 82: by default can we not require volumes in the counter demo?

Only enable it for e2e or opt-in

hajiler: That added flag is just to set which storageClass that would be used. 

This existing flag actually determines if 
https://github.com/hajiler/substrate/blob/23633f571a49e344ddb5e262e4be699089a5beea/cmd/ate-setup/internal/demos/counter/counter.go#L78-L79.

I updated the flag description to make it clear --with-external-volumes must be set to true first. Alternatively, we could merge the flags into an enum type situtation, where "", or "false" is the default and means externalVolumes are not added.

msau42: awesome thanks

msau42 (approved):

PR #1010, opened 2026-08-17
atelet: close ateom conns evicted from the dialer cache

### LINKED ISSUE #1009: # atelet: AteomDialer never closes conns evicted from its LRU cache
Found this bug while reviewing #961

`AteomDialer` caches one `grpc.ClientConn` per worker pod UID in `lru.New(256)` (`cmd/atelet/main.go:183`). `k8s.io/utils/lru` without an eviction func just drops the oldest entry when the cache is full — and nothing anywhere calls `Close()` on the dropped conn. Cache entries are also never removed when a worker pod dies, so the cache fills with dead UIDs over time.

A `ClientConn` that's dropped but not closed keeps its goroutines and memory around; gRPC requires an explicit `Close()`. So every eviction is a leaked conn.

Today this is slow: the cache only sees pods that lifecycle RPCs explicitly target, so it takes >256 distinct worker pods dialed *since atelet started* before evictions begin, and an atelet restart resets the count. On a long-running atelet on a churny node it's a steady drip; mostly nobody notices.

## Repro

In-process repro with the repo's exact deps (grpc v1.83.0, same `k8s.io/utils/lru`, same dial options as `DialAteomPod`): 32-entry cache, 100 fake ateom sockets swept per tick, one failed RPC per probe, 10 ticks.

- No eviction func (current code): goroutines grow 1 → 4001, heap 0.4 → 12.7 MiB, linear, ~400 goroutines leaked per tick. GC reclaims nothing.
- With `NewWithEvictionFunc` closing evicted conns: flat at 129 goroutines / ~1 MiB across all ticks.
- gRPC's channel idle timeout (default 30 min) is not a rescue: after it fires, only 1 of each abandoned conn's 4 goroutines exits. ~3 goroutines and ~5 KiB heap per conn stay leaked permanently.

## Fix

Use `lru.NewWithEvictionFunc` and close the evicted conn. One-liner, fixes the flaw at the source regardless of who dials.

Separately, #961's poller should probably not go through this cache at all (open/close per probe), but that's that PR's call — this issue is just: don't drop conns without hanging up.

git286:
AteomDialer caches one grpc.ClientConn per worker pod UID in a 256-entry LRU, but the cache had no eviction function: a conn pushed out to make room was forgotten, never closed. grpc does not reclaim an un-Closed conn -- measured against a dead socket, each one holds 4 goroutines, and the channel idle timeout later parks only one of them, so ~3 goroutines and their buffers leak per eviction for the life of the process.

Today evictions need >256 distinct worker pods dialed since atelet started, so the leak is a slow drip reset by restarts. Close evicted conns via the cache's eviction hook. The trade is that an RPC still in flight on a conn that aged to the LRU tail now fails visibly instead of completing on a leaked conn; that needs the same >256-pod churn, and the failure is retryable.

Fixes #1009.

- [ ] Tests pass

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -1796,6 +1796,12 @@ type AteomDialer struct {
 // (the channel idle timeout parks only one of them). Closing on eviction can
 // fail an RPC still in flight on a conn that aged to the LRU tail, but that
 // failure is visible and retryable, unlike the leak.
+//
+// TODO: Consider pool semantics instead of a cache: a conn evicted for
+// capacity would drain — close only once its last in-flight RPC finishes
+// (e.g. refcounted checkout/release) — rather than being closed out from
+// under a caller. Worth revisiting if the retryable eviction failures show
+// up in practice.
 func newAteomDialer(size int) *AteomDialer {
 	return &AteomDialer{
 		conns: lru.NewWithEvictionFunc(size, func(_ lru.Key, value interface{}) {
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -184,9 +184,7 @@ func main() {
 		}()
 	}
 
-	ateomDialer := &AteomDialer{
-		conns: lru.New(256),
-	}
+	ateomDialer := newAteomDialer(256)
 
 	var gcpRegistryAuthn authn.Authenticator
 	if *gcpAuthForImagePulls {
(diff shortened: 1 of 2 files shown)


cmd/atelet/main.go
```diff
@@ -1552,6 +1550,20 @@ type AteomDialer struct {
...
+// newAteomDialer builds a dialer whose cache closes the connections it
+// evicts. A conn pushed out of the LRU without Close is not reclaimed: grpc
+// keeps most of its goroutines and buffers alive for the life of the process
+// (the channel idle timeout parks only one of them). Closing on eviction can
+// fail an RPC still in flight on a conn that aged to the LRU tail, but that
+// failure is visible and retryable, unlike the leak.
+func newAteomDialer(size int) *AteomDialer {
+	return &AteomDialer{
+		conns: lru.NewWithEvictionFunc(size, func(_ lru.Key, value interface{}) {
+			value.(*grpc.ClientConn).Close()
```
juli4n, line 1808: A made the same comment in #480 : If we blindly close on eviction, can this close a connection which is already in use?

juli4n:
IIRC ate-apiserver has the same issue as it also caches connections (to atelet).

cmd/atelet/main.go
git286: Get moves an entry to the front. Every RPC starts with DialAteomPod → Get, so a conn with an RPC in flight was most-recently-used the moment the RPC began. For it to reach the eviction end of the queue mid-RPC, atelet would have to dial 256 other distinct pod UIDs in the window between that RPC starting and finishing, which is extremely unlikely, right?

juli4n: I think this is true for atelet. How about the other leak in ate-apiserver?

git286: Correct me if I am wrong, but I think the same logic applies to apiserver: for a connection to be evicted while in use, the api server would have to dial 1024 other different atelets during that one call, which basically never happens?

Also checked the atelet proto: all 5 RPCs (MintActorCertificate, Run, Checkpoint, Restore,UploadPausedCheckpoint) are plain one-shot calls, no streaming, which makes the live eviction even less unlikely?

juli4n (approved):

PR #1013, opened 2026-08-17
Add workqueue as sole caller of updateState

### LINKED ISSUE #1012: Possible Race in K8sResolver
k8sResolver runs updateState in 3 possible locations, we should either encapsulate in a mutex or move to workqueue. Otherwise updates could get interleaved in a bad order.

Image

michaelasp:
Fixes #1012

Adds workqueue as the sole caller of updateState, which has 2 benefits. One is the race condition described in the bug, and second is aggregation of multiple callers into a single update to prevent spurious calls.

### DIFF
--- a/internal/k8sresolver/resolver.go
+++ b/internal/k8sresolver/resolver.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	discoverylisters "k8s.io/client-go/listers/discovery/v1"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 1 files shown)


michaelasp:
Looks like a flake, not related to changes.

juli4n (approved):

michaelasp:
Hitting #1016, seems like main is busted :(

BenTheElder:
Yeah, we have to land https://github.com/agent-substrate/substrate/pull/1021 before we can land the other flake fixes

BenTheElder (approved):

PR #1019, opened 2026-08-17
e2e: stop TestGracefulWorkerTerminationSuspend racing the actor to CRASHED

### LINKED ISSUE #1016: TestGracefulWorkerTerminationSuspend failing on main
This test is blocking merges currently (aside from other flakes), we just landed #893, seems to be after that change.

BenTheElder:
The test deletes the worker pod, sleeps a fixed 2s, then suspends the actor, which is meant to land while the graceful shutdown is under way. On micro-VM it lands after the actor has already reached CRASHED, so the suspend is rejected and the test fails on every run since #893 enabled it there - on main as well as on unrelated PRs:

>  termination_test.go: failed to suspend Actor: rpc error: code = FailedPrecondition desc = MarkSuspending prerequisite not met for Actor: demo/graceful-term-suspend-... (got: STATUS_CRASHED, want STATUS_RUNNING or STATUS_PAUSED)

This PR updates the test to poll for STATE_DRAINING instead of a fixed sleep.

Fixes https://github.com/agent-substrate/substrate/issues/1016

- [x] Tests pass

### DIFF
--- a/internal/e2e/suites/demo/termination_test.go
+++ b/internal/e2e/suites/demo/termination_test.go
@@ -16,6 +16,7 @@ package demo
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


BenTheElder:
/hold

so while we should avoid fixed timeouts, and tests *are* passing locally with this change where they failed before it, this doesn't make sense. it should stay crashed.

BenTheElder:
focusing on fixing the mechanism in ateom

PR #1021, opened 2026-08-17
ateom-microvm: fix graceful termination, cleanup dead code

### LINKED ISSUE #1016: TestGracefulWorkerTerminationSuspend failing on main

BenTheElder:
Fixes #1016

1. Fix graceful termination, which was broken by the drop of the "carrier container" hack when we switched to overlay for the rootfs writes
2. Drop dead code attempting to enable compatibility before that change. It's a lot of complexity and we're not there yet

- [x] Tests pass


sfunkenhauser (dismissed):

BenTheElder:
Finally dodged the other flakes.

PR #1022, opened 2026-08-17
re-assemble micro-VM assets when version pins move

sfunkenhauser:
install-microvm-deps.sh decided whether to run assemble.sh by testing only whether the five asset filenames existed in $OUT. Those names don't change when a version pin moves, so an asset dir built before a bump looks complete while holding the old bytes.

### DIFF
--- a/hack/microvm-assets/assemble.sh
+++ b/hack/microvm-assets/assemble.sh
@@ -36,6 +36,13 @@
 #
 # Env: ARCH (arm64|amd64, default arm64), KATA_VER (4.0.0), CH_VER (v53.0),
 #      OUT (default ./bin/microvm-assets/$ARCH, under the gitignored bin/).
+#
+# Always re-downloads and overwrites — there is no incremental mode. It clears
+# $OUT/.asset-versions before the first write and re-stamps it with the versions that
+# produced the set only once the run completes, so the stamp is present only on a dir
+# assembled end-to-end by those pins. install-microvm-deps.sh uses it to decide whether
+# a cached $OUT is still current.
+# `--print-stamp` prints that stamp for the current env and exits without downloading.
 
 set -o errexit -o nounset -o pipefail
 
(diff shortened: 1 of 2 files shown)


sfunkenhauser:
@BenTheElder PTAL

hack/microvm-assets/assemble.sh
```diff
@@ -36,6 +36,11 @@
 #
 # Env: ARCH (arm64|amd64, default arm64), KATA_VER (4.0.0), CH_VER (v53.0),
 #      OUT (default ./bin/microvm-assets/$ARCH, under the gitignored bin/).
+#
+# Always re-downloads and overwrites — there is no incremental mode. On success it
```
BenTheElder: We should probably fix that.

hack/microvm-assets/assemble.sh
```diff
@@ -117,6 +140,9 @@ cd "${OUT}"
 for f in cloud-hypervisor virtiofsd vmlinux rootfs.img configuration-clh.toml; do
   [ -f "$f" ] || { echo "MISSING: $f" >&2; exit 1; }
 done
+# Written only once all five are present, so an interrupted run can't leave behind a
```
BenTheElder: 🤖 **should-fix** 🟡 – This holds for a fresh `$OUT` but not for a re-assembly, where the *previous* run's stamp is still sitting there.

The script overwrites in place — `curl -fSL -o "${OUT}/cloud-hypervisor"` and the kata copies all write straight into `$OUT` — so an interrupted run leaves a half-written asset set beside a stamp written by the last successful one. Interrupt a re-run at the same pins (a flaky download, a Ctrl-C) and the stamp still matches what `--print-stamp` reports, so `install-microvm-deps.sh` skips assembly and stages a truncated binary. That surfaces at runtime as exactly the kind of confusing wedge this PR set out to remove.

A version bump is safe by luck rather than design: the stale stamp names the old versions, so the mismatch survives the interruption and forces a re-assemble. Only the same-pin case is exposed.

Removing the stamp before the first download rather than only writing it at the end closes it, and then the comment is true as written: no stamp at all until the set is complete.

BenTheElder: this only works the first time, then it will have a stale stamp file, right? we should remove it before we start writing so a half-done run doesn't show as complete

sfunkenhauser: Done

BenTheElder (commented):
/lgtm

BenTheElder (approved):

Issue #988, opened 2026-08-17
Update on postgres impl is inefficient


thockin:
Edit: This was originally about  updates being non-atomic but I was wrong.  

There does still seem to be a significant opportunity to optimize.

I am not a pgsql expert, so please correct me if I am wrong. 

@supreme-gg-gg 

Roughly, it looks like:
```
Query:
          SELECT proto FROM actors
          WHERE atespace = $1 AND name = $2
          FOR UPDATE`, atespace, name

Unmarshal
call mutate(oldVal)
validate
Marshal

Query:
          UPDATE actors
          SET version = $1, proto = $2
          WHERE atespace = $3 AND name = $4`,
          dbActor.GetMetadata().GetVersion(), protoBytes, atespace, name
```

I am not a PGSQL expert but I thought "FOR UPDATE" requires a txn.  Also gemini says not to use "FOR UPDATE" but "FOR NO KEY UPDATE".  IIUC, also: row-locks are not auto-released, so we need to handle a crash mid-txn. 

I assumed I would find something more like:

```
UPDATE actors
          SET version = $5, proto = $6
          WHERE atespace = $1 AND name = $2 AND uid = $3 AND version = $4,
          atespace, name, expectUid, expectVersion, newVersion, protoBytes
```

...and then check that it affected exactly 1 row.

EItanya:
I dug further into this last night and it appears as though the update is transaction safe, but there are definitely a couple of optimizations we could make.

1. We use `FOR UPDATE` even though no keys are updated so we can indeed switch to `FOR NO KEY UPDATE`.
2. As @jpbetz pointed out, we hold the transaction across client side marshal/unmarshal which absolutely presents potential performance issues.

jpbetz:
+100. I highly recommend switching entirely away from the pessimistic locking approach implemented for writes and switching to an optimistic locking approach like @thockin describes.  For the target scale numbers I've seen, the write path needs to be dramatically simplified.  You'll probably also need to drop the DB enforced uniqueness check on UID and rely on the server to enforce that uniqueness guarantee.

EDIT: See the scale notes I put here: https://github.com/agent-substrate/substrate/pull/940#issuecomment-5317189350

thockin:
Indeed, I missed the `tx, err := p.pool.Begin(ctx)` and `defer tx.Rollback(ctx)`

[...]

Issue #731, opened 2026-08-04
Introduce a PostgreSQL based backend


juli4n:
There have been multiple discussions around ValKey's durability guarantees and whether it's the right fit for Substrate. In #640 , @supreme-gg-gg proposed a PostgreSQL alternative backend (see https://docs.google.com/document/d/12-ko_BFHcBo_nJkx9f4B7zMbiiWKC2saGhMhZG3aQ-s/edit?pli=1&tab=t.0 for a detailed benchmark).

This task is to introduce an alternative store PostgreSQL implementation, controlled by a binary flag.

mtaufen:
Why an "alternative" solution, if ValKey isn't the right fit to begin with? Seems like we should just switch to PostgreSQL. 

Have we considered what to do for scale-out, after? What solutions are available for sharding PostgreSQL (anything like Vitess?)?

BenTheElder:
This is done. And it is now the default / only.

PR #1023, opened 2026-08-17
Harden PostgreSQL concurrency, integrity, and error handling

supreme-gg-gg:
Part of #731 and follow up to #640. This PR makes improvements to the Postgres store by addressing review comments on #940 and optimizing Postgres updates as suggested in #988.

- Replaces Postgres row-locking updates for actors, templates, and snapshot tags with bounded optimistic concurrency control using UID/version CAS checks.
- Strengthens relational integrity with missing FKs and indexes, and replace snapshot-tag creation pre-check with FK
- Removes unnecessary UID uniqueness constraints
- Standardize pagination and validation across stores, map malformed tokens and invalid sizes to appropriate gRPC errors
- Improves actor resume race handling when workers disappear or concurrent assignments exhaust their retry budget
- Improves reliability with startup retries, expired-lease cleanup, worker-watch reconnection after malformed msg
- Expand Postgres, Redis, Contract, and Control API tests to cover new constraints, error semantics, concurrency

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_resume.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume.go
@@ -83,6 +83,18 @@ func (w *ActorWorkflow) ResumeActor(ctx context.Context, actorRef resources.Acto
 			lifecycleOpAttrs(actor, actorTemplate, tele.SnapshotKind, tele.WireSnapshotScope)...)
 	}()
 
+	// Routed requests call ResumeActor even when the actor is already running.
+	// Read before taking the distributed lease so that hot-path checks do not
+	// upsert and delete a PostgreSQL lease row. Any state that needs work is read
+	// again under the lock below.
+	actor, err = w.store.GetActor(ctx, actorRef)
+	if err != nil {
+		return nil, false, err
+	}
+	if wasRunning = actor.GetStatus().GetState() == ateapipb.ActorState_ACTOR_STATE_RUNNING; wasRunning {
+		return actor, false, nil
+	}
+
 	lockCtx, lock, err := w.acquireActorLock(ctx, actorRef)
 	if err != nil {
 		return nil, false, err
@@ -330,6 +342,9 @@ func (w *ActorWorkflow) ensureWorkerAssigned(ctx context.Context, actorRef resou
 		return false, attemptErr
 	})
 	if err != nil {
+		if wait.Interrupted(err) && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
+			return nil, nil, store.ErrVersionConflict
+		}
 		return nil, nil, err
 	}
 	return assignedActor, assignedWorker, nil
@@ -506,6 +521,10 @@ func (w *ActorWorkflow) assignWorkerAttempt(ctx context.Context, actorRef resour
 	}
 
 	if err := w.store.UpdateWorker(ctx, assignedWorker, assignedWorker.GetMetadata().GetVersion()); err != nil {
+		if errors.Is(err, store.ErrNotFound) {
+			w.workerCache.Forget(assignedWorker.GetWorkerNamespace(), assignedWorker.GetWorkerPod())
+			return nil, nil, fmt.Errorf("selected worker disappeared before claim: %w", store.ErrVersionConflict)
+		}
 		return nil, nil, err
 	}
 
--- a/cmd/ateapi/internal/controlapi/workflow_resume.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume.go
@@ -522,7 +522,7 @@ func (w *ActorWorkflow) assignWorkerAttempt(ctx context.Context, actorRef resour
 
 	if err := w.store.UpdateWorker(ctx, assignedWorker, assignedWorker.GetMetadata().GetVersion()); err != nil {
 		if errors.Is(err, store.ErrNotFound) {
-			w.workerCache.Forget(assignedWorker.GetWorkerNamespace(), assignedWorker.GetWorkerPod())
+			w.workerCache.Forget(assignedWorker.GetMetadata().GetName())
 			return nil, nil, fmt.Errorf("selected worker disappeared before claim: %w", store.ErrVersionConflict)
 		}
 		return nil, nil, err
--- a/cmd/ateapi/internal/store/atepg/atepg.go
+++ b/cmd/ateapi/internal/store/atepg/atepg.go
@@ -128,6 +128,14 @@ func pgErrCode(err error) string {
 	return ""
 }
 
+func pgErrConstraint(err error) string {
+	var pgErr *pgconn.PgError
+	if errors.As(err, &pgErr) {
+		return pgErr.ConstraintName
+	}
+	return ""
+}
+
 // --- Atespaces ---
 
 func (p *Persistence) CreateAtespace(ctx context.Context, atespace *ateapipb.Atespace) (*ateapipb.Atespace, error) {
@@ -183,6 +191,10 @@ func (p *Persistence) AtespaceExists(ctx context.Context, name string) (bool, er
 }
 
 func (p *Persistence) ListAtespaces(ctx context.Context, opts store.ListOptions) (store.ListResponse[*ateapipb.Atespace], error) {
+	opts, err := store.NormalizeListOptions(opts)
+	if err != nil {
+		return store.ListResponse[*ateapipb.Atespace]{}, err
+	}
 	pageSize, pageTokenStr := opts.PageSize, opts.PageToken
 	token, err := decodePageToken(pageTokenStr, kindAtespace, "", 1)
 	if err != nil {
@@ -367,6 +379,10 @@ func (p *Persistence) UpdateActorTemplate(ctx context.Context, templateRef resou
 }
 
 func (p *Persistence) ListActorTemplates(ctx context.Context, atespace string, opts store.ListOptions) (store.ListResponse[*ateapipb.ActorTemplate], error) {
+	opts, err := store.NormalizeListOptions(opts)
+	if err != nil {
+		return store.ListResponse[*ateapipb.ActorTemplate]{}, err
+	}
 	pageSize, pageTokenStr := opts.PageSize, opts.PageToken
 	keyParts := 2
 	if atespace != "" {
@@ -633,9 +649,12 @@ func (p *Persistence) DeleteActor(ctx context.Context, actorRef resources.ActorR
 }
 
 func (p *Persistence) ListActors(ctx context.Context, atespace string, opts store.ListOptions) (store.ListResponse[*ateapipb.Actor], error) {
+	opts, err := store.NormalizeListOptions(opts)
+	if err != nil {
+		return store.ListResponse[*ateapipb.Actor]{}, err
+	}
 	var items []*ateapipb.Actor
 	var nextToken string
-	var err error
 	if atespace != "" {
 		items, nextToken, err = p.listActorsScoped(ctx, atespace, opts.PageSize, opts.PageToken)
 	} else {
@@ -806,9 +825,12 @@ func (p *Persistence) GetActorSnapshotTag(ctx context.Context, atespace, name st
 }
 
 func (p *Persistence) ListActorSnapshots(ctx context.Context, atespace string, opts store.ListOptions) (store.ListResponse[*ateapipb.ActorSnapshot], error) {
+	opts, err := store.NormalizeListOptions(opts)
+	if err != nil {
+		return store.ListResponse[*ateapipb.ActorSnapshot]{}, err
+	}
 	var items []*ateapipb.ActorSnapshot
 	var nextToken string
-	var err error
 	if atespace != "" {
 		items, nextToken, err = p.listActorSnapshotsScoped(ctx, atespace, opts.PageSize, opts.PageToken)
 	} else {
@@ -928,10 +950,6 @@ func (p *Persistence) CreateActorSnapshotTag(ctx context.Context, snapshotAtespa
 		return nil, fmt.Errorf("beginning actor snapshot tag create: %w", err)
 	}
 	defer tx.Rollback(ctx) //nolint:errcheck // no-op once committed
-	if _, err := getActorSnapshotRow(ctx, tx, snapshotAtespace, snapshotName); err != nil {
-		return nil, err
-	}
-
 	var inserted []byte
 	err = tx.QueryRow(ctx, `
 		INSERT INTO actor_snapshot_tags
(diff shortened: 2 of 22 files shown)


cmd/ateapi/internal/store/atepg/schema.go
```diff
@@ -74,18 +77,24 @@ CREATE TABLE IF NOT EXISTS actor_snapshots (
 );
 
 CREATE TABLE IF NOT EXISTS actor_snapshot_tags (
-    atespace           text NOT NULL
-        REFERENCES atespaces(name) ON DELETE RESTRICT,
+    atespace           text NOT NULL,
     name               text NOT NULL,
     snapshot_atespace  text NOT NULL,
     snapshot_name      text NOT NULL,
+    uid                text NOT NULL,
```
jpbetz: This is the one that worries me. CREATE TABLE IF NOT EXISTS skips tables that already exist, so a database created before this PR never gets these changes. No uid column here, no new FKs, no named constraints, and the dropped UNIQUEs stay behind. Only the two new CREATE INDEX statements would apply. On such a database every tag create and update fails with a column does not exist error, and nothing notices the mismatch at startup.

If we can say no durable deployment is running the postgres backend yet, that is worth stating in the PR description. Either way I think this shows we need schema versioning and a migration path before the next change like this.

cmd/ateapi/internal/store/atepg/schema.go
```diff
@@ -58,9 +58,12 @@ CREATE TABLE IF NOT EXISTS actor_template_versions (
...
     actor_template_atespace   text NOT NULL,
     actor_template_name       text NOT NULL,
-    uid                       text NOT NULL UNIQUE,
+    uid                       text NOT NULL,
     proto                     bytea NOT NULL,
-    PRIMARY KEY (atespace, name)
+    PRIMARY KEY (atespace, name),
+    CONSTRAINT actor_template_versions_parent_fk
+        FOREIGN KEY (actor_template_atespace, actor_template_name)
+        REFERENCES actor_templates(atespace, name) ON DELETE RESTRICT
```
jpbetz: Glad to see this FK. One small follow-on. DeleteActorTemplate still uses its NOT EXISTS check and doesn't map FK errors, so I think this sequence is now possible.

1. DeleteActorTemplate runs. Its NOT EXISTS snapshot sees no versions, so the DELETE proceeds.
2. A concurrent CreateActorTemplateVersion for the same template commits first.
3. The DELETE trips actor_template_versions_parent_fk.
4. That error isn't mapped, so the client gets Internal instead of FailedPrecondition.

Rare, and the database now protects the data either way. Mapping the FK violation to ErrFailedPrecondition in DeleteActorTemplate would close it. Fine as a follow-up.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1100,26 +1141,30 @@ func (p *Persistence) CreateActorSnapshotTag(ctx context.Context, snapshotAtespa
...
 	if err == nil {
 		if err := tx.Commit(ctx); err != nil {
 			return nil, fmt.Errorf("committing actor snapshot tag create: %w", err)
 		}
 		return dbTag, nil
 	}
 	if isForeignKeyViolation(err) {
-		return nil, store.ErrFailedPrecondition
+		switch pgErrConstraint(err) {
+		case "actor_snapshot_tags_snapshot_fk":
```
jpbetz: Matching on constraint names is a nice way to split NotFound from FailedPrecondition. Same caveat as my schema comment though. On a database created before this PR the constraints still have the old auto generated names, so this switch falls to the default and a missing snapshot surfaces as Internal.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -105,6 +105,20 @@ func newUpdateMetadata(current *ateapipb.ResourceMetadata) *ateapipb.ResourceMet
 	return metadata
 }
 
+// updateMaxAttempts bounds how many times a read-modify-write is retried after
+// its optimistic uid/version check loses to a concurrent writer.
+const updateMaxAttempts = 5
```
jpbetz: nit: the five attempts run back to back with no backoff, so under a genuinely contended row they can all lose within a few milliseconds. A short sleep or a little jitter between attempts would make the budget do more work. Fine as a follow-up.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1518,6 +1567,9 @@ const defaultLockTTL = 30 * time.Second
 func (p *Persistence) AcquireLock(ctx context.Context, key string) (*store.Lock, error) {
 	ttl := p.lockTTL
 	token := uuid.NewString()
+	if err := p.cleanupExpiredLeases(ctx); err != nil {
```
jpbetz, line 1365: Recommend opening an issue to move this to a dedicated cleanup task for the long term. It's not super urgent, but it's worth tracking and circling back on.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -83,6 +83,18 @@ func (w *ActorWorkflow) ResumeActor(ctx context.Context, actorRef resources.Acto
...
+	// Routed requests call ResumeActor even when the actor is already running.
+	// Read before taking the distributed lease so that hot-path checks do not
+	// upsert and delete a PostgreSQL lease row. Any state that needs work is read
+	// again under the lock below.
+	actor, err = w.store.GetActor(ctx, actorRef)
+	if err != nil {
+		return nil, false, err
+	}
+	if wasRunning = actor.GetStatus() == ateapipb.Actor_STATUS_RUNNING; wasRunning {
+		return actor, false, nil
```
jpbetz: I like this fast path, it takes the lease table out of routed requests entirely, and the authoritative re-check still happens under the lock.

One question though. The e2e failure on this PR doesn't look like the usual flakes. TestGracefulWorkerTerminationSuspend fails on both the cert and token legs with the same error, the actor is already CRASHED when the test tries to suspend it during graceful termination. Main's recent e2e failures are on different tests. I haven't pinned it to a line, but the resume changes in this file are the nearest suspects. Have you been able to reproduce it locally?

cmd/ateapi/internal/store/ateredis/ateredis.go
```diff
@@ -487,6 +487,14 @@ func (s *Persistence) DeleteActorTemplate(ctx context.Context, templateRef resou
 }
 
 func (s *Persistence) CreateActorTemplateVersion(ctx context.Context, atv *ateapipb.ActorTemplateVersion) (*ateapipb.ActorTemplateVersion, error) {
+	parent := resources.ActorTemplateRefFromObjectRef(atv.GetActorTemplate())
+	exists, err := s.ActorTemplateExists(ctx, parent)
+	if err != nil {
+		return nil, fmt.Errorf("while checking actor template parent %s: %w", parent, err)
+	}
+	if !exists {
+		return nil, store.ErrFailedPrecondition
```
jpbetz: Consider adding a comment that is best-effort but orphans here are possible.

cmd/ateapi/internal/store/atepg/schema.go
```diff
@@ -36,7 +36,7 @@ CREATE TABLE IF NOT EXISTS actors (
     atespace  text NOT NULL
         REFERENCES atespaces(name) ON DELETE RESTRICT,
     name      text NOT NULL,
-    uid       text NOT NULL UNIQUE,
+    uid       text NOT NULL,
```
thockin, line 41: Should we include UID in *every* resource table?  Not everything is updateable now, but that tends to not last, and we want consistent precondition semantics for every resource.  This way we do not need to back-populate them when we inevitably need them later.

I see atespaces and actor_snapshots do not have UID

cmd/ateapi/internal/store/atepg/atepg.go
thockin: Why do we retry at all if there are preconditions specified on input?  If I asked for version==6 and I find DB has version 7, no retries will ever fix it.  If we support non-precondition updates, this could matter.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -105,6 +105,20 @@ func newUpdateMetadata(current *ateapipb.ResourceMetadata) *ateapipb.ResourceMet
 	return metadata
 }
 
+// updateMaxAttempts bounds how many times a read-modify-write is retried after
+// its optimistic uid/version check loses to a concurrent writer.
+const updateMaxAttempts = 5
+
+func validateMetadataProjection(resource string, metadata *ateapipb.ResourceMetadata, uid string, version int64) error {
```
thockin: Is Projection a term I don't know or did you mean Precondition?

thockin (commented):
I skimmed - this is not my expertise area, but a few thoughts

jpbetz (commented):
Thanks for getting this all addressed, this all looks like it's moving in the right direction to me. A few minor comments and I saw @thockin had some comments, after that this looks ready to me.

cmd/ateapi/internal/store/atepg/schema.go
supreme-gg-gg: `actor_snapshots` should be immutable, so it should not need UID for update? I can add version and UID to atespaces in case it will become updatable

cmd/ateapi/internal/store/atepg/atepg.go
supreme-gg-gg: The current update interface does support non-precondition updates, so the retry loop intends to support unpinned mutations which wish to reapply the change to the latest stored value, there is also a test in `atepg_test.go` that exhibits this behaviour. You're right about version-pinned updates, they will return with `ErrVersionConflict` after a failed CAS.

I just double checked that all ateapi callers are pinned using `WithPrecondition`, so we might want to remove the retry loop and make the update interface always pinned. However, this seems to me to be partially reversing #763 which replaced `expectedVersion` with a transactional `mutate` closure.

@juli4n is the goal of #763 to support retryable, unpinned read-modify-writes?

jpbetz: If we end up keeping the retry loop, recommend adding a small amount of jittered backoff. Without it this is a relatively hot loop that expires quickly. More bang for the buck if you spread the requests over a bit more time.

cmd/ateapi/internal/store/atepg/atepg.go
supreme-gg-gg: Here "projection" refers to the version/uid columns "projected" from the stored proto, which is different from precondition which refers to caller provided expectations. I agree it's a bit unfamiliar, I'll change it to something like `validateProtoMetadataMatchesColumns`, which is more literal.

cmd/ateapi/internal/store/atepg/atepg.go
supreme-gg-gg: Opened #1040 to track

cmd/ateapi/internal/store/atepg/schema.go
thockin: Immutability tends to erode, unless there's a clear reason not to we should probably just be consistent?

cmd/ateapi/internal/store/atepg/atepg.go
thockin: Julian's proposal for update is:
* Whole-object updates, require preconditions now
* Maybe whole-object updates without preconditions later
* Probably patch-updates without preconditions later

I'm fine to keep the loop now as long as we only loop when there are no preconditions in the user input.  Note that (I think?) a patch is still:

```
user calls patch with no preconditions
loop
  read old value
  call mutate()
  submit to storage with preconditions
  if success; break loop
```

?

cmd/ateapi/internal/store/atepg/atepg.go
thockin: Thanks!  A comment would suffice.

cmd/ateapi/internal/store/atepg/atepg.go
supreme-gg-gg: I think we should keep the loop then to support future updates without user preconditions. Currently, a preconditioned update that loses the CAS will exit the loop on the next iteration when it rereads the resource and `WithPrecondition` returns conflict, so we're effectively only looping for non-preconditioned updates

EDIT: regarding jittered backoff, since all callers are with precondition right now, I worry they will cause unnecessary latency on conflict. I suggest coming back to it when we start supporting future non-preconditioned patch updates, and perhaps alongside a storage interface change that allows us to identify retryable versus pinned updates (to skip the loop).

juli4n: Considering that we are making preconditions required (this is in-progress, see https://github.com/agent-substrate/substrate/issues/954), I would probably simplify the code and remove the loop? I don't see much value in it as it will be extra complexity / dead code.

juli4n: Considering that we are making preconditions required for updates (see https://github.com/agent-substrate/substrate/issues/954), I don't see much value in keeping the retry loop.

jpbetz: > Currently, a preconditioned update that loses the CAS will exit the loop on the next iteration when it rereads the resource and `WithPrecondition` returns conflict, so we're effectively only looping for non-preconditioned updates

That's not what we want. My interpretation of @thockin's pseudo-code is that after each "read old value" we also update the precondition to be predicated off of that read.  That way the mutation is still safely applied in the presence of conflicts.

I also think it's critical that we only auto-retry writes is if we get a clear response from the server that the write failed due to a precondition.  If we retry in any other situation it's possible we could accidentally double-write.

cmd/ateapi/internal/store/atepg/atepg.go
juli4n: Why a separate method? Just merge into `GetAtespace`?

juli4n: Same comment here, inline into `GetActorTemplate`?

cmd/ateapi/internal/store/atepg/atepg.go
supreme-gg-gg: I've removed the retry loop given #954

cmd/ateapi/internal/store/atepg/atepg.go
supreme-gg-gg: Merged them, some of these helpers are used elsewhere before this PR and I forgot to remove them

juli4n: Why is this needed? Should this block be just `return nil, store.ErrNotFound`?

juli4n, line 502: Not in this PR, but I think we should have a dedicated sentinel error for "atespace doesn't exist" so we can document the contract of `Create<Resource>` methods clearly, and clients (i.e. RPC handlers) can properly map that to RPC statuses / error messages.

juli4n: Inline into `GetActor`? That's the only caller, and it's a single line method.

juli4n, line 604: Not for this PR: Do we really need transactions here? IIUC this wants to enforce that you only delete if the status is DELETING, but once we have preconditions for delete this can be just enforced by the caller, which will turn this into a single `DELETE ... WHERE`.

juli4n: Inline into `GetActorSnapshot`?

supreme-gg-gg: Not needed, I'll remove it, I realized it's a left over from removing `ActorTemplateVersion`

supreme-gg-gg: I agree, I'll do this probably after we have #940 merged, since this would be difficult to enforce in Redis and we're getting rid of it anyways

supreme-gg-gg: Yes we can remove the transaction and the pessimistic locking once we have preconditions for delete

supreme-gg-gg:
@juli4n sorry for pinging but could you take another look at this PR, thanks!

juli4n (approved):

PR #1025, opened 2026-08-17
Add `status` fields to Substrate resources

### LINKED ISSUE #1006: Encapsulate all output-only fields into a top-level `status` field
Currently, output-only (server-managed) and user-controlled fields are mixed together at the top level (e.g., within `Actor` resources). 

This creates friction during update operations:
- Updating user-controlled fields requires extra care to preserve existing server-managed fields.
- It complicates lifecycle management in scenarios where a server-owned field needs to outlive or be decoupled from user-controlled state.

### Proposal

Separate user-controlled state from server-managed state by encapsulating all output-only fields into a dedicated top-level `status` field. 

This aligns with standard Kubernetes-style resource conventions (e.g., `spec` vs. `status`), providing:
- Clear ownership boundaries between client intent and server state.
- Cleaner write/update semantics without accidental overwrites of output-only metadata.
- A familiar mental model for users already accustomed to declarative APIs.

Looking at the resources we have today, almost all output only fields are top level, so the change is mechanical, there are a few exemptions:

* `ResourceMetadata`: Mixes user controlled (atespace and name) and output only (UID, version, timestamps): I think we can probably keep this as is and special case. ResourceMetadata is already special during updates as UID and version are used for optimistic locking, atespace and ate to identify the resource to update, etc.
* `Actor.source_snapshot`: has `tag` which is user controlled and two output-only fields. We could just split them, making `tag` a top level field and moving the other two under `Actor.status.source_snapshot`.

juli4n:
Add `status` fields to all Substrate resources that need it. Move server-owned fields under it. 

Fixes #1006 .


thockin:
Nevermind, I see.

~~What is "partial" about this?~~

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -208,52 +208,73 @@ message Actor {
...
+  // find eligible pools.
+  // Changes take effect on the next ResumeActor call.
+  Selector worker_selector = 5;
+
+  // The tag specified by the caller at CreateActor to seed this Actor from an
+  // ActorSnapshot. Unset if the Actor was not created from a snapshot. Set
+  // once at creation and immutable afterward.
+  ObjectRef source_snapshot_tag = 6;
+
+  ActorStatus status = 7;
```
thockin, line 205: Should we give status a high tag number (e.g. 1024)?  Presumably we are going to add more fields in these resources.  ISTR the tags are encoded as varint, so that has some downside...

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -208,52 +208,73 @@ message Actor {
...
+}
+
+message ActorStatus {
+  enum State {
```
thockin: Aside: the resulting Go names for these are awful - WDYT about moving this to a file-level `ActorState` ?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -278,16 +299,19 @@ message WorkerAssignment {
...
   ResourceMetadata metadata = 1;
-  ObjectRef source_actor = 2;
-  string source_actor_uid = 3;
-  int64 source_actor_version = 4;
-  string actor_template_namespace = 5;
-  string actor_template_name = 6;
-  string actor_template_uid = 7;
-  SnapshotContentScope content_scope = 8;
-  string snapshot_uri = 9;
-  // Immutable reference to the actor_template_version where the snapshot was created from.
```
thockin: You lost the only comment in this whole section

docs/architecture.md
```diff
@@ -436,7 +436,7 @@ Triggered by an inbound request at the Gateway or an explicit API call.
 
   3. **Hydration**: The `atelet` supervisor coordinates with the `ateom` process inside the worker pod to restore the ActorTemplate's golden `ActorSnapshot` (for first-run) or the Actor's latest `ActorSnapshot` (for recurring runs) into the sandbox.
 
-  4. **Status**: Status transitions to `STATUS_RUNNING`. The actor now has an
+  4. **Status**: Status transitions to `STATE_RUNNING`. The actor now has an
```
thockin: State?

internal/e2e/suites/demo/demo_test.go
```diff
@@ -701,8 +701,8 @@ func createActor(ctx context.Context, t *testing.T, clients *e2e.Clients, nsObj
 	if actor.GetActorTemplateName() != at.Name {
 		t.Errorf("expected actor template name %s, got %s", at.Name, actor.GetActorTemplateName())
 	}
-	if actor.Status != ateapipb.Actor_STATUS_SUSPENDED {
-		t.Errorf("expected actor status to be SUSPENDED, got %v", actor.Status)
+	if actor.Status.State != ateapipb.ActorStatus_STATE_SUSPENDED {
+		t.Errorf("expected actor status to be SUSPENDED, got %v", actor.Status.State)
```
thockin: state

I found a few of these, I am sure it's not all of them.  Maybe `git grep -i 'actor.*status'` and see what hits?

docs/architecture.md
```diff
@@ -453,7 +453,7 @@ Triggered by an explicit `SuspendActor` call.
 
   3. **Reclaim**: The physical worker is wiped and returned to the `WorkerPool`.
 
-  4. **Status**: Status transitions back to `STATUS_SUSPENDED`, now pointing to
+  4. **Status**: Status transitions back to `STATE_SUSPENDED`, now pointing to
```
thockin: State

demos/claude-code-multiplex/ui/server.go
```diff
@@ -198,15 +198,15 @@ func newKubeClient() (*kubernetes.Clientset, error) {
 // actorStatusString maps the proto Status enum to the human-readable
 // phase string the UI's badge logic understands (running / suspended
 // / etc).
-func actorStatusString(s ateapipb.Actor_Status) string {
+func actorStatusString(s ateapipb.ActorStatus_State) string {
```
thockin: actorStateString

thockin (commented):
LGTM overall - as you said "mechanical"

pkg/proto/ateapipb/ateapi.proto
juli4n: Done

pkg/proto/ateapipb/ateapi.proto
juli4n: Added it back. It was intentional, all comments in ateapi.proto are pretty bad and I'm thinking about a PR dedicated just to documenting the API but I don't think it hurts.

docs/architecture.md
juli4n: Done.

internal/e2e/suites/demo/demo_test.go
juli4n: Done. Thanks for the command hint.

docs/architecture.md
juli4n: Done.

demos/claude-code-multiplex/ui/server.go
juli4n: Done

juli4n:
FYI: I've added one more commit for ActorTemplate as @zoez7 changes are merged already.

pkg/proto/ateapipb/ateapi.proto
juli4n: Is there any value in doing this? It's a single field. I've seen this being done in oneof fields so you can carve out a range for the oneof fields, but not sure about what value would that bring here.

zoez7 (approved):

pkg/proto/ateapipb/ateapi.proto
thockin: As we add more fields, we should put them in the source in logical grouping.  This means all spec fields either before or after status.

Scanning a proto for "what's the next tag number" is annoying.

That's my whole use-case :)

OTOH, if we know that every instance will have status, maybe status should be tag number 1 (smallest binary encoding).  With a billion actors an extra byte of encoding is a gigabyte of storage.

PR #1030, opened 2026-08-17
 microvm: exclude overlay workdirs from rootfs snapshots

Oneimu:
#### microvm: Exclude Overlay Workdirs from Rootfs Snapshots

With `index=off` pinned on the merged rootfs mount, a restored workdir is inert: overlayfs wipes and rebuilds it at mount time, and staging recreates the directory regardless. 

Archiving `<cid>/work` was therefore dead weight on the suspend path—and not always trivial weight, since a copy-up in flight at pause can leave file-sized temporary data there.

---

#### Key Changes:

* **`tarutil.CreateFiltered`**: Added a skip predicate over `Create` (skipping a directory prunes its entire subtree).
* **Workdir Pruning**: Drops exactly the second-level `work` entries during snapshot archive creation.
* **Backward Compatibility**: Extraction logic remains unchanged; existing snapshots containing `workdir` entries will still restore without issue.

---

> **Follow-up to #846 review note:** With `index=off` pinned, a restored workdir is inert and overlayfs rebuilds it at mount, making its archival unnecessary overhead on the suspend path.

### DIFF
--- a/cmd/ateom-microvm/internal/tarutil/tarutil_test.go
+++ b/cmd/ateom-microvm/internal/tarutil/tarutil_test.go
@@ -239,10 +239,43 @@ func TestRoundTripFifo(t *testing.T) {
 // snapshots: a socket in the archived tree — which agents leave behind routinely
 // — must not fail the archive, because that would make the actor impossible to
 // suspend.
-// CreateFiltered must drop exactly the entries the skip function selects — for
-// a directory, the whole subtree — and leave everything else identical to an
-// unfiltered archive. The rootfs snapshot leans on this to exclude overlay
-// workdirs.
+func TestCreateSkipsSockets(t *testing.T) {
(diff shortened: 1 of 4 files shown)


cmd/ateom-microvm/internal/tarutil/tarutil_test.go
```diff
@@ -239,6 +239,50 @@ func TestRoundTripFifo(t *testing.T) {
 // snapshots: a socket in the archived tree — which agents leave behind routinely
 // — must not fail the archive, because that would make the actor impossible to
 // suspend.
+// CreateFiltered must drop exactly the entries the skip function selects — for
```
BenTheElder: This test was inserted between `TestCreateSkipsSockets`'s doc comment and the function. Please fix.

BenTheElder (approved):

PR #1034, opened 2026-08-18
microvm: serve durable-dir volumes through the one kataShared virtiofsd

Oneimu:
## What Changed and Why

Durable-dir volumes were previously served to the guest by a second per-actor `virtiofsd`. That arrangement predates the writable `kataShared` share: when the rootfs share was a read-only lower, a writable durable share had to be its own device. Since #846, the `kataShared` tree is writable and served with `--announce-submounts`, making the second daemon redundant—it cost an extra process and `vhost` socket per actor, an extra `fs` device in every snapshot config, and a restore-time revival of all three.

This PR folds the durable-dir volumes into the single existing share:

* **Subtree Bind Mounting (`_durable`):**  
  `kata.BindIntoShare` bind-mounts the `atelet`-owned volumes directory into the served tree as its `_durable` subtree. The leading underscore keeps it out of the container-ID namespace (container names are RFC 1123 labels and cannot begin with an underscore). The guest sees it as a submount of the `kataShared` mount, and containers bind their volumes from `<shared>/_durable/<volume>` exactly as they previously did from the second share.
* **Streamlined VM Configuration:**  
  Cold boot no longer spawns the durable `virtiofsd`, and the VM configuration carries exactly one `virtio-fs` device.
* **Unchanged Ownership Semantics:**  
  `atelet` still owns the directory (creates it before boot, wipes it on actor reset). The bind is `ateom`-owned mount state, detached by `CleanupSandboxState` before any removal, ensuring `atelet`'s data is never touched through it.
* **Unchanged Snapshot Content:**  
  Checkpoints tar the host directory directly under every scope, exactly as before.

> [!NOTE]  
> The bind uses the same mechanism the per-container merged rootfs mounts already use through this `virtiofsd` (submounts inside the served tree, re-opened by `find-paths` on restore), introducing no new guest-side behavior. `BindIntoShare`'s doc comment establishes this pattern for future per-actor shares: **mount a reserved subtree rather than adding a device** (relevant to in-flight work such as #803; #923 already follows this subtree approach).

---

## Compatibility with Existing Snapshots

Restore is self-describing in both directions:

* **Two-Share Era Snapshots:**  
  A snapshot whose `config.json` carries an `ateDurable` fs device originates from the two-share era. The resumed guest still expects that device, so restore revives the second `virtiofsd` exactly as before (`stageLegacyDurableShare`). Such a lineage remains two-share across its own re-checkpoints since `cloud-hypervisor` re-emits the device.
* **Single-Share Snapshots:**  
  A snapshot without the device gets the volumes re-bound into the shared tree before `virtiofsd` starts, allowing `find-paths` to re-open the guest's open durable files at their `_durable/...` paths, which the restored tar reproduces exactly.

Detection logic lives in `rewriteSnapshotSocketPaths`, which already inspects the configuration's `fs` devices. The configuration acts as the authority because the device is what `cloud-hypervisor` re-opens, regardless of the actor's spec.

---

## How This Was Tested

* **Unit Tests:**  
  * Added assertions verifying that `rewriteSnapshotSocketPaths` classifies single-device and two-device snapshot configs correctly (routing restore to the bind vs. the legacy share).
  * Added a new `kata-package` test pinning `_durable` subtree invariants: the underscore namespace reservation, the guest path residing inside the single `kataShared` mount, and host/guest agreement on the relative path re-opened by `find-paths`.
  * Verified that existing durable-volume unit tests (tar/untar round-trip, container mount construction, spec isolation) continue to pass unchanged.
* **End-to-End Verification:**  
  * Ran the counter demo on a GKE cluster with nested virtualization: verified cold boot, suspend, and resume of a durable-volume actor on this branch (volume contents survived across worker pods).
  * Verified restore of a legacy snapshot taken on `main` before this change to exercise the legacy two-share fallback path.

### DIFF
--- a/cmd/ateom-microvm/durable.go
+++ b/cmd/ateom-microvm/durable.go
@@ -41,12 +41,9 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"os/exec"
 	"path/filepath"
-	"strings"
 
 	"github.com/agent-substrate/substrate/cmd/ateom-microvm/internal/kata"
-	"github.com/agent-substrate/substrate/cmd/ateom-microvm/internal/reaper"
 	"github.com/agent-substrate/substrate/cmd/ateom-microvm/internal/tarutil"
 	"github.com/agent-substrate/substrate/internal/ateompath"
 	"github.com/agent-substrate/substrate/internal/proto/ateompb"
@@ -115,19 +112,8 @@ func (s *AteomService) stageDurableVolumes(ctx context.Context, actorUID string)
 	if _, err := os.Stat(src); err != nil {
 		return fmt.Errorf("while checking durable-dir volumes dir %q: %w", src, err)
 	}
-	dst := filepath.Join(kata.SharedDir(actorUID), "durable")
-	// Drop any stale mount first (lazy if busy), then ensure clean mountpoint.
-	if err := reaper.Run(exec.Command("umount", dst)); err != nil {
-		_ = reaper.Run(exec.Command("umount", "-l", dst))
-	}
-	if err := os.MkdirAll(dst, 0o755); err != nil {
-		return fmt.Errorf("creating %q: %w", dst, err)
-	}
-	cmd := exec.CommandContext(ctx, "mount", "--bind", src, dst)
-	var stderr strings.Builder
-	cmd.Stderr = &stderr
-	if err := reaper.Run(cmd); err != nil {
-		return fmt.Errorf("bind-mounting durable-dir volumes at %q: %w (%s)", dst, err, strings.TrimSpace(stderr.String()))
+	if err := kata.BindIntoShare(ctx, src, actorUID, "durable"); err != nil {
+		return fmt.Errorf("while binding durable-dir volumes into the shared tree: %w", err)
 	}
 	return nil
 }
--- a/cmd/ateom-microvm/internal/kata/overlay_linux.go
+++ b/cmd/ateom-microvm/internal/kata/overlay_linux.go
@@ -195,19 +195,12 @@ func StageImageVolume(ctx context.Context, src, id, cid, volumeName string) erro
 	if cid == "" || volumeName == "" {
 		return fmt.Errorf("StageImageVolume: empty container id or volume name")
 	}
+	if err := BindIntoShare(ctx, src, id, filepath.Join(cid, "volumes", volumeName)); err != nil {
+		return fmt.Errorf("while binding image volume %q into the shared tree: %w", volumeName, err)
+	}
+	// Read-only is this volume type's contract (the image is someone else's,
+	// mounted for its contents); the other subtree consumers stay writable.
 	dst := SharedVolumeDir(id, cid, volumeName)
-	if err := reaper.Run(exec.Command("umount", dst)); err != nil {
-		_ = reaper.Run(exec.Command("umount", "-l", dst))
-	}
-	if err := os.MkdirAll(dst, 0o755); err != nil {
-		return fmt.Errorf("creating shared volume dir %q: %w", dst, err)
-	}
-	cmd := exec.CommandContext(ctx, "mount", "--rbind", src, dst)
-	var stderr strings.Builder
-	cmd.Stderr = &stderr
-	if err := reaper.Run(cmd); err != nil {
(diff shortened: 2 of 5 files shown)


cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -231,30 +241,59 @@ func UnmountMergedRootfs(restoreID, cid string) {
...
-			Fstype:     typeVirtioFS,
-			MountPoint: guestDurableDir,
-		})
+// This is THE pattern for exposing another host directory to the guest: a new
+// share must otherwise pay for its own virtiofsd — a process, a vhost socket,
+// an fs device in every snapshot config and a restore-time revival of all
+// three — per actor, forever. A bind costs one mount, and teardown already
+// covers it: CleanupSandboxState lazily detaches every mount under the sandbox
+// dir before removing it, so the source directory (which may belong to atelet,
+// as the durable-dir volumes do) is never touched. Callers must stage binds
```
BenTheElder: 🤖 **should-fix** 🟡 – `CleanupSandboxState` doesn't enforce this, so a failed detach deletes the actor's durable data.

It logs each `unix.Unmount` failure and continues, then runs `os.RemoveAll(d)` unconditionally. Worse, the whole unmount loop sits inside `if b, err := os.ReadFile("/proc/self/mountinfo"); err == nil` with no else — if that read fails, nothing is unmounted and the `RemoveAll` still runs, silently.

Either way `RemoveAll` walks into a live `_durable` bind and deletes through it, and what it deletes is `ateompath.DurableDirVolumeMountsDir` — the data durable volumes exist to preserve. Nothing under `SharedDir` had that property before: the merged rootfs overlays would take their whiteouts to an ateom-owned upper that gets wiped anyway, and their lower is reconstructible from the image.

Both triggers are unlikely, which is why I'd not block on it — but the guarantee is asserted here and in `durable.go` ("nothing of atelet's is ever deleted through it") rather than enforced. Skipping the `RemoveAll` for a directory whose unmounts didn't all succeed turns silent data loss into a leaked directory, which is the right trade.

cmd/ateom-microvm/durable.go
```diff
@@ -103,22 +108,41 @@ func workloadSpec(c actorContainer) *specs.Spec {
...
 	return filepath.Join(kata.VMDir(id), "virtiofsd-durable.log")
 }
 
-// stageDurableShare starts the virtiofsd serving the actor's durable-dir volumes.
-//
-// It serves ateompath.DurableDirVolumeMountsDir directly — no bind into the
-// kataShared tree — so teardown has nothing extra to unmount. Unlike the RO
-// lower's virtiofsd this one runs with cache=auto: the host contents change
-// underneath the guest whenever a snapshot is restored into them.
+// stageLegacyDurableShare starts the second virtiofsd that snapshots from the
```
BenTheElder: 🤖 **question** 🟢 – Is this worth carrying? We don't support in-place upgrade, and the position on snapshot compatibility elsewhere this week has been that we don't care about it yet.

Unlike a stub, this one genuinely works — you've kept the device detection, the second daemon, its teardown, and the tests, and you verified a legacy restore end to end. That's the cost: a branch through `restoreFullScope`, a `legacyDurable` return threaded out of `rewriteSnapshotSocketPaths`, `DurableFsTag`/`DurableVirtiofsdSocketPath`/`durableVirtiofsdLogPath` kept alive, and a lineage that stays permanently two-share once restored.

If pre-change snapshots can just be discarded, dropping all of it lets `rewriteSnapshotSocketPaths` keep its single return value and lets `default:` reject `ateDurable` the way it already rejects the retired `ateUpper` tag.

BenTheElder: Human: Yes, don't try to maintain compatibility yet. Just clean up the legacy code.

cmd/ateom-microvm/internal/kata/overlay_linux.go
Oneimu: Fixed — removal is now gated on the unmounts: an unmount failure (or unreadable mountinfo) leaves the dir in place instead of RemoveAll  through a live bind. The stagers tolerate leftover dirs, so the degraded case is a retry, not data loss.

cmd/ateom-microvm/durable.go
Oneimu: Done: dropped entirely, and the PR is re-scoped after #840 landed (which already folded durable/CSI into the single share).

PR #1035, opened 2026-08-18
(chore): consolidate logging to ateattr

### LINKED ISSUE #886: Consolidate actor log labels onto the ate.* attribute registry
Actor logs and actor spans use two different spellings for the same values. Actorlog sets `ate.dev/*`, while `internal/ateattr` defines `ate.*` for spans and metrics, which follows proper conventions.

The actorlog convention predates the OTel work, that's is why I left it as-is in #412. But now it is now the only place in the telemetry surface that does not use the registry, and / is not a proper shape for an OTel attribute name, so the diff gets permanent as we send logs via OTLP.

We should set this direction before #853, so we are consistent.

Direction I would propose:

  - internal/ateattr is the single registry for every signal, logs included
  - Actor log records set the registry keys, plus trace and span IDs so a log line joins the request it belongs to (this is missing today)
  - Fix the asymmetry where the structured path omits actor_uid, the fallback path includes it

Open questions

  1. Can we decommission the old format? It will get only more expensive if we delay this.

cc @git286 @JeffLuoo @baizhenyu

krisztianfekete:
Actor logs used `ate.dev/actor_*` while spans and metrics use `ate.*` registry in internal/ateattr. This PR makes ateattr the single source of truth for everything telemetry-related.

- Renamed the six actor log labels onto the registry: 
  - `ate.atespace`, `ate.actor.name`, `ate.actor.uid`, `ate.template.namespace`, `ate.template.name`, `ate.actor.container.name`
- Logs join traces now. Records set `trace_id`, `span_id` and `trace_flags`, so you can go from Actor restored to the resume that caused it. Our own lines only, not an actor's stdout: one goroutine forwards a whole container stream and can't know which request produced a given line. Per line correlation comes with #853.
- Actors can't fake platform labels. They already couldn't overwrite ours, but they could invent new ones like `ate.tenant` that look platform issued downstream. Anything under `ate.` from an actor is now dropped.
- Fixed the asymmetry that was actually left: actor supplied label values weren't stringified, and one non string value makes Cloud Logging discard the labels for that whole entry.
- Note: the actor_uid bullet in the issue is stale, #841 fixed it earlier. Lifecycle records still set five labels rather than six, on purpose as they're about the actor, so no container produced them.

Fixes #886

- [x] Tests pass

### DIFF
--- a/internal/actorlog/logger.go
+++ b/internal/actorlog/logger.go
@@ -66,11 +66,18 @@ type ActorLogger struct {
 	labelsKey string
 }
 
+// The two spellings of the label group. Cloud Logging promotes the second into
+// LogEntry.labels, so that is the one to use on GCE.
+const (
+	labelsKeyPlain = "labels"
+	labelsKeyGCE   = "logging.googleapis.com/labels"
+)
+
 // NewActorLogger creates a new ActorLogger wrapping the provided destination writer.
 func NewActorLogger(w io.Writer, isOnGCE bool) *ActorLogger {
-	labelsKey := "labels"
+	labelsKey := labelsKeyPlain
 	if isOnGCE {
-		labelsKey = "logging.googleapis.com/labels"
+		labelsKey = labelsKeyGCE
 	}
 	return &ActorLogger{
 		writer:    w,
@@ -152,7 +159,7 @@ func (al *ActorLogger) WrapContainerLogs(r io.Reader, a resources.ActorAttributi
 						delete(m, k)
 					}
 				}
-				labels := sanitizeLabels(m[al.labelsKey])
+				labels := al.foldLabelGroups(m)
 				for k, v := range ateattr.ActorLogLabels(a, containerName) {
 					labels[k] = v
 				}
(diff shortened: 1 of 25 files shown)


krisztianfekete:
cc. @JeffLuoo, @git286

internal/actorlog/logger.go
```diff
@@ -134,45 +138,82 @@ func (al *ActorLogger) WrapContainerLogs(r io.Reader, actorRef resources.ActorRe
...
+						delete(m, k)
+					}
 				}
-				labels["ate.dev/actor_atespace"] = actorRef.Atespace
-				labels["ate.dev/actor_name"] = actorRef.Name
-				labels["ate.dev/actor_uid"] = actorUID
-				labels["ate.dev/actor_template_namespace"] = actorTemplateNamespace
-				labels["ate.dev/actor_template_name"] = actorTemplateName
-				labels["ate.dev/container_name"] = containerName
+				labels := sanitizeLabels(m[al.labelsKey])
```
git286: Should we sanitize labels for both "labels" and "logging.googleapis.com/labels", regardless what al.labelsKey is because nothing prevent users setting those labels, right?

krisztianfekete: Good catch, I just pushed this, thanks!

git286 (approved):

PR #1038, opened 2026-08-18
atenet/dns:  (Not Intended to Merge) manual CoreDNS rcode harness

ygao-g:
**Test signals only — not intended to merge.**

Stacked on #874, whose commit appears in this diff until it merges.

#874's unit test pins the rendered zone, but nothing in the tree exercises
CoreDNS itself, so the rcode behavior the PR is about goes unchecked. There is
no way to close that in a Go test here — neither `coredns` nor `miekg/dns` is
vendored, and no test in the repo runs a container. This is the reference
procedure instead:

    DOCKER="limactl shell docker-nested docker" hack/dns-manual-test.sh

It renders the real zone via the test binary, serves it with the pinned
`coredns/coredns:1.11.1`, and checks the rcode for A, AAAA, HTTPS, SRV, a name
in the zone that is not an actor, and a malformed name — plus an Alpine
`getent ahosts` for the musl path.

Passing `--corefile` a zone rendered from before the fix makes the run a
negative control: AAAA SERVFAILs and `getent` stops resolving, while plain
`dig A` still succeeds. That last combination is exactly the failure #874
describes.

### DIFF
--- a/cmd/atenet/internal/dns/README.md
+++ b/cmd/atenet/internal/dns/README.md
@@ -10,7 +10,6 @@ Cluster resources:
 
 * Deployment `ate-system:dns`. Label: app=dns
 * Service `ate-system:dns`.
-* ConfigMap `ate-system:dns`.
 
 These are defined in manifests/ate-install/atenet-dns.yaml.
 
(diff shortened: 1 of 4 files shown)


ygao-g:
Superseded by #874, which now carries `TestActorDNSZone` — the AAAA and out-of-pattern rcode assertions run in e2e on every PR, so the manual harness is no longer the only signal.

Closing. The script stays reachable in this PR's history if the extra qtypes it covers are ever worth adding as subtests.

PR #1042, opened 2026-08-18
atunnel: restrict the actor egress listener to IPv4

### LINKED ISSUE #943: atunnel: egress listener accepts IPv6 connections it cannot route

ygao-g:
Fixes #943

- [x] Tests pass

The actor egress listener was opened with `net.Listen("tcp", "0.0.0.0:15001")`.
Go treats every unspecified address as a wildcard, so that binds `::`
dual-stack: the listener has been accepting IPv6 connections all along, and
`TCPOriginalDestination` cannot resolve them because it reads only the IPv4
`SOL_IP`/`SO_ORIGINAL_DST`. Such a connection is accepted and then fails.
Egress now opens through `atunnel.ListenEgressIPv4`, where the network string
decides the family rather than an address that never did; the family is in the
name so the constraint is visible at the call site while the IPv6 work is in
flight.

This does not hold IPv6-only clusters back: the actor leg is IPv4 link-local by
construction (`ateomnet.ActorVethIP`, IPv4-family nftables), so nothing that
works today changes. Widening egress in earnest needs actor networking to gain
IPv6 (#945) and an `IP6T_SO_ORIGINAL_DST` lookup (#686). The ingress flag
default is spelled `:443` to record that it is deliberately dual-stack, which is
a comment on existing behavior rather than a change to it.

### DIFF
--- a/cmd/ateom-gvisor/main.go
+++ b/cmd/ateom-gvisor/main.go
@@ -65,8 +65,10 @@ var (
 	podUID = pflag.String("pod-uid", "", "The UID of the current pod")
 
 	// TODO(liorlieberman) have a sub package for all atunnel releated things like that
-	atunnelListenAddress        = pflag.String("atunnel-listen-address", "0.0.0.0:443", "Address for actor ingress HTTPS")
-	atunnelConnectListenAddress = pflag.String("atunnel-connect-listen-address", "0.0.0.0:444", "Address for actor ingress mTLS CONNECT")
+	//
+	// Ingress is dual-stack; egress is pinned to IPv4 by atunnel.ListenEgressIPv4.
+	atunnelListenAddress        = pflag.String("atunnel-listen-address", ":443", "Address for actor ingress HTTPS")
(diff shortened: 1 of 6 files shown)


ygao-g:
Superseded. The egress pin here was overtaken by #753 and #979, which make
actor egress work over IPv6 rather than keeping the listener off it. The
other half of this PR, the ateom ingress listen defaults, ships on its own
in #1080.

PR #1043, opened 2026-08-18
[do not merge] hack: local verification script for IPv6-only DNS

ygao-g:
**Not a merge candidate — please don't merge this.** It exists so anyone looking at #958 can reproduce that PR's local testing without hand-building a harness. Stacked on #958: the first commit is that PR's, only the second is new here.

`hack/verify-ipv6-dns.sh` checks the CoreDNS rewrite that IPv6-only kind clusters need. The offline half runs the transform against a pinned copy of kind's default Corefile and needs neither Docker nor a cluster. The live half detects the cluster's address family, asserts ipv4 and dual-stack are left alone, and probes from a pod rather than the node — the node is dual-stack and answers either way, so a node-side probe proves nothing.

```
hack/verify-ipv6-dns.sh --offline          # ~1s, no Docker, no cluster
IP_FAMILY=ipv6 hack/create-kind-cluster.sh
hack/verify-ipv6-dns.sh                    # offline checks, then live ones
```

Two of the offline checks are controls that must fail, so a transform that silently did nothing cannot pass. Locally: 9/9 offline, 11/11 against an IPv4 cluster, 14/14 against a fresh IPv6-only one.

To make the transform testable without a cluster it moved into a sourceable function; `create-kind-cluster.sh` behaviour is unchanged.

### DIFF
--- a/hack/create-kind-cluster.sh
+++ b/hack/create-kind-cluster.sh
@@ -21,6 +21,7 @@ KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
 KUBECTL_CONTEXT="kind-${KIND_CLUSTER_NAME}"
 reg_name="kind-registry"
 reg_port="${KIND_REGISTRY_PORT:-5001}"
+IPV6_DNS_UPSTREAM="${IPV6_DNS_UPSTREAM:-2001:4860:4860::8888 2001:4860:4860::8844}"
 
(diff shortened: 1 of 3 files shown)


ygao-g:
Subsumed by #958, which now carries `hack/verify-ipv6-dns.sh` itself. What is left here is the sourceable-function refactor; if that is worth having it belongs as a commit on #958 rather than a standing branch.

PR #1014, opened 2026-08-17
ateapi - Require preconditions in Update methods 

laoj2:
https://github.com/agent-substrate/substrate/issues/954

> It's a good idea to open an issue first for discussion.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

PR #1052, opened 2026-08-19
ateapi: authorize MintCert against the store, not the worker cache

### LINKED ISSUE #964: ateapi: MintCert flakes on stale worker-cache assignments

orangeCatDeveloper:
Fixes #964

Resume can deny `MintCert` after the worker assignment is committed, so atelet fails `CallAteletRestore` and the resume is lost.

`authorizeActor` resolves the worker through the watch-fed worker cache, which lags the store by the pub/sub round trip; authorization already reads the actor from the store, so only the worker lookup could observe the stale state.

## Fix

Read the worker from the store too (`s.store.GetWorker`), the same authoritative source the actor already comes from. Credential minting must see committed assignment state; cache retries or a version barrier would only shrink the window. This adds one point read per `MintCert` and removes the `workercache` dependency from `actoridentity`. The cache keeps serving scheduling and worker-count paths, where eventual consistency is fine.

Since #1000 workers are global resources named by pod UID, so the request already carries the exact key the store needs; the earlier revision of this PR re-keyed workers by `(namespace, pod)` and hardened the syncer around that, all of which #1000 and #1014 now cover on main.

## Evidence

Before, in [run 31628573958, attempt 1](https://github.com/agent-substrate/substrate/actions/runs/31628573958/attempts/1):

```
18:50:06.844788  UpdateWorker commits the assignment
18:50:06.856480  ActorIdentity denies: worker has no actor assignment
18:50:06.858441  ResumeActor fails at CallAteletRestore

--- PASS: TestMintCertAuthorization (0.13s)
--- PASS: TestMintCertUsesAnAssignmentAsSoonAsItIsWritten (0.01s)
ok  	github.com/agent-substrate/substrate/cmd/ateapi/internal/actoridentity	1.591s
```

The new test seeds an unassigned worker, writes the assignment directly to the store, and mints immediately; it is the 11.7 ms window from the log with no cache in between.

- [x] Tests pass

Supersedes #965, which was closed by an accidental force-push to its branch and cannot be reopened.

### DIFF
--- a/cmd/ateapi/internal/actoridentity/actoridentity.go
+++ b/cmd/ateapi/internal/actoridentity/actoridentity.go
@@ -29,7 +29,6 @@ import (
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/actoridjwt"
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
-	"github.com/agent-substrate/substrate/cmd/ateapi/internal/workercache"
 	"github.com/agent-substrate/substrate/internal/localca"
 	"github.com/agent-substrate/substrate/internal/localjwtauthority"
 	"github.com/agent-substrate/substrate/internal/principal"
@@ -55,19 +54,17 @@ type Server struct {
 
 	// store is the actor database. MintCert consults it to confirm the caller
 	// is entitled to the actor it is asking for a credential for.
-	store   store.Interface
-	workers *workercache.Cache
+	store store.Interface
 }
 
 var _ ateapipb.ActorIdentityServer = (*Server)(nil)
 
-func New(actorIdentityJWTIssuer, actorIDJWTPoolFile, actorIDCAPoolFile string, store store.Interface, workers *workercache.Cache) *Server {
+func New(actorIdentityJWTIssuer, actorIDJWTPoolFile, actorIDCAPoolFile string, store store.Interface) *Server {
 	return &Server{
 		actorIdentityJWTIssuer: actorIdentityJWTIssuer,
 		actorIDJWTPoolFile:     actorIDJWTPoolFile,
 		actorIDCAPoolFile:      actorIDCAPoolFile,
 		store:                  store,
-		workers:                workers,
 	}
 }
 
@@ -326,7 +323,7 @@ func (s *Server) authorizeActor(ctx context.Context, caller *ateletCaller, req *
 		return status.Errorf(codes.PermissionDenied, "caller is not permitted to mint credentials for this actor")
 	}
 
-	worker, err := s.workers.Worker(req.GetWorker().GetName())
+	worker, err := s.store.GetWorker(ctx, req.GetWorker().GetName())
 	if err != nil {
 		if errors.Is(err, store.ErrNotFound) {
 			return nil, resources.ActorRef{}, deny("worker not found")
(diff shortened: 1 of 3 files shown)


orangeCatDeveloper:
https://github.com/agent-substrate/substrate/pull/965 was closed by my agent by mistake and I couldn't reopen it due to a github dead lock. Create a new PR here. @ahmedtd @juli4n @BenTheElder

thompsonmax (approved):

orangeCatDeveloper:
Superseded by #934, which closes the same window from the other side: the worker
is still resolved from the cache, but a denial is re-checked against the store
before it stands, so an assignment committed moments earlier is seen.

Closing in favour of that.

PR #1056, opened 2026-08-19
CI e2e: run all suites for microvm

BenTheElder:
We were only running a subset of tests. That was meant to be temporary.

Fix the tests and enable the full suite for uVM.

Consolidate on a single env `E2E_SANDBOX_CLASS` as input to select microvm vs gvisor.

NOTE: we still skip identity. That will be replaced by #268


BenTheElder:
FYI @thompsonmax, I think you hit this.

HavenXia (approved):

PR #1057, opened 2026-08-19
[WIP] ateomnet: give the actor an IPv6 address when the pod has one

ygao-g:
Part of #246. **Do not merge before #979 and #753.**

Actor networking is IPv4-only, so an actor on a dual-stack worker pod cannot
reach an IPv6-only destination at all. `SetupActorNetwork` now assigns the
`fd00:169:254::/126` counterparts of the existing point-to-point pair to both
ends of the actor veth, installs an IPv6 default route in the interior netns,
and adds the matching rules to the inet-family actor table.

Whether the actor gets IPv6 is decided once in the worker pod netns and carried
into the interior one, which is created fresh and so always reports IPv6
available whatever the cluster's families are. Both halves have to hold: the pod
needs a global IPv6 address of its own, and the veth has to accept an IPv6
address — IPv4-only GKE sets `disable_ipv6` and netlink then rejects the
assignment with EPERM.

## Stacked on #1116

The first commit here is #1116, which moves the actor nftables table from the
`ip` family to `inet` with no behaviour change on IPv4. Review it there; GitHub
cannot target a base branch that lives on a fork, so it shows up in this diff
too. **The second and third commits are new in this PR.**

**Update:** a third commit makes `SetupActorNetwork` return the family decision
as an `ActorNetwork` value instead of reporting it only by side effect on the
interior netns. gVisor can read that back (runsc adopts the namespace
wholesale); the micro-VM path cannot — its guest has to be told over the
kata-agent channel — so the upcoming micro-VM parity stack consumes this
return. Also names the /126 the pair lives on (`ActorVethIPv6Subnet`) for the
guest's on-link route. Every current call site still ignores the value, so
there is no behavior change in this PR.

## Dependencies

- **#979** — without `net.ipv6.conf.all.forwarding=1` in the pod netns, the new
  default route is a black hole. Measured: 100% packet loss in a netns replica
  at `all.forwarding=0`, 0% at `=1`.
- **#753** — `atunnel` recovers the destination with
  `getsockopt(SOL_IP, SO_ORIGINAL_DST)` into a 16-byte `RawSockaddrInet4`, so an
  IPv6 connection REDIRECTed into the tunnel is accepted and then dies with no
  usable CONNECT authority. That function's own TODO asks for the IPv6 variant
  "when actor veth setup gains dual-stack support" — which is this PR.

Even with both, dual-stack egress is red: #1089, caught by the test in #1104.
This lands the address, not a working dual-stack egress path.

## Testing

**Green CI here means the family gate stays shut on an IPv4 cluster, not that
the IPv6 datapath works.** There is no dual-stack CI lane, so none of the IPv6
behaviour below is exercised there.

`TestSetupActorNetworkIPv6Gate` is the truth table, root-gated, five cases.
Reverting either half of the gate fails exactly one case and nothing else, and
reverting `IFA_F_NODAD` fails exactly the two that expect IPv6.

`TestActorNftablesRuleExprs` pins the IPv6 rule expressions alongside the IPv4
ones, and needs no root. `TestSetupActorNetworkInstallsEgressRedirect` checks
the kernel takes both redirects in one `inet` nat chain; the IPv6 rule is
installed whatever the gate decided, and simply matches nothing when the actor
has no IPv6 source address.

## Not covered

- The interior `eth0` is not observable under gVisor, so only the netns-level
  setup is asserted.
- Router → worker is still IPv4.
- The micro-VM guest is still configured IPv4-only, so a micro-VM actor gets no
  IPv6 even where the host veth has one. There is a `TODO(#246)` at that call
  site.

### DIFF
--- a/cmd/ateom-microvm/run.go
+++ b/cmd/ateom-microvm/run.go
@@ -1061,6 +1061,10 @@ func tailString(s string, n int) string {
 // agent: configure eth0 (IP/MAC/MTU), install the connected + default routes, and
 // pin the gateway's ARP entry to its fixed MAC (so a restored guest's frozen
 // neighbor entry stays valid).
+//
+// TODO(#246): the guest is configured IPv4-only, so a micro-VM actor sees no
+// IPv6 even on a dual-stack pod where the host veth has one. gVisor reads the
+// interior netns and picks the address up; this path has to be told.
 func (s *AteomService) configureGuestNetwork(ctx context.Context, ac *kata.AgentClient, mtu uint64) error {
 	if err := ac.UpdateInterface(ctx, &agentpb.Interface{
 		Device: ateomnet.ActorVethName,
--- a/cmd/ateom-microvm/run.go
+++ b/cmd/ateom-microvm/run.go
@@ -411,7 +411,7 @@ func (s *AteomService) coldBootActor(ctx context.Context, p actorBootParams) (re
 
 	// Networking (host side): per-activation veth into the interior netns. The
 	// tap + TC mirror is built below (after the VM exists) so its FDs are fresh.
-	if err := ateomnet.SetupActorNetwork(ctx, ateomnet.NetworkConfig{
+	if _, err := ateomnet.SetupActorNetwork(ctx, ateomnet.NetworkConfig{
 		InteriorNetNS:      s.interiorNetNS,
 		HostVethHWAddr:     hostVethHWAddr,
 		SweepInteriorLinks: true,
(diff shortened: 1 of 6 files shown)


cmd/ateom-microvm/run.go
```diff
@@ -1085,6 +1085,10 @@ func tailString(s string, n int) string {
 // agent: configure eth0 (IP/MAC/MTU), install the connected + default routes, and
 // pin the gateway's ARP entry to its fixed MAC (so a restored guest's frozen
 // neighbor entry stays valid).
+//
+// TODO(#246): the guest is configured IPv4-only, so a micro-VM actor sees no
+// IPv6 even on a dual-stack pod where the host veth has one. gVisor reads the
+// interior netns and picks the address up; this path has to be told.
```
BenTheElder, line 1067: Please do not merge new gVisor only features. This should be resolved prior to merge.

e2e tests should exercise both in lockstep.

PR #1059, opened 2026-08-19
ci: IPv6-only kind e2e plus the actor veth [DO NOT MERGE — signal only]

ygao-g:
**Not for merging.** Same job as #939, one layer further down the stack. #939 shows the IPv6-only e2e failing at `TestActorEgress`/`TestActorEgressHTTPS` because an actor has no IPv6 stack at all; this draft adds the actor-veth work on top so we can see what the next failure actually is, rather than predicting it. The run results are the deliverable, not the branch — nothing here should land in this shape.

#1065 supersedes this run with the full stack on a newer base.

Where each commit comes from. This branch is based on the #993 merge, so several are earlier revisions of what their source PR carries today:

| commit | subject | source |
|---|---|---|
| `09196821` | hack: fix DNS on IPv6-only kind clusters | #958, earlier revision |
| `e4ca0fcf` | hack: retry the IPv6-only kind DNS probe instead of asking once | #958, since superseded |
| `6d94c78b` | atenet/router: bind the Envoy ingress listeners dual-stack | #911, earlier revision |
| `301bf6f7` | atenet/router: bind the admin socket and Service dual-stack | #911, earlier revision |
| `f804208a` | atenet/egress: bind the Envoy sockets and Service dual-stack | #911, earlier revision |
| `c91689fa` | ateomnet: enable IPv6 forwarding in worker pod netns | #979 (@krsnaSuraj) |
| `3055db5e` | ateomnet: return nil when sysctl path missing in writeSysctlIfUnset | #979 (@krsnaSuraj) |
| `1ea34bc4` | ateomnet: give the actor veth an IPv6 address and dual-stack nftables rules | #1057 |
| `aa8e395a` | ateomnet: skip the actor veth IPv6 setup when the netns has IPv6 disabled | #1057 |
| `e5703c4f` | ci: add an IPv6-only kind e2e job | #939, earlier revision |

#877, #958, #911 and #979 all ride along because GitHub will not take a base branch that lives only in a fork; each drops out as it merges.

**A failure here is still the expected outcome, one layer lower.** The prediction to check the run against: with the veth dual-stacked, egress should get past `network is unreachable` and fail instead inside `atunnel`, which reads the redirected destination with the IPv4-only `SO_ORIGINAL_DST`. Behind that sits a third layer, `dns_lookup_family: V4_ONLY` in the egress Envoy config. Locally verified on this exact stack: builds and vets clean for `GOOS=linux`, and the five root-gated `TestSetupActorNetwork*` tests pass under `sudo` in a Linux VM, including the IPv4-only-netns case. The IPv6 CI job itself has never been run locally — `ubuntu-latest` has no IPv6 egress so the job stands up tayga/NAT64 that a local cluster does not need, which is precisely why this needs to run in CI.

Coverage and gating are unchanged from #939: 2 of 7 suites, gVisor only, `cert` auth only, single node, no CSI, and the job stays out of the `e2e-test` merge gate so it can never block a PR. Part of #246.

### DIFF
--- a/.github/workflows/e2e-ipv6.yaml
+++ b/.github/workflows/e2e-ipv6.yaml
@@ -1,0 +1,370 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: e2e-ipv6
+# Separate from pr-workflow.yaml so this can be gated independently -- and so a
+# 40-minute IPv6 run never delays that workflow's merge-gating jobs.
+#
+# TODO(246): gate this before merging. It runs on every PR today so the IPv6-only
+# results are visible without a maintainer having to act first; the intended
+# steady state is a `ci/ipv6` label, which needs the label created upstream:
+#
+#     on: {pull_request: {types: [labeled, opened, synchronize, reopened]}}
+#     if: contains(github.event.pull_request.labels.*.name, 'ci/ipv6')
+#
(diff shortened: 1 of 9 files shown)


ygao-g:
Superseded by #1065, which runs the same job on a newer base with the full stack — including #753, #1080 and #1060, which this branch predates. The run there is green; the results here are the older, partial picture.

PR #1060, opened 2026-08-19
atenet/egress: resolve upstream names on both address families

ygao-g:
Part of #246

The egress Envoy pinned `dns_lookup_family` to `V4_ONLY` in both the dynamic
forward proxy filter and the dynamic forward proxy cluster. On an IPv6-only
cluster every upstream connection failed: Envoy reported a DNS resolution
failure and the actor got a 503. Seven sites across the two egress manifests,
including the sdsmint variant, now use `ALL`, which returns both families and
enables Happy Eyeballs.

This is a prerequisite for IPv6 egress, not the fix on its own. An actor's
connection is redirected by nftables and atunnel recovers the original
destination with `getsockopt(SOL_IP, SO_ORIGINAL_DST)`, which returns ENOENT
for a v6-redirected connection — so egress fails before Envoy is ever asked to
resolve anything.

## Testing

`make verify` is clean. A new test walks the shipped manifests and requires
`ALL` on every `dns_cache_config`, so a new egress variant cannot reintroduce
the pin. It has already paid for itself: the seventh site arrived with the
sdsmint MITM leg while this was in review, still pinned to `V4_ONLY`, and the
test caught it on rebase.

Measured on an IPv6-only kind cluster with #911, #958, #979 and #753 applied.
Each value was deployed, Envoy restarted, and the live `/config_dump` checked
before running the suites:

| `dns_lookup_family` | `TestActorEgress` / `TestActorEgressHTTPS` |
| --- | --- |
| `ALL` (shipped) | pass / pass, `code=200` to `[2606:4700:10::ac42:93f3]:80` |
| `V4_ONLY` (before) | fail / fail, `code=503 flags=DF` |
| `AUTO` | pass / pass |

`AUTO` is not distinguishable from `ALL` on this path: the dynamic forward
proxy is handed the IP literal that atunnel recovered, never a hostname, so
`dns_lookup_family` only decides which literal families it accepts. `ALL` is
chosen as the value that strands neither family.

The IPv4 lane is unaffected: `TestActorEgress` and `TestActorEgressHTTPS` pass
with the change in place, across two runs of the standard e2e job.

- [x] Tests pass

### DIFF
--- a/cmd/ate-setup/internal/steps/egress_dns_test.go
+++ b/cmd/ate-setup/internal/steps/egress_dns_test.go
@@ -1,0 +1,248 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package steps
+
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+	"testing"
+
+	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/yaml"
+
+	"github.com/agent-substrate/substrate/cmd/ate-setup/internal/config"
+	"github.com/agent-substrate/substrate/cmd/ate-setup/internal/kustomize"
+)
+
+// TestEgressDNSLookupFamily requires ALL on every dynamic forward proxy DNS
+// cache in the install tree, so a new egress variant is checked the day it is
+// added rather than the day it is installed. atenet-egress.yaml says why ALL.
+func TestEgressDNSLookupFamily(t *testing.T) {
+	for _, path := range manifestPaths(t) {
+		caches := dnsCacheConfigs(t, path)
+		if len(caches) == 0 {
+			continue
+		}
+		t.Run(filepath.Base(path), func(t *testing.T) {
+			requireLookupFamilyALL(t, caches)
+		})
+	}
+}
+
+// TestRenderedOverlaysDNSLookupFamily repeats the check against the output of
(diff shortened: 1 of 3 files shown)


cmd/atenet/internal/router/extproc/egress_manifest_test.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+// The install tree, not a fixture, so this guards the Envoy config that ships.
+const manifestsDir = "../../../../../manifests"
+
+// What the install ships today. Falling below it means the walk stopped
+// matching, not that the config got better.
+const minDNSCacheConfigs = 6
+
+// TestEgressDNSLookupFamily requires ALL on every dynamic forward proxy DNS
```
ygao-g: We need to focus on the guide to verify the indices instead of why we want this inside the paper. That comment should actually be where we set it.

cmd/atenet/internal/router/extproc/egress_manifest_test.go
haiyanmeng, line 1: This test has no relationship to package extproc. We currently have https://github.com/agent-substrate/substrate/blob/main/cmd/atenet/internal/sdsmint/manifest_test.go to validate some part of `atenet-egress-with-sdsmint.yaml`. Please look into a better location for this test.

cmd/atenet/internal/router/extproc/egress_manifest_test.go
```diff
@@ -0,0 +1,136 @@
...
+			t.Fatalf("reading %s: %v", path, err)
+		}
+		var object struct {
+			Kind string            `json:"kind"`
+			Data map[string]string `json:"data"`
+		}
+		if err := yaml.Unmarshal(doc, &object); err != nil {
+			t.Fatalf("parsing a document of %s: %v", path, err)
+		}
+		if object.Kind != "ConfigMap" {
```
haiyanmeng: Will this be able to handle kustomize overlays?

cmd/atenet/internal/router/extproc/egress_manifest_test.go
ygao-g: Moved to `cmd/ate-setup/internal/steps/egress_dns_test.go`.

That package is what decides which egress manifest gets applied, so the test now goes through the installer.

cmd/atenet/internal/router/extproc/egress_manifest_test.go
ygao-g: Good catch! It does now.

`TestRenderedOverlaysDNSLookupFamily` runs every kustomization root through `kustomize.Build` and applies the same requirement to the rendered output.

haiyanmeng (approved):

Issue #922, opened 2026-08-13
atenet/dns: actor zone returns NXDOMAIN for empty non-terminals


ygao-g:
The terminal NXDOMAIN catch-all added in #874 also answers NXDOMAIN for the zone apex, `actors.resources.substrate.ate.dev`, and for `<atespace>.actors.resources.substrate.ate.dev`. Neither is an absent name: the apex exists by definition, and the zone answers for any well-formed `<actor>.<atespace>` pair, so every `<atespace>` label has children. Both are empty non-terminals, and the correct answer for one is NOERROR with an empty answer section, not NXDOMAIN.

NXDOMAIN asserts something stronger -- RFC 8020, "NXDOMAIN really means there's nothing below" -- so a resolver that caches it for `<atespace>` may decline to resolve any actor in that atespace. That is a hard resolution outage rather than a degradation, which is why it is worth tracking even though nothing hits it today.

Latent for now: it needs a resolver in front that both queries the intermediate name (QNAME minimisation, RFC 7816) and applies NXDOMAIN cuts. kube-dns `stubDomains` and the CoreDNS `forward` plugin pass the full qname through, the CoreDNS `cache` plugin caches per (qname, qtype) without aggressive negative caching, and musl and glibc do not cache at all.

The fix is one more `template` block before the catch-all, scoped by a `match` carrying the apex and single-label patterns, with `rcode NOERROR`, the same SOA authority record, and `fallthrough`. Ordering is load-bearing: after the actor NODATA block, before the catch-all. Worth deciding first whether an apex `SOA` or `NS` query should be answered properly instead, in which case `file`/`auto` with a real zone is a better shape than a fourth `template` block.

Reproduced against `coredns/coredns:1.11.1`, the pinned image: `A ns1.actors.resources.substrate.ate.dev` and `SOA actors.resources.substrate.ate.dev` both return NXDOMAIN. `corefile.go` carries a TODO pointing here. Follow-up to #874.

PR #1062, opened 2026-08-19
atenet/dns: return NODATA for non-A queries on actor zone

### LINKED ISSUE #888: atenet/dns: actor zone returns SERVFAIL for every query it does not answer
**Overview**
When resolving a hostname, modern operating systems concurrently query for both IPv4 (`A`) and IPv6 (`AAAA`) records. If a service only has an IPv4 address, the DNS server is expected to return the `A` record and explicitly return a successful but empty response (`NODATA`) for the `AAAA` query. 

Currently, our actor DNS zone does not handle this correctly. It responds to `A` queries but crashes out with a server error (`SERVFAIL`) for anything else. While tolerant resolvers (like Go's) ignore the error and proceed using the IPv4 address, strict POSIX resolvers treat the server error as a hard failure, breaking intra-cluster communication for certain actor images.

**Desired Behavior**
*   **For valid actor names (non-A queries):** The zone should return a `NODATA` response instead of a server failure. This means returning an `rcode NOERROR` with an empty answer section, alongside an SOA record in the authority section. 
*   **For invalid names:** The zone should include a terminal `NXDOMAIN` catch-all to explicitly state the name does not exist. 
*   By returning proper `NODATA` and `NXDOMAIN` responses, strict resolvers like `musl libc` will understand that no IPv6 address exists and will correctly fall back to the successfully resolved IPv4 `A` record.

krsnaSuraj:
## atenet/dns: return NODATA for non-A queries on actor zone

Fixes #888

### Problem

The actor zone only had a `template IN A` block. CoreDNS's `template` plugin returns `SERVFAIL` when the plugin's zone matches a query name but no template block matches the query type. Strict POSIX resolvers query A and AAAA in parallel and treat SERVFAIL as a hard failure — breaking intra-cluster resolution for actor images that only have IPv4 addresses.

### Fix

Add a `template IN ANY` block (with no `answer`) that matches the same actor names. Per the [CoreDNS template plugin docs](https://coredns.io/plugins/template/), specifying no answer results in a response with an empty answer section and default rcode NOERROR — the NODATA response RFC 8020 expects for a name that exists but has no records of the requested type.

### Testing

- `go test ./cmd/atenet/internal/dns/ -count=1` — passes
- `go build ./cmd/atenet/internal/dns/` — clean
- `gofmt -l cmd/atenet/internal/dns/` — clean
- Generated Corefile verified to contain both blocks:
  ```
  template IN A actors.resources.substrate.ate.dev {
    match "^[a-z0-9]([-a-z0-9]*[a-z0-9])?\.[a-z0-9]([-a-z0-9]*[a-z0-9])?\.actors\.resources\.substrate\.ate\.dev\.$"
    answer "{{ .Name }} 60 IN A 10.240.0.10"
  }
  template IN ANY actors.resources.substrate.ate.dev {
    match "^[a-z0-9]([-a-z0-9]*[a-z0-9])?\.[a-z0-9]([-a-z0-9]*[a-z0-9])?\.actors\.resources\.substrate\.ate\.dev\.$"
  }
  ```

### Notes

- The `ANY` template comes after `A`, so A queries are still answered first.
- No `answer` in the ANY block → empty NOERROR (NODATA), not NXDOMAIN — the zone apex and empty non-terminals remain distinct (see #922).

### DIFF
--- a/cmd/atenet/internal/dns/corefile.go
+++ b/cmd/atenet/internal/dns/corefile.go
@@ -42,11 +42,28 @@ func buildTemplate() string {
 	// Construct match pattern for <ActorName>.<atespace>.<dnsDomain>. Both the
 	// actor name and the atespace are DNS-1123 labels (same regex).
 	directives = append(directives, fmt.Sprintf("template IN A %s {", resources.ActorDNSSuffix))
-	// Escape the suffix's dots so they match literally; the final \. matches the FQDN's trailing dot.
+	// Escape the suffix's dots so they match literally; the final \\. matches the FQDN's trailing dot.
 	escapedSuffix := strings.ReplaceAll(resources.ActorDNSSuffix, ".", `\.`)
 	directives = append(directives, fmt.Sprintf(`  match "^%s\.%s\.%s\.$"`, resources.ResourceNameRegexPattern, resources.ResourceNameRegexPattern, escapedSuffix))
 	// Note the %s -- this will be filled with the router IP.
 	directives = append(directives, `  answer "{{ .Name }} 60 IN A %s"`)
+	directives = append(directives, "}")
+
+	// Non-A queries for valid actor names must return NODATA (NOERROR with an
+	// empty answer section), not SERVFAIL. CoreDNS's template plugin returns
+	// SERVFAIL when a template's zone matches a query name but no template
+	// block matches the query type (see plugin/template README: "Without
+	// fallthrough, when the template's ZONE matches a query but no regex match
+	// then a SERVFAIL response is returned"). Strict POSIX resolvers query A
+	// and AAAA in parallel and treat SERVFAIL as a hard failure, breaking
+	// intra-cluster resolution for actor images that don't have AAAA records.
+	//
+	// template IN ANY matches every type, and specifying no answer results in
+	// a response with an empty answer section and the default rcode NOERROR —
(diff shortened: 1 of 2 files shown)


ygao-g:
FYI I already opened a PR for the issue https://github.com/agent-substrate/substrate/issues/888
image

krsnaSuraj:
Thanks for the heads-up @ygao-g — you're right, #874 covers the same ground (Fixes #888) and was opened a week earlier.

I compared both implementations:

- **#874** adds `template ANY ANY` with a **SOA authority record** and `fallthrough`, plus README docs. The SOA is what makes the NODATA response negatively cacheable per RFC 2308, and the fallthrough keeps non-actor names in the zone from being swallowed.
- **#1062** (mine) adds `template IN ANY` with an empty answer and no SOA/fallthrough. It returns NOERROR/NODATA for non-A qtypes on valid actor names, which fixes the musl/AAAA-strict-resolver failure mode, but it's the thinner fix of the two.

#874 is the more complete and correct approach — it landed first and handles the negative-caching and NXDOMAIN cases properly. I'll close #1062 in favor of it. Thanks for the catch.

krsnaSuraj:
Closing in favor of #874 (same fix, opened earlier, more complete with SOA authority + fallthrough + docs). Thanks @ygao-g for the FYI.

PR #1063, opened 2026-08-19
Skip the template selector when resuming a paused actor

HavenXia:
This was found during upgrade PoC.

If an actor was suspended, its `RequiredNodes` is the source of truth of placement when resuming. However, our current scheduling mechanism will check also `ActorTemplateSpec.WorkerSelector` and `Actor.WorkerSelector`.

It's a chance that the `ActorTemplateSpec.WorkerSelector` was updated to pick a worker not on the node this actor got paused, to allow paused actor get resumed, we should only pick node as constraints when this happen.

- [ ] Tests pass


HavenXia:
I think it actually depends on whether we could change worker selector for templates. It was a step in blue/green upgrade PoC but we are getting away from that idea. 

@juli4n What do you think?

PR #1065, opened 2026-08-19
[WIP] ci: run the IPv6-only e2e job against the full IPv6 stack [DO NOT MERGE — signal only]

ygao-g:
Part of #246.

The IPv6-only kind e2e job brings a cluster up on main, but the four actor
egress tests fail: the actor's netns has no IPv6 route, so every dial to an
in-cluster address returns `network is unreachable`. This branch stacks the
commits that close that gap and runs the job against them, so the result is
visible before any of them merge.

Where each commit comes from, in branch order:

| commit | subject | source |
|---|---|---|
| `8cd2a012` | hack: drop the IPv6 kubeconfig repoint | #958 |
| `24567105` | hack: fix DNS on IPv6-only kind clusters | #958 |
| `4d00b5e6` | hack: add opt-in NAT64 support to IPv6-only kind clusters | #1275 |
| `00c5406d` | atunnel: support IPv6 original destination lookup | #753 |
| `f3f3814b` | atunnel: preserve IPv4 original destination errors | #753 |
| `ef064bb6` | atunnel: stabilize original destination tests | #753 |
| `a9bade05` | atunnel: select original destination by family | #753 |
| `af37c2ac` | atunnel: identify original destination family in errors | #753 |
| `5506f7a1` | atunnel: isolate original destination tests | #753 |
| `d1469ad2` | atunnel: test original destination formatting | #753 |
| `cf0fbcd5` | atunnel: document original destination buffer sizes | #753 |
| `404ec608` | ateom: drop the family from the atunnel ingress listen defaults | #1080 |
| `b3ef35f9` | atecontroller: drop the family from the atunnel ingress args | #1080 |
| `fcb1a890` | ateom: fix the comment about which listen addresses are dual-stack | #1080 |
| `606825e6` | ateomnet: enable IPv6 forwarding in worker pod netns | #979 |
| `1cd4703c` | ateomnet: return nil when sysctl path missing in writeSysctlIfUnset | #979 |
| `b357f52c` | ateomnet: rename EnableIPv4Forwarding to EnableForwarding | #979 |
| `a67785c6` | ateomnet: cover writeSysctlIfUnset's missing-path branch | #979 |
| `c501a6ee` | atenet/egress: resolve upstream names on both address families | #1060 |
| `aeb1b7cf` | ci: add an IPv6-only kind e2e job | #939 |
| `47655ec7` | ateomnet: move the actor nftables table to the inet family | #1116 |
| `703b8ee9` | ateomnet: give the actor an IPv6 address when the pod has one | #1057 |

Every row is that PR's commit unchanged, with its original author, matched
back to its source by `git patch-id` rather than by subject. The branch no
longer carries any fix of its own: the two original-destination reworks it
used to hold have been taken upstream into #753, #911 has merged, and #1275
joins because the runner needs a translator before an IPv6-only cluster can
reach anything.

Rebased onto main at `69428103`; the job runs on this pull request directly
and is green — 11 passed, 0 skipped — with all four egress tests going from
fail to pass and `TestActorEgress` dropping from 30.6s to 0.5s, the first dial
succeeding instead of retrying to the deadline. Signal only; it will never be
merged.


internal/atunnel/original_dst_linux.go
```diff
@@ -27,8 +26,10 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// IP6T_SO_ORIGINAL_DST is not generated by golang.org/x/sys/unix. It is
-// defined as 80 in linux/netfilter_ipv6/ip6_tables.h.
+// IP6T_SO_ORIGINAL_DST is not generated by golang.org/x/sys/unix. It is defined
```
ygao-g: Nit. make comment concise.

internal/atunnel/original_dst_linux.go
```diff
@@ -42,9 +43,10 @@ func TCPOriginalDestination(conn net.Conn) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("atunnel: acquiring TCP syscall connection: %w", err)
 	}
-	// The IPv6 option is only meaningful on an AF_INET6 socket: on AF_INET the
-	// kernel returns EOPNOTSUPP, which would mask the real IPv4 error. A
-	// v4-mapped local address still means an IPv4 flow, so To4 is the test.
+	// Each family keeps its original destination under its own socket option
```
ygao-g: Nit. make comment concise.

PR #1067, opened 2026-08-19
kubectl-ate: add --container  log filters for actor logs

HavenXia:
Part of  #294 as #311 has been inactive for over a month.

Example

```
kubectl ate logs actors test -a demo                                      # default: all containers + lifecycle


kubectl ate logs actors test -a demo -c counter                           # only specified containers
```

Scoped this down to --container only after discussing with @BenTheElder  — we are not sure the naming of the container related logs, the supervisor output may end up as a describe/events-style like how k8s does?

- [x] Tests pass

### DIFF
--- a/cmd/kubectl-ate/internal/cmd/logs_actors.go
+++ b/cmd/kubectl-ate/internal/cmd/logs_actors.go
@@ -42,6 +42,7 @@ import (
 
 var followLogs bool
 var logsAtespaceFlag string
+var logsContainerFlag string
 
 var logsActorsCmd = &cobra.Command{
 	Use:     "actors <actor-name>",
@@ -55,6 +56,7 @@ func init() {
 	logsActorsCmd.Flags().BoolVarP(&followLogs, "follow", "f", false, "Specify if the logs should be streamed.")
 	logsActorsCmd.Flags().StringVarP(&logsAtespaceFlag, "atespace", "a", "", "Atespace the actor lives in")
 	_ = logsActorsCmd.MarkFlagRequired("atespace")
+	logsActorsCmd.Flags().StringVarP(&logsContainerFlag, "container", "c", "", "Show only logs from this container.")
 	logsCmd.AddCommand(logsActorsCmd)
 }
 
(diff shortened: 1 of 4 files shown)


cmd/kubectl-ate/internal/cmd/logs_actors.go
```diff
@@ -340,10 +374,10 @@ func filterAndDisplayLogLine(line string, target resources.ActorRef, w io.Writer
 	// actors from different atespaces over time, so match on both. A partial
 	// target never matches: a line missing either label would otherwise be
 	// attributed to whichever actor was asked for.
-	matched := emitter == target && target.Atespace != "" && target.Name != ""
```
juli4n, line 343: How about moving all the checks to a `matches` method in `logLineFilter`?

HavenXia: moved, and added example in pr description

cmd/kubectl-ate/internal/cmd/logs_actors.go
```diff
@@ -55,6 +57,8 @@ func init() {
 	logsActorsCmd.Flags().BoolVarP(&followLogs, "follow", "f", false, "Specify if the logs should be streamed.")
 	logsActorsCmd.Flags().StringVarP(&logsAtespaceFlag, "atespace", "a", "", "Atespace the actor lives in")
 	_ = logsActorsCmd.MarkFlagRequired("atespace")
+	logsActorsCmd.Flags().StringVarP(&logsContainerFlag, "container", "c", "", "Show only logs from this container; with --supervisor, also show lifecycle events.")
+	logsActorsCmd.Flags().BoolVar(&logsSupervisorFlag, "supervisor", false, "Show only the supervisor's actor lifecycle events; with --container, also show that container's logs.")
```
BenTheElder: we might want to think of another name for this, is this supposed to be ateom? while I don't want to center it on ateom, "supervisor" is going to mean e.g. openshell to a lot of our users

HavenXia: Scoped down to only --container after offline discussion

BenTheElder (approved):

PR #1070, opened 2026-08-19
manifests: count telemetry by service on kind

JeffLuoo:
The volume measurement in current benchmarking repo worked on GKE only. There the managed collector cannot hold a connector, thus benchmarking/telemetry/meter.yaml puts a second collector in front of it and counts the data as it goes through.

This PR adds the support to kind cluster to count the telemetry data.

Follow-up to #749.

- [x] Tests pass

### DIFF
--- a/manifests/ate-install/kind/otel-collector.yaml
+++ b/manifests/ate-install/kind/otel-collector.yaml
@@ -33,6 +33,47 @@ data:
             endpoint: 0.0.0.0:4318
     processors:
       batch:
+
+      # The count connector below sends deltas, and the Prometheus exporter
+      # reads the new start timestamp of each batch as a reset. Without
+      # deltatocumulative the counts do not accumulate and each scrape shows
+      # only the last batch.
+      #
+      # max_streams bounds the memory. A stream is for each process and
+      # count, and not for each service, because serverboot gives each
+      # process its own service.instance.id. Above the limit new streams drop
+      # with no log line and the totals become too low.
+      deltatocumulative:
+        max_streams: 10000
+
+    # The volume of telemetry for each service, the same two counts that
+    # benchmarking/telemetry/meter.yaml gives on GKE. On GKE the managed
+    # collector cannot hold a connector, thus a measurement there needs a
+    # second collector in front of it. Here the collector is ours, thus the
+    # counts belong in it and kind needs no meter.
+    connectors:
+      count:
+        spans:
+          substrate_spans:
+            description: Spans received, by emitting service.
+            attributes:
+              # The connector reads each key from the attributes of the item,
+              # then the scope, then the resource, thus service.name resolves
+              # from the resource with no transform in front of it. The
+              # Prometheus exporter writes the key as service_name.
+              #
+              # Without a default value the connector removes each item that
+              # has no service.name from the count. The item does not show as
+              # unknown, and the totals are too low with no error message.
+              - key: service.name
+                default_value: unknown
+        datapoints:
+          substrate_datapoints:
+            description: Metric datapoints received, by emitting service.
+            attributes:
+              - key: service.name
+                default_value: unknown
+
     exporters:
       otlp/jaeger:
         endpoint: jaeger.otel-system.svc.cluster.local:4317
@@ -52,6 +93,39 @@ data:
           receivers: [otlp]
           processors: [batch]
           exporters: [prometheus, debug]
+
+        # The count pipelines read the same receiver as the two pipelines
+        # above, and they only count. The connector declares MutatesData
+        # false, thus it changes nothing that reaches Jaeger or Prometheus.
+        traces/count:
+          receivers: [otlp]
+          exporters: [count]
+        metrics/count:
+          receivers: [otlp]
+          exporters: [count]
+
+        # The counts that the connector makes, out to the Prometheus exporter.
+        metrics/count-out:
+          receivers: [count]
+          processors: [deltatocumulative]
+          exporters: [prometheus]
+
+      # The otelcol_* counters of the collector itself. A service that reports
+      # zero volume gives no data by itself: otelcol_receiver_accepted_* shows
+      # the difference between a service that sent nothing and data that never
+      # arrived. The default reader binds to localhost, thus give the address
+      # of each interface so the Prometheus pod can reach it.
+      #
+      # The level stays at the default. It already gives each counter above,
+      # and detailed adds histograms only, which no query here reads.
+      telemetry:
+        metrics:
+          readers:
+            - pull:
+                exporter:
+                  prometheus:
+                    host: 0.0.0.0
+                    port: 8888
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -69,6 +143,10 @@ spec:
     metadata:
       labels:
         app: opentelemetry-collector
+      # There is no prometheus.io/scrape annotation here. The annotation can
+      # name one port only, and the collector has two to read: 8889 for the
+      # substrate_* counts and 8888 for the otelcol_* counters. The
+      # otel-collector-self job in prometheus.yaml reads both.
     spec:
       containers:
       - name: otel-collector
@@ -85,6 +163,8 @@ spec:
           containerPort: 4318
         - name: prom
           containerPort: 8889
+        - name: self
+          containerPort: 8888
       volumes:
       - name: config
         configMap:
@@ -112,6 +192,9 @@ spec:
   - name: prom
     port: 8889
     targetPort: 8889
+  - name: self
+    port: 8888
+    targetPort: 8888
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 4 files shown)


JeffLuoo:
cc - @krisztianfekete

manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -33,6 +33,55 @@ data:
...
+      transform/svc:
+        error_mode: ignore
+        trace_statements:
+          - context: span
+            statements:
+              - set(attributes["svc"], resource.attributes["service.name"])
+        metric_statements:
+          - context: datapoint
+            statements:
+              - set(attributes["svc"], resource.attributes["service.name"])
```
krisztianfekete: The connector resolves each configured dimension against first attributes > scope > resource, finally `default_value`, so I don't think we even need this processor.

krisztianfekete: Can you please test this, and if it works without it, also remove it from here: https://github.com/agent-substrate/substrate/blob/bc51ef2452c4bf4c0542cd6850040c9ed1033421/benchmarking/telemetry/meter.yaml#L67-L90?

manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -69,6 +150,12 @@ spec:
     metadata:
       labels:
         app: opentelemetry-collector
+      annotations:
+        # The substrate_* counts, on the port of the Prometheus exporter. The
+        # otelcol_* counters are on 8888; the annotation can name one port
+        # only, thus a scrape job in prometheus.yaml collects that one.
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "8889"
```
krisztianfekete: 8889 re-exports every metric the components push, and ateapi, atelet, and atenet-router already have this annotation on 9090, so every series will be ingested twice and `sum()` doubles.

manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -33,6 +33,55 @@ data:
...
+      count:
+        spans:
+          substrate_spans:
+            description: Spans received, by emitting service.
+            attributes:
+              # Without a default value the connector removes each item that
+              # has no service.name from the count. The item does not show as
+              # unknown, and the totals are too low with no error message.
+              - key: svc
+                default_value: unknown
```
krisztianfekete: Let's group on the semconv key instead, which is also what the spanmetrics connector uses by default.

manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -52,6 +101,38 @@ data:
...
+        # that the count needs does not reach Jaeger.
+        traces/count:
+          receivers: [otlp]
+          processors: [transform/svc]
+          exporters: [count]
+        metrics/count:
+          receivers: [otlp]
+          processors: [transform/svc]
+          exporters: [count]
+        metrics/counts:
```
krisztianfekete: Let's make these easier to distinguish: 
```suggestion
        metrics/count-out:
```

manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -33,6 +33,55 @@ data:
             endpoint: 0.0.0.0:4318
     processors:
       batch:
+
+      # The count connector below sends deltas, and the Prometheus exporter
+      # reads the new start timestamp of each batch as a reset. Without
+      # deltatocumulative the counts do not accumulate and each scrape shows
+      # only the last batch.
+      deltatocumulative:
```
krisztianfekete: Nit: Let's set `max_streams` to 1000 or similar reasonable number to have some control over cardinality.

manifests/ate-install/kind/prometheus.yaml
```diff
@@ -107,6 +107,31 @@ data:
...
+              - __meta_kubernetes_pod_label_app
+              - __meta_kubernetes_pod_container_port_number
+            action: keep
+            regex: opentelemetry-collector;8888
+          - source_labels: [__meta_kubernetes_pod_name]
+            action: replace
+            target_label: pod
+          - source_labels: [__meta_kubernetes_namespace]
+            action: replace
+            target_label: namespace
```
krisztianfekete, line 142: Something like this should work to avoid duplicated reads I think, see the comment above for port 8889.
```suggestion
      - job_name: otel-collector-self
        kubernetes_sd_configs:
          - role: pod
            namespaces:
              names:
                - otel-system
        relabel_configs:
          - source_labels:
              - __meta_kubernetes_pod_label_app
              - __meta_kubernetes_pod_container_port_number
            action: keep
            regex: opentelemetry-collector;(8888|8889)
          - source_labels: [__meta_kubernetes_pod_name]
            action: replace
            target_label: pod
          - source_labels: [__meta_kubernetes_namespace]
            action: replace
            target_label: namespace
        metric_relabel_configs:
          - source_labels: [__name__]
            action: keep
            regex: (substrate|otelcol)_.*
```

manifests/ate-install/kind/otel-collector.yaml
JeffLuoo: tested, and updated to remove it

manifests/ate-install/kind/otel-collector.yaml
JeffLuoo: done

manifests/ate-install/kind/otel-collector.yaml
JeffLuoo: updated

manifests/ate-install/kind/prometheus.yaml
JeffLuoo: updated

manifests/ate-install/kind/otel-collector.yaml
JeffLuoo: Confirmed, and thank you.

I removed the prometheus.io/scrape and prometheus.io/port annotations from the collector pod

manifests/ate-install/kind/otel-collector.yaml
JeffLuoo: Done, in both files.

manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -52,6 +92,37 @@ data:
...
+          exporters: [prometheus]
+
+      # The otelcol_* counters of the collector itself. A service that reports
+      # zero volume gives no data by itself: otelcol_receiver_accepted_* shows
+      # the difference between a service that sent nothing and data that never
+      # arrived. The default reader binds to localhost, thus give the address
+      # of each interface so the Prometheus pod can reach it.
+      telemetry:
+        metrics:
+          level: detailed
```
krisztianfekete: Isn't the default verbosity enough?

manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -33,6 +33,46 @@ data:
...
+      # The count connector below sends deltas, and the Prometheus exporter
+      # reads the new start timestamp of each batch as a reset. Without
+      # deltatocumulative the counts do not accumulate and each scrape shows
+      # only the last batch.
+      #
+      # max_streams bounds the cardinality. There is one stream for each pair
+      # of a count and a service, thus a limit far above the count of
+      # services is sufficient.
+      deltatocumulative:
+        max_streams: 1000
```
krisztianfekete: I just tested this locally and  in `max_streams` streams are actually per source resource, and since serverboot gives every process its own `service.instance.id` 1000 is not a lot. Maybe we can use locally, but in meter.yaml we can configure a lot higher value?

JeffLuoo: Consider this will be used by benchmarking. How about 10000 in kind and 100k in meter.yaml?

At a few hundred bytes per tracked datapoint that's roughly 40~50 MB of data in memory, negligible against the memory limit of collector.

manifests/ate-install/kind/otel-collector.yaml
JeffLuoo: You are right, the default is sufficient. Tested in my local cluster and confirmed.

manifests/ate-install/kind/otel-collector.yaml
krisztianfekete: 10000 on kind sounds right to me. I measured ~5KiB per stream end to end, so 10k is ~ 50MiB.

For 100k on the meter the memory is OK with 263MiB live heap, 675MiB RSS, but it becomes 38.5 MiB per scrape at 100k series, and we scrape every 10s from a local Prometheus. 

I'd  take 20000 there and add `sum(otelcol_deltatocumulative_datapoints{error="limit"})` must be 0 + `streams_tracked` / `streams_limit` to the pass criteria, since a limit hit deletes series from the scrape with no log line at all.

Wdyt?

JeffLuoo: sg, updated

krisztianfekete (approved):
There are some flakes (?), but LGTM!

git286 (approved):

PR #1071, opened 2026-08-19
benchmarking: run the observability ladder as one test

JeffLuoo:
Follow-up to #749, for comment of adding the ladder design in benchmarking for observability.

### DIFF
--- a/benchmarking/automation/tests.yaml
+++ b/benchmarking/automation/tests.yaml
@@ -48,6 +48,13 @@
 # Optional `flags` is a list of arbitrary strings appended to runner.py's
 # argv; runner.py forwards them to locust verbatim.
 #
+# `--ladder users:duration[:trace_probability],...` in `flags` turns the test
+# into a sequence of steps in one run (benchmarking/locust/shapes/ladder_shape.py).
+# The shape then owns the load and the end of the run, so runner.py stops
+# forwarding `-t` and `users` becomes the count before the first tick only.
+# Keep `duration` at the total of the steps: the orchestrator sizes the Job
+# timeout from it.
+#
 # ---- type: nighthawk-ingress --------------------------------------------------------
 #
 # Runs the Nighthawk adaptive router-capacity benchmark. `duration` is the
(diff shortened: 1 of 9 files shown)


benchmarking/automation/tests.yaml
```diff
@@ -20,6 +20,13 @@
 # Optional `flags` is a list of arbitrary strings appended to runner.py's
 # argv; runner.py forwards them to locust verbatim.
 #
+# `--ladder users:duration[:trace_probability],...` in `flags` turns the test
```
maxsmythe: how does trace probability set in the ladder interact with trace probability set via flag?

benchmarking/locust/runner.py
```diff
@@ -44,6 +49,15 @@
 # benchmarking/locust/Dockerfile.
 BOOMER_BINARY = "/app/boomer-glutton"
 
+# Loaded next to the test file for a --ladder run. It holds no user class, so
+# locust must never get it on its own.
+LADDER_SHAPE = str(Path(__file__).resolve().parent / "shapes" / "ladder_shape.py")
```
maxsmythe: To help maintain extensibility, can we just require the tests.yaml author to explicitly mix in the ladder_shape.py file? I believe `file` takes a comma-separated list of paths (perhaps it should have been "files")

This way users can mix and match test classes and load shapes and runner.py doesn't need to be as coupled to the test suites.

benchmarking/locust/runner.py
```diff
@@ -114,6 +151,8 @@ def log_run_config(args: argparse.Namespace, dest_prefix: str, work_dir: Path, l
         f"  duration:       {args.duration}",
         f"  users:          {args.users}",
         f"  uses_boomer:    {needs_boomer(args.file)}",
+        f"  ladder:         {ladder_spec(args.locust_extra) or '(none)'}",
```
maxsmythe: Instead of making special flags per-test, maybe we can make the "extra flags" output more legible (e.g. not all in one line, but multiple indented lines)

benchmarking/locust/runner.py
```diff
@@ -185,19 +224,31 @@ def run_test(args: argparse.Namespace, csv_prefix: Path, logs: TextIO, traces: T
     siphoned into traces.txt as a deduped one-per-line list.
     """
     with_boomer = needs_boomer(args.file)
+    ladder = ladder_spec(args.locust_extra)
+
+    # A ladder holds a step's user count for the step's duration and then
```
maxsmythe: I think if we want to evaluate test-specific or shape-specific arguments, we should do that where the test or shape is defined to try to maintain separation-of-concerns

maxsmythe (changes requested):
Looks good! Main feedback is to try to separate test and shape definitions from runner.py code to help establish a pattern.

I expect there will be multiple new shapes/tests in the future.

benchmarking/automation/tests.yaml
JeffLuoo: --trace-probability gives the rate before the first step that names one. A step that names a rate changes it when the step starts, and that rate stays until another step changes it. Thus a ladder with no rate in any step keeps the flag from start to end.

Also make it clear in a comment.

benchmarking/locust/runner.py
JeffLuoo: Done - Decoupled runner.py completely from ladder_shape.py

benchmarking/locust/runner.py
JeffLuoo: Updated

Removed `ladder_spec()` and all custom ladder inspection from `runner.py`.

benchmarking/locust/runner.py
JeffLuoo: Done. Removed the dedicated `ladder` log line and added `pair_flags()` to format all `extra flags` across multiple indented lines with flags and values grouped together:
```text
==== Run config ====
  name:           observability_s1_user_sweep
  ...
  extra flags:
    --ladder 5:3m,10:3m,15:3m
    --trace-probability 0.1
    --min-wait-time 1.0
    --max-wait-time 1.0
====================

maxsmythe (approved):
LGTM

PR #1073, opened 2026-08-19
csi: Use default csi node hostpath

hajiler:
Mount `kubelet-pods` host path to atelet for CSI node operations. This hostpath is the default for the average CSI driver, and will enable compatibility with default CSI node daemonsets without additional customization. 

Volumes are mounted to the node at the following path: `/var/lib/kubelet/pods/[worker-pod-uuid]/substrate-volumes/[actor-uuid]/[vol-name]`". By mounting the directory in substrate-volumes, this avoids interferences from the volume reconstruction feature, which uses `var/lib/kubelet/pods/*/volumes/` for discovery. 

TAG=agy
CONV=945df49f-7efa-4dc1-a9e4-03ecbbed676f

#232  

- [X ] Tests pass


mattcary:
lgtm from a CSI driver perspective

msau42:
We discussed offline, but  I don't think we want to depend on "/var/lib/kubelet/pods/[worker-pod-uuid]/substrate-volumes/[actor-uuid]/[vol-name]" because we won't be resilient to worker pod crashes/restarts/upgrades, etc.

I suggest exploring something under "/var/lib/kubelet/plugins"

PR #1075, opened 2026-08-19
ateapi: stamp the worker pool on successful suspend and pause

### LINKED ISSUE #957: ateapi: suspend and pause stamp the worker pool only when they fail
## Expected Behavior

`ate.actor.lifecycle.operation.duration` carries the `ate.workerpool.namespace` + `ate.workerpool.name` pair on every suspend and pause that ran on a worker. A successful suspend of an actor on `ate-workers/pool-a` reports that pool, so pool-level latency and error rates stay comparable across the five operations.

## Actual Behavior

The pair is inverted for suspend and pause: a **successful** operation reports no pool, and a **failed** one reports it.

Both workflows record the histogram from a deferred call that reads the `actor` variable:

https://github.com/agent-substrate/substrate/blob/2b3a4715c6b7af4debe42c32791ae7307f18ca1b/cmd/ateapi/internal/controlapi/workflow_suspend.go#L45-L48

The happy path reassigns that variable to the finalized record as its last step:

- `workflow_suspend.go:85-90` — `actor = finalized` from `ensureSuspendedFinalized`
- `workflow_pause.go:77-82` — `actor = finalized` from `ensurePausedFinalized`

`FinalizeSuspended` and `FinalizePaused` commit `STATUS_SUSPENDED`/`STATUS_PAUSED` and the cleared `WorkerAssignment` in one update. By the time the defer runs, `actor.GetWorkerAssignment()` is nil, so `lifecycleOpAttrs` omits both keys. A failure returns early with the pre-finalize record, which still names the worker, so the pool pair lands there instead.

The re-entrant fast-forward path has the same result for a different reason: a workflow that finds the actor already `SUSPENDED`/`PAUSED` returns a record that carries no assignment.

Of the five operations, `resume` is the only one where the pool is reliable. `create` has no assignment yet, and `delete` builds its labels by hand (`actor.go:313-322`) and never stamps a pool at all.

Net effect: any dashboard that breaks lifecycle latency down by pool sees suspend and pause only when they fail.

1. Resume an actor so it holds a `WorkerAssignment` in a named pool.
1. Suspend it, and let the workflow reach `FinalizeSuspended`.
1. Read `ate.actor.lifecycle.operation.duration` for `ate.actor.operation.name=suspend`: the sample has neither `ate.workerpool.namespace` nor `ate.workerpool.name`.
1. Repeat with a suspend that fails after the assign step. That sample carries both keys.

## Suggested Fix

Snapshot the labels before the record loses the pointers they read, the way `crash.go` already does for the crash counter:

https://github.com/agent-substrate/substrate/blob/2b3a4715c6b7af4debe42c32791ae7307f18ca1b/cmd/ateapi/internal/controlapi/crash.go#L74-L76

Concretely, in `SuspendActor` and `PauseActor`, build `lifecycleOpAttrs(...)` into a variable just before the finalize step and let the defer use it when set, falling back to the current computation for paths that end earlier.

Worth deciding at the same time whether `delete` should carry the pool, since it is released before the operation completes and hits the same ordering problem.

## Specifications

- Version: `main` @ 2b3a4715
- Platform: any

Follow-up from review of #953 (comment https://github.com/agent-substrate/substrate/pull/953#discussion_r3785442209).

JeffLuoo:
Fixes #957

`ate.actor.lifecycle.operation.duration` carried the pool pair on suspend and pause only when they failed, so per-pool dashboards saw those two operations exclusively as failures.

Both workflows record the histogram from a defer that reads the `actor` variable, and the happy path reassigns it to the finalized record. The finalize step commits the new state and the cleared `WorkerAssignment` in one update, so the defer found no assignment and dropped both keys. A failure returns the pre-finalize record, which still names the worker.

Both now snapshot `lifecycleOpAttrs(...)` just before the finalize step — the same snapshot-before-clear crash.go does for the crash counter. Paths that end earlier keep the current computation.

`delete` stays without a pool: it only runs from SUSPENDED or CRASHED, which already released the worker, so there is none to name.

TestLifecycleOpPoolAttributesOnSuccess drives a real suspend and pause through the gRPC service; both subtests fail without the fix.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_suspend.go
+++ b/cmd/ateapi/internal/controlapi/workflow_suspend.go
@@ -27,6 +27,7 @@ import (
 	"github.com/agent-substrate/substrate/internal/resources"
 	atev1alpha1 "github.com/agent-substrate/substrate/pkg/api/v1alpha1"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+	"go.opentelemetry.io/otel/attribute"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
@@ -41,10 +42,16 @@ func (w *ActorWorkflow) SuspendActor(ctx context.Context, actorRef resources.Act
 	var actor *ateapipb.Actor
 	var actorTemplate *atev1alpha1.ActorTemplate
 	var wireSnapshotScope string
+	// Set just before finalize; nil until then, so earlier exits label
+	// themselves from the record they hold.
+	var finalAttrs []attribute.KeyValue
 
 	defer func() {
-		w.instruments.recordLifecycleOp(ctx, ateattr.OperationSuspend, start, err,
-			lifecycleOpAttrs(actor, actorTemplate, "", wireSnapshotScope)...)
+		attrs := finalAttrs
+		if attrs == nil {
+			attrs = lifecycleOpAttrs(actor, actorTemplate, "", wireSnapshotScope)
+		}
+		w.instruments.recordLifecycleOp(ctx, ateattr.OperationSuspend, start, err, attrs...)
 	}()
 
 	lockCtx, lock, err := w.acquireActorLock(ctx, actorRef)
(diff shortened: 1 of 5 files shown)


cmd/ateapi/internal/controlapi/workflow_suspend.go
krisztianfekete, line 73: Nice, but this still leaves the third path the issue calls out, where a re-entered suspend that finds the actor already SUSPENDED returns here as a success with no pool. 

Not fixable here I guess as the worker's already released, so probably a comment here + a reply on the issue would be fine for now, just so no one expects that to work.

krisztianfekete (approved):

cmd/ateapi/internal/controlapi/workflow_suspend.go
JeffLuoo: Updated.

git286 (approved):

PR #1078, opened 2026-08-19
Add Docker pull secret support

brendandburns:
If a registry where images are pushed is a private-only registry the subsequent deployment fails. This adds support for reading secrets from the docker registry config and adding it to pull secrets for the deployed images.

- [X] Tests pass


BenTheElder:
See discussion on #869 

There is a draft for adding credential provider support in #917 pending further verification that this can be a workable solution for AKS.

brendandburns:
@BenTheElder I think this is different, this is for deploying the agent-substrate infrastructure itself (e.g. the controller etc) if I run install-ate.sh today and my registry is fully private, the script fails because k8s can't pull the images that are built/pushed as part of the install.

I guess this is a bit of an edge case since people will eventually just use publically released images, but for development today it's a bit of a blocker at least for my personal setup where my OCI registry is configured to block anonymous images for security reasons.

BenTheElder:
> [Benjamin Elder (@BenTheElder)](https://github.com/BenTheElder) I think this is different, this is for deploying the agent-substrate infrastructure itself (e.g. the controller etc) if I run install-ate.sh today and my registry is fully private, the script fails because k8s can't pull the images that are built/pushed as part of the install.

Ah, apologies, yes that makes sense.

I didn't look closely enough, the PR volume has gotten quite high and we had recent conversations about the actor side of this. The "infrastructure" side should embrace k8s concepts IMHO.

> I guess this is a bit of an edge case since people will eventually just use publically released images, but for development today it's a bit of a blocker at least for my personal setup where my OCI registry is configured to block anonymous images for security reasons.

To be honest, I think the majority of development is on kind clusters with local unathenticated pulls. It's not intended to only work there + GCP though ...

FYI @juli4n @sfunkenhauser re: worker API changes.

hack/install-ate.sh
```diff
@@ -207,6 +208,128 @@ default_postgres_connection_string() {
...
+
+install_docker_pull_secret_in_namespace() {
+  local namespace="$1"
+  shift
+  local secret_name=""
+  secret_name="$(docker_pull_secret_name)"
+  local docker_config=""
+  docker_config="$(docker_config_file)"
+
+  if [[ ! -f "${docker_config}" ]]; then
```
BenTheElder, line 244: non-blocking: This script is already getting rather complicated, but it's pretty common that the system docker config contains only credential helpers, we might have to point people at something like this:
https://kind.sigs.k8s.io/docs/user/private-registries/#:~:text=To%20generate%20a%20config%20file%20with%20plain%20credentials%2C%20create%20a%20temporary%20config%20that%20bypasses%20the%20credential%20store%20and%20log%20in%20with%20it%3A

benchmarking/workloads/deploy.sh
```diff
@@ -87,6 +87,14 @@ resolve_otlp_endpoint() {
   fi
 }
 
+workerpool_pull_secret_sed_expr() {
+  if [[ -n "${ATE_DOCKER_PULL_SECRET_NAME:-}" ]]; then
```
BenTheElder: 🤖 **should-fix** 🟡 – This gate can never be true when driven from `install-ate.sh`, so benchmarking workloads silently get no pull secret — the private-registry case the PR exists for.

Two reasons, either one sufficient:

- `ATE_DOCKER_PULL_SECRET_NAME` is assigned at `install-ate.sh:1146` but never exported, and `deploy_benchmarks` shells out to `deploy_locust.sh` (line 1029), which runs this script. The child process never sees it.
- The gate differs from its twin. `install-ate.sh` keys on `INSTALL_DOCKER_PULL_SECRET` and defaults the name inside `docker_pull_secret_name`; this keys on the name being non-empty. So plain `--install-docker-pull-secret`, which takes the default name and never sets the variable, would skip here even if it were exported.

Exporting the variable and giving this copy the same default would line the two up. Failing that, forwarding the name to `deploy_locust.sh` as a flag would at least make the coupling visible.

While you are here: this copy interpolates the name straight into the `sed` expression, without the `validate_docker_pull_secret_name` its twin runs first — so a name containing `|` breaks the expression rather than being rejected.

BenTheElder (approved):

PR #1079, opened 2026-08-19
imagecache: Fix TestConcurrentEnsureImageAndEvict flake

### LINKED ISSUE #1076: TestConcurrentEnsureImageAndEvict flakes
## Expected Behavior

TestConcurrentEnsureImageAndEvict passes: per its doc comment, whatever interleaving happens between EnsureImage and a free-everything EvictUnused pass, "either its touch wins and the evictor skips, or the evictor wins and EnsureImage re-pulls fresh dirs" — so EnsureImage never returns an error.

## Actual Behavior

Rarely, under CI load, an iteration fails with the pull path's defensive backstop error:

2026/08/19 20:32:41 INFO Image cache removed retired layer dirs count=1 took=432.006µs
    gc_test.go:507: iteration 24: EnsureImage: layer dir vanished during pull (evicted?): stat /tmp/TestConcurrentEnsureImageAndEvict2288464666/001/layers/sha256/99630e.../fs: no such file or directory
--- FAIL: TestConcurrentEnsureImageAndEvict (0.09s)

The test and the implementation disagree about the contract. The reuse/retire protocol serializes per layer — ensureLayer refreshes the dir mtime inside the layer singleflight, and retireLayer re-checks it inside the same flight and vetoes fresh dirs — which is airtight for a retire racing a concurrent reuse of the same layer. But it leaves a window: after ensureLayer returns a dir and before the image-level record rewrite re-references it, a retire can enter the flight with a cutoff at or past the refresh instant and legitimately rename the dir away. The pull path knows this and handles it by design — the check that fired is its explicit backstop (imagecache.go: "Never return LayerDirs that are not on disk right now: a vanished dir fails the pull into a clean RPC retry instead of a bundle spec naming a missing lowerdir"). In production this is a transient, self-healing error; the test treats it as a failure.

So this is a flake in effect (timing-dependent, pre-existing, harmless in production), but it points at a real gap: the implementation provides "fail clean and let the retry heal it," while the test asserts "eviction winning is a bug." The triggering PR (#803) contains no internal/imagecache changes.

Suggested resolution: make the test's stated invariant true rather than weakening the test — when the final stat in EnsureImage finds a vanished layer dir, loop back and re-pull (bounded; one retry suffices). The retry re-enters ensureLayer, which re-unpacks the layer under a fresh mtime, and the record rewrite re-references it — exactly the "evictor wins and EnsureImage re-pulls" behavior the test describes, and it also spares the production path a full RPC retry for the same interleaving.

1. Run go test -race -run TestConcurrentEnsureImageAndEvict ./internal/imagecache/ repeatedly on a loaded machine (observed on a GitHub Actions ubuntu-latest runner during a PR build; the failure hit iteration 24 of 25 in one run).
2. Note it does not reproduce readily on idle hardware: 100 runs × 25 iterations with -race -cpu=2,4 (~4,000 race windows) on unmodified main @ 43fa6ccb produced zero failures on an idle machine (darwin/arm64).
3. The interleaving needing to land is: ensureLayer returns a layer dir → evictor's retireLayer enters that layer's singleflight with cutoff ≥ the refresh mtime and renames the dir → EnsureImage's final stat loop finds the dir missing and returns the backstop error before the record rewrite would have protected it.

## Specifications

- Version: main @ 43fa6ccb (also present on earlier commits containing the GC/eviction work; the failing CI run was on a PR branch with no internal/imagecache changes)
- Platform: observed on GitHub Actions ubuntu-latest (linux/amd64), Go per go.mod; not reproduced in ~4,000 iterations on idle darwin/arm64

thompsonmax:
The reuse/retire protocol leaves one legitimate window: after ensureLayer refreshes a reused dir's mtime inside the layer singleflight, but before the image record rewrite re-references it, an eviction pass whose cutoff postdates the refresh may retire the dir. The pull's final re-verify caught this and failed the pull into an RPC retry, which TestConcurrentEnsureImageAndEvict (correctly, per its stated invariant) flagged as a flake under CI load.

Make the test's invariant true: when the re-verify finds a vanished layer dir, loop back and unpack the missing layers again, bounded to one extra attempt. One retry suffices because the winning pass listed its candidates before our record rewrite, so it cannot retire the re-unpacked dirs, and any later pass sees them referenced. Exhausting the attempts still fails with the original "vanished during pull" error.

TestPullReverifyFailsCleanlyOnYankedLayer asserted the old fail-clean behavior; it becomes TestPullReverifyRepullsYankedLayer, asserting the yanked layer is unpacked again and the pull succeeds.

Fixes #1076

### DIFF
--- a/internal/imagecache/imagecache.go
+++ b/internal/imagecache/imagecache.go
@@ -485,58 +485,78 @@ func (s *Store) pull(ctx context.Context, parsedRef name.Reference, digest v1.Ha
 		}
 	}
 
+	// An eviction pass can legitimately win the window between ensureLayer's
+	// mtime refresh and the record rewrite below: its cutoff may postdate the
+	// refresh, letting retireLayer rename a just-reused dir before the rewrite
+	// re-references it. When the re-verify catches that, loop back and unpack
+	// the missing layers again. One extra attempt suffices — the pass that
+	// won listed its candidates before this rewrite, so it cannot retire the
+	// re-unpacked dirs, and any later pass sees them referenced.
+	const maxUnpackAttempts = 2
 	layerDirs := make([]string, len(layers))
-	g, gctx := errgroup.WithContext(ctx)
-	g.SetLimit(layerPullConcurrency)
-	for i, layer := range layers {
-		g.Go(func() error {
-			diffID, err := layer.DiffID()
-			if err != nil {
-				return fmt.Errorf("while reading layer diffID: %w", err)
-			}
-			if diffID.String() != diffIDs[i] {
-				// The record must reference exactly what lands on disk.
-				return fmt.Errorf("layer %d diffID %s does not match config rootfs diffID %s", i, diffID, diffIDs[i])
-			}
-			dir, err := s.ensureLayer(gctx, diffID, layer)
-			if err != nil {
-				return fmt.Errorf("while unpacking layer %s: %w", diffID, err)
-			}
-			layerDirs[i] = dir
-			// Each completed layer refreshes the record's mtime: a pull
-			// making progress stays fresh indefinitely; a wedged one ages
-			// into ordinary LRU eviction. The twin is not touched — the
-			// primary keeps the layers referenced, and the final rewrite
-			// recreates the twin if it ages out mid-pull.
-			s.touchRecord(digest)
-			return nil
-		})
-	}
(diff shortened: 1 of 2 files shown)


thompsonmax:
Superseded by https://github.com/agent-substrate/substrate/pull/1509, will close this PR.

PR #1080, opened 2026-08-19
ateom: drop the family from the atunnel ingress listen defaults

ygao-g:
Part of #246.

Both ateom herders default the actor ingress flags to `0.0.0.0:443` and `0.0.0.0:8443`, and the
worker Deployment passes those same two values explicitly — so the command line on a running worker
reads IPv4-only. It never was: Go treats an unspecified IPv4 address as a wildcard on a listen and
binds it dual-stack, so those sockets already served both families. This spells all four `:443` and
`:8443` so the flag says what it does.

No behavior change. The egress listen address stays `0.0.0.0:15001` until the actor datapath
carries v6. #1042 carried the same lines under its egress pin and is closed as superseded.


bowei:
Can you merge the commits into one?

bowei (approved):

ygao-g:
> Can you merge the commits into one?

Done, thx for the reminder

haiyanmeng (approved):

PR #1088, opened 2026-08-20
microvm: document and log CSI volume snapshot behavior

mowangdk:
### Summary

   A Data-scope checkpoint for a microVM actor with only CSI volumes (no durable-dirs) correctly produces no snapshot files — CSI volumes persist independently on network storage and are re-mounted at restore.
   However, the silent behavior was confusing and flagged by a TODO: Revisit handling for CSI volumes since snapshots are currently quietly ignored.

   This PR replaces the TODO with clear documentation and adds explicit info logs when CSI volumes are present but not captured.

### Changes

 cmd/ateom-microvm/checkpoint.go
   - Replaces the TODO with documentation explaining the CSI snapshot contract: CSI volumes are NOT captured under any scope because they persist independently on the storage backend and are re-mounted at restore
   - Adds info log when CSI volumes are present but not captured, distinguishing:
     - CSI-only (no durable-dirs) → snapshot will produce no files
     - CSI + durable-dirs → durable-dirs will be captured, CSI is not

### DIFF
--- a/cmd/ateom-microvm/checkpoint.go
+++ b/cmd/ateom-microvm/checkpoint.go
@@ -88,16 +88,30 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 	// under either scope — and are the ONLY thing a Data-scope snapshot
 	// captures. DATA_ON_GOLDEN is restore-only (a DataOnGolden commit arrives
 	// here as plain DATA) and lands in the default rejection.
+	//
+	// CSI (external) volumes persist independently on network storage and are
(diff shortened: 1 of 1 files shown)


BenTheElder:
I don't think this should log. That's inconsistent with expected behavior. The TODO is more so to discuss it with dberkov when he's not on vacation

mowangdk:
> I don't think this should log. That's inconsistent with expected behavior. The TODO is more so to discuss it with dberkov when he's not on vacation

Got it, thanks. Closing this PR

PR #1091, opened 2026-08-20
atunnel: preserve HTTP/2 for actor requests

EItanya:
## Summary

- preserve HTTP/2 ingress requests when proxying to h2c Actors
- retain HTTP/1 behavior for existing Actors
- add a protocol regression test

## Verification

- `go test -race ./internal/atunnel`
- `make verify`


EItanya:
I'm actually going to convert this to a draft because I think we should get rid of this HTTP logic altogether in favor of the connect path recently added by @keithmattix

PR #1096, opened 2026-08-20
Add controller in ate apiserver to reconcile ActorTemplate state

zoez7:
Part of #477.

The controller has 2 components
1. A "producer" that periodically lists ActorTemplates and add work items to a client-go workqueue.
2. A "consumer" that gets work items from the queue, and reconcile it.

AT will have the FAILED condition if we encounter non-retriable errors during the transition


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -395,10 +395,11 @@ message Limits {
 enum ActorTemplatePhase {
```
juli4n: Can we avoid this 'phase' field? k8s had this pattern in the past and it's discouraged now because of how hard to evolve it is. 

if the process of provisioning the golden snapshot is idempotent, maybe the controllr doesn't need to stage its current step? We might need some way to indicate if the process failed though, but that could be a separate field.

cmd/ateapi/internal/controlapi/actor_snapshot_test.go
```diff
@@ -526,7 +526,7 @@ func TestUpdateActorSnapshotTag_RejectsUnknownFields(t *testing.T) {
 			in := proto.Clone(stored).(*ateapipb.ActorSnapshotTag)
 			tt.injectUnknownField(in)
 
-			_, err := svc.UpdateActorSnapshotTag(ctx, &ateapipb.UpdateActorSnapshotTagRequest{Tag: in})
+			_, err := svc.UpdateActorSnapshotTag(ctx, &ateapipb.UpdateActorSnapshotTagRequest{ActorSnapshotTag: in})
```
zoez7, line 530: Self review: this fixes existing unit test failurs.

cmd/ateapi/internal/controlapi/actor_template_test.go
```diff
@@ -218,8 +218,12 @@ func TestCreateActorTemplateIgnoresServerOwnedFields(t *testing.T) {
 		tmpl.Resources = &ateapipb.Resources{Limits: []*ateapipb.Limits{{Name: "memory", Quantity: "1Gi"}}}
 		// Server-owned status a client must not be able to set.
 		tmpl.Status = &ateapipb.ActorTemplateStatus{
-			Phase:          ateapipb.ActorTemplatePhase_ACTOR_TEMPLATE_PHASE_READY,
-			Message:        "forged",
+			Conditions: []*ateapipb.Condition{{
```
zoez7: Self review: this is to ensure clients can not set this field.

cmd/ateapi/internal/controlapi/template_reconciler.go
```diff
@@ -208,73 +213,126 @@ func (r *ActorTemplateReconciler) reconcileOne(ctx context.Context, ref resource
...
-				if r.goldenCrashed(ctx, goldenRef) {
-					return 0, r.fail(ctx, ref, phase, fmt.Sprintf("golden actor crashed while booting: %v", err))
+		switch state := actor.GetStatus().GetState(); state {
+		case ateapipb.ActorState_ACTOR_STATE_CRASHED:
+			return 0, r.fail(ctx, ref, reasonGoldenActorCrashed, "golden actor crashed before its snapshot was taken")
+
+		case ateapipb.ActorState_ACTOR_STATE_RUNNING:
+			takeAt := tmpl.GetStatus().GetTakeGoldenSnapshotAt()
+			if takeAt == nil {
+				// Resumed, but the deadline write was lost (a replica died
```
zoez7, line 223: Self review- this means the golden actor snapshot will be taken at a later time.

pkg/proto/ateapipb/ateapi.proto
zoez7: Updated to 
1. Using Conditions instead of Phases as described in https://screenshot-v2.corp.google.com/6ctmlmia4ruh0
2. Removed intermediate conditions, and only have "Ready" and "Failed" conditions.

cmd/ateapi/internal/controlapi/template_reconciler.go
```diff
@@ -0,0 +1,367 @@
...
+	tmpl, err := r.persistence.GetActorTemplate(ctx, ref)
+	if err != nil {
+		// Template was deleted after we enqueued the work item.
+		if errors.Is(err, store.ErrNotFound) {
+			return 0, nil
+		}
+		return 0, err
+	}
+
+	goldenRef := &ateapipb.ObjectRef{
```
juli4n: Nit: maybe call this goldenActorRef? Golden is kind of overloaded as there is a golden actor and a golden snapshot.

cmd/ateapi/internal/controlapi/template_reconciler.go
```diff
@@ -0,0 +1,367 @@
...
+	// fact demands, and re-observes; the pass ends at a terminal condition,
+	// a deadline wait, or an error the workqueue retries.
+	for {
+		conds := tmpl.GetStatus().GetConditions()
+		if conditionIsTrue(conds, conditionReady) || conditionIsTrue(conds, conditionFailed) {
+			return 0, nil
+		}
+
+		// TODO: Freeze sandbox assets before creating the golden actor.
+		actor, err := r.control.GetActor(ctx, &ateapipb.GetActorRequest{Actor: goldenRef})
```
juli4n: Can this logic be abstracted into a `ensureActor() (*Actor, err)` method? That would simplify this method a bit.

cmd/ateapi/internal/controlapi/template_reconciler.go
```diff
@@ -0,0 +1,367 @@
...
+	"k8s.io/client-go/util/workqueue"
+)
+
+const (
+	templateResyncInterval = 5 * time.Second
+	templateListPageSize   = 100
+
+	// templateWorkerCount is the number of goroutines draining the work
+	// queue.
+	templateWorkerCount = 2
```
juli4n: Because ResumeActor is blocking on the readyz probe, this means that worst case we cannot resume more than 1 actor every 15 seconds? It sounds like this would be a significant bottleneck, should we set this to a larger value?

cmd/ateapi/internal/controlapi/template_reconciler.go
```diff
@@ -0,0 +1,367 @@
...
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/timestamppb"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/workqueue"
+)
+
+const (
+	templateResyncInterval = 5 * time.Second
```
juli4n: Isn't this too frequent? How about starting with 15 / 20 seconds?

cmd/ateapi/internal/controlapi/template_reconciler.go
zoez7: Done in 9b3ef02ee9678ec1656ace8cee96b5b3c0eba5f2.

cmd/ateapi/internal/controlapi/template_reconciler.go
zoez7: Done.

cmd/ateapi/internal/controlapi/template_reconciler.go
zoez7: I increased this to 5 workers. With 2 ateapi replicas, we can have 10 ATs being reconciled at a given time.

cmd/ateapi/internal/controlapi/template_reconciler.go
zoez7: Done.

juli4n (dismissed):

PR #1097, opened 2026-08-20
observability: add a metric registry that CI checks

### LINKED ISSUE #1093: observability: adopt OpenTelemetry Weaver metric registry for system metrics
## Summary
  
The project now has 20 metric. Six components send them. Only the text in `docs/observability.md` describes these instruments. No check compares that text with the code.

This FR is for a metric registry. The registry must be the single source of truth for the system metrics. It must also have a machine-readable form. A check must compare the registry with the code on each build.

I propose to use the OpenTelemetry Weaver format for semantic conventions.

### For AI agents

An agent that finds faults must query the metrics. To do this, the agent must know three things: the correct instrument names, the correct label keys, and the permitted values of each label.

### For developers

A developer declares an instrument in the code. A different document describes that instrument. A dashboard queries it. Nothing connects the three.

JeffLuoo:
`docs/observability.md` was the only description of the metrics of Substrate, and nothing compared that text with the code. It listed 13 instruments; the code sends 21. The request-parking and actor resource-usage instruments were
missing entirely.

This change adds an [OpenTelemetry Weaver](https://github.com/open-telemetry/weaver) semantic convention registry.

  ### What is in it
  
  | Path | Content |
  |---|---|
  | `docs/metrics/registry/manifest.yaml` | The Weaver manifest: name and schema URL. |
  | `docs/metrics/registry/metrics.yaml` | 21 instruments and 12 attribute groups: every label, its permitted values, and its buckets. |
  | `docs/metrics/substrate.yaml` | The rules Weaver cannot express. |
  | `hack/verify/verify-metrics.sh` | The check. |

Fixes #1093

- [x] Tests pass

### DIFF
--- a/docs/metrics/registry/manifest.yaml
+++ b/docs/metrics/registry/manifest.yaml
@@ -1,0 +1,28 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The OpenTelemetry Weaver manifest of this registry. Weaver reads it when you
+# run `weaver registry check -r docs/metrics/registry`.
+#
+# schema_url holds the version of this registry, and not the version of the
+# upstream semantic conventions. docs/metrics/substrate.yaml records the
+# upstream version, because Substrate borrows attributes from it.
+#
+# The URL is an identifier. Nothing is published there at this time. ate.dev is
+# the domain of the API group of Substrate, thus it is the correct place if the
+# project publishes this registry later.
+
+name: substrate
+description: The metric semantic conventions of Agent Substrate.
+schema_url: https://ate.dev/schemas/substrate/1.0.0
--- a/docs/metrics/registry/metrics.yaml
+++ b/docs/metrics/registry/metrics.yaml
@@ -1,0 +1,1187 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file is an OpenTelemetry Weaver semantic convention registry. It defines
+# every metric instrument that the ate system components send, and the
+# permitted values of each label. Use it as the only source of this data.
+#
+# `weaver registry check -r docs/metrics/registry` validates this file.
+# `hack/verify/verify-metrics.sh` runs that command in CI.
+#
+# Weaver permits only `groups` and `imports` at the top level. Thus the rules
+# that Weaver cannot hold live in docs/metrics/substrate.yaml. That file is
+# outside this directory on purpose: Weaver reads each YAML file in the
+# registry directory and refuses a file with an unknown top-level key.
+#
+# The metrics section of docs/observability.md tells you how to use this file.
+
+groups:
+  # ---------------------------------------------------------------------------
+  # Attribute groups
+  # ---------------------------------------------------------------------------
+
+  - id: registry.ate.actor
+    type: attribute_group
+    brief: The labels of an actor.
+    attributes:
+      - id: ate.actor.operation.name
+        stability: development
+        brief: The actor lifecycle operation that ateapi did.
+        type:
+          members:
+            - id: create
+              stability: development
+              value: create
+              brief: ateapi made the actor record.
+            - id: resume
+              stability: development
+              value: resume
+              brief: ateapi put the actor on a worker.
+            - id: suspend
+              stability: development
+              value: suspend
+              brief: ateapi wrote the durable snapshot and released the worker.
+            - id: pause
+              stability: development
+              value: pause
+              brief: ateapi wrote a node-local snapshot.
+            - id: delete
+              stability: development
+              value: delete
+              brief: ateapi removed the actor record.
+            - id: unknown
+              stability: development
+              value: unknown
+              brief: The operation is not in this list. This is the default value.
+
+  - id: registry.ate.template
+    type: attribute_group
+    brief: >
+      The identity of an ActorTemplate. The values are not limited to a list.
+      But an operator makes each template. No value comes from a request. Thus
+      the number of values stays small.
+    attributes:
+      - id: ate.template.namespace
+        stability: development
+        type: string
+        brief: The Kubernetes namespace of the ActorTemplate of the actor.
+        examples: [ate-demo-counter]
+      - id: ate.template.name
+        stability: development
+        type: string
+        brief: The name of the ActorTemplate of the actor.
+        examples: [counter]
+
+  - id: registry.ate.workerpool
+    type: attribute_group
+    brief: >
+      The identity of a WorkerPool. The two keys identify a pool together. A
+      component sets both keys or no key. A WorkerPool has a namespace. Thus the
+      name alone puts the pools of different namespaces into the same series.
+    attributes:
+      - id: ate.workerpool.namespace
+        stability: development
+        type: string
+        brief: The Kubernetes namespace of the WorkerPool.
+        examples: [ate-demo-counter]
+      - id: ate.workerpool.name
+        stability: development
+        type: string
+        brief: The name of the WorkerPool.
+        examples: [counter]
+      - id: ate.worker.state
+        stability: development
+        brief: >
+          The assignment state of a worker. The key starts with worker and not
+          with the name of the pool. Thus you can add related keys later.
+        type:
+          members:
+            - id: idle
+              stability: development
+              value: idle
+              brief: No actor has this worker.
+            - id: assigned
+              stability: development
+              value: assigned
+              brief: An actor has this worker.
+
+  - id: registry.ate.sandbox
+    type: attribute_group
+    brief: >
+      The class of the sandbox. The class comes from the template. Thus it adds
+      no series to the template labels. It lets a dashboard group the data by
+      class without a list of the template names.
+    attributes:
+      - id: ate.sandbox.class
+        stability: development
+        brief: The sandbox runtime of the actor.
+        type:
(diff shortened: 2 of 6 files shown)


git286 (changes requested):
1. The scanner only picks up instrument names that are string literals or same-package consts. If someone creates an instrument through a helper the name is invisible to the check, and an undeclared instrument passes silently.  Could the tool fail (or at least warn) when it sees an instrument constructor whose name argument it can't resolve?Then the gap is loud instead of silent, and the author either inlines the name or adds it to the registry.

2. Can we add tests for tools/metricsregistry/main.go?

JeffLuoo:
> The scanner only picks up instrument names that are string literals or same-package consts. If someone creates an instrument through a helper the name is invisible to the check, and an undeclared instrument passes silently. Could the tool fail (or at least warn) when it sees an instrument constructor whose name argument it can't resolve?Then the gap is loud instead of silent, and the author either inlines the name or adds it to the registry.

Good catch, updated it new patch. Any instrument constructor whose name argument the scanner cannot read is reported with the file, the line, the constructor and the expression.

> Can we add tests for tools/metricsregistry/main.go?

Updated.

docs/metrics/registry.yaml
```diff
@@ -0,0 +1,1187 @@
...
+    stability: stable
+    brief: The time, the rate and the errors of each gRPC method on the server.
+    note: >
+      The otelgrpc interceptor sends this instrument. The code of Substrate does
+      not make it. Thus the check does not compare it with the Go code. This
+      instrument is the only instrumentation of ateom. ateom has it on the gRPC
+      interface that atelet calls.
+    annotations:
+      substrate:
+        emitted_by: [ateapi, atelet, ateom-gvisor, ateom-microvm]
```
krisztianfekete: I think we are missing:
 - atenet-router as amitted_by
 - rpc.client.call.duration metric
 - the bridged controller_runtime_* / workqueue_* / go_* / process_* metrics

docs/metrics/registry.yaml
```diff
@@ -0,0 +1,1187 @@
...
+# Weaver does not read this file at this time. tools/metricsregistry does all
+# the checks.
+#
+# The metrics section of docs/observability.md tells you how to use this file.
+
+registry_version: 1
+
+# The version of the upstream semantic conventions. Substrate takes error.type,
+# file.name and the rpc.* attributes from this version.
+semconv_version: 1.44.0
```
krisztianfekete: Let's stick with v1.40.0, we are using that in code.

docs/metrics/registry/metrics.yaml
```diff
@@ -0,0 +1,1187 @@
...
+        golden_signals: [saturation]
+        code_anchor: cmd/atelet/main.go
+        cuj: The snapshots become larger. Which template, and which image?
+    attributes:
+      - ref: file.name
+        requirement_level: required
+      - ref: ate.template.namespace
+        requirement_level: required
+      - ref: ate.template.name
+        requirement_level: required
```
krisztianfekete, line 855: Looks like we are missing the buckets from this one?

docs/metrics/registry.yaml
```diff
@@ -0,0 +1,1187 @@
...
+# the first rule. The other rules are here so that a reviewer and an agent read
+# the same contract.
+# -----------------------------------------------------------------------------
+cardinality_rules:
+  - id: no-actor-identity
+    brief: >
+      The name, the UID and the atespace of an actor are never metric labels.
+      They have too many values at the target of 1 billion actors. Put them on
+      spans and logs. The trace_id field connects a log to a metric.
+    enforced: true
```
krisztianfekete: Not sure if this can fire at all. Weaver probably recommends a Rego/similar policy for this, so maybe we can turn `enforced` off?

docs/metrics.json
krisztianfekete, line 1: Can we delete this? registry.yaml should be enough.

krisztianfekete (changes requested):
There's a lot of value in this registry, thanks! Not fully sure about the ~3.5k lines of code we are also adding. 

There's a "Go AST scanner" that is very close to what `weaver registry live-check` does natively, and we're handcrafting stuff that `weaver registry generate` could emit straight into ateattr. 

Can we move the five substrate top-level keys into a sibling file so `weaver registry check` runs today. Can we land the YAML and skip the tool for now?

EItanya (dismissed):
I think it's important we address @krisztianfekete's comments before we move forward with this. Placing requires changes as he doesn't have "write" yet.

docs/metrics/registry.yaml
JeffLuoo: good catch. updated to include missing three components

docs/metrics/registry.yaml
JeffLuoo: updated

docs/metrics/registry/metrics.yaml
JeffLuoo: added

docs/metrics/registry.yaml
JeffLuoo: updated

docs/metrics.json
JeffLuoo: Deleted

For the record on the precedent: `docs/threats.json` is generated from `docs/threat-model.md`, checked in, verified by `hack/verify/verify-threats.sh`, and read by the security-status-report skill. So the pattern exists here.

We can work on this as a follow-up if we need to follow the similar pattern later.

JeffLuoo:
> Can we move the five substrate top-level keys into a sibling file so `weaver registry check` runs today. Can we land the YAML and skip the tool for now?

 A sibling file in the same directory does not work. Weaver reads every YAML file in the registry directory and refuses any top-level key other than groups and imports, I see errors like:
 ```
 Object contains unexpected properties: cardinality_rules
 ```
 
 Hence, the directory looks like:
 ```
docs/metrics/registry/manifest.yaml   name + schema_url
docs/metrics/registry/metrics.yaml    groups only
docs/metrics/substrate.yaml           the substrate keys, one level up
 ```

hack/verify/verify-metrics.sh
```diff
@@ -0,0 +1,73 @@
...
+  "${DOCKER}" run \
+    --rm \
+    --user "$(id -u):$(id -g)" \
+    --volume "${ROOT}:/workspace:ro" \
+    --workdir /workspace \
+    "${WEAVER_IMAGE}" \
+    "$@"
+}
+
+if ! run_weaver registry check --registry "${REGISTRY_DIR}"; then
```
krisztianfekete: Can we add `--future`. Right now a duplicate metric_name is only a warning and the script exits 0, it exists with 1 with this flag.

hack/verify/verify-metrics.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Allow overriding the docker CLI, as hack/third_party/kubernetes does.
+DOCKER="${DOCKER:-docker}"
+
+run_weaver() {
+  if command -v weaver >/dev/null 2>&1; then
+    echo "Using the local weaver binary: $(weaver --version)"
+    weaver "$@"
+    return
+  fi
```
krisztianfekete, line 60: Can we ensure we're running the same version both locally and in CI?

docs/metrics/registry/metrics.yaml
```diff
@@ -0,0 +1,1637 @@
...
+  # atecontroller gives the controller-runtime Prometheus registry to the OTLP
+  # pipeline at cmd/atecontroller/main.go:113. Thus these series leave the
+  # cluster with the ate.* metrics. The code of Substrate does not make them.
+  # The names keep the Prometheus form with underscores.
+  #
+  # This list is not complete. The same registry also holds the full Go runtime
+  # set (go_*) and the process set (process_*). There are approximately 100 of
+  # them and none is specific to Substrate. docs/metrics/substrate.yaml records
+  # them as a family.
+  # ---------------------------------------------------------------------------
```
krisztianfekete: Can we collapse these into `bridged_metric_families` same way as `go_*` and `process_*`?

I gathered `crmetrics.Registry` and it's already missing four registered families: 
 - certwatcher_read_certificate_total,
 - certwatcher_read_certificate_errors_total,
 - controller_runtime_webhook_panics_total,
 - controller_runtime_conversion_webhook_panics_total

so it's really error prone to maintain these manually, but handling these families this way can help I think. Also if we add `controller_runtime_version: v0.24.1` key next to `upstream_semconv_version` we can try to re-read each time we bump it.

docs/metrics/registry/manifest.yaml
```diff
@@ -0,0 +1,24 @@
...
+# The OpenTelemetry Weaver manifest of this registry. Weaver reads it when you
+# run `weaver registry check -r docs/metrics/registry`.
+#
+# schema_url holds the version of this registry, and not the version of the
+# upstream semantic conventions. docs/metrics/substrate.yaml records the
+# upstream version, because Substrate borrows attributes from it.
+
+name: substrate
+description: The metric semantic conventions of Agent Substrate.
+schema_url: https://raw.githubusercontent.com/agent-substrate/substrate/main/docs/metrics/registry/1.0.0
```
krisztianfekete: Nit: do we want to do something like https://ate.dev/schemas/substrate/1.0.0 where we publish this registry at some point? Even if we don't want to do that, at least it's not a 404.

hack/verify/verify-metrics.sh
JeffLuoo: Added

hack/verify/verify-metrics.sh
JeffLuoo: Pin WEAVER_VERSION in this file

docs/metrics/registry/metrics.yaml
JeffLuoo: Agreed, collapsed

docs/metrics/registry/manifest.yaml
JeffLuoo: updated

krisztianfekete (approved):
Awesome, we can update the PR desc to match the new approach, but otherwise LGTM!

git286 (approved):

JeffLuoo:
@EItanya Can you PTAL of the change as well? Thanks

git286 (approved):

EItanya:
Oops sorry, LGTM

hack/verify/verify-metrics.sh
```diff
@@ -0,0 +1,93 @@
...
+set -o errexit -o nounset -o pipefail
+
+ROOT="$(git rev-parse --show-toplevel)"
+cd "${ROOT}"
+
+REGISTRY_DIR="docs/metrics/registry"
+
+# Keep the version in sync with the digest below.
+WEAVER_VERSION="v0.25.1"
+WEAVER_IMAGE="docker.io/otel/weaver:${WEAVER_VERSION}@sha256:9ad46ca9cd4fa5974b121f886aa3e9946a8ef8ea905001a96c018d21f9db87ca"
```
BenTheElder, line 33: nit: There's no point interpolating the tag here, the digest is authoritative and the tag will be discarded.

it also might confuse the next person editing this into thinking that they can just update the version above.

doing this sometimes makes sense for handing out a reference to use, since it's "documentation" but in this case where we're auto-generating it .... seems counter productive

JeffLuoo: I will work on a follow-up PR

PR #1098, opened 2026-08-20
Fix upstream mtls cert rotation

### LINKED ISSUE #1066: atenet-router: upstream mTLS client cert is never rotated; expiry breaks all actor routing until pod restart
On a long-lived dev cluster, a ~2-day-old atenet-router pod started failing every actor-bound request with:

```
503: upstream connect error or disconnect/reset before headers. reset reason: remote connection failure,
transport failure reason: TLS_error:|268436501:SSL routines:OPENSSL_internal:SSLV3_ALERT_CERTIFICATE_EXPIRED
```

atunnel was rejecting Envoy's client certificate. `kubectl -n ate-system rollout restart deployment/atenet-router` fixed it immediately (worker pods untouched), confirming the stale credential was on the router side.

**Root cause:** `buildUpstreamTransportSocket()` (`cmd/atenet/internal/router/xds.go`) configures the podidentity credential bundle as inline filename `DataSource`s in a static `UpstreamTlsContext`. Envoy reads filename-based `TlsCertificates` once at cluster load and never re-reads them, so kubelet's rotation of the projected `credential-bundle.pem` never reaches Envoy — the presented cert goes stale and eventually expires.

**Fix shape:** deliver the upstream client cert the same way the ingress serving cert is delivered — an SDS secret with `WatchedDirectory`. #531 fixed this exact bug class for the serving cert ("Otherwise, envoy's serving certificate expires after 24hours"); the upstream client cert surface was missed and needs the same treatment.

Found while running the ingress capacity benchmark (#998): the benchmark's actor warm-up 503s on any dev cluster whose router pod outlives the cert TTL.

skamboj:
Fixes #1066 

- [x] Tests pass
- [x] ~~Appropriate changes to documentation are included in the PR~~ No documentation changes needed.

## Summary

- Move actor-facing Envoy upstream mTLS credentials from inline file `DataSource`s to SDS secrets served over ADS.
- Publish separate SDS secrets for the upstream podidentity client cert and upstream trust bundle.
- Preserve SPIFFE URI SAN validation by using a combined validation context: trust bundle via SDS, SAN matcher inline.
- Reuse the ADS config source helper for downstream TLS SDS references.
- Add tests covering upstream SDS references, snapshot secret publication, and projected-volume rotation behavior.

### DIFF
--- a/cmd/atenet/internal/router/xds.go
+++ b/cmd/atenet/internal/router/xds.go
@@ -81,6 +81,15 @@ const (
 	OtlpClusterName      = "otel_collector_cluster"
 	HTTPSCertSecretName  = "https_serving_cert"
 
+	// UpstreamCertSecretName and UpstreamTrustSecretName are the SDS secrets
+	// backing the actor-facing mTLS transport socket: the podidentity client
+	// cert the router presents to atunnel, and the CA bundle it validates
+	// atunnel with. Both are delivered over SDS rather than inlined into the
+	// transport socket so that Envoy re-reads them after kubelet rotates the
+	// projected volume (see buildCertSecret).
+	UpstreamCertSecretName  = "upstream_client_cert"
+	UpstreamTrustSecretName = "upstream_trust_bundle"
+
 	// httpProtocolOptionsName is the well-known extension key Envoy looks for in
 	// a cluster's typed_extension_protocol_options. It must match the message's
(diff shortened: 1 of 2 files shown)


chuangw6 (approved):

chuangw6:
@skamboj can you take a look at failed tests?

PR #1101, opened 2026-08-20
fix(atunnel): close both relay ends

### LINKED ISSUE #1100: E2E test flake: TestRelayIngressCancellationClosesBothSides
`TestRelayIngressCancellationClosesBothSides` fails intermittently on unrelated branches:

- https://github.com/agent-substrate/substrate/actions/runs/32270460698 (`yelshall/workerpool-pod-metadata`)
- https://github.com/agent-substrate/substrate/actions/runs/32289001614 (`logs-source-filters`)
- https://github.com/agent-substrate/substrate/actions/runs/32311722045 (`nighthawk-router-capacity`)
- https://github.com/agent-substrate/substrate/actions/runs/32325183622 (`kind-ipv6-coredns`)

```
    ingress_test.go:123: client stream remained open after relay cancellation
--- FAIL: TestRelayIngressCancellationClosesBothSides (0.00s)
```

## Expected Behavior

The test asserts that cancelling the request context closes both sides of a CONNECT relay.

## Actual Behavior

The assertion at `ingress_test.go:123` races the close it is checking for.

`relayIngressWithHalfClose` closes both sides from a `context.AfterFunc` callback, which runs on its own goroutine. `cancel()` returning says nothing about whether that goroutine has run. The test asserts immediately afterwards.

`clientReader`/`clientInput` is an unbuffered `io.Pipe`, so the test's `Write` is claimed by whichever goroutine reaches it first:

- the `AfterFunc` goroutine closing `clientCloser` — `Write` fails, test passes
- the still-running `io.Copy(upstream, clientReader)` — its `Read` takes the bytes, `Write` returns nil, test fails

The preceding `actor.Read` assertion is much harder to hit because it carries a one-second read deadline, which gives the close a full second of slack. The `Write` has none, which is why the failure is always line 123.

```
go test ./internal/atunnel/ -race -run TestRelayIngressCancellationClosesBothSides -count=10000
```

Measured on two platforms, 10000 iterations per cell:

| | Linux/x86_64, 4 cores (same as CI) | macOS/arm64, 10 cores |
|---|---:|---:|
| plain | 1 failure | 8 failures |
| `-race` | **89 failures** | **202 failures** |

`-race` widens the window by two orders of magnitude, and CI runs `go test -race -v ./...`, which is the path these runs fail on. 89/10000 is consistent with 4 hits across the last 70 failed `pr-workflow` runs.

## Specifications

- Version: `origin/main` @ 3eaf6c2f
- Platform: Linux 6.17.0-1022-azure x86_64, Go 1.26.3; also reproduced on darwin/arm64

orangeCatDeveloper:
Fixes #1100

`TestRelayIngressCancellationClosesBothSides` fails on about 1% of `-race` runs, on branches unrelated to the relay.

## Root cause

The test asserts a close that happens on a goroutine it never waits for.

`relayIngressWithHalfClose` closed both sides from a `context.AfterFunc` callback, which runs on its own goroutine. `cancel()` returning therefore says nothing about whether the streams are closed — and the test asserted right after it.

The client side is an unbuffered `io.Pipe`, where a `Write` cannot complete on its own; it needs someone on the other end. Two goroutines can be that someone:

| Whichever gets there first | Outcome |
|---|---|
| the `AfterFunc` callback closes the pipe | `Write` fails → **test passes** |
| the relay's `io.Copy`, still parked in `Read` | it takes the bytes, `Write` returns nil → **test fails** |

The scheduler picks the winner, so the test is a coin flip. `-race` widens the window by two orders of magnitude — 1 failure per 10000 plain, 89 with `-race` — and CI runs `go test -race -v ./...`.

The `actor.Read` assertion just above races the same close, but carries a one-second deadline and so almost always wins. The `Write` has no such slack, which is why the failure is always line 123.

## Fix

**Production.** Close both sides in the `ctx.Done()` branch the relay loop already selects on, rather than from a detached goroutine. This turns

> both sides are *eventually* closed

into

> both sides are closed *before this returns*

which is what a caller can build on, and it drops a goroutine plus its `stop()` bookkeeping.

**Test.** Wait for the relay to return, then assert. The `SetReadDeadline` error check goes with it: once the relay has closed the pipe, setting a deadline on it fails by design, and the deadline was only ever a guard against hanging if the connection had been left open.

Nothing changes about the bytes the relay moves — this is a shutdown-ordering fix, not a data-plane one.

## Before / after

```
$ go test ./internal/atunnel/ -race -run TestRelayIngressCancellationClosesBothSides -count=10000
--- FAIL: TestRelayIngressCancellationClosesBothSides (0.00s)
    ingress_test.go:123: client stream remained open after relay cancellation
... 89 of 10000 iterations
FAIL	github.com/agent-substrate/substrate/internal/atunnel

$ go test ./internal/atunnel/ -race -run TestRelayIngressCancellationClosesBothSides -count=10000
ok  	github.com/agent-substrate/substrate/internal/atunnel	3.035s
```

Linux/x86_64, 4 cores, matching CI. On darwin/arm64 the same runs give 202 failures before and 0 after.

- [x] Tests pass

### DIFF
--- a/internal/atunnel/ingress.go
+++ b/internal/atunnel/ingress.go
@@ -309,12 +309,6 @@ func (s *Server) serveH2Connect(w http.ResponseWriter, r *http.Request, upstream
 // stream ends, it half-closes the actor connection and continues forwarding
 // actor output until that stream ends too.
 func relayIngressWithHalfClose(ctx context.Context, upstream net.Conn, clientReader io.Reader, clientWriter io.Writer, clientCloser io.Closer) {
-	stop := context.AfterFunc(ctx, func() {
-		_ = upstream.Close()
-		_ = clientCloser.Close()
-	})
-	defer stop()
-
 	done := make(chan struct{}, 2)
 	go func() {
 		_, _ = io.Copy(upstream, clientReader)
@@ -328,6 +322,8 @@ func relayIngressWithHalfClose(ctx context.Context, upstream net.Conn, clientRea
 	for range 2 {
 		select {
 		case <-ctx.Done():
+			_ = upstream.Close()
+			_ = clientCloser.Close()
 			return
 		case <-done:
 		}
--- a/internal/atunnel/ingress_test.go
+++ b/internal/atunnel/ingress_test.go
@@ -113,19 +113,18 @@ func TestRelayIngressCancellationClosesBothSides(t *testing.T) {
 	}()
 
 	cancel()
-	if err := actor.SetReadDeadline(time.Now().Add(time.Second)); err != nil {
-		t.Fatal(err)
+	select {
+	case <-done:
+	case <-time.After(time.Second):
+		t.Fatal("relay did not return after cancellation")
 	}
+	// Deadline only guards against hanging; it fails once the relay closed the pipe.
+	_ = actor.SetReadDeadline(time.Now().Add(time.Second))
 	if _, err := actor.Read(make([]byte, 1)); err == nil {
 		t.Fatal("actor connection remained open after relay cancellation")
 	}
 	if _, err := io.WriteString(clientInput, "request"); err == nil {
 		t.Fatal("client stream remained open after relay cancellation")
-	}
-	select {
-	case <-done:
-	case <-time.After(time.Second):
-		t.Fatal("relay did not return after cancellation")
 	}
 }
 


keithmattix (approved):
Thanks for the fit

aditya-shantanu (approved):
Verified the analysis and the fix.

- The race is exactly as described: the `context.AfterFunc` closes ran on a detached goroutine, so `cancel()` returning guaranteed nothing about stream state, and the unbuffered `io.Pipe` write in the test raced the relay's `io.Copy` against the close. Moving the closes into the `ctx.Done()` branch of the relay loop makes "both sides closed before return" an actual postcondition, and the copy goroutines are still unblocked by those same closes (the `done` channel is buffered, so no leak).
- Behavior on the normal completion path is unchanged — the relay still doesn't close either side when both copies finish, same as before with `defer stop()`.
- Ran `go test -race ./internal/atunnel/ -run TestRelayIngressCancellationClosesBothSides -count=5000` locally (darwin/arm64): green. Also green with this branch merged into current main, `-race -count=50` over the relay/serve tests.

LGTM.

aditya-shantanu:
Heads-up: adopted this fix (unchanged, with Co-authored-by credit) into the consolidated flake-fix PR #1160 to get CI green faster. If maintainers prefer landing this PR individually, #1160 can drop the commit or be closed — whatever lands first wins.

Issue #1089, opened 2026-08-20
Actor egress has no fallback when a destination family has no path


ygao-g:
Part of #246.

A worker pod can hold a global unicast address in a family that has no path off the node. Dual-stack kind is the case in hand: pods get `fd00:10:244::/64`, a ULA with no route out, and `curl -6` fails from the node itself. #1057 gates the actor's IPv6 on the pod having a global IPv6 address, which is correct and still says yes here — `IsGlobalUnicast()` is true for `fc00::/7`, and no addressing check can tell a ULA from a routable prefix.

The actor then picks the AAAA and the connection is accepted locally rather than refused, because the egress REDIRECT makes every destination look reachable. It dies mid-stream instead of failing to connect, so Happy Eyeballs never falls back to IPv4. Refusing at accept time when the destination's family has no path would turn that into an ordinary connect failure and let the fallback work.

This is what blocks the `dual` axis of the IP-family e2e matrix: `TestActorEgress` and `TestActorEgressHTTPS` fail on a dual-stack kind cluster today, so adding the axis lands two red tests on day one.

PR #1104, opened 2026-08-20
[WIP] e2e: reach an in-cluster destination over each address family

ygao-g:
The egress suite fetches example.com, so which address families a test exercises is decided by
whatever resolver the cluster inherited. When the name has no AAAA the actor never attempts IPv6 and
a broken IPv6 path passes; when it does, the same test fails. Neither outcome says anything about
the system, and neither single-family lane can notice.

This adds an in-cluster destination fronted by Services pinned to each family, so which families are
in play is a property of the test, and asserts the family each request actually arrived over rather
than trusting a 200. The dual-homed case is the one no single-family cluster can construct: an actor
that has an IPv6 address prefers the AAAA, so that destination becomes unreachable the moment IPv6
egress breaks, even with a working A record right beside it and every IPv4-only destination still
working. Measured on a dual-stack kind cluster it passes on main and fails with #979 and #1057
applied, which is #1089. On a single-family cluster the cases it cannot build are skipped. Part of
#246.

### DIFF
--- a/internal/e2e/echodest.go
+++ b/internal/e2e/echodest.go
@@ -1,0 +1,118 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package e2e
+
+import (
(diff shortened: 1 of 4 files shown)


aspire488:
I took a close look at this against #1089. The in-cluster destination approach is useful, but I think there is one important gap in the dual-stack assertion: the "dual" case currently has "want == """, so it only proves that the request eventually returns a 200.

It does not prove that the dual destination actually exercised the IPv6-preferred path or, more importantly for #1089, that an unreachable IPv6 path causes a connection-level failure and Happy Eyeballs can fall back to the IPv4 address. A regression that silently forces IPv4 for dual-stack destinations could therefore pass this test.

I would make the dual case explicitly verify the failure/fallback behavior (or otherwise inject/construct the no-route IPv6 condition) rather than leaving the family unconstrained. That would make the test directly encode the bug in #1089 instead of only adding family coverage.

The single-family cases are still valuable as baseline coverage.

Issue #1106, opened 2026-08-20
Race condition in Ateom startup and TestActorIdentity_AfterRestore_IsOwnID_NotGolden Test Flake


Stevenjin8:
## Expected Behavior

TestActorIdentity_AfterRestore_IsOwnID_NotGolden pass consistently https://github.com/agent-substrate/substrate/actions/runs/32402609221/job/96534114102

## Actual Behavior

TestActorIdentity_AfterRestore_IsOwnID_NotGolden  does not 

## Steps to Reproduce the Problem

This is pretty hard to repro, but I have ateom startup with a random delay between 0 to 60 seconds before it listens on a unix socket.
Then the issue appears about 1/3-ish of the time.
If we get lucky (or unlucky), substrate will create a golden actor on an ateom and then create an actor and schedule it on an ateom that is not yet ready (inside the wait)

The issue is that we check whether an ateom is ready by whether the pod has an address, but there could be time between the pod is given an address and from when it listening on the unix socket.

## Specifications

- Version:
- Platform:

aditya-shantanu:
No failures for TestActorIdentity_AfterRestore_IsOwnID_NotGolden since 66300d5c (#1243) across 56 e2e-gvisor and 46 e2e-microvm runs in the last 7 days — at the prior ~11% gvisor flake rate, 56 clean runs is ~0.1% probability by chance. Marking both lanes resolved on the flake dashboard. Note: the separately-observed post-restore container-readyz timeout (run 32891995583) is a different failure layer not claimed by #1243; it gets its own issue if it recurs.

PR #1107, opened 2026-08-20
check readiness for scheduling and fix TestActorIdentity_AfterRestore_IsOwnID_NotGolden flake

Stevenjin8:
Fixes # https://github.com/agent-substrate/substrate/issues/1106

Add ateom readiness and require readiness to schedule actor.

Note that this readiness isn't fully correct as we start the readiness server just before the main server. The proper solution is to add a bunch of plumbing that makes the readiness server only return ready after calling `Accept`. I think this will be very much good enough to fix `TestActorIdentity_AfterRestore_IsOwnID_NotGolden`

- [ ] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/syncer_test.go
+++ b/cmd/ateapi/internal/controlapi/syncer_test.go
@@ -1,0 +1,1557 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controlapi
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"maps"
+	"strconv"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
+	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store/storetest"
+	"github.com/agent-substrate/substrate/internal/ateattr"
+	"github.com/agent-substrate/substrate/internal/resources"
+	atev1alpha1 "github.com/agent-substrate/substrate/pkg/api/v1alpha1"
+	atefake "github.com/agent-substrate/substrate/pkg/client/clientset/versioned/fake"
+	"github.com/agent-substrate/substrate/pkg/client/informers/externalversions"
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+// testPodUID is the pod UID most syncer tests give their single worker pod.
+// The syncer names Workers after their pod UID, so it is also the name the
+// resulting store row is keyed by.
+const testPodUID = "11111111-1111-1111-1111-111111111111"
+
+// setupSyncerTest sets up a real PostgreSQL store and a fake K8s client with informer.
+func setupSyncerTest(t *testing.T, ctx context.Context, initPools ...*atev1alpha1.WorkerPool) (store.Interface, *fake.Clientset, *atefake.Clientset, func()) {
+	persistence, fakeK8s, fakeAte, _, cleanup := setupSyncerTestWithStore(t, ctx, nil, initPools...)
+	return persistence, fakeK8s, fakeAte, cleanup
+}
+
+func setupSyncerTestWithStore(t *testing.T, ctx context.Context, wrapStore func(store.Interface) store.Interface, initPools ...*atev1alpha1.WorkerPool) (store.Interface, *fake.Clientset, *atefake.Clientset, *WorkerPoolSyncer, func()) {
+	t.Helper()
+
+	persistence, cleanup := storetest.SetupTestStore(t)
+	if wrapStore != nil {
+		persistence = wrapStore(persistence)
+	}
+
+	fakeK8s := fake.NewSimpleClientset()
+	workerFactory, workerInformer := WorkerPodInformer(fakeK8s)
+
+	objects := make([]runtime.Object, len(initPools))
+	for i, pool := range initPools {
+		objects[i] = pool
+	}
+	//nolint:staticcheck // NewSimpleClientset is the only available fake clientset for versioned CRDs.
+	fakeAte := atefake.NewSimpleClientset(objects...)
+	ateInformerFactory := externalversions.NewSharedInformerFactory(fakeAte, 0)
+	workerPoolLister := ateInformerFactory.Api().V1alpha1().WorkerPools().Lister()
+
+	syncer := NewWorkerPoolSyncer(persistence, workerInformer, workerPoolLister)
+	syncer.Start(ctx)
+
+	workerFactory.Start(ctx.Done())
+	ateInformerFactory.Start(ctx.Done())
+
+	workerFactory.WaitForCacheSync(ctx.Done())
+	ateInformerFactory.WaitForCacheSync(ctx.Done())
+
+	return persistence, fakeK8s, fakeAte, syncer, cleanup
+}
+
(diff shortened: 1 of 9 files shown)


aditya-shantanu (changes requested):
The direction is right (gating scheduling on ateom readiness closes a real gap), but as-is this breaks a large part of the controlapi test suite — `run-tests` is red on the PR.

**Blocking: existing tests create worker pods that never satisfy the new gate.** `isWorkerEligible` now requires `PodReady=True`, but the fake pods in `cmd/ateapi/internal/controlapi/functionaltest/common_test.go` (worker pod creation around line 499) and several syncer tests (`TestSyncer_DeleteBoundWorker_ClearsActor`, `TestSyncer_OmittedFields`, `TestSyncer_UpdateWorker_RetryOnVersionConflict`, `TestSyncer_SoftDelete_ViaInformer`, `TestSyncer_PodRecreatedWithNewUID`, ...) only set `PodIP`/`Phase`, so workers never register and dozens of tests time out with `failed to wait for worker to be registered`. Those helpers need the `PodReady=True` condition added (only the one pod in `TestSyncer_Lifecycle` was updated).

**Minor:**
- The comment at the `isWorkerEligible` call site in `reconcile` still says "No IP yet" — worth updating to cover not-Ready too.
- The TCP probe on the `https` port shows up in ateom's logs as a `TLS handshake error ... EOF` every 10s (visible in the e2e job log). Fine for now, but it's another nudge toward the `/readyz` HTTP endpoint you mentioned in the description.

One thing I checked that is *not* a problem: the gate only guards worker creation/update — once the Worker row exists, a later readiness flap just makes `reconcile` return nil, so it won't deregister a worker with a bound actor.

The `e2e-test` job is also red (`TestActorIdentity_AfterRestore_IsOwnID_NotGolden`, SuspendActor failing to dial ateom's socket); I couldn't conclusively tie that to this change vs. a flake, but it should be green before merge either way.

aditya-shantanu (commented):
Inline notes for the blocking test breakage (details in my earlier review).

cmd/atecontroller/internal/workersync/syncer.go
```diff
@@ -304,7 +304,15 @@ func (s *WorkerPoolSyncer) createOrUpdateWorker(ctx context.Context, key workerK
 }
 
 func isWorkerEligible(pod *corev1.Pod) bool {
-	return pod.Status.PodIP != ""
+	if pod.Status.PodIP == "" {
+		return false
+	}
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == corev1.PodReady {
```
aditya-shantanu, line 323: **Blocking — this is what's turning `run-tests` red.** The fake worker pods in `functionaltest/common_test.go` (worker pod creation ~L499) and in `TestSyncer_DeleteBoundWorker_ClearsActor`, `TestSyncer_OmittedFields`, `TestSyncer_UpdateWorker_RetryOnVersionConflict`, `TestSyncer_RequeueOnMissingWorkerPool`, `TestSyncer_SoftDelete_ViaInformer`, and `TestSyncer_PodRecreatedWithNewUID` only set `PodIP`/`Phase`, never `PodReady=True`, so workers never register and dozens of tests time out on `failed to wait for worker to be registered`. Those helpers need the Ready condition added (only `TestSyncer_Lifecycle` was updated).

Also: the caller-side comment at L230 ("No IP yet; a later update event re-enqueues the pod.") should mention not-Ready too.

One thing I checked that is fine: this gate only guards registration — an already-registered worker isn't deregistered by a readiness flap, so bound actors aren't disrupted.

cmd/atecontroller/internal/controllers/workerpool_apply.go
```diff
@@ -104,6 +105,9 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool, otel ateomOTelSettin
 				WithName("connect").
 				WithContainerPort(444).
 				WithProtocol(corev1.ProtocolTCP)).
+		WithReadinessProbe(corev1ac.Probe().
```
aditya-shantanu, line 112: The TCP probe against the TLS port logs `http: TLS handshake error ... EOF` in ateom every 10s (visible in the e2e job log). Fine as a first step, but it's another nudge toward the `/readyz` HTTP endpoint you floated in the description.

cmd/ateapi/internal/controlapi/syncer_test.go
```diff
@@ -0,0 +1,1575 @@
+// Copyright 2026 Google LLC
```
aditya-shantanu, line 1: **Blocking — doesn't compile.** This file is the pre-migration copy: the syncer moved to `cmd/atecontroller/internal/workersync/` on main, so `go vet` fails here with `undefined: WorkerPoolSyncer` at line 56. Delete this file and move its `PodReady` fixture fixes into `workersync/syncer_test.go`, which is where the 7 currently-failing tests live.

cmd/atecontroller/internal/workersync/syncer.go
```diff
@@ -316,7 +316,15 @@ func (s *WorkerPoolSyncer) createOrUpdateWorker(ctx context.Context, key workerK
 }
 
 func isWorkerEligible(pod *corev1.Pod) bool {
```
aditya-shantanu, line 318: **Blocking — this package's own tests weren't updated for the new gate.** `workersync/syncer_test.go` fixtures set only `PodIP`/`Phase`, never a `PodReady=True` condition, so 7 tests fail locally (`TestSyncer_Lifecycle`, `_OmittedFields`, `_UpdateWorker_RetryOnVersionConflict`, `_RequeueOnMissingWorkerPool`, `_SoftDelete_ViaInformer`, `_PodRecreatedWithNewUID`, `_ReconcileErrorsAreClassified`). Same fix as you made in `functionaltest/common_test.go` — add the Ready condition wherever the fixture marks a pod Running.

internal/serverboot/serverboot.go
```diff
@@ -331,20 +331,22 @@ func StartMetricsServer(ctx context.Context, opts MetricsServerOptions) {
 	}
 }
 
+// StartReadinessServer runs an HTTP server exposing only /readyz. Blocks until
+// http.ListenAndServe returns; designed to be `go`-launched.
+func StartReadinessServer(ctx context.Context, addr string, readiness *Readiness) {
```
aditya-shantanu, line 336: Nit: if `ListenAndServe` fails here (port taken, bad flag value), the error is logged once and the pod then just never becomes Ready — a quiet failure mode that presents as "worker never registers." Since a worker that can't serve its readiness endpoint is useless anyway, exiting non-zero (or plumbing the error to the caller like the gRPC serve path does) would make the failure loud and let the kubelet restart it.

aditya-shantanu (changes requested):
Re-reviewed at 9938b658. The design is now right and my main v1 concerns are addressed: a real HTTP `/readyz` on a dedicated port replaces the TCP probe against the TLS listener (no more handshake-error log spam), it's wired into graceful shutdown (`MarkNotReady` on SIGTERM), both ateom variants get it, the stale "No IP yet" comment is fixed, and the functionaltest fake pods now set `PodReady`. Readiness-as-ready-by-default (`Readiness{}` zero value) gives the intended semantics here: the probe turns 200 once the process is up and serving, which is exactly the startup window #1106's scheduling race needs closed.

But the PR doesn't build and its target package's tests fail — same class of issue as last round, relocated by the syncer migration that landed on main in the meantime:

1. `cmd/ateapi/internal/controlapi/syncer_test.go` is a resurrected pre-migration file: `go vet` fails with `undefined: WorkerPoolSyncer` (the syncer now lives in `cmd/atecontroller/internal/workersync/`). It looks like the v1 test fixes were applied to the old path during rebase.
2. The tests that actually exist on main — `cmd/atecontroller/internal/workersync/syncer_test.go` — were not updated, and 7 of them fail (`TestSyncer_Lifecycle`, `_OmittedFields`, `_UpdateWorker_RetryOnVersionConflict`, `_RequeueOnMissingWorkerPool`, `_SoftDelete_ViaInformer`, `_PodRecreatedWithNewUID`, `_ReconcileErrorsAreClassified`): their pod fixtures set only IP/Phase, so nothing is eligible under the new gate. Ran locally to confirm; `serverboot` and `controllers` packages pass.

One scope note, not a blocker: the identity flake's only two post-merge (#1160) failures show a *resume-side* signature — the restored actor's own container readyz timing out (`readyz for "probe" never returned 200 within 1m0s`, actor stuck RESUMING) — which worker-registration gating doesn't obviously cover. Worth checking against #1106's race analysis whether this PR closes that path too or whether a second fix is still needed; if the latter, the issue should stay open after this merges.

Fix the two test items and I'm happy to approve — the production change itself looks correct.

aditya-shantanu:
Heads-up @Stevenjin8: to get #1106 closed faster we've adopted this PR's approach into #1243 (your commits credited with Co-authored-by), plus the review items: the pre-migration syncer_test copy is dropped, the workersync fixtures/tests are updated for the #1203 move, and the readiness server now fails loudly. If you'd rather land this PR instead, #1243 can yield — whichever is ready first wins.

Stevenjin8:
@aditya-shantanu Yeah I don't super care, Ill leaveit up to you

aditya-shantanu:
this is mostly your work. I've made sure you get the credit.
i just want to make sure we can get to less flaky tests asap. thanks for your understanding.

Issue #1011, opened 2026-08-17
Simplify update semantics to do whole object replace


juli4n:
The way we do updates today is using AIP's field masks. Each `Update<Resource>Request` message has a `google.protobuf.FieldMask update_mask` field to specify the paths of fields to merge.  

Considering that merge semantics have proven to be hard to get right (Kubernetes went through many iterations to get it right, with strategic merge patch, JSON patch, JSON merge patch, SSA) and that [AIP-161](https://google.aip.dev/161) is vague on how to deal with things like non-scalar values (maps and repeated fields), I'm worried we won't get it right either, even more so if we rush it.  

This task proposes we start with whole-object updates. This would remove the `update_mask` field completely and leave the resource field as the only field in the `Update<Resource>Request` message. Clients will be able to fence the update on either version or UID in the same way they do it today to prevent clients from overwriting each other (i.e. blind updates).  

Note that this decision doesn't preclude us from adding some form of patch updates in the future. For example, we could add the `update_mask` field and let the default (no mask) be whole-object replacement. On the other hand, getting the merge semantics wrong would put us in a bad place where we have to support that forever.

thockin:
> Clients will be able to fence the update on either version or UID in the same way they do it today to prevent clients from overwriting each other (i.e. blind updates).

Clarify - are you proposing that clients MAY or MUST specify version and uid preconditions for updates?

> On the other hand, getting the merge semantics wrong would put us in a bad place where we have to support that forever.

As the author of and repeated triager of several such bugs in k8s, I appreciate the sentiment :)

I'm OK with this.

jpbetz:
I support this.

The combination of whole object update, with required preconditions for optimistic concurrency, and with preserved fields on the client side, is safe and defensible (and is skew safe!).

> Considering that merge semantics have proven to be hard to get right (Kubernetes went through many iterations to get it right

This is too generous, actually. Only JSON patch is able to perform all possible patch operations, and it still requires the use of optimistic read/modify/write loops to manage conflicts safely.. SSA has capability gaps that we have not been able to fix. And there is client-side apply hidden there too, to the grand total is 5 patch mechanisms, none of which fully achieve both soundness and fire-and-forget.

> Note that this decision doesn't preclude us from adding some form of patch updates in the future

For posterity, my current thinking on this is:

[...]

Issue #954, opened 2026-08-14
Consider rejecting blind API writes


juli4n:
Update calls for Substrate resources (Actor, ActorSnapshot, etc) expose an optimistic locking schema. Requests can specify `metadata.version` and/or `metadata.uid`, and the ate-apiserver will use those values to fence the update and make it only succeed if it matches the value in the DB. This check is implemented in a transactional read-modify-write.

Now, do we want to support this? Clients doing blind writes are dangerous as they can override each others requests (last request wins) when doing concurrent updates. 

This task is to evaluate if we want to make `metadata.version` required for update methods. Also, should we require `metadata.uid` as well or just keep it optional?

@laoj2 @thockin @jpbetz

laoj2:
@juli4n just clarifying: the idea is to reject blind API writes only for Update methods, right? I'm not sure pinning a version  makes sense for Delete methods

juli4n:
Ah, good clarification (I updated the title to reflect this better). Yes, this issue was mostly about Update<Resource> calls. 

I think deletes are different, and making them optional makes sense as requiring clients to always do a read just to delete seems excessive / unnecessary.

jpbetz:
> Ah, good clarification (I updated the title to reflect this better). Yes, this issue was mostly about Update calls.
> 
> I think deletes are different, and making them option makes sense as requiring clients to always do a read just to delete seems excessive / unnecessary.

This is very lifecycle dependent so you should make sure that callers have the option of only deleting if a precondition is met, but I would expect a lot of deletion use cases not to require the preconditions for safety.

thockin:
Forcing RMW for updates amplifies the risk of data loss in the face of version skew.  AFAICT there are 2 fixes - PATCH and preserving unknown fields in clients.

At this point in time I am OK to cut PATCH support (for scope - it will come back) and require preconditions, but we should decide very soon how we want clients to behave.  We don't yet have an SDK beyond "here's a proto file", but I think we should make sure all our examples allow RMW to work correctly, without data loss.

@HavenXia since this intersects with upgrades.

laoj2:
> At this point in time I am OK to cut PATCH support (for scope - it will come back) and require preconditions, but we should decide very soon how we want clients to behave. We don't yet have an SDK beyond "here's a proto file", but I think we should make sure all our examples allow RMW to work correctly, without data loss.

Sounds good. For now, I'll make sure the examples are correct and also add a note to the api-guide and api-style-guide docs to make this clearer

[...]

PR #1108, opened 2026-08-20
Simplify Update methods to do a whole object replace

laoj2:
* Removed field_mask from the API
* Added a new protoupdate package to handle replacing mutable fields. This makes sure that unknown fields in the server are not dropped by an update from a stale/old client.

#1011 

- [x] Tests pass


cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -268,11 +268,15 @@ func validateListActorsRequest(req *ateapipb.ListActorsRequest) field.ErrorList
...
 }
 
-// actorMutableFields lists the Actor field paths a client may name in an
-// UpdateActor update_mask.
-var actorMutableFields = fieldmask.NewMutableFields(
-	"worker_selector",
-	"worker_selector.match_labels",
+// actorImmutableFields lists the Actor fields an update never takes from the
+// request: the stored values win, whatever the client sends.
+var actorImmutableFields = protoupdate.ImmutableFields[*ateapipb.Actor](
```
juli4n: I hope we don't need to enumerate this and can have a generic merge logic that works for all resources / fields. See my comment in https://github.com/agent-substrate/substrate/issues/954#issuecomment-5371533363

laoj2: Updated this PR based on our conversations. Now, we're doing:

```
serverOwned := storedObj[...]

proto.Reset(storedObj)
proto.Merge(storedObj, inputObj)
		
storedObj[...] = serverOwned		
```

One thing to note is that this operation retains all unknown fields from the input actor. But, we've been discussing what behaviour we want to keep in #954

juli4n (approved):

PR #1109, opened 2026-08-20
Fix TestActorIdentity_AfterRestore_IsOwnID_NotGolden flaky test

shrutiyam-glitch:
`TestActorIdentity_AfterRestore_IsOwnID_NotGolden` (and other E2E tests) failed intermittently with an Envoy 503 "upstream connect error".
For example: https://github.com/agent-substrate/substrate/actions/runs/32400171542/job/96526194319#step:12:549 

Issue: The test called `ResumeActor` and then immediately fired an HTTP request at the actor. While the control plane successfully resumed the actor, it takes a fraction of a second for the new routing information to propagate to the Envoy proxy (via xDS). The test's HTTP request slammed into Envoy before its routing tables were updated to point to the newly restored pod.

Fix: updated the shared test client (`RouterClient.request`in `router_client.go`). It now explicitly intercepts 503 "upstream connect error" responses and retries the request for up to 30 seconds (`time.Sleep(1 * time.Second)`).

- [ ] Tests pass

### DIFF
--- a/internal/e2e/router_client.go
+++ b/internal/e2e/router_client.go
@@ -123,7 +123,9 @@ func (c *RouterClient) request(ctx context.Context, method string, actorRef reso
 		}
 	}
 
-	deadline := time.Now().Add(30 * time.Second)
+	retryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
 	for {
 		var reqBody io.Reader
 		if bodyBytes != nil {
@@ -146,10 +148,10 @@ func (c *RouterClient) request(ctx context.Context, method string, actorRef reso
 		if resp.StatusCode == http.StatusServiceUnavailable {
 			respBodyBytes, _ := io.ReadAll(resp.Body)
 			resp.Body.Close()
-			if strings.Contains(string(respBodyBytes), "upstream connect error") && time.Now().Before(deadline) {
+			if strings.Contains(string(respBodyBytes), "upstream connect error") {
 				select {
-				case <-ctx.Done():
-					return nil, ctx.Err()
+				case <-retryCtx.Done():
+					return nil, retryCtx.Err()
 				case <-time.After(1 * time.Second):
 					continue
 				}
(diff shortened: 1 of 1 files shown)


internal/e2e/router_client.go
```diff
@@ -114,16 +114,50 @@ func (c *RouterClient) PostJSON(ctx context.Context, actorRef resources.ActorRef
...
+
+		resp, err := c.http.Do(req)
+		if err != nil {
+			return resp, err
+		}
+
+		if resp.StatusCode == http.StatusServiceUnavailable {
+			respBodyBytes, _ := io.ReadAll(resp.Body)
+			resp.Body.Close()
+			if strings.Contains(string(respBodyBytes), "upstream connect error") && time.Now().Before(deadline) {
```
zoez7: Nit: use the status code only and avoid string matching with error message?

zoez7: Nit: use `retryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)` above and check `case <-retryCtx.Done():` below instead?

zoez7 (dismissed):

internal/e2e/router_client.go
shrutiyam-glitch: done, thanks.

shrutiyam-glitch: looks like if we retry all 503s, we'll probably break the `TestRequestParking/BudgetExhaustion` e2e test. 
 `ext_proc` also uses `503 Service Unavailable` to signal that a parked request ran out of time waiting for capacity (no free workers available). 
 If this client loops on all 503s, it masks that legitimate timeout and retries it up to the 30s deadline, causing the parking test to fail its timing assertions. 
 Or in order to differentiate between the router rejected requests (ex:parking budget exhaustion) and the other normal cases, we could add a header here - https://github.com/agent-substrate/substrate/blob/main/cmd/atenet/internal/router/extproc/errors.go#L67 
`{
							Header: &corev3.HeaderValue{Key:"x-substrate-error",
								RawValue: []byte("1"),
							},
						},`
not sure if it is in scope for this PR. wdyt ?

shrutiyam-glitch:
Closing this PR as it is superseded by #1107

PR #1113, opened 2026-08-20
fix(ateapi): add store fallback in ActorIdentity to resolve test flakiness

shrutiyam-glitch:
Fix `TestDurableDirLifecycle` flaky test
https://github.com/agent-substrate/substrate/actions/runs/32416349094/job/96578260374 

Cause: 
- In the specific test (`_suspend_from_PAUSED`), the actor is suspended (which strips its worker assignment, returning the worker to the pool) and then immediately resumed.
- When the test calls `ResumeActor`, the control plane selects a new worker, writes the assignment to the Kubernetes database, and synchronously makes a direct gRPC Restore() call to the atelet on that new worker.
- The `atelet` boots the actor and `atunnel`. `atunnel` then immediately asks the `ateapi` (via the `ActorIdentity` service) to mint an actor certificate so it can connect to the network.

Race: The `ActorIdentity` service verifies that the worker is actually assigned to the actor by checking its internal `workercache.Cache`. Because `ResumeActor` was so fast, the cache hasn't processed the new worker assignment yet.
The `ActorIdentity` service looks at the stale cache, sees the worker is unassigned, and hard-rejects the certificate minting with caller is not permitted to mint credentials for this actor.

I faced this issue too when working on the postgresql updateWorker optimization (#934)

Fix: Updated `authorizeActor` inside `cmd/ateapi/internal/actoridentity/actoridentity.go`.
Now, if `authorizeActor` fails an authorization check (like discovering a missing or mismatched worker assignment), it will automatically do a one-time "live fetch" bypass (s.store.GetWorker()) directly against the primary store to bypass the lagging informer cache.

- [ ] Tests pass


BenTheElder:
> Now, if authorizeActor fails an authorization check (like discovering a missing or mismatched worker assignment), it will automatically do a one-time "live fetch" bypass (s.store.GetWorker()) directly against the primary store to bypass the lagging informer cache.

If we're failing through commonly ... maybe we shouldn't even be using it?

I haven't had a chance to dig yet, but IIRC @zoez7 was moving actor out of CRD => substrate API?

LiorLieberman:
@zoez7 Any feedback here?

cmd/ateapi/internal/actoridentity/actoridentity.go
```diff
@@ -317,60 +317,94 @@ func validateWorkerRef(worker *ateapipb.ObjectRef) error {
...
 		}
 		slog.ErrorContext(ctx, "ActorIdentity: failed to read worker", slog.Any("err", err))
 		return nil, resources.ActorRef{}, status.Error(codes.Internal, "failed to look up worker")
 	}
+
+	actor, actorRef, reason, err := s.authorizeWithWorker(ctx, worker, caller, req)
+	if err != nil {
+		return nil, resources.ActorRef{}, err // Internal error (e.g., actor lookup failed)
+	}
+	if reason == "" {
```
zoez7: nit: explicitly return a authorized boolean instead of relying on empty string.

cmd/ateapi/internal/actoridentity/actoridentity.go
```diff
@@ -317,60 +317,94 @@ func validateWorkerRef(worker *ateapipb.ObjectRef) error {
...
+		append([]any{slog.String("worker", req.GetWorker().GetName()), slog.String("callerPod", caller.podName), slog.String("callerNode", caller.nodeName)}, args...)...)
+	return status.Errorf(codes.PermissionDenied, "caller is not permitted to mint credentials for this actor")
+}
+
+// authorizeWithWorker returns a (reason string) instead of calling denyMint directly.
+// If reason is non-empty, the authorization failed.
+func (s *Server) authorizeWithWorker(ctx context.Context, worker *ateapipb.Worker, caller *ateletCaller, req *ateapipb.MintCertRequest) (*ateapipb.Actor, resources.ActorRef, string, error) {
 	if worker.GetNodeName() != caller.nodeName {
-		return nil, resources.ActorRef{}, deny("worker is hosted on a different node", slog.String("workerNode", worker.GetNodeName()))
+		return nil, resources.ActorRef{}, "worker is hosted on a different node", nil
```
zoez7: so we return "error = nil" here because we want to retry it? Should we do
```
const assignmentMismatchErr = fmt.Error("mismatch)"
...

func (s *Server) authorizeWithWorker() {
  if worker.GetNodeName() != caller.nodeName {
    return ..., assignmentMismatchErr
  }
}


...

func (s *Server) authorizeActor() {
   ...

  actor, actorRef, reason, err := s.authorizeWithWorker(ctx, worker, caller, req)
   if errors.Is(err, assignmentMismatchErr) {
       // refetch
   }
}

```

zoez7 (commented):
> I haven't had a chance to dig yet, but IIRC [Zoe Zhao (@zoez7)](https://github.com/zoez7) was moving actor out of CRD => substrate API?

This is WIP, not fully cut over yet.

> Race: The ActorIdentity service verifies that the worker is actually assigned to the actor by checking its internal workercache.Cache. However, this cache is powered by Kubernetes Informers, which receive updates asynchronously.

I got confused by this part of PR description. The worker cache used here is a in-memory cache of the store data, has nothing to do with k8s informers.

zoez7 (approved):

PR #1116, opened 2026-08-21
ateomnet: move the actor nftables table to the inet family

ygao-g:
Part of #246. First of two: #1057 builds on this to give the actor an IPv6 address. This PR has no dependency of its own.

### Motivation & Changes

The actor's NAT and filter rules currently live in an `ip` table, which can only ever carry IPv4. This PR moves them to an `inet` table so a single table can hold both address families once the actor veth becomes dual-stack. 

Key mechanics:
* **Table Family Migration:** Moved the actor's NAT and filter rules to the `inet` family. The IPv6 traffic in the worker pod netns is now conntracked. The forward chain's `accept` now covers IPv6 too.
* **IPv4 Payload Guards:** Every IPv4 match now opens with an `NFPROTO_IPV4` comparison.
* **Robust Teardown:** Teardown now sweeps tables of both `ip` and `inet` to ensures we handle and clear the `ip` table an earlier ateom might have left behind. 

### Testing

Root-gated tests:
* `TestSetupActorNetworkEgressRedirect` checks that an `inet` nat chain actually takes the redirect by reading the installed rule back from the kernel.
* `TestRemoveActorNftablesRules` covers teardown for every family mix — `ip` only (the upgrade case), `inet` only, and both. 
* `TestSetupActorNetworkIsRepeatable` asserts the table is gone after each teardown.

### DIFF
--- a/internal/ateomnet/net.go
+++ b/internal/ateomnet/net.go
@@ -229,8 +229,8 @@ func InstallActorNftablesRules(egressPort uint16) error {
 	// rules in an ateom-owned table makes cleanup simple and avoids mutating
 	// Kubernetes or CNI-managed chains directly.
 	//
-	// TODO: Add IPv6 veth addressing, forwarding, and nftables rules once actor
-	// networking supports dual-stack pods. The current actor network is IPv4-only.
+	// TODO(#246): Add the IPv6 veth addressing and forwarding. The actor
+	// network itself is still IPv4-only.
 	//
 	// The rules do three things:
 	//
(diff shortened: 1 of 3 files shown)


internal/ateomnet/net.go
ygao-g, line 306: to minimize diff, would it be better to introduce a helper function `func RemoveActorNftablesRules(ipFamily)`  and have the `Remove*` one iterate through the two family, call the function, aggregate errors if needed

PR #1118, opened 2026-08-21
WIP: hack: only repoint ipv6 kubeconfig when [::1] healthz fails

### LINKED ISSUE #1099: hack: the ipv6 kubeconfig repoint breaks kubectl on a native-Linux Docker host

asjdf:
- Fixes #1099
- This is the kubeconfig-only extract of the gate already in #958, so native-Linux Docker hosts are not blocked while that larger IPv6 DNS PR waits for review
- Happy to close this if #958 merges first
- Cannot run full `IP_FAMILY=ipv6 hack/create-kind-cluster.sh` here (no Docker); `bash -n` clean

### DIFF
--- a/hack/create-kind-cluster.sh
+++ b/hack/create-kind-cluster.sh
@@ -146,14 +146,16 @@ if [[ "${IP_FAMILY}" != "ipv4" &&
 fi
 
 # For ipv6 kind writes a kubeconfig pointing at [::1], the address it published
-# the apiserver on, which only works for a client on the Docker host itself: a
-# VM-hosted daemon (Lima on macOS) forwards the port to the *v4* loopback, so
-# every kubectl below fails at connect. localhost is a SAN on the apiserver
-# cert and lets the client pick a family that works from either side.
+# the apiserver on. That works for a client on the Docker host itself but not
+# behind a VM-hosted daemon (Lima on macOS), which forwards the port to the *v4*
+# loopback. localhost is a SAN on the apiserver cert and covers that case.
+# Don't guess the platform; ask the apiserver. Debian-family hosts name ::1
+# ip6-localhost / ip6-loopback, not localhost. See issue #1099 and PR #958.
 if [[ "${IP_FAMILY}" == "ipv6" ]]; then
   server="$(kubectl config view \
     -o jsonpath="{.clusters[?(@.name==\"${KUBECTL_CONTEXT}\")].cluster.server}")"
-  if [[ "${server}" == "https://[::1]:"* ]]; then
+  if [[ "${server}" == "https://[::1]:"* ]] &&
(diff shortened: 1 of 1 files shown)


hack/create-kind-cluster.sh
```diff
@@ -146,14 +146,16 @@ if [[ "${IP_FAMILY}" != "ipv4" &&
 fi
 
 # For ipv6 kind writes a kubeconfig pointing at [::1], the address it published
-# the apiserver on, which only works for a client on the Docker host itself: a
-# VM-hosted daemon (Lima on macOS) forwards the port to the *v4* loopback, so
```
BenTheElder, line 150: This seems like a bug we should report to kind upstream. Does lima expose this information in any way?

EDIT: I might argue this is a limactl bug. Seems like it should forward consistently to the macOS host.

BenTheElder: https://github.com/lima-vm/lima/issues/1540

BenTheElder: cc @aojea we're running into https://github.com/kubernetes-sigs/cloud-provider-kind/pull/422#issuecomment-4671122497 again

aojea: anything we can do?

BenTheElder: I think we might need to follow-up with limactl, I'm not sure how kind can reasonably detect this, limactl should really forward the port on the host with the same IP family?

BenTheElder: If we inspect the network / containers is there any indication of this there? E.g. docker desktop sets a label IIRC.

aojea: I remember I commented somewhere about how should it work, I can not find it now

PR #1125, opened 2026-08-21
hack: make the egress ext_proc script run on macOS

### LINKED ISSUE #1124: hack/install-ate.sh fails to parse on macOS bash 3.2
`hack/install-ate.sh` fails on macOS before installing anything:

```
hack/experimental-additional-egress-extproc.sh: line 96: unexpected EOF while looking for matching `)'
```

`install-ate.sh:52` unconditionally sources that file, and it does not parse under macOS's `/bin/bash` (3.2). Line 115 has an apostrophe inside a heredoc nested in `$( )`:

```
    # The actor's verified identity
```

bash 3.2 scans `$( ... )` for the closing paren without tracking heredoc bodies. Repro on any macOS shell:

```bash
$ printf 'x="$(cat <<E\nit%ss fine\nE\n)"\n' "'" > t.sh && /bin/bash -n t.sh
t.sh: line 2: unexpected EOF while looking for matching `''
```

Every local install breaks, not just runs passing `--experimental-additional-egress-extproc-service`. Introduced in 57c3e192 (#1085). CI runs bash 5 on Linux, where the file parses fine.

orangeCatDeveloper:
Fixes #1124

`hack/install-ate.sh` exits after one second and installs nothing on macOS, with no error saying so.

It sources `experimental-additional-egress-extproc.sh` unconditionally, and that file hits two things macOS's tools reject: bash 3.2 mis-parses the heredocs inside `$( )` because a comment in one holds an apostrophe, and awk refuses a newline inside a `-v` assignment. The first breaks every install, not only runs passing the experimental flag.

Both blocks now use `read -r -d ''` and reach awk through `ENVIRON[]`. Not "delete the apostrophe": three more are already in the file and only happen to pair up. `read` keeps the trailing newline `$( )` strips, hence the two `%$'\n'` lines — without them each spliced block gains a blank line.

Verified on macOS 26 (bash 3.2.57, awk 20200816) and a Linux container (bash 5.2.37, mawk 1.3.4). The manifest is byte-identical before and after on Linux, and identical on macOS to what Linux produces (56115 bytes all three). This does not make the experimental flag usable end to end on macOS — deploying a real ext_proc service is out of scope.

```
$ bash hack/install-ate-kind.sh --deploy-ate-system --store-backend=postgres \
    --experimental-use-sdsmint \
    --experimental-additional-egress-extproc-service=ate-system/extprocd:50051
hack/experimental-additional-egress-extproc.sh: line 96: unexpected EOF while looking for matching `)'
$ echo $?; kubectl -n ate-system get pods --no-headers | wc -l
2
0

$ bash hack/install-ate-kind.sh --deploy-ate-system --store-backend=postgres \
    --experimental-use-sdsmint \
    --experimental-additional-egress-extproc-service=ate-system/extprocd:50051
...
deployment "atenet-egress" successfully rolled out
$ echo $?; kubectl -n ate-system get cm atenet-egress -o yaml | grep -c additional_egress_ext_proc
0
5
```

### DIFF
--- a/hack/experimental-additional-egress-extproc.sh
+++ b/hack/experimental-additional-egress-extproc.sh
@@ -92,8 +92,10 @@ patch_atenet_egress_manifest() {
 
   # Both blocks are written at column zero; awk below re-indents each to the
   # marker it replaces, so the depth they nest to is the manifest's to decide.
+  # bash 3.2, which macOS ships, mis-parses a heredoc inside $( ) whose body
+  # holds an apostrophe. read keeps the trailing newline that $( ) would strip.
   local filter_block cluster_block
-  filter_block="$(cat <<EOF
+  IFS= read -r -d '' filter_block <<EOF || true
 # Added by hack/install-ate.sh
 # --experimental-additional-egress-extproc-service=${ATE_ADDITIONAL_EGRESS_EXTPROC_SERVICE}.
 # Spliced over each #ATE_MITM_EXTPROC_FILTER marker in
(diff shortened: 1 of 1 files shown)


haiyanmeng:
@orangeCatDeveloper , thanks for flagging this. I reverted https://github.com/agent-substrate/substrate/pull/1085 (commit https://github.com/agent-substrate/substrate/commit/57c3e1927c4c946b8684fc0dcfacc226b066a289) in https://github.com/agent-substrate/substrate/pull/1129 for now.

PR #1046, opened 2026-08-18
benchmarking: right-size actor memory, default 256Mi

Oneimu:
#### benchmarking: right-size actor memory, default 256Mi

Benchmark `ActorTemplates` declared no resources, so microvm actors fell through to the 2 GiB kata default guest — and everything the guest kernel caches rides along in the memory snapshot, inflating snapshot size and suspend/resume latency, which the benchmarks then measure.

Set `spec.resources.limits.memory` on the `glutton` and `sleep` templates, parameterized as `ACTOR_MEMORY` with a `256Mi` default (the smallest size microvm admits: 128Mi VMM reserve + 128Mi guest floor), and thread it through the deploy chain:
- `--actor-memory` on `workloads/deploy.sh` and `deploy_locust.sh`
- `--benchmark-actor-memory` on `install-ate.sh`
- Optional per-test `actorMemory` field in the automation's `tests.yaml` for future `WriteRAM` stress suites.

PR #1130, opened 2026-08-21
 benchmarking(memory usage phase 1 baseline): large-memory glutton workloads (--mem-targetLarge mem bench

Oneimu:
## What this PR does:

Adds large-memory benchmark suites: glutton actors that hold a resident 1–2Gi working set, so the suspend/resume path is measured at realistic application sizes on both runtimes — {1Gi, 2Gi} × {gvisor, microvm}, tracked.

The boomer GluttonUser fills each actor to a configured target via chunked `WriteRAM` calls (64Mi per keyed allocation — the proto size field is int32) after resume and **before the first suspend**, so every snapshot from cycle one onward carries the full working set. `WriteRAM` writes incompressible random bytes, so snapshots are genuinely target-sized rather than zstd-compressing away (a zero-filled working set compresses ~35:1 and would shrink the upload/download phases to a few MB).

## How it's configured

The target flows through the established boomer dynconfig channel, like the durdir knobs: `--mem-target-bytes` in the suite's locust `flags:` → `/boomer-config` → the Go worker. Because it's per-user-class runtime config, heterogeneous and changing workload shapes are expressible with no redeploy, and the deploy stack needs no changes.

Changes:
- `cmd/benchmarking/glutton`: route `WriteRAM` on the HTTP-mode mux (it was gRPC-only, unreachable in `--mode=http` deployments)
- `internal/benchmarking/boomer/glutton`: `ensureRAMFilled` on the user cycle — runs once per actor (glutton holds the allocations across suspend/resume), retried on failure, reported as its own `GluttonFillRAM` stats row so it never pollutes ping/resume numbers
- `internal/benchmarking/boomer/dynconfig` + `common/boomer_config.py`: new `mem_target_bytes` knob
- `tests.yaml`: the four tracked suites, with `actorMemory` sized above the target for headroom

## Behavior notes

- The golden template snapshot and each actor's cold boot stay small: the working set exists from the first fill onward. All steady-state suspend/resume cycles measure at size; `ResumeActorColdStart` does not.
- Each actor's first suspend uploads the first at-size snapshot and follows the fill — cycle one is an expected outlier on the dashboards.
- A mid-run change to `mem_target_bytes` applies to newly spawned users; already-filled actors keep their size, so each actor's cycles stay comparable.
- Existing suites are untouched: with no flag the target is 0 and the fill is a no-op.

## Testing

- `go test -race` across `cmd/benchmarking/glutton`, `internal/benchmarking/boomer/...`, and the glutton fake: PASS. New tests cover chunking to an exact target, disabled-by-default, and fail-then-retry.
- Cluster verification (microvm, 1Gi target): verified that the RAM fill completes before the initial suspend, snapshot upload reflects the expected ~1Gi payload, and steady-state suspend/resume cycles succeed without error.


cmd/benchmarking/glutton/memload.go
```diff
@@ -0,0 +1,214 @@
...
+// size everywhere we run; writing one byte per page dirties the whole page.
+const pageBytes = 4096
+
+// startMemLoad allocates target bytes, dirties all of it once, and then keeps
+// re-dirtying it: one full pass per interval ("sequential"), or the same
+// write rate spread over uniformly random pages ("random"). It returns after
+// the initial full dirtying pass, so a caller that waits for readyz after
+// this knows the working set is resident.
+func startMemLoad(ctx context.Context, target int64, interval time.Duration, pattern string) (*memLoad, error) {
+	if target <= 0 {
```
BenTheElder: nit: if we're doing this sort of check, we should ensure time.Duration is not negative either (ticker will panic on negative)

maxsmythe:
I like the idea of patterns  of memory usage and we definitely need a memory benchmark. I wonder about using flags to configure it, which may not be super flexible.

What do you think about adding the ability to invoke this behavior via HTTP/proto api?

https://github.com/agent-substrate/substrate/blob/ecca2cac00b3511e5f598522a60718483e849e41/internal/proto/glutton/glutton.proto#L68-L76

Advantages:
* No need to customize launch flags
* Can easily run heterogenous/changing workloads throughout the fleet, using user classes and workload shapes.
* Can control the pacing of the behavior

Disadvantages
* Every behavioral change requires a network request
* More complex behaviors are hard to model (though flags have this problem too, some kind of scripting solution would be needed)

maxsmythe:
Also we would need to have something that drives suspend/resume/etc. and gathers the latency measurements

cmd/benchmarking/glutton/main.go
```diff
@@ -63,6 +63,10 @@ var (
 	dataDir           = pflag.String("data-dir", "", "Directory under which WriteDisk files are stored. Required.")
 	mode              = pflag.String("mode", "grpc", "Wire protocol for the main listener: grpc (default) or http.")
 
+	memTarget        = pflag.String("mem-target", "", "If set (e.g. 512Mi, 2Gi), allocate this much memory at startup and keep re-dirtying it, so the actor holds a resident working set without needing the gRPC API (see memload.go).")
```
maxsmythe: As before, we should instead use the GRPC protocol to determine the test behavior instead of flags. This allows for a greater diversity of workloads in any given test.

benchmarking/deploy_locust.sh
```diff
@@ -74,6 +78,8 @@ while [[ "$#" -gt 0 ]]; do
     --otlp-endpoint=*) OTLP_ENDPOINT="${1#*=}" ;;
     --actor-memory) shift; ACTOR_MEMORY="$1" ;;
     --actor-memory=*) ACTOR_MEMORY="${1#*=}" ;;
+    --mem-target) shift; MEM_TARGET="$1" ;;
```
maxsmythe: this would go away if we removed flags

benchmarking/automation/orchestrator.py
```diff
@@ -348,6 +349,10 @@ def deploy_workloads(
     # Empty keeps deploy.sh's own default; large fleets set workerWaitTimeout.
     if wait_timeout:
         cmd += ["--wait-timeout", wait_timeout]
+    # Empty keeps glutton's self-driving memory load disabled; large-memory
+    # suites set memTarget in tests.yaml (with actorMemory sized above it).
+    if mem_target:
```
maxsmythe: no flag usage would lead to no code changes to this file.

benchmarking/automation/tests.yaml
```diff
@@ -41,6 +41,11 @@
 #                  the ateom worker pods to become ready. Raise it for
 #                  large workerCount fleets whose image pulls or scheduling
 #                  outrun the default.
+#   memTarget      Optional (default "" = disabled). Resident working set the
```
maxsmythe: ironically, this should live in "flags" -- these are the flags passed to locust to configure workers.

maxsmythe (commented):
We still need to refine how behavior is determined -- a global configuration likely isn't flexible enough. I strongly suggest looking at the .proto definitions for glutton -- there is room to extend them and this is an established pattern.

You can modify an existing glutton user to send the triggers.

cmd/benchmarking/glutton/main.go
Oneimu: Reworked to drive the working set via chunked `WriteRAM` calls before the actor's first suspend (configured via `--mem-target-bytes` in locust flags), dropping all deploy-stack changes.

cmd/benchmarking/glutton/main.go
```diff
@@ -185,6 +185,10 @@ func newMux(svc *gluttonService) *http.ServeMux {
 	mux.HandleFunc("/ping", protoRoute("Ping", svc.Ping))
 	mux.HandleFunc("/writedisk", protoRoute("WriteDisk", svc.WriteDisk))
 	mux.HandleFunc("/readdisk", protoRoute("ReadDisk", svc.ReadDisk))
+	// WriteRAM is how benchmark drivers give an actor a resident working set
```
maxsmythe: If we're not documenting the other routes, we should be consistent

internal/benchmarking/boomer/glutton/lifecycle.go
```diff
@@ -54,6 +54,11 @@ const (
 	templateNS   = "benchmark-workloads"
 	actorDomain  = "actors.resources.substrate.ate.dev"
 	pingPath     = "/ping"
+	writeRAMPath = "/writeram"
+
+	// ramChunkBytes bounds one WriteRAM call: the proto size field is int32,
+	// so a >2GiB target must arrive as several keyed allocations.
+	ramChunkBytes = 64 << 20
```
maxsmythe: Let's just modify the API to allow suffixes.

cmd/benchmarking/glutton/main.go
Oneimu: done.

internal/benchmarking/boomer/glutton/lifecycle.go
Oneimu: Done.

benchmarking/automation/tests.yaml
```diff
@@ -91,6 +91,79 @@ tests:
...
+    description: "Large-memory: 2Gi resident working set, microvm"
+    targetCluster: dev
+    file: /app/tests/glutton.py
+    duration: 1m
+    users: 1
+    workerCount: 1
+    sandboxClass: microvm
+    actorMemory: "2560Mi"
+    flags:
+      - "--mem-target-bytes"
```
maxsmythe: just change this to mem-target (if we're gonna add suffixes, we should carry the readability benefit all the way through)

cmd/benchmarking/glutton/main.go
```diff
@@ -384,10 +385,17 @@ func (s *gluttonService) WriteRAM(ctx context.Context, req *glutton.WriteRAMRequ
 	if req.GetKey() == "" {
 		return nil, status.Error(codes.InvalidArgument, "key is required")
 	}
-	if req.GetSize() < 0 {
+	size64 := int64(req.GetSize())
```
maxsmythe: lets just drop the number-based one. The API isn't used for anything yet, so no worries changing it.

Oneimu: Done.

benchmarking/automation/tests.yaml
Oneimu: Done.

benchmarking/automation/tests.yaml
```diff
@@ -91,6 +91,79 @@ tests:
...
+      - "1.0"
+  - name: glutton_mem_2gi_gvisor
+    type: locust
+    description: "Large-memory: 2Gi resident working set, gvisor"
+    targetCluster: dev
+    file: /app/tests/glutton.py
+    duration: 1m
+    users: 1
+    workerCount: 1
+    actorMemory: "2560Mi"
```
maxsmythe, line 142: shouldn't these go away? Moved to `flags` field.

cmd/benchmarking/glutton/main.go
```diff
@@ -384,10 +385,14 @@ func (s *gluttonService) WriteRAM(ctx context.Context, req *glutton.WriteRAMRequ
 	if req.GetKey() == "" {
 		return nil, status.Error(codes.InvalidArgument, "key is required")
 	}
-	if req.GetSize() < 0 {
+	size64, err := parseBytes(req.GetSize())
```
maxsmythe: nit: size64 -- "64" suffix unnecessary

maxsmythe (dismissed):
LGTM after nits

benchmarking/automation/tests.yaml
Oneimu: `memTarget` was moved to flag. 

However,  `actorMemory` is something different:  `actorMemory` is the VM's physical RAM size, set at deploy time; implemented in #1046 to right-size the memory. `actorMemory` unspecified defaults to `256Mi`

maxsmythe: ah, thank you for this. Makes sense.

Probably we'll need to eventually figure out how to more easily customize actors as we got more and different types of them, but this works for now.

Issue #1131, opened 2026-08-21
Mirror systemInfo volumes into the ActorTemplate substrate resource


thompsonmax:
The ActorTemplate substrate resource (#824, updated in #1026) corresponds to the ActorTemplate CRD, but its Volume union predates the newer CRD volume sources: it only has durableDir and externalVolumeTemplate. systemInfo (#803, plus trustBundle from #941) is missing, so a template created through the substrate API cannot carry SystemInfo projections.

Work:
  - [x] Add SystemInfoVolumeSource (actorMetadata and trustBundle data sources) to the ateapipb Volume union, using the plain-field union shape from #962, and cover it in the store contract round-trip.
  - [x] Once controlapi starts serving ActorTemplates (#477): enforce the CRD's CEL rules (path shape, exactly-one data source, MaxItems, path uniqueness) and defaulting in the create/update handlers, since the kube-apiserver no longer does it for free, and convert from the substrate resource in the actor-start path.
  - [ ] Keep the mirror current as data sources are added (actorIdentityToken is next, #802 Part 2).

Note: image volumes (#923) have the same gap and could be backfilled in the same pass.

thompsonmax:
First 2 tasks are complete. For keeping the mirror current (task 3), this can be moved into #802.

PR #1026, opened 2026-08-17
Update ActorTemplate substrate resource definition per latest discussion

zoez7:
Part of #477 

* The ActorTemplate substrate resource now corresponds to the ActorTemplate CRD.
* Included resource limits that was added in https://github.com/agent-substrate/substrate/pull/679
* Included suggestions from https://github.com/agent-substrate/substrate/issues/962

Issue #962, opened 2026-08-14
Remove oneofs from ate-apiserver API


juli4n:
We plan to use k8s declarative validation framework for validating requests in the Substrate API. The DV framework doesn't play well with proto oneofs so we need to avoid them for discriminated unions. 

Oneof can be just replaced with two message-type fields (for each option) plus an optional string discriminator field.

Today we use oneof in a couple of places:
* `EnvVar` source (only one value: string).
* `Volume` source (durable dir or external volume template).

juli4n:
There are no oneofs in the API service messages now. Additionally, after #1274 presubmit checks will fail if an oneof field added.

PR #1134, opened 2026-08-21
Mirror systemInfo volumes into the ActorTemplate substrate resource

thompsonmax:
Part of #1131.

The substrate ActorTemplate's Volume union (#824, #1026) predates the systemInfo volume source (#803, trustBundle in #941): it only had durableDir and externalVolumeTemplate, so a template created through the substrate API could not carry SystemInfo projections. This PR adds:

- `SystemInfoVolumeSource` with the `actorMetadata` and `trustBundle` data sources, mirrored from the CRD types in the plain-field union shape (#962) that `Volume` already uses ("SystemInfo" joins the `type` discriminator).
- Declarative validation tags (#1215) mirroring the CRD's kubebuilder rules: the SystemInfoDataSource one-of union, list bounds, string lengths, and unique projected fields, with generated validators and tests in controlapi.
- Store contract round-trip coverage: the shared fixture now carries a systemInfo volume through both the redis and postgres backends.

Not in this PR (tracked in #1131): the CRD's cross-item CEL rules (duplicate paths across data sources, at most one actorMetadata entry) and the clean-relative-path checks, which have no declarative tag equivalent and belong in the controlapi create/update handlers when ActorTemplate serving lands (#477), and conversion of the substrate resource in the actor-start path.


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -678,6 +680,50 @@ message ExternalVolumeTemplate {
...
+// atelet on every Run/Restore. Mirrors the CRD's systemInfo volume source.
+message SystemInfoVolumeSource {
+  // data_sources lists the projections to place within the volume. At most
+  // one actor_metadata entry may appear, and file paths must be unique
+  // across all entries.
+  repeated SystemInfoDataSource data_sources = 1;
+}
+
+message SystemInfoDataSource {
+  // Exactly one of actor_metadata / trust_bundle must be set.
```
zoez7, line 700: Could you add validation annotations for the new fields similar to https://github.com/agent-substrate/substrate/pull/1215 ?

zoez7 (dismissed):

pkg/proto/ateapipb/ateapi.proto
thompsonmax: done

zoez7 (approved):
Approved. waiting for tests to pass.

Issue #372, opened 2026-07-01
Ensure Atelet API is idempotent


juli4n:
Atelet's `Run/Checkpoint/Restore` methods need to be idempotent to make sure we can recover from workflows that fail half-way through. I believe this is not true today.

This task is to go over atelet (and ateom) and ensure operations can be retried.

mjlshen:
I'm seeing that idempotency for `Run`, `Checkpoint`, and `Restore` is not guaranteed, especially if a gRPC `Unavailable` (atelet --> ateom connection interrupted, e.g. an atelet restart) or `DeadlineExceeded` ([10min deadline](https://github.com/agent-substrate/substrate/blob/76bac5d051421d9176e8ae24f3cbccd8859a7739/cmd/ateapi/main.go#L58)) is returned in the middle of these operations. In these cases, the caller has no information about whether the initial operation never started, partially ran, or completed without a response. I have a small repro for `Checkpoint` below:

## Checkpoint-specific reproduction

1. Setup locally with demo counter
```bash
KIND_CLUSTER_NAME=issue-372 hack/create-kind-cluster.sh
kubectl config use-context kind-issue-372

hack/install-ate-kind.sh --deploy-ate-system
hack/install-ate-kind.sh --deploy-demo-counter
```
Relevant output:
```
/dev/kvm not available: micro-VM support disabled (gVisor still works).
Creating cluster "issue-372" ...
Set kubectl context to "kind-issue-372"
daemon set "atelet" successfully rolled out
deployment "counter" successfully rolled out
actortemplate.ate.dev/counter condition met
```

2. Create and resume a counter actor
```bash
kubectl ate create atespace demo

kubectl ate create actor repro-main \
  -a demo \
  --template ate-demo-counter/counter

kubectl ate resume actor repro-main -a demo
kubectl ate get actor repro-main -a demo
```
Observed:
```
ATESPACE   NAME         TEMPLATE                   STATUS           ATEOM POD                                   ATEOM IP      VERSION
demo       repro-main   ate-demo-counter/counter   STATUS_RUNNING   ate-demo-counter/counter-7c877c5cfd-qt9ml   10.244.0.30   3
```
Save the worker location before suspending:
```bash
kubectl ate get actor repro-main -a demo -o json \
  > /private/tmp/repro-main-running.json

export WORKER_NS="$(
  jq -r '.actors[0].ateomPodNamespace' \
    /private/tmp/repro-main-running.json
)"

export WORKER_POD="$(
[...]

PR #1136, opened 2026-08-21
microvm: find a restored actor's CH api-socket without ateom's record

troychiu:
Towards #372

When `ateom` restarts, it loses in-memory actor records and falls back to socket lookup on disk. Restored actors listen on `clh-api-restore.sock` rather than `clh-api.sock`, causing `terminateWorkload` to miss the VMM (leaving orphan guests running) and `CheckpointWorkload` to time out.

This change:
- Uses `chSocketFor` to probe disk for whichever API socket exists when no in-memory record is present.
- Adds `kata.RestoredCLHSocketPath` alongside `kata.CLHSocketPath` for consistent socket naming across launch, restore, and teardown.

AI has assisted with this PR and I have verified all the changes.

- [x] Tests pass


troychiu:
Rebased onto main to pick up CI fixes

PR #1137, opened 2026-08-21
atelet: serialize node-local operations per actor

troychiu:
Towards #372 

Serializes node-local operations (Run, Checkpoint, UploadPausedCheckpoint, Restore, Terminate) on a per-actor basis in atelet to prevent concurrent operations from interfering with each other's on-disk state (e.g. reading or wiping checkpoint directories concurrently).

- [X] Tests pass

---

### DIFF
--- a/cmd/atelet/actorlocks.go
+++ b/cmd/atelet/actorlocks.go
@@ -1,0 +1,61 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


cmd/atelet/actorlocks.go
```diff
@@ -0,0 +1,61 @@
...
+package main
+
+import (
+	"sync"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// actorLocks serializes concurrent node-local operations for the same actor.
```
zoez7, line 24: Do we need this given that the ate apiserver already holds a global lock per actor before each Actor operation?

troychiu: I think it's still needed.

If an `ateapi` request cancels or times out, `ateapi` exits and releases the distributed lock. However, the in-flight gRPC call on `atelet` may still be running or unwinding. If a retry arrives immediately, `ateapi` re-acquires the lock and issues a second RPC to `atelet`, resulting in concurrent operations interfering with the actor's local state.

PR #1145, opened 2026-08-22
Reapply reverted commit "hack: revert the optional ext_proc filter on the MITM leg"

bowei:
I kept the fixes separate to keep the revert commit clear.


haiyanmeng:
@bowei , thanks for fixing this on macOS.

I verified this change on my Linux OS, and it works correctly.

PR #1148, opened 2026-08-22
Rename controlapi.Service -> RpcService

thockin:
Having been reading this code for a while, and in light of changes we want to make, I think something like this will help clarify.

I am not particularly attached to the name or style `RpcService` - it's not usual Go convention but it is protobuf style.  If anyone feels strongly it could be `RPCService` or `ControlService` or something else.

@juli4n @laoj2

### DIFF
--- a/cmd/ateapi/internal/controlapi/actor.go
+++ b/cmd/ateapi/internal/controlapi/actor.go
@@ -34,7 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
-func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequest) (created *ateapipb.Actor, err error) {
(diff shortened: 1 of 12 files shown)


laoj2:
> I am not particularly attached to the name or style RpcService - it's not usual Go convention but it is protobuf style. If anyone feels strongly it could be RPCService or ControlService or something else.

I prefer RPCService, otherwise some golangci-lint configs will complain about it (although there are lots of `xyzHttp`, etc everywhere)

laoj2 (dismissed):

juli4n (dismissed):

juli4n:
It looks like a flake. Re-running the failed ones.

thockin:
Pushed with RpcService -> RPCService

thockin:
rebased

thockin:
w00t it passed CI

Issue #989, opened 2026-08-17
Non-atomic atespace existence check on Create


thockin:
@supreme-gg-gg 

In CreateActor we see:

```
func (s *Service) CreateActor(ctx context.Context, req *ateapipb.CreateActorRequest) (created *ateapipb.Actor, err error) {    
    if errs := validateCreateActorRequest(req); len(errs) > 0 {    
        return nil, toGRPCStatusError(errs)    
    }    

    // ...

    atespace := in.GetMetadata().GetAtespace()    
    name := in.GetMetadata().GetName()    
    
    // The atespace must already exist.    
    exists, err := s.persistence.AtespaceExists(ctx, atespace)    
    if err != nil {    
        return nil, fmt.Errorf("while checking atespace: %w", err)    
    }    
    if !exists {    
        return nil, status.Errorf(codes.FailedPrecondition, "Atespace %s not found", atespace)    
    }    

    // ...

    stored, err := s.persistence.CreateActor(ctx, actor)
```

The problem I see is that the atespace might exist when we check for it, but could be deleted before we actually store the actor.  Now we have an actor in an atespace which doesn't exist.

IIUC we need this to be in a txn or to fail if the atespace has been deleted (foreign key?).  This has to be part of the contract with the storage implementation, I think.

EItanya:
So the answer here is yes and no.  The `Service` layer has a race because it calls `exists, err := s.persistence.AtespaceExists(ctx, atespace)` without a lock. However, the DB layer does have the foreign key constraint which will fail the `Create`. Now that we are moving to PG we should unify this layer.

juli4n:
I think this check in the service layer should just go away now that we have posgresql. The contract with the store is that Create<Resource> should fail if the resource is atespaced and the atespace doesn't exist. There was no way to enforce this (at least in an scalable way) in Redis, but PostgreSQL enforces this via a FK.

thockin:
> I think this check in the service layer should just go away now that we have posgresql

Agree - we need to translate that particular failure (failed insert) into a good error message, but it's part of the storage contract.

PR #1149, opened 2026-08-23
Compartmentalize AtespaceExists()

thockin:
This needs to be a property of the storage layer.  It is in PG but not in redis.  This commit moves it out of the store interface and down into redis.  It's still not CORRECT but it constrains it.

This requires a lot of tests to create the atespace before creating other resources.  These tests are wrong in the face of a correct storage layer, anyway.

xref #989


laoj2 (dismissed):

thockin:
Rebased

laoj2 (approved):

PR #1154, opened 2026-08-23
fix(ateom): distinguish canceled runsc commands

### LINKED ISSUE #1153: Several e2e flakes cannot be diagnosed from the failed run
## Expected Behavior

A failed CI run carries enough to say why a test failed, without reproducing it.

## Actual Behavior

Working through the recent e2e flakes, several could not be explained from the failed run at all. The diagnostics were collected, the run was red, and the line that would have said why was not there.

Three cases from the last 80 failed runs:

- **`identity` failing with `502 bad gateway`** (4 runs). `internal/atunnel/ingress.go:148` logs a WARN carrying the real error next to every 502 it returns. None of the four runs has that line — the dump tails each pod at 300 lines and it had scrolled off.
- **`TestExternalVolumeLifecycle` failing with `DataLoss desc = actor … crashed`.** Same truncation; the crash reason is still unknown.
- **Ten runs failing with `while running \`runsc restore\`: signal: killed`**, two of them on `main`. Nothing in the run says what killed it.

The third is the worst of them, because the error looks specific and is not.

### What the error does not say

`runsc` is a subprocess of ateom, and `signal: killed` is what it reports in every case. A minimal program covering the three ways it dies:

```
scenario                            cmd.Wait()          ctx.Err()
context deadline expired            signal: killed      context deadline exceeded
context cancelled                   signal: killed      context canceled
outside SIGKILL, context untouched  signal: killed      <nil>
```

`ctx.Err()` is the only thing that separates them, and `cmd/ateom-gvisor/runsc.go` throws it away: every subcommand wraps the failure as `fmt.Errorf("while running \`runsc X\`: %w", err)`.

| date | branch | run |
|---|---|---|
| 2026-08-20 | non-http-egress | [32398330062](https://github.com/agent-substrate/substrate/actions/runs/32398330062) |
| 2026-08-20 | ctb-source | [32402389884](https://github.com/agent-substrate/substrate/actions/runs/32402389884) |
| 2026-08-20 | experiment-flag-extproc | [32428132897](https://github.com/agent-substrate/substrate/actions/runs/32428132897) |
| 2026-08-21 | extproc-example | [32431501158](https://github.com/agent-substrate/substrate/actions/runs/32431501158) |
| 2026-08-21 | **main** | [32436004611](https://github.com/agent-substrate/substrate/actions/runs/32436004611) |
| 2026-08-21 | ateomnet-ipv6-dualstack | [32448121452](https://github.com/agent-substrate/substrate/actions/runs/32448121452) |
| 2026-08-21 | e2e-ipfamily-dns | [32521617871](https://github.com/agent-substrate/substrate/actions/runs/32521617871) |
| 2026-08-21 | e2e-ingress-family | [32522618861](https://github.com/agent-substrate/substrate/actions/runs/32522618861) |
| 2026-08-21 | ctb-source | [32530161712](https://github.com/agent-substrate/substrate/actions/runs/32530161712) |
| 2026-08-22 | **main** | [32545349474](https://github.com/agent-substrate/substrate/actions/runs/32545349474) |

### What the dump does not collect

An OOM kill leaves a dmesg line and a container termination state. The dump has neither, nor node conditions, nor events — and it tails every pod at 300 lines.

### It also reads as a timeout when it is not

`RestoreWorkload` runs a deferred cleanup on a 30s budget (`cmd/ateom-gvisor/main.go:911`) and that time lands on the same RPC. Thirteen `elapsed-time` readings clustered between 30.21s and 30.52s, which reads as a 30-second timeout on the restore. The restore had already failed in the first 0.2–0.5s.

### What it cost

Attributing those ten runs took an instrumented ateom image, a dedicated VM, and an experiment that deleted a worker pod mid-restore to catch it. The answer — `ctx.Err() == context.Canceled`, arriving 112–442ms in — is a value the process already had in hand when it returned the error.

1. Delete a worker pod while an actor is resuming on it. Gracefully — `--grace-period=0` kills ateom before it can report anything.
2. The test fails with `while running \`runsc restore\`: signal: killed`.
3. Read the failed run: nothing says whether the context was cancelled or the kernel killed the process.

## Specifications

- Version: main @ 6210ad86
- Platform: kind, gVisor sandbox class. micro-VM is unaffected: it drives the VMM over a socket rather than a subprocess, and deliberately avoids `exec.CommandContext` for it (`cmd/ateom-microvm/internal/ch/restorefds.go:66`).

orangeCatDeveloper:
Fixes #1153

A CI run where runsc was killed says neither why it died nor enough to work it out afterwards.

`runsc` reports `signal: killed` for an OOM kill, an outside SIGKILL and `exec.CommandContext` acting on a done context alike, and every subcommand in `runsc.go` dropped `ctx.Err()`, the one thing that separates them. Meanwhile the dump tailed each pod at 300 lines and never looked at termination state, node conditions, events or dmesg, so the line that would have explained a failure had usually scrolled off.

`ctxErrorSuffix` names the context state next to the failed command, and stays silent when the error already carries it — which is what `Run` returns if the context died before `Start`. It says what was observed rather than naming a killer: a done context makes cancellation the strong candidate, but the process could still have been killed for another reason at the same moment.

On the CI side the workers now dump untailed, plus their previous container's log and `describe`, while the control plane keeps its tail: it runs the length of the job, and its history is not where a failing actor's evidence is. dmesg is grepped for OOM rather than tailed, since the kill line scrolls off a busy ring buffer, with a plain tail kept alongside for anything the grep misses.

Reading the two together: a context error points at cancellation, a dmesg record confirms an OOM, and neither means the cause is still open.

Before and after, for the three ways the subprocess dies:

```
scenario                            before                after
context deadline expired            signal: killed        signal: killed (context: context deadline exceeded)
context cancelled                   signal: killed        signal: killed (context: context canceled)
outside SIGKILL, context untouched  signal: killed        signal: killed
```

The dump was run against a live cluster: every section produced output, and the OOM grep found a real record a `tail -100` would have missed.

```
Memory cgroup out of memory: Killed process 1133445 (prometheus) total-vm:3625916kB, anon-rss:1039984kB
```

`hack/verify-all.sh` passes except `proto-fmt`, which needs a clang-format this machine does not have; no proto files are touched.

- [x] Tests pass


.github/workflows/pr-workflow.yaml
```diff
@@ -137,17 +137,37 @@ jobs:
       if: failure()
       run: |
         kubectl --context kind-kind get actortemplate,workerpool,pods -A -o wide || true
+        echo "=== node ==="
```
BenTheElder: this is way too much shell to put inline to the github workflow, where we can't shellcheck it etc. Please move this out to a script.

orangeCatDeveloper: Updated. Created a new script for this

PR #1147, opened 2026-08-22
test(e2e): isolate probe worker pools

orangeCatDeveloper:
Fixes #1146

`TestActorIdentity_AfterRestore_IsOwnID_NotGolden` is the most frequent failure in CI — 26 of the last 80 failed runs — and fails three different ways: a dial to a worker's `ateom.sock` that is not there, ``runsc restore`: signal: killed``, and a router 502/503.

All three are one bug. `probe.yaml.tmpl` gives every suite its own namespace, ActorTemplate and snapshot prefix through `${FIXTURE_SUFFIX}`, but the pool label and the `workerSelector` are the constant `workload: probe`. Selection is by label, not namespace, so `identity`, `egressmitm` and `imagevolume` share one 9-worker pool, and the first suite to tear its namespace down deletes workers the others are still running actors on. The victim's symptom is decided only by where it happened to be: between calls it cannot dial, mid-restore ateom's handler context is cancelled and `exec.CommandContext` SIGKILLs runsc, and after routing is up the router answers 5xx.

Putting the suffix on both labels restores the isolation the rest of the template already has — `parking_test.go:190` does the same with `demo: <namespace>`.

`capabilities.yaml.tmpl` and `probe-sized.yaml.tmpl` carry constant labels too, but each has a single caller today, so they cannot collide and are left alone.

Two identical 4-vCPU VMs, run concurrently, differing only in whether these two labels carry the suffix. A round is the full e2e suite; a hit is any of the three signatures:

```
unfixed   114 rounds   28 hits    dial 15   runsc-killed 6   router 5xx 7
fixed      92 rounds    0 hits    dial  0   runsc-killed 0   router 5xx 0

Fisher two-sided, any signature: p = 1.2e-8
```

The unfixed arm also carries CI's own evidence — an identity actor running on an imagevolume worker:

```
runs/32548404077
  identity_test.go:164: SuspendActor "probe-alpha": rpc error: code = Unavailable
    desc = "transport: Error while dialing: dial unix /var/lib/ateom-gvisor/ateoms/63f963d9-…/ateom.sock"
  ateapi: "Syncer: removing worker from store (pod deleted)"
    worker=63f963d9-…  pod=ate-e2e-probe-imagevolume/probe-6cfd69bbbd-xl29r
```

- [x] Tests pass
- [ ] Appropriate changes to documentation are included in the PR

Issue #1050, opened 2026-08-19
Multiple flaky networking tests


BenTheElder:
TestActorArbitraryPortAccess
TestActorDirectAccess
TestActorEgress
TestActorEgressHTTPS

/cc @LiorLieberman @bowei @keithmattix 

failing on main post merge:
https://github.com/agent-substrate/substrate/actions/runs/32200086998/job/95911826556

keithmattix:
#993 adds some retries to TestActorArbitraryPortAccess

BenTheElder:
thank you! everything passed when that merged, watching for if the next commit does as well. unfortunately we're starting to get throttled so it may be a bit. will close if we don't see more.

haiyanmeng:
From https://github.com/agent-substrate/substrate/actions/runs/32200086998/job/95911826556,
```
=== RUN   TestActorArbitraryPortAccess
    arbitraryport_test.go:51: creating actor networking-e2e/arbitraryport-1787098383986053833
    arbitraryport_test.go:51: resumed actor networking-e2e/arbitraryport-1787098383986053833
=== RUN   TestActorArbitraryPortAccess/regression:_default_port_still_reachable
    arbitraryport_test.go:58: default-port access through ingress returned HTTP 503; retrying...
    arbitraryport_test.go:58: default-port access through ingress returned HTTP 503; retrying...
    arbitraryport_test.go:58: default-port access through ingress returned HTTP 503; retrying...
    arbitraryport_test.go:58: default-port access through ingress returned HTTP 503; retrying...
    arbitraryport_test.go:58: default-port access through ingress returned HTTP 503; retrying...
    arbitraryport_test.go:61: default-port access through ingress succeeded; body: ok
=== RUN   TestActorArbitraryPortAccess/arbitrary_port_reachable
    arbitraryport_test.go:73: tunneled request returned HTTP 503, want 200; body: upstream connect error or disconnect/reset before headers. reset reason: connection timeout
=== RUN   TestActorArbitraryPortAccess/unlisted_port_rejected
    arbitraryport_test.go:130: tunneled request to an unlisted port correctly returned HTTP 503; body: upstream connect error or disconnect/reset before headers. reset reason: connection timeout
--- FAIL: TestActorArbitraryPortAccess (42.69s)
    --- PASS: TestActorArbitraryPortAccess/regression:_default_port_still_reachable (30.50s)
[...]

PR #991, opened 2026-08-17
atenet: never cancel an in-flight resume at the park budget

omeryahud:
The park budget doubled as the ResumeActor RPC deadline, so a resume that outlived it was cancelled mid-restore. ateapi claims the worker and persists RESUMING before the restore begins, rolls back neither on cancellation, and nothing reclaims a RESUMING actor on a live worker - a budget cancel therefore discarded the restore and stranded the worker, which is the TestRequestParking/ParkThenServed flake (#675).

The budget now bounds the retry loop only: an attempt still in flight when it elapses runs to completion on a non-cancellable context (bounded by ateapi's own server-side RPC deadline, with Envoy's ext_proc timeout unchanged as the client-side ceiling), and its real result is classified - an overshooting restore is served late instead of failed, and a late retryable error still surfaces as the capacity
503.

The e2e window widens to admit a served-late overshoot, and the test now pins the invariants the flake exposed: actor B reaches RUNNING and a follow-up request is served warm.

Fixes #675

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

PR #1160, opened 2026-08-24
fix: land the three root-caused flaky-test fixes (identity, parking, relay)

### LINKED ISSUE #675: E2E test flaky: TestRequestParking

### LINKED ISSUE #1100: E2E test flake: TestRelayIngressCancellationClosesBothSides

### LINKED ISSUE #1146: E2E test flake: TestActorIdentity_AfterRestore_IsOwnID_NotGolden — three suites share one worker pool
## Expected Behavior

Each suite that renders `internal/e2e/fixtures/probe/probe.yaml.tmpl` gets its own worker pool. `identity`, `egressmitm` and `imagevolume` already get their own namespace, ActorTemplate and snapshot prefix from `${FIXTURE_SUFFIX}`, so their workers should be private too.

## Actual Behavior

The pool label and the `workerSelector` are the constant `workload: probe`, so all three render the same pair:

```yaml
kind: WorkerPool
metadata:
  namespace: ate-e2e-probe${FIXTURE_SUFFIX}   # per suite
  labels:
    workload: probe                            # constant
---
kind: ActorTemplate
spec:
  workerSelector:
    matchLabels:
      workload: probe                          # constant
```

Worker selection is by label, not by namespace, so the three 3-replica pools behave as one 9-worker pool that any of the three can schedule onto. When a suite tears its namespace down it deletes workers another suite's actors are still running on.

Two runs where the CI logs show it directly — an identity actor on an imagevolume worker:

```
runs/32548404077   identity_test.go:164: SuspendActor "probe-alpha": rpc error: code = Unavailable
                   desc = "transport: Error while dialing: dial unix /var/lib/ateom-gvisor/ateoms/63f963d9-…/ateom.sock"
   ateapi          "Syncer: removing worker from store (pod deleted)"
                   worker=63f963d9-…  pod=ate-e2e-probe-imagevolume/probe-6cfd69bbbd-xl29r

runs/32417243258   identity_test.go:143: SuspendActor … same dial failure, worker=ddcba475-…
   ateapi 15:34:58 "Syncer: marking worker draining (pod deleting)"
                   worker=ddcba475-…  pod=ate-e2e-probe-imagevolume/probe-8588b687c8-qxrr6
   ateapi 15:34:59 "Syncer: removing worker from store (pod deleted)"   same worker
```

The victim's symptom depends only on where it was when the worker went away, which is why this reads as several unrelated flakes:

| when the worker is deleted | what the test reports |
|---|---|
| between calls | `dial unix .../ateom.sock: no such file or directory` |
| mid-restore | ateom's handler context is cancelled, `exec.CommandContext` SIGKILLs runsc → ``while running `runsc restore`: signal: killed`` |
| after routing is up | router `502 bad gateway` / `503` |

`TestActorIdentity_AfterRestore_IsOwnID_NotGolden` is the most frequent failure in CI — 26 of the last 80 failed runs — and splits into several causes; the three signatures above are this bug.

`parking_test.go:190` shows the shape the probe fixture is missing: `Labels: {"demo": nsObj.Name}` with a matching `MatchLabels`.

### Failing CI runs

Thirteen runs in the last 80 failed ones, spanning all three signatures. In each the diagnostics show the worker the identity actor was using belongs to the **imagevolume** namespace, with a Syncer line recording that pod being drained or removed:

| symptom | date | branch | run | worker's pod |
|---|---|---|---|---|
| `router-5xx` | 2026-08-20 | atepg-change-feed | [32400171542](https://github.com/agent-substrate/substrate/actions/runs/32400171542) | `ate-e2e-probe-imagevolume/probe-79d8fff675-bp8z8` |
| `runsc-killed` | 2026-08-20 | ctb-source | [32402389884](https://github.com/agent-substrate/substrate/actions/runs/32402389884) | `ate-e2e-probe-imagevolume/probe-9bbb6684c-v7f8p` |
| `dial` | 2026-08-20 | **main** | [32402609221](https://github.com/agent-substrate/substrate/actions/runs/32402609221) | `ate-e2e-probe-imagevolume/probe-dc94b795f-4dmxz` |
| `router-5xx` | 2026-08-20 | atenet-egress-dns-lookup-family | [32414610688](https://github.com/agent-substrate/substrate/actions/runs/32414610688) | `ate-e2e-probe-imagevolume/probe-676bb955df-d2hqx` |
| `dial` | 2026-08-20 | flaky-tests | [32417243258](https://github.com/agent-substrate/substrate/actions/runs/32417243258) | `ate-e2e-probe-imagevolume/probe-8588b687c8-qxrr6` |
| `runsc-killed` | 2026-08-21 | extproc-example | [32431501158](https://github.com/agent-substrate/substrate/actions/runs/32431501158) | `ate-e2e-probe-imagevolume/probe-86c866dbb6-j4pjk` |
| `dial` | 2026-08-21 | experiment-flag-extproc | [32436089702](https://github.com/agent-substrate/substrate/actions/runs/32436089702) | `ate-e2e-probe-imagevolume/probe-55694b85d6-pkz6z` |
| `runsc-killed` | 2026-08-21 | ateomnet-ipv6-dualstack | [32448121452](https://github.com/agent-substrate/substrate/actions/runs/32448121452) | `ate-e2e-probe-imagevolume/probe-578597c69b-xm6hw` |
| `router-5xx` | 2026-08-21 | obs-ladder-automation | [32498115146](https://github.com/agent-substrate/substrate/actions/runs/32498115146) | `ate-e2e-probe-imagevolume/probe-68677485df-cnxlv` |
| `runsc-killed` | 2026-08-21 | ctb-source | [32530161712](https://github.com/agent-substrate/substrate/actions/runs/32530161712) | `ate-e2e-probe-imagevolume/probe-5888c9c85d-ls4j5` |
| `runsc-killed` | 2026-08-22 | **main** | [32545349474](https://github.com/agent-substrate/substrate/actions/runs/32545349474) | `ate-e2e-probe-imagevolume/probe-8474965b8d-gdcrb` |
| `dial` | 2026-08-22 | atunnel-test-socket-path | [32548404077](https://github.com/agent-substrate/substrate/actions/runs/32548404077) | `ate-e2e-probe-imagevolume/probe-6cfd69bbbd-xl29r` |
| `router-5xx` | 2026-08-22 | wait-for-workers | [32549263943](https://github.com/agent-substrate/substrate/actions/runs/32549263943) | `ate-e2e-probe-imagevolume/probe-857d9f47b7-qfnft` |

Eleven more runs in the same window fail with one of these signatures but cannot be attributed from the run itself — the dump tails each pod at 300 lines and the Syncer lines had already scrolled off.

### Relationship to #1106

#1106 describes a different race — a pod that has an address before ateom listens on its socket — and cites [32402609221](https://github.com/agent-substrate/substrate/actions/runs/32402609221). That run is in the table above: the worker identity failed to dial is `ate-e2e-probe-imagevolume/probe-dc94b795f-4dmxz`, and ateapi logged `Syncer: marking worker draining (pod deleting)` for it 34ms before the dial. The socket was gone because the pod was being deleted, not because it had not been created yet.

That does not disprove #1106's race, which is real in principle — `isWorkerEligible` did accept a pod on `PodIP` alone. It means the run offered as its evidence is this bug, so #1107 landing will not make this test green.

1. Run the full e2e suite on one cluster so `identity`, `egressmitm` and `imagevolume` overlap (`hack/run-e2e-kind.sh -v -args --no-color`).
2. While a suite is running, `kubectl get workerpool -A -o jsonpath='{range .items[*]}{.metadata.namespace}/{.metadata.name} {.metadata.labels}{"\n"}{end}'` — all three probe pools carry `{"workload":"probe"}`.
3. Repeat. `TestActorIdentity_AfterRestore_IsOwnID_NotGolden` fails on roughly a quarter of the rounds, with one of the three signatures above.

Measured on two identical 4-vCPU VMs running concurrently, the only difference being whether the two labels carry `${FIXTURE_SUFFIX}`:

```
unfixed   114 rounds   28 rounds hit one of the three signatures
fixed      92 rounds    0
                        Fisher two-sided p = 1.2e-8
```

Per signature: dial 15/114 → 0/92 (p=1.6e-4), runsc killed 6/114 → 0/92 (p=3.4e-2), router 5xx 7/114 → 0/92 (p=1.8e-2).

## Specifications

- Version: main @ 6210ad86
- Platform: kind, gVisor sandbox class

aditya-shantanu:
Fixes #675, fixes #1100, fixes #1146; addresses the CI flake in #1106.

## Why this PR

Flaky tests are the single biggest drag on this repo's velocity right now: the three flakes fixed here account for the majority of red CI runs over the last 7 days (identity: 35 failures, parking: 32, relay: 9 — from the flake dashboard's cross-PR analysis of ~630 runs). Every red run costs a contributor a rebase-and-rerun cycle and costs reviewers signal. **This PR consolidates the three root-caused, in-flight fixes into one change to get CI green now and unblock the community — the goal is velocity, not authorship.**

## Credit where it's due

All three fixes were root-caused and written by others; this PR adopts them onto latest main with their tests, unchanged in substance. Each commit carries a `Co-authored-by` trailer:

| Commit | Original PR | Author | Root cause |
|---|---|---|---|
| e2e: give each probe fixture its own worker pool | #1147 | @orangeCatDeveloper | identity/egressmitm/imagevolume suites share one `workload: probe` pool label; cross-suite selection under concurrent suite processes dials workers that are not there |
| atenet: never cancel an in-flight resume at the park budget | #991 | @omeryahud | the park budget doubled as the ResumeActor RPC deadline; a mid-restore cancel strands a RESUMING actor on a live worker |
| atunnel: close the relay's both ends before returning | #1101 | @orangeCatDeveloper | the relay closed both ends from a `context.AfterFunc` goroutine the test never waits for |

@Stevenjin8's #1107 correctly diagnosed the ateom readiness race in #1106; the control-plane readiness gap it targets remains real and open — this PR only removes the e2e-fixture contention that makes it fire constantly in CI.

If maintainers prefer to land the original PRs individually instead, closing this one is completely fine — the point is that the fixes land somewhere, soon.

## Evidence the flakes are actually fixed

**TestRelayIngressCancellationClosesBothSides (unit, `-race`):**
- Unpatched main, `-count=3000`: **83 failures (2.8%)** — matches the 2.9% observed across 308 CI runs this week
- This branch, `-count=10000`: **0 failures**

**TestRequestParking (park-budget cancellation):**
- The new `InFlightAttemptRunsToCompletion` and `LateRetryableErrorIsBudgetExhaustion` unit tests (from #991) encode the exact failure mode from #675 and pass under `go test -race -count=100 ./cmd/atenet/internal/router/ingress/`
- The pre-fix behavior (budget cancelling the in-flight RPC) is deterministically reproduced by the old test it replaces

**TestActorIdentity_AfterRestore_IsOwnID_NotGolden (probe pool isolation):**
- Not reproducible outside CI (needs concurrent suite processes on a contended kind node), so verified statically: `${FIXTURE_SUFFIX}` is always `-<suite>` (internal/e2e/sandbox.go:189,201 — never empty), `probe-sized` already uses its own label, and no other manifest or selector references `workload: probe`. #1147's CI data shows all three failure signatures (missing `ateom.sock`, `runsc restore` killed, router 502/503) trace to cross-suite pool sharing; per-suite labels make the selector suite-local by construction
- The definitive check is this PR's own CI plus the flake dashboard's 7-day window after merge — I will report the post-merge rates on #1106

Also run: `go build ./...`, `go vet` and the full `-race` suites of both touched packages — all green.

## What this PR deliberately does NOT fix

`TestActorEgressHTTPS` (#1050, 4.6% this week, below the 5% flake threshold) has no root-caused fix yet — the 503 `upstream connect error` path needs investigation in a live cluster. #1103 (@orangeCatDeveloper) tightens the related `TestActorArbitraryPortAccess` assertion so those 503s stop passing silently; it should land after #1050's cause is fixed, or it converts hidden flakiness into visible red.

## Update (post-CI investigation)

The first e2e runs failed on `TestRequestParking/ParkThenServed` (micro-VM lane). Investigation showed this is the **pre-existing dominant mode** of #675 — identical failures in main-era runs 32305728993 / 32397291519 / 32487958077 — not a regression: on micro-VM, `SuspendActor` returns before the snapshot upload completes, so the worker legitimately isn't free within the 5s park budget and the router's 503 is correct behavior. #991 fixes the *other* (mid-restore cancellation/stranding) mode. Commit d637690d makes the subtest retry while the worker is still freeing; a stranded worker still fails every attempt, so the regression stays pinned.

**Additional validation:**
- CI e2e-test now **passes both lanes** (run 32754691017)
- Local kind cluster built from this branch: parking suite **10/10 consecutive passes**; identity + egressmitm + imagevolume run **concurrently** (the exact contention behind the identity flake) × 3 iterations — **9/9 suite passes**

### DIFF
--- a/internal/e2e/suites/parking/parking_test.go
+++ b/internal/e2e/suites/parking/parking_test.go
@@ -115,10 +115,28 @@ func TestRequestParking(t *testing.T) {
 		if !strings.Contains(res.body, "hello from") {
 			t.Errorf("parked request body = %q, want the counter greeting", res.body)
 		}
-		if elapsed >= routerParkBudget+2*time.Second {
-			t.Errorf("parked request served after %v, want inside the %v budget window", elapsed, routerParkBudget)
+		// The budget bounds retries, not a committed resume: a restore that
+		// overshoots the budget under CI contention is served rather than
+		// canceled, so the window allows for the overshoot while staying
+		// under Envoy's ext_proc timeout (budget + 5s).
+		if elapsed >= routerParkBudget+4*time.Second {
+			t.Errorf("parked request served after %v, want inside the %v budget window (+overshoot)", elapsed, routerParkBudget)
 		}
 		t.Logf("parked request served after %v", elapsed)
+
+		// The flake's root cause stranded actors in RESUMING with the worker
+		// claimed (#675): pin that B really converges and a follow-up request
+		// is served warm — a stranded actor would 503 it.
+		waitForActorState(ctx, t, clients, actorB, ateapipb.ActorState_ACTOR_STATE_RUNNING)
+		followUp, err := router.Get(ctx, resources.ActorRef{Atespace: parkingAtespace, Name: actorB}, "/")
+		if err != nil {
+			t.Fatalf("follow-up request failed transport-level: %v", err)
+		}
+		followUpBody, _ := io.ReadAll(followUp.Body)
+		followUp.Body.Close()
+		if followUp.StatusCode != http.StatusOK {
+			t.Errorf("follow-up request: status = %d (body %q), want 200 from the resumed actor", followUp.StatusCode, string(followUpBody))
+		}
 
 		// The slot must be released once served.
 		waitForParkedCount(ctx, t, statusz, func(active int) bool { return active == 0 })
--- a/internal/e2e/suites/parking/parking_test.go
+++ b/internal/e2e/suites/parking/parking_test.go
@@ -80,34 +80,52 @@ func TestRequestParking(t *testing.T) {
 		resumeActor(ctx, t, clients, actorA)
 		waitForActorState(ctx, t, clients, actorA, ateapipb.ActorState_ACTOR_STATE_RUNNING)
 
-		// Request actor B: the pool is full, so the request parks.
+		// Request actor B: the pool is full, so the request parks. Freeing
+		// the worker is asynchronous — SuspendActor(A) returns before the
+		// suspend completes, and on the micro-VM class the snapshot upload
+		// routinely outlives the 5s park budget under CI contention. A
+		// budget-exhausted 503 while the suspend is still in flight is the
+		// router behaving correctly, so the request is retried: each attempt
+		// parks anew, and the suspend's completion lets one of them resume B.
+		// A stranded worker (#675's root cause) fails every attempt, so the
+		// regression this subtest pins still fails it.
 		type result struct {
 			resp *http.Response
 			body string
 			err  error
 		}
 		resCh := make(chan result, 1)
-		start := time.Now()
-		go func() {
-			resp, err := router.Get(ctx, resources.ActorRef{Atespace: parkingAtespace, Name: actorB}, "/")
-			var body string
-			if err == nil {
-				b, _ := io.ReadAll(resp.Body)
-				resp.Body.Close()
-				body = string(b)
+		var res result
+		var elapsed time.Duration
+		for attempt := 1; ; attempt++ {
+			start := time.Now()
+			go func() {
+				resp, err := router.Get(ctx, resources.ActorRef{Atespace: parkingAtespace, Name: actorB}, "/")
+				var body string
+				if err == nil {
+					b, _ := io.ReadAll(resp.Body)
+					resp.Body.Close()
+					body = string(b)
+				}
+				resCh <- result{resp, body, err}
+			}()
+			if attempt == 1 {
+				// Free the worker only once the request is observably parked —
+				// the statusz gauge, not a sleep, is the synchronization point.
+				waitForParkedCount(ctx, t, statusz, func(active int) bool { return active >= 1 })
+				suspendActor(ctx, t, clients, actorA)
 			}
-			resCh <- result{resp, body, err}
-		}()
-
-		// Free the worker only once the request is observably parked — the
-		// statusz gauge, not a sleep, is the synchronization point.
-		waitForParkedCount(ctx, t, statusz, func(active int) bool { return active >= 1 })
-		suspendActor(ctx, t, clients, actorA)
-
-		res := <-resCh
-		elapsed := time.Since(start)
-		if res.err != nil {
-			t.Fatalf("parked request failed transport-level: %v", res.err)
+			res = <-resCh
+			elapsed = time.Since(start)
+			if res.err != nil {
+				t.Fatalf("parked request failed transport-level: %v", res.err)
+			}
+			if res.resp.StatusCode == http.StatusServiceUnavailable &&
+				strings.Contains(res.body, "no free workers available") && attempt < 3 {
+				t.Logf("attempt %d budget-exhausted while the worker was still freeing (503 after %v); retrying", attempt, elapsed)
+				continue
+			}
+			break
 		}
 		if res.resp.StatusCode != http.StatusOK {
 			t.Fatalf("parked request: status = %d (body %q), want 200", res.resp.StatusCode, res.body)
--- a/internal/e2e/suites/parking/parking_test.go
+++ b/internal/e2e/suites/parking/parking_test.go
@@ -133,13 +133,9 @@ func TestRequestParking(t *testing.T) {
 		if !strings.Contains(res.body, "hello from") {
 			t.Errorf("parked request body = %q, want the counter greeting", res.body)
 		}
-		// The budget bounds retries, not a committed resume: a restore that
-		// overshoots the budget under CI contention is served rather than
-		// canceled, so the window allows for the overshoot while staying
-		// under Envoy's ext_proc timeout (budget + 5s).
-		if elapsed >= routerParkBudget+4*time.Second {
-			t.Errorf("parked request served after %v, want inside the %v budget window (+overshoot)", elapsed, routerParkBudget)
-		}
+		// No upper bound on elapsed here: a 200 proves the router served the
+		// request before Envoy's ext_proc timeout, and a slow-but-successful
+		// restore under CI contention is a pass, not a flake.
 		t.Logf("parked request served after %v", elapsed)
 
 		// The flake's root cause stranded actors in RESUMING with the worker
--- a/cmd/atenet/internal/router/ingress/resumer.go
+++ b/cmd/atenet/internal/router/ingress/resumer.go
@@ -182,15 +182,24 @@ func (r *ActorResumer) ResumeActor(ctx context.Context, actorRef resources.Actor
 		// one control-plane RPC per hot actor (see docs/request-parking.md).
 		bgCtx, bgCancel := context.WithTimeout(context.Background(), r.budget)
 		defer bgCancel()
+		// The budget bounds the RETRY LOOP only — it never cancels an
+		// in-flight ResumeActor. ateapi durably claims the worker and marks
+		// the actor RESUMING before the expensive snapshot restore begins,
+		// rolls back neither on cancellation, and nothing reclaims a RESUMING
+		// actor whose worker pod is alive — a budget cancel therefore throws
+		// the restore away and strands the worker (#675). An attempt still
+		// running when the budget elapses is waited for and its real result
+		// classified below; ateapi's own server-side RPC deadline bounds it.
+		attemptCtx := context.WithoutCancel(bgCtx)
 
 		backoff := r.backoff
 
 		var resumeResp *ateapipb.ResumeActorResponse
 		var lastRetryErr error
 
-		err := wait.ExponentialBackoffWithContext(bgCtx, backoff, func(ctx context.Context) (bool, error) {
+		err := wait.ExponentialBackoffWithContext(bgCtx, backoff, func(context.Context) (bool, error) {
 			var err error
-			resumeResp, err = r.apiClient.ResumeActor(ctx, &ateapipb.ResumeActorRequest{
+			resumeResp, err = r.apiClient.ResumeActor(attemptCtx, &ateapipb.ResumeActorRequest{
 				Actor: actorRef.ToObjectRef(),
 			})
 			if err == nil {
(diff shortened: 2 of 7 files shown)


aditya-shantanu:
The e2e failure here is worth a close look — it is **not** a regression from this PR, and it exposed that #991 fixes only part of the #675 flake:

**What failed:** `TestRequestParking/ParkThenServed` (micro-VM lane): `parking_test.go:113: 503 "no free workers available", want 200`, with `BudgetExhaustion` failing as a cascade of the shared subtest state.

**Why it's pre-existing:** the identical failure (same line, same 503 body, same lane) appears in main-era runs [32305728993](https://github.com/agent-substrate/substrate/actions/runs/32305728993), [32397291519](https://github.com/agent-substrate/substrate/actions/runs/32397291519) and [32487958077](https://github.com/agent-substrate/substrate/actions/runs/32487958077) — this is the *dominant* mode of the 13.5% micro-VM flake. Mechanism: `SuspendActor(A)` returns before the suspend completes, and on micro-VM the guest snapshot upload routinely outlives the 5s park budget under CI contention. The router's budget-exhausted 503 is then **correct product behavior**; the test's assumption that the worker always frees within one budget is the bug. #991 fixes the other (mid-restore cancellation → stranded worker) mode.

**Fix pushed (d637690d):** `ParkThenServed` now retries the parked request (bounded at 3 attempts) while the worker is still freeing — each attempt parks anew, and a genuinely stranded worker (#675's root cause) still fails every attempt plus the follow-up assertions, so the regression stays pinned.

**Validation:** 10/10 consecutive suite passes against a local kind cluster built from this branch (gVisor lane; micro-VM needs KVM, so this PR's CI is the venue for that lane). The relay fix separately holds at 0 failures over 10,000 `-race` iterations vs 83/3,000 on main.

internal/e2e/suites/parking/parking_test.go
```diff
@@ -80,46 +80,82 @@ func TestRequestParking(t *testing.T) {
...
 		}
 		if res.resp.StatusCode != http.StatusOK {
 			t.Fatalf("parked request: status = %d (body %q), want 200", res.resp.StatusCode, res.body)
 		}
 		if !strings.Contains(res.body, "hello from") {
 			t.Errorf("parked request body = %q, want the counter greeting", res.body)
 		}
-		if elapsed >= routerParkBudget+2*time.Second {
-			t.Errorf("parked request served after %v, want inside the %v budget window", elapsed, routerParkBudget)
+		// The budget bounds retries, not a committed resume: a restore that
```
yufan-su: should we remove this check? 

in the case when the resume eventually succeeded within the ext_proc timeout, but takes more than 9s, it looks like an expected successful run and should not fail the e2e. If it takes longer and timed out, we already checks for the resp.StatusCode above.

omeryahud:
Thanks for consolidating all this @aditya-shantanu 
parking-related fixes lgtm

yufan-su (dismissed):

internal/e2e/suites/parking/parking_test.go
aditya-shantanu: Agreed — a 200 already proves the router beat the ext_proc timeout, and failing a slow-but-successful restore would just be a new flake. Removed in cefbb076.

PR #1162, opened 2026-08-24
Introduce a `apitool` CLI for the substrate API.

juli4n:
The idea is to bundle automation related to the API here. For now, there is only a single `validate` command that runs a set of lint rules to enforce that the API adheres to the style guide. This should help us prevent regressions / deviations when modifying it.

These are not enforced as part of the presubmit checks, although we might want to enforce that soon (right after we close the existing gaps? see #1163). There is no support for violation exemptions for now.

In the future, we could add another command to, for example, generate a reference documentation for the whole API.


tools/apitool/internal/cmd/root.go
```diff
@@ -0,0 +1,32 @@
...
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "apitool",
+	Short: "CLI tool for the validating the Substrate API definition",
```
BenTheElder, line 23: This sounds like an internal dev tool, maybe don't expose it for import by users at all?

juli4n: Done. We still need to fix this in other tools/s though. Will file an issue.

tools/apitool/README.md
```diff
@@ -0,0 +1,13 @@
+# Substrate API Tool (`apitool`)
+
+The `apitool` is a CLI that automates tasks related to the Substrate API.
+
+The only command it has today is a command to run a set of validations that enforce
+that the API adheres to the Substrate API style guide.
```
BenTheElder, line 6: Any reason not to add a verify script to enforce it?

BenTheElder:
> These are not enforced as part of the presubmit checks, although we might want to enforce that soon (right after we close the existing gaps? see https://github.com/agent-substrate/substrate/issues/1163). There is no support for violation exemptions for now.

We should add even just a hardcoded exception list that will be removed over time, so there doesn't need to be a concept in the future to allow this, but so we can start ratcheting fixes?

Doing everything in bulk is harder.

BenTheElder (commented):
I can't help but feel like there's a lot of repetitive and verbose code looking for a better abstraction but I'd need more time to think about how we might approach that

golangci-lint like kube-api-linter perhaps

juli4n:
> I can't help but feel like there's a lot of repetitive and verbose code looking for a better abstraction but I'd need more time to think about how we might approach that
> 
> golangci-lint like kube-api-linter perhaps

I'm sure abstractions will emerge (e.g. some sort of visitor over the API model, etc). I would let them emerge though, no need to overthink it initially.

juli4n:
> > These are not enforced as part of the presubmit checks, although we might want to enforce that soon (right after we close the existing gaps? see #1163). There is no support for violation exemptions for now.
> 
> We should add even just a hardcoded exception list that will be removed over time, so there doesn't need to be a concept in the future to allow this, but so we can start ratcheting fixes?
> 
> Doing everything in bulk is harder.

I've been fixing violations incrementally. I don't want to make this PR even larger (it's already fairly large) by adding presubmit automation. Will add that as a follow up PR.

BenTheElder:
> I'm sure abstractions will emerge (e.g. some sort of visitor over the API model, etc). I would let them emerge though, no need to overthink it initially

selfishly: it's a LOT of mechanical code to read through is all, I didn't feel confident I caught everything through the boilerplate

BenTheElder (approved):

PR #1170, opened 2026-08-24
postgres: cert reload

### LINKED ISSUE #1169: postgres cert expires
## Expected Behavior

cluster should remain stable when installed for O(days)

## Actual Behavior

cluster breaks when configured with postgres storage, because rotated certs are not picked up unless you manually roll the pod

1. Install substrate with postgres enabled
1. Wait O(days)
1. Observe broken cluster

## Specifications

This part of the template needs revisiting .. we don't have versions yet.

- Version:
- Platform:

BenTheElder:
Fixes https://github.com/agent-substrate/substrate/issues/1169

- [x] Tests pass

### DIFF
--- a/manifests/ate-install/postgres.yaml
+++ b/manifests/ate-install/postgres.yaml
@@ -21,17 +21,55 @@ data:
   postgresql.conf: |
     listen_addresses = '*'
     ssl = on
-    ssl_cert_file = '/run/tls/credential-bundle.pem'
-    ssl_key_file = '/run/tls/credential-bundle.pem'
+    ssl_cert_file = '/run/servicedns.podcert.ate.dev/credential-bundle.pem'
+    ssl_key_file = '/run/servicedns.podcert.ate.dev/credential-bundle.pem'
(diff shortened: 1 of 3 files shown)


manifests/ate-install/postgres.yaml
```diff
@@ -63,28 +113,47 @@ spec:
       labels:
         app: postgres
     spec:
-      # The official postgres image requires ssl_key_file to contain only a
-      # private key with restrictive (0600) permissions. The projected
-      # podCertificate source is world-readable, so copy it into a writable
-      # emptyDir and fix permissions/ownership before postgres starts
+      # PostgreSQL requires ssl_key_file to be unreadable by anyone but the
```
ahmedtd: Pod Certificates can write the private key and certificate chain to separate files, if that's easier.

ahmedtd (dismissed):

manifests/ate-install/postgres.yaml
BenTheElder: Thanks, reworking.

PR #1174, opened 2026-08-24
Reject object updates containing unknown fields

laoj2:
https://github.com/agent-substrate/substrate/issues/1011

An update containing unknown fields during a RMW indicates there's a version skew between client and server. 

So, the server enforces a policy to fail explicitly when it receives an unknown proto field: so that  it does not silently drop fields that the user intended to set but the current server version cannot process 

This also helps to avoid behaviour drift during rolling upgrades: During rolling upgrades, different components (like the data plane and control plane) can be on different versions. If an older server replica . If an older server replica processes a request coming from a newer client, it will reject it (it doesn't know how to validate/handle it) and it's the client's responsibility to retry.

- [x] Tests pass


HavenXia (dismissed):

HavenXia:
Thanks, this matches exactly the 2nd proposal in [substrate upgrade proposal](https://docs.google.com/document/d/1JduAyGZFyqdNp4UhKiv0EN-is3iW5tf5BwGTWbt2ouI/edit?resourcekey=0-MXG12QCkleIxhY7cOB7g6A&tab=t.0#bookmark=id.kjmz6zkxy24v)

HavenXia (approved):

PR #1181, opened 2026-08-25
Accept both Google LLC and The Agent Substrate Authors in boilerplate

brendandburns:
## Summary

This change updates the boilerplate verification to accept either of the approved copyright attribution forms:

- `Copyright YYYY Google LLC`
- `Copyright YYYY The Agent Substrate Authors`

## Why

The repo now allows either owner attribution, and the boilerplate checker should enforce that both valid forms pass without rejecting the other.

I retained the Google LLC so that we didn't have to re-copyright all the existing files.

## Verification

- Ran the focused matcher assertions for both allowed strings and an invalid holder.
- Ran `hack/verify/boilerplate.sh` successfully.

### DIFF
--- a/hack/util/verify-boilerplate.py
+++ b/hack/util/verify-boilerplate.py
@@ -37,7 +37,7 @@ ASLV2_HEADER = [
 ]
 
 def match_copyright(line):
-    match = re.search(r"Copyright\s+(\d{4})\s+Google LLC", line)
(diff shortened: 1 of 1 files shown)


BenTheElder:
I see @thockin suggested this, I know we had to do the current approach for launch but beyond that blissfully unaware / defer to tim.

Triaging to @thockin for review.

thockin:
FTR I am awaiting a consult from the powers that be about this.  I think it is correct.

thockin:
LGTM

PR #1183, opened 2026-08-25
atenet: h2 on the HTTPS ingress and mirror the protocol to actors

LiorLieberman:
atenet now mirrors the protocol it received. adds support for grpc.

h2 on https is gated with a flag. 

cc: @bowei @rakyll 

- [x] Tests pass


cmd/atenet/internal/router/xds.go
```diff
@@ -1192,6 +1188,13 @@ func buildDownstreamTlsTransportSocket() *corev3.TransportSocket {
 func (x *XdsServer) buildHttpsListener() *listenerv3.Listener {
 	hcm := x.buildHcm("ingress_https", true)
 
+	// gRPC requires a negotiated "h2"; http/1.1 keeps plain HTTPS clients
+	// working alongside it.
+	var alpn []string
+	if x.httpsH2 {
+		alpn = []string{"h2", "http/1.1"}
```
rakyll: I wonder what happens when you curl an HTTP/1.1 actor once you advertise this ALPN. I think some clients like curl assumes the endpoint is H2 even though the actor may have HTTP/1.1 endpoints.

We want actors to be able to serve a mixture of protocols from the same port.

cmd/atenet/internal/router/xds.go
```diff
@@ -757,22 +771,23 @@ func (x *XdsServer) buildOriginalDstCluster() *clusterv3.Cluster {
...
+	// Mirror the downstream protocol on the atunnel connection: HTTP/1.1
+	// requests proxy as HTTP/1.1, while HTTP/2 requests (gRPC) stay HTTP/2
+	// end to end, which trailer-based protocols require. atunnel forwards
+	// to the actor with the same protocol it receives.
+	cluster.TypedExtensionProtocolOptions = map[string]*anypb.Any{
+		httpProtocolOptionsName: newAny(&httpv3.HttpProtocolOptions{
+			UpstreamProtocolOptions: &httpv3.HttpProtocolOptions_UseDownstreamProtocolConfig{
+				UseDownstreamProtocolConfig: &httpv3.HttpProtocolOptions_UseDownstreamHttpConfig{
+					HttpProtocolOptions:  &corev3.Http1ProtocolOptions{},
+					Http2ProtocolOptions: &corev3.Http2ProtocolOptions{},
```
rakyll: Don't we need this line only if h2 is enabled?

rakyll (dismissed):
I think the current way we set ALPN is not allowing us to serve a mixture of HTTP/1.1 and H2 traffic. Left some comments.

cmd/atenet/internal/router/xds.go
aojea: I'm looking into the setup — @LiorLieberman could you confirm how many hops exist and where request reassembly happens?

This is what gemini tells me is the setup, is this correct?

- Hop 1 (Client to Envoy atenet-router): Envoy terminates downstream TLS and the downstream HTTP connection (HTTP/1.1 or HTTP/2). It parses the request completely (headers, path, query, payload stream), executes L7 filters (such as ext_proc for actor activation/routing), and selects the upstream cluster.
- Hop 2 (Envoy to atunnel in the worker pod): Envoy dials atunnel over a new, independent mTLS connection. Because of UseDownstreamProtocolConfig, Envoy speaks HTTP/2 on this leg if the client negotiated HTTP/2.

- Hop 3 (atunnel to Actor process at 127.0.0.1:8080): atunnel is also an L7 reverse proxy (built with Go's httputil.ReverseProxy). It terminates mTLS from Envoy and dials the actor on a local plaintext TCP connection.

Because every hop is an independent L7 HTTP connection then we can always choose where to change the protocol transparently and avoid the requirement of 1to1 match between client and backend.

@LiorLieberman the test you set up is allowing the backend to answer in both

In `TestProtocolMirrorTransport`, the test backend was configured with:
```go
backendProtocols.SetUnencryptedHTTP2(true)
```

You should include a table test with http1 only backends , also grpc , if you apply something like the diff below you can get more coverage and you can detect exactly when you are requesting grpc or http2, it also will fail for the reasons @rakyll indicate above, is better to not assume http2 == grpc

```diff
diff --git a/internal/atunnel/ingress.go b/internal/atunnel/ingress.go
index a813b96d..aea836e5 100644
--- a/internal/atunnel/ingress.go
+++ b/internal/atunnel/ingress.go
@@ -181,10 +181,34 @@ func NewServer(cfg Config) (*Server, error) {
        return s, nil
 }
 
-// protocolMirrorTransport forwards a request to the actor with the same
-// HTTP major version it arrived with: HTTP/1.1 stays HTTP/1.1, and HTTP/2
-// (gRPC) goes out as cleartext prior-knowledge HTTP/2, preserving trailers
-// and full-duplex streaming end to end.
+// isGRPC reports whether an incoming HTTP request conforms to the gRPC over
+// HTTP/2 specification: HTTP/2 framing, POST method, and a Content-Type of
+// "application/grpc" (optionally with a "+<subtype>" or ";<parameters>").
+// See https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+func isGRPC(r *http.Request) bool {
+       if r.ProtoMajor != 2 || r.Method != http.MethodPost {
+               return false
+       }
+       ct := strings.ToLower(r.Header.Get("Content-Type"))
+       const base = "application/grpc"
+       if ct == base {
+               return true
+       }
+       if strings.HasPrefix(ct, base) {
+               switch ct[len(base)] {
+               case '+', ';':
+                       return true
+               }
+       }
+       return false
+}
+
+// protocolMirrorTransport forwards a request to the actor based on whether the
+// request requires cleartext HTTP/2 (gRPC) or standard HTTP/1.1. Plain HTTP
+// requests (e.g. REST calls, web traffic) are translated to HTTP/1.1 even if
+// they arrived over HTTP/2 at the edge, ensuring compatibility with standard
+// HTTP/1.1 actors. gRPC requests (which require HTTP/2 trailers and streaming)
+// go out as cleartext prior-knowledge HTTP/2 (h2c).
 type protocolMirrorTransport struct {
        h1, h2c *http.Transport
 }
@@ -199,7 +223,7 @@ func newProtocolMirrorTransport() protocolMirrorTransport {
 }
 
 func (t protocolMirrorTransport) RoundTrip(r *http.Request) (*http.Response, error) {
-       if r.ProtoMajor == 2 {
+       if isGRPC(r) {
                return t.h2c.RoundTrip(r)
        }
        return t.h1.RoundTrip(r)
diff --git a/internal/atunnel/ingress_test.go b/internal/atunnel/ingress_test.go
index 695fb877..6623a208 100644
--- a/internal/atunnel/ingress_test.go
+++ b/internal/atunnel/ingress_test.go
@@ -29,11 +29,16 @@ import (
        "net"
        "net/http"
        "net/http/httptest"
+       "net/http/httputil"
        "net/url"
        "os"
        "path/filepath"
        "testing"
        "time"
+
+       "google.golang.org/grpc"
+       "google.golang.org/grpc/credentials/insecure"
+       healthpb "google.golang.org/grpc/health/grpc_health_v1"
 )
 
 func TestRelayIngressWithHalfClose(t *testing.T) {
@@ -617,47 +622,254 @@ func tlsHandshake(serverConfig, clientConfig *tls.Config) (serverErr, clientErr
        return serverErr, clientErr
 }
 
-// TestProtocolMirrorTransport verifies the upstream leg mirrors the inbound
-// HTTP major version: HTTP/1.1 stays HTTP/1.1 and HTTP/2 goes out as
-// cleartext prior-knowledge HTTP/2, which proxied gRPC requires.
-func TestProtocolMirrorTransport(t *testing.T) {
-       protoSeen := make(chan string, 2)
-       backendProtocols := new(http.Protocols)
-       backendProtocols.SetHTTP1(true)
-       backendProtocols.SetUnencryptedHTTP2(true)
+type testHealthServer struct {
+       healthpb.UnimplementedHealthServer
+}
+
+func (s *testHealthServer) Check(ctx context.Context, req *healthpb.HealthCheckRequest) (*healthpb.HealthCheckResponse, error) {
+       return &healthpb.HealthCheckResponse{Status: healthpb.HealthCheckResponse_SERVING}, nil
+}
+
+// TestIsGRPC tests the request detection heuristics specified by the gRPC over
+// HTTP/2 specification (PROTOCOL-HTTP2.md).
+func TestIsGRPC(t *testing.T) {
+       tests := []struct {
+               name        string
+               method      string
+               protoMajor  int
+               contentType string
+               want        bool
+       }{
+               {
+                       name:        "standard gRPC base content-type",
+                       method:      http.MethodPost,
+                       protoMajor:  2,
+                       contentType: "application/grpc",
+                       want:        true,
+               },
+               {
+                       name:        "gRPC with protobuf sub-type",
+                       method:      http.MethodPost,
+                       protoMajor:  2,
+                       contentType: "application/grpc+proto",
+                       want:        true,
+               },
+               {
+                       name:        "gRPC with json sub-type",
+                       method:      http.MethodPost,
+                       protoMajor:  2,
+                       contentType: "application/grpc+json",
+                       want:        true,
+               },
+               {
+                       name:        "gRPC with parameters",
+                       method:      http.MethodPost,
+                       protoMajor:  2,
+                       contentType: "application/grpc; charset=utf-8",
+                       want:        true,
+               },
+               {
+                       name:        "case-insensitive media type",
+                       method:      http.MethodPost,
+                       protoMajor:  2,
+                       contentType: "Application/GRPC+Proto",
+                       want:        true,
+               },
+               {
+                       name:        "wrong method (GET)",
+                       method:      http.MethodGet,
+                       protoMajor:  2,
+                       contentType: "application/grpc",
+                       want:        false,
+               },
+               {
+                       name:        "wrong method (PUT)",
+                       method:      http.MethodPut,
+                       protoMajor:  2,
+                       contentType: "application/grpc",
+                       want:        false,
+               },
+               {
+                       name:        "wrong protocol major version (HTTP/1.1)",
+                       method:      http.MethodPost,
+                       protoMajor:  1,
+                       contentType: "application/grpc",
+                       want:        false,
+               },
+               {
+                       name:        "non-gRPC JSON request",
+                       method:      http.MethodPost,
+                       protoMajor:  2,
+                       contentType: "application/json",
+                       want:        false,
+               },
+               {
+                       name:        "grpc-web content type (not native gRPC over h2)",
+                       method:      http.MethodPost,
+                       protoMajor:  2,
+                       contentType: "application/grpc-web+proto",
+                       want:        false,
+               },
+               {
+                       name:        "empty content-type",
+                       method:      http.MethodPost,
+                       protoMajor:  2,
+                       contentType: "",
+                       want:        false,
+               },
+       }
+
+       for _, tt := range tests {
+               t.Run(tt.name, func(t *testing.T) {
+                       req, err := http.NewRequest(tt.method, "http://localhost/test", nil)
+                       if err != nil {
+                               t.Fatal(err)
+                       }
+                       req.ProtoMajor = tt.protoMajor
+                       if tt.contentType != "" {
+                               req.Header.Set("Content-Type", tt.contentType)
+                       }
+                       if got := isGRPC(req); got != tt.want {
+                               t.Errorf("isGRPC() = %v, want %v", got, tt.want)
+                       }
+               })
+       }
+}
+
+// TestProtocolMirrorTransport_HTTP1Backend verifies that requests to an HTTP/1.1-only
+// actor backend (standard web/REST actors without h2c support) always arrive as
+// HTTP/1.1, even when the downstream client connected to the ingress proxy via HTTP/2 (ALPN h2).
+func TestProtocolMirrorTransport_HTTP1Backend(t *testing.T) {
+       protoSeen := make(chan string, 1)
        backend := &http.Server{
                Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
                        protoSeen <- r.Proto
+                       w.WriteHeader(http.StatusOK)
                }),
-               Protocols: backendProtocols,
        }
        lis, err := net.Listen("tcp", "127.0.0.1:0")
        if err != nil {
                t.Fatal(err)
        }
-       go backend.Serve(lis)
        t.Cleanup(func() { backend.Close() })
+       go backend.Serve(lis)
 
        transport := newProtocolMirrorTransport()
-       for _, tt := range []struct {
-               protoMajor int
-               want       string
+       t.Cleanup(transport.CloseIdleConnections)
+
+       tests := []struct {
+               name        string
+               protoMajor  int
+               contentType string
+               wantProto   string
        }{
-               {protoMajor: 1, want: "HTTP/1.1"},
-               {protoMajor: 2, want: "HTTP/2.0"},
-       } {
-               req, err := http.NewRequest(http.MethodGet, "http://"+lis.Addr().String()+"/", nil)
-               if err != nil {
-                       t.Fatal(err)
-               }
-               req.ProtoMajor = tt.protoMajor
-               res, err := transport.RoundTrip(req)
-               if err != nil {
-                       t.Fatalf("RoundTrip (inbound HTTP/%d): %v", tt.protoMajor, err)
-               }
-               res.Body.Close()
-               if got := <-protoSeen; got != tt.want {
-                       t.Fatalf("upstream saw %s for inbound HTTP/%d, want %s", got, tt.protoMajor, tt.want)
-               }
+               {
+                       name:        "downstream HTTP/1.1 client",
+                       protoMajor:  1,
+                       contentType: "application/json",
+                       wantProto:   "HTTP/1.1",
+               },
+               {
+                       name:        "downstream HTTP/2 client (e.g. curl over HTTPS with ALPN h2)",
+                       protoMajor:  2,
+                       contentType: "application/json",
+                       wantProto:   "HTTP/1.1",
+               },
+       }
+
+       for _, tt := range tests {
+               t.Run(tt.name, func(t *testing.T) {
+                       req, err := http.NewRequest(http.MethodGet, "http://"+lis.Addr().String()+"/hello", nil)
+                       if err != nil {
+                               t.Fatal(err)
+                       }
+                       req.ProtoMajor = tt.protoMajor
+                       if tt.contentType != "" {
+                               req.Header.Set("Content-Type", tt.contentType)
+                       }
+
+                       res, err := transport.RoundTrip(req)
+                       if err != nil {
+                               t.Fatalf("RoundTrip failed: %v", err)
+                       }
+                       defer res.Body.Close()
+
+                       if res.StatusCode != http.StatusOK {
+                               t.Fatalf("status = %d, want %d", res.StatusCode, http.StatusOK)
+                       }
+
+                       if got := <-protoSeen; got != tt.wantProto {
+                               t.Fatalf("backend received protocol %s, want %s", got, tt.wantProto)
+                       }
+               })
+       }
+}
+
+// TestProtocolMirrorTransport_RealGRPCServer verifies that gRPC calls through
+// atunnel's reverse proxy correctly reach a real gRPC server over HTTP/2 (h2c)
+// and complete end-to-end with trailers intact.
+func TestProtocolMirrorTransport_RealGRPCServer(t *testing.T) {
+       // 1. Start a real gRPC backend server (speaking cleartext HTTP/2)
+       grpcLis, err := net.Listen("tcp", "127.0.0.1:0")
+       if err != nil {
+               t.Fatal(err)
+       }
+       t.Cleanup(func() { grpcLis.Close() })
+
+       grpcSrv := grpc.NewServer()
+       healthpb.RegisterHealthServer(grpcSrv, &testHealthServer{})
+       t.Cleanup(grpcSrv.Stop)
+       go grpcSrv.Serve(grpcLis)
+
+       // 2. Setup atunnel reverse proxy pointing to the gRPC actor
+       grpcUpstreamURL, err := url.Parse("http://" + grpcLis.Addr().String())
+       if err != nil {
+               t.Fatal(err)
+       }
+       transport := newProtocolMirrorTransport()
+       t.Cleanup(transport.CloseIdleConnections)
+
+       proxy := &httputil.ReverseProxy{
+               Rewrite: func(pr *httputil.ProxyRequest) {
+                       pr.SetURL(grpcUpstreamURL)
+               },
+               Transport: transport,
+       }
+
+       // 3. Run atunnel proxy server supporting unencrypted HTTP/2 for testing
+       proxyProtocols := new(http.Protocols)
+       proxyProtocols.SetHTTP1(true)
+       proxyProtocols.SetUnencryptedHTTP2(true)
+       proxyServer := &http.Server{
+               Handler:   proxy,
+               Protocols: proxyProtocols,
+       }
+       proxyLis, err := net.Listen("tcp", "127.0.0.1:0")
+       if err != nil {
+               t.Fatal(err)
+       }
+       t.Cleanup(func() { proxyServer.Close() })
+       go proxyServer.Serve(proxyLis)
+
+       // 4. Dial the proxy with a real gRPC client
+       conn, err := grpc.NewClient(
+               proxyLis.Addr().String(),
+               grpc.WithTransportCredentials(insecure.NewCredentials()),
+       )
+       if err != nil {
+               t.Fatal(err)
+       }
+       t.Cleanup(func() { conn.Close() })
+
+       client := healthpb.NewHealthClient(conn)
+       ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+       defer cancel()
+
+       resp, err := client.Check(ctx, &healthpb.HealthCheckRequest{})
+       if err != nil {
+               t.Fatalf("gRPC Check through atunnel failed: %v", err)
+       }
+       if resp.Status != healthpb.HealthCheckResponse_SERVING {
+               t.Fatalf("gRPC status = %v, want SERVING", resp.Status)
        }
 }
```

It is also important to decide where are you going to implement these conversions and the roles of each component, as later you may probably require websockets that is a different beast, since you need to implement the http handhsake for upgrade first ... or to support non-gRPC h2c actors

cmd/atenet/internal/router/xds.go
haiyanmeng, line 812: `defaultRouteTimeout` is 10s. Any server-streaming or bidi RPC longer than that is cut off. Streaming gRPC won't actually work. We can either `make it work in this PR` or `explain the limitation of this PR and fix it in a followup PR`

haiyanmeng (commented):
Can you add some e2e test coverage? Thanks!

EItanya (commented):
Rather than adding more logic to the http parsing in `atunnel`. Can we instead move to the CONNECT tunnel approach added by @keithmattix? It's far simpler both now and in the long-run, and will allow us to delete custom HTTP logic from the atunnel which is a huge win.

FWIW I originally wanted to open a PR to do exactly this, but I realized that the CONNECT path would save us from constantly fixing our HTTP logic.

LiorLieberman:
> Rather than adding more logic to the http parsing in `atunnel`. Can we instead move to the CONNECT tunnel approach added by Keith Mattix II (@keithmattix)? It's far simpler both now and in the long-run, and will allow us to delete custom HTTP logic from the atunnel which is a huge win.
> 
> FWIW I originally wanted to open a PR to do exactly this, but I realized that the CONNECT path would save us from constantly fixing our HTTP logic.

Is this finished? I dont think its wired up in the router still. Also aligned with @keithmattix comment in https://github.com/agent-substrate/substrate/pull/993#issue-5173345281

keithmattix:
+1 to @EItanya - atunnel shouldn't be in the business of doing L7 parsing. Standardizing on (h2) CONNECT across the substrate networking data plane has a few benefits, especially around auth, telemetry, and identity.

@LiorLieberman atunnel side is totally finished. I'm working on the agentgateway bits now and should have that leg wired up end to end by the end of the week.

LiorLieberman:
ok, I agree. but I dont see ingress CONNECT dialing wired up in atenet --> atunnel. am I missing something?
I thought you said its a follow up.  

From https://github.com/agent-substrate/substrate/pull/993#issue-5173345281

>I haven't wired up atenet router yet because it's nontrivial; we should do that in a second step so we can have a baseline for performance

cmd/atenet/internal/router/xds.go
LiorLieberman: >is better to not assume http2 == grpc

Right, thats a good idea.

Regarding the hops - yes. Gemini assessment is correct. (1) client --> atenet-router(envoy) (1.5) atenet-router(envoy) --> local sidecar (ext-proc) (2) atenet-router(envoy) --> pod (atunnel) (3) atunnel --> actor

LiorLieberman: Fixed via Antonio's suggestion: atunnel forwards h2 to the actor only for gRPC and translates everything else to HTTP/1.1. Also added tests with HTTP/1-only backends.

LiorLieberman: >where are you going to implement these conversions

regarding where conversions would live, we are in discussions about it. We need a followup to support atenet-router --> atunnel with CONNECT, so atunnel is not aware of L7 at all. This would require moving some of its l7 logic as well. 

@EItanya just opened an issue about it. I believe we would end up, at least for now, supporting two listeners on atunnel, one for CONNECT and one for no CONNECT - I will post my comments on the issue.

cmd/atenet/internal/router/xds.go
LiorLieberman: The flag is for h2 over TLS, we dont have an h2 enable flag. gRPC can also arrive as cleartext h2c on the plain listener with the flag off, and this line is needed for that path too. It's safe to keep on because atunnel negotiates h1 or h2 and downgrades everything except gRPC to HTTP/1.1 (to not break non h2 actors). I did move it inside the the if condition though so it only applies when atunnel is the upstream, though that distinction is kinda not relevant anymore since atunnel should always be in the upstream ( and we always targets atunnel:443 today); left a TODO to drop the if branch, and make this unconditional.

I think that a better approach would be to have CONNECT to atunnel but this will be discussed in a separate issue

LiorLieberman:
Have had discussions with many of you offline but posting here for posterity;

I agree that its a good direction to go to atenet --> atunnel is over CONNECT (and, removing / avoiding L7 logic in atunnel). But this requires more thinking and I think, for the short term at least, would likely result in having two listeners/ports in atunnel - one for CONNECT, one for non-CONNECT. 

Eitan was faster and opened #1208 already - will post my thoughts there.

cmd/atenet/internal/router/xds.go
LiorLieberman: added a comment

internal/e2e/suites/networking/protocol_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+// TestIngressProtocolDowngrade pins the ingress protocol contract end to end:
+// a client that negotiates HTTP/2 with the router must still be able to reach
+// an HTTP/1.1-only actor (the counter demo), because the atunnel leg
+// downgrades non-gRPC traffic to HTTP/1.1. A gRPC-shaped request, by
+// contrast, is carried to the actor as real HTTP/2 — so against this
+// non-gRPC actor it must fail loudly rather than silently fall back to
+// HTTP/1.1 (which would strip the trailers gRPC needs).
+//
+// TODO(liorlieberman): add the gRPC-positive counterpart (a gRPC actor
```
haiyanmeng: FYI, we have https://github.com/agent-substrate/substrate/tree/main/internal/e2e/fixtures/grpcecho today. Currently, it is used for testing egress grpc traffic. You can extend it if needed. 

Ideally, we should have a single binary for ingress/egress testing.

haiyanmeng: In fact, I just realized that your PR branch does not include https://github.com/agent-substrate/substrate/tree/main/internal/e2e/fixtures/grpcecho . You could rebase, and add gRPC-positive e2e test coverage.

internal/e2e/suites/networking/protocol_test.go
haiyanmeng, line 1: We currently have https://github.com/agent-substrate/substrate/blob/main/internal/e2e/suites/networking/grpcegress_test.go.

We can rename this file to https://github.com/agent-substrate/substrate/blob/main/internal/e2e/suites/networking/grpcingress_test.go

LiorLieberman: Thanks

internal/e2e/suites/networking/protocol_test.go
haiyanmeng: FYI, https://github.com/agent-substrate/substrate/pull/1245 tries to merge the networking e2e test fixtures into one binary called `testserver`.

internal/atunnel/ingress.go
```diff
@@ -181,6 +181,65 @@ func NewServer(cfg Config) (*Server, error) {
...
+			return true
+		}
+	}
+	return false
+}
+
+// protocolMirrorTransport picks the actor-leg protocol per request. gRPC goes
+// out as prior-knowledge h2c, since it needs HTTP/2 trailers and full-duplex
+// streaming; everything else is translated to HTTP/1.1 even if it arrived over
+// HTTP/2 at the edge. This is to not break previous logic so an HTTP/1.1-only actor keeps working whatever the
```
haiyanmeng, line 210: This line is too long

cmd/atenet/internal/router/xds.go
```diff
@@ -759,19 +778,21 @@ func (x *XdsServer) buildOriginalDstCluster() *clusterv3.Cluster {
...
+		// gates its own actor leg: it forwards HTTP/2 only for gRPC and
+		// downgrades everything else, so an HTTP/2 client cannot break an
+		// HTTP/1.1-only actor. Without upstream mTLS there is no atunnel leg
+		// to carry HTTP/2, so that mode keeps Envoy's implicit HTTP/1.1.
+		cluster.TypedExtensionProtocolOptions = map[string]*anypb.Any{
+			httpProtocolOptionsName: newAny(&httpv3.HttpProtocolOptions{
+				UpstreamProtocolOptions: &httpv3.HttpProtocolOptions_UseDownstreamProtocolConfig{
+					UseDownstreamProtocolConfig: &httpv3.HttpProtocolOptions_UseDownstreamHttpConfig{
+						HttpProtocolOptions:  &corev3.Http1ProtocolOptions{},
+						Http2ProtocolOptions: &corev3.Http2ProtocolOptions{},
```
haiyanmeng, line 778: Should this be gated by `--https-h2`?

haiyanmeng:
LGTM. 

Can you create a github issue for the grpc support on ingress and put the TODOs (such as make it work for long gRPC streaming) either on the issue itself or as sub-issues?

Another question: do we plan to let an actor advertise the application protocol it uses in the future?

cmd/atenet/internal/router/config.go
```diff
@@ -94,6 +94,10 @@ type routerConfig struct {
 	ConnectPlainTextPort int
 	ConnectTLSPort       int
 	EnvoyCertPath        string
+	// HttpsH2 offers HTTP/2 via ALPN on the HTTPS ingress listener, which
+	// gRPC over TLS requires. Off preserves the historical no-ALPN
+	// behavior.
+	HttpsH2 bool
```
bowei: Is this truly an option that we would switch on/off? Why not have this always on?

internal/atunnel/ingress.go
```diff
@@ -181,6 +181,65 @@ func NewServer(cfg Config) (*Server, error) {
 	return s, nil
 }
 
+// isGRPC reports whether an incoming HTTP request conforms to the gRPC over
+// HTTP/2 specification: HTTP/2 framing, POST method, and a Content-Type of
+// "application/grpc" (optionally with a "+<subtype>" or ";<parameters>").
+// Notably this excludes gRPC-Web that works over HTTP/1.1.
+// See https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+func isGRPC(r *http.Request) bool {
```
bowei, line 189: Is there a reason why atunnel needs to parse HTTP? (or TLS?). Can we just dumb copy the bytes and avoid having to look inside for every kind of TCP-based protocol?

keithmattix: Big +1

LiorLieberman: Here are all the reasons why atunnel **IS already** L7 today:

- atunnel checks the actor (from Host) is active before it passes the request to it, it returns 421 if its stale or not active
- **without** CONNECT (between atenet->atunnel) implemented, we currently have `X-Ate-Target-Port` that is used to route to different actor ports
- atunnel already had an L7 policy before this PR, it was just "downgrade everything to h1"

since it is l7 today, it has to know what it needs to re-originate, and for that it needs to know what was the request it received. 

if we remove atunnel L7 logic right now, CONNECT (or alternatives) is required to keep host activation check and support multiple ports. 

As said multiple times before - I do believe making atunnel not do any l7 is the best approach but we do need to talk about it and its definitely not the scope of this PR.

internal/atunnel/ingress.go
```diff
@@ -181,6 +181,65 @@ func NewServer(cfg Config) (*Server, error) {
...
+	if strings.HasPrefix(ct, base) {
+		switch ct[len(base)] {
+		case '+', ';':
+			return true
+		}
+	}
+	return false
+}
+
+// protocolMirrorTransport picks the actor-leg protocol per request. gRPC goes
```
bowei, line 207: Discussed -- we should take a TODO to negotiate H1/2 if the Actor supports it.

bowei (dismissed):

EItanya (dismissed):

cmd/atenet/internal/router/config.go
LiorLieberman: removed

cmd/atenet/internal/router/xds.go
LiorLieberman: removed the flag

internal/atunnel/ingress.go
LiorLieberman: discussed again but adding for posterity: cleartext negotiation is really not common, even the http2 rfc have removed it - https://www.rfc-editor.org/rfc/rfc9113.html#name-http-2-version-identificati

We should not negotiate here. We should build towards a model where atunnel dont do l7s. Will open issues for that, and see how far we can go

PR #1195, opened 2026-08-25
atelet: fall back to a dense stream when a checkpoint source lacks SEEK_DATA

mesutoezdil:
writeSparseZstd only special-cased ENXIO from the initial SEEK_DATA call. Any other error, for example a filesystem that does not support SEEK_DATA/SEEK_HOLE at all, hard-failed the whole checkpoint upload, which the caller turns into a DataLoss error and crashes the actor. The local checkpoint copy path (copySparse) already falls back to a dense copy in this situation. This gives the upload path the same fallback: probe SEEK_DATA before writing anything to the output, and if it is unsupported, rewind and stream a plain dense zstd instead of failing the upload.

### DIFF
--- a/cmd/atelet/internal/ategcs/objects.go
+++ b/cmd/atelet/internal/ategcs/objects.go
@@ -150,10 +150,17 @@ type writeContentResult struct {
 func writeContent(out io.Writer, content io.Reader) (writeContentResult, error) {
 	if f, ok := content.(*os.File); ok {
 		logical, populated, err := writeSparseZstd(out, f)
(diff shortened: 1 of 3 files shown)


BenTheElder:
What file systems don't support it?

at the moment this is so performance load bearing that I almost wonder if we should continue to hard fail ...

mesutoezdil:
> What file systems don't support it?
> 
> at the moment this is so performance load bearing that I almost wonder if we should continue to hard fail ...

hm yeah y r right
i close it

PR #1199, opened 2026-08-25
atenet: cap Envoy worker concurrency

### LINKED ISSUE #1193: atenet: Envoy sizes its worker pool from the host CPU count
## What happened

I ran an agent-substrate cluster on a 128 CPU node, and received these errors:

```
[warn] evutil_make_internal_pipe_: pipe: Too many open files
[err]  evsig_init_: socketpair: Too many open files
```

Both atenet pods crashlooped. The Go container next to each Envoy stayed healthy the whole time, which is why this reads like an atenet bug rather than a sizing problem.

Envoy picks its worker count from the host CPU count. Kubernetes CPU limits are CFS quota and don't constrain that, so on a big node Envoy starts a worker per host thread and burns through the default nofile before it finishes booting.

## What I did

I solved it by pinning the worker count instead of letting Envoy pick one, across all three Envoy containers:

- `--concurrency 4` on the Envoy container in `atenet-router.yaml` and `atenet-egress.yaml`. This is the change I ran on the box, and it cleared the crashloop.
- The same flag on `atenet-egress-with-sdsmint.yaml`. That file runs a third Envoy, selected by `--experimental-use-sdsmint` rather than by an overlay, so it inherited nothing from the other two. I only found it afterwards while checking coverage, and I have not run that path.

For the path I haven't run, I confirmed every Envoy container across the three manifests resolves to `--concurrency 4`, and that `base` and `kind` still build clean.

## Why this still needs work

The 4 is not measured. I picked something comfortably under the fd limit and a bit above what other projects use for sidecars. Nothing has been profiled against actor traffic.

What the number should track is the CPU the pod actually gets, and we can't express that today, because none of these containers declare CPU requests or limits.

## What to do instead

One option is to declare CPU requests on the three proxy containers, following what `kind/prometheus.yaml` already does: CPU requests, memory limits, no CPU limit.

Then pass the request in through the downward API:

```yaml
env:
- name: ENVOY_CONCURRENCY
  valueFrom:
    resourceFieldRef:
      resource: requests.cpu
      divisor: "1"
```

And read it. Kubernetes expands `$(VAR)` in `command` as well as `args`, so no wrapper script is involved:

```yaml
- --concurrency
- "$(ENVOY_CONCURRENCY)"
```

Another option is to be able to configure the concurrency itself using flags and not touching CPU at all.

## What this costs

- kind is the real constraint. `manifests/ate-install/kind/kustomization.yaml` pulls `../atenet-router.yaml` in directly with no resource patches, so whatever request we set lands on laptop clusters too. Either pick something small enough for kind and live with one thread everywhere, or patch it in the kind overlay and accept that the number differs per environment.

## What other projects do

- Istio derives it, reading `limits.cpu` into `ISTIO_CPU_LIMIT` through the downward API. It is only safe because their charts ship a default 2000m sidecar limit.
- Consul hardcodes 2, overridable per pod by annotation. Their values.yaml says to keep the number low for sidecars and high for edge proxies. Ours are edge.
- Envoy Gateway exposes `EnvoyProxy.spec.concurrency` and only emits `--concurrency` when you set it. No derivation at all.

## Related

- #665 measures atenet-router request capacity, which is what would give the worker count a real number instead of a guess.

eliranw:
Fixes #1193

With `--concurrency` unset, Envoy sizes its worker pool from the host CPU count, which cgroup limits do not constrain. On a many-core node it exhausts the default nofile during startup and both atenet pods crashloop, with the Go container beside each one healthy:

```
[warn] evutil_make_internal_pipe_: pipe: Too many open files
[err]  evsig_init_: socketpair: Too many open files
```

Each Envoy container now takes its worker count from a variable defaulting to 4, so a cluster can be resized without editing manifests:

```yaml
- --concurrency
- "$(ENVOY_CONCURRENCY)"
env:
- name: ENVOY_CONCURRENCY
  value: "4"
```

`hack/install-ate.sh` gains `--atenet-router-concurrency` and `--atenet-egress-concurrency`, following the shape of `--podcert-workers-per-signer`. They are separate flags because Envoy pins a connection to a worker for its lifetime, and egress holds CONNECT tunnels open far longer than the router's request/response connections. `atenet-egress-with-sdsmint.yaml` carries a third Envoy that no overlay reaches, so it is changed directly.

No CPU requests or limits are added. #1193 lists deriving the count from a cpu request as the other option; the measurements against it are in a comment there.

Tested on kind: full install, both proxies healthy, and the flags verified against Envoy's own `server.concurrency`. Whether 4 is the right number is still open, and #665 would answer it.

## Known gap

`--atenet-router=agentgateway` runs no Envoy, so this does nothing there. agentgateway sizes its tokio runtime from `num_cpus` and needs its own change.

- [x] Tests pass

### DIFF
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -72,6 +72,8 @@ function usage() {
   echo "  --delete-all                           Delete core system and all registered demos"
   echo "  --atenet-router=envoy|agentgateway     Select the ingress and egress dataplane (default: envoy)"
   echo "  --podcert-workers-per-signer N         Concurrent workers per podcertificate-controller signer (default: 1)"
+  echo "  --atenet-router-concurrency N          Envoy worker threads in atenet-router (default: 4)"
+  echo "  --atenet-egress-concurrency N          Envoy worker threads in atenet-egress (default: 4)"
   echo "  --rollout-timeout DURATION             Per-workload readiness wait timeout, kubectl-style Go duration (default: 60s)"
   echo "  --otlp-endpoint URL                    Send all control plane telemetry to URL, not to the cluster default (see benchmarking/telemetry/README.md)"
   echo ""
@@ -183,6 +185,21 @@ podcert_workers_per_signer() {
   local workers="${ATE_INSTALL_PODCERT_WORKERS_PER_SIGNER:-1}"
   if ! [[ "${workers}" =~ ^[1-9][0-9]*$ ]]; then
     echo "Error: --podcert-workers-per-signer must be a positive integer, got '${workers}'" >&2
+    exit 1
+  fi
+  echo "${workers}"
+}
+
+# atenet_concurrency validates one proxy's Envoy worker thread count and echoes
+# it. The two atenet proxies take separate flags: the router accepts short
+# request/response connections, while the egress gateway holds CONNECT tunnels
+# open for the life of an actor's connection, and Envoy pins a connection to a
+# worker for its lifetime.
+atenet_concurrency() {
+  local flag="$1"
+  local workers="${2:-4}"
+  if ! [[ "${workers}" =~ ^[1-9][0-9]*$ ]]; then
+    echo "Error: ${flag} must be a positive integer, got '${workers}'" >&2
     exit 1
   fi
   echo "${workers}"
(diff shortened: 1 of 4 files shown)


eliranw:
Looking at other flags in similar scenarios (`--podcert-workers-per-signer`), not sure if we need additional documentation.

eliranw:
@botengyao @EItanya Could someone take a peek at this?

PR #1200, opened 2026-08-25
ateom-microvm: break down guest RAM by what is holding it

chw120:
A micro-VM's guest RAM is a black box from the host. `ateom` knows the VM was given N MiB; it has no way to say whether that memory is empty, page cache, the workload's anonymous pages, or the kata-agent. That matters beyond curiosity: CH's snapshot is sparse and faulted-only, so everything that is *not* free is a
page a Full checkpoint has to write — and therefore bytes in `ate.actor.checkpoint.duration` and `ate.actor.restore.duration`.

This adds `ate.microvm.guest.memory.bytes`, an observable gauge that splits the guest into six components which tile the guest kernel's MemTotal:

| component | what it is |
|---|---|
| `free` | pages the guest kernel considers free |
| `page_cache` | file-backed pages, minus shmem |
| `containers` | the workload cgroups' anonymous pages |
| `tmpfs` | shmem, given a slice of its own |
| `kata_agent` | the agent's RSS |
| `kernel_and_other` | the residual |

The source is the kata-agent's own metrics scrape (`kata_guest_meminfo` plus per-container cgroup stats), so it measures the guest kernel's accounting rather than this code's model of it: whatever holds RAM lands in one of the six.

Notes on the choices:

* **Closed set, sums to MemTotal.** That is the property that makes a stacked chart honest, and the e2e assertion checks all six are present for exactly that reason. `kernel_and_other` is a residual, so something new that starts holding memory shows up rather than disappearing.
* **`tmpfs` is separate from `page_cache` on purpose.** It is where write storage living in guest RAM would appear — see the retired guest-tmpfs-upper design in `rootfsupper.go`, which "pinned every written byte in guest memory". Folding it into page cache would hide exactly the regression worth catching.
* **`containers` charges anonymous pages, not cgroup usage.** Usage would double-count the container's file pages against `page_cache` and push the total over MemTotal.
* **Attribution is by template, not by actor.** A worker is recycled through actor after actor and each would leave a dead series behind; the repo draws the same line in `ateattr.ActorMetricAttributes`.
* **The sample is taken in the callback, not by a background goroutine.** The two pieces of state it reads are the same atomics `GetWorkloadStats` reads, which exist so a reader never waits on a boot, snapshot or restore. With no lock to take, a sampler goroutine would add a staleness window and a lifecycle to get wrong, for nothing.
* **Registration failure is not fatal.** An ateom that cannot publish the breakdown can still run actors; failing the boot would trade a blind spot for an outage.
* **Nothing is observed when there is no guest.** A zero would read as "the guest is empty", which is a different and false claim.

micro-VM only — the gVisor sentry publishes no equivalent decomposition — so the e2e assertion is gated on `SandboxClassMicroVM` and the prefix list is additive.

- [x] Tests pass

### DIFF
--- a/cmd/ateom-microvm/guestmem.go
+++ b/cmd/ateom-microvm/guestmem.go
@@ -1,0 +1,190 @@
+//go:build linux
+
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/agent-substrate/substrate/cmd/ateom-microvm/internal/agentstats"
+	"github.com/agent-substrate/substrate/cmd/ateom-microvm/internal/guestmem"
+	"github.com/agent-substrate/substrate/internal/ateattr"
+)
+
+// guestMemoryMetric is the gauge the breakdown is published on: one series per
+// slice, distinguished by guestMemComponentKey, all of them adding up to the
+// guest's MemTotal.
+//
+// Deliberately NOT the same shape as GetWorkloadStats. That RPC answers "how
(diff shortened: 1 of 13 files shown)


BenTheElder:
What does this cost? Do we have a use case that requires this on a permanent basis, versus intermittent debugging the runtime itself? Is there an issue for this feature?

On highly loaded systems, the context switches may add up.

Nit: let's not expose the implementation detail of "kata agent", we're not committed to that longterm. And if you boot an older snapshot we're booting systemd, not a single process.

chw120:
> What does this cost? Do we have a use case that requires this on a permanent basis, versus intermittent debugging the runtime itself? Is there an issue for this feature?
> 
> On highly loaded systems, the context switches may add up.
> 
> Nit: let's not expose the implementation detail of "kata agent", we're not committed to that longterm. And if you boot an older snapshot we're booting systemd, not a single process.

This is runtime-debugging instrumentation. Close the PR.

PR #1202, opened 2026-08-25
ateom: detect actor process exits and expose GetWorkloadHealth

zoez7:
Part of #292 

* Added a new RPC, `GetWorkloadHealth(actor_uid)` in ateom.proto so atelet can probe each ateom to get the status.
* Both gVisor and microVM ateom will monitor the health of the containers inside.


cmd/ateom-gvisor/main.go
```diff
@@ -307,12 +308,65 @@ type activeRPCInfo struct {
...
+	cmdWait(ctx context.Context, containerName string) (*int32, error)
+	cmdStateStatus(ctx context.Context, containerName string) (string, error)
+}
+
 // workloadSession captures the in-memory metadata for the workload currently running
 // in the sandbox, so the SIGTERM handler knows which containers to signal and
 // wait on during graceful shutdown. The sandbox runs one workload at a time.
 type workloadSession struct {
 	rcmd       *runsc
+	prober     containerProber
```
sfunkenhauser: It looks like containerProber just exposes a subset of runsc methods. Is that correct? Can we just use the existing rcmd variable instead? If the reason for the interface was to enable testing, let's still combine them and create an interface for all runsc methods.

cmd/ateom-gvisor/main.go
sfunkenhauser, line 787: Seems a little weird that we have multiple defer func() that both unset `s.activeActor`. Are they both necessary?

counter
sfunkenhauser, line 1: We shouldn't push the counter binary. Maybe we are missing this from .gitignore?

sfunkenhauser (commented):
Couple small review comments.

A larger overall question though. Do we need to re-implement this logic, or can we instead use the pod informer to get the container status?

cmd/ateom-gvisor/main.go
zoez7: Yes both are needed.
1. If we remove the first defer, any error between line 759 and 786 does not register the cleanup of `s.activeActor`.
2. If we remove the second one, during `deactivateActorNetworking` etc, the activeActor would not be unset yet (due to LIFO of defer functions)

cmd/ateom-gvisor/main.go
zoez7: I was using the interface for testing. Created interface for all runsc methods.

counter
zoez7: Looks like our .gitignore already ignores all bin/, I updated my GOBIN env var locally to point there.

BenTheElder: you shouldn't need to touch system GOBIN, we should use a make target or add an agent instruction to put binaries under bin/

zoez7:
As discussed during standup, for the immediate milestone, we just need to know (on the next suspend / resume) that an actor has crashed, since we are still exploring mega-atelet topics, the changes here may soon become obsolete. This kind of active monitoring makes sense for long running workload, but a core usage pattern of substrate is frequent suspend and resume, so I think "crash on next suspend" is ok for M2. I will close this PR for now.

PR #1203, opened 2026-08-25
Move WorkerPoolSyncer to ate-controller

### LINKED ISSUE #729: Move WorkerPoolSyncer out of api-server
WorkerPoolSyncer runs an informer to synchronize K8s pods to Substrate workers. This currently runs inside the ate-api-server, which is mixing responsibilities. We should move the WorkerPoolsyncer out of the ate-api-server and likely into the ate-controller, which is a better spot for an informer like this to run.

sfunkenhauser:
Fixes #729 

- [ x  ] Tests pass


cmd/ateapi/internal/controlapi/worker.go
```diff
@@ -188,6 +207,88 @@ func validateDeleteWorkerRequest(req *ateapipb.DeleteWorkerRequest) field.ErrorL
...
+// are cleared.
+//
+// Nothing to release is the common case and reports success: an unassigned
+// Worker, a superseded assignment, and an Actor that has since moved elsewhere
+// all leave no Actor pointing at this Worker, which is the state this is driving
+// towards.
+//
+// A concurrent SuspendActor or ResumeActor wins the optimistic version check;
+// this attempt fails as ABORTED so the caller retries against the newer state.
+func (s *RPCService) releaseActorOnDeadWorker(ctx context.Context, worker *ateapipb.Worker) error {
```
juli4n: Should we move this into a DeleteWorker workflow? That would be consistent with other RPC handlers that need to orchestrate an idempotent sequence of steps.

juli4n (approved):

Issue #796, opened 2026-08-07
Add ate.actors gauge


krisztianfekete:
#433 shipped everything, except `ate.actors`, opening this one so we can chat about the design and it doesn't get lost.

Now we can now see pool saturation, lifecycle latency, scheduling outcomes and cold start phases, but nothing says how many actors exist and what state they're in.

Original idea for `ate.actors` was an observable updowncounter labelled by status and template. Both bounded, so it's fine cardinality-wise.

The problem is that a metric callback runs on every collection interval, and a scan across ~1B actors every interval isn't something we can (or at least should) do. So we need something cheaper, e.g. a store maintained counters incremented on status transitions, or a separate actor index. That's more like a store design question and I'd rather discuss it first.

One related thing is addressing the new `ate.actor.crashes` metric that counts crashes as they happen. This would count actors that are in CRASHED right now. Both are useful, but we should say which is which so operators are aware.

cc @JeffLuoo @maxsmythe @juli4n @EItanya @git286

JeffLuoo:
Put some of my thought:

-   The cost is different for each status. Four statuses hold a Worker: RESUMING, RUNNING, SUSPENDING and PAUSING. The fleet size limits their counts, not the Actor. Assignment already contains `actor_template`, so ateapi can count these from `workercache.Cache.Workers()` in memory. This needs no store change, and it suppliesthe density and ratio use cases from #433. Only SUSPENDED, PAUSED, CRASHED and DELETING need the design.

When you mention `separate actor index`, what does it mean?

krisztianfekete:
Assignment has `actor_template`, `actor` and `actor_uid` but no status, so we can get actors by template from cache, not by status. To get statuses, we need a read per worker, which is probably fine, but more expensive than what we can get from cache.

A worker hosts one actor, so the count is `ate.workerpool.workers{state=assigned}` labeled by template, which is mostly fine. The interesting case is the SUSPENDED ones, which is the part we'd be deferring. The per-template view of what the fleet is doing is still useful, but that's a bit different that what I was originally thinking of, and prefer no to ship that like that as it can be off by a lot vs. what users would expect from a metric called `ate.actors`.

Regarding the index, I am thinking of counters per status, and template, written in the same transaction as the status change (it can drift if we miss updating the counters), or a set per status, and template and take the cardinality cost. It's a drift vs. memory question really, and  regardless of which one, we write it at transition-time, which is why it feels like a it's store side question.

[...]

PR #1204, opened 2026-08-25
feat(atepg): project actor state into a column

krisztianfekete:
This PR adds
- a `state` column to atepg's `actors` table, projected out of the proto and written by the same INSERT/UPDATE, so it can't drift from the record
- `CountActors`, which is a plain `count(*) group by state`

Note that there's no index on`state`, which is the opposite of what I planned to do. Benchmarked it at 10M actors:

| | clean table | bloated (29% dead) |
|---|---|---|
| seq scan | 0.72s | 0.57s |
| index-only scan | 0.33s | 1.70s (13.2M heap fetches) |

Transitions land randomly all over the table, so it takes very little churn to wreck the visibility map. 10.6% dead tuples leaves only 4.9% of pages all-visible, and autovacuum doesn't even wake up until 20%. So the index only wins in the state the table is basically never in. It also costs 14% write throughput and kills HOT updates entirely (0 HOT with the index, 99.8% without), on the hottest write path we have.

Seq scan is flat at ~0.6s no matter how bloated things get, so that's what we're doing.

Tested all this with 10M actors on a single-node container. Scaling to 1B _should_ be linear based on this.

Delivers first half of #796

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/store/atepg/atepg.go
+++ b/cmd/ateapi/internal/store/atepg/atepg.go
@@ -583,9 +583,10 @@ func (p *Persistence) CreateActor(ctx context.Context, actor *ateapipb.Actor) (*
 	}
 
 	_, err = p.pool.Exec(ctx, `
-		INSERT INTO actors (atespace, name, uid, version, proto)
-		VALUES ($1, $2, $3, $4, $5)`,
-		atespace, name, dbActor.GetMetadata().GetUid(), dbActor.GetMetadata().GetVersion(), protoBytes)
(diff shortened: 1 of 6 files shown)


BenTheElder (changes requested):
I was about to write a comment along these lines, but I see they've already been written on the issue:
https://github.com/agent-substrate/substrate/issues/796#issuecomment-5417191954
https://github.com/agent-substrate/substrate/issues/796#issuecomment-5417437047

Issue #861, opened 2026-08-11
Reanme `ateomImage` as `workerImage` for `WorkerPool`


HavenXia:
If customer has their own ateom implementation, they can provide one in `ateomImage`.

Otherwise this field should be optional, and leaving it empty will let substrate system (specially `atecontroller`) inject a versioned `ateomImage`, based on the `sandboxClass` to pick `ateom-gvisor`/`ateom-microVM` image that `atecontroller` carries.

This is a critical part of substrate system upgrade, it doesn't make senses to edit customer's CR for every ateom upgrade.

HavenXia:
@BenTheElder  Let me know what's your opinions.

HavenXia:
Status update
--------

Discussed this with @BenTheElder in the context of the [substrate upgrade proposal](http://doc/1JduAyGZFyqdNp4UhKiv0EN-is3iW5tf5BwGTWbt2ouI?resourcekey=0-MXG12QCkleIxhY7cOB7g6A), Where it stands now:

The original shape of this issue still holds, when the image field is not set,  the system injects a versioned default based on `sandboxClass`. 

Open concerns:
1. Coupling: if the default is baked into the `atecontroller` binary or through a flag, patching an worker image CVE requires rolling `atecontroller`. Dataplane seems patch happen much more often than control plane changes, so per-component upgrade must stay possible.
2. Editing the worker image on the CR is a legitimate customer operation, but today it is broken: atecontroller reconciles every worker pod immediately, which kills the actors on them. Any design here must keep the old worker set intact while new workers come up. Similar to the rolling upgrade of dataplane in [substrate upgrade proposal](https://docs.google.com/document/d/1JduAyGZFyqdNp4UhKiv0EN-is3iW5tf5BwGTWbt2ouI/edit?resourcekey=0-MXG12QCkleIxhY7cOB7g6A&tab=t.0#bookmark=id.75lm729hv048).  @sfunkenhauser have we considered that?
3. Rolling worker pools without losing actors is core functionality independent of system upgrade. Image is one case; pod shape changes (e.g. uVM dropping privileged) are another. 


Direction we are converging on, but need more discussion:

~~1. My idea:  Make dataplane rolling a `ate-controller`-owned behavior. On any change to a WorkerPool CR, `ate-controller` keeps two worker sets, drains nodes one by one, and fills in new workers. A system upgrade driver triggers it the same way a customer would.
[...]

PR #1210, opened 2026-08-25
api: rename ateomImage to workerImage and make it optional

HavenXia:
Rename `WorkerPoolSpec.AteomImage`to `WorkerImage` and drop the constraints so the field can be left unset. 

This is the basis for let an empty workerImage lets the controller inject a versioned default image chosen by the pool's sandbox class.

Part of #861

- [x] Tests pass

### DIFF
--- a/docs/api-guide.md
+++ b/docs/api-guide.md
@@ -11,7 +11,7 @@ The `WorkerPool` defines the pool of physical "warm" compute capacity. It manage
 | Field | Type | Description |
 | :--- | :--- | :--- |
 | `replicas` | `int32` | **Required.** Number of physical standby pods to maintain in the cluster. |
-| `ateomImage` | `string` | **Required.** The container image for the `ateom` herder process (e.g. `ko://github.com/agent-substrate/substrate/cmd/ateom-gvisor`). |
+| `workerImage` | `string` | Optional. The container image for the `ateom` herder process (e.g. `ko://github.com/agent-substrate/substrate/cmd/ateom-gvisor`). When unset, the controller injects a versioned default image for the pool's `sandboxClass`. |
 | `sandboxClass` | `string` | Optional. The sandbox runtime family for the pool: `gvisor` (default) or `microvm`. Drives the worker pod shape (e.g. KVM device mounts, node placement) and which `SandboxConfig`s are eligible. |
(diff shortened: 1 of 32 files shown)


docs/api-guide.md
```diff
@@ -11,7 +11,7 @@ The `WorkerPool` defines the pool of physical "warm" compute capacity. It manage
 | Field | Type | Description |
 | :--- | :--- | :--- |
 | `replicas` | `int32` | **Required.** Number of physical standby pods to maintain in the cluster. |
-| `ateomImage` | `string` | **Required.** The container image for the `ateom` herder process (e.g. `ko://github.com/agent-substrate/substrate/cmd/ateom-gvisor`). |
+| `workerImage` | `string` | Optional. The container image for the `ateom` herder process (e.g. `ko://github.com/agent-substrate/substrate/cmd/ateom-gvisor`). When unset, the controller injects a versioned default image for the pool's `sandboxClass`. |
```
zoez7, line 14: > When unset, the controller injects a versioned default image for the pool's `sandboxClass`.

Just to understand -  substrate will maintain a mapping of "sandboxClass" -> "default worker image" for each substrate version?

zoez7 (approved):

docs/api-guide.md
HavenXia: No need for a maintained mapping. 

In my next pr, `ate-controller` will compose the default workerImage by: <repo>/ateom-<sandboxClass>:<its own build version>, so each controller build defaults to the ateom images published alongside it.

In the near term, the install script stamps the repo it pushes to into the controller; later once we have prebuilt image for each version, will replace to pull that.

BenTheElder: That is a really weird API. I dislike this immensely.

BenTheElder: This means if I want to push a quick fix to ate-controller I must also push and roll my workers.

As I have said repeatedly: we should not do that. This is a bad coupling.

HavenXia: (this should be a discussion on next pr but happy to clarify now)

> This means if I want to push a quick fix to ate-controller I must also push and roll my workers.

So I can think of tweaking the upgrade model to not use the ate-controller's version but read the entire substrate version to decouple.

 But that incurs a real question: if you push a quick fix to ate-controller, what's the version for this fixture release, equal to previous one, or a new version? If it's from v0.1.0 to v.0.1.1 to just fix a ate-controller, do you want this system upgrade to be clever enough to only upgrade the `ate-controller`?

BenTheElder: [discussed more 1:1, I think we should make it required again and take another approach to upgrade]

PR #1212, opened 2026-08-25
atecontroller: enable leader election

HavenXia:
A Deployment rollout briefly runs old and new controller replicas side by side. Take a `coordination.k8s.io` lease (`ate-controller.ate.dev`, in the pod's own namespace) so only one manager reconciles the WorkerPool CR at a time.

- [x] Tests pass

### DIFF
--- a/cmd/atecontroller/internal/controllers/gen.go
+++ b/cmd/atecontroller/internal/controllers/gen.go
@@ -23,3 +23,7 @@ package controllers
 //+kubebuilder:rbac:groups=discovery.k8s.io,resources=endpointslices,verbs=get;list;watch,namespace=ate-system
 
 //go:generate bash ../../../../hack/run-tool.sh controller-gen rbac:headerFile=../../../../hack/boilerplate/sh.txt,roleName=ate-controller paths="./..." output:rbac:artifacts:config=../../../../manifests/ate-install/generated/
+
+// The manager's leader-election lease and the events it records on it.
+//+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch;delete,namespace=ate-system
+//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch,namespace=ate-system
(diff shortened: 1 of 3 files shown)


ahmedtd:
Use the rendezvous package --- that's what it's for!

With rendezvous hashing, each controller instance can be active and splitting work fairly between them, rather than all the work being handled by a single leader.

HavenXia:
> Use the rendezvous package --- that's what it's for!
> 
> With rendezvous hashing, each controller instance can be active and splitting work fairly between them, rather than all the work being handled by a single leader.

The rendezvous package says 

> the assignment is not atomic, so the controllers' actions need to be safe even if multiple replicas attempt to process the same item simultaneously.

I think it's for same-version replicas, where reconcile is idempotent. However, The window this PR cares about is the rolling replacement of the controller itself during system upgrade. 

In this scenario the two live replicas are different builds that disagree about the desired worker set, and simultaneous processing of the same WorkerPool by both versions is exactly what we needs to avoid. 

Leader election make the trasition serial, and we run a single replica of `atecontroller` today, so I feel there is no work to use  `rendezvous ` yet?

@ahmedtd

ahmedtd:
K8s leader election is also not guaranteed atomic, it's best effort.  If that is problematic, we need to solve it.

HavenXia:
We don't rely on it to be atomic, `ate-controller` will be made to only create and manage worker sets that corresponding to its own build version.  So a short dual controller window is still safe.
The leader election here is to minimize the situation that old and new `atecontroller` writes stauts to the same `WorkerPool` CR.

bowei:
We should also use the K8s hashing for watchers when it moves to beta

HavenXia:
> We don't rely on it to be atomic, `ate-controller` will be made to only create and manage worker sets that corresponding to its own build version. So a short dual controller window is still safe. The leader election here is to minimize the situation that old and new `atecontroller` writes stauts to the same `WorkerPool` CR.

Update: had more discussion with @BenTheElder on atecontoller and it won't has customized logic to only manages workers corresponding to its own build version. 

But the leader election itself should still be useful to reduce the possibility of two atecontroller reconciling same CR.

PR #1173, opened 2026-08-24
Compartmentalize ActorTemplateExists

thockin:
This seems mostly vestigial?

PR #1215, opened 2026-08-25
Use Kubernetes declarative validation

thockin:
This mostly eliminates the need to hand-write validation code.

This PR is a long series of commits which add DV for most of Actor and all of Atespace.

Here is a map to the commits:
* The first few take a dep on a new Kubernetes tag, import the code into third_party, and apply a single patch.  Because we have different Go modules for tools, I had to do it twice.  When that patch lands, we can revert these commits, but that won't be until the 1.38 cycle in a few months.
* The next commits slowly add DV support, so a human can review each of them in a reasonable amount of time.  The emphasis is on great test cases.
* I made a bad choice early on as to where to generate the code into, so I moved it.  Rebasing on that was exceedingly hard, so I left it as a move.
* This required changing update/go-generate -> update/codegen -- we need to get the ordering of tools right, which `go generate` does not guarantee.
* Then I added a "middle" layer called "ServiceImpl" between the RPC and storage layers.  This allows things like workflow to call the same business logic as the RPC layer, including validation.  Lots of test fixes.
* Then I finished the Create() and Update() paths for Actor.  Those represent the "right" (or closest to) way to implement resources, and tests for validation.

I strongly encourage reviewers to read it commit-by-commit.  Rebasing this is VERY tedious, so the sooner it lands or dies completely, the better.  Then we can start converting the rest.

`./hack/run-tool.sh validation-gen --docs` will produce some docs on the tool and the available tags.

@laoj2 @juli4n @EItanya @HavenXia
@lalitc375 @yongruilin @jpbetz  FYI


juli4n (dismissed):

HavenXia (dismissed):

thockin:
Rebased on #1173 to remove a few hunks from this

thockin:
Green!

HavenXia (approved):

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -61,40 +83,39 @@ func (s *RPCService) CreateActor(ctx context.Context, req *ateapipb.CreateActorR
...
-		},
+	// Verify that the result is properly valid before storing it.
+	outActor := proto.CloneOf(inActor)
+	outActor.Status = &ateapipb.ActorStatus{
+		State:          ateapipb.ActorState_ACTOR_STATE_SUSPENDED,
+		ActorVolumes:   initVols,
+		LatestSnapshot: sourceSnapshotStatus.GetSnapshot(),
+		SourceSnapshot: sourceSnapshotStatus,
+	}
+	if errs := validateActorUpdate(ctx, field.NewPath("actor"), outActor, inActor, true); len(errs) > 0 {
```
yongruilin, line 113: IIUC, this is a `create`, but it calls the `validateActorUpdate` which  runs with `operation.Update`. Since `outActor` is a clone of `inActor` + `status`, DV ratcheting would make it only validate the `status`?  Should we have a `validateActorCreate()` which use `operation.Create`?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -144,30 +144,76 @@ enum ActorSnapshotTagScope {
 // Selector matches worker pools by label.
 // Only equality-based matching is supported.
 message Selector {
+  // match_labels selects by exact equality on a set of labels.
+  //
+  // +k8s:optional
+  // +k8s:unionMember
+  // +k8s:minProperties=1
```
yongruilin, line 151: `+k8s:minProperties=1`  — the `+k8s:optional` with `OptionalMap` short-circuit runs before `MinProperties`, so an empty map never reaches it and a non-empty map trivially passes.

pkg/proto/ateapipb/ateapi.pb.go
```diff
@@ -1465,10 +1576,19 @@ func (x *Atespace) GetMetadata() *ResourceMetadata {
...
+	// References to non-atespaced resources should add `+k8s:subfield(atespace)=+k8s:forbidden`
+	// +k8s:optional
+	// +k8s:format=k8s-short-name
 	Atespace string `protobuf:"bytes,1,opt,name=atespace,proto3" json:"atespace,omitempty"`
-	// The name of the resource. Required. Unique within an atespace, or globally
-	// unique if the resource is global-scoped.
+	// The name of the referenced resource. This field is required.
+	//
+	// +k8s:required
+	// +k8s:format=k8s-short-name
```
yongruilin, line 1591: This is `k8s-short-name` while Actor.ActorTemplateName is `k8s-long-name`, is it intended?

cmd/ateapi/internal/controlapi/zz_generated.validation.go
```diff
@@ -0,0 +1,1367 @@
...
+					errs = append(errs, e...)
+					earlyReturn = true
+				}
+				if e := validate.Subfield(ctx, op, fldPath, obj, oldObj, "atespace",
+					func(o *ateapipb.ResourceMetadata) *string { return &o.Atespace }, validate.DirectEqual, validate.Immutable).MarkShortCircuit(); len(e) != 0 {
+					earlyReturn = true
+				}
+				if e := validate.Subfield(ctx, op, fldPath, obj, oldObj, "atespace",
+					func(o *ateapipb.ResourceMetadata) *string { return &o.Atespace }, validate.DirectEqual, validate.OptionalValue).MarkShortCircuit(); len(e) != 0 {
+					earlyReturn = true
```
yongruilin, line 72: Non-blocking -- but this definitely worth improvement. requiredness check followed by a optional check is redendant.

cmd/ateapi/internal/controlapi/zz_generated.validation.go
```diff
@@ -0,0 +1,1367 @@
...
+				if e := validate.Subfield(ctx, op, fldPath, obj, oldObj, "atespace",
+					func(o *ateapipb.ResourceMetadata) *string { return &o.Atespace }, validate.DirectEqual, validate.OptionalValue).MarkBeta().MarkShortCircuit(); len(e) != 0 {
+					earlyReturn = true
+				}
+				if e := validate.Subfield(ctx, op, fldPath, obj, oldObj, "atespace",
+					func(o *ateapipb.ResourceMetadata) *string { return &o.Atespace }, validate.DirectEqual, validate.Immutable).MarkBeta().MarkShortCircuit(); len(e) != 0 {
+					earlyReturn = true
+				}
+				if e := validate.Subfield(ctx, op, fldPath, obj, oldObj, "atespace",
+					func(o *ateapipb.ResourceMetadata) *string { return &o.Atespace }, validate.DirectEqual, validate.OptionalValue).MarkBeta().MarkShortCircuit(); len(e) != 0 {
```
yongruilin, line 449: `fobidden` followed by a `optional` doesn't look right. Should be a problem for validation-gen's subfield validator.

yongruilin (commented):
Late to the party, sorry — left some comments from DV perspective.

cmd/ateapi/internal/controlapi/actor.go
thockin: `operation.Create` and `operation.Update` don't apply as obviously in gRPC as they do it k8s REST.

Your understanding is correct, but we have an implicit `validateActorCreate` because the `CreateActorRequest` holds an Actor, so DV already recurses.  The point of this pattern is to only validate things set by us (it is an update as compared to the user's input)

pkg/proto/ateapipb/ateapi.proto
thockin: Yeah, it still serves as documentation :)

pkg/proto/ateapipb/ateapi.pb.go
thockin: We are moving templates from a CRD to a built-in resource, which does change the name format.

cmd/ateapi/internal/controlapi/zz_generated.validation.go
thockin: Lalit said this was a bug in how DV "inherits" tags from the subfield - it should have no impact but it's weird.  If the field was specified it will pass both required and optional.  If it was not specified it will "fail" both, but required actually produces an error, so that's what the final result is.

cmd/ateapi/internal/controlapi/zz_generated.validation.go
thockin: This is just how forbidden is implemented.

If the field was specified, forbidden will generate an error.  If the field was not specified, then optional avoids any further validation (e.g. type-attached rules)

PR #1216, opened 2026-08-25
ateapi: reject unknown fields on the request, not just the resource

### LINKED ISSUE #1213: Unknown fields in requests should be disallowed.
Today we call `validateNoUnknownFields()` on a request's resource (e.g. `CreateFooRequest.Foo`) but we don't check the request itself.

If someone, somehow, sends us a request that has a field we don't understand, we should probably balk, don't you think?

@juli4n @HavenXia @laoj2

HavenXia:
Fix #1213, we should also reject unknown fields at request level.

Refactor all functionality back into a gRPC interceptor to check every incoming request, no need to handle every new request method.

- [x] Tests pass


thockin:
Needs rebase, sorry

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -159,15 +159,15 @@ func validateCreateActorRequest(req *ateapipb.CreateActorRequest) field.ErrorLis
 	var fldPath *field.Path
 	var errs field.ErrorList
 
+	errs = append(errs, validateNoUnknownFields(req, fldPath)...)
```
thockin: The fldPath is nil - when you rebase you will find it easier to just say `nil`

thockin: Or just skip that argument - if we validate every request, it will always start as nil.

thockin (commented):
Should we do this on EVERY incoming request?  E.g. as an interceptor?

HavenXia:
> Should we do this on EVERY incoming request? E.g. as an interceptor?

Refactor into an ate interceptor.

HavenXia:
green

internal/ateinterceptors/rejectunknown.go
```diff
@@ -0,0 +1,83 @@
...
+		msg, ok := p.Index(-1).Value.Interface().(protoreflect.Message)
+		if !ok || len(msg.GetUnknown()) == 0 {
+			return nil
+		}
+		path := toFieldPath(p.Path)
+		if path == nil {
+			// Use "request" for the root message which has no field path.
+			path = field.NewPath("request")
+		}
+		errs = append(errs, field.Invalid(path, field.OmitValueType{}, ""))
```
thockin: We almost never use the last arg to "Invalid()" but this may be the exception -- should we say something like "unknown field with protobuf tag 99" ?

thockin (commented):
Small points then LGTM

internal/ateinterceptors/rejectunknown.go
HavenXia: SG, added.

HavenXia:
green again

PR #1218, opened 2026-08-26
feat: fix Envoy router to support WebSockets and add Jupyter notebook demo

### LINKED ISSUE #1201: Support bidirectional WebSocket streaming for interactive Substrate actors
## Expected Behavior

When deploying an interactive Substrate actor (like, say, Jupyter Notebook) via `kubectl ate create actor`, users should be able to access the UI in their browser and successfully open a terminal or execute kernel cells. The underlying WebSocket connection (`Upgrade: websocket`) used by such actors should cleanly establish a bidirectional duplex stream natively through the Substrate atenet-router (Envoy) directly back into the payload container.

## Actual Behavior

In the case of the Jupyter example, the UI loads perfectly in the browser, but opening a terminal results in a black screen, and kernel connections endlessly spin or immediately disconnect. The WebSocket connection fails silently because the Substrate `atenet-router` (Envoy) defaults to strict HTTP/1.1 tracing and scrubs Upgrade: websocket headers, dead-ending the handshake before the duplex stream can establish.

1. Build and run Substrate without Envoy `UpgradeConfigs` specified in `cmd/atenet/internal/router/xds.go`.
2. Apply an `ActorTemplate` for a workload that relies on WebSockets (e.g., `jupyter/base-notebook`).
3. Create the actor: `kubectl ate create actor jupyter-notebook -a demo --template ate-demo-jupyter/jupyter`
4. Open the Jupyter UI in your browser and attempt to launch a terminal or connect to a Python kernel endpoint.

## Specifications

- Version: aa1d14a
- Platform: kind

linde:
- Configured atenet-router (Envoy HttpConnectionManager) to explicitly allow `Upgrade: websocket`.
- Added a Substrate Jupyter notebook demo under `demos/jupyter/` that's browser friendly (handles CORS etc).
- Added a small proxy service to help with host-based routing in demo/exploration usage.

Fixes #1201

- [x] Tests pass

### DIFF
--- a/demos/jupyter/README.md
+++ b/demos/jupyter/README.md
@@ -95,7 +95,7 @@ To **resume** the notebook, you can either explicitly resume it via CLI:
 kubectl ate resume actor jupyter-notebook -a demo
 ```
 
-Or, even easier, you can rely on "transparent resume" — just refresh the page in your browser or make another request to the URL while it's suspended! Substrate will automatically restore its state and serve your request without any downtime. The kernel state (including any variables you defined) will be exactly as you left it.
+Or, even easier, you can rely on "transparent resume" — just refresh the page in your browser or make another request to the URL while it's suspended. Substrate will automatically restore its state and serve your request without any downtime. 
 
 ### Clean up
 
--- a/demos/jupyter/README.md
+++ b/demos/jupyter/README.md
@@ -1,0 +1,116 @@
+# Jupyter Notebook Demo
+
+This demo runs a Jupyter Notebook on Agent Substrate. It demonstrates how standard unmodified container images (like `jupyter/base-notebook`) can be deployed as Substrate actors that transparently suspend when idle and resume when you access them via your browser.
+
+## Prerequisites
+
+- A local Kubernetes cluster via Kind (or GKE).
+- `ko` installed for building images.
+- A GCS bucket for storing snapshots, or an in-cluster snapshot repository (handled automatically by the kind install script).
+
+## Quickstart on Kind
+
+For local development, Agent Substrate provides helper scripts to get a Kind cluster running and configured.
+
+1. **Create the Kind cluster:**
+
+   ```bash
+   ./hack/create-kind-cluster.sh
+   # This will create a local cluster named "ate-demo" and configure a local registry.
+   ```
+
+2. **Install Agent Substrate core system to the Kind cluster:**
+
+   ```bash
+   ./hack/install-ate-kind.sh --deploy-ate-system
+   ```
+
+3. **Deploy the Jupyter Notebook Demo:**
+
+   ```bash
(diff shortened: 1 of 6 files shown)


demos/jupyter/README.md
```diff
@@ -0,0 +1,116 @@
...
+# Create an atespace (required before creating actors)
+kubectl ate create atespace demo
+
+# Create the actor in the atespace via the jupyter template
+kubectl ate create actor jupyter-notebook -a demo --template ate-demo-jupyter/jupyter
+```
+
+### 2. Access Jupyter via the Proxy!
+
+Substrate routes HTTP traffic using the `Host` header. To make this easy without modifying local `/etc/hosts` files, this demo includes a lightweight NGINX reverse proxy (`jupyter-proxy`) that automatically injects the proper `Host` header (`jupyter-notebook.demo.actors.resources.substrate.ate.dev`) and forwards traffic internally to the Substrate router.
```
LiorLieberman, line 58: any future path / recommendation for a future path where we wont need this proxy?

demos/jupyter/README.md
```diff
@@ -0,0 +1,116 @@
...
+```
+Notice how it shows `STATUS_SUSPENDED`.
+
+To **resume** the notebook, you can either explicitly resume it via CLI:
+
+```bash
+kubectl ate resume actor jupyter-notebook -a demo
+```
+
+Or, even easier, you can rely on "transparent resume" — just refresh the page in your browser or make another request to the URL while it's suspended! Substrate will automatically restore its state and serve your request without any downtime. The kernel state (including any variables you defined) will be exactly as you left it.
```
LiorLieberman:
```suggestion
Or, even easier, you can rely on "transparent resume" — just refresh the page in your browser or make another request to the URL while it's suspended. Substrate will automatically restore its state and serve your request without any downtime. 
```

LiorLieberman (dismissed):
LGTM overall, two nits

demos/jupyter/README.md
linde: After some reflection, I'd suggest some sequence of checks we use in a deterministic order.

This could take the form of first checking to see if the `Host` header matches an actor (current behavior); next check for a different well known header along the lines of `X-substrate-actor`; and then fall back to a cookie with that key/value.  The key point is to expand to sequence of checks/fallbacks.

One note -- the cookie might feel like overkill since the header is great for most programmatic clients; it's just that browsers aren't in that group and could easily not be a corner case. Having something that works without browser extensions might be worth the complexity of another fallback check.  

We could definitely do this in stages and only cookie approach iff warranted. Main point of the suggestion is some sequence of checks.

LiorLieberman: ack, maybe put an issue on this so we dont lose track of this. I know its a demo but could be interesting to explore.

PR #1220, opened 2026-08-26
Crash actor when Ateom Checkpoint or Restore fails; Also when the error is not marked as retriable

### LINKED ISSUE #292: Add the CRASHED state to Actor State Machine

zoez7:
Fixes https://github.com/agent-substrate/substrate/issues/292 

This flips the condition to crash actors: any errors returned by atelet will cause actors to crash by default. 
If we later discover specific error classes that are safe to retry, we can explicitly mark them as retriable.

Follow ups to this: We should consider return a error code that signals "Pick another worker" if the current worker cannot be scheduled, but the actor snapshot themselves are intact, today the actor are stuck with retrying with the same worker.

### DIFF
--- a/cmd/atelet/internal/ategcs/classify.go
+++ b/cmd/atelet/internal/ategcs/classify.go
@@ -1,0 +1,94 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ategcs
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"syscall"
+
+	"google.golang.org/api/googleapi"
+
+	"github.com/agent-substrate/substrate/internal/ateerrors"
+)
+
+// tagTransientErr tags a backend failure of a retryable class — a 5xx/429/408
+// response, connection trouble, a stream cut mid-transfer — with
+// ateerrors.ReasonTransientObjectStorage: a fact for the RPC handler's call
+// site, which alone decides whether the request retries. Errors already
+// carrying a Reason, context errors, and local file-system errors pass
+// through untouched.
+func tagTransientErr(err error) error {
+	if err == nil {
+		return nil
+	}
+	// An error classified at the source (e.g. ReasonFailedGetExternalObject
+	// for a missing object) keeps its classification.
+	if _, ok := errors.AsType[ateerrors.Reason](err); ok {
+		return err
+	}
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return err
+	}
+	if !isTransientBackendErr(err) {
+		return err
+	}
+	return fmt.Errorf("%w: %w", ateerrors.ReasonTransientObjectStorage, err)
+}
+
+// isTransientBackendErr reports whether err is a backend failure of a class
+// worth retrying.
+func isTransientBackendErr(err error) bool {
+	if gerr, ok := errors.AsType[*googleapi.Error](err); ok {
+		return retryableHTTPStatus(gerr.Code)
+	}
+	if re, ok := errors.AsType[interface {
+		error
+		HTTPStatusCode() int
+	}](err); ok {
+		return retryableHTTPStatus(re.HTTPStatusCode())
+	}
+	// Skip network errors.
+	if _, ok := errors.AsType[*net.OpError](err); ok {
+		return true
+	}
+	return errors.Is(err, io.ErrUnexpectedEOF) ||
+		errors.Is(err, syscall.ECONNRESET) ||
+		errors.Is(err, syscall.ECONNREFUSED) ||
+		errors.Is(err, syscall.EPIPE)
+}
+
+func retryableHTTPStatus(code int) bool {
+	return code == http.StatusTooManyRequests ||
+		code == http.StatusRequestTimeout ||
+		code >= http.StatusInternalServerError
+}
+
+// transientReader applies tagTransientErr to failures surfacing mid-stream from
+// a reader handed to the caller, where they bypass the function-boundary
+// tagging in this package.
+type transientReader struct{ io.ReadCloser }
+
+func (r transientReader) Read(p []byte) (int, error) {
+	n, err := r.ReadCloser.Read(p)
+	if err != nil && err != io.EOF {
+		err = tagTransientErr(err)
+	}
+	return n, err
+}
--- a/cmd/atelet/internal/ategcs/objects.go
+++ b/cmd/atelet/internal/ategcs/objects.go
@@ -41,6 +41,11 @@ type ObjectStorage interface {
 }
 
 func FetchFromGCS(ctx context.Context, client ObjectStorage, gsURL string) ([]byte, error) {
+	content, err := fetchFromGCS(ctx, client, gsURL)
+	return content, tagTransientErr(err)
+}
+
+func fetchFromGCS(ctx context.Context, client ObjectStorage, gsURL string) ([]byte, error) {
 	ctx, span := tracer.Start(ctx, "fetchFromGCS")
 	defer span.End()
 
@@ -66,6 +71,11 @@ func FetchFromGCS(ctx context.Context, client ObjectStorage, gsURL string) ([]by
 // Open streams the object at gsURL; the caller must Close the returned reader.
 // Unlike FetchFromGCS it does not buffer the whole object in memory.
 func Open(ctx context.Context, client ObjectStorage, gsURL string) (io.ReadCloser, error) {
+	rc, err := openObject(ctx, client, gsURL)
+	return rc, tagTransientErr(err)
+}
+
+func openObject(ctx context.Context, client ObjectStorage, gsURL string) (io.ReadCloser, error) {
 	bucket, object, err := parseGCSURL(gsURL)
 	if err != nil {
 		return nil, fmt.Errorf("%w: while parsing url: %w", ateerrors.ReasonInvalidObjectURL, err)
@@ -74,12 +84,18 @@ func Open(ctx context.Context, client ObjectStorage, gsURL string) (io.ReadClose
 	if err != nil {
 		return nil, fmt.Errorf("while getting object bucket=%q object=%q: %w", bucket, object, err)
 	}
-	return rc, nil
+	// Mid-stream failures surface from the caller's reads, not from this
+	// function, so the reader carries its own tagging.
+	return transientReader{rc}, nil
 }
 
 // SendBytesToGCS uploads the given bytes (uncompressed) to gsURL. Intended for
 // small objects such as the snapshot manifest.
 func SendBytesToGCS(ctx context.Context, client ObjectStorage, gsURL string, content []byte) error {
+	return tagTransientErr(sendBytesToGCS(ctx, client, gsURL, content))
+}
+
+func sendBytesToGCS(ctx context.Context, client ObjectStorage, gsURL string, content []byte) error {
 	ctx, span := tracer.Start(ctx, "sendBytesToGCS")
 	defer span.End()
 
@@ -93,7 +109,11 @@ func SendBytesToGCS(ctx context.Context, client ObjectStorage, gsURL string, con
 	return nil
 }
 
-func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, localFilePath string) (err error) {
+func SendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, localFilePath string) error {
+	return tagTransientErr(sendLocalFileToGCSWithZstd(ctx, client, gsURL, localFilePath))
+}
+
+func sendLocalFileToGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, localFilePath string) (err error) {
 	ctx, span := tracer.Start(ctx, "sendLocalFileToGCSWithZstd")
 	defer span.End()
 
@@ -297,7 +317,11 @@ func plainZstd(w io.Writer, src io.Reader) (int64, error) {
 	return n, zw.Close()
 }
 
-func FetchLocalFileFromGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, localFilePath string) (err error) {
+func FetchLocalFileFromGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, localFilePath string) error {
+	return tagTransientErr(fetchLocalFileFromGCSWithZstd(ctx, client, gsURL, localFilePath))
+}
+
+func fetchLocalFileFromGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, localFilePath string) (err error) {
 	ctx, span := tracer.Start(ctx, "fetchLocalFileFromGCSWithZstd")
 	defer span.End()
 
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -458,7 +458,12 @@ func NewService(
 	return wms
 }
 
-func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (resp *ateletpb.RunResponse, err error) {
+func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (*ateletpb.RunResponse, error) {
+	resp, err := s.run(ctx, req)
+	return resp, ateerrors.CrashUnlessRetriable(ctx, err)
+}
+
+func (s *AteomHerder) run(ctx context.Context, req *ateletpb.RunRequest) (*ateletpb.RunResponse, error) {
 	if err := validateRunRequest(req); err != nil {
 		// status.Error so the interceptor surfaces InvalidArgument and the
 		// message instead of masking both as Internal.
@@ -474,6 +479,10 @@ func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (resp *
 	}
 	assetPaths, err := s.ensureSandboxAssets(ctx, sandboxRec)
 	if err != nil {
+		// Nothing ran yet; transient storage trouble heals on retry.
+		if errors.Is(err, ateerrors.ReasonTransientObjectStorage) {
+			return nil, ateerrors.MarkRetriable(err)
+		}
 		return nil, err
 	}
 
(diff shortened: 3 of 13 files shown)


cmd/atelet/internal/ategcs/objects.go
```diff
@@ -40,7 +40,8 @@ type ObjectStorage interface {
 	PutObject(ctx context.Context, bucket, object string, reader io.Reader) error
 }
 
-func FetchFromGCS(ctx context.Context, client ObjectStorage, gsURL string) ([]byte, error) {
+func FetchFromGCS(ctx context.Context, client ObjectStorage, gsURL string) (_ []byte, err error) {
+	defer func() { err = tagTransient(err) }()
```
sfunkenhauser: I personally find these types of defer funcs (that mutate values) make it very difficult to understand the code path. Can we make these transformations explicit, either by enumerating the return sites if there are only a couple, or moving the logic to a private function and explicitly tagging the return value.

eg.

```
func FetchFromGCS() {
  b, err := fetch()
  return b, tagTransient(err)
}

func fetch() {
...
}
```

This seems to be a new pattern that is introduced through this PR, and I'd suggest removing all of them.

cmd/ateom-microvm/health.go
```diff
@@ -0,0 +1,150 @@
...
+	for _, wid := range ra.workloadIDs {
+		// The agent is handed over rather than re-read in the goroutine:
+		// teardownActor nils ra.guestAgent under s.lock, which the waiter
+		// does not hold.
+		go s.watchWorkloadExit(ctx, ra, attribution, ra.guestAgent, wid)
+	}
+}
+
+func (s *AteomService) watchWorkloadExit(ctx context.Context, ra *runningActor, attribution resources.ActorAttribution, agent *kata.AgentClient, wid string) {
+	wait := s.waitWorkloadExit
```
sfunkenhauser: Instead of having test specific overrides for nil, can we provide the real version of these methods when constructing the AteomService?

internal/ateomnet/net.go
```diff
@@ -405,24 +405,25 @@ func NetNSDo(ctx context.Context, targetNS netns.NsHandle, do func(context.Conte
...
 	curNetNS, err := netns.Get()
 	if err != nil {
 		return fmt.Errorf("while getting current netns: %w", err)
 	}
 	// Registered before the restoring defer below since deferred calls are LIFO.
 	defer curNetNS.Close()
+
+	// The switch never happened: the thread is still in its original
+	// netns, so there is nothing to restore.
+	if err := netns.Set(targetNS); err != nil {
```
sfunkenhauser: Is this re-ordering required for the crash changes? Seems like it might be an unrelated change that we should put in a separate PR?

cmd/atelet/internal/ategcs/objects.go
zoez7: Done in 6b640247dfa2b33cc84351a1fedbf9e330ca3bb2. I agree, "defer" has a line dependency issue that this method removes.

cmd/ateom-microvm/health.go
zoez7: Done in 6b640247dfa2b33cc84351a1fedbf9e330ca3bb2.

cmd/atelet/main.go
```diff
@@ -1698,6 +1752,17 @@ func (s *AteomHerder) dialAteom(ctx context.Context, targetAteomUid string) (ate
 	return ateompb.NewAteomClient(conn), nil
 }
 
+// markAteomTransportRetriable keeps an unreachable ateom retriable: a restart
+// mid-rollout is routine and surfaces as a transport-level Unavailable on the
+// workload RPC (never at dialAteom, which dials lazily). ateom handlers never
+// return Unavailable themselves, so the code alone identifies the transport.
```
BenTheElder: 🤖 **nit** 🟢 – The invariant this rests on isn't true: both ateoms return `Unavailable` from `rejectIfDraining` ("worker draining: not accepting new workloads"), in `cmd/ateom-gvisor/main.go:446` and `cmd/ateom-microvm/main.go:591`.

The behaviour is still right, which is why nothing breaks — a draining worker *should* be retried, and `cmd/ateom-microvm/main.go:380` says so outright ("rejected with codes.Unavailable so the control plane reschedules"). So the code lands on the correct answer for the one case that falsifies the comment.

Worth restating anyway, because fail-closed makes the cost of a future misreading high. The rule that actually holds is "every `Unavailable` from ateom is retriable, whether it came from the transport or from a handler" — not "only the transport produces it". As written, someone adding an ateom-side `Unavailable` for a reason that *shouldn't* be retried would have this comment telling them the code can't reach here.

cmd/atelet/sandbox_assets.go
```diff
@@ -448,7 +449,7 @@ func (s *AteomHerder) openAsset(ctx context.Context, url string) (io.ReadCloser,
 	}
 	rc, mainErr := ategcs.Open(ctx, s.gcsClient, url)
 	if mainErr != nil {
-		return nil, fmt.Errorf("anonymous open failed (%v); main client open failed: %w", anonErr, mainErr)
+		return nil, fmt.Errorf("anonymous open failed (%w); main client open failed: %w", anonErr, mainErr)
```
dberkov, line 452: **Anon-client `%w` puts the wrong error in the unwrap chain.** The `%v→%w` change here means a transient-tagged anonymous-leg failure can mask a definitive main-client failure: if anonymous egress to `storage.googleapis.com` is blocked (dial refused → `*net.OpError` → tagged `ReasonTransientObjectStorage`) while the main client gets a definitive 404 (`ReasonFailedGetExternalObject`), the combined error satisfies `errors.Is(err, ReasonTransientObjectStorage)` via the anon leg. `run()`/`checkpoint()`/`restore()` then `MarkRetriable` and return `Unavailable`, so the control plane retries a permanently missing asset forever — the exact infinite-retry pathology this PR exists to eliminate. `errors.AsType[Reason]` also reports the anon leg's reason first (DFS over multi-`%w`). Pre-PR, `%v` kept `anonErr` out of the chain; consider keeping it out, or ensuring the main-client error's reason wins.

internal/imagecache/imagecache.go
```diff
@@ -378,6 +383,48 @@ func (s *Store) EnsureImage(ctx context.Context, ref string) (_ *Image, err erro
...
+	}
+	if _, ok := errors.AsType[ateerrors.Reason](err); ok {
+		return err
+	}
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return err
+	}
+	if regErr, ok := errors.AsType[*transport.Error](err); ok {
+		if regErr.Temporary() {
+			return fmt.Errorf("%w: %w", ateerrors.ReasonTransientImageRegistry, err)
```
dberkov, line 404: **`transport.Error.Temporary()` classifies on parsed JSON body codes, not the HTTP status.** In go-containerregistry v0.21.7, when the body parses into `Errors`, `Temporary()` ignores `StatusCode` and requires every code ∈ {BLOB_UPLOAD_INVALID, TOOMANYREQUESTS, UNKNOWN, UNAVAILABLE}. Two failure modes:

- Artifact Registry returns **403 with `{"errors":[{"code":"DENIED"}]}`** during IAM/token propagation → `Temporary()==false` → tagged `ReasonFailedGetExternalObject` → actor crashes even though a retry would succeed. (ggcr's own retry layer even retries 503s regardless of body, so a 503-with-DENIED-body is retried by the SDK and then declared "definitive" here.)
- A **404 with body code `UNKNOWN`** is tagged `ReasonTransientImageRegistry` → infinite retry of a missing image.

Classify on `StatusCode` (like ategcs `retryableHTTPStatus`) instead.

cmd/atelet/main.go
```diff
@@ -989,9 +1035,9 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 			snapshotLogAttrs(attribution, op, restoreDurationMetric, outcome, phases)...)
 	}()
 
-	// Not crashing the actor, because terminal errors here indicate problems with atelet,
-	// node or the disk itself.
 	if err := resetActorDirs(actorUID); err != nil {
+		// TODO: return an error code that signals the control plane to
+		// reschedule the actor onto a different node when this happens.
 		return nil, fmt.Errorf("while resetting actor dirs: %w", err)
 	}
```
dberkov, line 1043: **External-volume mount failures (just below, and in `run()` ~line 493) are neither transient-tagged nor `MarkRetriable`'d.** A routine CSI `ABORTED` ("operation pending for volume", i.e. a NodeStage/NodePublish still in flight) or `UNAVAILABLE` (driver pod restarting) — codes the CSI spec defines as retry-me — now permanently crashes an actor whose snapshot is intact. The status survives the `%w` wraps (`internal/volume/csi/plugin.go:195,213`; `cmd/atelet/volumes.go:47`), and `CrashUnlessRetriable`'s pass-through set is only Canceled/DeadlineExceeded/InvalidArgument/FailedPrecondition, so the crash directive is appended and ateapi's `maybeCrashActor` sets terminal `ACTOR_STATE_CRASHED`. Pre-PR the untagged error passed `CrashIfReason` unchanged and the workflow retried. Consider passing through `ABORTED`/`UNAVAILABLE` from the CSI plugin or marking these retriable.

cmd/atelet/internal/ategcs/classify.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	// Skip network errors.
+	if _, ok := errors.AsType[*net.OpError](err); ok {
+		return true
+	}
+	return errors.Is(err, io.ErrUnexpectedEOF) ||
+		errors.Is(err, syscall.ECONNRESET) ||
+		errors.Is(err, syscall.ECONNREFUSED) ||
+		errors.Is(err, syscall.EPIPE)
+}
```
dberkov, line 75: **The hand-rolled transient allowlist is narrower than the SDKs' own retry predicates.** `isTransientBackendErr` here (and `isNetworkErr` in `internal/imagecache/imagecache.go`) misses fault classes the SDKs call retryable: TLS handshake timeout, `url.Error`-wrapped `io.EOF`, `net.ErrClosed`, temporary DNS errors, "http2: client connection lost". Example: Restore fetches the snapshot manifest over a reused connection the server closed → `url.Error{Err: io.EOF}` — not a `*net.OpError`, not `ErrUnexpectedEOF`, no errno — so `tagTransientErr` passes it through untagged and `CrashUnlessRetriable` crashes the actor on a one-shot blip. `cloud.google.com/go/storage`'s exported `ShouldRetry` and ggcr's `defaultRetryPredicate` both retry these exact classes; GCS uploads without preconditions get no SDK-internal retry at all, making `UploadPausedCheckpoint` the most exposed path. Suggest delegating to `storage.ShouldRetry` rather than maintaining a narrower copy.

P.S. - storage retry on 429 has been added to #1416 for GCS only

cmd/atelet/main.go
```diff
@@ -1715,6 +1769,16 @@ func (s *AteomHerder) dialAteom(ctx context.Context, targetAteomUid string) (ate
 	return ateompb.NewAteomClient(conn), nil
 }
 
+// markAteomTransportRetriable keeps an unreachable ateom retriable: a restart
+// mid-rollout is routine and surfaces as a transport-level Unavailable on the
+// workload RPC (never at dialAteom, which dials lazily).
+func markAteomTransportRetriable(err error) error {
+	if status.Code(err) == codes.Unavailable {
+		return ateerrors.MarkRetriable(err)
```
dberkov, line 1777: **`codes.Unavailable` does not imply "the workload never saw the request".** gRPC also returns Unavailable when the connection dies *after* the handler ran: if ateom is OOM-killed during `CheckpointWorkload` after destroying the sandbox (ateom-gvisor checkpoints, then tears down before replying), grpc-go closes all active streams with Unavailable — atelet marks it retriable, and the control plane retries Checkpoint against a torn-down workload, which then fails unclassified and crashes as UNKNOWN with the real cause (interrupted checkpoint) invisible.

The reverse also holds: grpc-go's `http2ErrConvTab` maps RST_STREAM/stream-level breakage to `codes.Internal`, which now gets the crash directive even when the request was never processed — crashing a healthy actor during routine ateom restarts.

cmd/atelet/main.go
```diff
@@ -989,9 +1035,9 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 			snapshotLogAttrs(attribution, op, restoreDurationMetric, outcome, phases)...)
 	}()
 
-	// Not crashing the actor, because terminal errors here indicate problems with atelet,
-	// node or the disk itself.
 	if err := resetActorDirs(actorUID); err != nil {
+		// TODO: return an error code that signals the control plane to
```
dberkov, line 1039: **Deleting the carve-out makes a node-local fault permanently destroy actors with intact snapshots.** The removed comment ("Not crashing the actor, because terminal errors here indicate problems with atelet, node or the disk itself") was deliberate, and this TODO admits the correct behavior (reschedule to another node) isn't implemented. If one node's disk fills or turns read-only, every actor resumed onto it fails `resetActorDirs` before anything runs; the error reaches `CrashUnlessRetriable`, which returns DataLoss + `ActorCrashedMetadata`, and ateapi sets terminal `ACTOR_STATE_CRASHED` and releases the worker — actors whose snapshots are fully intact in GCS are permanently destroyed. Pre-PR they stayed retriable. Suggest keeping this path retriable (or non-crashing) until rescheduling exists.

internal/ateerrors/ateerrors.go
```diff
@@ -127,15 +134,71 @@ func NewGRPCError(ctx context.Context, grpcCode codes.Code, reason Reason, metad
...
+	}
+	r, ok := errors.AsType[Reason](err)
+	if !ok {
+		r = ReasonUnknown
+	}
+	var grpcErr interface{ GRPCStatus() *status.Status }
+	if errors.As(err, &grpcErr) {
+		st := grpcErr.GRPCStatus()
+		switch st.Code() {
+		case codes.Canceled, codes.DeadlineExceeded,
```
dberkov, line 188: **The `InvalidArgument`/`FailedPrecondition` pass-through can't distinguish request validation from downstream deterministic rejections.** ateom-microvm/gvisor return `InvalidArgument` for deterministic spec-projection rejections and `FailedPrecondition` from checkpoint preconditions; `restore()` itself returns `FailedPrecondition` for a golden/actor sandbox-class mismatch after fetching manifests. `errors.As` finds the wrapped status ("while calling ateom.CheckpointWorkload: %w"), the boundary returns it uncrashed, `maybeCrashActor` doesn't crash, and every subsequent resume re-runs the identical deterministic failure — the fail-open pathology this PR's contract claims to remove. Fix at the right depth: run `validateXRequest` *outside* the crash-wrapped inner function instead of inferring "request-level" from codes.

internal/ateerrors/ateerrors.go
```diff
@@ -127,15 +134,71 @@ func NewGRPCError(ctx context.Context, grpcCode codes.Code, reason Reason, metad
...
+// CrashUnlessRetriable will append ActorCrashedMetadata unless the
+// error was marked as retriable.
+func CrashUnlessRetriable(ctx context.Context, err error) error {
+	if err == nil {
+		return nil
+	}
+	if ActorCrashRequested(err) {
+		return err
+	}
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
dberkov, line 170: **Checking the error chain instead of the RPC's own ctx lets library-internal timeouts escape unclassified.** `net/http`'s `timeoutError` implements `Is(context.DeadlineExceeded)` even while the RPC context is alive, so a storage-client HTTP transport timeout during `UploadPausedCheckpoint` passes through raw: `tagTransientErr` deliberately skips context-matching errors, the `ReasonTransientObjectStorage` check misses, and the control plane gets a bare DeadlineExceeded with neither a crash directive nor a TRANSIENT classification — exactly the ambiguous unclassified state this PR exists to eliminate. (`TestCheckpointExternalUploadFailureCrashes` pins a workaround at Checkpoint's persist leg only; every other timeout path lacks it.) Gating the pass-through on `ctx.Err() != nil` closes the hole in one place.

cmd/atelet/internal/ategcs/objects.go
```diff
@@ -297,7 +317,11 @@ func plainZstd(w io.Writer, src io.Reader) (int64, error) {
 	return n, zw.Close()
 }
 
-func FetchLocalFileFromGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, localFilePath string) (err error) {
+func FetchLocalFileFromGCSWithZstd(ctx context.Context, client ObjectStorage, gsURL string, localFilePath string) error {
+	return tagTransientErr(fetchLocalFileFromGCSWithZstd(ctx, client, gsURL, localFilePath))
```
dberkov, line 321: **Pre-existing inverted `rc.Close` defer in `fetchLocalFileFromGCSWithZstd` (~line 366) now destroys the transient tag this wrapper adds.** The defer reads `if err != nil { err = closeErr }` — inverted relative to its two sibling defers: it overwrites a real, transient-classifiable download error with the close error, and on success drops `closeErr`. Example: a snapshot download dies mid-stream with ECONNRESET (would be tagged `ReasonTransientObjectStorage`); `rc.Close()` on the broken stream also errors with a non-classifiable error, the defer replaces the real error, `tagTransientErr` finds nothing to tag, and `CrashUnlessRetriable` crashes the actor as UNKNOWN for a network blip. Harmless before this PR (untagged == retriable); should be flipped to match the siblings: `if err == nil { err = closeErr }`.

cmd/atelet/main.go
```diff
@@ -678,27 +703,34 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
...
 	}
 
-	// Note: we do not crash the actor if resetting the directory fails.
-	if err := resetActorDirs(actorUID); err != nil {
-		return nil, fmt.Errorf("while resetting actor dirs: %w", err)
-	}
+	// Best-effort: the snapshot is already persisted, so failing (and thereby
+	// crashing) here would lose a healthy actor over cleanup. Run and Restore
+	// reset these dirs again before reusing them.
+	resetActorDirsBestEffort(ctx, actorUID)
```
dberkov, line 733: **If the actor never returns to this node, nothing else cleans the checkpoint-state dir.** The justification "Run and Restore reset these dirs again before reuse" only holds when the actor comes back: the delete workflow skips Terminate when the worker assignment is already released (`workflow_delete.go:115-119`), and atelet has no actor-dir GC or startup sweep. So if `resetActorDirs` fails once here (e.g. EBUSY from a leaked mount), the multi-GB memory-ranges files persist on the hostPath for the node's lifetime — no metric, no retry. Accumulated per suspended actor this drives the node toward ENOSPC, which under the new regime permanently crashes other actors whose Restores hit the full disk. Consider a startup sweep or GC for orphaned actor dirs, plus a metric on cleanup failure.

HavenXia:
need rebase

cmd/atelet/sandbox_assets.go
zoez7: Resolved this comment since its based on a old version of the PR.

internal/imagecache/imagecache.go
zoez7: Resolved this comment (and below) since its based on a old version of the PR.

juli4n (dismissed):
Needs rebase

juli4n (approved):

PR #1224, opened 2026-08-26
microvm: propagate guest_hook_path so guest OCI hooks actually run

### LINKED ISSUE #1223: guest_hook_path is never set, so guest OCI hooks are silently ignored
`CreateSandboxRequest` carries a `GuestHookPath` field and ateom never sets it.

The kata-agent only scans for guest hooks when that field is non-empty
(`src/agent/src/rpc.rs`):

```rust
if !req.guest_hook_path.is_empty() {
    let _ = s.add_hooks(&req.guest_hook_path).map_err(|e| {
```

So a guest image carrying hooks under `/usr/share/oci/hooks` has them ignored.
Nothing logs it and nothing fails, which makes it read as a broken hook rather
than a runtime that never looked for one.

Found while adding an OCI prestart hook to a microVM guest image: the hook was
correct, executable, in the right directory, and simply never ran. The only way
to tell was reading the agent source.

Happy to send the fix, it is a couple of lines.

Harihara04sudhan:
Fixes #1223

The kata-agent only scans for guest hooks when `guest_hook_path` is non-empty,
and ateom never set it, so hooks baked into a guest image under
`/usr/share/oci/hooks` were never discovered. Nothing logs it and nothing fails,
so the symptom looks like a broken hook rather than a runtime that never looked.

Sets it to the conventional path the Kata guest images use. A guest with no such
directory is unaffected: `add_hooks` finds nothing and the sandbox comes up
exactly as before.

Verification, on `linux/amd64`:

- `gofmt` clean
- `go build ./cmd/ateom-microvm/...` and `go vet ./cmd/ateom-microvm/...` clean
- `go test ./cmd/ateom-microvm/...` passes, with one exception unrelated to this
  change: `tarutil.TestRoundTripDeviceNode` fails with
  `creating whiteout device node: operation not permitted`. It fails identically
  on unmodified `main` in the same environment, including `--privileged`, because
  the container filesystem will not accept device nodes. Not caused by this patch
  and not touched by it.

- [x] Tests pass

No documentation change: this makes an existing field behave as documented.

### DIFF
--- a/cmd/ateom-microvm/internal/kata/overlay_linux.go
+++ b/cmd/ateom-microvm/internal/kata/overlay_linux.go
@@ -100,6 +100,10 @@ func UpperWorkDirs(upperBase, containerID string) (upper, work string) {
 // stock kata flow.
 func GuestSharedRootfs(containerID string) string { return guestSharedDir + containerID + "/rootfs" }
 
+// GuestHookPath is where the guest image carries OCI hooks, in
+// {prestart,poststart,poststop} subdirectories.
+const GuestHookPath = "/usr/share/oci/hooks"
+
 // GuestSharedVolumeDir is the in-guest path one image volume's contents appear
 // at, beside the container's rootfs in the same kataShared tree.
 func GuestSharedVolumeDir(containerID, volumeName string) string {
@@ -384,10 +388,13 @@ func (a *AgentClient) CreateSandboxForActor(ctx context.Context, opts CreateSand
 		Fstype:     typeVirtioFS,
 		MountPoint: guestSharedDir,
 	}}
+	// GuestHookPath must be set or the agent never scans for guest hooks, which
+	// makes the documented guest_hook_path mechanism inert.
 	return a.CreateSandbox(ctx, &agentpb.CreateSandboxRequest{
-		Hostname:  opts.Hostname,
-		SandboxId: opts.SandboxID,
-		Storages:  storages,
+		Hostname:      opts.Hostname,
+		SandboxId:     opts.SandboxID,
+		Storages:      storages,
+		GuestHookPath: GuestHookPath,
 	})
 }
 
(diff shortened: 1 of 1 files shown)


BenTheElder (changes requested):
> [...] and ateom never set it, so hooks baked into a guest image under
> /usr/share/oci/hooks were never discovered.

What hooks? Is there a real use case for this?

kata doesn't enable these by default either:
https://github.com/kata-containers/kata-containers/blob/main/docs/design/hooks-handling.md#kata-hooks

Harihara04sudhan:
We answered both of these in [our comment on #1223](https://github.com/agent-substrate/substrate/issues/1223#issuecomment-5541762981).

On what hooks: one guest-side OCI `prestart` hook shipped in the guest image. It writes the objective's rules into our pseudo-filesystem and exits. No network calls and no new agent capability. `createRuntime` or `createContainer` would work just as well, we used this path because it already exists.

On kata not enabling them by default: you're right, and we're not asking Substrate to either. Setting the path is what lets an image opt in, and an image without that directory is unaffected. The part we'd still call a bug is the silence, since a guest image carrying hooks today gives no error and no log, so you cannot tell whether they ran.

The use case and the fuller detail are in [intentd on Agent Substrate: the use case, and what we are asking for](https://docs.google.com/document/d/1wln6Gg2Bw2UisVzvDpOXZNKjkfBaqrltXfYs4y4-PkY/view).

Harihara04sudhan:
We answered both of these in [our comment on #1223](https://github.com/agent-substrate/substrate/issues/1223#issuecomment-5541762981).

On what hooks: one guest-side OCI `prestart` hook shipped in the guest image. It writes the objective's rules into our pseudo-filesystem and exits. No network calls and no new agent capability. `createRuntime` or `createContainer` would work just as well, we used this path because it already exists.

On kata not enabling them by default: you were right, and the first version of this patch did the thing you objected to. It passed a hardcoded path on every `CreateSandbox` call, so scanning was on for every sandbox with no way to turn it off.

Pushed a second commit for that. The path is a field on `CreateSandboxOpts` now and empty is the default, so the agent never scans unless a caller sets it, and `run.go` leaves it unset. Nothing changes for anyone until a deployment opts in, and the constant is only a suggested location rather than something the runtime imposes.

The one part we'd still call a bug is the silence: a guest image carrying hooks today gives no error and no log, so you cannot tell whether they ran.

The use case and the fuller detail are in [intentd on Agent Substrate: the use case, and what we are asking for](https://docs.google.com/document/d/1wln6Gg2Bw2UisVzvDpOXZNKjkfBaqrltXfYs4y4-PkY/view).

PR #1229, opened 2026-08-26
atenet-router: take parking-lot slots at the park transition, not at admission

### LINKED ISSUE #1081: atenet-router: requests to running actors consume parking-lot slots
The parking lot admits every request, not just ones waiting for a resume: [HandleRequestHeaders](https://github.com/agent-substrate/substrate/blob/a0523de4e7e81a70734511b32665b3072d2d9682/cmd/atenet/internal/router/ingress/ingress.go#L119-L130) calls parking.enter() before ResumeActor, so even a request to a RUNNING actor holds a slot during its ateapi lookup.

This breaks the [design's own guarantee](https://github.com/agent-substrate/substrate/blob/main/docs/request-parking.md#behavior)
> a saturated lot cannot starve requests to already-running actors, at any lot size

[TestHandleRequestHeaders_ParkingLotFull](https://github.com/agent-substrate/substrate/blob/a0523de4e7e81a70734511b32665b3072d2d9682/cmd/atenet/internal/router/ingress/ingress_test.go#L343) explicitly asserts the contradicting behavior: a request to an already-running actor "is shed with a 503 before any resume is attempted" when the lot is full - i.e., the starvation the design doc says cannot happen is locked in as expected behavior.

omeryahud:
Fixes agent-substrate/substrate#1081

## Context

`HandleRequestHeaders` admitted **every** request to the parking lot before its `ResumeActor` lookup, so a request to an already-`RUNNING` actor held a slot for the duration of a millisecond-scale control-plane call — and once the lot was full, such requests were shed with `503 "router at capacity"` before the lookup even ran. That contradicts the design guarantee in `docs/request-parking.md` ("a saturated lot cannot starve requests to already-running actors, at any lot size"): the fast-path headroom built into the 2× ext_proc circuit-breaker sizing was nullified by the lot check sitting in front of the resume, and one saturated WorkerPool could black-hole routing for every other pool's healthy actors. `TestHandleRequestHeaders_ParkingLotFull` pinned the broken behavior as expected.

## Change

A caller now occupies a lot slot only from the moment its resume actually *parks* — the flight's first retryable failure — so the guarantee holds by construction, with no status probe:

- **Flight registry replaces `singleflight.Group`** (`resumeFlight{parked, done, result}` in a mutex-guarded map, split into `flight.go`). The registry keeps singleflight's exact semantics — per-actor dedup, forget-on-completion, leader/joiner outcome labels, caller-cancel-doesn't-abort-flight — and adds the one hook `singleflight` cannot express: a `parked` channel closed at the first retryable error. A side-map next to `Group` can't do this race-free, because `DoChan` never tells a joiner which flight generation it attached to.
- **Two-phase caller wait**: slot-free while the flight resolves; on the `parked` signal the caller must acquire a slot (`lot.enter`) to keep waiting, and is shed with the existing `503 "router at capacity"` when the lot is full — after exactly the one resume attempt that revealed it would have to wait.

| Scenario | Before | After |
|---|---|---|
| Full lot + RUNNING actor | shed 503, no lookup | **served** — never touches the lot |
| Full lot + request that must wait | shed before any attempt | shed after one attempt (per-actor deduped) |
| Parking disabled (`Max=0`) | unchanged | unchanged |
| `parking.active` / `wait.duration` | counted every lookup | count only genuinely parked requests |

- [x] Tests pass


chuangw6:
cc @bowei

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -162,118 +182,158 @@ func (r *ActorResumer) retryable(err error) bool {
...
+		return f.callerResult(reqID)
+	case <-f.parked:
+	}
 
-		// On error, return ResumeOutcomeNone ("none") so the failure is tagged
-		// under the 'outcome' label rather than misreported as an activation.
-		if callRes.err != nil {
-			return nil, ResumeOutcomeNone, callRes.err
-		}
+	// The flight parked. If its result raced in anyway, serve it without
```
bowei: Do we really need this extra `select`? We got out of the `select` from `case <-f.parked` and not `<-f.done`, this should be extra.

Please keep the `// Flight parked` comment.

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -162,118 +182,158 @@ func (r *ActorResumer) retryable(err error) bool {
...
-			if res.Err != nil {
-				return nil, ResumeOutcomeNone, res.Err
-			}
-			return nil, ResumeOutcomeNone, status.Error(codes.Internal, "resume call returned nil result")
-		}
+	case <-f.done:
+		// Fast path: the flight resolved without ever parking (or finished
+		// before this caller reacted to parking) — the lot is never touched.
+		return f.callerResult(reqID)
+	case <-f.parked:
```
bowei: I would add a comment here to remind readers that Golang select on a closed channel will always return here -- meaning that future requests encountering a parked request will reach the parked branch.

bowei (commented):
Sorry for taking so long reviewing this. The logic looks ok, although I am thinking we may be able to simplify this by explicitly maintaining the state of the request in the flight map instead of relying on channel state.

I am ok with taking this PR in and doing the fix in a follow-up.

cmd/atenet/internal/router/ingress/flight.go
```diff
@@ -0,0 +1,103 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// resumeFlight is one in-flight, per-actor resume that concurrent requests
+// share. It replaces a singleflight.Group entry so callers can observe the
+// flight mid-run: parked exposes the park transition, which singleflight's
+// result-only channel cannot. Its lifecycle is park (at most once, flight
+// goroutine only) followed by ActorResumer.publish (exactly once).
+type resumeFlight struct {
```
bowei: Let's name this resumeActorFlight

cmd/atenet/internal/router/ingress/flight.go
```diff
@@ -0,0 +1,103 @@
...
+
+package ingress
+
+import (
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// resumeFlight is one in-flight, per-actor resume that concurrent requests
```
bowei:
```
// resumeActorFlight encapsulates a single per-actor resume that aggregates
// multiple requests to avoid initiating separate RPCs in parallel for the
// same Actor.
//
// There are two signals for the flight that users can `select` on:
//
// - retrying chan: channel is closed when the request is retrying. The request
//   should attempt to enter the parking lot in this state.
// - done chan: this is closed when the request is finished.
```

cmd/atenet/internal/router/ingress/flight.go
```diff
@@ -0,0 +1,103 @@
...
+// share. It replaces a singleflight.Group entry so callers can observe the
+// flight mid-run: parked exposes the park transition, which singleflight's
+// result-only channel cannot. Its lifecycle is park (at most once, flight
+// goroutine only) followed by ActorResumer.publish (exactly once).
+type resumeFlight struct {
+	// parked is closed at most once, by park, at the flight's first retryable
+	// error — the moment the flight stops resolving and starts waiting. A
+	// caller woken by it (or attaching after it) must hold a parking-lot slot
+	// to keep waiting. Never closed on the fast path.
+	parked chan struct{}
```
bowei: This should be called `retrying`, not parked, as flights will never be parked -- requests are parked.

cmd/atenet/internal/router/ingress/flight.go
```diff
@@ -0,0 +1,103 @@
...
+	// parkedSignaled guards the parked close. Only the flight goroutine calls
+	// park, so a plain bool suffices.
+	parkedSignaled bool
```
bowei: Fix the name of this to match (retyringSignaled)

cmd/atenet/internal/router/ingress/flight.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func newResumeFlight() *resumeFlight {
+	return &resumeFlight{parked: make(chan struct{}), done: make(chan struct{})}
+}
+
+// park signals the flight's park transition: it stopped resolving and started
+// waiting, so from here on callers must hold parking-lot slots to keep
+// waiting. Idempotent; only the flight goroutine may call it.
+func (f *resumeFlight) park() {
```
bowei: park -> signalRetrying()

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -116,19 +119,13 @@ func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestM
...
-	// in-flight the request occupies a slot; if the actor's worker pool is
-	// momentarily saturated the resumer parks (retries) here rather than failing
-	// fast. A full lot sheds the request immediately so the router applies
-	// backpressure instead of queueing without bound.
-	release, ok := h.parking.enter(ctx)
-	if !ok {
-		return extproc.Result{}, parkingFullErr(actorRef.String())
-	}
-
+	// The resumer parks the request if the actor's worker pool is momentarily
```
bowei: Remove this comment. It's confusing in context, given that the implementation has moved inside ResumeActor.

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -72,9 +72,12 @@ type Handler struct {
 }
 
 func New(apiClient ateapipb.ControlClient, parkCfg ParkedRequestConfig, parkMetrics *ParkingMetrics) *Handler {
+	// The lot is shared: the resumer charges it at each caller's park
```
bowei: Remove this comment, it's talking about the old code, which no longer exists.

cmd/atenet/internal/router/ingress/errors.go
```diff
@@ -51,10 +51,16 @@ func statusDescription(err error) string {
 	return status.Convert(err).Message()
 }
 
+// errParkingLotFull is returned by ActorResumer.ResumeActor when a request
+// reached its park transition but the lot had no free slot. mapResumeError
+// turns it into the client-facing 503 "router at capacity" denial.
+var errParkingLotFull = errors.New("parking lot full")
```
bowei: Why do we need another level of error?

bowei: Is it possible to use the original parkingFullErr directly?

cmd/atenet/internal/router/ingress/parking.go
```diff
@@ -135,9 +135,12 @@ func DefaultParkedRequestConfig() ParkedRequestConfig {
 	}
 }
 
-// parkingLot is a bounded, non-blocking admission gate for resume-gated
-// requests. Each admitted request holds a slot for the duration of its resume
-// attempt; when the lot is full further requests are shed immediately so the
+// parkingLot is a bounded, non-blocking admission gate for parked requests.
```
bowei: I would keep the old comment.

cmd/atenet/internal/router/ingress/parking.go
```diff
@@ -154,12 +157,12 @@ func newParkingLot(cfg ParkedRequestConfig, m *ParkingMetrics) *parkingLot {
 	return &parkingLot{cfg: cfg, metrics: m}
 }
 
-// enter attempts to reserve a parking slot. On success it returns a release
-// func and ok=true; the caller MUST invoke release exactly once (passing the
-// request outcome, e.g. parkOutcomeServed) when the resume attempt completes.
-// ok=false means the lot is full and the request should be shed without
-// waiting. When parking is disabled every request is admitted and no slot
-// accounting or metrics are recorded.
+// enter reserves a parking slot for a caller whose resume flight just parked.
```
bowei: Did this change anything?

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -90,7 +90,19 @@ type resumeCallResult struct {
...
-	flight    singleflight.Group
+
+	// mu guards flights, the per-actor registry of in-flight resumes.
+	mu sync.Mutex
+	// flights deduplicates concurrent resumes per actor, singleflight-style:
+	// the first caller creates the flight and later callers attach to it. An
+	// entry is removed the moment its flight completes.
+	flights map[string]*resumeFlight
+
+	// lot bounds how many callers may wait on parked flights at once. A slot
```
bowei: Make this comment more concise.

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -162,118 +182,158 @@ func (r *ActorResumer) retryable(err error) bool {
 
 // ResumeActor ensures the requested actor is running. It deduplicates concurrent
 // requests within the process and, when parking is enabled, holds the request
-// while retrying transient failures until the budget elapses.
+// while retrying transient failures until the budget elapses. A caller occupies
```
bowei: We should probably just say 

// ResumeActor ensures the actor is running.
// This method will block until the actor is resumed or error.

The caller does not need to understand anything about the parking stuff.

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -162,118 +182,158 @@ func (r *ActorResumer) retryable(err error) bool {
...
-			// so the loop cannot end before its first attempt has completed —
-			// lastRetryErr is always the attempt's real answer here, and a
-			// definitive error (NotFound, ...) still passes through untouched.
-			if lastRetryErr != nil && (bgCtx.Err() != nil || wait.Interrupted(err)) {
-				return &resumeCallResult{leaderID: reqID, err: &budgetExhaustedError{lastErr: lastRetryErr}}, nil
-			}
-			return &resumeCallResult{leaderID: reqID, err: err}, nil
+// runFlight executes one shared resume for actorRef and publishes the outcome
+// to every caller attached to f. reqID identifies the caller that created the
+// flight (the cold-activation leader).
```
bowei: remove text (the cold-activation leader). Terminology is kind of confusing.

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -162,118 +182,158 @@ func (r *ActorResumer) retryable(err error) bool {
...
+func (r *ActorResumer) runFlight(f *resumeFlight, key string, actorRef resources.ActorRef, reqID uint64) {
+	// We detach the context from the first caller using a fixed background budget.
```
bowei: Simplify this comment.

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -162,118 +182,158 @@ func (r *ActorResumer) retryable(err error) bool {
...
+	// resume operation continues running for Caller 2 and Caller 3 without failing.
+	//
+	// The budget is therefore per-FLIGHT, not per-caller: its clock starts with
+	// the first caller, and later callers de-duplicated onto this flight share
+	// its remaining budget and outcome. A late joiner can see budget_exhausted
+	// after waiting far less than a full budget itself — the accepted cost of
+	// one control-plane RPC per hot actor (see docs/request-parking.md).
+	bgCtx, bgCancel := context.WithTimeout(context.Background(), r.budget)
+	defer bgCancel()
+	// The budget bounds the RETRY LOOP only — it never cancels an
```
bowei: Remove this comment.

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -162,118 +182,158 @@ func (r *ActorResumer) retryable(err error) bool {
...
+// flightResult classifies the retry loop's terminal state into the shared
+// result every caller attached to the flight receives. bgCtx is the flight's
+// budget context, consulted only for whether the budget expired.
+func flightResult(bgCtx context.Context, resumeResp *ateapipb.ResumeActorResponse, err, lastRetryErr error, reqID uint64) *resumeCallResult {
+	result := &resumeCallResult{leaderID: reqID}
+	switch {
+	case err == nil:
+		result.actor = resumeResp.GetActor()
+		result.resumed = resumeResp.GetResumed()
+	// If the budget elapsed while we were still blocked on a retryable
```
bowei: Simplify this comment...

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -162,118 +182,158 @@ func (r *ActorResumer) retryable(err error) bool {
...
+		result.err = &budgetExhaustedError{lastErr: lastRetryErr}
+	default:
+		result.err = err
+	}
+	return result
+}
+
+// awaitFlight waits for f's outcome on behalf of one caller. The wait is
+// two-phase: while the flight is resolving the caller holds nothing; once the
+// flight parks (or if it already has), the caller must hold a parking-lot slot
```
bowei: once the flight is retrying

omeryahud:
Thanks @bowei , I've fixed the comments, and added a clarifying comment to explain the additional select here https://github.com/agent-substrate/substrate/pull/1229/changes#diff-480555d3a5d237ae93d5718aadd2d6cdc2d4ad382fd12ce57b985f78beba0271R285

> Sorry for taking so long reviewing this. The logic looks ok, although I am thinking we may be able to simplify this by explicitly maintaining the state of the request in the flight map instead of relying on channel state.
> 
> I am ok with taking this PR in and doing the fix in a follow-up.

I'm looking into the alternative you below, and if I think it simplifies the code I'll open a follow-up PR

bowei (dismissed):
Thanks, this looks simpler

bowei (approved):

PR #1177, opened 2026-08-24
AGENTS.md: comment and commit message conventions & add CLAUDE.md pointer

BenTheElder:
Lots of contributors are using claude code, but claude does not read AGENTS.md by default.

I'd rather not add a lot of these, but this is the official documented solution:
https://code.claude.com/docs/en/memory#agents-md

Also, I've noticed both in my own usage and in a lot of other usage some defects:
- Excessively verbose code comments
- Failing linter due to british spellings
- spamming github due to repeated pushes with commits referencing PRs / issues in the commit instead of the PR body

Adding agents.md rules to address those, based on my local config.

NOTE: any existing sessions will need to reload memory to pick these up (restart claude / vscode, or `/compact`)

> It's a good idea to open an issue first for discussion.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

PR #1231, opened 2026-08-26
atelet: live refresh of projected trust bundles for running actors

### LINKED ISSUE #1266: [EPIC] Multi-Actor Worker Support
> **NOTE**: to read the design doc join the ate-dev mailinglist in the repo README.md

[Substrate: Multiple Actors Per Worker](https://docs.google.com/document/d/1yaagfSgh6Y7Qxq4gOG0GGh45yVk1VJsQv1Aa3W_O6RA/edit?tab=t.0) outlines an incremental approach to enabling multiple actors per worker.

Here is a markdown conversion of the current iteration of the doc, but it is missing comments on the doc
https://gist.github.com/BenTheElder/1fd8d7b456c66999594b791b2f96fc34

https://github.com/agent-substrate/substrate/issues/44 & https://github.com/agent-substrate/substrate/issues/793 are some early issues that previously asked about the concept.

----

At a very high level the plan is to make these changes incrementally. 

I have a working version but it's an unreasonably large diff at this stage of the project.

The doc outlines some of those incremental steps. 

I'm going to start filing some of the PRs that aren't strictly blocked on it today, and a **draft** of the API changes for review.

thompsonmax:
Part of #932 (PR 2 of 3). PR 1 (#941) added the `trustBundle` SystemInfo data source, resolved on the node at Run/Restore. This PR keeps those projections current while the actor runs.

## Live refresh

A `systemInfoVolumeRefresher` in atelet, modeled on kubelet's projected volumes: `collectData` is the one place that builds a volume's complete contents from its spec, `write` applies them, and every lifecycle point uses the pair. Run/Restore registers the actor's volumes, which writes them fail-closed before the sandbox boots; ClusterTrustBundle events rewrite them while it runs; Checkpoint, Terminate, and failed starts deregister. No API, proto, or RBAC changes: the wire still carries only `{name, path}`.

- Informer events only enqueue bundle names; a single run loop writes. Failed writes requeue with backoff through a rate-limited workqueue; resolution failures keep last-good contents and wait for the bundle's next event. Resync (24h) is only a guard against missed watch events.
- Change detection hashes the raw backing contents, not the projected output (sanitization shuffles). Files are replaced by temp-and-rename at stable paths and byte-identical files are left untouched: a needless rewrite replaces an inode that suspended guests re-bind on resume.
- Nothing is persisted. Registrations are in memory; after an atelet restart, a running actor's files keep their last-written contents until its next Run/Restore rewrites them from current cluster state.

## virtiofsd: `--migration-on-error=guest-error`

CI confirmed the hazard (3 of 3 micro-VM runs): a rotation renames over a file whose inode the guest still references (a dcache reference is enough, no held fd), the actor suspends before re-reading, and find-paths serializes an inode with no findable path. Under the default `abort`, the destination virtiofsd rejects the device state at vm.restore: suspend succeeds, every restore fails, and the actor is permanently stuck. Full analysis in the PR comments.

With `guest-error` the restore succeeds and only the stale reference is faulty: EIO on access until a fresh lookup at the stable path heals it. That matches the documented reader contract (the platform rewrites the file, applications re-read it) and also covers guest-created `O_TMPFILE`s held across suspend, which already fail restores under `abort`. The flag is backend-process configuration, so existing snapshots (template goldens included) restore unchanged. Trade-off: a share-reconstruction bug now surfaces as post-resume EIO and a readyz failure instead of a loud restore failure; virtiofsd names the faulty inodes in the worker pod log.

## E2E

The identity suite covers live refresh on both sandbox classes: rotate the pool under two running actors and wait for both to observe the new bundle live; rotate/suspend/resume without waiting for propagation, so the resume must deliver rotated contents whether the rewrite landed before or after the guest went down (the exact `abort` repro); assert a sibling that never cycled also got the rotation live. The probe fixture now keeps its namespace when a test fails so the worker pod logs survive, and the cloud-hypervisor client includes the vm.restore response body in errors.

## Not in this PR

Auto-injected egress trust volume (#932 PR 3) and the configurable backend registry (tracked in #932).


cmd/ateom-microvm/internal/kata/overlay_linux.go
```diff
@@ -129,6 +129,18 @@ func virtiofsdArgs(o VirtiofsdOptions) []string {
...
+		// reference inodes with no findable path — a system-info file
+		// live-rewritten under the guest (trust-bundle rotation), or a
+		// guest-unlinked temp file held open across the suspend — and abort
+		// would make such snapshots permanently unrestorable. Under
+		// guest-error the stale reference degrades to EIO on access, and a
+		// fresh lookup at the stable path heals it. The mode is daemon-wide:
+		// a path genuinely missing elsewhere in the share (a rootfs or
+		// durable-dir reconstruction gap) also surfaces as guest EIO instead
+		// of failing vm.restore; virtiofsd's log names the inodes it could
+		// not re-establish.
```
BenTheElder: nit: can you refresh your session(e.g. do a developer reload in vscode or claude --resume or ...) and ask your agent to apply the current AGENTS.md rules? I'm pretty sure it should write a terser comment following https://github.com/agent-substrate/substrate/pull/1177 

(also, do you have any idea if that commit was in the branch when you started? if it's not taking effect we need to iterate)

some of these agents have been defaulting to excessively verbose code comments, we don't really need 11 lines checked in to justify this flag.

thompsonmax: Hmm, seems the `AGENTS.md` was in fact loaded in my session. It's possible it was pushed too far back in the context window when this huge comment was generated. Or perhaps we need to update `AGENTS.md` to be a bit stricter with its rules around comments.

I also have this in my CLAUDE.md (but maybe it's the same issue of it got pushed too deep in the ctx window):

```
## Code comments
Write sparse comments. A comment must say something the code cannot: a
constraint, an invariant, a non-obvious why. Default to 1–3 lines; if a
comment needs a second paragraph, the content probably belongs in a doc
comment, an issue, or the commit message. Never write comments that:
- restate what the adjacent code does
- justify the change to a reviewer ("per the review", "as discussed in the
  PR") — that context dies when the PR merges
- explain the same fact in more than one place; pick the site a reader hits
  first and cross-reference from the others
- give background essays for hypothetical questions nobody has asked
Match the comment density and tone of the surrounding file. When editing
existing code, do not add comments narrating the edit.
```

At least for this PR, I refreshed that guidance in my session, which seems to have trimmed down most of the unnecessarily long comments.

BenTheElder: Interesting, thank you.

[Also explicit +1 for the flag]

haiyanmeng (commented):
LGTM

cmd/atelet/main.go
```diff
@@ -1602,47 +1642,58 @@ func (s *AteomHerder) prepareOCIBundles(
...
 //
-// TODO(#802): rotating data sources (identity JWTs, certificates) will need
-// these files refreshed while the actor runs, not just at Run/Restore — and
-// must keep the per-file rename discipline so visible paths never move.
-// actorMetadata never changes after start, so writing here is enough for it.
+// Every file must be a plain file at a stable real path across
+// regenerations: restore re-binds suspend-time guest state to files by
+// recorded path (virtiofsd find-paths, gVisor's gofer), so a scheme that
+// relocates files on write (kubelet's atomic writer) would leave that state
+// dangling. Per-file write-to-temp-and-rename is atomic enough — a reader
```
ahmedtd: Is this still true with the guest-error flag change?

Atomicwriter lets kubelet simplify the refresh logic somewhat --- first Kubelet builds a model of all the content that should be in the projected volume and then atomically applies it.

cmd/atelet/trustbundle.go
ahmedtd, line 39: Technically speaking, the way you are supposed to use a signer-linked cluster trust bundle is by using a combination of a selector on the signer name and a label selector.  This is something to keep in mind for the future, since it will let us canary changes to a fraction of workloads.

If everyone is just reading one direct ClusterTrustBundle, then there's no way to trial adding a new root for n% of workloads.

cmd/atelet/trustbundle.go
```diff
@@ -47,28 +49,63 @@ func supportedTrustBundleNames() string {
...
+// rawTrustBundle returns the unsanitized contents of the named trust
+// bundle's backing ClusterTrustBundle, resolving the name against the
+// allowlist and reading through atelet's informer-backed lister.
+func rawTrustBundle(lister certlisters.ClusterTrustBundleLister, name string) (objectName, raw string, err error) {
 	objectName, supported := supportedTrustBundles[name]
 	if !supported {
-		return nil, fmt.Errorf("trust bundle %q is not supported by this deployment (supported: %s)", name, supportedTrustBundleNames())
+		return "", "", fmt.Errorf("trust bundle %q is not supported by this deployment (supported: %s)", name, supportedTrustBundleNames())
 	}
 	if lister == nil {
```
ahmedtd: Is this something that can actually happen in practice?  If not, it would be better to just panic on the nil access.

cmd/atelet/trustbundle.go
```diff
@@ -47,28 +49,63 @@ func supportedTrustBundleNames() string {
...
 	} else if err != nil {
-		return nil, fmt.Errorf("trust bundle %q: while reading ClusterTrustBundle %q: %w", name, objectName, err)
+		return "", "", fmt.Errorf("trust bundle %q: while reading ClusterTrustBundle %q: %w", name, objectName, err)
 	}
-	pemBundle, err := pemutil.SanitizeCertificateBundle([]byte(bundle.Spec.TrustBundle))
+	return objectName, bundle.Spec.TrustBundle, nil
+}
+
+// sanitizeTrustBundle sanitizes raw backing contents into the projected PEM.
+func sanitizeTrustBundle(name, objectName, raw string) ([]byte, error) {
```
ahmedtd: I don't think this function adds anything over pemutil.SanitizeCertificateBundle?

cmd/atelet/trustbundlerefresh.go
```diff
@@ -0,0 +1,330 @@
...
+	"k8s.io/client-go/tools/cache"
+)
+
+// trustBundleProjectionsFileName is the per-actor record recoverFromDisk
+// rebuilds the registry from after an atelet restart. It sits beside (not
+// inside) the volume roots under SystemInfoVolumeRootsDir, sharing their
+// lifecycle: wiped by resetActorDirs, never snapshotted. The micro-VM class
+// serves the whole roots dir into the guest read-only, so nothing sensitive
+// belongs here; the dot cannot collide with a volume root (DNS-label names).
+const trustBundleProjectionsFileName = "trust-bundle-projections.json"
```
ahmedtd: Let's not persist anything.  We tried something like this in gke-spiffe-node-agent and it was a huge pain in the ass.  It's OK if an atelet restart changes file contents.

cmd/atelet/trustbundlerefresh.go
```diff
@@ -0,0 +1,330 @@
...
+
+// trustBundleResyncPeriod is the ClusterTrustBundle informer's resync, and
+// with it the refresher's retry cadence: a failed rewrite gets no further
+// event until the bundle changes again, so the periodic replay re-drives it
+// (a no-op otherwise, per the AppliedHash compare).
+const trustBundleResyncPeriod = 10 * time.Minute
```
ahmedtd: Resync should be probably be 0 (disabling resync) or 24 hours.  Every resync period, the client will do a complete list of all ClusterTrustBundles --- it could drive significant traffic if we set it to 10 minutes.

It's not an appropriate mechanism to redrive failed actions on our side; it is originally intended to be a guard against bugs in the k8s informer libraries.

If we need retries for actions that can fail on our end, we need to use a workqueue.  The podcert signer should have an example.

cmd/atelet/trustbundlerefresh.go
```diff
@@ -0,0 +1,330 @@
...
+
+// trustBundleProjection is one projected trustBundle file of one actor.
+type trustBundleProjection struct {
```
ahmedtd: This will probably be more work, but I think it will be important for future flexibility here:  instead of tracking specific files, the refresh logic should re-interpret the volume and volume mount fields of the actor spec.

Right now, you've got the logic ping-ponging between writeSystemInfoVolume, trustBundleRefresher, and writeSystemInfoFile.

Instead, we should have a systemInfoVolumeRefresher with a write() function that knows how to write the complete contents of a system info.  The write() function should be called once at actor startup, and then re-called from an timer or informer hook.

cmd/atelet/trustbundlerefresh.go
```diff
@@ -0,0 +1,330 @@
...
+		objectName, raw, err := rawTrustBundle(r.lister, p.Bundle)
+		if err != nil {
+			slog.WarnContext(ctx, "Trust bundle unreadable at registration; keeping the projection as written", slog.String("actor_uid", actorUID), slog.String("bundle", p.Bundle), slog.Any("err", err))
+			continue
+		}
+		h := trustBundleHash(raw)
+		if p.AppliedHash == h {
+			continue
+		}
+		pemBundle, err := sanitizeTrustBundle(p.Bundle, objectName, raw)
```
ahmedtd: Here's a litmus test for if this system is factored correctly --- there should only be one call to sanitizeTrustBundle in the source code.

cmd/atelet/main.go
thompsonmax: Yeah you're right, we don't have this restriction any more. Since we don't need multi-file atomic updates yet though, I'd prefer to keep the simple temp + rename updates to keep the PR simple, but feel free to push back if there's something AtomicWriter gives us that would be helpful now.

Left a TODO to switch to AtomicWriter once we need multi-file atomic updates.

cmd/atelet/trustbundle.go
thompsonmax: Good point, removed the check

cmd/atelet/trustbundle.go
thompsonmax: Deleted, collectData calls pemutil.SanitizeCertificateBundle directly now

cmd/atelet/trustbundlerefresh.go
thompsonmax: Hmm, I think we still need some way to track the projected volumes on each actor so that we can reload which actors need live refresh if the kubelet restarts.

I believe how this works for k8s projected volumes is: when the kubelet restarts, it just fetches the specs of the pods scheduled on the node from kube-apiserver, then repopulates those volumes.

For our case, perhaps we can just have each atelet, when restarted, read the specs of currently scheduled actors attached to it to figure out what SystemInfo CTBs volumes are present, then re-hook up the systemInfoVolumes to be refreshed by the informer.

Let me know if you have other thoughts. I think we can leave how to persist this and recover during atelet restarts out of this PR and include it in a follow up.

cmd/atelet/trustbundle.go
thompsonmax: Ack, added a TODO for handling selection in that way, once we add more general support for CTB volumes.

cmd/atelet/trustbundlerefresh.go
thompsonmax: Done, I switched re-sync to 24 hours and added a workqueue mechanism to handle transient node-side write failures.

cmd/atelet/trustbundlerefresh.go
thompsonmax: Done, I roughly followed how [kubelet implements projected volumes](https://github.com/kubernetes/kubernetes/blob/4faad4bd2bfb3d303e0d8ff523857f71974bcb39/pkg/volume/projected/projected.go#L249
) and added a consolidated systemInfoVolumeRefresher.

cmd/atelet/trustbundlerefresh.go
thompsonmax: Done, now only called in collectData in systemInfoVolumeRefresher.

cmd/atelet/trustbundlerefresh.go
ahmedtd: I guess this is a larger decision --- should atelet reacquire state by requesting it from ate-api-server, or persist it locally.

@BenTheElder WDYT?

cmd/atelet/main.go
```diff
@@ -284,7 +283,7 @@ func main() {
 	// only the egress trust bundle). The v1beta1 API is feature-gated: on a
 	// cluster that does not serve it, startup blocks at WaitForCacheSync
 	// below, with the reflector's errors naming the missing API.
-	coreFactory := informers.NewSharedInformerFactoryWithOptions(k8sClient, 0,
+	coreFactory := informers.NewSharedInformerFactoryWithOptions(k8sClient, trustBundleResyncPeriod,
```
ahmedtd: Nit: ClusterTrustBundles are not part of the core API group.

You should name this informer factor something specific to ClusterTrustBundles, because the field selector makes it unsuitable for use with any other API types.

cmd/atelet/main.go
ahmedtd, line 288: We should construct the systemInfoVolume refresher in main, and pass it as an argument to NewService (dependency injection hygiene)

cmd/atelet/main.go
```diff
@@ -308,6 +307,14 @@ func main() {
 		csiDriverConfigLister,
 		clusterTrustBundleLister,
 	)
+
+	// Live-refresh projected trust bundles: bundle events enqueue, and the
+	// run loop rewrites the files of registered running actors.
+	if _, err := coreFactory.Certificates().V1beta1().ClusterTrustBundles().Informer().AddEventHandler(wmService.systemInfoVolumes.eventHandler()); err != nil {
```
ahmedtd: Conventionally, this registration should go in newSystemInfoVolumeRefresher

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,305 @@
...
+	certsv1beta1 "k8s.io/api/certificates/v1beta1"
+	certlisters "k8s.io/client-go/listers/certificates/v1beta1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+)
+
+// trustBundleResyncPeriod is the ClusterTrustBundle informer's resync: a
+// guard against missed watch events, not a retry mechanism — failed writes
+// redrive through the refresher's workqueue.
+const trustBundleResyncPeriod = 24 * time.Hour
```
ahmedtd: This doesn't need to be a constant --- you can just hardcode 24 hours in main.go.  Even if it is a constant, it belongs in main.go because it's only used there.

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,305 @@
...
+	Root string
+	Spec *ateletpb.SystemInfoVolume
+
+	// appliedHashes maps each projected bundle name to the trustBundleHash
+	// last written into this volume. In memory only: after an atelet restart
+	// the next Run/Restore rewrites the volume from current cluster state.
+	appliedHashes map[string]string
+}
+
+type registeredActor struct {
```
ahmedtd, line 67: @BenTheElder It seems like atelet probably needs a central component that holds the protobuf specs for each Actor it is currently running?

cmd/atelet/trustbundlerefresh.go
BenTheElder: Atelet isn't even tracking actors at all currently, it's all on-demand. It does periodically identify available ateoms (!= actors).

I don't think it should persist state for this. I also don't know that I want it doing a live read from the API of all the actors for each of the workers (keep in mind that's moving from a singleton O(soon)).

Would we be OK with querying the ateoms on startup (so it's all local?)

cmd/atelet/systeminfovolume.go
ahmedtd: @thompsonmax Maybe let's open an umbrella issue for "atelet state persistence" and leave a TODO referencing it here.  I don't think the optimal strategy is for each subsystem to hold onto its own bit of state, but I also don't think you need to solve it in this PR.

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,305 @@
...
+//
+// TODO(#802): adopt kubelet's whole-volume AtomicWriter swap if a source
+// ever needs multi-file atomicity.
+type systemInfoVolumeRefresher struct {
+	lister certlisters.ClusterTrustBundleLister
+
+	// queue carries bundle names from informer events to the run loop.
+	queue workqueue.TypedRateLimitingInterface[string]
+
+	// mu serializes registry mutations AND projection writes: Deregister
```
ahmedtd: Just "mu covers actors".

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+}
+
+// Register records actorUID's system-info volumes and writes their complete
+// contents from current cluster state, replacing any prior registration.
+// Fail-closed: an actor that declared a projection must not start without it.
+func (r *systemInfoVolumeRefresher) Register(actorUID string, ref resources.ActorRef, volumes []systemInfoVolume) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if len(volumes) == 0 {
```
ahmedtd: I am suspicious of this.

a) If we think our state tracking has gotten corrupted, we should fail loudly (panic) rather than paper over it.  So we shouldn't be `delete`-ing any entry in Register.

b) Probably it makes sense to track actors even if they don't have any systemInfoVolumes, just to lower the number of edge cases we have to reason about.

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,305 @@
...
+		delete(r.actors, actorUID)
+		return nil
+	}
+	actor := &registeredActor{ref: ref, volumes: volumes}
+	for i := range actor.volumes {
+		v := &actor.volumes[i]
```
ahmedtd: Either registeredActor.volumes should be `[]*systemInfoVolume` (if we don't want copies), or write should take `[]systemInfoVolume` (if we don't care about copies)

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,305 @@
...
+	return payload, bundleHashes, nil
+}
+
+// write makes the volume's on-disk contents match collectData: files are
+// replaced via temp-and-rename and unchanged ones left alone, because
+// restores re-bind suspend-time guest state by real path and inode.
+func (r *systemInfoVolumeRefresher) write(ref resources.ActorRef, actorUID string, v *systemInfoVolume) error {
+	payload, bundleHashes, err := r.collectData(ref, actorUID, v.Spec)
+	if err != nil {
+		return err
```
ahmedtd: Wrap errors, we typically shouldn't return a bare error

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,305 @@
...
+func writeSystemInfoFile(rootPath, relPath string, data []byte) error {
+	if relPath == "" || strings.HasPrefix(relPath, "/") {
+		return fmt.Errorf("invalid system-info path %q: must be a non-empty relative path", relPath)
+	}
+	for _, seg := range strings.Split(relPath, "/") {
+		if seg == ".." || seg == "." || seg == "" {
+			return fmt.Errorf("invalid system-info path %q: must not contain empty, '.', or '..' segments", relPath)
+		}
+	}
+	dst := filepath.Join(rootPath, filepath.FromSlash(relPath))
```
ahmedtd: We should probably use os.Root (https://pkg.go.dev/os#Root) to be sure that these operations are constrained within the volume root.

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// run drains the refresh queue until ctx ends, requeueing failed refreshes
+// with backoff.
+func (r *systemInfoVolumeRefresher) run(ctx context.Context) {
+	go func() {
+		<-ctx.Done()
+		r.queue.ShutDown()
+	}()
+	for {
```
ahmedtd: Traditionally, there's a lot more ceremony in this loop: https://github.com/agent-substrate/substrate/blob/main/cmd/podcertcontroller/internal/signercontroller/signercontroller.go#L104

Some of it goes towards intercepting panics, some of it might be useless.  I would investigate closely before choosing a different form.

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,305 @@
...
+
+// refreshBundle rewrites every registered volume that projects bundleName
+// and has not applied its current contents, keeping last-good files on any
+// failure: resolution errors wait for the next event, write errors requeue.
+func (r *systemInfoVolumeRefresher) refreshBundle(ctx context.Context, bundleName string) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	_, raw, err := rawTrustBundle(r.lister, bundleName)
+	if err != nil {
+		if r.projectsLocked(bundleName) {
```
ahmedtd: Why would we be trying to fetch the bundle if one of the pods weren't using it?

internal/e2e/fixtures/probe/probe.yaml.tmpl
```diff
@@ -21,6 +21,10 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: ate-e2e-probe${FIXTURE_SUFFIX}
+  labels:
+    # e2e.NamespaceLabel: a fixture retained by a failed test stays visible
```
ahmedtd: This comment doesn't really explain what it's doing.

cmd/atelet/systeminfovolume.go
BenTheElder: Should atelet be the one doing this whole thing at all, or ateoms?

Yes, today there's one ateom per active actor, but we will get away from that O(soon) #1266, and it feels awkward to make atelet stateful about the set of actors. Atelets do not own actors.

I guess the argument is caching?

BenTheElder: It feels like this needs to be more tightly coupled with the sandbox implementations so atelet is the wrong layer.

Imagine if we use a block-only sandbox in the future (e.g. firecracker, no virtiofs support), we could conceivably coordinate writing through via the guest agent, but it would be a layering violation for atelet to know how to do thsi.

ahmedtd: I mean, atelet handles all filesystem composition today, and some of the SystemInfoVolume data sources will need to make outbound calls to ate-api-server (JWTs and certs, at least).

However, I think at this point you're probably the person who is best positioned to make calls about the division of functionality between atelet and ateom.  I also think we need someone (probably you) to decide what the on-node source of truth for state is.  With Kubelet, it's the pod informer cache.  It seems like we need something similar in atelet/ateom, or we need to have a local DB where we record the Actor specs.

internal/e2e/fixtures/probe/probe.yaml.tmpl
thompsonmax: updated to explain its used to mark the resource for deletion at the end of e2e tests

cmd/atelet/systeminfovolume.go
thompsonmax: done

cmd/atelet/systeminfovolume.go
thompsonmax: done

cmd/atelet/systeminfovolume.go
thompsonmax: done

cmd/atelet/main.go
thompsonmax: done

cmd/atelet/main.go
thompsonmax: done

cmd/atelet/main.go
thompsonmax: Renamed to simply "clusterTrustBundleInformerFactory"

cmd/atelet/systeminfovolume.go
thompsonmax: done

cmd/atelet/systeminfovolume.go
thompsonmax: The idea here is to stop tracking an actor which no longer defined systemInfoVolumes. But yeah I agree that just tracking every actor is likely to be less error-prone, so switched to that instead.

cmd/atelet/systeminfovolume.go
thompsonmax: Switched to  `[]*systemInfoVolume` throughout (registeredActor, Register, the call site)

cmd/atelet/systeminfovolume.go
thompsonmax: Good point, I updated this to not resolve the bundle if no actor is using it.

cmd/atelet/systeminfovolume.go
thompsonmax: Created https://github.com/agent-substrate/substrate/issues/1372 to track the broader issue of where to persist this state and left a TODO on it here for now.

cmd/atelet/systeminfovolume.go
thompsonmax: Sure, I copied most of the structure there (intercepting panics seems useful). I kept out a few things that didn't seem useful here, e.g. having multiple consumer workers since for now everything locks on the mutex inside `systemInfoVolumeRefresher`.

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,361 @@
...
+	r.queue.Forget(bundleName)
+	return true
+}
+
+// refreshBundle rewrites every registered volume that projects bundleName
+// and has not applied its current contents, keeping last-good files on any
+// failure: resolution errors wait for the next event, write errors requeue.
+// A bundle no registered actor projects is not even resolved.
+func (r *systemInfoVolumeRefresher) refreshBundle(ctx context.Context, bundleName string) error {
+	r.mu.Lock()
```
BenTheElder: I'm a bit concerned about this ... we're trying to make it cheap to spawn a lot of actors on the same node, concurrently ...

This code looks like it presents a pretty serious new scaling limitation, can we avoid holding a shared lock while walking every actor?

cmd/atelet/systeminfovolume.go
BenTheElder: > [...] and some of the SystemInfoVolume data sources will need to make outbound calls to ate-api-server (JWTs and certs, at least).

Slightly a problem for current ateoms, but longterm we'll be putting all the actor traffic on the ingress/egress tunnel, I think we don't really support direct ingress traffic anymore so it might be ok to do this already. cc @bowei I'm not sure we still need to preserve "actor gets the full pod network" and we definitely can't in multi-actor ateom anyhow.

> However, I think at this point you're probably the person who is best positioned to make calls about the division of functionality between atelet and ateom. I also think we need someone (probably you) to decide what the on-node source of truth for state is. [...]


ACK. I think it will be the ateoms, which may at some point absorb more of atelet. Even in main, ateom is the one that actually knows what actors are currently running.

> With Kubelet, it's the pod informer cache. It seems like we need something similar in atelet/ateom, or we need to have a local DB where we record the Actor specs.

Do we need the whole actor spec, or just context on the live-updated volumes?

Most of what an actor does is just on resume / cold boot, during suspend we generically handle actors.

I don't think we want a node-local database (image cache aside).

ahmedtd: > Do we need the whole actor spec, or just context on the live-updated volumes?

Saving the whole actor spec (or at least, the actor spec as passed in the atelet boot/resume call) seems like a simple way to keep us flexible for the future.  

> I don't think we want a node-local database

We have one, even if it's just in-memory in the ateom.  The state lives somewhere.  We need to think about it holistically.

Edit: I think "ateom holds all the actor state in-memory, and ateom is not allowed to crash while an actor is running" is a valid position", but we need to write it down wherever we write down project architecture decisions.

cmd/atelet/systeminfovolume.go
ahmedtd: Should we move the volume writing into ateom?

thompsonmax: Hmm, moving the volume writing into ateom seems like a big shift (although perhaps is the ideal end state).

Here's an alternative to unblock this without moving things into ateom yet -- adopt a dual locking strategy with two locks: a registry lock for the actors map, and a per-actor lock for that actor's file writes, appliedHashes, and a stale flag. The registry lock is never held across I/O, only for in-memory map ops.

```
Two lock domains
════════════════

  ┌──────────────────────────────┐      ┌────────────────────────────────┐
  │ registry lock  r.mu          │      │ per-actor lock  actor.mu       │
  │ covers: actors map only      │      │ covers: volume file writes,    │
  │ (insert/delete/lookup/snap)  │      │ appliedHashes, stale flag      │
  │ never held across I/O        │      │ held across fsync writes       │
  └──────────────────────────────┘      └────────────────────────────────┘
        one per refresher                     one per registered actor


Register(uid, spec)                              locks held
───────────────────────────────────────────────────────────
 build new entry                                 —
 lock new.mu          (pre-publication)          new
 r.mu { old := actors[uid]; actors[uid] = new }  new + r.mu   (µs)
 if old:  lock old.mu ──── drains in-flight ──┐  new + old
          old.stale = true; unlock old.mu   ◄─┘  new
 write volumes (fsync), set appliedHashes        new          ← parallel
 unlock new.mu                                   —              across actors


refreshBundle(name)                              locks held
───────────────────────────────────────────────────────────
 r.mu { snapshot actors projecting name }        r.mu         (µs)
 resolve bundle via lister, hash it              —
 for each target a:
     lock a.mu
     a.stale? ── yes ──► skip (removed/replaced)
        │ no
     write changed volumes, update hashes        a only       ← blocks just
     unlock a.mu                                                this actor


Deregister(uid)                                  locks held
───────────────────────────────────────────────────────────
 r.mu { a := actors[uid]; delete }               r.mu         (µs)
 lock a.mu ─── waits out any write to a ───┐     a
 a.stale = true; unlock a.mu             ◄─┘     —
 return  ⇒ no future write can touch a's dirs;
           Checkpoint/Terminate may wipe them
```

Spawns stop serializing behind other actors' fsyncs, and a rotation blocks only the one actor being rewritten.

thompsonmax: Implemented this 2 level locking, which should prevent holding a shared lock while we walk every actor.

cmd/atelet/trustbundlerefresh.go
thompsonmax: Will explore moving volume injection into ateom in #1372. I'd like to still get the PR in which keeps this in atelet in the short term.

cmd/atelet/systeminfovolume.go
thompsonmax: Will explore moving this state tracking into ateom in #1372.

cmd/atelet/systeminfovolume.go
thompsonmax: Long-term, exploring moving this state tracking into individual ateoms in #1372.

cmd/atelet/main.go
```diff
@@ -266,22 +264,25 @@ func main() {
 	ateFactory := externalversions.NewSharedInformerFactory(ateClient, 0)
 	csiDriverConfigLister := ateFactory.Api().V1alpha1().CSIDriverConfigs().Lister()
 
-	// Start an informer on the ClusterTrustBundle we care about (currently
-	// only the egress trust bundle). The v1beta1 API is feature-gated: on a
-	// cluster that does not serve it, startup blocks at WaitForCacheSync
-	// below, with the reflector's errors naming the missing API.
-	coreFactory := informers.NewSharedInformerFactoryWithOptions(k8sClient, 0,
+	// Watch only the ClusterTrustBundle we care about (currently the egress
```
ahmedtd: We don't really need this comment.

cmd/atelet/main.go
```diff
@@ -1105,6 +1123,15 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 	// (uncached assets + image, ~2.5s unpack) that overlap is large.
 	// TODO(dberkov): the old pause checkpoint files are not deleted after they are
 	// copied to checkpointDir for the LOCAL case.
+
+	// prepareOCIBundles registers system-info volumes for live refresh; leave
+	// no registration behind for a restore that fails.
+	defer func() {
+		if err != nil {
+			s.systemInfoVolumes.Deregister(actorUID)
```
ahmedtd: If possible, can we extract register() from prepareOCIBundles, so that we can see the matched register() / deregister() pair of calls?  That should lower the cognitive burden on someone reading this code.

cmd/atelet/main.go
```diff
@@ -1558,12 +1592,19 @@ func (s *AteomHerder) prepareOCIBundles(
...
-				return fmt.Errorf("while populating system-info volume %q: %w", vol.GetName(), err)
-			}
+			siVolumes = append(siVolumes, &systemInfoVolume{
+				Name: vol.GetName(),
+				Root: ateompath.SystemInfoVolumeRoot(actorUID, vol.GetName()),
+				Spec: volSrc.SystemInfo,
+			})
 		}
 	}
+	// Registering before boot makes the files carry the values of the actor
```
ahmedtd: This comment seems irrelevant, at least at this point in the code.

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,395 @@
...
+	if informer != nil {
+		informer.AddEventHandler(r.eventHandler())
+		r.hasSynced = informer.HasSynced
+	}
+	return r
+}
+
+// Register records actorUID's system-info volumes — possibly none, every
+// actor is tracked — and writes their complete contents from current cluster
+// state, replacing any prior registration. Fail-closed: an actor that
```
ahmedtd: Can you take a look through the whole PR and scrub Claude-isms like the "Fail-closed" sentence here?  

SystemInfoVolumeRefresher doesn't and can't enforce that property.

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,395 @@
...
+	// blocks until the initial population below finishes.
+	actor.mu.Lock()
+	defer actor.mu.Unlock()
+
+	r.mu.Lock()
+	old := r.actors[actorUID]
+	r.actors[actorUID] = actor
+	r.mu.Unlock()
+
+	if old != nil {
```
ahmedtd: I don't think this should be happening unless several system invariants have been violated.  It's probably better to panic if we see an actor being double-registered, at least for now.  That makes the problem nice and loud.

Once we get the system more stable and battle-tested, we can think about removing the panics for extra resilience.

cmd/atelet/systeminfovolume.go
```diff
@@ -0,0 +1,395 @@
...
+		if err := r.write(ref, actorUID, v); err != nil {
+			actor.stale = true // half-populated; the caller deregisters on failure
+			return fmt.Errorf("while populating system-info volume %q: %w", v.Name, err)
+		}
+	}
+	return nil
+}
+
+// Deregister drops actorUID's registration once the sandbox is down (or
+// never came up). It returns only after any in-flight write to the actor's
```
ahmedtd: This comment does not seem correct.

After drawing some diagrams of different racing Deregister() and Refresh() calls, I think the invariant the code enforces is, "After Deregister(A) returns successfully, there will be no further writes to the actor's volumes".  It's not actually waiting for any in-flight writes to succeed, but instead setting a flag to cancel any future pending writes.

cmd/atelet/systeminfovolume.go
thompsonmax: done

cmd/atelet/systeminfovolume.go
thompsonmax: Makes sense, switched to panic if we double register.

cmd/atelet/systeminfovolume.go
thompsonmax: I think the invariant is true, but I don't think it's clear about how it's achieved:

1. It takes the actor lock, which drains any in-flight writes for that actor
2. Setting stale prevents any further writes which might occur.

But yeah, I think it's simpler to just say "After Deregister returns, no more system-info volumes will be written for the actor.". Updated the comment.

cmd/atelet/main.go
thompsonmax: Removed the comment

cmd/atelet/main.go
thompsonmax: done

cmd/atelet/main.go
thompsonmax: Sure, made it so `Run` and `Restore` call Register/Deregister directly.

ahmedtd (approved):

PR #1233, opened 2026-08-26
feat(authz): define 3-tier atespace authorization model and node restriction

SHRUTI6991:
Working on [#1563](https://github.com/agent-substrate/substrate/issues/1563)

## Overview
This PR defines the initial OpenFGA authorization model (`internal/authz/model.fga`) and comprehensive test suite (`internal/authz/model_test.fga.yaml`) for Substrate control-plane and data-plane operations.

## Key Features & Design

1. **Global Cluster Scope (Singleton)**:
   - Modeled as a singleton (`cluster:root`).
   - `owner`: Can create Atespaces (`can_create_atespace`), and automatically inherits `owner` across all child Atespaces via `owner from parent_cluster`.
   - `viewer`: Read-only access (`can_get`), allowing cluster-level inspection/listing, and automatically inherits `viewer` across all child Atespaces via `viewer from parent_cluster`.

2. **3-Tier Atespace Governance**:
   - `owner`: Full administrative control; can set policy (`can_set_policy`), get policy (`can_get_policy`), create templates/actors, update, and delete the Atespace (`can_delete`).
   - `editor`: Can create templates/actors (`can_create_actor`, `can_create_actor_template`), get/update resources and the Atespace (`can_update`, `can_get`).
   - `viewer`: Read-only access (`can_get`, `can_get_policy`). All subjects require explicit authentication (no wildcard `user:*`).

3. **Direct Binding Boundaries**:
   - User role bindings stop at the Atespace boundary.
   - Child `actor` and `actor_template` instances derive their access directly from `parent_atespace`.

4. **Node Restriction for Credential Minting**:
   - Links the physical topology graph (`scheduled_worker: [worker]`, `host_node: host_node from scheduled_worker`).
   - `can_mint_ateom_actor_credential` is restricted exclusively to the active `host_node` running the scheduled worker pod.

5. **Lifecycle Permissions**:
   - `can_resume`: Ingress routers or Atespace editors can resume actors.
   - `can_suspend`: Actors' workload identities or Atespace editors can trigger suspend.

6. **Naming Conventions**:
   - Roles and graph edges are nouns (`owner`, `viewer`, `parent_cluster`, `host_node`).
   - Permission checks evaluated at runtime consistently use the `can_` prefix (`can_get`, `can_set_policy`, `can_create_actor`, `can_mint_ateom_actor_credential`).

## Verification

- **OpenFGA Model Tests**: All 5 test suites (101/101 checks) pass:
  ```bash
  docker run --rm -v $(pwd)/internal/authz:/workspace -w /workspace openfga/cli:latest model test --tests model_test.fga.yaml
  # Tests 5/5 passing, Checks 101/101 passing


internal/authz/model.fga
```diff
@@ -0,0 +1,87 @@
...
+
+# Physical host node running atelet
+type node
+
+# Worker pod running ateom
+type worker
+  relations
+    define host_node: [node]
+
+# Global Cluster Scope
```
ahmedtd: Nit: mention that this is a singleton, there will only ever be one instance.

internal/authz/model.fga
```diff
@@ -0,0 +1,87 @@
...
+type cluster
+  relations
+    define owner: [user]
```
ahmedtd, line 36: Do we want viewer/editor roles here?  Viewer, at least, should be well-defined.

internal/authz/model.fga
```diff
@@ -0,0 +1,87 @@
...
+    define parent_cluster: [cluster]
+
+    # --- 1. Aligned 3-Tier Roles ---
+    # Cluster owner automatically inherits owner without explicit per-atespace tuples
+    define owner: [user] or owner from parent_cluster
+    define editor: [user] or owner
+    define viewer: [user, user:*] or editor
+
+    # --- 2. Policy Governance (SetPolicy / GetPolicy) ---
+    define can_set_policy: owner
```
ahmedtd, line 53: I like the convention of starting all the "permission check" relations with "can_" --- can you put a comment at the top stating this?  Just thinking about future evolution of file.

internal/authz/model.fga
```diff
@@ -0,0 +1,87 @@
...
+
+type atespace
+  relations
+    define parent_cluster: [cluster]
+
+    # --- 1. Aligned 3-Tier Roles ---
+    # Cluster owner automatically inherits owner without explicit per-atespace tuples
+    define owner: [user] or owner from parent_cluster
+    define editor: [user] or owner
+    define viewer: [user, user:*] or editor
```
ahmedtd: What does the `user:*` syntax indicate?

internal/authz/model.fga
```diff
@@ -0,0 +1,87 @@
...
+
+    # --- User-Facing Operations (Inherited from Atespace) ---
+    define can_get: viewer from parent_atespace
+    define can_update: editor from parent_atespace
+    define can_delete: editor from parent_atespace
+
+    # --- Hardened System & Infrastructure Operations ---
+
+    # Node Restriction: Only the node hosting the scheduled worker can mint tokens
+    define can_mint_credential: host_node
```
ahmedtd: Can you make this `can_mint_ateom_actor_credential`?  I am working on splitting apart ateom credentials (used by ateom to connect to the gateway) vs actor credentials (injected by the gateway for the actor's own actions).

ahmedtd (dismissed):

internal/authz/model.fga
SHRUTI6991: Editor role is someone who has permissions for create / update of resources but doesn't have delete permissions. I don't see the need just yet to have cluster editor permissions. Owner seems sufficient for now.

But yes viewer is a good one. They can get or list atespaces. I added it.

internal/authz/model.fga
SHRUTI6991: Done.

internal/authz/model.fga
SHRUTI6991: Done.

internal/authz/model.fga
SHRUTI6991: Done.

internal/authz/model.fga
SHRUTI6991: In OpenFGA, `user:*` represents a wildcard matching any valid identity of type user (similar to `system:authenticated` in Kubernetes or allAuthenticatedUsers in GCP). 

But thinking about it, we shouldn't open atespace public for all. Since that's the security boundary. So I have only kept viewer as `[user] or editor`.

I was initially thinking how we can give public access to Actor Templates: For eg: Here is a Python 3.12 sandbox template that anyone in the company can launch, this is a very common use case for Sandboxing. 

But I think, it is actually much cleaner and safer to handle this via an explicit shared catalog (e.g. a system atespace `atespace:catalog` where members are granted read access via groups like `group:all-devs#member`), rather than opening wildcards (user:*) at the schema level.

ahmedtd: OK, that makes sense.  I agree we shouldn't worry about user:* for now.

PR #1235, opened 2026-08-26
ci: stop the testcontainers reaper from deleting live test databases

### LINKED ISSUE #1230: Unit test race: cmd/ateapi/internal/controlapi
### Agent reported `go test -race` failure

Transient Failure: Testcontainer Database Connection Drop under Load
  
  Failing Package: github.com/agent-substrate/substrate/cmd/ateapi/internal/controlapi

### Symptoms

```  
storetest.go:74: dropping PostgreSQL test database: conn closed
workflow_suspend_test.go:211: creating PostgreSQL test database: failed to connect to `user=postgres database=postgres`:

127.0.0.1:32781 (localhost): failed to receive message: read tcp 127.0.0.1:...->127.0.0.1:32781: read:    
connection reset by peer
```

### Potential Root Cause (need verification)

When go test -race ./... executes every package concurrently, multiple packages spin up Docker PostgreSQL containers and create/drop dozens of databases in parallel via storetest.go:51-78.

Under the high CPU/memory overhead of the race detector, the Docker container port/connection dropped mid-suite, causing subsequent subtests to fail with connection reset / refused errors.
  
When tested individually (go test -race ./cmd/ateapi/...), all tests in controlapi pass cleanly with zero race warnings.

orangeCatDeveloper:
Fixes #1230

`cmd/ateapi/internal/controlapi` fails ~30 tests at once in `run-tests`, every one on `creating PostgreSQL test database: conn closed`. A package's PostgreSQL container is deleted while that package is still using it.

## Root cause

Every package binary under one `go test ./...` derives the same testcontainers session ID from the shared parent process, so all five packages that start a PostgreSQL container share one Ryuk reaper and one set of container labels. Ryuk deletes everything carrying that label once its connected-client count reaches zero.

The original failure had two upstream contributors:

- Reusing an existing reaper waited only for its Docker port mapping. Docker exposes the port before Ryuk is listening, so the handshake could fail with `read ack: EOF`. [testcontainers-go#3743](https://github.com/testcontainers/testcontainers-go/issues/3743) was fixed by [#3761](https://github.com/testcontainers/testcontainers-go/pull/3761); testcontainers-go v0.44.0 is now in main via #1237.
- `Reaper.connect` logs a failed handshake or startup and returns success anyway, so a package can run unregistered while its containers still carry the shared session label. [#3827](https://github.com/testcontainers/testcontainers-go/issues/3827) remains open upstream. Under v0.44.0 the observed failure moves to `wait for reaper <id>: context deadline exceeded` rather than eliminating the shared-reaper race.

One package then exits, Ryuk sees zero clients, and prunes the session, including a database another package is mid-suite on. `storetest`'s `sync.Once` blocks recreation, so every remaining test in that package fails on the same line.

The package count has since grown from four to five. That increases contention but does not change the defect.

## Fix

Disable the reaper for `run-tests`.

The reaper exists to clean up after a process that dies without doing so. `storetest` and `atepg` now terminate their own containers, and a GitHub runner is destroyed with the job, so CI does not need a shared reaper. Local development is untouched and keeps it.

No test accompanies this: the failure lives in CI job composition and a dependency, neither of which is reachable from a Go test. Evidence is below instead.

## Evidence

Ryuk's own log during a failing run; the container is removed, not crashed:

```
client disconnected  clients=0
prune check          clients=0
removed containers=1 networks=0 volumes=0 images=0
```

Forcing exactly one package to lose its reaper connection, with nothing else changed:

```
control   ok    cmd/ateapi/internal/store/atepg              10.229s
          ok    cmd/ateapi/internal/controlapi/functionaltest 30.673s

forced    ok    cmd/ateapi/internal/store/atepg              10.229s
          FAIL  cmd/ateapi/internal/controlapi/functionaltest 15.461s
                creating PostgreSQL test database: failed to connect to `user=postgres database=postgres`:
                  127.0.0.1:55110 (localhost): failed to receive message: read: connection reset by peer
```

30 of the 30 failing test names are among the 32 seen in run [32926359026](https://github.com/agent-substrate/substrate/actions/runs/32926359026).

## Related changes

- #1237 upgraded testcontainers-go to v0.44.0 and fixed the early handshake race for local runs, where the reaper stays enabled. It does not replace this CI change: on a 4-core Linux VM with cold build caches, database tests remained unavailable in 3/3 rounds as `wait for reaper <id>: context deadline exceeded`.
- #1345 proposed making CI container startup errors fail instead of skip, but closed without merging. It is separate from preventing a live container from being deleted.

---

- [x] Tests pass

### DIFF
--- a/.github/workflows/pr-workflow.yaml
+++ b/.github/workflows/pr-workflow.yaml
@@ -29,6 +29,10 @@ permissions:
 jobs:
   run-tests:
     runs-on: ubuntu-latest
+    # A lost reaper handshake gets a package's containers pruned mid-run
+    # (testcontainers-go#3827); a throwaway runner needs no reaping anyway.
+    env:
+      TESTCONTAINERS_RYUK_DISABLED: "true"
     steps:
     - name: Checkout
       uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0


orangeCatDeveloper:
Split out the dependency bump into #1237 so this stays reviewable at four lines. The bump does not replace this change — see the cold-cache numbers there.

BenTheElder:
> Cost: the three packages that share storetest's fixture leave their container behind until the runner is torn down, because storetest has no termination hook. atepg terminates its own.

That seems unacceptable. Wherever possible CI should match local development behavior.

Leaking containers is a bug.

.github/workflows/pr-workflow.yaml
```diff
@@ -29,6 +29,10 @@ permissions:
 jobs:
   run-tests:
     runs-on: ubuntu-latest
+    # A lost reaper handshake gets a package's containers pruned mid-run
+    # (testcontainers-go#3827); a throwaway runner needs no reaping anyway.
```
BenTheElder, line 33: we should just fix the handshake. relying on "it's in CI" leads to bugs that cannot be tested locally.

.github/workflows/pr-workflow.yaml
```diff
@@ -29,6 +29,10 @@ permissions:
...
+    env:
+      TESTCONTAINERS_RYUK_DISABLED: "true"
```
jpbetz: I think the leak half of the objection is lifted now, by #1297: it merged 2026-08-29 and this branch is already rebased past it, so `storetest.Shutdown` terminates the shared container from `TestMain` the way `atepg` always did for its own. I tried it locally: with `TESTCONTAINERS_RYUK_DISABLED=true` the actoridentity package starts no Ryuk container at all and logs `Container terminated` on the way out. One honest gap: if `go test -timeout` panics the binary, `TestMain` never gets there and that package's container lingers until the runner is torn down. The description's "Cost" paragraph still says storetest has no termination hook, which is the sentence that got quoted — worth refreshing to the above.

On "just fix the handshake": the reuse race (testcontainers-go#3743) is already fixed and shipped in v0.44.0, which we picked up in #1237. What's left is #3827, where `connect` logs the failed handshake and hands back a live termination channel anyway (`vendor/.../reaper.go:538-554`); its fix (#3841) is open upstream, so fixing that means waiting on a release either way. There may be one in-tree option worth ruling out before we accept a CI-only divergence, though: the root cause is the shared session ID (`bootstrap.SessionID()` hashes the parent pid, so every package binary under one `go test ./...` lands on one reaper), and `TESTCONTAINERS_SESSION_ID` (`internal/config/config.go:133`) overrides it. Setting it per test binary from `storetest`/`dockerenv` before the first container starts would give each package its own reaper that prunes only its own containers, with identical behavior in CI and locally. I've only seen it tried on one package, not the full four-package suite, so I may be missing something — but if it holds, that's the fix Ben is asking for and this env var isn't needed. If it doesn't, I'd state the tradeoff plainly in the description: once `TestMain` terminates the container, Ryuk is only a crash safety net, and a runner destroyed with the job doesn't need one.

.github/workflows/pr-workflow.yaml
```diff
@@ -29,6 +29,10 @@ permissions:
 jobs:
   run-tests:
     runs-on: ubuntu-latest
+    # A lost reaper handshake gets a package's containers pruned mid-run
```
jpbetz: Not blocking this PR, but the follow-up you flag in the description is the piece I'd most like to see land right behind it, because it's what tells us whether this change worked. At head, `storetest.go:186` and `atepg_test.go:112` still `t.Skipf` on *any* `containerErr`, and the message says "(requires Docker)" without ever checking for Docker — #1297 actually widened that path a bit by routing `dockerenv.Configure` failures into the same bucket.

It isn't hypothetical: job 98021431239 (branch k8s_dv, 2026-08-26, pre-v0.44.0) logs 154 `PostgreSQL testcontainer unavailable ... wait for reaper ...` skips in controlapi and still prints `ok  .../controlapi  2.992s`; that run only went red on an unrelated later step. `go test -v` does put the `--- SKIP` lines in the log, so it isn't invisible, but nothing in job status distinguishes "the contract suite passed" from "the contract suite never ran".

What I'd want: `t.Fatalf` when `CI` is set (GitHub sets it and the runner has Docker, so a container failure there is by definition a real failure), and skip locally only after a preflight that shows Docker is genuinely absent. I have a follow-up ready for it unless you'd rather fold it in here. Small thing while you're in there: the description's line references (`storetest.go:153` / `atepg_test.go:107`) went stale in the rebase.

.github/workflows/pr-workflow.yaml
jpbetz: nit: the comment cites the upstream issue but not ours. Worth adding #1230 so the env var doesn't get tidied out later by someone who can't see what it's holding up — the failure mode is subtle enough that the local breadcrumb earns its line. Fine as a follow-up if you'd rather not respin.

```suggestion
    # (testcontainers-go#3827, see #1230); a throwaway runner needs no reaping anyway.
```

orangeCatDeveloper: Locally the reaper stays on and the blast radius of #3827 is small — worst case a flaky full-suite run, and the fix for that is upstream ([testcontainers-go#3841](https://github.com/testcontainers/testcontainers-go/pull/3841), not yet merged).

Disabling Ryuk in CI is also what the testcontainers [docs](https://golang.testcontainers.org/features/garbage_collector/) recommend: "We recommend using it only for Continuous Integration services that have their own mechanism to clean up resources." — a GitHub-hosted runner is a fresh VM discarded after the job, so there's nothing left to reap.

PR #1239, opened 2026-08-26
ateom-microvm: promote checkpoint/restore phase timings to metrics

Oneimu:
## What this pr do:

Exports ateom-microvm's checkpoint and restore phase breakdowns — until now visible only as structured log lines — as OTel histograms, and records the size of the OnDemand delta a checkpoint merges away.

New instruments (registered in `docs/metrics/registry/metrics.yaml`):

- `ateom.actor.checkpoint.duration` — phases `pause`, `snapshot`, `durable_dir`, `rootfs_upper`, `merge`, `teardown`, `total`. The three captures run concurrently on the paused guest, so they are independent observations, not a partition of the total.
- `ateom.actor.restore.duration` — phases `prep`, `bundles`, `upper_join`, `lowers`, `tap`, `vmm_launch`, `vm_restore`, `resume`, `readyz`, `total`. These are sequential and do partition the total. A Data-scope restore (cold boot) records only the total.
- `ateom.snapshot.delta.size` — populated bytes of the OnDemand delta before `MergeDeltaIntoBase` overlays it onto the restore source. The merge code already computed this and discarded it; both merge functions now return it.

All series carry the shared `ate.snapshot.phase` / `ate.snapshot.scope` / template attributes, so they decompose the existing `ateom_restore` / `ateom_checkpoint` phases of atelet's `ate.actor.*.duration` histograms. Everything rides the meter provider ateom already pushes through atelet's OTLP relay — no new export plumbing. Phases are recorded where the logs are (success path); failures remain visible on atelet's histograms. A nil `*Instruments` is a no-op, so hand-built test services need no meter.

## Why is it needed:

The suspend/resume benchmarking effort needs percentiles, not anecdotes, for where time goes *inside* the micro-VM lifecycle. Today the atelet-level phases end at opaque `ateom_restore` / `ateom_checkpoint` blocks, and the finer breakdown exists only in logs. With the large-memory benchmark workloads landing separately, these histograms are what splits a slow resume between rootfs assembly, the VMM memory restore, and the readyz wait — and `ateom.snapshot.delta.size` vs. the memory image size is the single number that says whether differential snapshot uploads are a 2x or a 10x win.

## Testing

- `GOOS=linux go build` / `go vet ./cmd/ateom-microvm/...` clean
- `go test ./cmd/ateom-microvm ./internal/ateattr` pass (new metrics tests run host-side against a ManualReader; merge tests assert the returned delta byte count)
- `gofmt` clean; registry validated with `hack/verify/metrics.sh`

### DIFF
--- a/cmd/ateom-microvm/checkpoint.go
+++ b/cmd/ateom-microvm/checkpoint.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/agent-substrate/substrate/cmd/ateom-microvm/internal/ch"
 	"github.com/agent-substrate/substrate/cmd/ateom-microvm/internal/kata"
+	"github.com/agent-substrate/substrate/internal/ateattr"
 	"github.com/agent-substrate/substrate/internal/ateompath"
 	"github.com/agent-substrate/substrate/internal/ateomstats"
 	"github.com/agent-substrate/substrate/internal/imagecache"
@@ -70,6 +71,7 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
 	s.setActiveRPC(rpcCheckpointWorkload, cancel)
 	defer s.clearActiveRPC()
 
+	tStart := time.Now()
 	if err := s.deactivateActorNetworking(ctx); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


BenTheElder:
I don't know that we want to do this, this was something @chw120 and I discussed after you dropped from the meet yesterday IIRC.

This exposes to the user what the phases are, but those are implementation details subject to change. Metrics are an API.

Keeping them as debug logs leaves them developer facing.

BenTheElder:
see also https://github.com/agent-substrate/substrate/pull/1200, we abandoned for the same reason. if the data is primarily for developers of substrate, we should not expose it as a metric

if the data is for *end users* to manage their workloads, then we should, but be careful about how we shape it

Issue #1168, opened 2026-08-24
Adopt Declarative Validation (DV) for resource operations


laoj2:
Today, we lack a robust, declarative mechanism to validate resource operations and enforce invariants.

Relying on manual validation code makes it harder to maintain, scale, and test confidently. We need a systematic way to wield validation to ensure solid tests and API invariants during resource creation.

zlammerts-svg:
@laoj2 Is this one done as this point? Thanks!!

laoj2:
@zlammerts-svg  Not yet. We have a couple of fields in `ateapi.proto` that are missing validation. I'll send a PR to close the gap

PR #1244, opened 2026-08-26
Extend declarative validation to Actor 

shrutiyam-glitch:
In this PR: 
**Actor read/lifecycle verbs — completes Actor end-to-end**
*   `Get`/`Delete`/`Suspend`/`Pause`/`Resume` `ActorRequest`: dropped the `+k8s:opaqueType` fence on the actor ref, added `+k8s:required` + `+k8s:subfield(atespace)=+k8s:required`. Combined with the recursion into `Validate_ObjectRef` this reproduces exactly what `resources.ValidateObjectRef` enforced (ref required, atespace required, name required, both DNS-1123 labels).
*   `ListActorsRequest`: tagged to match `ListAtespacesRequest` (atespace optional + short-name format, `page_size` minimum 1, `page_token` max length 256).
*   All six `validate*Request()` functions are now one-line calls to the generated validators. With this, every RPC verb on `Actor` and `Atespace` is fully declarative; `resources.ValidateObjectRef` has no callers left in `actor.go`.
  
**Note:**`ListActorsRequest.page_token` now capped at 256 chars (parity with `ListAtespaces`).
  
**Testing**
*   Existing verb tests updated to the generated error shapes (`field.ErrorMatcher` with origin matching).
*   New coverage: `page_token` boundary cases for `ListActors`; 
*   `controlapi`, `functionaltest`, `actoridentity`, `store`, and `atepg` suites all pass; `go vet` and `gofmt` clean.

- [ ] Tests pass


cmd/ateapi/internal/controlapi/actor.go
thockin: Remove TODOs now that all of these paths are implemented

thockin: remove TODO

thockin: remove TODO

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -113,12 +113,26 @@ service Control {
...
+  // names are server-generated UUIDs, but any resource name is valid here.
+  //
+  // +k8s:optional
+  // +k8s:format=k8s-short-name
   string snapshot_name = 1;
+
   // Node VMs that have local snapshots for this actor, while it's PAUSED.
+  //
+  // +k8s:optional
+  // +k8s:listType=atomic
```
thockin: Is it OK to have duplicate values in this list?  If no, set `listType=set`

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -113,12 +113,26 @@ service Control {
 }
 
 message LocalSnapshotInfo {
-  // The name of the local checkpoint on each of the nodes below.
+  // The name of the local checkpoint on each of the nodes below. Checkpoint
+  // names are server-generated UUIDs, but any resource name is valid here.
+  //
+  // +k8s:optional
+  // +k8s:format=k8s-short-name
```
thockin, line 133: I do not see tests for this

pkg/proto/ateapipb/ateapi.proto
thockin: Every list should come with a `+k8s:maxItems`  That IMMEDIATELY brings this whole aspect of the API into question.  Is it allowed to have a snapshot propagate to all machines?  I would think yes -- are we really going to update this field to have 1000 or 10000 or 100000 entries?

@juli4n @BenTheElder @dberkov 

I think, for now, set it to something like 256, but we need to revisit this design

thockin: Every list should come with some sort of per-item validation, too.  `+k8s:maxLength=128` or something?

thockin: And tests! :)

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -219,13 +234,20 @@ message ResourceMetadata {
 
 message ExternalVolume {
   // Name of the volume specified in the actor template.
+  //
+  // +k8s:required
```
thockin, line 257: Every string should come with a max-length check if not a full format check.  @msau42 what defines the name format?

thockin: And this needs a testcase

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -219,13 +234,20 @@ message ResourceMetadata {
...
   string volume_name = 1;
 
   // The globally unique volume_id returned from the storage system.
-  // This will be initially empty during volume creation
+  // This will be initially empty during volume creation. Its format is the
+  // storage system's own, so it is not validated beyond presence.
+  //
+  // +k8s:optional
```
thockin, line 271: Can we at least put a max-len on this?  @msau42

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -219,13 +234,20 @@ message ResourceMetadata {
...
   string storage_volume_id = 2;
 
   // Internal volume plugin name or CSI driver name.
+  //
+  // +k8s:optional
```
thockin: Same - some sort of validation

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -236,10 +258,19 @@ message ExternalVolume {
...
+  // +k8s:minimum=1
+  // +k8s:maximum=3 # keep this in sync with the Status enum
   Status status = 4;
+
   // volume_context contains metadata returned by the CSI driver during volume
-  // provisioning, needed by the node plugin for mounting (e.g. attachment info).
+  // provisioning, needed by the node plugin for mounting (e.g. attachment
+  // info). Keys and values are the driver's own, so they are not validated.
+  //
+  // +k8s:optional
```
thockin, line 310: We need some `maxProperties` limit and some per-key and per-value limits at least.

```
// +k8s:maxProperties: 32
// +k8s:eachKey=+k8s:maxLength=128
// +k8s:eachVal=+k8s:maxLength=128
```

Something like that.  And tests!

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -322,33 +353,51 @@ message ActorStatus {
   // +k8s:update=NoModify # can be set and cleared, but not changed in place
   WorkerAssignment worker_assignment = 2;
 
-  // TODO: Add DV (optional, maxLength?)
+  // The name the in-progress durable snapshot will be stored under. Snapshot
+  // names are server-generated UUIDs, but any resource name is valid here.
+  //
+  // +k8s:optional
+  // +k8s:format=k8s-short-name
```
thockin, line 517: I don't see a test case for this

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -322,33 +353,51 @@ message ActorStatus {
...
   string in_progress_snapshot_name = 3;
 
-  // The latest durable snapshot created for this Actor.
-  // +k8s:opaqueType
-  // TODO: Add DV (required?)
+  // The latest durable snapshot created for this Actor. Unset until the first
+  // snapshot exists.
+  //
+  // +k8s:optional
+  // +k8s:subfield(atespace)=+k8s:required
```
thockin, line 524: I don't see a test case for this

thockin: I won't repeat this comment, but it should be easy to add a testcase for  each of these

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -322,33 +353,51 @@ message ActorStatus {
...
+  // +k8s:optional
+  // +k8s:minimum=1
   int64 in_progress_snapshot_source_actor_version = 6;
 
   // Volumes attached to the actor. These volumes only live as long as the actor.
   // They are deleted when the actor is deleted.
-  // TODO: Add DV (optional, need to recurse into this type, immutable?)
+  //
+  // +k8s:optional
+  // +k8s:listType=atomic
```
thockin: we probably do not need to declare atomic lists - that is the default

thockin: Actually, let's poke at this one.

I am assuming that this list should not have 2 entries for the same volume_name, right?  @msau42 

If that is right, you can do something like:

```
// +k8s:listType=map
// +k8s:listMapKey=volume_name
```

@juli4n should we venture into list-maps or actually write these things as literal maps?  That would be a different PR, of course

thockin: Also needs a maxLength (or maxProperties if map)

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -322,33 +353,51 @@ message ActorStatus {
...
   string in_progress_local_snapshot_name = 8;
 
   // How this Actor was seeded from an ActorSnapshot at CreateActor. Unset if
-  // the Actor was not created from a snapshot.
-  // TODO: Add DV (optional, need to recurse into this type, immutable?)
+  // the Actor was not created from a snapshot. Set once at creation and
+  // never changed afterward.
+  //
+  // +k8s:optional
+  // +k8s:update=NoModify # set at creation, never changed in place
```
thockin: use `immutable` here?  `NoModify` means it can be set or cleared.  The utility of that is rare.

thockin (commented):
Mostly about testing and bounds checking.  Thanks for picking this up!

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: I've made it `NoModify` + `NoUnset`, which together give "set once, never changed, never cleared."
`immutable` didnt work here since `CreateActor` validates the built object against the scrubbed request with `op=Update` - https://github.com/agent-substrate/substrate/blob/main/cmd/ateapi/internal/controlapi/actor.go#L106 , and since the request's status is nil, the server populating `source_snapshot` is a `nil→set` that `immutable` rejects.

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: For now, considered `actor_volumes` as map with a todo.

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: set `maxItems` as 256 with a TODO.

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: sorry, not sure how I missed the test cases for all these fields :(
added now, thanks

pkg/proto/ateapipb/ateapi.proto
msau42: I have a PR specifically for externalvolumes here: https://github.com/agent-substrate/substrate/pull/678/changes#diff-f29aa971d0e861ade55215208c2d5e2eaa7d9f743969517826128207b1e4095e

Do you want to pull in my changes?

msau42: For this specific field, k8s-short-name is fine.

pkg/proto/ateapipb/ateapi.proto
msau42: CSI spec default limit is 128 if not otherwise specified. Although this isn't actually enforced in k8s so it may be possible some drivers have gone over. Let's start with 128 but be open to changing it.

pkg/proto/ateapipb/ateapi.proto
thockin: Ooooh, that's UGLY.

thockin: https://github.com/agent-substrate/substrate/issues/1359

thockin:
LGTM but it has an actual error flagged

Error: cmd/ateapi/internal/controlapi/actor_test.go:100:23: undefined: refOnly

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -231,14 +251,39 @@ message ResourceMetadata {
...
 
 message ExternalVolume {
-  // Name of the volume specified in the actor template.
+  // Name of the volume specified in the actor template. Template volume
+  // names are DNS labels.
+  //
+  // +k8s:required
+  // +k8s:format=k8s-short-name
+  // +k8s:update=NoModify
+  // +k8s:update=NoUnset # set-once; immutable would reject the create
```
msau42, line 260: this has to be set on creation so immutable is fine

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -231,14 +251,39 @@ message ResourceMetadata {
...
 
-  // Internal volume plugin name or CSI driver name.
+  // Internal volume plugin name or CSI driver name, from the StorageClass
+  // provisioner. Must be a DNS subdomain, optionally prefixed with
+  // "substrate.io/" (per the CSI spec's driver-name syntax).
+  //
+  // +k8s:required
+  // +k8s:maxLength=253
+  // +k8s:update=NoModify
+  // +k8s:update=NoUnset # set-once, like volume_name above
```
msau42, line 285: This also needs to be set at creation

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: Right, but the way we are verifying the creation object doesn't allow us to use `immutable` in the DV. 
Related github issue - https://github.com/agent-substrate/substrate/issues/1359 
Maybe we could handle this separately ? And update the DV all of such fields in a separate PR ?

thockin: The way we use DV today has an issue with immutable in status.  We will have to fix that and then fix this.  This will have to be "close enough" for now.

thockin:
LGTM

PR #1246, opened 2026-08-26
Delete local snapshots when an actor is terminated

laoj2:
We were leaking snapshots after actor termination.

Note that this is a temporary fix: it only removes local snapshots in one node. We should clean up the copies on any other NodeVmsWithLocalSnapshots. This is fine *as of the day this was written* because today NodeVmsWithLocalSnapshots has at  most one item.

Related to https://github.com/agent-substrate/substrate/issues/668 and #664 

- [x] Tests pass
- [] Appropriate changes to documentation are included in the PR

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -660,7 +660,11 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	// all: the actor's current state was just captured by CheckpointWorkload,
 	// and the control plane tracks only a single local snapshot, which this
 	// checkpoint either overwrites (pause) or clears (suspend).
-	pruneLocalCheckpoints(ctx, actorUID)
+	//
+	// Best-effort: if this fail, the actor's terminate prunes again.
+	if err := pruneLocalCheckpoints(ctx, actorUID); err != nil {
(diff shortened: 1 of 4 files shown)


cmd/atelet/main.go
```diff
@@ -1264,6 +1264,13 @@ func (s *AteomHerder) Terminate(ctx context.Context, req *ateletpb.TerminateRequ
 		return nil, fmt.Errorf("failed to unmount external volumes during terminate (actor: %s, actorUID: %s): %w", actorRef, actorUID, err)
 	}
 
+	// The actor is gone, so no pause snapshot of it can ever be restored again.
+	// TODO(#664): this only removes local snapshots in one node. We should clean
+	// up the copies on any other NodeVmsWithLocalSnapshots. This is fine *as of
+	// the day this was written* because today NodeVmsWithLocalSnapshots has at
+	// most one item.
+	pruneLocalCheckpoints(ctx, actorUID)
```
zoez7: `pruneLocalCheckpoints` might fail and still leak the local snapshots on disk, Should this return an error and have control plane retry?

zoez7 (dismissed):

cmd/atelet/main.go
laoj2: good point. Done

zoez7 (approved):

PR #1249, opened 2026-08-26
Switch default region/zone from us-central1(-c) to us-west1(-c).

cheftako:
Switch default location from us-central1 to us-west1.

### DIFF
--- a/hack/tools/code-generator/third_party/k8s.io/apimachinery/pkg/util/managedfields/node.yaml
+++ b/hack/tools/code-generator/third_party/k8s.io/apimachinery/pkg/util/managedfields/node.yaml
@@ -13,10 +13,10 @@ metadata:
     kubernetes.io/os: linux
     cloud.google.com/gke-nodepool: default-pool
     cloud.google.com/gke-os-distribution: cos
-    failure-domain.beta.kubernetes.io/region: us-west1
-    failure-domain.beta.kubernetes.io/zone: us-west1-b
-    topology.kubernetes.io/region: us-west1
-    topology.kubernetes.io/zone: us-west1-b
+    failure-domain.beta.kubernetes.io/region: us-central1
+    failure-domain.beta.kubernetes.io/zone: us-central1-b
+    topology.kubernetes.io/region: us-central1
+    topology.kubernetes.io/zone: us-central1-b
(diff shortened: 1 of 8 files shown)


hack/tools/code-generator/third_party/k8s.io/apimachinery/pkg/util/managedfields/node.yaml
LiorLieberman, line 1: likely an artifact of search and replace? probably dont need to change it?

third_party/k8s.io/apimachinery/pkg/util/managedfields/node.yaml
```diff
@@ -13,10 +13,10 @@ metadata:
     kubernetes.io/os: linux
     cloud.google.com/gke-nodepool: default-pool
     cloud.google.com/gke-os-distribution: cos
-    failure-domain.beta.kubernetes.io/region: us-central1
-    failure-domain.beta.kubernetes.io/zone: us-central1-b
-    topology.kubernetes.io/region: us-central1
-    topology.kubernetes.io/zone: us-central1-b
```
LiorLieberman, line 19: likely an artifact of search and replace? probably dont need to change it?

LiorLieberman (dismissed):
Makes sense to me. 
LGTM, will merge it by EOD if no one objects

third_party/k8s.io/apimachinery/pkg/util/managedfields/node.yaml
cheftako: We have things like hack/tools/code-generator/third_party/k8s.io/apimachinery/pkg/util/managedfields/fieldmanager_test.go which reference node.yaml. It seemed like we either didn't actually use it, in which case it should be f=safe to change it OR we do use it and it should be changed.

BenTheElder: this code is vendored from Kubernetes, so we should not change it.

hack/tools/code-generator/third_party/k8s.io/apimachinery/pkg/util/managedfields/node.yaml
cheftako: Done

Issue #1005, opened 2026-08-17
Publish a guide on how to make actors trust the egress MITM trust bundle


haiyanmeng:
After both https://github.com/agent-substrate/substrate/issues/823 and https://github.com/agent-substrate/substrate/issues/932 land, we should publish a guide on how to wire the provided ClusterTrustBundle for `ate-egress-trust` into its trust store.

haiyanmeng:
https://github.com/agent-substrate/substrate/issues/932 makes the egress MITM trust bundle available in the filesystem of an actor. For example, the following config would make the egress MITM trust bundle available at `/run/ate/trust-bundle.pem`.

```
apiVersion: ate.dev/v1alpha1
kind: ActorTemplate
metadata:
  name: egress
  namespace: ate-demo-egress
spec:
  # Requires an sdsmint install (--experimental-use-sdsmint): only that path
  # creates the egress-mitm-ca-pool Secret the bundle is derived from, and an
  # actor referencing a bundle that does not resolve fails to start.
  volumes:
  - name: system-info
    systemInfo:
      dataSources:
      # The trust anchors for the per-SNI leaves the MITM egress gateway mints.
      - trustBundle:
          name: egress-mitm.ate.dev
          path: trust-bundle.pem
  containers:
  - name: egress
    image: ko://github.com/agent-substrate/substrate/demos/egress
    command: ["/ko-app/egress"]
    volumeMounts:
    - name: system-info
      mountPath: /run/ate   # the bundle lands at /run/ate/trust-bundle.pem
```

To make an actor trust `/run/ate/egress-ca.pem`, we can recommend the ActorTemplate owner to add the environment variables `SSL_CERT_FILE` and `SSL_CERT_DIR` into each container:

```
apiVersion: ate.dev/v1alpha1
kind: ActorTemplate
metadata:
  name: egress
  namespace: ate-demo-egress
spec:
  # Requires an sdsmint install (--experimental-use-sdsmint): only that path
  # creates the egress-mitm-ca-pool Secret the bundle is derived from, and an
  # actor referencing a bundle that does not resolve fails to start.
  volumes:
  - name: system-info
    systemInfo:
      dataSources:
      # The trust anchors for the per-SNI leaves the MITM egress gateway mints.
      - trustBundle:
          name: egress-mitm.ate.dev
          path: trust-bundle.pem
  containers:
  - name: egress
    image: ko://github.com/agent-substrate/substrate/demos/egress
[...]

PR #1252, opened 2026-08-27
ateapi: auto-inject the egress trust volume into every actor

thompsonmax:
Part of #932 (PR 3 of 3; #941 is merged, #1231 is orthogonal). Closes #871's open item: on a deployment that intercepts actor egress TLS, every actor gets the CA anchors with nothing declared in its template.

### What this does

This PR adds an ateapi flag, --inject-egress-trust-bundle (default off), that adds one systemInfo volume, trust.ate.dev, projecting the egress-mitm.ate.dev bundle to /run/substrate/certs/egress-mitm.ate.dev.pem in every container. Injection happens in ActorWorkflow.workloadSpec, the one place every atelet RPC builds its spec, so Run, Restore, both Checkpoint types, and Terminate agree on the volume set. The wire still carries only {name, path}; atelet resolves the bundle node-side as for a declared projection, so a missing bundle fails actor start with no new code. No atelet, CRD, or RBAC changes.

- /run/substrate is reserved for platform-injected volumes. CreateActorTemplate rejects any mountPath at or under it, flag or not, so nothing a template declares can shadow the injected file, and later injections (actor identity) get a home without another API change. As a backstop the injector fails the spec build on a mount under the prefix, on an image volume above the injected path (micro-VM binds image volumes after systemInfo), and on a volume named trust.ate.dev (impossible from a template, since volume names are DNS labels). The api-guide's manual projection example moves to /run/ate. An earlier revision treated a template mount at the path as an opt-out and logged it. Review preferred the reservation, since an unrelated mount there would have quietly cost the actor its trust anchors.
- ateapi has no runtime signal for "MITM is deployed", so hack/install-ate.sh passes the flag whenever it deploys ate-api-server under --experimental-use-sdsmint, and after every rollout it fails the install if a deployed sdsmint gateway is missing the flag. CI's MITM step redeploys ateapi with it and recreates the counter demos through a new --deploy-demo-counter-microvm handler.
- A FULL snapshot only restores against the spec it was captured with (gVisor refuses the mismatch; a micro-VM guest resumes its old mount table without the file), so the flag belongs at install time, before templates and workloads exist. A later flip means regenerating goldens and discarding paused or suspended actors. The docs and the flag help state this.

TestValidateVolumeMountMountPath in controlapi's validation_test.go covers the reserved-prefix rule at template validation, and TestInjectEgressTrustVolume in egresstrust_test.go covers each injector rule, including the backstop errors. TestActorLifecycle_InjectsEgressTrustVolume in the controlapi functional tests asserts the injected volume in the Run, Checkpoint, Restore, and Terminate wire specs. TestActorEgressTrustAutoInjection in the egressmitm suite, on both sandbox classes, reads the injected file from a probe that declares nothing about trust, checks it against the published ClusterTrustBundle, and completes TLS through the MITM gateway with only those anchors.

Not in this PR: a per-template opt-out (a follow-up field; templates are immutable, so it stays snapshot-stable), env vars (#1424), live refresh (#1231), the backend registry (#932), and trust-pickup guidance (#1005).


haiyanmeng:
> A new ateapi flag, --inject-egress-trust-bundle (default off), adds one systemInfo volume, trust.ate.dev, projecting the egress-mitm.ate.dev bundle to /run/substrate/certs/egress-mitm.ate.dev.pem in every container.

@thompsonmax , do you know the time overhead of auto-injecting the egress trust volume? IIUC, this will be on the critical path of every actor resume operation. 

AFAIK, we currently plan to auto-inject MITM trust bundle and actor identity. Each of these auto-injection will have a overhead. If the number of auto-injected info keep going up, the time overhead may be too high for our sub-100ms actor-resume goal. 

> Opt-out: a container keeps its own mounts and gets no injection when it mounts a volume at /run/substrate/certs or below it (the injected parent bind would shadow a nested mount on the micro-VM runtime), or an image volume above it (micro-VM binds image volumes after systemInfo, shadowing the injected mount). Every opt-out is logged with the actor, container, and colliding path. If all containers opt out, no volume is injected.

Should we disallow users to mount things at `/run/substrate` and fail to start an actor if it happens?

Strictly speaking, the egress MITM trust volume is only needed for actors generating HTTPS traffic. I wonder whether we should allow users to indicate whether an actor generates HTTPS traffic. This can be a followup PR. @bowei , WDYT?

haiyanmeng:
> Env vars (SSL_CERT_FILE and friends) stay out of scope; https://github.com/agent-substrate/substrate/issues/1005's proposal has no signoff, and whatever lands there can point at this path.

I filed https://github.com/agent-substrate/substrate/issues/1424 to this

.github/workflows/pr-workflow.yaml
haiyanmeng, line 1: @thompsonmax ,  what is the impact of this change on `e2e-test` presubmit? How much longer does it make?

cmd/ateapi/main.go
```diff
@@ -67,8 +67,9 @@ var (
 	authenticationConfigFile = pflag.String("authentication-config", "", "YAML file configuring trusted JWT providers.")
 	postgresConnectionString = pflag.String("postgres-connection-string", "", "PostgreSQL connection string (libpq DSN or URI).")
 
-	actorIDJWTPoolFile   = pflag.String("actor-id-jwt-pool", "", "The file that contains the serialized JWT authority pool for signing actor JWTs")
-	egressGatewayAddress = pflag.String("egress-gateway-address", "", "Address of the egress PEP. Empty disables tunneled egress.")
+	actorIDJWTPoolFile      = pflag.String("actor-id-jwt-pool", "", "The file that contains the serialized JWT authority pool for signing actor JWTs")
+	egressGatewayAddress    = pflag.String("egress-gateway-address", "", "Address of the egress PEP. Empty disables tunneled egress.")
+	injectEgressTrustBundle = pflag.Bool("inject-egress-trust-bundle", false, "Project the egress-mitm.ate.dev trust bundle into every actor at /run/substrate/certs. Enable only where the deployment publishes the bundle (actors fail to start without it) and before workloads exist; see docs/api-guide.md.")
```
haiyanmeng: remove `; see docs/api-guide.md`

haiyanmeng: `Enable only where the deployment publishes the bundle (actors fail to start without it) and before workloads exist`: I don't fully understand this

thompsonmax:
> > A new ateapi flag, --inject-egress-trust-bundle (default off), adds one systemInfo volume, trust.ate.dev, projecting the egress-mitm.ate.dev bundle to /run/substrate/certs/egress-mitm.ate.dev.pem in every container.
> 
> Max Thompson (@thompsonmax) , do you know the time overhead of auto-injecting the egress trust volume? IIUC, this will be on the critical path of every actor resume operation.
> 
> AFAIK, we currently plan to auto-inject MITM trust bundle and actor identity. Each of these auto-injection will have a overhead. If the number of auto-injected info keep going up, the time overhead may be too high for our sub-100ms actor-resume goal.

Good question, I did some microbenchmarking on a kind cluster and came up with these numbers on how this affects actor resume latency. Checked the operation latency for both gVisor and microVM.

Local restores, p50 in ms, 20 samples per cell:

| | gVisor off | gVisor on | micro-VM off | micro-VM on |
|---|---|---|---|---|
| atelet `oci_unpack` | 1.0 | 4.5 | 0.9 | 4.4 |
| atelet `download` | 3.3 | 3.4 | 43.3 | 43.5 |
| atelet `ateom_restore` | 364.5 | 359.6 | 177.2 | 181.2 |
| atelet `total` | 372.1 | 368.3 | 228.0 | 230.7 |
| `ResumeActor` round trip | 420.1 | 421.4 | 256.6 | 267.0 |
| `ResumeActor` round trip, mean | 427.8 | 429.6 | 271.4 | 271.8 |

Golden restores, p50 in ms, 5 samples per cell:

| | gVisor off | gVisor on | micro-VM off | micro-VM on |
|---|---|---|---|---|
| atelet `oci_unpack` | 1.0 | 4.7 | 1.1 | 5.5 |
| atelet `download` | 8.9 | 9.2 | 217.2 | 216.1 |
| `ResumeActor` round trip | 414.8 | 382.6 | 399.5 | 405.6 |

So it seems to extend the E2E latency by about 3-5 ms for either runtime. Doesn't seem like an immediate concern, but perhaps something to keep an eye on as/if we increase auto-injected volumes.

Here's the branch with the microbenchmarking code: https://github.com/thompsonmax/substrate/tree/egress-trust-injection-bench

cmd/ateapi/main.go
thompsonmax: Clarified this a bit to mention it requires the sdsmint (MITM) egress gateway to be installed, which publishes the bundle. Actors fail to start when the auto-injection is set but the bundle is not available.

thompsonmax: done

.github/workflows/pr-workflow.yaml
thompsonmax: Quick timing check:

Both runs are from today. Main is commit 9b333c6f, run 33915867922 at 20:22 UTC. The PR is 6ae844e0, run 33923715673 at 22:01 UTC.

| Step | main | PR |
|---|---|---|
| Deploy MITM egress (sdsmint) | 24s | 53s |
| Recreate counter demos under injection | none | 56s |
| E2E egress MITM trust, gVisor | 18s | 45s |
| E2E egress MITM trust, micro-VM | 21s | 41s |
| Total | 63s | 195s |

So, this adds about 2 minutes at the step level.

thompsonmax:
> Should we disallow users to mount things at `/run/substrate` and fail to start an actor if it happens?
> 
> Strictly speaking, the egress MITM trust volume is only needed for actors generating HTTPS traffic. I wonder whether we should allow users to indicate whether an actor generates HTTPS traffic. This can be a followup PR. Bowei Du (@bowei) , WDYT?

Yeah I think rejecting user mounts inside `/run/substrate` makes more sense than silently failing the automount. Added that validation instead of treating it as an opt out.

I think opt-out config can be in a follow-up PR.

PR #1254, opened 2026-08-27
ateom microvm: drop `privileged: true`, using atelet device plugin for KVM

BenTheElder:
This is the companion to #496 

In order to grant access to `/dev/kvm` we have to either:
- use a device plugin
- use a DRA driver
- use an NRI plugin

An NRI plugin is highly privileged in it's own right for all pods on the host and is difficult to ship portably at the moment.
DRA is promising, but not enough functionality is GA yet at our current 1.35+ target.
Device plugin fits reasonably well. We do wind up publishing an ~arbitrarily high limit, which has some cost in kubelet memory, but otherwise is relatively clean.

This approach is what kata uses currently. Their device plugin is not available unbundled, and we anyhow have a per-node daemonset.

atlet is taught to sniff if /dev/kvm appears on the host at all, so we can also stop using the manually labeled nodes for microVM class and instead schedule to the KVM + TUN devices on nodes that advertise them.

Later we can migrate to device plugin by using 

I implemented that already, but I don't think it's worth merging at the moment. We would want consumable capacity to be on by default. We can migrate later without changing the pod spec by using `extendedResourceName`.
https://github.com/agent-substrate/substrate/compare/main...BenTheElder:substrate:ateom-microvm-dra

NOTE: I confirmed with upstream that device plugin is not going anywhere despite being "v1beta1", it's GA in all but name. It won't receive new features but we don't really need anyhow. We'll move to DRA down the line.

---

By doing this, we can drop `privileged: true` from the uVM ateom pods.

We can also drop the `ate.dev/sandboxClass` node label hacks, reducing friction to deploy.


aojea:
check if bind mounting work for /dev/kvm too, it does for /dev/tun https://github.com/google/sam/blob/1fc706fc2a302bc19983caa22f8caf5c238f0597/site/content/docs/user/running-agents.md#in-a-kubernetes-pod

BenTheElder:
> check if bind mounting work for /dev/kvm too, it does for /dev/tun https://github.com/google/sam/blob/1fc706fc2a302bc19983caa22f8caf5c238f0597/site/content/docs/user/running-agents.md#in-a-kubernetes-pod

No it does not work for /dev/kvm, the cgroup device controller will block it. You have to either be privileged or be specifically granted access to the device.

This is why kata ships a device-plugin for KVM as well.

I assumed for tun as well, which turns out to be in the default allowlist ...

BenTheElder:
Need to double check if that's the same in say, crun.

https://github.com/opencontainers/runc/pull/4555

BenTheElder:
crun:
https://github.com/containers/crun/blob/4e862393a767c7ab1ede6146106ad1642ac0277e/NEWS#L254-L256

HavenXia (approved):

PR #1257, opened 2026-08-27
Make sandbox_class in workers immutable

sfunkenhauser:
Changing the sandbox_class requires creating a new worker, you can't change the sandbox_class of a running worker.

- [ x ] Tests pass

### DIFF
--- a/cmd/atecontroller/internal/workersync/syncer.go
+++ b/cmd/atecontroller/internal/workersync/syncer.go
@@ -282,22 +282,27 @@ func (s *WorkerPoolSyncer) createOrUpdateWorker(ctx context.Context, key workerK
 		return fmt.Errorf("getting worker: %w", err)
 	}
 
-	// UpdateWorker replaces the whole resource, so the two mutable fields are
+	// UpdateWorker replaces the whole resource, so the one mutable field is
 	// edited onto the Worker as it was read and the rest is sent back unchanged
 	// — anything else altered here, including a field cleared by omission, is
 	// rejected as INVALID_ARGUMENT. Everything else on a Worker is immutable
 	// after create, so drift there cannot be repaired by an update; it takes a
 	// new pod, which arrives under a new key.
 	var changed bool
-	if w.GetSandboxClass() != string(pool.Spec.SandboxClass) {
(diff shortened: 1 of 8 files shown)


BenTheElder:
> Changing the sandbox_class requires creating a new worker, you can't change the sandbox_class of a running worker.

What does this buy us?

I ask because I'm looking at having workers self register which classes they can handle, which may vary (thinking forward a bit, but imagine if gvisor + uVM were one binary)

sfunkenhauser:
> > Changing the sandbox_class requires creating a new worker, you can't change the sandbox_class of a running worker.
> 
> What does this buy us?
> 
> I ask because I'm looking at having workers self register which classes they can handle, which may vary (thinking forward a bit, but imagine if gvisor + uVM were one binary)

This prevents unsupported modifications to the worker resource, preventing accidental updates that could break the system.

If we want to allow workers to self register their sandbox class, I would think we would expand the WorkerService to handle sandbox class as well as capacity. This does not require the sandbox class in the UpdateWorker API to be mutable.

cmd/atecontroller/internal/workersync/syncer.go
```diff
@@ -282,23 +282,28 @@ func (s *WorkerPoolSyncer) createOrUpdateWorker(ctx context.Context, key workerK
...
-		w.SandboxClass = string(pool.Spec.SandboxClass)
-		changed = true
-	}
 	if !maps.Equal(w.GetLabels(), pool.GetLabels()) {
 		slog.InfoContext(ctx, "Syncer: updating worker (labels changed)", key.logAttrs()...)
 		w.Labels = pool.GetLabels()
 		changed = true
 	}
+	if w.GetSandboxClass() != string(pool.Spec.SandboxClass) {
+		// Expected mid-rollout: sandboxClass drives the worker pod's shape, so
```
juli4n, line 298: I wonder if the WorkerPool controller should add the sandbox class to the Pod as an annotation (in `buildDeploymentApplyConfig`). That way, the Pod is self-descriptive and this whole race condition goes away.

Not part of this PR for sure.

juli4n (approved):

PR #1260, opened 2026-08-27
Extend declarative validation to ActorSnapshotTags

shrutiyam-glitch:

#### Schema (`ateapi.proto`)
* **`ActorSnapshotTag`:** Added DV tags (atespaced metadata, immutable snapshot, and min/max scope bounds).
* **Requests:** Applied standard DV tags (required, atespaced) for Create, Get, Delete, and List verbs. Update follows the two-step pattern (`opaqueType` with custom validation).
* *Note:* The `ActorSnapshot` resource remains fenced out of scope for a future server-written-state migration.

#### Service & Storage Layers
* **Validation:** Replaced manual request validators with 1-line generated DV calls. 
* **Immutability:** Moved tag immutability enforcement from the storage layer (`atepg.go`) to the service layer (`actor_snapshot.go`) via a validate-in-closure pattern on Update.
* **Creation:** `CreateActorSnapshotTag` now properly scrubs server-assigned metadata and enforces that a tag must live in its snapshot's Atespace.

#### Notes
* `ListActorSnapshotsRequest.page_token` is now capped at 256 characters.
* Scope out-of-enum errors now return `INVALID_ARGUMENT` as an `Invalid` type rather than `NotSupported`.
* Metadata `uid`/`version` presence checks moved to the store's precondition check. 
* Snapshot/scope immutability validations moved to the update closure (against-stored-time).

#### Testing
Added comprehensive tests covering all tagged fields, List pagination, immutability guarantees, cross-atespace rejections, server-metadata scrubs, and end-to-end RPC workflows.

- [ ] Tests pass


shrutiyam-glitch:
will be superseded by #1417

PR #1261, opened 2026-08-27
Extend declarative validation to MintJWT/MintCert requests

shrutiyam-glitch:
#### Identity RPCs (`MintJWT` / `MintCert`)
* **Requests:** Added full declarative validation tags (required, formatting, enum bounds) to all fields in `MintJWTRequest` and `MintCertRequest`.
* **Handlers:** Validation now occurs immediately after authentication (the `purpose != ATUNNEL` policy check remains in the handler).

#### Notes
* **Stricter Validation:** Empty `atespace`/`actor_name` can no longer mint malformed JWT subjects. UIDs must now be valid UUIDs.
* **Better Errors:** `MintJWT` empty-audience error is now an `InvalidArgument` (was `Unknown`) checked *before* disk I/O. `MintCert`'s atespace-on-global-ref error is now typed `Forbidden` (was `Invalid`).

#### Testing
* Added 20 new table-driven test cases (`TestValidateMintJWTRequest` / `TestValidateMintCertRequest`), as these requests previously lacked validation tests.
* Updated existing verb tests to match the new generated error shapes.
* All test suites (`controlapi`, `functionaltest`, etc.), `go vet`, and `gofmt` pass cleanly.

- [ ] Tests pass


cmd/ateapi/internal/actoridentity/actoridentity_test.go
```diff
@@ -777,7 +783,9 @@ func TestMintCertActorUID(t *testing.T) {
 		wantCode   codes.Code
 	}{
 		"Matching": {requestUID: func(actorUID string) string { return actorUID }, wantCode: codes.OK},
-		"Stale":    {requestUID: func(string) string { return "uid-of-a-previous-incarnation" }, wantCode: codes.FailedPrecondition},
+		// The stale uid is well-formed on purpose: a malformed one is rejected
+		// as INVALID_ARGUMENT by request validation before the guard runs.
+		"Stale": {requestUID: func(string) string { return "9d1f7b06-3c58-4a2e-8b40-5f7c1e9a2d63" }, wantCode: codes.FailedPrecondition},
```
LiorLieberman: what is this string?

LiorLieberman: this uid should come from somewhere?

LiorLieberman (commented):
LGTM, one question

ahmedtd:
FYI I am working on a PR to move MintJWT / MintCert into the main API, and remove the dedicated ActorIdentity API.  It seems clear from the authorization discussion that we will be able to handle any needed authorization using the overall authz framework we choose.

cmd/ateapi/internal/actoridentity/actoridentity_test.go
ahmedtd: This is a random UID that is assumed not to collide with the UID randomly picked by the store when it created the actor.  I don't think we need the agent comment.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1292,24 +1305,35 @@ message MintCertRequest {
...
   // Actor incarnation expected by the activation. This is only a stale-request
   // guard: ateapi derives the actor and its identity from the worker assignment.
+  //
+  // +k8s:required
+  // +k8s:format=k8s-uuid
   string expected_actor_uid = 3;
 
+  // +k8s:required
+  // +k8s:minimum=1
+  // +k8s:maximum=1 # keep this in sync with the ActorCertificatePurpose enum
```
ahmedtd, line 1495: I am not sure we need to try to guard the enum values using k8s validation logic.  Our implementation has to throw an error on an unrecognized purpose, and the error message will be better from there.

ahmedtd (dismissed):

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: Yeah
Actually, there is `+k8s:enum` in DV - https://kubernetes.io/docs/reference/using-api/declarative-validation/#tag-enum  (stability level:beta though)
But, this was the kind of practice everywhere - https://github.com/agent-substrate/substrate/blob/main/pkg/proto/ateapipb/ateapi.proto#L443-L449, hence followed it for uniformity. 

cc: @thockin

shrutiyam-glitch: sorry, nvm. 
looks like `+k8s:enum` only works on string types. The proto enums compile to `int32`, so the tag can't be used on ours.

cmd/ateapi/internal/actoridentity/actoridentity_test.go
shrutiyam-glitch: done.

pkg/proto/ateapipb/ateapi.proto
thockin: What the DV error gives you is a consistent path to the field in question.

What it does not give you is a good symbolic name for the enum, but since we have an out-of-bounds value anyway, that's sort of moot.

I agree that "Invalid value: purpose: must be less than or equal to 1" isn't awesome for a proto enum.  What error would we like it to throw?

"Unsupported value: purpose: 6: must be one of [ ACTOR_CERTIFICATE_PURPOSE_ATUNNEL ]" isn't awesome either (especially for enums with lots of values).

We could handle all enums with a custom override and produce any error we want - what's a good error?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1438,10 +1438,23 @@ service ActorIdentity {
 }
 
 message MintJWTRequest {
+  // The audiences the minted JWT is bound to. Tokens are only issued with
+  // audience bindings, so at least one is required.
+  //
+  // +k8s:required
+  // +k8s:listType=atomic
```
thockin: I'm assuming you want unique values here, so `listType=set` seems better?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1438,10 +1438,23 @@ service ActorIdentity {
...
   repeated string audience = 1;
 
+  // +k8s:required
+  // +k8s:format=k8s-short-name
   string atespace = 2;
```
thockin, line 1430: Should these 2 fields become a single ObjectRef ?

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: yeah, makes sense. 
But I am not sure what the `maxItems` and `maxLength` should be ? would this work - 
```  
// +k8s:maxItems=16 // upper bound for number of audiences (?) 
  // +k8s:listType=set
  // +k8s:eachVal=+k8s:maxLength=512 # since audiences are caller-defined URIs
```

pkg/proto/ateapipb/ateapi.proto
ahmedtd: I don't think we need to solve this problem in declarative validation for now.

pkg/proto/ateapipb/ateapi.proto
ahmedtd: 16 audiences is a reasonable upper bound.

pkg/proto/ateapipb/ateapi.proto
ahmedtd: Yes, but I'm handling that in my PR to move these RPCs into the main API.

ahmedtd (dismissed):

ahmedtd (approved):

PR #1265, opened 2026-08-27
Flesh out agentgateway implementation

keithmattix:
Flesh out the agentgateway implementation as a router and egress PEP. This also changes the atunnel CONNECT port to do 8443. Most of this is agentgateway contained, but the one other system we touch is localca; agentgateway's MITM support requires the TLS cert and key in the secret instead of the JSON pool (we don't rely on sdsmint).

- [X] Tests pass

### DIFF
--- a/internal/localca/localca.go
+++ b/internal/localca/localca.go
@@ -39,6 +39,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"os"
 	"sync"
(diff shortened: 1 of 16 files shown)


EItanya (commented):
~Can you add more info to the PR description explaining why you needed to add the extra methods to `localca`, and why they were necessary for agentgateway~

Nvm it's there

internal/localca/localca.go
```diff
@@ -215,6 +216,29 @@ type CA struct {
 	RootCertificate *x509.Certificate
 }
 
+// TLSCertificateChainPEM returns the CA certificate in the PEM encoding used
+// by TLS servers.
+func (ca *CA) TLSCertificateChainPEM() ([]byte, error) {
+	if ca == nil || ca.RootCertificate == nil {
+		return nil, fmt.Errorf("ca certificate: is nil")
+	}
```
EItanya: Do we really need this check, can this ever happen?

internal/localca/localca.go
```diff
@@ -215,6 +216,29 @@ type CA struct {
...
+	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: ca.RootCertificate.Raw}), nil
+}
+
+// TLSPrivateKeyPEM returns the CA signing key in the PKCS#8 PEM encoding used
+// by TLS servers.
+func (ca *CA) TLSPrivateKeyPEM() ([]byte, error) {
+	if ca == nil || ca.SigningKey == nil {
+		return nil, fmt.Errorf("ca key: is nil")
+	}
```
EItanya: Same as above

internal/localca/localca.go
keithmattix: I think so; we directly construct CA's a couple of times: https://github.com/agent-substrate/substrate/blob/54264bb49551351ba798331135e4398d19421f57/internal/localca/localca.go#L298. If someone forgets to assign RootCertificate or something, we can find ourselves calling this function without one.

internal/localca/localca.go
keithmattix: Same for me actually :) There's no `New` function for CAs; we construct them directly so I feel like we should guard against nils

EItanya: Do we need to guard against `ca == nil` though. That's classic proto getter behavior, would require passing in a `nil` ca object, not just constructing it wrong.

i won't push too hard, I've just noticed that agents LOVE these kind of checks

keithmattix: Eh, fair enough; I'll just keep the RootCertificate check

EItanya (approved):

PR #1245, opened 2026-08-26
e2e: merge the networking e2e test fixtures into one binary called testserver

yufan-su:
The egress e2e suites stood up in-cluster helper pods from two separately-built ko images: `grpcecho` (a gRPC echo origin) and `egressprobe` (an HTTP binary that both drives the gateway via `/handshake` and doubles as a plain HTTP target via `/healthz`). This collapses both into a single `internal/e2e/fixtures/testserver` binary that dispatches on a subcommand:

 - `testserver grpc` — cleartext-h2 gRPC echo origin (was grpcecho)
 - `testserver http` — plain HTTP origin serving /healthz (the httpTarget role)
 - `testserver probe` — the gateway client-driver (was egressprobe)

Each pod runs one subcommand on one listener, so per-pod wire behavior is unchanged.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

#1017

PR #1271, opened 2026-08-27
test: Verify atenet-router Layer 7 WebSocket upgrades

linde:
Introduces an end-to-end networking test to verify that the atenet Envoy ingress router successfully honors and proxies HTTP `Upgrade: websocket` requests to isolated Actors.

This adds a minimal ping-pong Go fixture which is materialized into the test cluster as an ActorTemplate. We also export the `BaseURL()` getter on the e2e test `RouterClient`. Exporting this is necessary to allow the `gorilla/websocket` client to construct its own L7 HTTP dial string, rather than relying on the client's existing `Connect` method, which would establish a blind L4 TCP tunnel and completely bypass the Envoy HTTP routing logic we aim to test.

Coverage for #1201

- [x] Tests pass


internal/e2e/fixtures/websocket/main.go
haiyanmeng, line 1: @linde , thanks for the e2e test coverage followup.

We are tying to minimize the number of the fixtures. @yufan-su submitted https://github.com/agent-substrate/substrate/pull/1245 earlier today to merge the networking e2e test fixtures into one binary called `testserver`: https://github.com/agent-substrate/substrate/tree/main/internal/e2e/fixtures/testserver. Can you try to extend `testserver` to support websocket?

linde: yes, that's a great call. will do

linde: done, PTAL

internal/e2e/suites/networking/websocket_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	actorRef := resources.ActorRef{Atespace: networkingAtespace, Name: actorName}
+	header := http.Header{}
+	header.Set("Host", resources.ActorDNSName(actorRef))
+
+	var c *websocket.Conn
+
+	dialer := websocket.DefaultDialer
+	dialer.Proxy = nil // Explicitly circumvent HTTP_PROXY so Cloudtop BeyondCorp doesn't block ws://127.0.0.1 dialing
```
yufan-su: Is this needed for running e2e on github action? for line 92

haiyanmeng: I think we can remove `dialer`. It will look like:
```
      var c *websocket.Conn

      // Ride out atenet-router xDS snapshot sync lag (up to ~30s), similar to waitForRouteReady
      deadline := time.Now().Add(30 * time.Second)
      for {
              var resp *http.Response
              c, resp, err = websocket.DefaultDialer.DialContext(ctx, u.String(), header)
```

internal/e2e/fixtures/testserver/main.go
haiyanmeng, line 24: Let us include a line for `websocket` here

internal/e2e/fixtures/testserver/websocket.yaml.tmpl
haiyanmeng, line 1: Let us register this template in https://github.com/agent-substrate/substrate/blob/69428103ee1ae5ad9db151f80ede8a6ee7b6bf52/internal/e2e/sandbox_test.go#L27-L31

internal/e2e/suites/networking/websocketingress_test.go
haiyanmeng, line 1: Let us rename this file to `websocketingress_test.go` since we plan to add websocket egress test coverage soon.

internal/e2e/suites/networking/websocketingress_test.go
```diff
@@ -0,0 +1,131 @@
...
+		if e2e.KubeContext != "" {
+			delArgs = append([]string{"--context=" + e2e.KubeContext}, delArgs...)
+		}
+		e2e.RunCmd(t, "kubectl", delArgs...)
+	})
+
+	return namespace
+}
+
+func TestWebsocketPingPong(t *testing.T) {
```
haiyanmeng: Rename the test to `TestWebsocketPingPongIngress` since we plan to add websocket egress test soon.

internal/e2e/suites/networking/websocket_test.go
linde: this is a good catch, thanks for flagging @yufan-su  and for the suggestion @haiyanmeng. done in b9dad6e

internal/e2e/fixtures/testserver/main.go
linde: ty, also done in b9dad6e

internal/e2e/suites/networking/websocketingress_test.go
linde: thanks, covered in 1bcd134

internal/e2e/suites/networking/websocketingress_test.go
linde: 👍  in 3760cc4

internal/e2e/fixtures/testserver/websocket.yaml.tmpl
linde: good call, done in 2ff2f65

haiyanmeng (approved):

PR #1275, opened 2026-08-27
[WIP] hack: add opt-in NAT64 support to IPv6-only kind clusters

ygao-g:
> **⚠️ Depends on #958:** This PR is built on top of #958. Please review and merge that PR first.

## Description

GitHub-hosted runners only support IPv4. When running an IPv6-only kind cluster on them, external endpoints are unreachable. This PR adds opt-in NAT64/DNS64 support to translate traffic and allow these clusters to reach the outside world.

## Key Changes

* **Opt-in NAT64:** Added `IPV6_DNS64_PREFIX` to `create-kind-cluster.sh`.
* **Agent Deployment:** Vendored and deploys the `kubernetes-sigs/nat64` agent, and configures CoreDNS for DNS64 synthesis.
* **CIDR Adjustments:** Narrows the node `podSubnet` to `/120` when enabled, due to the NAT64 agent's IPv4 pool limits.
* **Automated Probing:** Runs a startup probe to fetch an external endpoint, ensuring translation works before proceeding.

## Testing Performed

- Ran the IPv6-only kind e2e workflow on `amd64` GitHub runners (no IPv6 egress).
- Minimal stack (`prs: 958 1275`, i.e. just this PR on its base):
  https://github.com/ygao-g/substrate/actions/runs/33929661265
  Cluster bring-up with the connectivity probe, DNS64 synthesis of
  `storage.googleapis.com` into `64:ff9b::/96`, the full Agent Substrate
  install, and the gVisor runtime download over the IPv6-only pod network
  are all green. The run ends red at the micro-VM demo on a pre-existing
  IPv6 parity gap (unbracketed ClusterIP in the kata-asset endpoint),
  unrelated to this change.
- Full IPv6 stack (`prs: 958 753 979 1275 1122`):
  https://github.com/ygao-g/substrate/actions/runs/33892185054
  41 passed / 0 failed / 0 skipped on both legs, including actor egress
  through `64:ff9b::`.

Both runs used commit 872a8fb5; the current head differs only by
commit-message amends (same tree).


BenTheElder:
why do we need egress off of the host machine from the test cluster at all?

BenTheElder:
we push images into the registry / cluster

ygao-g:
> why do we need egress off of the host machine from the test cluster at all? 

@BenTheElder  It's the actor-egress e2e tests, which fetch real external origins:

- [TestActorEgress](https://github.com/agent-substrate/substrate/blob/main/internal/e2e/suites/networking/networking_test.go#L98) fetches http://example.com/
- TestActorEgressHTTPS fetches https://example.com/
- egressmitm uses the same origin

ygao-g:
> > why do we need egress off of the host machine from the test cluster at all?
> 
> Benjamin Elder (@BenTheElder) It's the actor-egress e2e tests, which fetch real external origins:
> 
> * [TestActorEgress](https://github.com/agent-substrate/substrate/blob/main/internal/e2e/suites/networking/networking_test.go#L98) fetches http://example.com/
> * TestActorEgressHTTPS fetches https://example.com/
> * egressmitm uses the same origin

Updates on this:
- Discuss with others and have a pending item to use an in-cluster origin instead of example.com in the egress tests;
- However, there is still traffic to external domain, as atelet that will fetch the runtime from gs://gviros/...

@BenTheElder Does this answer your questions?

PR #1276, opened 2026-08-27
ateapi: document the container security context and image volume fields

BenTheElder:
The Capabilities, SecurityContext, ImageVolumeSource and Volume.image fields landed without doc comments, so apitool's documented rule fails on them and they are not in the exemption backlog. Document them rather than growing the backlog.

Fixes failure on main

- [x] Tests pass


juli4n (commented):
Can you also regenerate the python code for the protos?

juli4n (approved):

PR #1280, opened 2026-08-28
benchmarking: re-dirty the glutton working set each cycle (--mem-churn)

Oneimu:
#### What this pr do

Re-dirties part of the glutton working set on every benchmark cycle, so repeated suspends snapshot an actor whose memory is changing — like a live application's — instead of a set that was filled once and never touched again.

Each iteration (after the one-time fill), the GluttonUser sends one `WriteRAM` request with `WRITE_MODE_OVERWRITE`, re-randomizing the first `mem_churn` bytes of the working set in place. Overwrite mode already existed on the glutton API; this PR adds no proto changes and no glutton server changes — it is driver + config wiring only.

This is the follow-up from #1130's: the original self-driving memload continuously re-dirtied its pages, and that property was dropped in the move to API-driven fill. This restores it in the API-driven shape with a dialable amount instead of the old all-or-nothing full pass.

#### Why it matters

A fill-once working set is static: every suspend after the first packs up identical memory. If snapshotting ever gains incremental / dirty-page optimizations, a static benchmark would measure almost nothing from cycle two onward — and would score "upload nothing" as an infinite win. With churn, every cycle carries a known amount of freshly-dirtied pages, and the knob is sweepable (64Mi, 256Mi, …) so a future differential-snapshot optimization can be demonstrated as "upload cost scales with churn size, not total size."

### DIFF
--- a/internal/benchmarking/boomer/glutton/lifecycle.go
+++ b/internal/benchmarking/boomer/glutton/lifecycle.go
@@ -111,6 +111,9 @@ func (r *taskRuntime) iterate() {
 	// carries the full working set; glutton keeps the allocations across
 	// suspend/resume, so this runs once per actor (retried if it fails).
 	user.ensureRAMFilled(ctx)
+	// Re-dirty part of the working set each cycle so repeated suspends
+	// snapshot an actor whose memory is changing, like a live application's.
+	user.churnRAM(ctx)
 	user.ping(ctx)
 	user.suspend(ctx)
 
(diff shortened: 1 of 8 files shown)


internal/benchmarking/boomer/glutton/lifecycle.go
```diff
@@ -365,14 +368,41 @@ func (u *gluttonUser) ensureRAMFilled(ctx context.Context) {
...
+	churn := u.cfg.Dyn.Load().MemChurn
+	if churn == "" || !u.ramFilled {
+		return
+	}
+
+	ctx, span := u.cfg.Tracer.Start(ctx, "GluttonChurnRAM")
+	defer span.End()
+	start := time.Now()
+
+	err := u.writeRAM(ctx, "memload", churn, gluttonpb.WriteMode_WRITE_MODE_OVERWRITE)
```
maxsmythe, line 388: Do we care about offsets here? We would always be over-writing the same part of the RAM array otherwise.

Not 100% sure that matters for the intent here, but wanted to raise the thought.

Oneimu: This is intentional for this PR to keep the re-dirty baseline simple. I'm exploring more realistic, continuous read/write patterns across the working set and will tackle offsets in a follow-up PR.

maxsmythe (approved):
LGTM

Issue #777, opened 2026-08-06
Per-snapshot authorization


mtaufen:
We don't have very granular authz for which snapshots atelet is allowed to read. Ideally, an atelet can only read snapshots for Actors that are actively scheduled to it. Several ideas have been discussed, such as GCP IAM conditions or Credential Access Boundaries. However, these are GCP-specific features, and it would be better to do something less cloud specific. GCS, S3, and Azure Storage all support variants of signed URLs, for example.

Potential option:
- On Create (for reading golden snapshot), Suspend, and Resume: ateapi requests a signed URL from the appropriate storage provider and passes it to atelet. 
- If this is a performance bottleneck, consider caching pre-signed URLs in ateapi memory to speed it up (we probably can't cache ALL of them, TBD). 

This can help mitigate the following threats in the [threat model](https://github.com/agent-substrate/substrate/blob/main/docs/threat-model.md):
- "Attacker who can create ActorTemplates can read or write any storage buckets atelet has access to."
- "A compromised node accesses all snapshots for the cluster."

PR #1282, opened 2026-08-28
feat: pluggable snapshot storage broker with per-snapshot signed URLs

aramase:
Adds a vendor-neutral snapshot storage broker. The control plane mints
short-lived, per-snapshot signed URLs and atelet reads and writes snapshots over
plain HTTP, so the node links no cloud SDK and holds no cloud credential. The
cloud-specific signer runs out of process as a sidecar over a Unix socket, which
keeps the control-plane core free of any cloud SDK as well. Ships an S3 sidecar.
Off by default.

This is the storage-decoupling touchpoint from the Cloud Provider Decoupling
design, and it stands up the same per-snapshot capability #777 needs: the node
stops holding a broad, permanent storage credential. It does not yet make the
authorization decision before minting; that needs the broader control-plane
authz layer.

Design doc: https://docs.google.com/document/d/1I2SlYey0Y2mcGQBZNpfYufWhwLbQDmaqI5T56sz33_4/edit?usp=sharing

## What's in it
- `storagebroker` interface and registry in ate-api-server. `New(backend)`
  selects one; empty disables signing and the existing in-atelet path is
  unchanged.
- ate-api-server pre-mints per-snapshot capabilities at resume (actor plus
  golden) and suspend (write), passed to atelet via a new `signed_access` map on
  the atelet proto.
- atelet `signedObjectStore` satisfies the existing `ategcs.ObjectStorage` and
  speaks S3 (presigned GET, POST starts-with policy) or Azure (PUT plus token)
  over net/http, with no SDK and no credential.
- UDS broker client so the signer is a sidecar. ate-api-server links 0 AWS SDK
  packages (was 39 with the in-tree signer).
- `cmd/s3-broker-server`: the S3 sidecar (presigned GET per object, one POST
  policy per snapshot).
- Selected via `ATE_SNAPSHOT_BROKER` (for example `uds`), separate from atelet's
  `ATE_STORAGE_BACKEND`.

## Testing
- Unit tests for the mint wiring and the signed HTTP read/write dispatch. Full
  `verify-all` is green.
- Broker-on kind e2e (parking suspend/resume) exercises the sidecar mint and
  atelet presigned GET/POST end to end.
- Benchmarks show the mint is off the byte path: resume p50 is 244 ms native vs
  253 ms in-tree signer vs 243 ms over the socket, so no measurable latency.

## Not in this PR
- The authorization decision before minting (the decision half of #777).
- A GCS broker, and retiring atelet's existing S3 object-store client to take the
  node to zero SDK.
- Chunked upload for very large snapshots.
- Wiring the sidecar into the in-repo manifests so CI exercises the broker path.

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_suspend.go
+++ b/cmd/ateapi/internal/controlapi/workflow_suspend.go
@@ -259,6 +259,10 @@ func (w *ActorWorkflow) ensureAteletSuspended(ctx context.Context, actorRef reso
 		Scope:    actorSnapshotContentScopeToAtelet(commitSnapshotScope(actor.GetMetadata().GetAtespace(), actorTemplate)),
 		ActorUid: actor.GetMetadata().Uid,
 	}
+	// The node writes the actor's snapshot; mint a write capability for it.
+	if req.SignedAccess, err = w.writeAccessFor(ctx, snapshotURI.String()); err != nil {
+		return "", err
+	}
 	wireSnapshotScope = ateattr.SnapshotScopeValue(req.Scope)
 
 	_, err = client.Checkpoint(ctx, req)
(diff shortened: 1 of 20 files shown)


cmd/ateapi/internal/controlapi/workflow_suspend.go
```diff
@@ -311,6 +315,11 @@ func (w *ActorWorkflow) ensurePausedSnapshotUploaded(ctx context.Context, actorR
 		// from the captured scope in the snapshot's manifest where possible.
 		DesiredScope: actorSnapshotContentScopeToAtelet(commitSnapshotScope(actor.GetMetadata().GetAtespace(), actorTemplate)),
 	}
+	// The node writes the paused snapshot to its destination; mint a write
+	// capability for it.
+	if req.SignedAccess, err = w.writeAccessFor(ctx, snapshotURI.String()); err != nil {
+		return "", err
+	}
```
swiftdiaries, line 322: niche edge case from Codex review:
for the case when there's a paused upload that has actually completed but doesn't have a local checkpoint + when the response is lost, should we add access to read the completion manifest as well (with write)?
that way, we check the completion manifest and return success without an upload.

PR #1250, opened 2026-08-26
extend declarative validation to Workers

shrutiyam-glitch:
Follow up on issue #1168  and the base PR #1215
 
**Overview:** 
Continues the declarative validation (DV) migration from #1215 by converting the `Worker` resource. This moves immutable-field enforcement from the storage layer up to the service layer. With `WorkerPoolSyncer` now using RPCs, nearly all write paths share the exact same validation.

**Key Changes:**
*   **Schema (`ateapi.proto`):** Added full DV tags (required, format, immutable) to `Worker` pod-coordinates and metadata. `WorkerStatus.state` is now strictly bounded to its enum range.
*   **Service Layer:** Replaced ~60 lines of hand-written validation with 1-line generated calls. `CreateWorker` and `UpdateWorker` now scrub server-owned fields and enforce immutability *before* hitting the store.
*   **Storage Layer:** Removed `store.CheckWorkerMutation`. `atepg.UpdateWorker` is optimized to only clone metadata instead of the whole worker. Immutability contract tests were moved to the service layer.
*   **Behavior Tweaks (Tightenings):** 
    *   `page_token` is now capped at 256 chars.
    *   `DeleteOptions.uid` must be a valid UUID.
    *   Global-ref atespace violations now correctly return `Forbidden` instead of `Invalid`.
    *   Immutability checks now natively cover the `WorkerPoolSyncer` path.
*   **Testing:** Added comprehensive positive/negative cases for all newly tagged fields, boundary cases, and moved the immutability test suite. All suites pass cleanly.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

Issue #1164, opened 2026-08-24
Stable public API for M2


juli4n:
This is a parent issue / epic to track all the work to stabilize the API for our milestone 2.

PR #1283, opened 2026-08-28
multi-actor worker API

BenTheElder:
Part of #1266 

This is a draft of the core API + data store changes.

It's still a large PR, apologies.

The "as rows" commit could be split out, but this takes it to ~all of the breaking changes we can't hide behind updating internals.

Same for the claimlock, but in both cases it seems these are worth understanding when considering the API shape. 

They're loadbearing for performance once we actually have multi-actor workers.


BenTheElder:
x-ref: https://github.com/agent-substrate/substrate/issues/1164 🙈

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1321,13 +1321,16 @@ message Worker {
   string sandbox_class = 8;
   map<string, string> labels = 9;
 
-  // The compute capacity this worker can give an actor sandbox. Immutable, set
-  // at creation: a worker pod's limits are fixed for its lifetime.
+  // What this worker has to give the Actors it hosts. Mutable: a pod can be
```
thockin: This comment is a little jumbled.  Prefer not to talk about Pods here, but frame it in terms of the worker.

What does `ReportWorkerCapacity` do that this field doesn't?

"would ask to lose one" - one what?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1342,22 +1345,45 @@ enum WorkerState {
 message WorkerStatus {
   WorkerState state = 1;
 
-  // The Actor currently bound to this Worker, if any.
-  ActorAssignment assignment = 2;
+  // The Actors bound to this Worker, as many as capacity admits.
+  //
+  // Populated only by GetWorker; ListWorkers leaves it empty and reports
```
thockin: @juli4n @EItanya @bowei 

In some ways this is very similar to the EgressPolicy "subresource" and I think we should think hard about the semantics we want to offer.

IIUC What you are saying Ben is that the "Worker" resource has this field, which could be quite large.  GetWorker(w) will populate it but ListWorkers() will not.  That's a net-new behavior (partial reads).  The EgressPolicy equivalent would be GetActor(a) will return policies and ListActors() will not.

For Egress we are going (trying) a different approach.  The Actor resource never includes policies, but a purpose-built set of RPCs lets us model them as subordinate to the Actor - GetActorEgressPolicy(a, name) / ListActorEgressPolicies(a) is how we read them.

So the analog here would be:
* GetWorker(w) and ListWorkers() return the worker resource
* ListWorkerAssignments(w) returns all the assignments for a single worker
* CreateWorkerAssignment(w, ...) makes a new assignment - similar to a full resource but subordinate to Worker

One aspect that may be significantly different is the need for atomicity WRT resource accounting?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1342,22 +1345,45 @@ enum WorkerState {
...
+  // occupancy through allocated, so a listing costs the fleet's size and not
+  // its actor count.
+  //
+  // Reuses field 2, which held the single ActorAssignment this replaces: the
```
thockin: You don't need this comment :)

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1342,22 +1345,45 @@ enum WorkerState {
...
+  // two are wire-compatible, so an old record reads back as a set of one.
+  repeated ActorAssignment assignments = 2;
+
+  // What the assignments have taken from capacity, as a running total moved
```
thockin: "taken" -> "consumed" or "claimed" all over?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1342,22 +1345,45 @@ enum WorkerState {
...
+  WorkerCapacity allocated = 3;
 }
 
-// WorkerCapacity is the worker pod's compute capacity available to host an
-// actor sandbox, taken from the ateom container's resource limits. The
-// scheduler only places an actor on a worker whose capacity is >= the actor's
-// declared resource limits. An unset message, or a zero field within it, means
-// "unknown/unset" for that dimension: treated as unconstrained so placement is
-// not blocked (matching the pre-capacity behavior).
+// WorkerCapacity is what a worker pod has to give the Actors it hosts, and also
```
thockin: This comment is also unclear.  I think you should clarify a resource model but "give/take" wording is sort of clunky.

IIUC this is the total capacity of the worker, independent of what may currently be allocated to actors.  Is that right?  So "available" is "worker.capacity - worker.status.allocated". 

How would we model overcommit?  Worker lies about capacity?  Maybe add an overcommit scalar (e.g. 1.5)?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1342,22 +1345,45 @@ enum WorkerState {
...
   int64 cpu_milli = 1;    // CPU capacity in millicores (1000 = one core).
   int64 memory_bytes = 2; // Memory capacity in bytes.
+
+  // How many Actors may be bound at once. A capacity dimension rather than a
+  // separate limit because it bounds the same thing — what one more Actor costs
+  // — for the costs that are not CPU or memory: netns, mounts, file
+  // descriptors, blast radius. Kubernetes bounds a node the same way, with
+  // allocatable pods beside allocatable cpu and memory.
+  //
+  // Unset means ONE, not unconstrained: a Worker that has not said it can hold
```
thockin: @juli4n should we run with a model where unspecified fields have a value, but clients need to do the logic:
```
n := func()int { if cap.count == 0 { return 1}; return cap.count }()
```

...or should we do a more principled form of explicit default reification (in the RPC handler)?  The main advantage (other than less LOC in clients) is that we can evolve defaults over time without breaking stored objects.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1370,6 +1396,13 @@ message ActorAssignment {
   string actor_uid = 3;
   // +k8s:opaqueType
   ObjectRef actor_template_ref = 4;
+
+  // What the Worker admitted this Actor for, from its declared limits at
+  // placement time. Recorded here rather than re-read from the Actor so that
+  // releasing gives back exactly what binding took, even if the template was
+  // edited in between. The actors dimension is unset; one assignment is one
+  // Actor.
+  WorkerCapacity resources = 5;
```
thockin: This is getting into the bad parts of reusing schema.  What if we modelled it as

```
WorkerCapacity {
    Resources resources
    Actors int
}
```

and then here you would use `Resources` ?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1413,6 +1446,33 @@ service ActorIdentity {
...
   rpc MintCert(MintCertRequest) returns (MintCertResponse);
+
+  // ReportWorkerCapacity records what a Worker can hold, as the Worker reports
+  // it. The actor ceiling is the ateom's — it owns the slot allocator, and only
+  // its node can observe the number — so a fleet may run mixed ateom versions.
+  //
+  // atelet calls this for the Worker with its own client certificate, as it
+  // does for MintCert; an atelet may speak only for the Workers on its node.
+  // Idempotent: re-reporting the same capacity is not an update.
+  rpc ReportWorkerCapacity(ReportWorkerCapacityRequest) returns (ReportWorkerCapacityResponse);
```
thockin: I don't understand why this is part of the `ActorIdentity` service?

Why is this not part of the CreateWorker() flow?

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1418,6 +1420,239 @@ func (p *Persistence) DeleteWorker(ctx context.Context, name string, pre store.D
...
+
+// saveWorkerAfterAssignment writes back a worker whose allocation just moved,
+// at the next version.
+func saveWorkerAfterAssignment(ctx context.Context, tx pgx.Tx, worker *ateapipb.Worker) error {
+	worker.Metadata = newUpdateMetadata(worker.GetMetadata())
+	protoBytes, err := proto.Marshal(worker)
+	if err != nil {
+		return fmt.Errorf("marshaling worker: %w", err)
+	}
+	_, err = tx.Exec(ctx, `UPDATE workers SET version = $1, proto = $2 WHERE name = $3`,
```
thockin: Shouldn't you do precondition here anyway? (where UID=... AND VERSION=...)

thockin (commented):
I focused on API stuff.  The atomic update of an assignment and the worker seems like something that will hurt us later, but maybe workers just not updated often enough for that to matter?

It's tricky logic and could do with more explanation (and specifically the "not updated often enough to matter" rational).

I did not focus on validation yet

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -525,21 +537,23 @@ func (w *ActorWorkflow) assignWorkerAttempt(ctx context.Context, actorRef resour
...
 		if errors.Is(err, store.ErrNotFound) {
 			w.workerCache.Forget(assignedWorker.GetMetadata().GetName())
 			return nil, nil, fmt.Errorf("selected worker disappeared before claim: %w", store.ErrVersionConflict)
 		}
 		return nil, nil, err
 	}
-	assignedWorker = stored
 
 	newAssignment := workerAssignmentFrom(assignedWorker)
 	storedActor, err := w.store.UpdateActor(ctx, actorRef, store.PreconditionFrom(actor), func(toUpdate *ateapipb.Actor) error {
```
jpbetz: This is the one I most want to see change before the store part merges. The claim and the actor's RESUMING+assignment are still two separate transactions — `BindActorToWorker` commits inside its own `writeAndAppendEventFor` tx, `UpdateActor` is a separate CAS — so an ateapi crash, a lost connection, or a context deadline in between leaves a `worker_assignments` row, and the allocation it books, for an actor that is still SUSPENDED.

To be fair, this PR makes it less bad rather than worse: the leak now costs one slot out of N instead of the whole worker, and `workerHoldingStaleClaim` recovering off `FindWorkerHostingActor` is a real improvement over the old cache scan. But it only runs on the next ResumeActor for that specific actor; otherwise the row sits until the worker itself is deleted. And unless I'm misreading, it survives the actor too: `releaseWorker` keys off `actor.Status.WorkerAssignment`, which is nil in exactly this case, so suspend/delete/crash all skip the release, and `worker_assignments` has an FK to workers but not to actors — delete the orphaned actor and the booking outlives it.

What I'd like is one store primitive that does the whole thing: workers FOR UPDATE, the assignment INSERT, an actors CAS on uid+version, the outbox INSERT, COMMIT. The awkward part used to be that the store had no way to run a caller's decision inside the transaction, and this PR just built exactly that (`writeAndAppendEventFor` plus `admit`), so what's left is mostly moving the actor precondition and mutation across the signature. The same shape covers unbind for crashActor / FinalizePaused / FinalizeSuspended / the delete path. This is #607, and I think Tim's "the atomic update of an assignment and the worker seems like something that will hurt us later" is the same instinct. I don't want to carry it into v1 with the data model changing underneath it; if it doesn't fit in the split-out store PR, can we at least name the primitive on #607 so it lands right behind?

cmd/ateapi/internal/store/atepg/schema.go
```diff
@@ -102,6 +102,20 @@ CREATE TABLE IF NOT EXISTS workers (
     proto    bytea NOT NULL
 );
 
+-- One row per Actor, keyed by Actor UID because an Actor has at most one Worker.
+-- Kept separate from workers so Worker reads, writes, and watch events do not
+-- grow with occupancy. The primary key finds an Actor's Worker; worker_name
+-- lists a Worker's Actors.
+CREATE TABLE IF NOT EXISTS worker_assignments (
```
jpbetz: This creates the table, but nothing moves existing `status.assignment` values into it and nothing seeds `status.allocated` — `applySchema` is still `CREATE TABLE IF NOT EXISTS` only, and the only `INSERT INTO worker_assignments` in the package is the one inside `BindActorToWorker`. Unless I'm misreading, on an upgraded cluster a worker that is currently hosting an actor reads as empty everywhere it matters:

- Placement: `HasRoom` does `used.GetActors() >= WorkerMaxActors(capacity)` with `Allocated` nil, so `0 >= 1` is false and a second actor gets placed on it. The new `admit` callback doesn't rescue this — it re-asks `HasRoom` against the locked row, but that row's `Allocated` is the same nil.
- Resume: `validateAssignedWorker` -> `workerHostsActor` finds no `worker_assignments` row and crashes the legacy actor with `ReasonWorkerReassigned`.
- Delete (I think): `ensureAteletTerminated` asks `workerHostsActor`, gets false, and skips the atelet terminate RPC, so deleting a legacy actor leaves its sandbox running.

And the proto comment at ateapi.proto:1355 ("an old record reads back as a set of one") isn't true in effect: controlapi's `GetWorker` overwrites `Status.Assignments` from `ListWorkerAssignments`, so the legacy value never reaches a reader, and `store.ClearAssignments` runs on every `CreateWorker`/`UpdateWorker` (atepg.go:1285, :1379), so the first unrelated worker write drops it from the proto too.

I think we need either a one-shot backfill here, under the same advisory lock `applySchema` already holds (an `INSERT ... SELECT` over the unmarshaled protos, or a small Go loop), or an explicit "upgrading requires a full drain" statement in the PR description and docs — and the proto comment corrected either way. A test that seeds a legacy-shaped worker row and pins what Bind/GetWorker/HasRoom do with it would lock in whichever we pick (it can't be seeded through the interface since Create/Update both clear the field, so probably an atepg-level test with raw SQL). Ordering note: if #1196 lands first this becomes a numbered migration plus a backfill step rather than DDL in this file.

cmd/ateapi/internal/store/store.go
```diff
@@ -222,11 +221,43 @@ type Interface interface {
...
+	// Actor already on this Worker is already counted against it.
+	//
+	// ErrNotFound if the Worker is gone.
+	BindActorToWorker(ctx context.Context, workerName string, assignment *ateapipb.ActorAssignment, admit func(*ateapipb.Worker) error) error
+
+	// ReleaseActorFromWorker removes an assignment and updates allocation,
+	// returning the Worker as it now stands so the caller can feed the
+	// watch-fed cache, which until then reports it full. It returns nil if the
+	// assignment was already absent.
+	ReleaseActorFromWorker(ctx context.Context, workerName string, expectedVersion int64, actorUID string) (*ateapipb.Worker, error)
```
jpbetz: Now that Bind has dropped its version precondition, `expectedVersion` here is dead: `lockWorkerForAssignment` no longer takes a version and the DELETE is keyed on `(actor_uid, worker_name)`, so the atepg implementation never reads it — and it's still threaded through the interface, the `ServiceImpl` pass-through, workflow.go and crash.go. The one visible consequence is that workflow_pause.go:243 still maps `store.ErrVersionConflict` to `codes.Aborted` for a conflict this call can no longer produce.

The row lock plus the actor-scoped DELETE is the whole safety story here, so I'd drop the parameter rather than leave a precondition-shaped thing that isn't one. If we do want a precondition on release (Tim is asking for something like this on the write in `saveWorkerAfterAssignment`), let's make it a real `store.Precondition` checked against the locked row, not a bare int64 nobody looks at.

```suggestion
	ReleaseActorFromWorker(ctx context.Context, workerName string, actorUID string) (*ateapipb.Worker, error)
```

cmd/ateapi/internal/store/store.go
```diff
@@ -315,8 +348,8 @@ func CheckWorkerMutation(stored, mutated *ateapipb.Worker) error {
 			return fmt.Errorf("%w: %s changed from %q to %q", ErrImmutableField, f.name, f.stored, f.mutated)
 		}
 	}
-	if !proto.Equal(stored.GetCapacity(), mutated.GetCapacity()) {
-		return fmt.Errorf("%w: capacity changed from %v to %v", ErrImmutableField, stored.GetCapacity(), mutated.GetCapacity())
+	if stored.GetCapacity() != nil && mutated.GetCapacity() == nil {
```
jpbetz: nit, semantics rather than something I can trigger from anything in tree: this only rejects a nil capacity, so an update carrying an explicitly-constructed `&WorkerCapacity{}` gets through and lands the same state a clear would — `WorkerMaxActors` reads an all-zero message as 1 and cpu/memory read as unconstrained, exactly as nil does. I ran it: `CheckWorkerMutation` with stored actors=4000 and mutated `&WorkerCapacity{}` returns nil, while the nil case correctly returns `ErrImmutableField`. Nothing we ship produces that (`workerCapacity()` returns nil when both dimensions are zero, and `mergeReportedCapacity` only overwrites non-zero dimensions), so this is only about a client whose JSON hands us `"capacity": {}`.

If you think it's worth a line, careful with `proto.Equal` — `proto.Equal(nil, &WorkerCapacity{})` is false, so swapping to it would stop catching the nil case. An all-zero check covers both and keeps the `proto` import out:

```suggestion
	mutatedCapacity := mutated.GetCapacity()
	cleared := mutatedCapacity.GetActors() == 0 && mutatedCapacity.GetCpuMilli() == 0 && mutatedCapacity.GetMemoryBytes() == 0
	if stored.GetCapacity() != nil && cleared {
```

Ordering note either way: #1250 is moving this function's rules to the service layer as declarative validation and #1257 adds sandbox_class to it, so whichever of the three lands second re-expresses this.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1418,6 +1420,239 @@ func (p *Persistence) DeleteWorker(ctx context.Context, name string, pre store.D
...
+			actorUID, workerName, assignmentBytes)
+		if err != nil {
+			return nil, fmt.Errorf("binding actor %s to worker %s: %w", actorUID, workerName, err)
+		}
+		if tag.RowsAffected() == 1 {
+			// A new binding, so the Worker has to have room for it. Asked here
+			// rather than before the insert because only the insert says whether
+			// this is new, and asked under the row lock so the answer holds
+			// until commit. Refusing rolls the speculative insert back.
+			if admit != nil {
```
jpbetz: I like this — deciding admission against the locked row is strictly better than the version CAS it replaces, and `RefusedAdmissionLeavesNothingBehind` / `AdmitSeesConcurrentBinds` pin down the part that matters.

nit: the doc on `BindActorToWorker` says admit "runs against the Worker as it stands with its row locked", which is the half that makes it correct. What it doesn't say is the obligation that comes with it: admit runs inside the open transaction on a pooled connection, so anything slow in there parks a `workers` row lock and a pool connection for the duration. Today's only caller passes `Applies` + `HasRoom`, which are pure reads over the proto, so this is fine as written — but the signature doesn't stop the next caller from doing an RPC, a store read, or taking a mutex in there (the #940 "no Go work under the row lock" concern in a new place). Worth one more clause in the store.go doc saying admit must not block — no I/O, no store calls, no locks.

cmd/ateapi/internal/store/store.go
```diff
@@ -222,11 +221,43 @@ type Interface interface {
...
+	// Rebinding the same Actor replaces its assignment.
+	//
+	// admit decides whether the Worker will take the Actor, and runs against the
+	// Worker as it stands with its row locked, so the answer cannot go stale
+	// between the check and the bind. Returning an error from it refuses the
+	// bind and is returned unchanged. It is consulted only for a new binding: an
+	// Actor already on this Worker is already counted against it.
+	//
+	// ErrNotFound if the Worker is gone.
+	BindActorToWorker(ctx context.Context, workerName string, assignment *ateapipb.ActorAssignment, admit func(*ateapipb.Worker) error) error
```
jpbetz: Release comes back with the updated Worker specifically so the caller can hand it to the watch-fed cache before the change event lands — your comment right below says so — and all four release sites do that. Bind returns only an error, so after a successful claim the cache still shows the worker at its old allocation until the outbox event arrives, and the caller has no fresh copy to `Observe` (the `assignedWorker = stored` that at least gave it one went away with the rows commit).

The part that bugs me is the refusal path more than the success path: `assignWorkerAttempt` only touches the cache on ErrNotFound (`Forget`); `errWorkerFilledUp` falls straight through, so the retry re-runs scheduling against the same stale entry and can pick the same full worker again — which I think is what the retry budget had to grow for. Returning the post-bind Worker so the caller can `Observe()` it would make the cache converge at claim time instead of event-delivery time, and it makes Bind and Release symmetric. The atepg closure already has the worker in hand (it hands it to `writeAndAppendEventFor` for the event), so it's close to free.

Not a correctness problem — admit under the row lock is the real arbiter and I'm happy with that. But note the contention test won't show this either way: it has one worker with capacity set to exactly the actor count, so admit never refuses for want of room. A number would need a pool where a worker actually fills.

cmd/ateapi/internal/controlapi/workflow_worker_delete.go
```diff
@@ -68,29 +68,50 @@ func (w *WorkerWorkflow) loadWorkerForDelete(ctx context.Context, name string) (
...
+
+	assignments, err := w.store.ListWorkerAssignments(ctx, worker.GetMetadata().GetName())
+	if err != nil {
+		return fmt.Errorf("while listing the assignments of worker %s: %w", worker.GetMetadata().GetName(), err)
+	}
+	if len(assignments) == 0 {
+		markSkipped(ctx, "worker has no actors assigned")
+		return nil
+	}
+	for _, assignment := range assignments {
```
jpbetz: This is a serial loop over every assignment the worker holds, each iteration a GetActor and (usually) an UpdateActor, all inline in one DeleteWorker RPC. At one actor per worker that was two round trips; at the occupancy this PR is aiming for it's up to `capacity.actors` × 2 sequential round trips in a single call, and `releaseBoundActor` returns Aborted on the first actor-level version conflict, which stops the whole pass.

It does converge, to be fair: the release is recorded on the actor, so a retry skips already-released ones via the "actor no longer points at this worker" branch. But the assignment rows only go away with the FK cascade at the end, so every retry re-lists and re-GetActors all N from the top before it reaches the ones still bound — and in between, the Worker record still claims N assignments and N allocated actors that are actually CRASHED and unbound, which is what the scheduler and `workerHoldingStaleClaim` read.

Two things would help, neither necessarily in this PR: don't abort on one actor's conflict (collect it, keep going, fail at the end with "N still bound" so the retry is bounded by what's left), and drop each actor's `worker_assignments` row as it's released so a retry costs O(remaining) and progress is visible in the table — a plain row delete rather than `ReleaseActorFromWorker`, which relocks and rewrites the worker row and appends an outbox event per actor. Mostly I don't want it left implicit that DeleteWorker is now O(occupancy) in the request path.

cmd/ateapi/internal/store/storecontract/contract.go
```diff
@@ -1738,6 +1783,602 @@ func runWorkerContractTests(t *testing.T, setup func(t *testing.T) store.Interfa
...
+	})
+
+	// TestAllocatedTracksAssignmentsUnderChurn is the test a stored total needs,
+	// because the risk of storing one is not that the arithmetic is wrong once
+	// -- it is that some sequence of binds and releases leaves it disagreeing
+	// with what it summarizes, quietly, forever after.
+	//
+	// So: churn a worker through hundreds of binds, rebinds and releases, and
+	// after every one require the total to equal the assignments.
+	t.Run("Allocated_TracksAssignmentsUnderChurn", func(t *testing.T) {
```
jpbetz: Coverage here is genuinely good — I ran the suite against a real Postgres at this head and every assignment subtest passes with no skips, and `AdmitSeesConcurrentBinds` is exactly the test the admission change needed. Three things I'd still want, all in the "confirm what the row lock should already give us" category:

1. Concurrent Bind and Release for *different* actors on the same worker, asserting `allocated` ends at exactly `sum(rows)` and neither side loses an update. That's the interleaving the row lock exists for and nothing exercises it.
2. nit: a Release racing DeleteWorker. Both take the worker row lock so I'd expect them to serialize; what's unpinned is what the losing side sees (ErrNotFound vs. the silent nil Release returns when the row is gone) and what the outbox's last event for that worker is.
3. This churn test is a single goroutine driving version off GetWorker each step, so it validates the arithmetic but never the arithmetic under concurrency. Since `allocated` is a denormalized running total that nothing recomputes in production (`SumAllocated` is only called from tests), I'd want at least one concurrent form of the `allocated == sum(rows)` invariant — a handful of goroutines binding and releasing overlapping actor UIDs, same assertion once they've drained.

And tied to the schema comment: the legacy-row case (a stored proto carrying an old field-2 assignment, no `worker_assignments` row, `allocated` nil) pinned down somewhere — probably an atepg-level test with raw SQL, since it can't be seeded through the interface.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1418,6 +1420,239 @@ func (p *Persistence) DeleteWorker(ctx context.Context, name string, pre store.D
...
+
+		// Something is already bound to this Actor. Reading it costs a second
+		// statement, which is why it is here and not above: this is the
+		// retried claim, not the common one.
+		previous, previousWorker, err := getAssignmentRow(ctx, tx, actorUID)
+		if err != nil {
+			return nil, err
+		}
+		if previousWorker != workerName {
+			return nil, fmt.Errorf("actor %s is already hosted by worker %s", actorUID, previousWorker)
```
jpbetz: nit, fine as a follow-up: this is the one bare error left on the bind path, so the resume retry loop doesn't recognize it (it only retries `ErrVersionConflict` and `errWorkerFilledUp`) and the interceptor turns it into Internal for the client. It's hard to hit since resume holds the per-actor lease, but per your own comment above the insert a read can't see a claim that commits after it, and that's the case this branch is here to catch.

A retry would actually fix it rather than just re-roll: `assignWorkerAttempt` re-runs `workerHoldingStaleClaim` each pass, so the next attempt would see the other replica's committed claim and land on the rebind path. So wrapping `store.ErrVersionConflict` here would do it — same treatment you already give the vanished-worker case a few lines down.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1418,6 +1420,239 @@ func (p *Persistence) DeleteWorker(ctx context.Context, name string, pre store.D
 	})
 }
 
+// Worker assignments and status.allocated are updated in one transaction.
```
jpbetz: Taking you up on the offer in the description: I'd like this split. Land (1) the proto changes plus the adaptation commit plus capacity mutability plus ReportWorkerCapacity, then (2) `worker_assignments`, the five store methods, and the claim-lock commit on top. Everything from here down is the part that changes the store's data model and its CAS discipline, and this close to v1 it's the part I want to read slowly and separately from the mechanical adaptation. The API-shape argument for keeping them together is fair, but the proto half is reviewable on its own and Tim is already deep in it.

The stack already splits cleanly at 9d09c6f2, FWIW: I checked it out and `go build ./...` plus the scheduling/controlapi/functionaltest/resources/kubectl-ate tests are green there without the last two commits.

Ordering note while you're splitting: #1196 rejects an unversioned schema, so if it lands first, part 2's DDL needs to be a numbered migration file rather than an addition to schema.go — plus whatever we settle on for existing bound workers.

BenTheElder:
> The atomic update of an assignment and the worker seems like something that will hurt us later, but maybe workers just not updated often enough for that to matter?

FWIW I've pushed 28k claims/s with a test of the storage layer, it scales pretty well with cores.. For overcomitted toy actors, 97k in 60s ... mostly bound by dataplane. This is only going to hurt us if the rest of the system gets really fast or if we have reason to do a lot of other writes to workers unrelated to scheduling. We're _by far_ dataplane bound even at fairly considerable scale with this.

We could do something else like ... a distributed scheduling architecture, but I'm not sure that's necessary.

---

I've realized some other defects in this as-is when I got some time to sit and think over the weekend. Working on it.

1) I'd really like registered capacity to be orthogonal to pod creation. The worker should self report CPU and memory, even if we initially just downward API that in from the pods. That's a minor tweak (we already do this for actor count) but it needs some work.

2) ... devices. They're under-modeled currently, but they do sort of work in main. This regresses. Not straighforward.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: I debated splitting the types here, but that seemed relatively painful for one repeated field we could just not set.

> So the analog here would be:

That makes it more API calls / roundtrips to get the full details of a worker. Maybe that's the right trade-off, but it's a real tradeoff.

From my POV how these are actually stored in the backend is not something we should necessarily model directly into the API.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: > How would we model overcommit? Worker lies about capacity? Maybe add an overcommit scalar (e.g. 1.5)?

My gut take was worker lies about capacity.

We will have to filter this stuff up to autoscaling, that approach seems easy enough to reason about but perhaps too limiting.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: This mostly comes down to when we land it. If we land it before we have any back-compat, then this is unnecessary and I'll just clean it up. 

If we have to land it later, we need to handle old ateoms that fail to self-report.

I'm leaning towards landing this just being a hard break and dropping this hack.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: agree

cmd/ateapi/internal/store/atepg/atepg.go
BenTheElder: This bit is oudated. See later commit?

pkg/proto/ateapipb/ateapi.proto
BenTheElder: ACK.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: taking the hard break

BenTheElder:
I will iterate on this a bit more, including some cleanup to the commits etc, and notify when it's ready.

For now, logging off.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: Moved it to a `WorkerReporting` service. It's not part of CreateWorker because CreateWorker fires when we create the Pods, this fires when a worker boots and self-reports what capacity the implementation can handle.

That also allows us to make local decisions about overcommit in the future, and to avoid scheduling to worker pods that are not ready yet.

I sort of want to make Worker and creating worker pods _completely_ distinct but that blows the scope even further ...

pkg/proto/ateapipb/ateapi.proto
BenTheElder: I've been bikeshedding this particular point back and forth. The resource slice is kind of annoying.

BenTheElder: Reworked.

pkg/proto/ateapipb/ateapi.proto
juli4n: Can the caller just use `UpdateWorker` and update this field?

cmd/ateapi/internal/store/atepg/atepg.go
BenTheElder: The row lock handles this currently. But we should probably defensively cover it as well.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1751,6 +1779,39 @@ service ActorIdentity {
...
+  // atelet calls this with its own client certificate, as it does for
+  // MintCert. Idempotent: re-reporting the same capacity is not an update.
+  rpc ReportWorkerCapacity(ReportWorkerCapacityRequest) returns (ReportWorkerCapacityResponse);
+}
+
+message ReportWorkerCapacityRequest {
+  // The Worker being reported on. atespace is always empty; Workers are
+  // global-scoped.
+  // +k8s:opaqueType
+  ObjectRef worker = 1;
```
EItanya, line 1861: Doesn't this need to be inferred from the certificate anyway, is this redundant?

pkg/proto/ateapipb/ateapi.proto
EItanya: This feels like an ownership thing, or maybe subresource ownership/rbac. We would need to model that only the `ateom` on the worker can modify that particular field, (which we may need to do anyway). This separate RPC at least allows the authz to be slightly easier at the handler level.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1679,33 +1679,52 @@ message WorkerStatus {
   // +k8s:maximum=2 # keep this in sync with the WorkerState enum
   WorkerState state = 1;
 
-  // The Actor currently bound to this Worker, if any.
+  // The Actors bound to this Worker.
+  //
+  // Populated only by GetWorker. ListWorkers leaves it empty and reports
```
EItanya: So to be clear, this field is stored in the new `worker_assignments` table, but represented in the API separately?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1679,33 +1679,52 @@ message WorkerStatus {
...
+  // so the two are one vocabulary and subtract directly. A name the Worker does
+  // not report is unconstrained rather than absent, so a Worker that has said
+  // nothing is not unschedulable.
+  //
+  // Sorted by name, so equal capacities compare equal.
   //
   // +k8s:optional
-  // +k8s:minimum=1
-  int64 cpu_milli = 1;
+  Resources resources = 1;
```
EItanya, line 1761: Feel free to throw this away. If we parse this into a map in go anyway, should it be a map in the proto as well?

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1450,6 +1452,253 @@ func (p *Persistence) DeleteWorker(ctx context.Context, name string, pre store.D
...
+
+// lockWorkerForAssignment serializes updates to a Worker's allocation.
+func lockWorkerForAssignment(ctx context.Context, tx pgx.Tx, name string) (*ateapipb.Worker, error) {
```
EItanya: Can we rename this function then, it's really `startTransaction` or something along those lines

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1450,6 +1452,253 @@ func (p *Persistence) DeleteWorker(ctx context.Context, name string, pre store.D
...
+		worker, err := lockWorkerForAssignment(ctx, tx, workerName)
+		if err != nil {
+			return nil, err
+		}
+		if worker.Status == nil {
+			worker.Status = &ateapipb.WorkerStatus{}
+		}
+
+		// The ordinary bind is a first bind, and the insert landing is proof
+		// nothing was there to account for: one statement, no read.
```
EItanya: I really hate to be that guy but these comments are horrendously confusing

EItanya (commented):
Overall this makes sense, I still need to mull the whole thing over, but I am really liking the direction.

Codex found a couple potential races that we can confirm:

```
  Binding commits before the Actor is updated to reference the Worker. If ateapi crashes in
  between, the Worker holds an assignment that the Actor does not know about.

  Later, DeleteActor skips Worker cleanup when actor.status.worker_assignment is empty. There is
  no assignment-to-Actor foreign key or reconciler, so the capacity can remain leaked until the
  Worker disappears.

  Smallest sound fix: deletion must look up and release an assignment by Actor UID even when the
  Actor lacks a backlink. Add a test covering bind → missing Actor update → delete.

```

```
  When an Actor is already assigned to the same Worker, BindActorToWorker subtracts the old
  resources and adds the replacement, but does not call admit.

  Because Actor.actor_template is mutable, a stale claim can be resumed with larger limits and
  push the Worker over capacity.

  Smallest fix: subtract the previous reservation, run admission against that effective Worker,
  then add the replacement.
```

pkg/proto/ateapipb/ateapi.proto
BenTheElder: We're going to do `ListWorkerAssignments` and model it as a subresource in the API response

pkg/proto/ateapipb/ateapi.proto
BenTheElder: We plan to consolidate this with status.

cmd/ateapi/internal/store/atepg/atepg.go
BenTheElder: fair :-)

pkg/proto/ateapipb/ateapi.proto
BenTheElder: We don't currently dial directly from the worker, it comes from ateom => atelet.

I have confidence that will change in the future but ... not sure when. Probably after we actually have multi-actor worker ... bit chicken and egg.

BenTheElder:
> Codex found a couple potential races that we can confirm:

These are real and should be fixed. 

I need to do some other remaining cleanup.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: We now have paginated `ListWorkerAssignments`, it's in `ateapi: store a Worker's assignments as their own records`

pkg/proto/ateapipb/ateapi.proto
BenTheElder: Yes. Though we dropped that in a later commit / iteration, for the subresource with `ListWorkerAssignments` as a dedicated RPC.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: @thockin had an earlier comment about reusing the Resources type. I'm on the fence. Later we probably need to model more things. I don't love the Resources type but I like that it matches what we have on ActorTemplate.

BenTheElder: So I lean towards keeping it for now. If we want to use cpu_milli / memory bytes consistently we should rewrite ActorTemplate too.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1466,6 +1470,36 @@ message DeleteOptions {
...
+  // +k8s:required
+  // +k8s:beta(since: "0.0")=+k8s:subfield(atespace)=+k8s:forbidden # TODO: get rid of beta prefix
+  ObjectRef worker = 1;
+
+  // Requested page size; the server may return fewer. If unspecified, defaults
+  // to a server-chosen value; values above 1000 are coerced to 1000.
+  //
+  // +k8s:optional
+  // +k8s:minimum=1
+  int32 page_size = 2;
```
thockin, line 1501: @juli4n Should we bundle these into a generic `ListOptions` ?

cmd/ateapi/internal/store/atepg/atepg.go
BenTheElder: The caller starts the transaction, renamed to `getWorkerForUpdate`

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1661,18 +1695,18 @@ message Worker {
...
-  // capacity is asking to clear it, which the immutability rule rejects.
+  // What this Worker can supply to the Actors it hosts.
+  //
+  // Mutable: a Worker can be resized, and it reports its own actor ceiling.
+  // Clearing is rejected. Shrinking below allocated stops new placements and
+  // evicts nothing.
   //
   // +k8s:optional
-  // +k8s:immutable
   WorkerCapacity capacity = 10;
```
thockin: As per discussion, my takeaway was this moves to status.  If you disagree respond, else close this in due course.  Or are we killing this entirely?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1688,40 +1722,61 @@ enum WorkerState {
 }
 
 message WorkerStatus {
+  reserved 2; // was assignments, now the ActorAssignment subresource
```
thockin: We really don't need this

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1661,18 +1695,18 @@ message Worker {
...
   map<string, string> labels = 9;
 
-  // The compute capacity this worker can give an actor sandbox. Immutable, set
-  // at creation: a worker pod's limits are fixed for its lifetime. An update
-  // replaces the worker rather than patching it, so a request that omits
-  // capacity is asking to clear it, which the immutability rule rejects.
+  // What this Worker can supply to the Actors it hosts.
+  //
+  // Mutable: a Worker can be resized, and it reports its own actor ceiling.
+  // Clearing is rejected. Shrinking below allocated stops new placements and
```
thockin: If you want to prevent clearing: `+k8s:update=NoUnset`

pkg/proto/ateapipb/ateapi.proto
thockin: I don't love Resources but it maps to k8s.  There are ...issues... properly handling quantity strings, which I am chasing down on the k8s side.  I think we should ultimately decide to either support quantity strings all over or enforce quantized int values (e.g. cpu_millicores, memory_mb) all over.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1688,40 +1722,61 @@ enum WorkerState {
...
+// WorkerStatus.allocated, which uses this same message, so the two subtract. An
+// Actor is placed only where free covers its limits in every dimension.
+//
+// A Worker overcommits by reporting more than it has.
 message WorkerCapacity {
-  // CPU capacity in millicores (1000 = one core).
+  // What a Worker supplies, named as an ActorTemplate names what it asks for,
+  // so the two are one vocabulary and subtract directly. A name the Worker does
+  // not report is unconstrained rather than absent, so a Worker that has said
+  // nothing is not unschedulable.
```
thockin: I don't think this semantic works -- if I ask for 128 cores, you can't put me somewhere unless you know I have them.  If I ask for GPU, you can't put me on a worker that has no GPUs.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1480,6 +1484,36 @@ message DeleteOptions {
   string uid = 2;
 }
 
+message ListWorkerAssignmentsRequest {
+  // The Worker whose Actors to list. atespace is always empty; Workers are
+  // global-scoped.
+  //
+  // +k8s:required
+  // +k8s:beta(since: "0.0")=+k8s:subfield(atespace)=+k8s:forbidden # TODO: get rid of beta prefix
+  ObjectRef worker = 1;
```
BenTheElder, line 1494: the apitool linter doesn't permit this, but the AIP it cites permits a `parent` ref so I'm going with that
https://google.aip.dev/132#:~:text=The%20collection%27s%20parent%20resource%20should%20be%20called%20parent%2C%20and%20should%20be%20the%20only%20variable%20in%20the%20URI%20path.%20All%20remaining%20parameters%20should%20map%20to%20URI%20query%20parameters

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1688,40 +1722,61 @@ enum WorkerState {
...
-// scheduler only places an actor on a worker whose capacity is >= the actor's
-// declared resource limits. An unset message, or a zero field within it, means
-// "unknown/unset" for that dimension: treated as unconstrained so placement is
-// not blocked (matching the pre-capacity behavior).
+// WorkerCapacity is a Worker's total, not what is free. Free is capacity minus
+// WorkerStatus.allocated, which uses this same message, so the two subtract. An
+// Actor is placed only where free covers its limits in every dimension.
+//
+// A Worker overcommits by reporting more than it has.
 message WorkerCapacity {
```
thockin: Do we need this intermediate struct if we use Resources?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1688,40 +1722,61 @@ enum WorkerState {
...
 }
 
-// ActorAssignment names the Actor currently bound to a Worker — the inverse of
-// WorkerAssignment.
+// ActorAssignment is an Actor bound to a Worker, a subresource of Worker
+// rather than a field on it: a Worker hosts thousands, and a listing that
+// carried them would grow with occupancy. The inverse of WorkerAssignment.
 message ActorAssignment {
+  // atespace is always empty, as it is on the Worker these belong to. name is
+  // the Actor's UID, so the Actor that caused an assignment is what addresses
```
thockin, line 1781: I thought we said that `metadata.name` would match `metadata.uid` but I am also fine with this approach.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1765,6 +1829,39 @@ service ActorIdentity {
...
+  //
+  // atelet calls this with its own client certificate, as it does for
+  // MintCert. Idempotent: re-sending the same capacity is not a write.
+  rpc SetWorkerCapacity(SetWorkerCapacityRequest) returns (SetWorkerCapacityResponse);
+}
+
+message SetWorkerCapacityRequest {
+  // The Worker being reported on. atespace is always empty; Workers are
+  // global-scoped.
+  // +k8s:opaqueType
```
thockin: Why opaque?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1765,6 +1829,39 @@ service ActorIdentity {
...
+  ObjectRef worker = 1;
+
+  // What the Worker can hold. An unset dimension keeps what is recorded rather
+  // than clearing it, so a reporter that knows only its actor ceiling does not
+  // erase the compute capacity taken from the pod's limits.
+  // +k8s:opaqueType
```
thockin: Why opaque?

thockin (commented):
Proto seems straight forward at this point

pkg/proto/ateapipb/ateapi.proto
BenTheElder: Agree ...

Aside: this is also a meh claude comment for sure.

my main takeaway from this PR is to cut off claude for API changes and stick to boilerplate fixes. even writing out what you want winds up with ... ehh not there yet.

next time I will entirely hand-craft the API commit at minimum.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: I suggest we bulk edit that in a follow-up, it's somewhat orthogonal and I'm not totally sold either way yet.

We're also going to need more complex resource types O(soon) to express GPUs properly, a count really isn't sufficient.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: We did, and this is what I told the agent, and it decided we were wrong I suppose.

cmd/ateapi/internal/workercache/workercache.go
```diff
@@ -120,6 +120,15 @@ func (c *Cache) Forget(name string) {
 	delete(c.workers, name)
 }
 
+// Observe applies a store result before its watch event arrives.
+// It is version guarded and nil-receiver safe.
+func (c *Cache) Observe(worker *ateapipb.Worker) {
```
juli4n: This seems to be implementing something very similar to https://github.com/agent-substrate/substrate/issues/1370. 

Can we leave it out of scope from this PR? Sounds like an optimization which is fairly orthogonal to the multi-worker change. Probably deserves to be reviewed separately.

pkg/proto/ateapipb/ateapi.proto
juli4n: +1.

IIUC we want a field under status that merges both worker capacity and allocated. Something like this:

```
message WorkerStatus {
...
  WorkerAllocation allocation;
...
}

message WorkerAllocation {
  WorkerResources capacity = 1;
  WorkerResources allocated = 2;
}

message WorkerResources {
  Resources resources = 1;
  int32 actors = 2;
}
```

WDYT?

cmd/ateapi/internal/workercache/workercache.go
BenTheElder: Fair enough. 

I've been validating the performance throughout, but it's not really essential to implementing this PR.
This was always borderline. The subresource bit was more essential.

tools/apitool/internal/lint/list_test.go
```diff
@@ -36,6 +36,57 @@ func listAPI(reqFields, respFields []model.Field) *model.API {
 	}
 }
 
+// listSubresourceAPI is a List over a parented collection: ListWidgetParts
```
juli4n: I think we should just file exemptions for this in this PR. We have #1427 to make the linter understand sub-resources, that will remove the exemptions.

BenTheElder: Yes, I'm reverting that. On principle I don't like adding lint exceptions, but in this case the linter is what needs to change. I looked through our policy and the AIP. Next push will just have an exception here.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: for these it disable the validation, which is bad, dropping these and testing. ditto the other one

pkg/proto/ateapipb/ateapi.proto
BenTheElder: I'm not sure if it will just be resources, and actor count is not a dimension we represent elsewhere.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: I think what I want here is to require that these are set, and make ateom fully set them now.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: Elaborating: I think we're going to need a more complex model for devices than <name: count>. So we'd have to add a new type anyhow.

cmd/ateapi/internal/workercache/workercache.go
BenTheElder: FWIW this is somewhat loadbearing, after removing it some pre-eisting test cases flaked, those tests do a pause -> resume immediately and race the watch.

cmd/ateapi/internal/store/atepg/atepg.go
BenTheElder: updated

cmd/ateapi/internal/workercache/workercache.go
BenTheElder: [This was reverting, running some more tests]

BenTheElder: This flake exists on upstream/main. Evidence for the fix, outside of this PR.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: done @ HEAD

pkg/proto/ateapipb/ateapi.proto
BenTheElder: Done.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: This is moved to status entirely

pkg/proto/ateapipb/ateapi.proto
BenTheElder: https://github.com/agent-substrate/substrate/pull/1283/changes/7d9ced407ebdb003b9be712a41f3a4b3c3bc0789

BenTheElder:
https://github.com/agent-substrate/substrate/pull/1283/changes/7d9ced407ebdb003b9be712a41f3a4b3c3bc0789 is substantial new code hoisted to this branch to manage https://github.com/agent-substrate/substrate/pull/1283#discussion_r3919983819 -- we have the ateoms fully self-report capacity now.

For now it's just count = 1 and cpu/memory from downward API, to keep complexity down and behavior close to before (but still only after ateom boots).

In the future we will report live capacity (read cgroups instead, consider IPPR), account for overcommit, etc.

pkg/proto/ateapipb/ateapi.proto
BenTheElder: left this for now.

demos/claude-code-multiplex/ui/server.go
```diff
@@ -220,7 +220,7 @@ func actorStateString(s ateapipb.ActorState) string {
 // The UI's badgeFor() treats "running" as green; "idle" falls through
 // to the neutral badge, which is the right visual treatment.
 func workerPhase(w *ateapipb.Worker) string {
-	if w.GetStatus().GetAllocated().GetActors() > 0 {
+	if w.GetStatus().GetAllocation().GetAllocated().GetActors() > 0 {
```
BenTheElder, line 223: a little clunky but oh well

BenTheElder:
I think all review comments are addressed. 💤

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1765,6 +1843,42 @@ service ActorIdentity {
...
+// authorization: Control is the client-facing API, while these RPCs are served
+// only to an atelet, and only for the Workers on its own node.
+service WorkerService {
+  // SetWorkerCapacity records what a Worker can hold. Capacity is the Worker's
+  // to report rather than the control plane's to infer: it is what the ateom
+  // can actually supply, only its node can observe it, and a fleet may run
+  // mixed ateom versions.
+  //
+  // atelet calls this with its own client certificate, as it does for
+  // MintCert. Idempotent: re-sending the same capacity is not a write.
```
EItanya, line 1851: > Idempotent: re-sending the same capacity is not a write.

Is this trying to signal that it won't kick off a watch event for that worker?

internal/ateomcapacity/ateomcapacity.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+	slog.InfoContext(ctx, "Reported worker capacity", slog.Any("capacity", capacity.GetCapacity()))
+	return nil
+}
+
+// retryReport calls send until it succeeds or ctx ends, backing off between
+// attempts.
+func retryReport(ctx context.Context, send func() error, backoff time.Duration) error {
```
EItanya, line 130: super nit: this function takes a callback to be generic,  (and maybe for the unit tests), but makes reading it slightly less clear as it's only actually retrying the `reportOnce`. Consider just getting rid of the callback and call directly.

cmd/atelet/workercapacity.go
```diff
@@ -0,0 +1,59 @@
...
+	workers ateapipb.WorkerServiceClient
+}
+
+// SetWorkerCapacity records what the calling worker says it has.
+//
+// It returns the control plane's error unwrapped so the caller retries: a
+// worker reports once, so an accepted call is the only thing that puts
+// capacity on the Worker, and a Worker record the syncer has not created yet
+// is the ordinary reason for a first attempt to fail.
+func (s *workerCapacityService) SetWorkerCapacity(ctx context.Context, req *ateletpb.SetWorkerCapacityRequest) (*ateletpb.SetWorkerCapacityResponse, error) {
```
EItanya, line 40: Are we doing this `ateom -> atelet -> ateapi` because that's what we already have, or because that's what we actually want?

cmd/ateapi/internal/store/atepg/outbox.go
```diff
@@ -104,6 +104,37 @@ func (p *Persistence) writeAndAppendEvent(ctx context.Context, eventType store.W
 	return worker, nil
 }
 
+// writeAndAppendEventFor is writeAndAppendEvent for a write whose event
+// payload is only known once the transaction has read what it is changing.
+// fn returns the worker to announce, or nil for a write that changed nothing.
+func (p *Persistence) writeAndAppendEventFor(ctx context.Context, eventType store.WorkerEventType, fn func(ctx context.Context, tx pgx.Tx) (*ateapipb.Worker, error)) error {
```
EItanya: These helpers appear identical except that `writeAndAppendEventFor` discards the `Worker` result. The existing helper already supports discovering the event payload inside the transaction. Could `Bind` ignore its returned `Worker` and Release return it directly, then remove this duplicate?

cmd/ateapi/internal/controlapi/workflow_worker_delete.go
EItanya, line 41: Codex pointed out something interesting here. Essentially `Worker` remains `schedulable` during this deletion run.

```
  A Worker remains ACTIVE while this sweep runs, so a concurrent resume can bind a new Actor after
  its page has been processed. finalizeDeleted then deletes the Worker and cascades that new
  assignment without clearing the Actor's backlink. The worker syncer supplies no version
  precondition, so the bind's Worker version bump does not prevent this. Could deletion first
  persist a state that makes Applies reject new binds, then sweep assignments?
```

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -1456,6 +1456,299 @@ func (p *Persistence) DeleteWorker(ctx context.Context, name string, pre store.D
...
+
+func (p *Persistence) BindActorToWorker(ctx context.Context, workerName string, assignment *ateapipb.ActorAssignment, admit func(*ateapipb.Worker) error) error {
+	actorUID := assignment.GetActorUid()
+	if actorUID == "" {
+		return fmt.Errorf("binding an assignment with no actor_uid to worker %s", workerName)
+	}
+	// The store assigns identity. atespace is empty because Workers are
+	// global-scoped; the name is the Actor's UID, which is also the row key.
+	assignment.Metadata = &ateapipb.ResourceMetadata{Name: actorUID}
+	setCreateMetadata(assignment.Metadata)
```
EItanya, line 1514: Codex comment, seems worth looking at

```
  Because create metadata is generated before the insert attempt, the rebind path replaces the
  existing assignment with a fresh UID, creation timestamp, and version 1. A retry of the same
  logical binding therefore appears to be a newly created subresource. Should the rebind path
  preserve the previous UID/creation time and apply update metadata instead?
```

EItanya (commented):
Mostly nits, but also a couple real potential races

cmd/atelet/workercapacity.go
BenTheElder: The former. But we could switch in the future with only authz changes. I'm not sure we want to introduce "ateom uses the network directly" in this PR.

pkg/proto/ateapipb/ateapi.proto
thockin: Obviously for a followup PR

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1693,35 +1721,76 @@ message WorkerStatus {
...
+
+  // What the assignments consume of capacity. A running total because placement
+  // reads it for every Worker on every decision, and summing the assignments
+  // would cost the fleet's actor count each time.
+  //
+  // +k8s:optional
+  WorkerResources allocated = 2;
+}
+
+// WorkerResources is a vector of what a Worker has or holds. Capacity and
```
thockin: Comments here are awkward since you use this type for "capacity" and "allocated".  Could probably use human wordsmithing.

The comment on this type should talk about the type's structure, not how it is used in the parent.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1693,35 +1721,76 @@ message WorkerStatus {
...
+// allocated share it so the two subtract: free is capacity minus allocated, and
+// an Actor is placed only where free covers its limits in every dimension.
+//
+// A Worker overcommits by reporting more than it has.
+message WorkerResources {
+  // What a Worker supplies, named as an ActorTemplate names what it asks for,
```
thockin: ditto on comments

pkg/proto/ateapipb/ateapi.proto
BenTheElder: Yes. reworded.

cmd/ateapi/internal/controlapi/workflow_worker_delete.go
BenTheElder: should be fixed

cmd/ateapi/internal/store/atepg/atepg.go
BenTheElder: should be fixed

pkg/proto/ateapipb/ateapi.proto
thockin: Added to https://github.com/agent-substrate/substrate/issues/1378

thockin (commented):
Just the proto file again

pkg/proto/ateapipb/ateapi.proto
juli4n: I don't have a strong opinion. We can wrap it and reuse the validations. Each response still needs to define a next page token.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1480,6 +1484,39 @@ message DeleteOptions {
...
+  //
+  // +k8s:optional
+  // +k8s:maxLength=256
+  string page_token = 3;
+}
+
+// ListWorkerAssignmentsResponse is one page of a Worker's Actors.
+message ListWorkerAssignmentsResponse {
+  // The Actors this page of the listing covers.
+  repeated ActorAssignment worker_assignments = 1;
```
juli4n: Are we using consistent terminology?  Should `ActorAssigment` be `WorkerAssigment`? We already have a rule that List<Resource> methods should have a repeated field with the pluralized name of the resource.

I noticed that `WorkerAssigment` message exists already, and it's use for the bookkeeping of the worker assigment from the other end: `Actor.status.worker_assigment` so I can see three options: a) don't do anything and keep the current (arguably confusing) terminology, b) rename `ActorAssigment` to `WorkerAssigment` (which will require we also rename the message on the other end to `ActorAssigment`), or c) call the subresource `ActorAssigment` (i.e. `ListActorAssigmentsResponse{actor_assigments}`).

juli4n: Option c) should have been 

"call the subresource `Assigment` (i.e. ListWorkerAssigmentsResponse{assigments})."

internal/resources/quantities.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+// Quantities is the parsed form of an ateapipb.Resources: what an Actor asks
+// for, what a Worker supplies, or the sum of what a Worker's Actors hold. The
+// same type serves all three so they subtract, which is the point of naming
+// Worker capacity the way an ActorTemplate names its limits.
+//
+// An absent name is none of that resource. A Worker reports every dimension it
+// has, so a name missing from its capacity is one it cannot supply at all.
+type Quantities map[string]resource.Quantity
```
thockin, line 32: OK for the moment, but we need to decide if we REALLY want to use `Quantity` the string format and if so if we really want to parse it into `Quantity` the type or into a `math/big.Float` or something

internal/ateomcapacity/ateomcapacity.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/agent-substrate/substrate/internal/resources"
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+
+	"github.com/agent-substrate/substrate/internal/proto/ateletpb"
+)
+
+// Environment variables the atecontroller sets on the ateom container from the
+// downward API, in milli-cores and bytes.
+const (
+	CPULimitEnv    = "ATEOM_CPU_LIMIT_MILLI"
```
thockin: Minor, but we should consume this thru a projected file instead of env, so IPPR can work.  Just noting for later.  No action needed

thockin (commented):
This PR is big enough that there are no obvious bugs.  I'm still worried about the multi-resource "dance" of creating the assignment. There are things I am not quite sure are right but are not obviously wrong either - for example a RESUMING actor bound to a DRAINING worker currently crashes but it's not clear to me if that is AI paranoia, if it is a legitimate "oh no" situation, or if it is actually an acceptable corner case.

Basically I'm trying to convince myself that the high-level flow is right.  There are long sequences of tricky, conditional-rich code with no clear comments on assumed preconditions or asserted postconditions.

I'm still walking thru it and I will send you a diff in bit just for some comments I think could be better.  I don't see a reason to block this going in at this point.

internal/ateomcapacity/ateomcapacity.go
BenTheElder: Ahh, the TODO I had is to switch to reading cgroups, TIL about the updated file. Probably worth folding that in quickly

internal/resources/quantities.go
BenTheElder: Yeah. Do we already have a TODO for this? Should we file one?

internal/ateomcapacity/ateomcapacity.go
BenTheElder: done.

Issue #1258, opened 2026-08-27
Default OTLP endpoint points at an addon namespace that usually does not exist


git286:
Default OTLP endpoint points at an addon namespace that usually does not exist

manifests/ate-install/ate-otel-config.yaml defaults to http://opentelemetry-collector.gke-managed-otel.svc.cluster.local:4317. On any cluster without GKE's managed OTel addon enabled - the default gke-managed-otel namespace does not exist, so every component fails to resolve the collector once a minute, forever, and telemetry is silently dead. It presents as a networking fault rather than a missing dependency.

The installer rewrites the ConfigMap, so patching it after install is silently reverted. The override has to go in at install time via ATE_OTLP_ENDPOINT, which hack/install-ate.sh already honours.

Ask: default to no exporter, or make the install refuse when the endpoint's namespace is absent.

JeffLuoo:
> Ask: default to no exporter, or make the install refuse when the endpoint's namespace is absent.

We should make Observability a must-have, despite which OTel collector users use, managed one or customized. I prefer "make the install refuse when the endpoint's namespace is absent", WDYT? @git286

git286:
yeah, I think we should have some preflight check in the install script and explicitly tell users what's wrong and what they should do, instead of silently failure

BenTheElder:
> We should make Observability a must-have,

_really_? If the user doesn't want it, why is this a hard requirement? I could imagine not wanting to collect otel because storing it will add up with a lot of actors and you may already be doing something else at the actor level.

Our components shouldn't even assume that the install scripts are in use.


> Ask: default to no exporter, or make the install refuse when the endpoint's namespace is absent.

We need to split out installing for GKE versus installing generally. install-ate.sh is currently primarily for GKE but quasi-pretending to be generic. Instead what we have here must become generic. vendor neutrality is actually a requirement for e.g. CNCF sandbox were we to go that route.

Installing for GKE can expect the GKE namespace, the components need to be well behaved without it though.

JeffLuoo:
> I could imagine not wanting to collect otel because storing it will add up with a lot of actors and you may already be doing something else at the actor level.

We don't have actor level metrics as we also discussed in other threads, but cardinality will grow with ate namespace or template name, which the system metric use as attributes.

[...]

PR #1289, opened 2026-08-28
ate-setup: make observability optional and configurable

JeffLuoo:
### Summary
Decouples Substrate from a hardcoded GKE OTel collector by making telemetry optional and explicit via `--observability none|otlp|gke|kind` (and `--otlp-endpoint`).

On clusters without a collector, Substrate now runs cleanly without export retry spam, while preserving request trace correlation IDs in logs and `/metrics` pull endpoints.

---

### Key Changes
* **Explicit Modes (`cmd/ate-setup`)**:
  * `none`: No OTLP push; runs cleanly without a collector.
  * `kind` (default for local): Points to the in-cluster Kind collector.
  * `gke`: Points to GKE managed OTel addon (`gke-managed-otel`).
  * `otlp`: Points to a custom `--otlp-endpoint`.
* **Cluster State & Idempotence**:
  * Persists active mode as an annotation (`ate.dev/observability-mode`) on the `ate-otel-config` ConfigMap.
  * Targeted redeploys (e.g. `ate-setup deploy atelet`) preserve the cluster's existing mode instead of resetting to defaults.
  * Pre-existing unannotated clusters have their mode inferred from the endpoint for backwards compatibility.
* **Graceful Zero-Collector Handling (`internal/serverboot`)**:
  * An empty endpoint or `none` disables OTLP export without failing back to `localhost:4317`.
  * Instantiates `TracerProvider` without an exporter rather than `noop`, keeping valid `trace_id`s in structured logs.
  * Preserves pull-based Prometheus `/metrics` endpoints across all components.
* **Safety & Restarts**:
  * **Preflight validation**: Verifies in-cluster collector Service/Namespace exists before deploying workloads.
  * **Minimal disruption**: Only triggers rolling restarts of consumer workloads if the collector configuration actually changed.

---

### Verification
* All unit tests passing:
  `go test -v ./internal/serverboot ./cmd/ate-setup/...`
* Verified clean startup in `--observability=none` (no dial errors, `trace_id` present in logs).
* Verified mode preservation during single-component redeploys and smooth migration on mode changes.
Part of #1258

- [x] Tests pass


JeffLuoo:
@BenTheElder as we discussed in https://github.com/agent-substrate/substrate/issues/1258, I have this PR to make observability stack more configurable by breaking it down into four options `--observability none|otlp|gke|kind`.

If we want to further make it abstract for the concept of cluster type like kind and GKE, I can change the implementation to just two options `--observability none|otlp`, and if `otlp` is selected, an endpoint needs to be passed from flag as well. Having GKE and Kind as options for the simplicity of configuring `--otlp-endpoint`.

manifests/ate-install/otel/none/ate-otel-config.yaml
```diff
@@ -0,0 +1,44 @@
...
+# NOTE: a change to this ConfigMap does not restart the pods that read it. The
+# pod template stays the same, thus there is no new rollout. After a change, do
+# `kubectl rollout restart` on the affected workloads.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ate-otel-config
+  namespace: ate-system
+data:
+  OTEL_EXPORTER_OTLP_ENDPOINT: ""
```
git286, line 50: The main exporters in serverboot.go are built unconditionally, and the OTel Go SDK's env parsing treats an empty variable as unset. So when the mode is none, the exporter falls back to the SDK default endpoint, localhost:4317, which will error, right?

hack/install-ate.sh
```diff
@@ -287,58 +295,17 @@ apply_atenet_egress() {
...
-  for workload in deployment/ate-api-server deployment/ate-controller \
-                  deployment/atenet-router daemonset/atelet; do
-    if run_kubectl -n ate-system get "${workload}" >/dev/null 2>&1; then
-      run_kubectl -n ate-system rollout restart "${workload}"
-    fi
-  done
+# No bundle carries a copy of this ConfigMap. Thus this is the one apply of it,
+# and the install needs no patch after the bundle and no restart of the
+# workloads that read it.
+apply_otel_config() {
```
git286, line 410: Every deploy path calls apply_otel_config, including the targeted ones (-- deploy-atelet etc.). The mode only lives in the flags, so if I installed with
--observability=gke a while ago and later run ./hack/install-ate.sh --deploy-atelet without repeating the flag, this quietly applies the none ConfigMap over my gke one. The redeployed component loses telemetry right away and the rest lose it whenever they next restart, which makes it really hard to trace back. Could we leave the ConfigMap alone (or read the mode back from the cluster) when --observability wasn't explicitly given and a config already exists?

hack/install-ate.sh
```diff
@@ -591,9 +559,6 @@ deploy_ate_system() {
   run_kubectl rollout status deployment/atenet-router -n ate-system --timeout="$(rollout_timeout)"
   run_kubectl rollout status deployment/atenet-egress -n ate-system --timeout="$(rollout_timeout)"
   run_kubectl rollout status daemonset/atelet -n ate-system --timeout="$(rollout_timeout)"
-
-  # After the bundle, which carries its own copy of ate-otel-config.
-  apply_otel_endpoint_override
```
git286, line 979: There's no way to switch modes on an existing cluster anymore. The components only read ate-otel-config when a pod starts — updating the ConfigMap does nothing to pods that are already running. So if I re-install with a different --observability mode, the ConfigMap changes but no pod restarts (nothing in the pod spec changed), and everything keeps using the old endpoint. The deleted apply_otel_endpoint_override handled this: when the endpoint changed, it restarted the four workloads that read the config. We need that restart back in some form, the docs say re-installing is how you change modes, and right now that silently does nothing.

manifests/ate-install/otel/none/ate-otel-config.yaml
JeffLuoo: I plan to do this in a follow-up PR.

Note that it is not a regression: before this PR the same components failed at the same rate against gke-managed-otel. I added a line to docs/observability.md that says mode none stops the export but not the SDK error log yet and will clean it up in my follow-up PR.

hack/install-ate.sh
JeffLuoo: Good catch.

The install now writes the mode on the ConfigMap as the ate.dev/observability-mode annotation, and reads it back when the command line gives no mode. Order: --observability, then --otlp-endpoint, then the mode of the cluster, then kind on a kind install, then none. Thus --deploy-atelet after an install with --observability=gke keeps mode gke, and the install reports Observability: mode gke (from the cluster).

hack/install-ate.sh
JeffLuoo: Updated:

- apply_otel_config now compares the new ConfigMap with the one in the cluster and records a change. restart_otel_consumers will restartsystem components when the collector changed.
- Updated documentation to align with what code does.

hack/observability.sh
```diff
@@ -0,0 +1,386 @@
...
+}
+
+# ATE_OTEL_CONFIG_CHANGED records that this install gives the cluster a
+# different collector than the one it had. restart_otel_consumers reads it.
+ATE_OTEL_CONFIG_CHANGED=false
+
+# note_otel_config_change compares the ConfigMap of this install with the one in
+# the cluster. Nothing to compare on a first install, thus no change: the
+# workloads that come after it read the new ConfigMap when they start.
+note_otel_config_change() {
```
git286: AI finding:

The cluster snapshot this compares against is read once and never updated, so with more than one deploy target the restart fires twice. E.g. on a mode-none cluster, ./hack/install-ate.sh --observability=gke --deploy-atelet --deploy-atenet: the atelet path sees none→gke and restarts all four consumers, then the atenet path compares against the same cached "none" (the ATE_OBSERVABILITY_CLUSTER_READ guard blocks a re-read), decides the collector changed again, and restarts everything a second time.

Since every ate-controller restart also rolls all the WorkerPools, that's the running workers and actors replaced twice in one install.

Could we update the cached mode/endpoint after the apply (or after the restart), so the second comparison sees the value we just wrote?

benchmarking/workloads/deploy.sh
```diff
@@ -77,15 +77,20 @@ usage() {
 # ate-system must exist, because the workloads below need its CRDs. Thus an
 # absent ConfigMap is an error and not a condition to work around: a default
 # here would send the actor telemetry to the wrong collector with no message.
+#
+# An empty endpoint is an error for the same reason. It is the value of
+# --observability=none, in which the control plane has no collector; the actors
+# then need --otlp-endpoint, or a collector for the whole install.
```
git286, line 90: I think this breaks the benchmark automation. orchestrator.py tears down and reinstalls with plain --deploy-ate-system without --observability flag, so the fresh install always ends up in mode none with an empty endpoint. So every automated run dies at setup before any benchmark starts?

hack/observability.sh
JeffLuoo: Fixed.

note_otel_config_change now takes the values it applied as the values of the cluster, in the same call, thus each target after the first compares with what the first one wrote. Your suggestion, with the update at the comparison and not after the restart, so a deploy path that ends before the restart cannot leave a stale snapshot behind.

benchmarking/workloads/deploy.sh
JeffLuoo: Fixed

JeffLuoo:
Because of #1503, I will refactor this change. cc - @git286

JeffLuoo:
## Impact on ai-on-gke/substrate-gke (https://github.com/ai-on-gke/substrate-gke)

This change has no effect on that repository today. Its installer keeps a pin to an upstream commit, Commit = "cbae8250" (installer/internal/snapshot/snapshot.go:59). That commit is older than this change. Thus nothing changes there before a person moves the pin.

After a move of the pin, that repository still needs no change. Its runs go run ./cmd/ate-setup deploy ate-system. It gives the image flags only, and no --observability flag. There are two cases:

- A first install. The script runs setup-gcp bootstrap before the install. That step enables the managed OTel addon, and the reconcile for a cluster that exists). The install then finds the collector of the addon, and selects mode gke. Thus the telemetry of the quickstart stays as it is today.
- An install on a cluster that has substrate. The ate-otel-config ConfigMap of that cluster holds the endpoint, and the install keeps the mode of it. A ConfigMap from an install before this change holds no annotation. The install then reads the mode from the endpoint in it.

cmd/ate-setup/internal/steps/observability.go
```diff
@@ -0,0 +1,421 @@
...
+	log.Infof("The collector changed. Restarting the workloads that read %s.", otelConfigMapName)
+	log.Infof("  ate-controller rolls each WorkerPool when it restarts, thus the running workers, " +
+		"and the actors on them, are replaced.")
+	now := time.Now()
+	for _, name := range otelConsumerDeployments {
+		if err := e.Kube.RolloutRestartDeployment(ctx, NamespaceAteSystem, name, now); err != nil {
+			return err
+		}
+	}
+	return e.Kube.RolloutRestart(ctx, NamespaceAteSystem, "atelet", now)
```
krisztianfekete, line 420: Does this work? Doesn't it need the suffix? You can likely use `e.AteletDaemonSetName()` I think.

cmd/ate-setup/internal/steps/observability.go
```diff
@@ -0,0 +1,421 @@
...
+)
+
+// gkeOtelNamespace is the namespace that the GKE managed OTel addon makes. The
+// endpoint in manifests/ate-install/otel/gke/ate-otel-config.yaml names it.
+const gkeOtelNamespace = "gke-managed-otel"
+
+// otelConsumerDeployments are the Deployments that read the ConfigMap with
+// envFrom. The atelet DaemonSet reads it too; it takes a different restart
+// call, thus restartOtelConsumers names it on its own.
+var otelConsumerDeployments = []string{"ate-api-server", "ate-controller", "atenet-router"}
```
krisztianfekete, line 49: Think you also need to include `atenet-egress` as well?

cmd/ate-setup/internal/steps/observability.go
```diff
@@ -0,0 +1,421 @@
...
+		mode, source = config.ObservabilityOTLP, "the --otlp-endpoint flag"
+	default:
+		e.readClusterObservability(ctx)
+		switch {
+		case e.observability.clusterMode != "":
+			mode, source = e.observability.clusterMode, "the cluster"
+			// Only mode otlp keeps its address in the ConfigMap. Each other
+			// mode holds it in a manifest.
+			if mode == config.ObservabilityOTLP {
+				endpoint = e.observability.clusterEndpoint
```
krisztianfekete, line 201: Would this get reverted?
https://github.com/agent-substrate/substrate/blob/4a4c59ad7274428f5aa57a5ffda80ebfb6bc9802/hack/install-ate.sh#L452-L454

cmd/ate-setup/internal/steps/observability.go
```diff
@@ -0,0 +1,421 @@
...
+// Mode otlp uses the file of mode none, and renderOtelConfig puts the given
+// endpoint in it.
+func (e *Env) otelConfigPath(mode string) string {
+	switch mode {
+	case config.ObservabilityKind:
+		return e.Cfg.Manifest("otel", "kind", "ate-otel-config.yaml")
+	case config.ObservabilityGKE:
+		return e.Cfg.Manifest("otel", "gke", "ate-otel-config.yaml")
+	default:
+		return e.Cfg.Manifest("otel", "none", "ate-otel-config.yaml")
```
krisztianfekete, line 89: We are missing `OTEL_METRIC_EXPORT_INTERVAL`, would this reintroduce https://github.com/agent-substrate/substrate/issues/634?

cmd/ate-setup/internal/steps/observability.go
JeffLuoo: Updated

cmd/ate-setup/internal/steps/observability.go
JeffLuoo: Good catch, fixed!

cmd/ate-setup/internal/steps/observability.go
JeffLuoo: Fixed.

The restart now uses `RestartAteletDaemonSets`, which selects the DaemonSets by the `app=atelet` label. I did not use `AteletDaemonSetName`, because a cluster in a rolling upgrade has one DaemonSet for each version. Each of them reads the ConfigMap, thus each of them needs the restart.

I tested this on a live GKE cluster.

cmd/ate-setup/internal/steps/observability.go
JeffLuoo: Issue #634 does not come back, but I found a different problem near it. Mode otlp holds its address in no manifest, thus it used the file of mode `none` and lost the tick. A kind cluster with its own collector went back to the 60s default of the SDK. The patch of the shell script kept the tick, because it changed the endpoint only.

Fixed in latest patch -> Mode `otlp` now uses the kind file on a kind install, and the none file on each other install. The tick belongs to the cluster, and not to the collector.

BenTheElder:
[I think we're trying to consolidate towards the go code now]

JeffLuoo:
> [I think we're trying to consolidate towards the go code now]

Yes. The PR has been refactored to only modify go code.

PR #1290, opened 2026-08-28
Switching e2e tests to use the new ActorTemplate substrate proto

zoez7:
Switches all e2e tests that uses counter demo to validate the new substrate proto ActorTemplate.

Made some changes to make the e2e test pass:
* cmd/ateapi/internal/controlapi/template_reconciler.go - golden actors still need the golden atespace, because the logic in suspend actor relies on it to always take FULL snapshot regardless of the config. (filed https://github.com/agent-substrate/substrate/issues/1299)
* internal/ateattr/ateattr.go - Added placeholder label for now, will fix in a follow up PR to wire the metric reporting when using new ActorTemplate substrate object.


.github/workflows/pr-workflow.yaml
```diff
@@ -83,23 +83,24 @@ jobs:
...
     - name: Install Agent Substrate
       run: hack/install-ate-kind.sh --deploy-ate-system
-    - name: Deploy micro-VM counter demo
-      # Stages the (cached) assets into the cluster's rustfs and applies the
-      # counter-microvm demo onto the control plane installed above.
-      run: hack/run-microvm-demo-kind.sh
-    - name: Deploy gVisor counter demo
-      run: hack/install-ate-kind.sh --deploy-demo-counter
+    - name: Deploy substrate micro-VM counter demo
+      # Stages the (cached) micro-VM assets into the cluster's rustfs (the
```
zoez7: Do not change the previous comment here.

.github/workflows/pr-workflow.yaml
```diff
@@ -83,23 +83,24 @@ jobs:
...
+      # egress-microvm demo and the micro-VM e2e lane need them too), then
+      # deploys the substrate counter-microvm demo and blocks on its golden
+      # snapshot.
+      run: hack/run-microvm-demo-kind.sh --substrate
+    - name: Deploy substrate gVisor counter demo
+      # The counter-based suites copy their per-test templates' resolved
```
zoez7: Same, do not change the previous comment here.

hack/install-demo-counter-substrate.sh
```diff
@@ -45,18 +45,21 @@ demo-counter-substrate_cmdline() {
       demo-counter-substrate_deploy_variant \
         demos/counter/counter-substrate.yaml.tmpl \
         demos/counter/counter-substrate-template.yaml.tmpl \
-        ate-demo-counter-substrate counter-substrate counter
+        ate-demo-counter-substrate counter-substrate counter 300
```
dberkov, line 48: what does 300 mean?

dberkov (approved):

PR #1293, opened 2026-08-28
Ateom child reaper

### LINKED ISSUE #1266: [EPIC] Multi-Actor Worker Support

BenTheElder:
Necessary for #1266, but functionally useful regardless.

This eliminates the MPL go-reap dependency in favor of our own implementation tuned to our needs.

Our package has utilities for spawning child processes without holding a RWMutex to avoid convoying lots of concurrent activations in the future.


cmd/ateom-gvisor/runsc.go
```diff
@@ -373,20 +349,17 @@ func (r *runsc) waitArgs(containerName string) []string {
...
+// and an entry held that long would hold off reaping and, past MaxDefer, every
+// other runsc invocation with it. The cost is that reaping can collect the
+// `runsc wait` process first, so a clean shutdown may report ECHILD.
 func (r *runsc) cmdWait(ctx context.Context, containerName string) error {
 	slog.InfoContext(ctx, "About to run runsc wait", slog.String("container", containerName))
 
 	cmd := exec.CommandContext(ctx, r.path, r.waitArgs(containerName)...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	if err := cmd.Run(); err != nil {
```
dberkov, line 363: In cmd/ateom-gvisor/main.go, killContainer() treats any non-timeout/non-cancellation error from cmdWait as a hard failure.

If the background reaper collects runsc wait before cmd.Run()'s internal wait finishes, cmd.Run() returns syscall.ECHILD / waitid: no child processes. As a result, killContainer() returns an error instead of treating it as a successful exit, and aborts the shutdown sequence without escalating to SIGKILL.

Proposal: In cmdWait, detect unix.ECHILD / syscall.ECHILD and treat it as success (or log at debug), because runsc wait exiting (even if reaped by the background reaper) indicates the container process exited

---
if err := cmd.Run(); err != nil {
    if errors.Is(err, unix.ECHILD) {
        slog.DebugContext(ctx, "runsc wait collected by background child reaper", slog.String("container", containerName))
        return nil
    }
    return fmt.Errorf("while running `runsc wait`: %w", err)
}
----

internal/childreap/childreap.go
```diff
@@ -0,0 +1,222 @@
...
+		if r.inFlight == 0 {
+			if !now.Before(drainAt) {
+				slog.WarnContext(ctx, "Reaped children only after holding off subprocesses",
+					slog.Duration("waited", MaxDefer))
+			}
+			r.draining = false
+			r.reaping = true
+			return true
+		}
+		if !now.Before(giveUpAt) {
```
dberkov, line 238: When acquire() gives up after MaxDefer + MaxDrain (90s), reapOnce() returns without calling wait4. In Run(ctx), execution loops back to select { case <-sigs: ... }.

* The dead child that originally triggered SIGCHLD 90 seconds earlier is still a zombie.
* Because Linux only generates SIGCHLD on process state transitions, the kernel will not send another SIGCHLD for that already-dead orphan.
* When the long-running subprocess finally exits (e.g. at t = 95s), leave() calls r.cond.Broadcast(), but the reaper goroutine is waiting on <-sigs, not on r.cond.
* If no further child processes terminate before pod termination, that orphan remains a zombie indefinitely.

Recommendation: If a drain round is abandoned, consider re-arming the reaper when inFlight drops back to 0 (or using a non-blocking re-send to sigs once inFlight == 0), so stranded zombies are cleaned up once the blocking subprocess finally exits:

---
func (r *Reaper) leave() {
    r.mu.Lock()
    defer r.mu.Unlock()
    r.inFlight--
    if r.inFlight == 0 {
        if r.abandoned {
            r.abandoned = false
            // Wake or signal reaper that a quiet moment is finally available
        }
        r.cond.Broadcast()
    }
}
---

cmd/ateom-microvm/internal/reaper/reaper_linux_test.go
```diff
@@ -0,0 +1,40 @@
...
+package reaper
+
+import (
+	"os/exec"
+	"testing"
+	"time"
+)
+
+// TestRunSurvivesAConcurrentReaper verifies the package wiring.
+func TestRunSurvivesAConcurrentReaper(t *testing.T) {
```
dberkov, line 25: The 8 goroutines spawn sleep 0.05 & exit 0 exactly once. They finish in ~50ms. For the remaining ~4.95 seconds of the 5-second test loop, no orphans or SIGCHLDs are being generated, so RunCombined is testing against an idle reaper.

Loop the orphan generators until deadline (similar to how TestShortSubprocessesAreNotPacedByLongOnes in childreap_linux_test.go uses for run.Err() == nil):

---
for range 8 {
    go func() {
        for time.Now().Before(deadline) {
            _ = exec.Command("sh", "-c", "sleep 0.05 & exit 0").Run()
        }
    }()
}
---

internal/childreap/childreap.go
```diff
@@ -0,0 +1,222 @@
...
+		default:
+			slog.WarnContext(ctx, "Reaping children failed", slog.Any("err", err))
+			return
+		}
+	}
+}
+
+// acquire waits for in-flight subprocesses, blocking new ones after MaxDefer.
+// It returns false if ctx ends or the drain exceeds MaxDrain.
+func (r *Reaper) acquire(ctx context.Context) bool {
```
dberkov, line 176: Fast-Path Allocation Optimization in acquire()

drainAt := time.Now().Add(MaxDefer)
giveUpAt := drainAt.Add(MaxDrain)
for _, at := range []time.Time{drainAt, giveUpAt} {
    timer := time.AfterFunc(time.Until(at), r.wake)
    defer timer.Stop()
}
stop := context.AfterFunc(ctx, r.wake)
defer stop()
r.mu.Lock()
defer r.mu.Unlock()
if r.inFlight == 0 { ... return true }

---


In almost every normal reap pass, r.inFlight == 0 upon entry. However, two time.AfterFunc timers and one context.AfterFunc are allocated and immediately cancelled via defer.

Recommendation: Acquire r.mu.Lock() first. If r.inFlight == 0 && !r.reaping, set r.reaping = true and return immediately without allocating timers or context hooks. Only register timers if inFlight > 0 or r.reaping == true.

internal/childreap/childreap.go
```diff
@@ -0,0 +1,222 @@
...
+func (r *Reaper) enter() {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	for r.reaping || r.draining {
+		r.cond.Wait()
+	}
+	r.inFlight++
+}
+
+func (r *Reaper) leave() {
```
dberkov, line 96: Defensive Underflow Check in leave()

If a caller accidentally invokes leave twice (or calls leave without enter), inFlight decrements below 0. Subsequent calls could hit inFlight == 0 while operations are still running. Adding a guard like sync.WaitGroup (if r.inFlight < 0 { panic("childreap: negative inFlight count") }) prevents silent desynchronization.

cmd/ateom-gvisor/runsc.go
BenTheElder: This is pre-existing but we should fix it.

internal/childreap/childreap.go
```diff
@@ -0,0 +1,222 @@
+// Copyright 2026 Google LLC
```
dberkov, line 1: nit: //go:build linux  similar to  childreap_linux_test.go.

internal/childreap/childreap.go
```diff
@@ -0,0 +1,222 @@
...
+		// Another reap is in progress (it drained on our behalf); let it.
+		if r.reaping {
+			r.cond.Wait()
+			continue
+		}
+		now := time.Now()
+		if r.inFlight == 0 {
+			if !now.Before(drainAt) {
+				slog.WarnContext(ctx, "Reaped children only after holding off subprocesses",
+					slog.Duration("waited", MaxDefer))
```
dberkov: nit:  logs the constant MaxDefer (60s) rather than the actual duration elapsed

dberkov (commented):
minor comments

internal/childreap/childreap.go
BenTheElder: ACK, We don't expose `leave` directly so it would have to be a bug in this package, but worth guarding against.

internal/childreap/childreap.go
BenTheElder: I think actually the opposite, linux isn't the constraint here, we don't use anything linux specific. maybe constrain to unix. This suite passes on darwin.

cmd/ateom-gvisor/runsc.go
BenTheElder: https://github.com/agent-substrate/substrate/pull/1293/commits/8f41691c1a94f733955fd6a5dc725780ae4f84f6

internal/childreap/childreap.go
BenTheElder: 8f41691c

internal/childreap/childreap.go
BenTheElder: 8f41691c

cmd/ateom-microvm/internal/reaper/reaper_linux_test.go
BenTheElder: 8f41691c

BenTheElder:
Addressed comments.

PR #1295, opened 2026-08-28
benchmarking: Refactor boomer based Locust benchmarking

### LINKED ISSUE #1248: Rename boomer-glutton and extract shared benchmarking utilities for Locust based boomer worker tests
The boomer benchmarking infrastructure for Locust needs to be more modular and extensible. Currently, all boomer-based virtual users and test implementations are tightly coupled inside the glutton package. This refactoring should isolate distinct test workloads and expose generic utilities so that new boomer-based users that do not strictly fall under the Glutton umbrella can be easily added.

pmandewalkar:
Fixes #1248 

Renamed boomer-glutton to boomer-worker.
Extracted the boomer shared utils that future non-glutton workloads may use to the boomerutil package.
Updated durdir and glutton to import and use the changes.
- [ X] Tests pass
- [ X] Appropriate changes to documentation are included in the PR

### DIFF
--- a/cmd/benchmarking/boomer-worker/main.go
+++ b/cmd/benchmarking/boomer-worker/main.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// boomer-glutton is the Go re-implementation of the GluttonUser locust test.
+// boomer-worker is the Go re-implementation of the GluttonUser locust test.
 // It speaks the locust worker protocol via myzhan/boomer, so it appears as a
 // regular worker to the Python locust master while sidestepping gevent's
 // scheduling tax.
--- a/cmd/benchmarking/boomer-worker/main.go
+++ b/cmd/benchmarking/boomer-worker/main.go
@@ -28,12 +28,15 @@ import (
 	"strings"
 	"time"
 
+	"github.com/agent-substrate/substrate/internal/benchmarking/boomer/boomerutil"
 	"github.com/agent-substrate/substrate/internal/benchmarking/boomer/dynconfig"
-	"github.com/agent-substrate/substrate/internal/benchmarking/boomer/glutton"
 	bmetrics "github.com/agent-substrate/substrate/internal/benchmarking/boomer/metrics"
 	btrace "github.com/agent-substrate/substrate/internal/benchmarking/boomer/trace"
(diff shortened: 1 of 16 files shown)


cmd/benchmarking/boomer-worker/main.go
```diff
@@ -28,8 +28,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/agent-substrate/substrate/internal/benchmarking/boomer/boomerutil"
 	"github.com/agent-substrate/substrate/internal/benchmarking/boomer/dynconfig"
-	"github.com/agent-substrate/substrate/internal/benchmarking/boomer/glutton"
+	_ "github.com/agent-substrate/substrate/internal/benchmarking/boomer/glutton"
```
roycaihw: Is it still needed (e.g. doing some package level init)? If not, let's remove this import.

internal/benchmarking/boomer/boomerutil/client.go
```diff
@@ -40,11 +40,6 @@ const (
 )
 
 // DialControl opens an authenticated gRPC connection to ateapi.
-//
-// ateapi rejects calls that carry no credential, so present the pod
-// certificate — the same client auth the base install gives ate-controller
-// and atenet-router. ClientCredBundle re-reads the bundle on every handshake,
-// so rotations are picked up without a restart.
```
roycaihw, line 47: Let's keep the documentation. Maybe this gets dropped accidentally during a rebase?

roycaihw (approved):
LGTM

roycaihw:
cc @maxsmythe PTAL

cmd/benchmarking/boomer-worker/main.go
maxsmythe: lifecycle.go has an init function.

I think it might make sense to include init imports in their own commented import block.

maxsmythe (changes requested):
After two small nits, LGTM

cmd/benchmarking/boomer-worker/main.go
pmandewalkar: addressed!

maxsmythe (approved):
LGTM

roycaihw (approved):
LGTM

PR #1296, opened 2026-08-28
docs: add symptom-first troubleshooting guides

### LINKED ISSUE #1294: troubleshooting guide
https://docs.google.com/document/d/1B_WBlaXnQtbhAC_-CO78wb8EXip3Axt7usoQi73w3bA/edit?tab=t.0#heading=h.stmc47ehviff

JeffLuoo:
Fixes #1294

It's converted from [this public doc](https://docs.google.com/document/d/1B_WBlaXnQtbhAC_-CO78wb8EXip3Axt7usoQi73w3bA/edit?tab=t.0). Please feel free to drop comments in the doc and I can update this PR.

### DIFF
--- a/docs/troubleshooting/resumes-are-slow.md
+++ b/docs/troubleshooting/resumes-are-slow.md
@@ -18,14 +18,11 @@ Two states give a resume, and their cost is not the same:
 | Paused | On the node VM, and the resume prefers that node | Low. No download. |
 | Suspended | In object storage (GCS or S3) | High. A download and an unpack. |
 
-`ate.snapshot.kind` tells you which snapshot the resume read:
-
-| Value | Meaning |
-|---|---|
-| `local` | A snapshot on the node. A pause wrote it. |
-| `latest` | The durable snapshot of the actor. |
-| `golden` | The image of the ActorTemplate. This is the first activation of a new actor. |
-| `boot` | A start from nothing. A boot is not a restore, thus this value never occurs on the atelet histograms. |
+`ate.snapshot.kind` tells you which snapshot the resume read. The permitted
+values and their meaning are in the `registry.ate.snapshot` group of
+[the registry](../metrics/registry/metrics.yaml). One of them changes what you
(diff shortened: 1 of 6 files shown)


git286:
I noticed several drifts between this guide and the metrics definition (like the error code etc) and I think it will become more programatic in the future because it will be very hard to keep the guide the metrics definition synced.

I think we should make the guide very thin. The durable structure is: guide owns the decision tree and interpretation, registry owns definitions, and the guide links rather than paraphrases. Every value taxonomy or default it copies inline is future debt.

docs/troubleshooting/resumes-are-slow.md
```diff
@@ -0,0 +1,410 @@
...
+
+```promql
+sum by("ate.imagecache.outcome") (
+  rate({__name__="ate.imagecache.requests"}[5m]))
+```
+
+Calculate the hit ratio as `hit / (hit + miss)`. Keep `error`, `cancelled` and
+`timeout` out of the denominator.
+
+If the outcome is `error`, `error.type` holds the HTTP status of the registry.
```
git286: It claims ate.imagecache.requests error.type is only 401/403 (credential), 429 (rate limit), and _OTHER for everything else, but the registry (docs/metrics/registry/metrics.yaml:973-976) and internal/imagecache/metrics.go pass 404, 500, 502, 503, 504 through as literal values

JeffLuoo: fixed

JeffLuoo:
> I noticed several drifts between this guide and the metrics definition (like the error code etc) and I think it will become more programatic in the future because it will be very hard to keep the guide the metrics definition synced.
> 
> I think we should make the guide very thin. The durable structure is: guide owns the decision tree and interpretation, registry owns definitions, and the guide links rather than paraphrases. Every value taxonomy or default it copies inline is future debt.

Agreed, and updated to remove some metrics details from the guide.

PR #1300, opened 2026-08-28
api: drop ActorTemplateStatus.sandbox_assets for now

zoez7:
Remove `ActorTemplateStatus.sandbox_assets` along with the `SandboxAssets` messages it referenced. Nothing ever wrote the field: sandbox assets are resolved from the WorkerPool and SandboxConfig objects at resume time and travel to atelet via ateletpb, so freezing them into the template status never materialized.

This requires more thought, one option is to add it as a field of GoldenSnapshotStatus in the future, and make GoldenSnapshotStatus a repeated field to allow multiple goldens.


dberkov (dismissed):

BenTheElder:
needs rebase / codegen. also prefer rebase to merge so the commit history is cleaner (even after squash merge this is helpful for tracing back)

PR #1303, opened 2026-08-28
Declarative validation for ActorTemplate

shrutiyam-glitch:
Continues the declarative-validation (DV) migration from #1215, covering the `ActorTemplate` resource, its full spec tree, the create path, and read/delete verbs. 

**Key Changes**
* **DV Tags & Hooks**: Added validation tags and custom hooks across the `ActorTemplate` spec (`Metadata`, `SandboxConfig`, `SnapshotsConfig`, `Container`, `Volume`, `Resources`).
* **Create Path**: Replaced hand-written validation with generated validators in `CreateActorTemplate`. Custom rules (e.g., `on_commit ⊆ on_pause`) are now handled via custom hooks.
* **Read/Delete Verbs**: Converted `Get`, `List`, and `Delete` requests to use DV. 
* **State Updates**: Refactored `atepg` metadata setters to update in-place.
* Regenerated apitool exemptions (documented 15 previously-exempt fields).

**Deliberate Behavior Changes**
The gRPC path now strictly enforces CRD rules. Specific tightenings include:
* Empty `worker_selector` is now rejected.
* `EnvVar.name` is strictly required.
* Negative values in optional enum fields are rejected (`minimum=1`).
* `page_token` is capped at 256 characters (matching other list requests).

**Deliberately Not Done**
* **No update DV / status tags**: `ActorTemplates` are immutable to clients. The reconciler updates the status directly against the store, so tagging the status subtree or adding update validation isn't necessary right now.
* **`config_name` matching `sandbox_class`**: Enforcing this requires calling SandboxConfigLister (ServiceImpl), so it's left as a TODO.

**Testing**
* Added ~70 positive and negative test cases in the validator tables. 
* Store-backed tests (`TestCreateActorTemplate*`, `TestUpdateActorTemplateMetadata`, and the `atepg` suite) verified against real Postgres via testcontainers.

- [ ] Tests pass


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -617,41 +618,69 @@ enum SandboxClass {
...
 
   // worker_selector restricts which worker pools actors from this template
   // may use.
   //
-  // +k8s:opaqueType
+  // +k8s:optional
   Selector worker_selector = 2;
 
+  // +k8s:required # at least one container
+  // +k8s:listType=atomic
```
juli4n: We should probably set a max number of containers (`+k8s:maxItems`).

We should also give this repeated field map semantics, and designate the container `name` field as key. Something like `+k8s:listType=map` and `+k8s:listMapKey=name`

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -617,41 +618,69 @@ enum SandboxClass {
...
   repeated Container containers = 3;
 
   // +k8s:optional
+  // +k8s:listType=atomic
   repeated Volume volumes = 4;
```
juli4n, line 642: Same comment as in `containers` about max length and map semantics.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -617,41 +618,69 @@ enum SandboxClass {
...
+  //
+  // +k8s:required
   SandboxConfig sandbox_config = 6;
 
   // Resource usage configuration.
+  //
+  // +k8s:optional
   Resources resources = 7;
 
   ActorTemplateStatus status = 8;
```
juli4n, line 659: `+k8s:optional`.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -724,39 +774,64 @@ enum ResumeSource {
...
+  // +k8s:required
+  // TODO: validate that this is a well-formed image reference
   string image = 2;
 
   // Entrypoint array; when set, the image's ENTRYPOINT and CMD are both
   // ignored and the process argv is command + args. Unlike Kubernetes,
   // $(VAR_NAME) references are NOT expanded.
+  //
+  // +k8s:optional
+  // +k8s:listType=atomic
```
juli4n, line 799: IIUC there is no need for listType here as we don't want to give the field set or map semantics.

Do we need to also enforce a max number of items and a max length for each of them? e.g.  `+k8s:eachVal=+k8s:maxLength=<someValue>`

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -783,47 +858,83 @@ message Capabilities {
...
   // name may be any printable ASCII character except '='.
+  //
+  // +k8s:required
+  // TODO: validate the printable-ASCII-except-'=' rule
   string name = 1;
 
   // value is the literal value.
+  //
+  // +k8s:optional
   string value = 2;
```
juli4n, line 897: +k8s:maxLength=<length> ?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -724,39 +774,64 @@ enum ResumeSource {
...
+  //
+  // +k8s:optional
+  // +k8s:listType=atomic
   repeated string command = 3;
 
   // Arguments to the entrypoint; the image's CMD is used if unset (unless
   // command is set, which discards the image's CMD).
+  //
+  // +k8s:optional
+  // +k8s:listType=atomic
```
juli4n, line 808: Same comment as in `command`.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -724,39 +774,64 @@ enum ResumeSource {
...
   repeated string args = 4;
 
+  // +k8s:optional
+  // +k8s:listType=atomic
```
juli4n: Set a limit? Also question: do we want `env` to have map semantics and keyed on `EnvVar.name`?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -724,39 +774,64 @@ enum ResumeSource {
...
 
   // readyz is an optional HTTP readiness probe; when set the actor is not
   // ready until the endpoint returns 200.
+  //
+  // +k8s:optional
   ContainerReadyz readyz = 6;
 
+  // +k8s:optional
+  // +k8s:listType=atomic
+  // +k8s:maxItems=32
```
juli4n, line 832: Do we want this list to have map semantic? Should it be keyed on VolumeMount.name?

juli4n, line 832: Can you validate with @msau42 if this (32) is a number we feel comfortable with for now?

shrutiyam-glitch: Referred to the CRD and added the validations for most of the fields - https://github.com/agent-substrate/substrate/blob/a0c4ee46221ce86ff75256d442c2d6dc7b23465e/pkg/api/v1alpha1/actortemplate_types.go#L343

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -783,47 +858,83 @@ message Capabilities {
 // expanded with Kubernetes-style $(VAR) references.
 message EnvVar {
   // name may be any printable ASCII character except '='.
+  //
+  // +k8s:required
+  // TODO: validate the printable-ASCII-except-'=' rule
```
juli4n: Any reason to not add this in this PR?

shrutiyam-glitch: There is another TODO validation that needs to be in the .go files - https://github.com/agent-substrate/substrate/pull/1303/changes#diff-54487cd2c9a8b26ec0ca05b934fd914a6c261a8c068b2b9f00e7f95e2b970c6cR66
Since the PR was already big, thought of doing a follow up. 
But I guess for this, I can add in this PR itself. Will add it. Thanks.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -617,41 +618,69 @@ enum SandboxClass {
...
 }
 
 message Resources {
+  // limits is the maximum amount of compute resources allowed. Only "cpu"
+  // and "memory" are supported, each at most once, and each quantity must be
+  // greater than zero; the cpu limit must be less than 1000 cores.
+  //
+  // +k8s:optional
+  // +k8s:maxItems=2
+  // +k8s:listType=atomic
```
juli4n: Use `listType=map`, then I dont think you need `k8s:unique`.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -617,41 +618,69 @@ enum SandboxClass {
...
+  // +k8s:unique=map
+  // +k8s:listMapKey=name
+  // +k8s:customValidation # names, quantity parse/positivity, cpu bound
   repeated Limits limits = 1;
 }
 
 message Limits {
+  // +k8s:required
   string name = 1;
```
juli4n, line 678: Can we enforce here that only "cpu" or "memory" are set?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -724,39 +774,64 @@ enum ResumeSource {
...
+  // +k8s:optional
+  // +k8s:listType=atomic
+  // +k8s:maxItems=32
   repeated VolumeMount volume_mounts = 7;
 
   // security_context adjusts the container's security settings. Unset leaves
   // the default capability set.
+  //
+  // +k8s:optional
   SecurityContext security_context = 8;
```
zoez7, line 841: Can we also copy over the validation for SecurityContext field here https://github.com/agent-substrate/substrate/blob/9bddf66fdacd7e687a100e968e02f617462c3a52/pkg/api/v1alpha1/actortemplate_types.go#L267

shrutiyam-glitch: done. thanks!

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: Since `VolumeMount.name` is just a reference to a declared volume, it can be repeated,  while `mount_path` is the mount's identity (one mount per path). so keyed on `mount_path`.

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: enforced in the custom validation method above, for the `Resources` - `ValidateCustom_Resources_Limits` in `actor_template.go`.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -627,41 +628,75 @@ enum SandboxClass {
...
 
   // worker_selector restricts which worker pools actors from this template
   // may use.
   //
-  // +k8s:opaqueType
+  // +k8s:optional
   Selector worker_selector = 2;
 
+  // +k8s:required # at least one container
+  // +k8s:maxItems=10 # matches the CRD's containers bound
```
juli4n: Nit: let's remove the comment as the CRD is going away soon and it will get obsolete.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -627,41 +628,75 @@ enum SandboxClass {
...
+  // +k8s:listType=map
+  // +k8s:listMapKey=name
   repeated Container containers = 3;
 
   // +k8s:optional
+  // +k8s:maxItems=32 # matches the CRD's volumes bound
```
juli4n: Nit: let's remove the comment as the CRD is going away soon and it will get obsolete.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -730,46 +787,80 @@ enum ResumeSource {
...
+  // +k8s:maxLength=512 # matches ImageVolumeSource.reference's bound
+  // TODO: validate that this is a well-formed image reference
   string image = 2;
 
   // Entrypoint array; when set, the image's ENTRYPOINT and CMD are both
   // ignored and the process argv is command + args. Unlike Kubernetes,
   // $(VAR_NAME) references are NOT expanded.
+  //
+  // +k8s:optional
+  // +k8s:maxItems=64 # matches the CRD's command bound
```
juli4n: Nit: let's remove the comment as the CRD is going away soon and it will get obsolete.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -730,46 +787,80 @@ enum ResumeSource {
...
+  // +k8s:eachVal=+k8s:maxLength=4096 # argv strings; guardrail, not a contract
   repeated string command = 3;
 
   // Arguments to the entrypoint; the image's CMD is used if unset (unless
   // command is set, which discards the image's CMD).
+  //
+  // +k8s:optional
+  // +k8s:maxItems=64 # matches the CRD's args bound
```
juli4n: Nit: let's remove the comment as the CRD is going away soon and it will get obsolete.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -730,46 +787,80 @@ enum ResumeSource {
...
+  // +k8s:eachVal=+k8s:maxLength=4096 # argv strings; guardrail, not a contract
   repeated string args = 4;
 
+  // +k8s:optional
+  // +k8s:maxItems=32 # matches the CRD's env bound
```
juli4n: Nit: let's remove the comment as the CRD is going away soon and it will get obsolete.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -730,46 +787,80 @@ enum ResumeSource {
...
   // ready until the endpoint returns 200.
+  //
+  // +k8s:optional
   ContainerReadyz readyz = 6;
 
+  // +k8s:optional
+  // +k8s:maxItems=32
+  // +k8s:listType=map # a volume may be mounted at several paths, so the
+  // path, not the name, is the key
+  // +k8s:listMapKey=mount_path
```
juli4n: Shouldn't this be name? That's what defines the ID of the volume, right? We can add a validation that mount paths don't clash, but the key of the map seems to be the name.

pkg/proto/ateapipb/ateapi.proto
juli4n: I don't think the DV framework will understand those annotations.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -778,58 +869,107 @@ message SecurityContext {
 // granted. Capabilities are named without the "CAP_" prefix; "ALL" in drop
 // removes the whole default set.
 message Capabilities {
-  // add lists capabilities to grant on top of the default set.
+  // add lists capabilities to grant on top of the default set. "ALL" is
+  // rejected: name the individual capabilities the container needs.
+  //
+  // +k8s:optional
+  // +k8s:maxItems=64 # matches the CRD's bound
```
juli4n: Nit: let's remove the comment as the CRD is going away soon and it will get obsolete.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -778,58 +869,107 @@ message SecurityContext {
...
+  // +k8s:customValidation # capability grammar; "ALL" not accepted
   repeated string add = 1;
 
-  // drop lists capabilities to remove from the default set.
+  // drop lists capabilities to remove from the default set. "ALL" drops the
+  // whole set, so drop+add expresses an exact set rather than a relative one.
+  //
+  // +k8s:optional
+  // +k8s:maxItems=64 # matches the CRD's bound
```
juli4n: Nit: let's remove the comment as the CRD is going away soon and it will get obsolete.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -778,58 +869,107 @@ message SecurityContext {
...
   repeated string drop = 2;
 }
 
 // EnvVar supplies one environment variable to a container. Values are not
 // expanded with Kubernetes-style $(VAR) references.
 message EnvVar {
   // name may be any printable ASCII character except '='.
+  //
+  // +k8s:required
+  // +k8s:maxLength=256 # guardrail; the CRD does not bound env names
```
juli4n: Nit: let's remove the comment as the CRD is going away soon and it will get obsolete.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -730,46 +787,80 @@ enum ResumeSource {
...
   string image = 2;
 
   // Entrypoint array; when set, the image's ENTRYPOINT and CMD are both
   // ignored and the process argv is command + args. Unlike Kubernetes,
   // $(VAR_NAME) references are NOT expanded.
+  //
+  // +k8s:optional
+  // +k8s:maxItems=64 # matches the CRD's command bound
+  // +k8s:eachVal=+k8s:maxLength=4096 # argv strings; guardrail, not a contract
   repeated string command = 3;
```
juli4n, line 801: +k8s:listType=atomic (all repeated fields *must* have the +k8s:listType tag).

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -730,46 +787,80 @@ enum ResumeSource {
...
 
   // Arguments to the entrypoint; the image's CMD is used if unset (unless
   // command is set, which discards the image's CMD).
+  //
+  // +k8s:optional
+  // +k8s:maxItems=64 # matches the CRD's args bound
+  // +k8s:eachVal=+k8s:maxLength=4096 # argv strings; guardrail, not a contract
   repeated string args = 4;
```
juli4n, line 810: +k8s:listType=atomic (all repeated fields *must* have the +k8s:listType tag).

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -778,58 +869,107 @@ message SecurityContext {
...
 // removes the whole default set.
 message Capabilities {
-  // add lists capabilities to grant on top of the default set.
+  // add lists capabilities to grant on top of the default set. "ALL" is
+  // rejected: name the individual capabilities the container needs.
+  //
+  // +k8s:optional
+  // +k8s:maxItems=64 # matches the CRD's bound
+  // +k8s:customValidation # capability grammar; "ALL" not accepted
   repeated string add = 1;
```
juli4n, line 871: Should `add` and `drop` have +k8s:listType=set?

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: So, I guess by default for all the lists, listType is `atomic`. However, to avoid any ambiguity, added it.

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: added a custom validation method to check this - https://github.com/agent-substrate/substrate/pull/1303/changes#diff-54487cd2c9a8b26ec0ca05b934fd914a6c261a8c068b2b9f00e7f95e2b970c6cR319-R324

pkg/proto/ateapipb/ateapi.proto
shrutiyam-glitch: k8s `VolumeMounts` are keyed by `mountPath` - https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3239-L3243 
However, for simplicity, we keying by volume name. But, this incorrectly validates the case when volume names are the different but the `mount_path` are the same. Added a TODO for the same - https://github.com/agent-substrate/substrate/pull/1303/changes/43fc4681da7bacb10762c66f3833606b21af4fdd#diff-f29aa971d0e861ade55215208c2d5e2eaa7d9f743969517826128207b1e4095eR842-R845

@msau42 could you please confirm this behavior? 

cc: @juli4n

msau42: Yeah for now lets keep it simple with one mount per volume. I vaguely recall that in Kubernetes there were some drivers/scenarios where having multiple mount paths on the same volume didn't work.

juli4n:
This PR is getting large and it adds a significant set of validations. I would say let's merge and iterate. If we find (we will :) ) missing validations, we can add them in follow up PRs.

juli4n (approved):

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -240,44 +165,163 @@ func (s *ServiceImpl) DeleteActorTemplate(ctx context.Context, templateRef resou
...
+		}
+		if limit.Name == "cpu" && q.Cmp(cpuLimitMax) >= 0 {
+			errs = append(errs, field.Invalid(fldPath.Index(i).Child("quantity"), limit.Quantity, "cpu limit must be less than 1000 cores"))
 		}
 	}
+	return errs
+}
 
+// ValidateCustom_ActorTemplate_SnapshotsConfig mirrors the ActorTemplate
+// CRD's CEL rule: on_commit must be a subset of on_pause. UNSPECIFIED means
```
thockin, line 274: Why are we allowing UNSPECIFIED to mean full - why not demand a value?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -611,41 +613,75 @@ enum SandboxClass {
...
+  // limits is the maximum amount of compute resources allowed. Only "cpu"
+  // and "memory" are supported, each at most once, and each quantity must be
+  // greater than zero; the cpu limit must be less than 1000 cores.
+  //
+  // +k8s:optional
+  // +k8s:maxItems=2
+  // +k8s:listType=map
+  // +k8s:listMapKey=name
+  // +k8s:customValidation # names, quantity parse/positivity, cpu bound
   repeated Limits limits = 1;
```
thockin, line 672: This is more about API design than validation, but:

K8s did this because we allow 3rd party resources.  It seems possible that we will ALSO allow those (e.g. GPU) -- is "map of string to quantity" what we want?

Alternative #1: `map[string]string` where keys are validated and values are quantities.

Alternative #2: `message Limits { cpu string=1; memory string=2}`

@juli4n

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -671,28 +706,49 @@ message ActorTemplateStatus {
...
   // supplying the sandbox binaries. Required; must match sandbox_class.
+  //
+  // +k8s:required
+  // +k8s:format=k8s-long-name
   string config_name = 2;
 }
 
 message SnapshotsConfig {
-  // on_pause selects what is captured during pause actor.
+  // on_pause selects what is captured during pause actor. UNSPECIFIED is
```
thockin, line 724: Why do we need to allow UNSPECIFIED?  We don't really have a compat problem yet.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -714,46 +771,88 @@ enum ResumeSource {
...
+  // +k8s:optional
+  // +k8s:maxItems=64
+  // +k8s:listType=atomic
+  // +k8s:eachVal=+k8s:maxLength=4096 # argv strings; guardrail, not a contract
   repeated string args = 4;
 
+  // +k8s:optional
+  // +k8s:maxItems=32
+  // +k8s:listType=map # each variable is set at most once
+  // +k8s:listMapKey=name
```
thockin, line 815: @juli4n should we use actual maps for things that are fundamentally map-like ?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -762,58 +861,109 @@ message SecurityContext {
...
 
 // ContainerReadyz configures the readiness signal for a container.
 message ContainerReadyz {
   // http_get specifies the HTTP request to perform. Required.
+  //
+  // +k8s:required
   HTTPGetAction http_get = 1;
 
   // timeout_seconds bounds how long to poll http_get before failing the
   // actor start. 0 means the server-applied default (30s).
```
thockin, line 908: @juli4n I asked elsewhere in a different PR - do you think we should keep this semantic or should we reify default values when they are stored?  If we do not, then we can never really change a default.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -762,58 +861,109 @@ message SecurityContext {
...
   // image mounts the contents of an OCI image, read-only.
+  //
+  // +k8s:optional
+  // +k8s:unionMember
   ImageVolumeSource image = 6;
 
   // type discriminates the source union: "DurableDir",
   // "ExternalVolumeTemplate", "Image" or "SystemInfo".
+  //
+  // +k8s:optional
```
thockin, line 963: @juli4n We discussed this one - I thought we agreed to not use discriminators - did I misremember?  Optional discrims are weird

thockin (commented):
I have a few comments I will send as a followup PR.  Some questions about API direction for @juli4n :)

PR #1304, opened 2026-08-28
Replace actor template namespace/name pair with ActorTemplate atespace and name pair in internal APIs

zoez7:
Both Atelet and Ateom internal APIs consume actor template namespace/name pair to telemetry. Updated them to use the new Substrate ActorTemplate proto instead of the legacy CRD one.

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -511,17 +511,17 @@ func (s *AteomHerder) Run(ctx context.Context, req *ateletpb.RunRequest) (resp *
 	// Tell ateom to start the workload. gVisor uses RunscPath; the micro-VM
 	// runtime uses the full RuntimeAssetPaths set.
 	if _, err := client.RunWorkload(ctx, &ateompb.RunWorkloadRequest{
-		Atespace:               actorRef.Atespace,
-		ActorName:              actorRef.Name,
-		ActorTemplateNamespace: req.GetActorTemplateNamespace(),
-		ActorTemplateName:      req.GetActorTemplateName(),
-		RunscPath:              runscPathFor(assetPaths),
-		RuntimeAssetPaths:      assetPaths,
-		Spec:                   spec,
-		ActorUid:               actorUID,
-		EgressGateway:          toAteomEgressGateway(req.GetEgressGateway()),
-		CpuMilli:               req.GetCpuMilli(),
-		MemoryBytes:            req.GetMemoryBytes(),
+		Atespace:              actorRef.Atespace,
+		ActorName:             actorRef.Name,
+		ActorTemplateAtespace: req.GetActorTemplateAtespace(),
+		ActorTemplateName:     req.GetActorTemplateName(),
+		RunscPath:             runscPathFor(assetPaths),
+		RuntimeAssetPaths:     assetPaths,
+		Spec:                  spec,
+		ActorUid:              actorUID,
+		EgressGateway:         toAteomEgressGateway(req.GetEgressGateway()),
+		CpuMilli:              req.GetCpuMilli(),
+		MemoryBytes:           req.GetMemoryBytes(),
 	}); err != nil {
 		return nil, fmt.Errorf("while calling ateom.RunWorkload: %w", err)
 	}
@@ -548,7 +548,7 @@ func initSnapshotSizeMetric() error {
 // recordSnapshotSize labels each image with the registry's file.name. That
 // label used to be spelled "kind", which means the snapshot's provenance
 // everywhere else in the ate.* namespace, not one of its files.
-func recordSnapshotSize(ctx context.Context, file, path, atNamespace, atName string) {
+func recordSnapshotSize(ctx context.Context, file, path, templateAtespace, templateName string) {
 	if snapshotSizeBytes == nil {
 		return
 	}
@@ -563,8 +563,8 @@ func recordSnapshotSize(ctx context.Context, file, path, atNamespace, atName str
 	}
 	snapshotSizeBytes.Record(ctx, fi.Size(), metric.WithAttributes(
 		semconv.FileNameKey.String(file),
-		ateattr.TemplateNamespaceKey.String(atNamespace),
-		ateattr.TemplateNameKey.String(atName),
+		ateattr.TemplateNamespaceKey.String(templateAtespace),
+		ateattr.TemplateNameKey.String(templateName),
 	))
 }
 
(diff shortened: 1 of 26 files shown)


internal/proto/ateompb/ateom.proto
```diff
@@ -104,13 +104,22 @@ service Ateom {
...
+// ObjectRef references a Substrate resource by its (atespace, name) identity.
+// Mirrors ateapipb.ObjectRef, which internal protos do not import.
+message ObjectRef {
+  string atespace = 1;
+  string name = 2;
+}
+
 message TerminateWorkloadRequest {
+  reserved 4, 5;
+  reserved "actor_template_namespace", "actor_template_name";
```
juli4n: Don't reserve fields. Just drop them and reuse the IDs. We are breaking the API anyways.

internal/proto/ateletpb/atelet.proto
```diff
@@ -62,15 +62,24 @@ service AteomHerder {
   rpc Terminate(TerminateRequest) returns (TerminateResponse) {}
 }
 
+// ObjectRef references a Substrate resource by its (atespace, name) identity.
+// Mirrors ateapipb.ObjectRef, which internal protos do not import.
+message ObjectRef {
```
juli4n: ObjectRef is a concept of the control plane API (mgmt plane) where we need a way for resources to reference another resources. We could introduce the same concept in the atelet API (control plane), although I'm not sure how useful that is. 

If we adopt it, we should use it consistently, for example, we have `atespace`, `actor_name` , which could be an `ObjectRef` too, but for actors we also have an `actor_uid` which isn't in the ObjectRef (I guess that could be a separate top level field alongside `actor` ?

```
ObjectRef actor;
string actor_uid;
ObjectReg  actor_template;
```

cmd/ateapi/internal/controlapi/workflow.go
```diff
@@ -146,6 +147,16 @@ type workerWorkflowStore interface {
...
 
+// ateletTemplateRef converts the actor's substrate template reference for the
+// atelet API; nil when the actor carries none.
+func ateletTemplateRef(actor *ateapipb.Actor) *ateletpb.ObjectRef {
+	ref := actor.GetActorTemplate()
+	if ref == nil {
+		return nil
+	}
+	return &ateletpb.ObjectRef{Atespace: ref.GetAtespace(), Name: ref.GetName()}
+}
```
krisztianfekete: This drops the CRD fallback, so actors now might send nil and `ate.template.*` becomes empty on all the atelet/ateom metrics and actor logs. 

Here:
https://github.com/agent-substrate/substrate/blob/698289452b33bf27da43a115594815278e8a55fd/cmd/ateapi/internal/controlapi/workflow_resume.go#L519-L526
we already do the if/else against actorTemplateObjectRef, can we do the same until the legacy one is gone?

cmd/atelet/main.go
```diff
@@ -581,8 +580,8 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	tStart := time.Now()
 	var dAssets, dAteom, dPersist time.Duration
 	op := snapshotOp{
-		templateNamespace: req.GetActorTemplateNamespace(),
-		templateName:      req.GetActorTemplateName(),
+		templateNamespace: req.GetActorTemplate().GetAtespace(),
```
krisztianfekete: `ActorMetricAttributes` coerces an empty template to `TemplateUnknown`, this, and at a few other places below emit `""`. Can we pick one to make it consistent?

cmd/atelet/main.go
```diff
@@ -563,8 +562,8 @@ func recordSnapshotSize(ctx context.Context, file, path, atNamespace, atName str
 	}
 	snapshotSizeBytes.Record(ctx, fi.Size(), metric.WithAttributes(
 		semconv.FileNameKey.String(file),
-		ateattr.TemplateNamespaceKey.String(atNamespace),
-		ateattr.TemplateNameKey.String(atName),
+		ateattr.TemplateNamespaceKey.String(templateRef.GetAtespace()),
+		ateattr.TemplateNameKey.String(templateRef.GetName()),
```
krisztianfekete: Note that `ate.template.namespace` is documented like this:
https://github.com/agent-substrate/substrate/blob/698289452b33bf27da43a115594815278e8a55fd/docs/metrics/registry/metrics.yaml#L74-L78
but it's an atespace now. Weaver only checks spellings so `verify/metrics.sh` stays green either way, we should fix the brief in this PR.

cmd/atelet/sandbox_assets.go
```diff
@@ -86,11 +93,10 @@ type sandboxAssetsRecord struct {
...
 	// persistence is unavailable.
-	Atespace               string `json:"atespace,omitempty"`
-	ActorName              string `json:"actorName,omitempty"`
-	ActorUID               string `json:"actorUid,omitempty"`
-	ActorTemplateNamespace string `json:"actorTemplateNamespace,omitempty"`
-	ActorTemplateName      string `json:"actorTemplateName,omitempty"`
+	Atespace      string               `json:"atespace,omitempty"`
+	ActorName     string               `json:"actorName,omitempty"`
+	ActorUID      string               `json:"actorUid,omitempty"`
+	ActorTemplate *actorTemplateRecord `json:"actorTemplate,omitempty"`
```
krisztianfekete: Can we lose template identity before/after this PR? Probably fine at this point though.

cmd/atelet/statspoller.go
```diff
@@ -287,8 +287,8 @@ func (p *statsPoller) collect(ctx context.Context) map[templateKey]*templateAggr
 			}
 
 			key := templateKey{
-				templateNamespace: sample.GetActorTemplateNamespace(),
-				templateName:      sample.GetActorTemplateName(),
+				templateNamespace: sample.GetActorTemplate().GetAtespace(),
+				templateName:      sample.GetActorTemplate().GetName(),
```
krisztianfekete: If the ref is nil every actor collapses into one `("", "")` bucket here, so the per-template memory/cpu gauges merge instead of just going unlabeled.  Similar issue to #761

internal/proto/ateompb/ateom.proto
zoez7: Done. I renamed "actor_template_namespace" to "actor_template_atespace" instead.

internal/proto/ateletpb/atelet.proto
zoez7: Dropped ObjectRef concept in internal API for now, since the internal APIs only need the `ActorTemplate`'s name and atepsace for telemetry for now.

cmd/ateapi/internal/controlapi/workflow.go
zoez7: I restricted the scope of this PR to just renaming the actor template "Namespace" to "Atespace" in internal APIs. Will fix this in a follow up PR.
The legacy one will be gone in the next 2 days. I'll have a PR to do the cutover so we won't have to maintain 2 branches.

cmd/atelet/sandbox_assets.go
zoez7: Yes, during the switch there are some places where the template might log empty. I'll fix it as part of the full cutover.

cmd/atelet/main.go
zoez7: The `TemplateUnknown` was a temporary hack to fix e2e test during the transition period. Will have a separate PR to use the new substrate atespace everywhere.

juli4n (approved):

PR #1311, opened 2026-08-29
benchmarking: move glutton's internals into `internal/benchmarking/glutton`

sairajp-rewind:
#### What is this PR and Why we need it ?
Glutton's implementation lived in `cmd/benchmarking/glutton/main.go`, which Go's
`internal/` rule makes importable only from `cmd/benchmarking/glutton/...`. That is
why `internal/benchmarking/glutton/fake/` existed as a subdirectory with no parent
package: the fake could not import the thing it was faking. It duplicated the route
strings instead, and so did boomer.

This moves the implementation to `internal/benchmarking/glutton/`, where the fake sits
next to it and can share its constants. `main.go` keeps flag parsing and boot. No behavior change.

#### Commits

**1. Move the core out of `cmd/`.** `main.go` → `internal/benchmarking/glutton/glutton.go`
verbatim, plus `sizes.go` / `sizes_test.go` relocated from where #1130 landed them.
`newGluttonService` → `New`, `handlerForMode` → `Handler`; the proto import is aliased
`gluttonpb` per repo convention.

**2. Split the package and export the routes.** The 640-line file becomes `glutton.go`
(identity and route consts), `server.go`, `service.go`, `metrics.go`, `gossip.go`; tests
split into `server_test.go` and `service_test.go`. `fake/server.go` now aliases the real
route constants rather than restating them, so the stand-in cannot answer a path the
actor does not.

#### Validation

- [x] `make verify`
- [x] `gofmt -l ./cmd/benchmarking ./internal/benchmarking` (clean)
- [x] `go build ./...`
- [x] `go vet ./cmd/benchmarking/... ./internal/benchmarking/...`
- [x] `go test ./cmd/benchmarking/... ./internal/benchmarking/...`
- [x] glutton smoke test on cluster
- [x] Rename detection preserved on all moved files (`git diff -M` shows
      `{cmd => internal}` for `sizes.go`, `sizes_test.go`)


maxsmythe (dismissed):
LGTM

maxsmythe:
Marked as draft, cannot merge

maxsmythe (approved):
LGTM

PR #1313, opened 2026-08-29
localjwtauthority: Prep for rotation support

ahmedtd:
* Add a RefreshingPool abstraction like localca.RefreshingPool.  This provides caching (so we don't read the keys fresh from the filesystem) for each signing operation.
* Extend ConcretePool so that it tracks which key is currently active for signing, instead of always picking the first key.

A followup PR will add admin commands for actually rotating the JWT pool.


cmd/ateapi/main.go
```diff
@@ -193,10 +194,15 @@ func main() {
 
 	actorIDCAPool, err := localca.NewRefreshingPool(*actorIDCAPoolFile)
 	if err != nil {
-		serverboot.Fatal(ctx, "while loading the Actor ID CA", err)
+		serverboot.Fatal(ctx, "while loading the Actor ID certificate authority pool: %w", err)
```
thompsonmax: [nit] `%w` is printed verbatim because we're calling `Fatal` rather than `Fatalf`

cmd/ateapi/main.go
```diff
@@ -193,10 +194,15 @@ func main() {
...
 	}
 
-	actorIdentitySrv := actoridentity.New(actorIdentityJWTIssuer, *actorIDJWTPoolFile, actorIDCAPool, persistence, workerCache)
+	actorIDJWTAuthorityPool, err := localjwtauthority.NewRefreshingPool(*actorIDJWTPoolFile)
+	if err != nil {
+		serverboot.Fatal(ctx, "while loading the Actor ID JWT authority pool: %w", err)
```
thompsonmax: [nit] %w is printed verbatim because we're calling Fatal rather than Fatalf

internal/localjwtauthority/localjwtauthority.go
```diff
@@ -20,37 +20,266 @@ import (
...
+		for _, authority := range p.Authorities {
+			if authority.ID == p.ActiveForSigning {
+				selectedAuthority = authority
+			}
+		}
+		if selectedAuthority == nil {
+			return "", fmt.Errorf("selected authority %q not present", p.ActiveForSigning)
+		}
+	} else {
+		// Fall back to first entry.
```
thompsonmax: Should we have some TODO on removing this fall back at some point? (similar to the fallback on the CA pool impl)

internal/localjwtauthority/localjwtauthority.go
```diff
@@ -20,37 +20,266 @@ import (
...
+	if err != nil {
+		return "", fmt.Errorf("while converting claims to wire model: %w", err)
+	}
+
+	payloadBytes, err := json.Marshal(wireClaims)
+	if err != nil {
+		return "", fmt.Errorf("while marshaling payload: %w", err)
+	}
+
+	// TODO(ahmedtd): Select authority
```
thompsonmax: [nit] stale TODO? Seems like you're selecting the authority here :)

internal/localjwtauthority/localjwtauthority_test.go
```diff
@@ -15,48 +15,83 @@
...
+	ca1, err := GenerateECDSAP256Authority("1")
 	if err != nil {
-		t.Fatalf("GenerateKey(): %v", err)
+		t.Fatalf("Unexpected error generating CA 1: %v", err)
 	}
-	keyDER, err := x509.MarshalECPrivateKey(key)
-	if err != nil {
-		t.Fatalf("MarshalECPrivateKey(): %v", err)
+	pool1 := &ConcretePool{
+		Authorities:      []*Authority{ca1},
```
thompsonmax, line 33: There's no tests here that exercise a pool that has multiple authorities set, unless I'm missing something. Should we add unit tests for that?

internal/localjwtauthority/localjwtauthority.go
```diff
@@ -20,37 +20,266 @@ import (
...
 )
 
-type Pool struct {
+// Pool is the interface for a JWT signing pool.
+//
+// Logically, a Pool is a collection of multiple authorities.  One or more are
+// designated as active for signing.  The rest are inactive, but are still
+// trusted for verifying JWTs.
+//
+// The active/inactive desngination allows a Pool to be seamlessly rotated.
```
thompsonmax: [typo] 'desngination' -> 'designation'

thompsonmax (commented):
Mostly LGTM, just some nits and potential testing gap

cmd/ateapi/main.go
ahmedtd: Done.

cmd/ateapi/main.go
ahmedtd: Done.

internal/localjwtauthority/localjwtauthority.go
ahmedtd: Done.

internal/localjwtauthority/localjwtauthority.go
ahmedtd: Whoops, done.

internal/localjwtauthority/localjwtauthority.go
ahmedtd: Done.

internal/localjwtauthority/localjwtauthority_test.go
ahmedtd: I've made this test case simulate a full rotation.

thompsonmax (approved):

thompsonmax:
Approved, waiting on CI tests to merge.

PR #1315, opened 2026-08-29
Identity: Move actor JWT/cert minting into the main API

ahmedtd:
This was originally a separate service to make it easy to apply separate authentication and authorization interceptors.

It now seems clear that our authn/z framework will be strong enough to support atelet and external callers in one system (based on OpenFGA).

This change moves the MintJWT and MintCert RPCs into the control API, and removes some inline authz checks that will be handled by our unified authorizer framework.


ahmedtd:
Whoops, this isn't ready to review yet.

ahmedtd:
Ready for review now

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1438,6 +1451,106 @@ message GetTagRequest {
   ObjectRef tag = 1;
 }
 
+message MintActorJWTRequest {
```
thompsonmax, line 1465: [nit] missing doc comment here and a few places below (looks like tests are complaining about this)

cmd/ateom-gvisor/main.go
```diff
@@ -1064,14 +1064,14 @@ func (s *AteomService) prepareActorEgress(ctx context.Context, actorUID string,
 		SocketPath:           ateompath.CredentialBrokerSocket,
 		CredentialBundlePath: s.workerCredentialBundlePath,
 		TrustBundlePath:      s.podIdentityTrustBundlePath,
-		ExpectedActorUID:     actorUID,
+		ActorUID:             actorUID,
```
thompsonmax, line 1089: IIUC we also need to set actor atespace and name here, otherwise the check in https://github.com/agent-substrate/substrate/pull/1315/changes#diff-7888c14ca7ad7f062e76770497045d59a0a9dd4c4a8c967c95e403c1fcab0954R74 fails

cmd/ateom-microvm/main.go
```diff
@@ -536,14 +536,15 @@ func (s *AteomService) prepareActorEgress(ctx context.Context, actorUID string,
 		SocketPath:           ateompath.CredentialBrokerSocket,
 		CredentialBundlePath: s.workerCredentialBundlePath,
 		TrustBundlePath:      s.podIdentityTrustBundlePath,
-		ExpectedActorUID:     actorUID,
+
+		ActorUID: actorUID,
```
thompsonmax: Also need to set actor atespace and name here, otherwise the check in https://github.com/agent-substrate/substrate/pull/1315/changes#diff-7888c14ca7ad7f062e76770497045d59a0a9dd4c4a8c967c95e403c1fcab0954R74 fails

cmd/ateapi/internal/controlapi/functionaltest/actor_test.go
```diff
@@ -3459,3 +3462,35 @@ func TestDeleteActor_ReleasesAnAssignmentTheActorDoesNotReference(t *testing.T)
...
+			},
+			ActorTemplate:  &ateapipb.ObjectRef{Atespace: testAtespace, Name: "tmpl1"},
+			WorkerSelector: &ateapipb.Selector{MatchLabels: map[string]string{"tier": "free"}},
+			Status:         &ateapipb.ActorStatus{State: ateapipb.ActorState_ACTOR_STATE_RUNNING},
+		},
+	})
+	if err != nil {
+		t.Fatalf("CreateActor failed: %v", err)
+	}
+	_, err = tc.client.MintActorJWT(t.Context(), &ateapipb.MintActorJWTRequest{
```
thompsonmax, line 3493: Should we check the error here?

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -537,3 +547,158 @@ func ValidateCustom_UpdateActorRequest_Actor(ctx context.Context, op operation.O
 	errs = append(errs, validate.RequiredValue(ctx, op, fldPath.Child("metadata", "atespace"), &actor.Metadata.Atespace, nil)...)
 	return errs
 }
+
+func (s *RPCService) MintActorJWT(ctx context.Context, req *ateapipb.MintActorJWTRequest) (*ateapipb.MintActorJWTResponse, error) {
```
thompsonmax, line 551: The old RPC handlers called these auto-generated validation functions like so:

```

func validateMintJWTRequest(ctx context.Context, req *ateapipb.MintJWTRequest) field.ErrorList {
	// Call the generated validation.
	op := operation.Operation{Type: operation.Create}
	return controlapi.Validate_MintJWTRequest(ctx, op, nil, req, nil)
}

func validateMintCertRequest(ctx context.Context, req *ateapipb.MintCertRequest) field.ErrorList {
	// Call the generated validation.
	op := operation.Operation{Type: operation.Create}
	return controlapi.Validate_MintCertRequest(ctx, op, nil, req, nil)
}
```

I believe we need to copy that over to enforce the tag-based validation?

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -537,3 +547,158 @@ func ValidateCustom_UpdateActorRequest_Actor(ctx context.Context, op operation.O
...
+		return nil, status.Error(codes.NotFound, "actor not found")
+	} else if err != nil {
+		return nil, fmt.Errorf("while retrieving actor: %w", err)
+	}
+	if dbActor.GetMetadata().GetUid() != req.GetActorUid() {
+		return nil, status.Error(codes.Aborted, "conflict; actor has been deleted and recreated")
+	}
+
+	// We only issue tokens with audience bindings.
+	if len(req.GetAudience()) == 0 {
```
thompsonmax, line 574: The old handler also validated the issuer (`caller.Issuer != s.actorIdentityJWTIssuer`), should we do that in the new handler as well?

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -537,3 +547,158 @@ func ValidateCustom_UpdateActorRequest_Actor(ctx context.Context, op operation.O
...
+
+func (s *RPCService) MintActorJWT(ctx context.Context, req *ateapipb.MintActorJWTRequest) (*ateapipb.MintActorJWTResponse, error) {
+	// TODO(authz): Authorization layer needs to check whether the caller has
+	// the mintActorJWT permission/relation with this actor.  This could be an
+	// atelet (via the relationship of the atelet running the actor), or the
+	// egress gateway (via a cluster-level grant?)
+
+	// Verify that this actor exists in the store.  It doesn't need to be
+	// running, since we may need to issue JWTs during actor boot / resume.
+	dbActor, err := s.impl.GetActor(ctx, resources.ActorRefFromObjectRef(req.GetActor()))
```
thompsonmax, line 563: Hmm, should we check that the actor is in a RUNNING state (or in some set of valid states) before minting tokens or certs for it?

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -537,3 +547,158 @@ func ValidateCustom_UpdateActorRequest_Actor(ctx context.Context, op operation.O
...
+	if errors.Is(err, store.ErrNotFound) {
+		return nil, status.Error(codes.NotFound, "actor not found")
+	} else if err != nil {
+		return nil, fmt.Errorf("while retrieving actor: %w", err)
+	}
+	if dbActor.GetMetadata().GetUid() != req.GetActorUid() {
+		return nil, status.Error(codes.Aborted, "conflict; actor has been deleted and recreated")
```
thompsonmax, line 570: Looks like egress.renew() only stops retrying on FailedPrecondition or PermissionDenied: https://github.com/agent-substrate/substrate/blob/2fcfa64a73adc682752fd08b75be20aee6cd533f/internal/atunnel/egress.go#L148

Should we add `Aborted` to the list of codes checked there, or adjust the code returned here, to stop it from retrying?

cmd/ateapi/internal/controlapi/functionaltest/actor_test.go
ahmedtd: Done.

cmd/ateapi/internal/controlapi/actor.go
ahmedtd: Good catch.  Done.

cmd/ateapi/internal/controlapi/actor.go
ahmedtd: I am not sure.  I think it's likely that components will need to issue credentials even if the actor is actively running (think about a frontend harness service that might need to access files, etc with actor identity, even if the LLM harness bits inside the actor proper are not currently running).

I think "not deleted" is the safest for now.

cmd/ateapi/internal/controlapi/actor.go
ahmedtd: FailedPrecondition is like, the one single code you are *not* supposed to retry on.  It means you need to fix something about your request or the state of the system.

I'll change that to Aborted.

cmd/ateapi/internal/controlapi/actor.go
ahmedtd: That didn't make any sense -- you shouldn't have to authenticate with an Actor JWT in order to mint an Actor JWT.  I would think typically, the thing calling this RPC would be the egress gateway, and it would be authenticating with a k8s client certificate.

cmd/ateom-gvisor/main.go
ahmedtd: You're right, fixed.  (Also in ateom-microvm)

cmd/ateom-microvm/main.go
ahmedtd: Done.

pkg/proto/ateapipb/ateapi.proto
ahmedtd: Done.

thompsonmax (approved):

cmd/ateapi/internal/controlapi/actor.go
ahmedtd: Whoops, I read that backwards.  As you point out, those are the codes it's *not* retrying on. I've added Aborted to that code list.

PR #1319, opened 2026-08-29
ateclient: spread RPCs over every ateapi replica

BenTheElder:
ateapi is a headless Service, so its DNS name resolves to one address per replica and the client picks. This one asked for no policy, which means gRPC's pick_first: it connects to whichever address answers first and sends everything there for the life of the connection.

So a client talks to one replica however many are running, and adding replicas moves no load. Measured with eight load generators against two replicas, one took 97% of the traffic.

internal/ateapiauth, which atelet and the controllers dial through, has asked for round_robin all along. This is the same policy for the clients that do not.

Pulled out from #1266, I ran into this while doing some custom benchmarking using ateclient, but this is already a bug for kubectl-ate, demos, ...

- [x] Tests pass

AI-assisted.

### DIFF
--- a/internal/ateclient/builder.go
+++ b/internal/ateclient/builder.go
@@ -63,6 +63,13 @@ type Client struct {
 }
 
 // Close closes the underlying gRPC connection and stops the port-forwarder.
+
+// roundRobinServiceConfig spreads RPCs over every address the resolver returns.
+// ateapi is a headless Service, so that is one address per replica, and gRPC's
+// default of pick_first would send an entire client's traffic to whichever one
+// it connected to first. internal/ateapiauth dials with the same policy.
+const roundRobinServiceConfig = `{"loadBalancingConfig": [{"round_robin":{}}]}`
+
 func (c *Client) Close() {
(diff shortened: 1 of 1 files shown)


internal/ateclient/builder.go
```diff
@@ -63,6 +63,13 @@ type Client struct {
 }
 
 // Close closes the underlying gRPC connection and stops the port-forwarder.
+
+// roundRobinServiceConfig spreads RPCs over every address the resolver returns.
+// ateapi is a headless Service, so that is one address per replica, and gRPC's
+// default of pick_first would send an entire client's traffic to whichever one
+// it connected to first. internal/ateapiauth dials with the same policy.
+const roundRobinServiceConfig = `{"loadBalancingConfig": [{"round_robin":{}}]}`
```
BenTheElder, line 71: This verbatim the config we use in `internal/ateapiauth`.

We could dedupe them but that seemed unnecessary for now. If this becomes more complex or we need N+1 client packages we probably should at that point.

EItanya (approved):

PR #1327, opened 2026-08-31
Add a doc on using DV to validate protobuf

thockin:
API reviewers should memorize this :)

@eitanya @juli4n @BenTheElder @dberkov

### DIFF
--- a/docs/api-validation.md
+++ b/docs/api-validation.md
@@ -1,0 +1,443 @@
+# API Validation
+
+All fields of all APIs must be validated.  This is especially critical for the
+main "resource oriented" API because (once we hit 1.0) when bad data has been
+written to a database, we can't easily fix it.  Retroactively fixing validation
+to make it stricter will break users.
+
+## Tooling
+
+We use the Kubernetes
+[validation-gen](https://github.com/kubernetes/code-generator/tree/master/cmd/validation-gen)
+tool to generate validation code for our APIs.  It is driven off of "tags"
+written as comments in Go code.  Protobuf comments are copied directly into the
+generated Go code, so we can use them to drive validation-gen.
+
+We call this "declarative validation" or "DV" for short.  Almost all validation
+should be done with DV.  Any case which does not should be considered rare and
+exceptional.
+
+## DV Tags
+
+All validation tags follow the same basic formats:
+
+```
+// +k8s:<tag-name>             # no payload
+// +k8s:<tag-name>=<tag-value> # with payload
+```
+
+Some tags have payloads and some do not. Sometimes the payload is
+optional.  Run `./hack/run-tool.sh validation-gen --docs` to see some
+documentation for the tags that validation-gen supports.
+
+All tags support optional EOL comments, using a `#` character.
+
+Some tags take arguments.  Single-arguments can be either named or unnamed, and
+multi-arguments are always named.  The sytax looks like:
+
+```
+// +k8s:<tag-name>(<value>)                        # unnamed single argument
+// +k8s:<tag-name>(<arg>: <val>)                   # single named argument
+// +k8s:<tag-name>(<arg1>: <val1>, <arg2>: <val2>) # multiple arguments
+```
+
+Arguments can be strings, numbers, or booleans.  Strings may be double-quoted,
+backtick-quoted, or unquoted, but if they are unquoted, they must match the
+regex `[A-Za-z_][A-Za-z0-9_-.]*`.
+
+### Nested tags
+
+Some tags take other tags as payloads.  The syntax looks like:
+
+```
+// +k8s:<tag-name>=+k8s:<other-tag-name>
+```
+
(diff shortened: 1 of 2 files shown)


docs/api-validation.md
```diff
@@ -0,0 +1,348 @@
...
+acknowledge that we are using an alpha or beta tag by adding a `+k8s:alpha` or
+`+k8s:beta` tag prefix.  Example:
+```
+// +k8s:alpha(since: "0.0")=+k8s:theAlphaTagWeWantToUse
+```
+
+### Rules for all fields
+
+Every field must be either `+k8s:required` or `+k8s:optional`.  Any field whose
+value si the zero value for that type is considered "unspecified".  Required
```
juli4n: typo: is

docs/api-validation.md
```diff
@@ -0,0 +1,348 @@
...
+
+If validation-gen does not support the format you need, you can use a
+combination of `+k8s:maxLength` and `+k8s:customValidation` (see below) tags.
+If a string field does not have an a priori known format, then it must at least
+have a `+k8s:maxLength` tag.
+
+### Rules for struct (message) fields
+
+Struct fields are usually just tagged as required or optional (relying on the
+fields within the type to be tagged). Sometimes the specific useage site wants
```
juli4n: typo: usage

docs/api-validation.md
```diff
@@ -0,0 +1,348 @@
...
+
+All of the rules established in this document for scalar fields also apply to
+the items in a repeated field.  For example, all lists of strings must have a
+format or a maxLength tag.
+
+### Rules for map fields
+
+Map fields must also be tagged as optional or required.  If a map field is
+required, then it must have at least one key-value pair.  If a map field is
+optionl then it may have zero key-value pairs (we do not distinguish between a
```
juli4n: typo: optional

docs/api-validation.md
```diff
@@ -0,0 +1,348 @@
...
+
+### Rules for repeated fields (lists)
+
+Repeated fields must also be tagged as optional or required.  If a repeated
+field is required, then it must have at least one element.  If a repeated field
+is optional, then it may have zero elements (we do not distinguish between a
+`nil` slice and a slice with 0 elements).
+
+All repeated fields must has a `+k8s:maxItems` tag, and some may need a
+`+k8s:minItems` tag.
```
juli4n, line 158: Nit: maybe worth mentioning that k8s:required will implicitly require min items >= 1?

juli4n: nvm, it's clear from the previous paragraph.

docs/api-validation.md
```diff
@@ -0,0 +1,348 @@
...
+
+Some repeated primitive fields (e.g. `string`) must hold unique values.  This
+can be designated with the `+k8s:listType=set` tag.  Such a list field will
```
juli4n: Should we mention that `listType=atomic` should not be used at all?

thockin: It is the default, and there's some value to having EVERY list declare `listType`.  I left it out for now, but I suspect almost all of our lists are going to be sets or maps.  Should I include it?

docs/api-validation.md
thockin: in next push

docs/api-validation.md
thockin: in next push

docs/api-validation.md
thockin: in next push

thockin (commented):
I pushed with a bit more detail about unions + primitive fields.

Optional/required for primitives is awkward.

docs/api-validation.md
juli4n: I kind of like the requirement of every repeated field to set a value, because it forces the author to make a conscious decision about the list semantics (i.e. list, map, set). We can also be more strict when validating the API proto in the `apitool`.

thockin: Will add

thockin: Pushed, PTAL

docs/api-validation.md
```diff
@@ -0,0 +1,444 @@
...
+}
+```
+
+The uniqueness of values may be used in the future to allow patch-like
+operations.
+
+##### Atomic lists
+
+Lists which are not set-like or map-like are considered "atomic" and are
+designated with the `+k8s:listType=atomic` tag.  Atomic lists are treated as a
```
juli4n: Nit: I find this a bit confusing as today map and set are also updated as a whole. IIUC the annotation is just adding a uniqueness contraint within the set of values. I know the naming is confusing, but how about just saying that this should be used if the list has no map/set semantics (i.e. it's just a list)?

juli4n (dismissed):

docs/api-validation.md
thockin: simplified and pushed

juli4n (approved):
Thanks for adding this, very useful.

Issue #1291, opened 2026-08-28
Timeouts are super confusing and hard to comprehend


LiorLieberman:
Really hard to comprehend envoy and related components timeout. 

- How much time we need for a graceful drain?
- how it effects terminationgraceperiod?
- Do we need/want different timout for atenet-egress vs atenet-ingress? (likely....)
- Do all these drains affect suspend? 

Many of these Qs have straight forward answers, but we have let agents do som many timeout configs - we need to consolidate and document.

(We do need to increase timouts for ingress, likely and soon)

Issue #1015, opened 2026-08-17
move egress ext proc to the inner tunnel


LiorLieberman:
#715 introduced connect for ingress and had good discussion over the PR for why we should do the ext proc within the inner tunnel. 

TLDR -- we want to trigger ext proc for every request in the inner tunnel. 

However we also need to figure out the story for TCP -- Network ext proc vs something. 
Compared to ingress where we dont have clear TCP usecases still, for egress its pretty clear that an actor would do non http protocols (e.g git close ssh://)

@keithmattix @bowei @EItanya @haiyanmeng

haiyanmeng:
@LiorLieberman , if we  move the ext_proc to the inner tunnel, where will the logic for "this is a valid ateom/atunnel certificate for actor X" live?

cc @ahmedtd @bowei

haiyanmeng:
Let us keep the ext_proc in the outer listener for now and think through the pros and cons first.

PR #1328, opened 2026-08-31
atenet-egress: disable the CONNECT route timeout on the default gateway

botengyao:
Envoy applies a route timeout to a CONNECT tunnel's whole lifetime, so the 15s default capped every actor's outbound connection. The sdsmint variant already disabled it; add the same line here, and a test over both manifests so they cannot drift apart again.

- [ ] Tests pass

### DIFF
--- a/cmd/atenet/internal/router/egressmanifest_test.go
+++ b/cmd/atenet/internal/router/egressmanifest_test.go
@@ -1,0 +1,149 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


botengyao:
cc @bowei @LiorLieberman @haiyanmeng

LiorLieberman (approved):
Thanks @botengyao !
PR LGTM, 

we also really need to move the egress Connect to the inner request and remove it from the outer.

And we also need to figure out all the different timeouts we have in the codebase for atenet(s) - we have so many timeout vars

refs:

- https://github.com/agent-substrate/substrate/issues/1015
- https://github.com/agent-substrate/substrate/issues/1291

botengyao:
> we also really need to move the egress Connect to the inner request and remove it from the outer.

could you elaborate more for this? we can sync up later.

LiorLieberman:
Sorry, was late night. I meant egress ext proc from the Connect to the inner requests. The issue has more details

haiyanmeng (approved):

Issue #1208, opened 2026-08-25
Unify `atenet -> atunnel` traffic path


EItanya:
Right now the ingress traffic paths in Substrate are a bit all over the place. Currently `atenet` can call into the `atunnel` in one of 2 ways:
1. :443 HTTP reverse proxy: parses and authorizes every HTTP request then proxies it to the actor
2. :444 CONNECT proxy: authorizes the CONNECT request once, opens actorIP:port, then ferries raw bytes bidirectionally.

It's not clear to me that we need the first path at all. 

Parsing L7 on every request leads to worse performance overall, it is equivalent to an L7 sidecar vs the ambient model. 
1. It leads to changes such as the [following](https://github.com/agent-substrate/substrate/pull/1183) which make atunnel ever more complicated.
2. There is no current or planned L7 per request behavior we want to put in atunnel, and ideally we'd keep it that way.

Edit: One additional datapoint is that there is a design initiative right now to pack multiple actors into a single worker, see [comment](https://github.com/agent-substrate/substrate/issues/853#issuecomment-5358052607). The `tunnel` performance will become even more critical in that world.

howardjohn:
+1 to dropping L7 parsing in atunnel, I think it negates the original goals of the component and has a lot of downsides with little-to-no upside

keithmattix:
+1 - I said similar in https://github.com/agent-substrate/substrate/pull/1183#issuecomment-5414418047. There's no need for atunnel to parse L7; in addition to the points made above, it creates technical debt that we'll have to unwind in the non-HTTP path in the future

haiyanmeng:
cc @bowei  @LiorLieberman

PR #1333, opened 2026-08-31
router: Move from Host header to explicit actor target header

keithmattix:
_This PR is purposefully controversial to try and encourage movement on the DNS issue (#445)._

There has been an emerging consensus that substrate is not a system that we expect end-users to interact with directly; we can assume a higher-level orchestrator is acting on the user's behalf to program substrate. Therefore, our networking scheme does not need to use the `Host` header to encode an actor reference. This PR encodes that reference in one explicit routing header: `ate-target-actor: <atespace>/<actor>`. We expect the higher-order system originating substrate requests to set this header, removing the need for us to rewrite DNS. This also applies to actor-to-actor requests: `atunnel` originates CONNECT to the egress PEP, so when addressing another actor, it can set the routing header on the CONNECT request and the PEP can forward it to the router.

NOTE: CONNECT vs. direct HTTP (#1208) is a similar conversation; I see no reason not to lean 100% into CONNECT as the standard datapath for tunneling if we expect substrate to be controlled by a higher-order system. I only want to bite off one controversial issue at a time, so I'll save that for later.

### Breaking change

Clients that previously selected an Actor with `Host: <actor>.<atespace>.actors.resources.substrate.ate.dev` must now send `ate-target-actor: <atespace>/<actor>`. The custom Actor DNS suffix is removed. `Host` and HTTP/2 `:authority` remain application metadata; CONNECT authority still carries the target host and port.

- [x] Tests pass


benchmarking/locust/tests/counter_demo.py
```diff
@@ -48,9 +48,8 @@
 tracer = get_tracer(__name__)
 
 
-# Atenet router fronts all actor traffic. Actors are addressed by setting
-# the HTTP Host header to <actor-name>.<atespace>.actors.resources.substrate.ate.dev;
-# the router resolves that to the actor's current worker pod.
+# Atenet router fronts all actor traffic. The actor routing headers select
+# the actor whose current worker pod the router resolves.
 ROUTER_URL = "http://atenet-router.ate-system.svc.cluster.local"
 ACTOR_DOMAIN = "actors.resources.substrate.ate.dev"
```
EItanya: What is this domain used for?

internal/resources/actor.go
```diff
@@ -18,8 +18,7 @@ const (
 	// ResourceNameRegexPattern is the regular expression pattern for a valid
 	// Substrate resource name.
 	ResourceNameRegexPattern = `[a-z0-9]([-a-z0-9]*[a-z0-9])?`
-	// ActorDNSSuffix is suffix to the DNS name for direct access to Actor
-	// "<actor_name>.<atespace>.actors.resources.substrate.ate.dev"
+	// ActorDNSSuffix is the suffix used by the conventional actor authority.
 	ActorDNSSuffix = "actors.resources.substrate.ate.dev"
```
EItanya: Can we remove this as well?

internal/e2e/suites/demo/demo_test.go
```diff
@@ -1297,6 +1298,8 @@ func callActorPathOnce(t *testing.T, actorRef resources.ActorRef, method, path s
 	if err != nil {
 		return "", fmt.Errorf("failed to create request: %w", err)
 	}
+	reqHttp.Header.Set(atunnel.ActorNameHeader, actorRef.Name)
```
EItanya: Should these constants be in `atenet` or something similar since they're really an ingress concern?

internal/atunnel/ingress_test.go
```diff
@@ -466,23 +489,29 @@ func TestServeNegotiatesH2(t *testing.T) {
...
 		cancel()
-		if err := <-served; err != nil {
-			t.Errorf("Serve: %v", err)
+		select {
+		case err := <-served:
+			if err != nil {
+				t.Errorf("Serve: %v", err)
+			}
+		case <-time.After(time.Second):
+			t.Error("Serve did not stop after cancellation")
```
EItanya, line 492: nit: is there a test cleanup context or something we can piggy back off of?

docs/threat-model.md
```diff
@@ -11,49 +11,49 @@ Substrate is an early, fast moving product. It is full of debate and subject to
...
-* **atenet-router:** Substrate runs Envoy with an `ext_proc` external processor to handle Actor ingress. The ext_proc extracts the Actor name and Atespace from the HTTP `Host` header, calls the Substrate API to resume the Actor and obtain its current worker assignment, and selects that worker as a dynamic backend. The router then connects with mTLS to `atunnel` on worker port 443; `atunnel` validates the router identity and forwards traffic only to the Actor currently assigned to that worker.
-* **Object Storage:** Used to store actor snapshots.  
-* **Filesystem support:** Container local filesystem is saved in snapshots, future integrations likely to include networked storage.  
+* **Actor Routing:** A higher-order system connects to `atenet-router` and identifies the target Actor with `X-Ate-Actor-Name` and `X-Ate-Atespace`. Host and authority values are application metadata and do not select the Actor.
```
EItanya: No need to mention a higher order system here, I think just mention how it works

docs/threat-model.md
```diff
@@ -11,49 +11,49 @@ Substrate is an early, fast moving product. It is full of debate and subject to
 
 # Intended Outcome
 
-* The suggestions from this threat model should be added to Substrate's official roadmap after review and agreement with the community.   
+* The suggestions from this threat model should be added to Substrate's official roadmap after review and agreement with the community.
```
EItanya: A lot of these are just md formatting fixes which make the diff unnecessarily difficult to check

demos/egress/README.md
```diff
@@ -113,6 +113,8 @@ kubectl ate resume actor egress-demo -a demo   # wait for ACTOR_STATE_RUNNING
 kubectl -n ate-system port-forward service/atenet-router 8000:80 &
 curl -s -X POST http://localhost:8000/ \
   -H 'Host: egress-demo.demo.actors.resources.substrate.ate.dev' \
```
EItanya: should we remove this host?

EItanya (commented):
Looking good to me overall. Just a few nits. Also I think we need to figure out how to message this to the community as it will definitely break existing demos and things.

benchmarking/locust/tests/counter_demo.py
keithmattix: We still use this for the host header; just need to know what to move to

EItanya: But why do we still need the host header, wouldn't that just be up to the implementation if they still need one?

bowei:
I think this needs to be discussed because the hostname-based integration is the simplest way for a system to integrate into being able to "send to a specific actor". There are a couple of topics that are somewhat independent:

* Whether or not we actual serve DNS
* How Actors discover each other
* How external entities will address Actors 

I would like to understand how this alternative ends up for our users.

Reference material:

* [Actor Addressing and Routing [Public]](https://docs.google.com/document/d/1PKdrUq9VH8VX5g1FGp3wsOH1o293rMO-BmnZH08MhBs/edit?tab=t.0)

bowei:
Add DO NOT MERGE for the discussion

jeffmccune:
@keithmattix Thanks for organizing the call on Friday.  Here's my take from the perspective of an operator of one of the higher order orchestrators the PR description mentions.  Hopefully this helps close out the questions.

> [!NOTE]
> Recommendation: one header, `ate-target-actor: my-atespace/my-actor`, sent to a host authority the orchestrator manages for the substrate.

Why one header?  One header is preferred so that it's easier for me to check and set as simply as possible.

Why ate-target-actor?  A header called just ate-actor is ambiguous about which end it names.  x- headers are deprecated.

Why not the alternatives, from where I sit:
1. Too onerous for operators.  DNS/Host asks operators to change DNS they often don't own. I don't control our vanity domain; with this PR I need one record and one single-name cert for the router, plus an HTTPRoute.
2. Too onerous on the agent-substate team.  For example, requires substrate to own at least some portions of DNS, inviting conflict with other teams.
3. Too onerous on everyone.  Implicit, unclear, or shared ownership of the DNS boundary means everyone renegotiates it forever.

I'd be happy to add a little backwards compatibility shim to set the ate-target-actor header to avoid disrupting existing demos.  The shim would accept requests like https://my-actor.my-atespace.actors.resources.substrate.ate.dev then forward the request to ate-router after dynamically setting the ate-target-actor header derived from the host header.  This could be implemented with a simple ext_authz provider, (or many other alternatives TBD) which is how I prefer to handle authentication and authorization at the ingress Gateway anyway.

keithmattix:
@jeffmccune Appreciate the feedback! Agreed; one (non-X-prefixed) header is better in several ways

bowei:
This is a follow-up to the discussion on 026-09-04:

* I think this choice is not a 1-way door and it does not block future addition of DNS.
* There are valid concerns about taking over a FQDN domain and TLS cert issuance due to two levels of wildcarding.

It would be good to be able to give users some "easy path" towards reaching their actors/agents. Mandating headers to be added is not possible with many clients without modification. We should keep a close watch on how many users end up building a shim anyway to use the system and incorporate the shim into Substrate if a common pattern emerges.

keithmattix:
> We should keep a close watch on how many users end up building a shim anyway to use the system and incorporate the shim into Substrate if a common pattern emerges.

Agree with this. I also think uid-based DNS is probably ideal as well, but there are concerns at the storage layer that shouldn't block this move

demos/egress/README.md
```diff
@@ -140,11 +140,15 @@ kubectl ate resume actor egress-demo -a ate-demo-egress   # wait for ACTOR_STATE
 # 3. Drive the Actor's egress through the ingress gateway.
 kubectl -n ate-system port-forward service/atenet-router 8000:80 &
 curl -s -X POST http://localhost:8000/ \
-  -H 'Host: egress-demo.ate-demo-egress.actors.resources.substrate.ate.dev' \
+  -H 'Ate-Target-Actor: ate-demo-egress/egress-demo' \
```
keithmattix: We probably want to still set a (dummy) host header just for completeness/spec correctness sake

EItanya:
Removed DO NOT MERGE from the title as we met and came to a joint conclusion, I will still not merge and wait for more input, but reviews should continue from here

benchmarking/nighthawk-ingress/tests/test_spec.py
```diff
@@ -67,14 +66,22 @@ def test_traffic_template_shape():
...
     options = plugin["typed_config"]["options_list"]["options"]
-    got = [o["request_headers"][0]["header"]["value"] for o in options]
-    assert got == hosts
+    got = [
+        {header["header"]["key"]: header["header"]["value"] for header in o["request_headers"]}
+        for o in options
+    ]
+    assert got == [
+        {
+            "ate-target-actor": f"benchmark/{actor_name}",
```
bowei, line 81: nit: We should be consistent as much as possible on the Ate-Target-Actor vs ate-target-actor in the codebase. While the header is case-insensitive, it makes it harder to grep for the string.

I see instances of both.

benchmarking/nighthawk-ingress/README.md
```diff
@@ -30,14 +30,14 @@ just don't steer the search.*
...
-warmed sandboxed actors, Host-header routing, the ext_proc routing
+warmed sandboxed actors, identity-header routing, the ext_proc routing
 decision, and the mTLS hop to the worker:
 
 ```mermaid
 flowchart LR
     subgraph job["nighthawk runner Job"]
         alc["nighthawk_adaptive_load_client<br/>exponential ramp + binary search"]
-        svc["nighthawk_service<br/>16 event loops, Host header<br/>rotated across all actors"]
+      svc["nighthawk_service<br/>16 event loops, actor headers<br/>rotated across all actors"]
```
bowei: nit: indent

cmd/atenet/internal/router/ingress/ingress.go
```diff
@@ -126,7 +121,7 @@ func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestM
 		return extproc.Result{Resume: string(resumeOutcome)}, mapResumeError(actorRef, err)
 	}
 
-	// Actor template identity, used as low-cardinality route-latency metric
+	// Actor template coordinates, used as low-cardinality route-latency metric
```
bowei: nit: weird comment change

demos/jupyter/README.md
```diff
@@ -53,7 +53,24 @@ kubectl ate create actor jupyter-notebook -a ate-demo-jupyter --template jupyter
 
 ### 2. Access Jupyter via the Proxy!
 
-Substrate routes HTTP traffic using the `Host` header. To make this easy without modifying local `/etc/hosts` files, this demo includes a lightweight NGINX reverse proxy (`jupyter-proxy`) that automatically injects the proper `Host` header (`jupyter-notebook.ate-demo-jupyter.actors.resources.substrate.ate.dev`) and forwards traffic internally to the Substrate router.
+Substrate routes HTTP traffic using `Ate-Target-Actor: <atespace>/<actor>`.
+A browser cannot add that header to an address-bar navigation or WebSocket
+upgrade, so this demo puts a lightweight NGINX reverse proxy (`jupyter-proxy`)
```
bowei, line 58: We need to be careful about forcing all of our clients to do this. 

Note: no action needed.

docs/api-guide.md
```diff
@@ -141,10 +141,39 @@ Unlike a Pod, an actor is sized by its **`limits`** (CPU and Memory): the size i
 
 Container environment variables support literal `value` entries only. Values are not interpolated (`$(VAR)` references are not expanded), and Kubernetes `envFrom`/`valueFrom` sources are not supported.
 
-### Workload Connectivity (Uniform DNS)
-Substrate uses a **Uniform DNS Mesh**: every actor created from a template is automatically reachable through the **Substrate Router** via its atespace and name:
+### Workload Connectivity
+
+A higher-order system reaches an actor through the **Substrate Router** by
```
bowei, line 146: We don't need to refer to the deleted DNS stuff in the description. It's only relevant for historical purposes, will be confusing for users.

docs/threat-model.md
```diff
@@ -11,49 +11,49 @@ Substrate is an early, fast moving product. It is full of debate and subject to
 
 # Intended Outcome
```
bowei, line 12: Did this doc actually change a lot?

internal/atunnel/ingress.go
```diff
@@ -105,7 +100,7 @@ func NewServer(cfg Config) (*Server, error) {
 		return nil, fmt.Errorf("atunnel: trust bundle path is required")
 	}
 	if cfg.AllowedClientID == "" {
-		return nil, fmt.Errorf("atunnel: allowed client identity is required")
+		return nil, fmt.Errorf("atunnel: allowed client ID is required")
```
bowei: spurious change

internal/atunnel/ingress.go
```diff
@@ -414,7 +409,7 @@ func (w flushingWriter) Write(p []byte) (int, error) {
 // active actor per worker.
 func (s *Server) Activate(atespace, actorName string) error {
 	if !resources.IsValidResourceName(atespace) || !resources.IsValidResourceName(actorName) {
-		return fmt.Errorf("atunnel: invalid actor identity %q/%q", atespace, actorName)
+		return fmt.Errorf("atunnel: invalid actor reference %q/%q", atespace, actorName)
```
bowei, line 412: spurious change?

internal/atunnel/ingress.go
```diff
@@ -474,31 +469,12 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	actorHost := r.Header.Get(OriginalHostHeader)
-	if actorHost == "" {
-		actorHost = r.Host
-	}
-	r.Header.Del(OriginalHostHeader)
-	r.Host = actorHost
-
-	// ReverseProxy changes the URL destination but intentionally retains Host,
-	// allowing the actor application to observe its stable actor DNS name.
+	// ReverseProxy changes the URL destination but intentionally retains Host.
```
bowei: This comment seems out of context (recommend deleting the comment altogether)

bowei (commented):
Mostly looks ok. Some comments.

internal/atunnel/ingress.go
keithmattix: Some of these were on purpose because I don't love the conflation of "actor identity" vs. "actor reference". Will take another pass

benchmarking/nighthawk-ingress/tests/test_spec.py
keithmattix: Aligned on lower case

benchmarking/nighthawk-ingress/README.md
keithmattix: Fixed

cmd/atenet/internal/router/ingress/ingress.go
keithmattix: Moved to reference

docs/threat-model.md
keithmattix: Fixed; there are actually changes here, and I removed some of the whitespace to make that more obvious

internal/atunnel/ingress.go
keithmattix: Reverted

internal/atunnel/ingress.go
keithmattix: Removed

bowei (approved):

bowei:
Thanks for the changes.

PR #1335, opened 2026-08-31
egress: add an example credential provider which reads from k8s secret

yufan-su:
Add an example gRPC credential-provider plugin which reads from k8s secrets for the egress
credential-injection path. 

 - It resolves `ate-secret://k8s.io/default/<namespace>/<secret>/<key>`
URIs to Kubernetes Secret values, so Substrate never stores secrets — it only
brokers a read that the provider is authorized to perform. 
 - It is the only
component in the injection path with Kubernetes secerts access; the egress gateway and
its injector never read Secrets directly.

### What's included

- **`cmd/credential-provider/kubernetes-secrets`** — the gRPC service:
  - Parses and validates `ate-secret://` URIs, resolving the requested Secret
    (with single-key fallback when the URI omits a key).
  - Enforces an **atespace→namespace authorization policy** (default-deny),
    derived from the caller's attested actor SPIFFE ID.
  - Serves over **mutual TLS**, requiring the caller's client cert to chain to
    the trust bundle *and* carry the egress injector's SPIFFE SAN.
- **Manifests** (`manifests/egress-credential-injection/`) — Deployment,
  Service, ServiceAccount + RBAC, a sample namespace-policy ConfigMap, and a
  sample Secret.
- Renames the default provider address/service from `credprovider` to
  `k8s-credential-provider` across `ate-setup`, install scripts, and the egress
  injection overlay.

### atespace → namespace authorization

Beyond the mTLS check that only the egress injector may call the provider, each
request is authorized against a **default-deny atespace→namespace policy**. The
provider derives the requesting actor's atespace from its attested SPIFFE ID and
resolves a Secret only if that atespace is explicitly granted access to the URI's
namespace.


EItanya:
I think this PR is honestly a bit noisy. There's both the credential provider, and the implementation of yet another `extproc` server which uses it. Can we instead review the credential provider in isolation, especially given that there are still open questions in the doc, and then we can tackle the thing which calls it?

Doc since it's not linked anywhere: https://docs.google.com/document/d/1-zQKvu-oYJlva_kwL3xbAOaFoJJaqO0qEqmqRlt1r0c/edit?resourcekey=0-TDrWohfRuYolcXYO-QPVTQ&tab=t.0#heading=h.rx0okwgtp5yu

manifests/egress-credential-injection/README.md
```diff
@@ -0,0 +1,136 @@
...
+                 | read K8s Secret
+                 v
+              header mutation: Authorization: Bearer <secret>
+                      |
+                      v
+              re-originated TLS to the real origin, credential attached
+```
+
+Actor identity comes from the CA-signed client cert the gateway verified on the
+CONNECT leg, relayed to the injector as the `ate.actor.identity` filter-state
```
LiorLieberman: nit: dev.ate.actor.identity.

LiorLieberman (commented):
We discussed offline but for posterity, @EItanya 's idea to split the pr sounds good.

manifests/egress-credential-injection/README.md
yufan-su: Done, updated in the injector PR.

yufan-su:
Split into two PRs #1335 and #1360 and this one only contains the credential provider change.

cmd/credprovider/internal/kubeprovider/kubeprovider.go
```diff
@@ -0,0 +1,198 @@
...
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/agent-substrate/substrate/internal/actorspiffe"
+	"github.com/agent-substrate/substrate/pkg/proto/credproviderpb"
+)
+
+// ProviderClass is the substrate-secret:// URI host this backend serves.
+const ProviderClass = "kubernetes.io"
```
LiorLieberman: nit: we removed "providerClass" concept from the doc and its now only provider name. the name is kubernetes.io

internal/proto/nsauthzpb/nsauthz.proto
```diff
@@ -0,0 +1,36 @@
...
+
+package nsauthz;
+
+option go_package = "github.com/agent-substrate/substrate/internal/proto/nsauthzpb";
+
+// NamespaceAuthorizationFile is the top-level container the credential provider
+// loads from a textproto file at startup. It maps each atespace to the
+// Kubernetes namespaces whose Secrets that atespace may resolve. Enforcement is
+// default-deny: an atespace absent from this file can resolve nothing. This is a
+// POC stand-in for a future authorization API.
```
LiorLieberman: comment is very verbose, can we remove things that are POC related and make it shorter?

manifests/egress-credential-injection/namespace-policy.yaml
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The atespace→namespace authorization policy credprovider enforces: an actor's
+# atespace may only resolve Secrets whose URI namespace is listed for it here.
+# Enforcement is default-deny — an atespace absent from this file resolves
+# nothing. credprovider loads it once at startup, so editing this ConfigMap
+# requires restarting the credprovider Deployment to take effect.
```
LiorLieberman: should we add a TODO to make it dynamically reload?

LiorLieberman (commented):
Thanks @yufan-su !

manifests/egress-credential-injection/credprovider.yaml
```diff
@@ -0,0 +1,157 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: credprovider-secret-reader
+subjects:
+- kind: ServiceAccount
+  name: credprovider
+  namespace: ate-system
+---
+apiVersion: apps/v1
+kind: Deployment
```
EItanya, line 51: Creating YET another deployment for this feels like it spreads our deployment footprint out quite a bit. What do you think about running this as a sidecar for the egress policy, or from the control-plane somehow? I'm worried about:
1. Latency
2. deployment footprint

internal/proto/nsauthzpb/nsauthz.proto
EItanya: I agree with @LiorLieberman here, it's not clear we really need these `proto` types at all?

pkg/proto/credproviderpb/credprovider.proto
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// SecretRequestContext is the attested context Substrate passes with a request.
+// The provider is trusted to have authenticated the caller (mutual TLS); the
+// caller is trusted to have attested the actor identity.
+message SecretRequestContext {
+  // The attested actor identity on whose behalf the secret is fetched. Today
+  // this is the actor's SPIFFE URI as verified by the egress gateway; a
+  // verifiable Actor JWT is the intended future form.
+  string actor_identity = 1;
```
EItanya, line 38: Is this secure? Wouldn't we need the actual actor identity document?

manifests/egress-credential-injection/namespace-policy.yaml
```diff
@@ -0,0 +1,32 @@
...
+  name: credprovider-namespace-policy
+  namespace: ate-system
+data:
+  namespace-policy.textproto: |
+    # atespace "team-a" may resolve secrets in namespace "ns1" (matches the
+    # sample policy's substrate-secret://kubernetes.io/team-secrets/ns1/example-api).
+    policy {
+      atespace: "team-a"
+      allowed_namespaces: "ns1"
+    }
```
EItanya: Has this API been discussed at all? This seems like an important piece we need to come to agreement on which was not in the doc @bowei @LiorLieberman

internal/proto/nsauthzpb/nsauthz.proto
EItanya: Doubly so given that this never goes over the wire, it could just be `golang` directly

internal/actorspiffe/actorspiffe.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"net/url"
+	"path"
+	"strings"
+)
+
+// TrustDomain is the SPIFFE host an actor identity URI carries, matching the URI
+// minted in cmd/ateapi/internal/actoridentity.
+const TrustDomain = "substrate-actor.local"
```
EItanya, line 30: This relates to my comment about the actual actor credential, I think this is a code smell from that

cmd/credprovider/main.go
```diff
@@ -0,0 +1,189 @@
...
+)
+
+const serviceName = "credprovider"
+
+var (
+	listenAddr   = pflag.String("listen-address", ":50051", "gRPC listen address")
+	metricsAddr  = pflag.String("metrics-address", ":9090", "Prometheus/health HTTP listen address")
+	serverBundle = pflag.String("server-cred-bundle", "", "credential bundle (PEM key+chain) presented for serving TLS; empty serves plaintext (dev only)")
+	clientCAFile = pflag.String("client-ca-file", "", "CA bundle that caller (injector) client certificates must chain to; empty accepts any client when TLS is on")
+	defaultKey   = pflag.String("default-secret-key", "", "Secret data key used when a credential URI omits one; empty requires a single-key Secret")
```
EItanya: I don't think this is a good idea

EItanya (changes requested):

manifests/egress-credential-injection/credprovider.yaml
bowei: I think we need to see if we consider this more of an example vs something that a lot of users will use.

Key things that would make some of the decisions:

1) Would this be enabled by default (with opt-out)
2) How we document the positioning of this (it's for demos vs it's for prod)

manifests/egress-credential-injection/namespace-policy.yaml
bowei: It's somewhat separate than the doc as this is specific to cred provider implementation.

I was thinking for this example, we do something very simple like a static YAML map of atespace(s) to k8s namespaces.

Nit: probably shouldn't use textproto, this is not a very popular format (outside of Google)

cmd/credprovider/internal/kubeprovider/kubeprovider.go
yufan-su: done

internal/proto/nsauthzpb/nsauthz.proto
yufan-su: done, removed the proto and updated to read from a yaml

manifests/egress-credential-injection/namespace-policy.yaml
yufan-su: done

cmd/credprovider/internal/kubeprovider/kubeprovider.go
```diff
@@ -0,0 +1,190 @@
...
+//	substrate-secret://kubernetes.io/<provider name>/<namespace>/<secret>[/<key>]
+type SecretRef struct {
+	// ProviderName is the registered provider instance name. It is opaque to
+	// this backend (a single kubernetes.io provider serves every name); it is
+	// retained only so it can be logged and, later, used to scope providers.
+	ProviderName string
+	Namespace    string
+	Name         string
+	// Key is the data key within the Secret, or "" when the URI omits it (the
+	// server then falls back to its configured default key).
```
keithmattix: Default keys feel weird to me

manifests/egress-credential-injection/credprovider.yaml
EItanya: I think that's a fair point, what do we consider it? How would a user actually deploy secrets and providers to their env?

manifests/egress-credential-injection/namespace-policy.yaml
EItanya: Ok, so this who PR is really meant to be an example of what one of these plugins could be? I think overall it's not clear to me how various plugins will work, what their lifecycle will be, how they will be accessed by the PEP, etc

pkg/proto/credproviderpb/credprovider.proto
yufan-su: resolving as proto was added in #1360

internal/actorspiffe/actorspiffe.go
yufan-su: resolving as proto was added in https://github.com/agent-substrate/substrate/pull/1360

cmd/credprovider/main.go
yufan-su: fixed

manifests/egress-credential-injection/k8s-credential-provider.yaml
```diff
@@ -0,0 +1,157 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: credprovider-secret-reader
+subjects:
+- kind: ServiceAccount
+  name: credprovider
+  namespace: ate-system
+---
+apiVersion: apps/v1
+kind: Deployment
```
yufan-su: This is more of a reference implementation of the credential provider in substrate. It is disabled by default.

manifests/egress-credential-injection/namespace-policy.yaml
LiorLieberman: On this particular aspect, its not an "example" to how plugins should work. Point is that 

All provider implementations are responsible for defining an authz mapping for themselves, in a k8s example, the thing that is most straightforward is "atespace to namespace" authz (to start). GCP Secret manager will have to implement something else, and vault has their own concepts (roles, etc.). 

Obvs this can be improve later but IMO it provides sufficient "security" to start. Agree on making this a configmap maybe and not a textProto

LiorLieberman (commented):
One of the comments Tim mentioned is multi cluster. E.g I am storing a secret in a different cluster. 

Not a requirement for this PR but the (code) design is important to support that. E.g 

"I install the kubernetes.io provider (and have it in as a registration flag), and together with this I either (a) provide a config map that has my cluster-name/dns/control-plane ip to a kubeconfig entry (b) store that mapping somewhere else not in a config map?"

how is the plugin authorized to read secrets? is it a service account only? can we support what I mentioned above to support fetching secrets from a different cluster?

EItanya:
> One of the comments Tim mentioned is multi cluster. E.g I am storing a secret in a different cluster.
> 
> Not a requirement for this PR but the (code) design is important to support that. E.g
> 
> "I install the kubernetes.io provider (and have it in as a registration flag), and together with this I either (a) provide a config map that has my cluster-name/dns/control-plane ip to a kubeconfig entry (b) store that mapping somewhere else not in a config map?"
> 
> how is the plugin authorized to read secrets? is it a service account only? can we support what I mentioned above to support fetching secrets from a different cluster?

Should that be in scope for this initial round of this plugin. I still don't think we've decided what the support level is for this component, if it's official or just an example etc.

Maybe let's start with this as the default local secret impl and then we can expand or let others bring their own multicluster vairant?

bowei:
I think we should not get too into the weeds on the multicluster aspect of it right now. We could leave a space in the path for later expansion if we really want to:

We have:

```
	ate-secret://kubernetes.io/<namespace>/<secret>[/<key>]
```

Maybe

```
	ate-secret://kubernetes.io/default/<namespace>/<secret>[/<key>]
```

Later we can add

```
	ate-secret://kubernetes.io/clusters/<cluster name>/<namespace>/<secret>[/<key>]
```

yufan-su:
Updated the substrate uri format to be 

```
ate-secret://k8s.io/default/<namespace>/<secret>[/<key>] - for local secrets
```


replaced `kubernetes.io` with `k8s.io` to make the uri shorter

manifests/egress-credential-injection/namespace-policy.yaml
yufan-su: updated to read the atespace -> ns policy from a yaml file.

cmd/atenet/internal/router/egress/provider.go
```diff
@@ -94,8 +94,8 @@ func ProviderName(name string) (string, error) {
 }
 
 // providerNameFromURI returns the provider name of a ate-secret:// URI —
-// the URI host, e.g. "kubernetes.io" in
-// ate-secret://kubernetes.io/<namespace>/<secret>. The gateway uses it to
+// the URI host, e.g. "k8s.io" in
+// ate-secret://k8s.io/default/<namespace>/<secret>. The gateway uses it to
```
bowei: We should be consistent on the documentation of the URI as it can be ns/secret or ns/secret/key

cmd/credential-provider/kubernetes-secrets/kubeprovider.go
```diff
@@ -0,0 +1,204 @@
...
+//
+// Future work: the secret uri can grow a "cluster/<cluster>" locator for fetching remote Secrets.
+//
+//	ate-secret://k8s.io/cluster/<cluster>/<namespace>/<secret>[/<key>]
+type SecretRef struct {
+	Namespace string
+	Name      string
+	// Key is the data key within the Secret, or "" when the URI omits it (only
+	// allowed when the secret contains one entry).
+	Key string
```
bowei, line 62: Why not always populate Key? Otherwise there is a weird asymmetry when this is called.

cmd/credential-provider/kubernetes-secrets/kubeprovider.go
```diff
@@ -0,0 +1,204 @@
...
+
+// selectKey resolves which Secret data entry to return: the URI's explicit key,
+// else the sole key of a single-key Secret. A URI without a key resolving a
+// multi-key Secret is an error.
+func selectKey(data map[string][]byte, uriKey string) ([]byte, error) {
+	if uriKey == "" {
+		if len(data) != 1 {
+			return nil, fmt.Errorf("no key given and the secret has %d keys; specify one in the URI", len(data))
+		}
+		for _, v := range data {
```
bowei: return data[0], nil? 

(why iterate)

cmd/credential-provider/kubernetes-secrets/kubeprovider_test.go
```diff
@@ -0,0 +1,272 @@
...
+		},
+		{name: "wrong scheme", uri: "https://k8s.io/default/ns1/example-api", wantErr: true},
+		{name: "wrong provider", uri: "ate-secret://vault.io/default/ns1/example-api", wantErr: true},
+		{name: "missing locator", uri: "ate-secret://k8s.io/ns1/example-api/token", wantErr: true},
+		{name: "remote form not yet supported", uri: "ate-secret://k8s.io/cluster/remote-east/ns1/example-api", wantErr: true},
+		{name: "too few segments", uri: "ate-secret://k8s.io/default/ns1", wantErr: true},
+		{name: "too many segments", uri: "ate-secret://k8s.io/default/ns1/example-api/token/extra", wantErr: true},
+		{name: "query not allowed", uri: "ate-secret://k8s.io/default/ns1/example-api?cluster=remote", wantErr: true},
+		{name: "fragment not allowed", uri: "ate-secret://k8s.io/default/ns1/example-api#token", wantErr: true},
+		{name: "unparseable", uri: "://://", wantErr: true},
```
bowei, line 55: * add %escape (we don't support right now)
* add with space " "

cmd/credential-provider/kubernetes-secrets/main.go
```diff
@@ -0,0 +1,210 @@
...
+			close(done)
+		}()
+		select {
+		case <-done:
+		case <-time.After(*drainGrace):
+			slog.Warn("graceful shutdown timed out; forcing stop", slog.Duration("grace", *drainGrace))
+			srv.Stop()
+		}
+	}()
+
```
bowei, line 145: do we have statusz page

bowei:
Minor comments -- generally looks good.

Should clarify URI escaping is not supported in the string.

cmd/atenet/internal/router/egress/provider.go
yufan-su: updated to have the key as required in uri, `ate-secret://k8s.io/default/<namespace>/<secret>/<key>`

cmd/credential-provider/kubernetes-secrets/kubeprovider.go
yufan-su: done

cmd/credential-provider/kubernetes-secrets/kubeprovider.go
yufan-su: done, made key as required field

cmd/credential-provider/kubernetes-secrets/kubeprovider_test.go
yufan-su: Added validation in egress policy and provider to reject percent-encoded uri

cmd/credential-provider/kubernetes-secrets/main.go
yufan-su: no we didn't have that yet, I can add it in a followup change

LiorLieberman:
> Updated the substrate uri format to be 
> 
> 
> 
> ```
> 
> ate-secret://k8s.io/default/<namespace>/<secret>[/<key>] - for local secrets
> 
> ```
> 
> 
> 
> 
> 
> replaced `kubernetes.io` with `k8s.io` to make the uri shorter

Wait why not starting with clusters now?

What are we loosing?

/clusters/_self_

LiorLieberman:
> I think we should not get too into the weeds on the multicluster aspect of it right now. We could leave a space in the path for later expansion if we really want to:
> 
> 
> 
> We have:
> 
> 
> 
> ```
> 
> 	ate-secret://kubernetes.io/<namespace>/<secret>[/<key>]
> 
> ```
> 
> 
> 
> Maybe
> 
> 
> 
> ```
> 
> 	ate-secret://kubernetes.io/default/<namespace>/<secret>[/<key>]
> 
> ```
> 
> 
> 
> Later we can add
> 
> 
> 
> ```
> 
> 	ate-secret://kubernetes.io/clusters/<cluster name>/<namespace>/<secret>[/<key>]
> 
> ```
> 
> 

We should start with clusters/cluster-name now. Adding another element later just complicates that.

Why not just go with clusters/'_self_' and support only local now or something

(Or self instead or underscore self)

LiorLieberman:
Errr

\_self\_

EItanya (commented):
🤖 AI-generated review.

Three findings below. The four focused test packages passed, and local TLS and request-load probes reproduced the first two findings on `55072e64`. I checked the changes through `6fba1c3`; the client trust loading, Kubernetes client limits, and image build lists are unchanged. No live Kubernetes deployment was tested.

cmd/credential-provider/kubernetes-secrets/main.go
```diff
@@ -0,0 +1,210 @@
...
+	if !pool.AppendCertsFromPEM(ca) {
+		return nil, fmt.Errorf("no certificates in --client-ca-file %q", *clientCAFile)
+	}
+
+	cfg := &tls.Config{
+		MinVersion:     tls.VersionTLS13,
+		GetCertificate: credbundle.Loader(*serverBundle),
+		// Require a client certificate that chains to the trust bundle.
+		ClientAuth:       tls.RequireAndVerifyClientCert,
+		ClientCAs:        pool,
```
EItanya: 1. **Reload the client trust bundle when it changes.**

`ClientCAs` is populated only at startup, although the deployment mounts a rotating ClusterTrustBundle. Once the injector presents a certificate signed by a newly published CA, new provider connections fail until the provider restarts.

A real TLS handshake probe confirmed that atomically publishing an old+new CA bundle still produces `x509: certificate signed by unknown authority`; rebuilding server credentials from the updated file makes the same certificate work.

Please reload the trust pool for new handshakes when the projected bundle changes, retaining both chain verification and the injector SAN check.

cmd/credential-provider/kubernetes-secrets/main.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	return nil
+}
+
+func newKubeClient() (kubernetes.Interface, error) {
+	cfg, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, fmt.Errorf("in-cluster config: %w", err)
+	}
+	return kubernetes.NewForConfig(cfg)
```
EItanya, line 166: 2. **Set Kubernetes client limits appropriate for the request path.**

`InClusterConfig` leaves QPS and Burst unset, so this client uses client-go's defaults: 5 QPS and a burst of 10. Every injected header performs an uncached Secret GET through this shared client, so the limit applies across all actors using the provider.

Against an immediate-response local HTTP API, 40 concurrent fetches with a five-second context budget matching the egress timeout produced 34 successes and six `Unavailable` errors from the client-side rate limiter. This probe used a real client-go client; it did not run Envoy.

Please set deliberate limits and expose QPS/burst tuning, following the existing podcertcontroller options.

cmd/ate-setup/internal/images/images.go
```diff
@@ -46,6 +46,7 @@ var Components = []string{
 	"cmd/atenet",
 	"cmd/ateom-gvisor",
 	"cmd/ateom-microvm",
+	"cmd/credential-provider/kubernetes-secrets",
```
EItanya, line 49: 3. **Include the provider in the standard image build.**

This registers a new deployable image, but the Makefile's image lists omit `./cmd/credential-provider/kubernetes-secrets`. `make -n build-images` confirms it is absent, so a registry populated by the standard image build lacks the provider image.

Please add the package to the appropriate Makefile image list so the build and installer lists stay consistent.

LiorLieberman (dismissed):
LGTM. I really prefer having the same element list in the uri and start with /clusters/\_self\_ or some other preserved thing but... not going to block on that, so good with whatever you decide

cmd/credential-provider/kubernetes-secrets/main.go
yufan-su: done, added a PoolLoader in provider to reload CA

cmd/ate-setup/internal/images/images.go
yufan-su: done

cmd/credential-provider/kubernetes-secrets/main.go
yufan-su: done, added flag to config qps and burst and set default to 50 and 100.

LiorLieberman (approved):

bowei (approved):

bowei:
What are the requested changes that are still pending? @EItanya I can't find any pending items from your comments.

EItanya (approved):

PR #1338, opened 2026-08-31
ateom: create the OCI mountpoints through os.Root

BenTheElder:
The proc/sys/dev mountpoints were created with os.MkdirAll on paths inside the container rootfs, whose lower is the actor's image and whose upper is restored from the guest's own snapshot. A symlink planted at one of those names sent the mkdir wherever it pointed on the worker pod, created by ateom as root. Confine it to the rootfs with os.Root, in both the merged-overlay and the reconstructed-from-image staging paths.


EItanya (approved):
Good catch

ahmedtd (approved):

PR #1344, opened 2026-08-31
csi: Cache CA certificate pool in CSI TLS verification

### LINKED ISSUE #1182: CSI TLS verification lacks caching, causing performance degradation
## Expected Behavior
The TLS configuration verification in `resolveTLSConfig` should efficiently verify CA certificates during connection handshakes while maintaining the ability to detect and reload CA certificate rotations without impacting performance.

## Actual Behavior
The CA certificate verification logic in `resolveTLSConfig` currently reads the CA trust bundle from disk on every TLS connection handshake. This causes redundant disk I/O operations on each connection, which may impact performance and scalability.

The current implementation is acceptable for the initial implementation but has been identified as needing optimization before general availability (GA).

1. Deploy CSI driver controller with TLS enabled via Pod Identity (PR #956)
2. Configure multiple ate-api instances making requests to the CSI controller
3. Monitor disk I/O and performance metrics during sustained load
4. Observe that CA certificate pool is re-read from disk on each TLS handshake

## Specifications

- **Version**: From PR #956 (identity: Secure CSI Driver TCP Connections via Pod Identity)
- **Component**: CSI Driver TLS Configuration
- **Related Files**:
  - `internal/volume/csi/plugin.go` (function: `resolveTLSConfig`)
- **Related PR**: #956
- **Target Milestone**: M2 (pre-GA)

## Additional Context

The `resolveTLSConfig` function uses `VerifyConnection` to dynamically verify server certificates against the latest CA bundle read from disk on each handshake. While this ensures CA rotations are detected without process restart, it currently lacks a caching mechanism.

Implementation should:
- Cache the CA certificate pool to reduce disk I/O
- Maintain dynamic reload capability to detect CA certificate rotations
- Reference implementation: https://github.com/ahmedtd/tinycert/blob/main/lib/spiffefsd/spiffefsd.go

hajiler:
Fixes #1182

### Description
In `resolveTLSConfig`, TLS verification was configured to dynamically read and parse the CA trust bundle on every handshake to support zero-downtime CA rotation. However, reading and parsing PEM certificates from disk on every connection causes unnecessary disk I/O and CPU overhead.

This PR introduces an in-memory caching mechanism for the CA certificate pool:
- Adds a thread-safe `caPoolCache` that caches the parsed `*x509.CertPool` in memory.
- Uses stat-matching (`os.SameFile`, modtime, size) to determine when a reload is needed.

- [X ] Tests pass

### DIFF
--- a/internal/volume/csi/plugin.go
+++ b/internal/volume/csi/plugin.go
@@ -22,6 +22,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"github.com/agent-substrate/substrate/internal/ateompath"
 	"github.com/agent-substrate/substrate/internal/credbundle"
@@ -336,9 +337,10 @@ func resolveTLSConfig(cfg *v1alpha1.CSIDriverConfig, paths tlsPaths) (*tls.Confi
 		return nil, fmt.Errorf("only pod identity TLS is supported in this configuration")
 	}
 
-	// Verify CA pool exists and is readable at construction time.
-	_, err := getCertPool(paths.caCert)
-	if err != nil {
+	caCache := newCAPoolCache(paths.caCert)
+
+	// Verify CA pool exists, is readable, and populate the initial cache.
+	if _, err := caCache.getCertPool(); err != nil {
 		return nil, fmt.Errorf("failed to load CA cert pool from %q: %w", paths.caCert, err)
 	}
 
@@ -358,15 +360,16 @@ func resolveTLSConfig(cfg *v1alpha1.CSIDriverConfig, paths tlsPaths) (*tls.Confi
 		// Standard tls.Config.RootCAs is a static cert pool evaluated at construction time.
 		// To automatically pick up CA trust bundle rotations on disk without restarting the process,
 		// we set InsecureSkipVerify=true and verify the server certificate chain dynamically
-		// against the latest CA bundle read from disk in VerifyConnection.
+		// against the CA bundle in VerifyConnection.
+		// caCache avoids re-reading and re-parsing the CA bundle from disk on every handshake
+		// unless the file has changed.
 		InsecureSkipVerify: true,
 		VerifyConnection: func(state tls.ConnectionState) error {
 			if len(state.PeerCertificates) == 0 {
 				return fmt.Errorf("server did not present certificates")
 			}
 
-			// Read CA trust bundle on each TLS connection handshake.
-			roots, err := getCertPool(paths.caCert)
+			roots, err := caCache.getCertPool()
 			if err != nil {
 				return fmt.Errorf("failed to load CA cert pool from %q: %w", paths.caCert, err)
 			}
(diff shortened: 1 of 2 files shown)


internal/volume/csi/plugin.go
```diff
@@ -391,14 +394,58 @@ func resolveTLSConfig(cfg *v1alpha1.CSIDriverConfig, paths tlsPaths) (*tls.Confi
...
+}
+
+// getCertPool returns the parsed CA cert pool, re-reading the file only when it has changed
+// on disk (identity, modification time, or size).
+func (c *caPoolCache) getCertPool() (*x509.CertPool, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	fi, err := os.Stat(c.path)
+	if err != nil {
```
ILLUM1N0X, line 425: If an error is returned due to a missing file, should we clear `ci.fi` and `ci.pool` here so that we do not hold stale memory?

internal/volume/csi/plugin.go
```diff
@@ -391,14 +394,58 @@ func resolveTLSConfig(cfg *v1alpha1.CSIDriverConfig, paths tlsPaths) (*tls.Confi
...
+	fi   os.FileInfo
+	pool *x509.CertPool
+}
+
+func newCAPoolCache(path string) *caPoolCache {
+	return &caPoolCache{path: path}
+}
+
+func (c *caPoolCache) hasFileChanged(newFi os.FileInfo) bool {
+	if c.fi == nil || newFi == nil {
```
ILLUM1N0X, line 412: Returning `false` when either is `nil` means we silently ignore state changes if a file appears or disappears.

What do you think about updating the logic to handle `nil` permutations more accurately?
```
if c.fi == nil && newFi == nil {
	return false
}
if c.fi == nil || newFi == nil {
	return true
}
```

ILLUM1N0X (changes requested):

internal/volume/csi/plugin.go
```diff
@@ -391,14 +394,58 @@ func resolveTLSConfig(cfg *v1alpha1.CSIDriverConfig, paths tlsPaths) (*tls.Confi
...
+		return false
+	}
+	return !os.SameFile(c.fi, newFi) || !c.fi.ModTime().Equal(newFi.ModTime()) || c.fi.Size() != newFi.Size()
```
Sneha-at, line 415: nit: it might be better to use a positive check here instead of looking for differences?
` c.pool != nil && sameFile && sameModTime && sameSize` ensures any unhandled case results in a re-read of cert instead of using old certificate (it might be expired)

internal/volume/csi/plugin.go
Sneha-at, line 412: If the file is nil should we trigger to read from the path here?

internal/volume/csi/plugin.go
hajiler: I'd rather leave the cache alone here. A stat failure is already fatal for this handshake — we return the error and the connection fails, so nothing stale is served on this path.

The next handshake stats from scratch, and the cached pool is only reused when that stat succeeds *and* still matches the stat the pool was parsed from (`os.SameFile` + mtime + size). So a transient failure can't lead to us serving a bundle that has since changed on disk. Clearing would only mean that a blip — an `EIO`, or a read that lands mid-rotation — throws away a still-valid pool and forces a re-parse, which is the cost this cache exists to avoid.

There isn't much memory to reclaim either: it's one `*x509.CertPool` plus a `FileInfo`, held for the life of the `tls.Config`.

internal/volume/csi/plugin.go
hajiler: Agreed. I'll fold this in along with @Sneha-at's suggestion below and flip the predicate to the positive form, so that any state we haven't reasoned about falls through to a re-read rather than to a cache hit:

```go
func (c *caPoolCache) isFileUnchanged(newFi os.FileInfo) bool {
	if c.fi == nil || newFi == nil {
		return false
	}
	return os.SameFile(c.fi, newFi) && c.fi.ModTime().Equal(newFi.ModTime()) && c.fi.Size() == newFi.Size()
}
```

Worth noting that both nil legs are unreachable as the code stands: `getCertPool` is the only caller and only calls this after a successful `os.Stat`, so `newFi` is never nil, and `c.fi == nil` only before the first successful parse, which the `c.pool == nil` guard already covers. But that's a fragile thing for the predicate to depend on, and stating it explicitly costs nothing.

internal/volume/csi/plugin.go
hajiler: Agreed on the shape — I'll flip it to `isFileUnchanged` and gate on `c.pool != nil && isFileUnchanged(fi)`, which also resolves @ILLUM1N0X's point about the nil permutations above.

One correction on the rationale, though: an expired certificate isn't a case a re-read would rescue. The bytes on disk don't change when a cert in the bundle lapses, so re-parsing produces the same pool. Expiry is caught downstream by `leaf.Verify` in `VerifyConnection`, which rejects a chain through an expired root regardless of how fresh the pool is. The reason to prefer the positive form is the other half of what you said: any case we haven't handled resolves to a re-read instead of to a cache hit.

internal/volume/csi/plugin.go
hajiler: Yes. With the predicate flipped to `isFileUnchanged` (see the thread above), a nil on either side reports "changed", so `getCertPool` falls through to `parseCertPool` and re-reads from `c.path`.

internal/volume/csi/plugin.go
hajiler: The stat error is already fatal for that handshake, and the cached pool is only ever reused after a successful stat that still matches (os.SameFile + mtime + size), so a transient failure can't cause a stale bundle to be served. 

Clearing would just throw away a valid pool on a blip.

ILLUM1N0X (approved):

msau42 (approved):

PR #1346, opened 2026-08-31
api: delete the DebugClear RPC and the Debug service

### LINKED ISSUE #999: Delete the DebugClear API before release
The public `Debug.DebugClear` RPC deletes all persistent application state. While it is useful for development, there is no method-level authorization or server side role restricting it to a dedicated administrator. 

Although authorization could be designed for this operation, exposing a remotely callable global wipe is unnecessary for the production API. It also requires the normal `ateapi` database role to have `TRUNCATE` privileges, which conflicts with the least-privilege runtime role design in #997

jpbetz:
Fixes #999.


jpbetz:

jpbetz:
OK, this is ready for review. It's mostly a reduction.

cmd/ateapi/internal/store/store.go
```diff
@@ -238,7 +239,10 @@ type Interface interface {
 	// Returns ErrLeaseConflict if the lease is already held by another client.
 	AcquireLease(ctx context.Context, key string) (*Lease, error)
 
-	// DebugClearAll drop all data from the database. Useful for debugging / local testing/
+	// DebugClearAll drops all data from the store. It is test-only: no RPC
+	// reaches it and nothing outside tests ever calls it (controlapi's
+	// pass-through exists only to satisfy this interface). Tests use it to get
+	// an empty store between cases without paying for a fresh database.
 	DebugClearAll(ctx context.Context) error
```
juli4n: This could be a follow-up PR, but how about we remove it from the `Interface` type and we move the logic to delete all (i.e. the `TRUNCATE` statement) to atepg_test.go (in `setupPostgresPersistence` ).

That way we avoid having test-only code mixed up with production code.

juli4n (dismissed):
It looks like you need to rebase and sign the CLA.

Will merge once the presubmit checks pass.

cmd/ateapi/internal/store/store.go
jpbetz: Good call. Fixing now.

juli4n (approved):

PR #1347, opened 2026-08-31
atepg: sweep expired leases on the maintenance tick

### LINKED ISSUE #1040: Move PostgreSQL expired-lease cleanup off the lock acquisition path
## Problem

In #1023 `atepg.AcquireLock` currently runs `DELETE FROM leases WHERE expires_at <= clock_timestamp()` before every lease acquisition. The `expires_at` index keeps this bounded, but it still adds an extra database write and round trip to every lock acquisition, including when there are no expired leases.

The acquisition query already reclaims an expired lease for the requested key through `INSERT ... ON CONFLICT ... DO UPDATE ... WHERE leases.expires_at <= clock_timestamp()`, so global cleanup is not required for lock correctness. It only prevents expired rows for inactive keys from accumulating.

## Proposed follow-up

Move global expired-lease cleanup off the acquisition path, for example to a periodic background task with a bounded interval and/or batch size. In a multi-replica deployment, ensure cleanup work is not needlessly duplicated or that concurrent cleaners remain cheap and safe.

jpbetz:
Fixes #1040.

### What

`AcquireLease` ran `DELETE FROM leases WHERE expires_at <= clock_timestamp()` —
a global, table-wide write — before every single acquisition. Since the actor
workflow lease (`lease:actor:<atespace>:<name>`) is taken on the ResumeActor
path, every resume paid for a scan and delete over rows belonging to unrelated
actors, and concurrent resumes contended with each other on those rows.

This moves the delete to the periodic loop that already maintains
`worker_outbox` partitions, and does nothing else to the lease protocol.

Changes, all in `cmd/ateapi/internal/store/atepg`:

- `AcquireLease` no longer calls `cleanupExpiredLeases`. `acquireLease` is
  untouched, only lifted into a named `acquireLeaseSQL` const so a test can run
  the same statement inside its own transaction.
- New `sweepExpiredLeases`, run from the maintenance loop on the watch pool
  (never the request pool), unelected on every replica.
- `outboxMaintenance` -> `backgroundMaintenance`, since it now covers more than
  the outbox; `outboxMaintenanceInterval` -> `maintenanceInterval` to match.
  `outboxMaintenancePassTimeout` keeps its outbox-specific name because it now
  bounds only the outbox step.
- Each maintenance step runs under its own deadline via a small
  `runMaintenanceStep` helper.
- Doc-comment fixes for the `watchPool` field and the `worker_outbox` schema
  comment, which both still named the old function/scope.

### Why this is safe

The sweep was never load-bearing for correctness, which is the whole reason it
can move:

- `acquireLease` upserts with `ON CONFLICT (key) DO UPDATE ... WHERE
  leases.expires_at <= clock_timestamp()`. An expired row is therefore already
  equivalent to a missing one: it is overwritten in place, with a fresh token
  and expiry. Nothing had to delete it first.
- `renewLease` and `releaseLease` both match on the holder's token, so a row
  the sweep removes can never strand a live holder or let a stale holder steal
  a lease back.

So deleting expired rows is pure garbage collection: skipping or failing a pass
costs a little storage and nothing else. That is what lets it run unelected on
every replica rather than needing the advisory-lock election the outbox
retention transaction uses.

### The one subtlety: the sweep's DELETE keeps its predicate on `leases`

The bounded statement is:

```sql
DELETE FROM leases
WHERE expires_at <= clock_timestamp()
  AND key IN (
    SELECT key FROM leases
    WHERE expires_at <= clock_timestamp()
    LIMIT $1
    FOR UPDATE SKIP LOCKED)
```

Both expiry predicates are load-bearing, and the outer one is easy to drop by
accident. Under READ COMMITTED, when the DELETE blocks on a row lock held by a
concurrent takeover of that same expired key, Postgres re-evaluates the DELETE's
`WHERE` against the committed tuple. Bounded by `key IN (...)` alone, that
re-check still matches (the key list was already materialized), and the sweep
deletes the *live* lease the takeover just wrote — after which a third caller
can acquire a key whose new holder still believes it holds it. Keeping
`expires_at <= clock_timestamp()` on the target relation makes the re-check
skip the row. `FOR UPDATE SKIP LOCKED` then avoids the block entirely and keeps
replicas' sweeps from serializing on one another.

`TestSweepExpiredLeases_ConcurrentTakeover` pins this: it fails deterministically
against the subquery-only form and passes against the committed one.

### Bounding and drain

`leaseSweepBatch = 1000` caps how many row locks one statement takes. The pass
loops until a batch comes back short, so a burst of abandoned rows — an ateapi
replica dying mid-flight leaves one row per in-flight actor — is reclaimed by a
single pass rather than 1000 rows per minute. The loop is bounded by
`leaseSweepPassTimeout` (30s), separate from the outbox pass's 5-minute budget.

### No metrics

Nothing in `docs/metrics/registry/metrics.yaml` is touched and no instrument is
added. A sweep-rows-collected counter is plausible follow-up but is not worth a
registry entry on its own here.

## Testing done

All in the worktree, against the package's real `postgres:18` testcontainer.

- `go build ./...` — exit 0, no output.
- `go vet ./cmd/ateapi/...` — exit 0, no output.
- `gofmt -l ./cmd ./internal ./pkg` — empty.
- `go test ./cmd/ateapi/internal/store/atepg/ -count=1` —
  `ok github.com/agent-substrate/substrate/cmd/ateapi/internal/store/atepg 8.755s`
- `go test ./cmd/ateapi/internal/store/atepg/ -count=1 -run 'Lease|Sweep' -v` —
  PASS `TestAcquireLease_DoesNotSweepOtherKeys`,
  `TestAcquireLease_ExpiresAfterHolderStops`,
  `TestAcquireLease_TakesOverExpiredRowWithoutSweep`,
  `TestSweepExpiredLeases`, `TestSweepExpiredLeases_DrainsPastOneBatch`,
  `TestSweepExpiredLeases_ConcurrentTakeover`,
  `TestAcquireLease_ConcurrentTakeover`.
- `go test ./cmd/ateapi/internal/controlapi/... -run 'Resume|Lease' -count=1` —
  `ok .../controlapi 3.720s`, `ok .../controlapi/functionaltest 11.465s`.
- `hack/verify/gofmt.sh`, `boilerplate.sh`, `go-modules.sh`, `golangci-lint.sh`
  — all exit 0, no output.
- `hack/verify/metrics.sh` not run: the diff touches no instrument, no label,
  and not `docs/metrics/registry/metrics.yaml`.

Negative check on the regression test: temporarily reverting only the sweep SQL
to the `key IN (...)`-only form makes `TestSweepExpiredLeases_ConcurrentTakeover`
fail with "the sweep deleted a lease that was taken over while the sweep ran;
two callers can now hold it". Reverted immediately after.

Test coverage added:

- `TestAcquireLease_CleansExpiredLeases` -> `TestAcquireLease_DoesNotSweepOtherKeys`,
  inverted: acquiring one key must leave other keys' rows, expired or live, alone.
- `TestAcquireLease_TakesOverExpiredRowWithoutSweep` — an expired row that is
  still present is taken over by the acquire statement itself.
- `TestSweepExpiredLeases` — expired rows go, the live row stays, a second pass
  is a no-op.
- `TestSweepExpiredLeases_DrainsPastOneBatch` — a 1500-row backlog clears in one
  pass.
- `TestSweepExpiredLeases_ConcurrentTakeover` — the mutual-exclusion regression
  test described above.

## Open questions for review

1. **Placement.** `backgroundMaintenance`, `maintenanceInterval`, and
   `runMaintenanceStep` still live in `outbox.go`, whose package comment is
   about the worker outbox. A follow-up could move them to a `maintenance.go`
   in the same package; kept out of this commit to keep the diff honest to its
   title.
2. **Unelected on every replica.** Deliberate — the sweep is idempotent GC and
   `SKIP LOCKED` keeps replicas out of each other's way — but if we would rather
   have exactly one sweeper, the `outboxMaintenanceLockKey` election pattern is
   right there.
3. **Sweep interval.** One minute is inherited from the outbox tick. Expired
   rows now live up to ~1 minute longer than before, which only matters for
   table size, not behavior.
4. **Metrics.** Deliberately none. Say the word if a rows-collected counter is
   wanted and it can go in with a registry entry.

### DIFF
--- a/cmd/ateapi/internal/store/atepg/atepg.go
+++ b/cmd/ateapi/internal/store/atepg/atepg.go
@@ -50,8 +50,8 @@ const (
 
 type Persistence struct {
 	pool *pgxpool.Pool
-	// watchPool serves the outbox side only: the WatchWorkers pollers
-	// and the partition-maintenance loop.
+	// watchPool serves the background side: the WatchWorkers pollers and the
+	// maintenance loop (outbox partitions and the expired-lease sweep).
 	watchPool             *pgxpool.Pool
 	ownsWatchPool         bool
 	leaseTTL              time.Duration
@@ -164,12 +164,12 @@ func newPersistence(ctx context.Context, pool, watchPool *pgxpool.Pool) (*Persis
 	}
 	go func() {
 		defer close(p.maintenanceDone)
-		p.outboxMaintenance(maintenanceCtx)
+		p.backgroundMaintenance(maintenanceCtx)
 	}()
 	return p, nil
 }
 
-// Close stops the outbox maintenance loop and waits for it to exit,
+// Close stops the background maintenance loop and waits for it to exit,
 // then closes the watch pool if Connect created one. It does not close the
 // main pool, which the caller owns.
 func (p *Persistence) Close() {
@@ -1479,9 +1479,6 @@ const defaultLeaseTTL = 30 * time.Second
 func (p *Persistence) AcquireLease(ctx context.Context, key string) (*store.Lease, error) {
 	ttl := p.leaseTTL
 	token := uuid.NewString()
-	if err := p.cleanupExpiredLeases(ctx); err != nil {
-		slog.WarnContext(ctx, "failed to clean up expired PostgreSQL leases", "error", err)
-	}
 
 	acquired, err := p.acquireLease(ctx, key, token, ttl)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


jpbetz:

Issue #606, opened 2026-07-30
[P1] Actor lock TTL (30s) exceeds router retry budget (15s) — ateapi crash causes guaranteed 15–30s user-visible outage


aditya-shantanu:
**Severity:** P1 (guaranteed user-visible failure window after any ateapi crash)
**Component:** Control Plane + Network Plane — `ateredis.go` + `resumer.go`
**Audit ID:** CP-4, NET-3, NET-6
**Maps to suspect:** S12

---

## Summary

When ateapi crashes mid-workflow, the actor's advisory lock (stored in Valkey with a
30-second TTL) remains held by the dead process. During the 30-second window before
the TTL expires, every `ResumeActor` attempt returns `codes.Aborted` ("another
operation is in progress"). The router's fail-fast retry budget is 15 seconds, and
Envoy's ext_proc timeout is 5 seconds — both shorter than the lock TTL. Every client
request during the 30-second window fails. Additionally, late-joining requests inherit
the remaining budget of the in-flight singleflight, potentially failing immediately if
they join near expiry.

---

## Impact

- Up to 30 seconds of hard actor unavailability after every ateapi pod crash or rolling
  restart.
- All client requests during this window fail with opaque 503/504 errors.
- The error message does not distinguish "ateapi crashed" from "no capacity" — on-call
  cannot immediately diagnose the cause.
- With a rolling restart of 3 ateapi replicas and a 10s restart interval, the total
  outage window can be 30s per pod × 3 = 90s for actors being actively used during
  the rollout.

---

## Root Cause

**File:** `cmd/ateapi/internal/store/ateredis/ateredis.go` line 805:
```go
defaultLockTTL = 30 * time.Second
```

**File:** `cmd/atenet/internal/router/resumer.go` line 36:
```go
failFastResumeBudget = 15 * time.Second
```

**File:** `cmd/atenet/internal/router/xds.go` line 83:
```go
defaultExtProcMessageTimeout = 5 * time.Second
```

The lock TTL goroutine that renews the lock dies with the ateapi process. The lock
key remains in Valkey until the TTL expires. The gap:

```
Lock TTL (30s) >> Envoy ext_proc timeout (5s) ≈ fail-fast budget (15s, but Envoy
cuts it at 5s) >> effective user retry budget (~5s)
```

Additionally, the singleflight in `resumer.go` uses a background context with a fixed
budget created when the **first** caller triggers the flight (lines 142–196). All
[...]

PR #1348, opened 2026-08-31
atepg: shorten the workflow lease TTL to 15s

jpbetz:
### What

Cuts `defaultLeaseTTL` in the atepg store from 30s to 15s, and lifts the two
timings the renew loop derives from it (`ttl/3` renew interval, `2/3 ttl` renew
deadline) out of `renewLeaseLoop` into named `leaseRenewInterval` /
`leaseRenewDeadline` helpers so the TTL stays the only number to tune.

No behavior change beyond the constant. `renewIntervalDivisor` and
`renewDeadlineFraction` are untouched, so the renew cadence goes 10s -> 5s and
the renew deadline 20s -> 10s, both purely as a consequence of the TTL.

Diff is 3 files: `cmd/ateapi/internal/store/atepg/atepg.go`,
`cmd/ateapi/internal/store/atepg/atepg_test.go`, and a one-test drift guard in
`cmd/atenet/internal/router/ingress/parking_test.go`.

Addresses #606.

### Why

A lease left behind by a hard-crashed ateapi replica (SIGKILL, node loss) stays
valid for the remainder of its TTL, and every cold workflow for that actor keeps
failing with `ErrLeaseConflict` until it expires. That surfaces to users: the
conflict becomes `codes.Aborted` (`controlapi/workflow.go`), the router treats
Aborted as retryable while the request is parked (`ingress/resumer.go`), and
once `DefaultParkedRequestBudget` (5s, `ingress/parking.go`) runs out the
request gets a 503. At 30s that is up to six consecutive park budgets of
user-visible 503s per crash; at 15s it is three at worst, roughly 1.5 on
average.

The TTL is bounded from below by two margins that both scale with it, and the
PR makes both explicit in the doc comment because the choice of 15s rests on
them:

1. **Stalled-renew grace.** A renewal still failing when its deadline passes
   cancels the lease context, and that context wraps the atelet restore call
   (`acquireActorLease` returns `lease.Context()`, `workflow_resume.go` passes
   it to `ensureAteletRestored`), so an expiring lease aborts the cold
   operations in flight under it. The renewal fires one interval (`ttl/3`)
   after the last success while its deadline is anchored at that same success
   (`2/3 ttl`), so the **guaranteed** tolerated store outage is the difference,
   `ttl/3` = **5s** — rising to 10s only when the outage happens to begin just
   after a renewal lands.
2. **Fencing.** A holder abandons its lease at `2/3 ttl` while the row only
   becomes acquirable at the full TTL, leaving `ttl - renewDeadline` = `ttl/3`
   = **5s** for an abandoned holder's already-issued atelet RPCs to unwind
   before another replica can start a restore of its own. This change halves
   that margin from 10s to 5s.

A 10s TTL — the other candidate — would cut both margins to 3.3s, which is why
15s and not 10s.

### Note on the 5s vs 10s figure

An earlier draft of this change described the
grace window as the full renew deadline: 10s at a 15s TTL, 6.7s at a 10s TTL.
That is the best case, not the floor. Because the first renewal attempt does not
start until one interval after the last success, the guaranteed window is
`renewDeadline - interval` = `ttl/3`. The comment, the tests and the commit
message now all state the floor.

### Testing done

All run in the worktree at `/home/jpbetz/projects/substrate-worktrees/lease-ttl-15s`.

- `go build ./...` — exit 0
- `go vet ./cmd/ateapi/...` — exit 0
- `go test ./cmd/ateapi/internal/controlapi/... -count=1`
  - `ok github.com/agent-substrate/substrate/cmd/ateapi/internal/controlapi 10.181s`
  - `ok github.com/agent-substrate/substrate/cmd/ateapi/internal/controlapi/functionaltest 18.836s`
- `go test ./cmd/ateapi/internal/store/atepg/ -run Lease -count=1 -v` — all pass:
  `TestAcquireLease_CleansExpiredLeases`, `TestAcquireLease_ExpiresAfterHolderStops`,
  `TestAcquireLease_ConcurrentTakeover`, `TestLeaseTimingConstraints`,
  `TestLeaseRenewGraceWindow/inside_the_renew_deadline`,
  `TestLeaseRenewGraceWindow/past_the_renew_deadline`
- `go test ./cmd/ateapi/internal/store/atepg/ -count=1` (whole package) — `ok ... 14.999s`
- `go test ./cmd/atenet/internal/router/ingress/... -count=1` — `ok ... 0.248s`
- `gofmt -l cmd internal pkg` — empty; `hack/verify/gofmt.sh`, `hack/verify/boilerplate.sh`,
  `hack/verify/golangci-lint.sh` — all exit 0

**New tests.** `TestLeaseTimingConstraints` pins the TTL against the router's
park budget and asserts both margins (grace window and fencing) are at least
`ttl/3` across three TTLs. `TestLeaseRenewGraceWindow` brackets the window
behaviorally at a 6s TTL (2s interval, 4s deadline) by holding a `FOR UPDATE`
row lock on the lease: a blip clearing at 3.5s must **not** cost the lease, and
one clearing at 4.5s **must**.

**Mutation-checked**, so these are load-bearing rather than tautological:
- `renewDeadlineFraction = 0.55` -> `TestLeaseTimingConstraints` fails on the
  grace window at all three TTLs, and `.../inside_the_renew_deadline` fails
  with "lease context was canceled by a renewal that stalled inside its deadline".
- `renewDeadlineFraction = 0.85` -> `TestLeaseTimingConstraints` fails on the
  fencing margin, and `.../past_the_renew_deadline` fails with "lease context
  still live after a renewal stalled past its deadline".

**Coverage caveat for reviewers:** the behavioral lease tests are Docker-gated.
In a lane without Docker, `requirePool` skips them and the suite still reports
`ok`, leaving only `TestLeaseTimingConstraints` (pure constant arithmetic)
running. That is this package's pre-existing convention, not something this
branch changes.

### Open questions

1. **Is a 5s guaranteed grace window enough?** It covers a pgbouncer blip or a
   connection reset, but likely not a full restart of the single-replica
   in-cluster Postgres. If we want ~10s guaranteed, the cheap lever is raising
   `renewDeadlineFraction` (e.g. to 0.9), *not* raising the TTL — but that
   spends the fencing margin, taking it to `ttl/10` = 1.5s. Deliberately not
   done here: it is a separate tuning decision with its own risk, and this PR
   keeps the two margins symmetric at `ttl/3`. Flagging it because it is easy to assume the larger number.
2. **Fencing margin halves from 10s to 5s.** Called out explicitly since it is
   a real reduction in the time an abandoned holder's in-flight atelet work has
   to unwind. Context cancellation reaches the caller immediately but not
   necessarily work already running on atelet. I believe 5s is still ample for
   an RPC to unwind, but it is the part of this change I would most want a
   second opinion on.
3. **Cross-tree constant duplication.** `cmd/ateapi` and
   `cmd/atenet/internal/router/ingress` are mutually unimportable, so the 5s
   park budget is hand-copied into the atepg test. Mitigated with a guard test
   on the ingress side naming atepg as the coupled consumer, so changing either
   trips the other. The cleaner fix — hoisting the budget into a shared
   `internal/` package — felt out of scope for a constant change; say the word
   and I will do it instead.

### DIFF
--- a/cmd/ateapi/internal/store/atepg/atepg.go
+++ b/cmd/ateapi/internal/store/atepg/atepg.go
@@ -1473,8 +1473,39 @@ func (p *Persistence) ListWorkers(ctx context.Context, opts store.ListOptions) (
 // --- Workflow leases ---
 
 // defaultLeaseTTL is how long a lease may go unrenewed before another client
-// can reclaim it.
-const defaultLeaseTTL = 30 * time.Second
+// can reclaim it. The renew cadence and the renew grace window are derived
+// from it by leaseRenewInterval and leaseRenewDeadline, so the TTL is the only
+// number to tune. It is bounded from both sides:
+//
+//   - From above by the router: after a hard crash (SIGKILL, node loss) the
+//     dead replica's row stays valid for the rest of its TTL, and every cold
+//     workflow for that actor keeps failing with ErrLeaseConflict until it
+//     expires. The router parks and retries such a request for its
+//     parked-request budget (5s, DefaultParkedRequestBudget in the atenet
+//     router's ingress package) and then returns 503, so the TTL has to stay
+//     within a small number of those budgets.
+//   - From below by the router as well: the TTL must exceed that budget plus
+//     the slack a renewal can take, or a healthy holder that is merely slow
+//     to renew could be preempted inside a single parked request.
+//   - From below by Postgres: a renewal that is still failing when its
+//     deadline passes cancels the lease context, which is the context that
+//     wraps the atelet restore call, so an expiring lease aborts the cold
+//     operations in flight under it. A renewal is attempted one
+//     leaseRenewInterval after the last successful one, and its deadline sits
+//     leaseRenewDeadline after that same success, so the store outage a
+//     stalled renewal is guaranteed to ride out is the difference between
+//     them: ttl/3, i.e. 5s here. It rises to the full leaseRenewDeadline
+//     (10s) only when the outage happens to begin just after a renewal
+//     lands. An outage past that 5s floor is what costs in-flight cold work,
+//     and it is the reason the TTL is not cut further.
+//   - From below by fencing: a holder abandons its lease at
+//     leaseRenewDeadline after its last successful renewal, while the row
+//     only becomes acquirable at the full TTL, so ttl - leaseRenewDeadline
+//     == ttl/3 (5s here) is the margin in which an abandoned holder's
+//     already-issued atelet RPCs have to unwind before a second replica can
+//     start a restore of its own. Canceling the lease context reaches the
+//     caller at once, but not necessarily work already running on atelet.
+const defaultLeaseTTL = 15 * time.Second
 
 func (p *Persistence) AcquireLease(ctx context.Context, key string) (*store.Lease, error) {
 	ttl := p.leaseTTL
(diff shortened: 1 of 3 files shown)


jpbetz:

PR #1349, opened 2026-08-31
docs: add a store policy for the ateapi persistence layer

jpbetz:
### What

Adds `docs/dev/store.md` (~2 pages) and links it from the README documentation
index. No code, proto, generated, manifest, or vendor file is touched.

The doc covers, in eight short sections plus a closing per-PR checklist:

1. **Data model** — proto is the source of truth; every proto-bearing table
   carries `uid` + `version`; `version` is a per-object counter and never a
   global revision or watch cursor; `atespace` leads every actor-family primary
   key; page tokens stay topology-free.
2. **Write path** — optimistic uid+version CAS only, in either of the two shapes
   the code uses (guards re-asserted in the `WHERE` clause, or checked against a
   row read `FOR UPDATE`); deletes guard differently on purpose; retry only on an
   explicit precondition failure and only after re-reading everything the attempt
   was built from; cross-row invariants live in one store transaction; the actor
   lease guards atelet side effects, not store consistency; no Go work under a
   row lock.
3. **Projected-column checklist** — name the query, write it under the same CAS
   guard, no index without a HOT/WAL benchmark, no write-path tax for
   observability.
4. **Outbox** — a cache, never a source of truth; the `pg_snapshot_xmin` fence
   and what stalls it; any future watch defines its own `xid8` cursor.
5. **Schema evolution** — forward-only, additive/expand-only, N-1 reads the new
   schema, the N-1 round-trip is lossless.
6. **Proto compatibility as a storage rule** — `bytea` is stored verbatim, so
   never reuse a tag, enums are add-only, a retype or removal ships with a row
   rewrite.
7. **Testing bar** — `storecontract` is the contract; concurrency is tested by
   mutate injection, not sleeps; database tests must run in CI rather than skip.
8. **Operational limits** — the honest v1 envelope, and the note that the first
   ceiling is the O(fleet) scheduling scan in ateapi, not PostgreSQL.

### Why

The rules governing the store are real but scattered across `store.go` contract
comments, the `atepg/schema.go` header, and a run of review positions on store
PRs. A reviewer had to remember them and a contributor had no way to read them
before writing a change. Nothing in the doc is a new rule: every claim is
grounded in the code as it stands on `main`, and the citations name the function
or test rather than a line number so they survive edits.

Three places where the doc records a gap rather than papering over it:

- `store.Interface`'s update comments reserve internal retry for backends
  ("mutate may run more than once", "if the retry budget is exhausted"). No
  `atepg` path implements one. The doc says the comment should be narrowed to
  match the backend, and that no caller may depend on a retry.
- The assignment retry loop in `controlapi/workflow_resume.go` refreshes the
  actor but rescans a stale worker cache, so it can claim a second worker and
  strand the first. The doc cites it as the known defect, not as the pattern to
  copy.
- Nothing asserts that unknown proto fields survive an N-1 read-modify-write.
  The doc makes that a `storecontract` case the first change relying on the
  guarantee has to write.

### Testing done

Docs-only change, so no test suite bears on it; nothing was regressed and no
suite was run green as evidence.

- `hack/verify/boilerplate.sh`, `gofmt.sh`, `licenses.sh`, `go-modules.sh`,
  `metrics.sh`, `shellcheck.sh`: all exit 0.
- `hack/verify/proto-fmt.sh`: exits 1 with "Failed to find clang-format: please
  make sure it is in your PATH". Environmental — the tool is not installed here
  and the diff touches no `.proto` file.
- `go vet ./cmd/ateapi/internal/store/...`: exit 0, no output.
  `gofmt -l cmd/ateapi/internal/store/`: no output.
- All five relative links and both anchors resolve
  (`../api-style-guide.md#63-uid`, `#7-resource-freshness-and-optimistic-concurrency`,
  `../code-style-guide.md`, `code-layout.md`), and the README entry's target
  exists.
- Every cited symbol, file, and test name verified present on `main`.
- No British spellings (misspell runs `locale: US`); no trailing whitespace; no
  issue or PR numbers in the doc or the commit message.

### Open questions for review

- **Home for the link.** It is in the README "Documentation & Guides" index, but
  it is a contributor doc and `docs/dev/code-layout.md` is not in that index at
  all. CONTRIBUTING.md or an `AGENTS.md` line under Repository Layout ("changes
  under `cmd/ateapi/internal/store` are reviewed against `docs/dev/store.md`")
  may be the better home. Happy to move it.
- **Overlap with the ops posture.** Section 8's last bullet (in-cluster
  PostgreSQL is dev/eval only, no StatefulSet backup path, an empty database
  reports healthy) belongs to a store-ops doc that does not exist yet. When one
  lands, this should reduce to a one-line pointer so the two do not drift.
- **`docs/api_changes.md`.** Section 6 states the proto-as-storage rules and says
  there is no `docs/api_changes.md` yet. If that doc is being written elsewhere,
  section 6 should shrink to a pointer.
- **Follow-ups the doc implies but does not do:** narrow the `store.Interface`
  retry comments to match `atepg`, and add the N-1 unknown-field round-trip case
  to `storecontract`. Both are small and can be separate PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,7 @@ We provide several sample applications demonstrating Agent Substrate's capabilit
 * [Glossary](docs/glossary.md): Core terms (Actor, Atespace, ActorTemplate, WorkerPool, Worker, ate-api-server, atenet, atelet, ateom) and how they relate.
 * [Integration Repositories](docs/integration-repos.md): Where integrations live, how their repositories are named, and how fixes flow back to core.
 * [Observability Guide](docs/observability.md): Guide to actor logging, metrics, and distributed tracing.
+* [Store Policy](docs/dev/store.md): Rules for the `ateapi` persistence layer — data model, write path, schema evolution, and the review checklist for store changes.
 * [Authentication Guide](docs/authentication.md): Configure trusted JWT providers and human credentials.
 * [Request Parking](docs/request-parking.md): How the router parks requests through transient worker-pool saturation.
 * [Threat Model](docs/threat-model.md): Trust boundaries, assumptions, and known risks.
--- a/docs/dev/store.md
+++ b/docs/dev/store.md
@@ -1,0 +1,262 @@
+# Store Policy
+
+Rules for `ateapi`'s persistence layer: the backend-neutral contract in
+`cmd/ateapi/internal/store` and its PostgreSQL implementation in
+`cmd/ateapi/internal/store/atepg`.
+
+Nothing here is new. It is the contract in `store.go`, the schema header comment
+in `atepg/schema.go`, and a run of store-PR review positions, collected so that a
+store change is reviewed against a written standard rather than reviewer memory.
+Read it before adding a table, a column, a `store.Interface` method, or a
+transaction.
+
+**Non-goals.** Not a design document, a tuning guide, or an operations runbook.
+The proto surface is governed by the [API style guide](../api-style-guide.md) —
+sections [6.3/6.4](../api-style-guide.md#63-uid) and
+[7](../api-style-guide.md#7-resource-freshness-and-optimistic-concurrency) are
+load-bearing here and are not repeated. Go conventions:
+[code style guide](../code-style-guide.md). Layout: [code layout](code-layout.md).
+There is no watch over actors and no global revision; sections 1 and 4 are what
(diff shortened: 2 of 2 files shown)


jpbetz:

Issue #997, opened 2026-08-17
Define least-privilege role design for Postgres


supreme-gg-gg:
ateapi currently connects to PostgreSQL as the `postgres` superuser. The HBA rule accepts any database and requested role from any client holding a certificate issued by the pod-identity CA. ateapi also applies its embedded schema during startup, coupling normal runtime access with DDL privileges.

https://github.com/agent-substrate/substrate/blob/517e48eef436f0a1454e7ae1c3c5db531209e4a9/manifests/ate-install/postgres.yaml#L34

We should move to least-privilege application roles, but the correct role split depends on the PostgreSQL migration design (tracked separately in #901) which has not yet been established. For example, migrations may need a separate owner / migrator role while the running API server only receives DML privileges. HBA rules also should be restricted appropriately.

https://github.com/agent-substrate/substrate/pull/940#discussion_r3793222156

iplay88keys:
@bowei, @jpbetz, @juli4n, Here's my proposal for review. I will work on an implementation:

## Problem

`ateapi` currently uses one PostgreSQL identity to create the schema, apply Goose migrations, serve requests, and maintain outbox partitions. The reference PostgreSQL manifest also accepts any database and requested role from a client with a trusted certificate. This makes every database operation run as the schema owner.

Startup migrations are only part of the problem: the running server also creates, truncates, and drops outbox partitions. Separating migrations without addressing that maintenance would not produce a DML-only runtime.

## Proposed design

Give `ateapi` two PostgreSQL connection strings and role classes:

- **DDL/maintenance:** owns the Substrate schema and may create the schema, apply Goose migrations, perform data migrations, and maintain outbox partitions. It is not a PostgreSQL superuser and has no cluster-wide administration privileges.
- **DML/runtime:** may connect, use the Substrate schema, and perform the table and sequence operations required to serve API traffic. It cannot create schemas, modify the migration ledger, or create, alter, truncate, or drop objects.

Keep the existing startup lifecycle. Schema creation and migrations use the DDL connection; request handling and watch polling use DML connections; outbox partition creation, stray cleanup, and retention drops use the DDL connection. The current advisory locks continue to coordinate migrations and maintenance across replicas.

The existing PostgreSQL connection remains the DML input. A new DDL connection input defaults to the DML connection, allowing simpler single-role deployments. For two-role deployments, deployment tooling supplies the credentials and provisions a clean runtime role without ownership or DDL privileges.

[...]

Issue #626, opened 2026-07-30
[P2] After Valkey data loss or flush, workers are not repopulated until next informer resync or ateapi restart


aditya-shantanu:
**Severity:** P2 (all ResumActor calls fail for up to minutes after Valkey data loss)
**Component:** Control Plane — `cmd/ateapi/internal/controlapi/syncer.go`
**Confirmed:** Yes — observed live during stress testing on GKE cluster `substrate-poc`

---

## Summary

After a Valkey flush or data loss event, the worker records for all live worker pods are
gone from Redis. The worker cache immediately shows zero workers (via periodic relist).
However, the `WorkerPoolSyncer` does not automatically repopulate worker records unless
it receives a Kubernetes pod event. The informer's re-sync period (typically 30–60 seconds)
eventually triggers update events for all pods, causing the syncer to repopulate the
workers — but during that window every `ResumeActor` call fails with
"no free workers available." A manual ateapi restart immediately triggers a full sync
and restores scheduling in ~5 seconds.

---

## Impact

- After any Valkey data loss (node crash, compaction failure, accidental flush), the
  cluster is completely unable to resume actors for 30–60 seconds.
- This is in addition to ISSUE-001 (Valkey wipe → double-placement risk). Even if an
  operator avoids double-placement by fencing, they face an extended scheduling outage.
- Operators who perform a `debug-flush-redis` for debugging and don't restart ateapi
  will observe a "no free workers" state that seems permanent until they know about
  the workaround.
- No log message or metric signals this condition — the only observable symptom is
  `100%` resume failure rate.

---

## Root Cause

**File:** `cmd/ateapi/internal/controlapi/syncer.go` lines 51–109 (`Start()`)

`WorkerPoolSyncer.Start()` calls `syncWorkerToStore` for all pods currently in the
informer cache at startup, then starts watching for new events. After startup, worker
records are only updated when the informer fires pod events (Add/Update/Delete).

After a Valkey flush:
1. Worker records are gone from Redis.
2. Worker cache calls `relist()` → finds 0 workers → `cache.ready = true` but worker
   list is empty.
3. No pod events fire (pods haven't changed).
4. Syncer has no trigger to call `syncWorkerToStore` again.
5. Workers remain invisible to the scheduler until the informer's resync fires.

The informer's resync period (`controller-runtime` default: 10 hours or whatever the
controller configures it to be) is not guaranteed to be short. In practice, informers
[...]

Issue #603, opened 2026-07-30
[P0] Valkey data loss orphans all running workloads and enables double-placement


aditya-shantanu:
**Severity:** P0 (permanent damage without manual repair)
**Component:** Control Plane — `cmd/ateapi/internal/store/ateredis`
**Audit ID:** CP-5
**Maps to suspect:** S6

---

## Summary

All actor and worker state lives exclusively in Valkey. After a full Valkey flush or
data loss, ateapi recreates worker records as free (no assignment). The ateletes on
those pods still host live workloads. The scheduler treats the newly-free workers as
available and assigns new actors to them, resulting in two workloads running inside
the same sandbox — a double-placement. There is no mechanism for ateapi to query
atelet about its current workloads; assignments cannot self-heal.

---

## Impact

- All existing actors become invisible (empty list).
- Running actor sandboxes continue executing but are unaddressable.
- New actor assignments race against live workloads — potential process collision,
  data corruption, or security boundary breach between two actors sharing a sandbox.
- Manual fencing of every node (cordon + atelet restart) is required before safe
  recovery.

---

## Root Cause

**File:** `cmd/ateapi/internal/store/ateredis/ateredis.go` lines 36–39  
**File:** `cmd/ateapi/internal/controlapi/syncer.go` lines 88–100

On ateapi startup, `syncWorkerToStore` upserts a fresh worker record for every live
pod with `Assignment = nil`. It does not query atelet for what workload (if any) is
currently running in that pod's sandbox. Once this upsert completes the worker is
treated as free. The very next `ResumeActor` call will ask the scheduler to assign it,
and the scheduler will succeed — sending a `Run`/`Restore` RPC to an atelet that
already hosts a live actor.

The package-level comment in `ateredis.go` (lines 36–39) explicitly notes:
> "Redis Lua is not ACID — power failure, etc. may leave us with half of the effects
> of a script applied."

There is no `atelet.ListWorkloads()` RPC or equivalent registration call that would
let ateapi reconstruct assignment state after data loss.

---

## Steps to Reproduce

1. Start a cluster with several RUNNING actors distributed across worker pods.
2. Confirm actors are healthy and responding to requests.
3. Flush all Valkey data:
   ```bash
   kubectl exec -n ate-system valkey-cluster-0 -- valkey-cli FLUSHALL
   ```
4. Restart ateapi (or wait for it to detect Valkey reconnection):
   ```bash
   kubectl rollout restart -n ate-system deployment/ate-api-server
[...]

PR #1350, opened 2026-08-31
postgres: add an ops guide and tune the shipped database

jpbetz:
### What

Adds `docs/ops/postgres.md` and moves the shipped PostgreSQL off compiled-in
defaults.

- **`docs/ops/postgres.md`** (new, linked from `README.md`'s doc index and from
  the state-store section of `docs/architecture.md`): what the store holds and
  why it is the system of record; the two supported postures; what actually
  loses data and what does not; a runbook for a lost or rolled-back store;
  sizing, tuning and connection budgeting with the derivations; what is and is
  not instrumented today, with `psql` queries where nothing is; operational
  timeouts; and the superuser/trust posture with pointers to
  `docs/threat-model.md` T-04, `docs/authentication.md`, and issues #997 and
  #999.
- **`manifests/ate-install/postgres.yaml`**: `shared_buffers = 512MB`,
  `effective_cache_size = 1GB`, `max_connections = 100` with the formula that
  produced it in a comment, `idle_in_transaction_session_timeout = 60s`, the
  memory request raised to meet the 2Gi limit, and a PodDisruptionBudget
  (`minAvailable: 1`, `unhealthyPodEvictionPolicy: AlwaysAllow`).
- **`hack/install-ate.sh`** and **`cmd/ate-setup/internal/config/config.go`**:
  `pool_max_conns=12` in the default DSN, plus a test that pins the two copies
  of that string to each other and asserts the parameter really does cap the
  pool.

### Why

The store is the system of record for actors, workers, leases and snapshot tags,
and nothing can re-derive that state from the data plane — `atelet` exposes no
"what are you hosting?" RPC. It shipped with no operator documentation and on
compiled-in defaults.

The doc's load-bearing claim is that **an empty store is worse than a down
store**. The schema is applied with `CREATE TABLE IF NOT EXISTS` at every start,
so a wiped database is indistinguishable from a first install; `atecontroller`'s
worker syncer then treats every `GetWorker` `NotFound` as a new worker and
re-registers the live fleet `STATE_ACTIVE` at full capacity
(`cmd/atecontroller/internal/workersync/syncer.go`, `createOrUpdateWorker`).
Those pods are still hosting sandboxes. The scheduler places on top of them.
Hence the runbook: if you lose the store, drain and re-create the worker fleet
before serving traffic again. That is the same mechanism as the P0 filed against
the previous Valkey store (#603) and its P2 follow-up (#626), both closed as
obsolete when Valkey was removed rather than fixed; the path is
backend-agnostic. Reopening #603 for the Postgres path is proposed separately.

Every tuning number is derived from the memory limit this manifest actually
declares rather than picked, and the derivation is in the manifest so the next
person changing the replica count or the pool size knows what else moves.
`pool_max_conns` matters most: without it pgxpool sizes the pool from
`max(4, runtime.NumCPU())`, which reports the *node's* cores regardless of the
container's CPU limit — four connections on a small node and sixty-four on a
large one, neither of them chosen, and both feeding into one `max_connections`.

`idle_in_transaction_session_timeout` targets the failure the doc spends the
most words on: `WatchWorkers` only delivers outbox rows below the cluster's xmin
horizon, so any transaction left open anywhere holds the horizon back and stalls
worker-event delivery for every replica — silently, since an empty poll batch
looks like an idle feed. 60s cannot touch legitimate work; the timeout ignores
sessions that are executing or waiting on a lock.

### Deliberately not in this PR

Backup CronJobs and PITR tooling for the in-cluster StatefulSet, store-side wipe
detection, and store HA (#181). A half-backup in the manifests would invite
operators to trust it; the supported answer for data that matters is a managed
instance reached through `ATE_API_POSTGRES_CONNECTION_STRING`, and the doc says
so in those words.

### Testing done

- `kubectl apply --dry-run=client -f manifests/ate-install/postgres.yaml` — all
  four objects accepted, including the new PDB.
- `kubectl kustomize manifests/ate-install/kind --load-restrictor
  LoadRestrictionsNone` — renders clean. The kind overlay patches only CPU and
  disk, so the 2Gi-derived tuning holds there; postgres renders with
  requests `cpu: 250m` / `memory: 2Gi` and limits `cpu: "2"` / `memory: 2Gi`,
  and total memory requests across the overlay are 2.1Gi.
- `go build ./...`, `go test ./cmd/ate-setup/...`, `go vet
  ./cmd/ate-setup/...`, `gofmt -l cmd/ate-setup/` — all clean. The new drift
  test was checked negatively: changing `pool_max_conns` in only one of the two
  files fails it with both strings printed.
- `hack/verify/`: `boilerplate`, `gofmt`, `go-modules`, `licenses`,
  `shellcheck`, `metrics`, `threats`, `golangci-lint` — all rc=0.
- `bash -n hack/install-ate.sh` — clean.
- All five relative links in the new doc resolve, as do the two new links into
  it.

Not run: the repo-wide test suite and the atepg testcontainers tests. No store
code is touched and there are no golden-file manifest tests.

### Open questions for review

1. `max_connections = 100` happens to equal the compiled-in default. It is
   stated anyway so the ceiling is an explicit thing to re-derive; say if you
   would rather not carry a line that changes nothing today.
2. The PDB makes a node drain on the postgres node stall until an operator
   intervenes. That is the intent, and the guide has a "Node maintenance"
   section saying so — but it is a behavior change for anyone running node-pool
   upgrades, and it is worth an explicit yes.
3. The memory request rises 1Gi → 2Gi, which raises what the pod reserves on a
   kind node too. Total memory requests in the kind overlay come to 2.1Gi, so
   this should be comfortable on a CI runner, but it is the one number that
   affects scheduling.
4. `pool_max_conns=12` is a first guess sized to the budget, not a measurement.
   There is no store instrumentation to size it against yet; that is itself
   noted in the doc as a blind spot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,7 @@ We provide several sample applications demonstrating Agent Substrate's capabilit
 * [Glossary](docs/glossary.md): Core terms (Actor, Atespace, ActorTemplate, WorkerPool, Worker, ate-api-server, atenet, atelet, ateom) and how they relate.
 * [Integration Repositories](docs/integration-repos.md): Where integrations live, how their repositories are named, and how fixes flow back to core.
 * [Observability Guide](docs/observability.md): Guide to actor logging, metrics, and distributed tracing.
+* [Operating the PostgreSQL Store](docs/ops/postgres.md): Supported database postures, sizing and connection budgeting, and what to do if the store is lost.
 * [Authentication Guide](docs/authentication.md): Configure trusted JWT providers and human credentials.
 * [Request Parking](docs/request-parking.md): How the router parks requests through transient worker-pool saturation.
 * [Threat Model](docs/threat-model.md): Trust boundaries, assumptions, and known risks.
--- a/cmd/ate-setup/internal/config/config.go
+++ b/cmd/ate-setup/internal/config/config.go
@@ -44,8 +44,10 @@ const DefaultRolloutTimeout = 60 * time.Second
 // DefaultPostgresConnectionString mirrors default_postgres_connection_string in
 // the shell installer: the apiserver reaches PostgreSQL over mTLS using the
 // podcertificate controller's projected servicedns trust bundle and its own
-// podidentity credential bundle.
-const DefaultPostgresConnectionString = "postgresql://postgres@postgres.ate-system.svc:5432/atepg?sslmode=verify-full&sslrootcert=/run/servicedns.podcert.ate.dev/trust-bundle.pem&sslcert=/run/podidentity.podcert.ate.dev/credential-bundle.pem&sslkey=/run/podidentity.podcert.ate.dev/credential-bundle.pem"
+// podidentity credential bundle. pool_max_conns caps the pgxpool that would
+// otherwise be sized from the node's CPU count; it is a term in the
+// max_connections budget in manifests/ate-install/postgres.yaml.
+const DefaultPostgresConnectionString = "postgresql://postgres@postgres.ate-system.svc:5432/atepg?sslmode=verify-full&sslrootcert=/run/servicedns.podcert.ate.dev/trust-bundle.pem&sslcert=/run/podidentity.podcert.ate.dev/credential-bundle.pem&sslkey=/run/podidentity.podcert.ate.dev/credential-bundle.pem&pool_max_conns=12"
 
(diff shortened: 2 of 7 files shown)


jpbetz:

PR #1354, opened 2026-08-31
[WIP] e2e: check an actor's address families against its worker pod's

ygao-g:
Part of #246.

Nothing outside a sandbox can see which address families the actor inside it
ended up with, and the two sandbox classes reach that state by different
routes: gVisor's runsc adopts the interior netns wholesale, while a micro-VM
guest has to be told over the kata-agent channel. The counter demo now serves
`/netinfo`, reporting its interfaces and the addresses on them, and
`TestActorAddressFamilies` compares that against the worker pod's own IPs.

The assertion is an equivalence rather than "expect IPv6", so it means
something on every cluster: on a dual-stack one it is the positive check, and
on the IPv4-only clusters CI runs today it asserts that an actor was *not*
given an address its pod cannot route. It runs on both sandbox classes through
`E2E_SANDBOX_CLASS`, so the merge-gating `e2e-test` job picks it up for both
without any workflow change. Link-local does not count as IPv6 — every link
has an `fe80::`, so counting it would make the check vacuous.

- [x] Tests pass

### DIFF
--- a/demos/counter/counter.go
+++ b/demos/counter/counter.go
@@ -21,11 +21,13 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"log/slog"
 	"net"
 	"net/http"
+	"net/netip"
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


BenTheElder:
... are we assuming that these will ever mismatch? why? does it even matter if we're sending traffic through the tunnel?

PR #1355, opened 2026-09-01
Update existing demos and benchmark tests to use substrate ActorTemplate resource

zoez7:
This PR is very large since it updates all existing demos and benchmark workloads to use the new ActorTemplate substrate proto. 
Please use the "Commits" tab to review individual commits.

Verifications done: 
* Used this script: gpaste/5143788763348992 to verify that the change from CRD -> proto are equivalent.
* The e2e tests are using the new susbtrate resources.
* Picked the parking demo to run e2e manually: gpaste/6193361380311040


benchmarking/automation/orchestrator.py
```diff
@@ -348,20 +348,8 @@ def deploy_workloads(
...
-        [
-            "kubectl",
-            "wait",
-            "--for=condition=Ready",
-            "--all",
-            "actortemplates",
-            "-n",
-            "benchmark-workloads",
-            "--timeout=300s",
-        ]
```
EItanya: Not necessarily for right now, but will we need something similar for the Substrate APIs? I saw an issue come through about async APIs a couple days ago

cmd/kubectl-ate/internal/printer/printer.go
```diff
@@ -273,9 +273,15 @@ func PrintActorTemplatesTo(out io.Writer, templates []*ateapipb.ActorTemplate, f
 		w := tabwriter.NewWriter(out, 0, 0, 3, ' ', 0)
 		fmt.Fprintln(w, "ATESPACE\tNAME\tSANDBOX CLASS\tSTATUS\tAGE")
 		for _, t := range templates {
-			status := "Failed"
-			if t.GetStatus().GetGoldenSnapshotStatus().GetGoldenSnapshot() != nil {
+			// An empty golden snapshot ref without an error means the golden
```
juli4n: Can we leave this as is and change the consumer to handle this? I think the CLI tool should print the resource exactly as is to avoid confusion.

zoez7: Done in 4fc0d3ce8b029664c60ab98de2f7b139be28ed53

README.md
HavenXia, line 100: Should be `./hack/install-ate.sh --deploy-demo-counter-substrate` right?

zoez7: I deleted the non-substrate variant, now `hack/install-ate-kind.sh --deploy-demo-counter` should deploy it in substrate.

HavenXia (approved):

Issue #811, opened 2026-08-08
Cold-node gVisor release extract (~20s, silent, uncancellable) exceeds the router resume budget — first resume after a release bump always fails


dberkov:
## Symptom

On a node whose gVisor asset cache is cold, the first `ResumeActor` reliably fails with:

```
workflow failed at step CallAteletRestore: while restoring durable snapshot:
rpc error: code = Canceled desc = context canceled
```

and atelet's `Restore` reports (with newly added pull logging):

```
while creating "counter" OCI bundle: in imageCache.EnsureImage:
in remote.Image (registry "gcr.io", gcp_auth=true, after 0s, ctx err: context canceled):
Get "https://gcr.io/v2/": context canceled     elapsed-time: 19.9s
```

The actor image is not the problem — zero bytes of it were fetched (`after 0s`); the context was already dead when the pull began.

## Timeline (GKE, `demo/my-counter-4`, trace `96d4b7e35eee327b1ec90a0dc3950df6`)

| t | event |
|---|---|
| 15:51:01.42–01.44 | snapshot files downloaded + zstd-decompressed (5.2MB total, <60ms) |
| ~15:51:06 | caller (atenet-router, ~5s resume deadline) cancels the context |
| 15:51:01.44–15:51:21.17 | **19.7s silent gap**: `ensureSandboxAssets` downloads `gs://gvisor/releases/release/20260803/x86_64/gvisor.tar.bz2` and extracts it via `extractTarBz2` |
| 15:51:21.17 | prep leg reaches the OCI image pull; first HTTP request dies instantly on the canceled context |

## Root cause

`cmd/atelet/sandbox_assets.go` `fetchGVisorRelease` → `extractTarBz2` (introduced in #684, when the gVisor asset became a release tarball instead of a bare `runsc` binary):

- **Silent**: no log lines for the download or the extraction — a 20s hole in the trace.
- **Uncancellable**: the bzip2+tar extraction loop has no `ctx` checks (Go's stdlib bzip2 is single-threaded and slow), so it runs ~15s past the caller's cancellation.
- **Fleet-wide cold**: #787 bumped the pinned release, invalidating the content-addressed cache (`StaticFilesDir`, hostPath) on every node at once — so every node's *first* resume after the rollout hits this.

#684's benchmarks measure warm-asset-cache steady state (which got faster); the per-node, per-release cold extraction cost was invisible to them.

[...]

PR #1207, opened 2026-08-25
microvm: support Microsoft Hypervisor devices

Bickor:
> This is an exploratory PR based on an AKS interoperability investigation. I am opening it as a draft to discuss whether hypervisor selection belongs at controller scope or should move into `SandboxConfig` / `WorkerPool` compatibility metadata.

## Summary

- keep `/dev/kvm` as the default MicroVM hypervisor device
- allow `ate-controller` to generate workers using `/dev/mshv`
- restrict MSHV workers to explicitly labeled x86-64 nodes
- reject arbitrary device paths before they become privileged `hostPath` mounts
- document backend and snapshot compatibility constraints

## Why

AKS `KataVmIsolation` nodes use an Azure Linux Microsoft Hypervisor root-partition image. These nodes expose `/dev/mshv`, not `/dev/kvm`, so the existing hard-coded KVM worker pod cannot start even though Cloud Hypervisor is built with MSHV support.

The new controller setting is available as either:

```text
--microvm-hypervisor-device=/dev/mshv
ATE_MICROVM_HYPERVISOR_DEVICE=/dev/mshv
```

Only `/dev/kvm` and `/dev/mshv` are accepted. MSHV workers also require:

```text
ate.dev/sandboxClass=microvm
ate.dev/microvm-hypervisor=mshv
kubernetes.io/arch=amd64
```

## AKS validation

Validated on:

- AKS Kubernetes 1.36.2
- `Standard_D8s_v5`
- `KataVmIsolation`
- `AKSAzureLinux-V3katagen2-202608.06.1`
- host kernel `6.6.137.mshv2-1.azl3`
- Cloud Hypervisor v53.0
- Kata 4.0 guest assets
- virtiofsd v1.14.0

Using the real `ateom-microvm` image and native gRPC API, the following succeeded through `/dev/mshv`:

1. `RunWorkload`
2. Cloud Hypervisor VM boot
3. kata-agent connection over vsock
4. actor container creation and HTTP readiness
5. full `CheckpointWorkload`
6. full `RestoreWorkload`
7. guest-agent resource metrics after restore

The checkpoint contained `base-id`, `config.json`, `memory-ranges`, `rootfs-upper.tar`, and `state.json`. An in-memory counter returned `1` before checkpoint and `2` after restore, confirming memory continuity.

This did not use the host's Kata containerd shim. The outer worker remained a normal `runc` pod; `ateom-microvm` launched Cloud Hypervisor and drove snapshot/restore directly.

## Current AKS limitation

[...]

PR #1358, opened 2026-09-01
atelet: pre-download sandbox assets from SandboxConfigs

dberkov:
## Summary

atelet now watches SandboxConfig objects and pre-downloads their sandbox assets into the node's content-addressed cache in the background, instead of only fetching them inside the first Run/Restore on the node.

Related to #811 — this does **not** fix it, it is one more optimization step toward it. It removes the cold-cache download+extract from the first Run/Restore in the common case (node boot, or a release bump while the node is idle), but the window between a SandboxConfig change and prewarm completion still exists, and the extraction itself remains on the critical path when a resume lands inside that window. The uncancellable/silent extraction called out in #811 is unchanged by this PR.

## How it works

- A handler on the existing shared informer factory enqueues SandboxConfig add/update events; registration happens after cache sync, so a freshly booted node replays every existing config as a synthetic Add and prewarms immediately.
- A single background worker drains the queue, so concurrent prewarms never compete for node bandwidth. Downloads are jittered by up to 30s so a config rollout doesn't hit the bucket from every node in the fleet at once.
- gVisor configs prewarm on every node. Micro-VM configs prewarm only where `/dev/kvm` exists (`microvmNodeCapable`, the same signal the device plugin advertises as `ate.dev/kvm`) — known at atelet startup, before any WorkerPool schedules to the node, and avoids pulling multi-hundred-MiB guest images onto nodes that can never run them.
- Prewarm is purely best-effort: failures are logged and the on-demand fetch in `ensureSandboxAssets` remains the correctness path. Racing the two is safe because both install content-addressed files via atomic rename.
- The atelet ClusterRole gains get/list/watch on `sandboxconfigs`.

Left as TODOs in the code: retry-with-backoff on prewarm failure, and GC of cached assets no longer referenced by any SandboxConfig or on-node actor record.

## Pause image prewarm (cherry-picked)

This PR also carries `atelet: prewarm the pause image alongside sandbox assets` (cherry-picked from 2c7dd31f). The prewarmer downloaded the runtime binaries but not the SandboxConfig's pause image, so every node still pulled it from the registry inside its first Run/Restore. At benchmark start that is a synchronized fleet-wide pull burst — registry.k8s.io answered a 1000-node run with 429s, failing Restores — and a failed pull writes no cache record, so each subsequent Restore re-pinged the registry and fed the rate limiter. The prewarm worker now pulls the pause image into the image cache concurrently with the asset downloads (different backends, so neither fetch waits on or fails the other), inheriting the existing per-config jitter. It remains best-effort: the pull inside `prepareOCIBundles` is still the correctness path.

## Testing

- Unit tests cover the arch projection (`recordFromSandboxConfig`), the enqueue filtering (KVM gating, unknown classes, full-queue drop), and `microvmNodeCapable` negative cases.
- An end-to-end test drives a SandboxConfig from a fake clientset through the real informer into the prewarm worker and verifies the asset lands in the static-files cache.
- The pause-image commit adds `TestPrewarmPauseImage`, pulling from a local test registry to verify the image lands in the image cache and that an asset-fetch failure does not stop the pull.
- `go vet` and the full `cmd/atelet` test package pass.

### DIFF
--- a/cmd/atelet/deviceplugin.go
+++ b/cmd/atelet/deviceplugin.go
@@ -30,6 +30,21 @@ import (
 // exist; workers are handed the real host paths, which kubelet resolves.
 const hostDevRoot = "/host/dev"
 
+// microvmNodeCapable reports whether this node can host micro-VM workers:
+// cloud-hypervisor needs /dev/kvm (VmCreate fails with EPERM without it), and
+// worker pods request the matching extended resource, so they only schedule to
+// nodes where the device exists. Device presence is therefore the earliest
+// reliable eligibility signal — known at atelet startup, before any WorkerPool
+// schedules here.
+func microvmNodeCapable(devRoot string) bool {
+	for _, d := range deviceplugin.SandboxDevices {
+		if d.ResourceName == deviceplugin.ResourceKVM {
+			return d.Present(devRoot)
+		}
+	}
+	return false
+}
+
 // startDevicePlugins advertises the sandbox host devices present on this node to
 // kubelet as extended resources, in the background for the lifetime of ctx. This
 // is what lets a worker be granted /dev/kvm without running privileged. atelet
--- a/cmd/atelet/deviceplugin.go
+++ b/cmd/atelet/deviceplugin.go
@@ -36,6 +36,12 @@ const hostDevRoot = "/host/dev"
 // nodes where the device exists. Device presence is therefore the earliest
 // reliable eligibility signal — known at atelet startup, before any WorkerPool
 // schedules here.
+//
+// TODO(https://github.com/agent-substrate/substrate/pull/1207): /dev/kvm is
+// not the only micro-VM hypervisor device; once /dev/mshv support lands, an
+// mshv-only node would be wrongly reported incapable here. Treat presence of
+// any micro-VM hypervisor device in SandboxDevices as capability, not KVM
+// alone.
 func microvmNodeCapable(devRoot string) bool {
 	for _, d := range deviceplugin.SandboxDevices {
 		if d.ResourceName == deviceplugin.ResourceKVM {
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -279,6 +279,9 @@ func main() {
 	// is unlikely to be used with frequency.
 	ateFactory := externalversions.NewSharedInformerFactory(ateClient, 0)
 	csiDriverConfigLister := ateFactory.Api().V1alpha1().CSIDriverConfigs().Lister()
+	// Requested before Start so the factory runs it; the prewarm handler is
+	// attached after the herder exists (see startSandboxAssetPrewarm below).
+	sandboxConfigInformer := ateFactory.Api().V1alpha1().SandboxConfigs().Informer()
 
 	// Start an informer on the ClusterTrustBundle we care about (currently
 	// only the egress trust bundle). The v1beta1 API is feature-gated: on a
@@ -308,6 +311,12 @@ func main() {
 		csiDriverConfigLister,
 		clusterTrustBundleLister,
 	)
+	// Pre-download sandbox assets as SandboxConfigs appear/change so the first
+	// Run/Restore on this node hits the cache. Best-effort: on failure the
+	// on-demand fetch in ensureSandboxAssets still covers correctness.
+	if err := startSandboxAssetPrewarm(ctx, sandboxConfigInformer, wmService, microvmNodeCapable(hostDevRoot)); err != nil {
+		slog.ErrorContext(ctx, "Sandbox asset prewarm disabled", slog.Any("err", err))
+	}
 	dialOpts, err := ateapiauth.DialOptions(ateapiauth.ClientConfig{
 		K8sClient:        k8sClient,
 		CAFile:           *ateapiCAFile,
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -279,9 +279,6 @@ func main() {
 	// is unlikely to be used with frequency.
 	ateFactory := externalversions.NewSharedInformerFactory(ateClient, 0)
 	csiDriverConfigLister := ateFactory.Api().V1alpha1().CSIDriverConfigs().Lister()
-	// Requested before Start so the factory runs it; the prewarm handler is
-	// attached after the herder exists (see startSandboxAssetPrewarm below).
-	sandboxConfigInformer := ateFactory.Api().V1alpha1().SandboxConfigs().Informer()
 
 	// Start an informer on the ClusterTrustBundle we care about (currently
 	// only the egress trust bundle). The v1beta1 API is feature-gated: on a
@@ -314,9 +311,21 @@ func main() {
 	// Pre-download sandbox assets as SandboxConfigs appear/change so the first
 	// Run/Restore on this node hits the cache. Best-effort: on failure the
 	// on-demand fetch in ensureSandboxAssets still covers correctness.
+	//
+	// The informer is requested only now, after the factory's blocking
+	// WaitForCacheSync above, so it cannot hold up atelet startup when its
+	// list/watch fails (e.g. Forbidden while the ClusterRole rollout lags the
+	// binary): the reflector retries in the background and prewarm stays cold
+	// until it recovers.
+	sandboxConfigInformer := ateFactory.Api().V1alpha1().SandboxConfigs().Informer()
 	if err := startSandboxAssetPrewarm(ctx, sandboxConfigInformer, wmService, microvmNodeCapable(hostDevRoot)); err != nil {
 		slog.ErrorContext(ctx, "Sandbox asset prewarm disabled", slog.Any("err", err))
 	}
+	// The factory only runs informers that exist when Start is called: the
+	// Start above predates the SandboxConfigs informer, so without this call
+	// it would never list or watch. Start is idempotent per informer — this
+	// launches the new one and leaves the already-running ones untouched.
+	ateFactory.Start(stopCh)
 	dialOpts, err := ateapiauth.DialOptions(ateapiauth.ClientConfig{
 		K8sClient:        k8sClient,
 		CAFile:           *ateapiCAFile,
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -318,7 +318,7 @@ func main() {
 	// binary): the reflector retries in the background and prewarm stays cold
 	// until it recovers.
 	sandboxConfigInformer := ateFactory.Api().V1alpha1().SandboxConfigs().Informer()
-	if err := startSandboxAssetPrewarm(ctx, sandboxConfigInformer, wmService, microvmNodeCapable(hostDevRoot)); err != nil {
+	if err := startSandboxAssetPrewarm(ctx, sandboxConfigInformer, wmService, imageCache, microvmNodeCapable(hostDevRoot)); err != nil {
 		slog.ErrorContext(ctx, "Sandbox asset prewarm disabled", slog.Any("err", err))
 	}
 	// The factory only runs informers that exist when Start is called: the
--- a/cmd/atelet/sandbox_prewarm.go
+++ b/cmd/atelet/sandbox_prewarm.go
@@ -24,16 +24,28 @@ import (
 	"sync"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 
 	"github.com/agent-substrate/substrate/pkg/api/v1alpha1"
+	listersv1alpha1 "github.com/agent-substrate/substrate/pkg/client/listers/api/v1alpha1"
 )
 
 // prewarmMaxJitter spreads the fleet's asset downloads after a SandboxConfig
 // change. Every atelet observes a create/update within about a second, and
-// without jitter they would all open the same bucket objects at once. A var so
-// tests can zero it.
+// without jitter they would all open the same bucket objects at once. The
+// jitter is applied as the enqueue delay, so duplicate events arriving inside
+// the window collapse into one queue entry instead of stacking waits in the
+// worker. A var so tests can zero it.
 var prewarmMaxJitter = 30 * time.Second
+
+// prewarmMaxRetries bounds how often one config's failed prewarm is retried
+// before it is dropped until the next config event (or first use, which stays
+// the correctness path). The backoff below caps at 5 minutes, so this covers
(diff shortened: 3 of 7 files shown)


BenTheElder:
> gVisor configs prewarm on every node. Micro-VM configs prewarm only where /dev/kvm exists (microvmNodeCapable, the same signal the device plugin advertises as ate.dev/kvm) — known at atelet startup, before any WorkerPool schedules to the node, and avoids pulling multi-hundred-MiB guest images onto nodes that can never run them.

Taking a note ... this is yet another place that hardcodes knowledge of the two classes :(

I think this factoring is going to prove badly later.

Also ... right now we run atelet on *all* nodes. So even nodes that never schedule an actor will start downloading these ..?

dberkov:
>Taking a note ... this is yet another place that hardcodes knowledge of the two classes :(
I think this factoring is going to prove badly later.
Also ... right now we run atelet on all nodes. So even nodes that never schedule an actor will start downloading these ..?

We need run atelet only on the nodes where we are planning to run workloads. I did not want to add dependency on the worker, since I wanted to download images ASAP.

https://github.com/google/gvisor/pull/14528 been merged today, so tomorrow we will update sandboxConfig parameters to use zstd images, so the first run will go down to 2sec from 23 sec.  In addition to it, based on https://github.com/agent-substrate/substrate/issues/811#issuecomment-5483138778, the size of gVisor image will be shrink even more, so might be all the pre-warming for gVisor will not be required. On the other hand, pre-warming non gVisor images might be still usefull, since they are large.

BenTheElder:
> On the other hand, pre-warming non gVisor images might be still usefull, since they are large.

Yeah, I'm just a little worried about space usage.

I have some ideas about how we can get that down, but we will need somewhere to host those scripts and pre-built artifacts.

cmd/atelet/deviceplugin.go
```diff
@@ -30,6 +30,21 @@ import (
 // exist; workers are handed the real host paths, which kubelet resolves.
 const hostDevRoot = "/host/dev"
 
+// microvmNodeCapable reports whether this node can host micro-VM workers:
```
BenTheElder, line 33: TODO: https://github.com/agent-substrate/substrate/pull/1207

cmd/atelet/main.go
```diff
@@ -308,6 +311,12 @@ func main() {
 		csiDriverConfigLister,
 		clusterTrustBundleLister,
 	)
+	// Pre-download sandbox assets as SandboxConfigs appear/change so the first
+	// Run/Restore on this node hits the cache. Best-effort: on failure the
+	// on-demand fetch in ensureSandboxAssets still covers correctness.
+	if err := startSandboxAssetPrewarm(ctx, sandboxConfigInformer, wmService, microvmNodeCapable(hostDevRoot)); err != nil {
```
BenTheElder: Putting the SandboxConfigs informer on ateFactory before Start makes WaitForCacheSync block forever if list/watch is Forbidden (RBAC skew during rollout).

This goes from a best effort optimization to hanging the node ...?

cmd/atelet/sandbox_prewarm.go
```diff
@@ -0,0 +1,201 @@
...
+	var assetErr, imageErr error
+	var wg sync.WaitGroup
+	wg.Go(func() {
+		_, assetErr = p.herder.ensureSandboxAssets(ctx, rec)
+	})
+	wg.Go(func() {
+		if rec.PauseImage == "" {
+			return
+		}
+		if _, err := p.herder.imageCache.EnsureImage(ctx, rec.PauseImage); err != nil {
```
BenTheElder: this is running on main's ctxt.Background() which is never cancelled that seems bad? we should probably put some kind of timeout on this

cmd/atelet/sandbox_prewarm.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+}
+
+// prewarm fetches every asset of one SandboxConfig into the static-files
+// cache and its pause image into the image cache. Racing an on-demand
+// ensureSandboxAssets for the same assets is safe: both paths install
+// content-addressed files via atomic rename, and the image cache collapses
+// concurrent pulls of one digest.
+func (p *sandboxPrewarmer) prewarm(ctx context.Context, cfg *v1alpha1.SandboxConfig) error {
+	rec, err := recordFromSandboxConfig(cfg)
```
BenTheElder: if this fails, we still fail to warm pause, which doesn't need the arch => config lookup

cmd/atelet/sandbox_prewarm.go
```diff
@@ -0,0 +1,201 @@
...
+
+	select {
+	case p.queue <- cfg:
+	default:
+		// Best-effort: dropping an event only costs a download at first use.
+		slog.WarnContext(ctx, "Sandbox asset prewarm queue full; skipping config", slog.String("config", cfg.Name))
+	}
+}
+
+func (p *sandboxPrewarmer) run(ctx context.Context) {
```
BenTheElder, line 183: we wind up doing this in a single worker with no concurrency or timeouts, so one failed config will hang the rest

imagine one config has a bad pause image reference, we'll never be able to pull it. and it will hang forever

cmd/atelet/sandbox_prewarm.go
```diff
@@ -0,0 +1,201 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case cfg := <-p.queue:
+			if prewarmMaxJitter > 0 {
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(rand.N(prewarmMaxJitter)):
```
BenTheElder: we don't dedupe configs, so a relist can cause this to stack additively?

cmd/atelet/sandbox_prewarm.go
```diff
@@ -0,0 +1,201 @@
...
+// without jitter they would all open the same bucket objects at once. A var so
+// tests can zero it.
+var prewarmMaxJitter = 30 * time.Second
+
+// sandboxPrewarmer downloads SandboxConfig assets into the node's
+// content-addressed static-files cache before any actor asks for them, so the
+// fetch inside the first Run/Restore on the node is a cache hit instead of a
+// download+extract on the critical path.
+type sandboxPrewarmer struct {
+	herder *AteomHerder
```
BenTheElder: this seems like kind of a weird touchpoint. we don't need most of it, right?

cmd/atelet/main.go
dberkov: Fixed in 3d7654c9: the SandboxConfigs informer is now requested only after the factory's blocking WaitForCacheSync, and started with a second factory Start call (which launches only informers added since the first). A Forbidden list/watch during RBAC skew now just leaves prewarm cold while the reflector retries in the background — atelet startup no longer waits on it. A watch error handler logs the degradation explicitly.

cmd/atelet/sandbox_prewarm.go
dberkov: Fixed in 224f35a2: the buffered channel is replaced with a rate-limited workqueue keyed by config name, so relists and event bursts coalesce into one entry. The worker resolves the name through the informer's lister at processing time, so it always prewarms the latest revision. This also brought retry-with-backoff (1s–5min, 8 attempts) for free, closing the retry TODO.

cmd/atelet/sandbox_prewarm.go
dberkov: Fixed in 8ec2d95d (plus the workqueue in 224f35a2): each attempt now runs under a 5-minute timeout, so a config with an unpullable pause image costs the worker at most 5 minutes before it requeues with backoff and moves on. Other configs process between its retries, and after 8 failed attempts it's dropped until the next config event. Kept the single worker deliberately so prewarms don't compete for node bandwidth — with the timeout it's no longer a liveness risk.

cmd/atelet/sandbox_prewarm.go
dberkov: Fixed in 8ec2d95d: prewarm now derives a 5-minute timeout context per attempt, so a download that stalls without failing can't wedge the worker. A timed-out attempt goes through the same requeue-with-backoff path as any failure. Added a test with a hanging bucket read verifying the attempt returns.

cmd/atelet/sandbox_prewarm.go
dberkov: Fixed in 8ec2d95d: the pause image pull no longer depends on recordFromSandboxConfig — it's scheduled first and runs even when the config has no assets for the node's architecture. The projection error is still returned so the failure stays visible. Test added for the no-assets-for-arch case.

cmd/atelet/sandbox_prewarm.go
dberkov: Fixed in f7e3392b: the prewarmer now takes a one-method sandboxAssetFetcher interface (still implemented by AteomHerder, so prewarm and the on-demand fetch can't diverge on cache layout) plus the image cache store directly, instead of the whole herder.

cmd/atelet/deviceplugin.go
dberkov: Added a TODO on microvmNodeCapable in b4d312bb: once /dev/mshv lands, capability should mean presence of any micro-VM hypervisor device in SandboxDevices, not KVM alone.

cmd/atelet/sandbox_prewarm.go
```diff
@@ -0,0 +1,275 @@
...
+	ctx, cancel := context.WithTimeout(ctx, prewarmTimeout)
+	defer cancel()
+	t := time.Now()
+
+	var imageErr error
+	var wg sync.WaitGroup
+	// schedule prewarm pause image if provided
+	if cfg.Spec.PauseImage != "" {
+		wg.Go(func() {
+			if _, err := p.images.EnsureImage(ctx, cfg.Spec.PauseImage); err != nil {
```
BenTheElder, line 253: 🤖 **blocking** 🔴 – Prewarm can fail a serving Restore.

`EnsureImage` collapses concurrent pulls of one digest, and the winning call's context governs the pull for every waiter (`internal/imagecache/imagecache.go:368`). Until now every caller was an RPC bounded by its own client. A prewarm that wins the slot imposes `prewarmTimeout` on a Restore waiting on the same digest, and its context is the daemon context, which cancels at SIGTERM while the gRPC server still drains for 5 minutes.

Fix in imagecache: run the singleflight body on a context detached from whichever caller won, bounded by its own pull timeout, so waiters only ever see real pull failures. The comment at `imagecache.go:368` states the old assumption and needs updating either way.

cmd/atelet/sandbox_prewarm.go
```diff
@@ -0,0 +1,275 @@
...
+func (p *sandboxPrewarmer) process(ctx context.Context, name string) {
+	defer p.queue.Done(name)
+	cfg, err := p.lister.Get(name)
+	if apierrors.IsNotFound(err) {
+		// Deleted since it was enqueued; nothing to prewarm anymore.
+		p.queue.Forget(name)
+		return
+	}
+	if err == nil {
+		err = p.prewarm(ctx, cfg)
```
BenTheElder, line 216: 🤖 **should-fix** 🟡 – The class gate is enforced against a stale object.

`enqueue` filters on `SandboxClass` but queues a name. `process` re-resolves the current revision and prewarms whatever class it holds by then. `sandboxClass` is mutable, so a gvisor→microvm edit inside the jitter or backoff window makes a node without `/dev/kvm` download the micro-VM guest assets the gate exists to avoid.

Factor the `switch` in `enqueue` into a `skipConfig(ctx, cfg) bool` and call it here too, after the lister `Get`.

cmd/atelet/sandbox_prewarm.go
```diff
@@ -0,0 +1,275 @@
...
+	return nil
+}
+
+// recordFromSandboxConfig projects a SandboxConfig's per-architecture assets
+// onto the local node's architecture, mirroring recordFromRequest.
+func recordFromSandboxConfig(cfg *v1alpha1.SandboxConfig) (*sandboxAssetsRecord, error) {
+	arch := runtime.GOARCH
+	files := cfg.Spec.Assets[arch]
+	if len(files) == 0 {
+		return nil, fmt.Errorf("sandbox config %q has no assets for architecture %q", cfg.Name, arch)
```
BenTheElder: 🤖 **nit** 🟢 – A permanent error travelling the transient-retry path.

A config with no assets for this node's architecture will never grow them without a config update, and an update re-enqueues anyway. `process` cannot tell this apart from a failed download, so every node of a non-listed architecture burns `prewarmMaxRetries` rate-limited attempts with a warning each. Every attempt also redoes the pause-image pull that already succeeded, because `prewarm` joins both errors into one retry decision.

Return a sentinel error here so `process` can log at debug and `Forget` the config immediately.

BenTheElder (commented):
three claude findings, given the time constraints, I think these are valid

cmd/atelet/sandbox_prewarm.go
dberkov: Fixed in 288edfb3: the singleflight body now runs on a context detached from whichever caller starts the flight (`context.WithoutCancel`, so trace/log values survive), bounded by a store-level pull timeout (default 10m, `WithPullTimeout` to override). A waiter only ever sees a real pull failure — never another caller's cancellation, deadline, or the daemon context ending at SIGTERM. `EnsureImage` switched to `DoChan`, so each caller also stops waiting when its own context ends while the pull runs on to warm the cache for the next attempt (the prewarm worker therefore stays bounded by `prewarmTimeout` even when the underlying pull runs longer). The stale comment is rewritten to state the new contract, and tests cover both halves: a cancelled caller doesn't kill the pull for a later joiner, and a wedged detached pull is reclaimed by the pull timeout.

cmd/atelet/sandbox_prewarm.go
dberkov: Fixed in d785c6b2: the switch is factored into `skipConfig`, applied both in `enqueue` and in `process` after the lister `Get`, so the gate holds for the revision actually prewarmed. A test covers the gvisor→microvm edit landing between enqueue and processing on a non-KVM node (skipped and forgotten, not fetched).

cmd/atelet/sandbox_prewarm.go
dberkov: Fixed in 02616d52: `recordFromSandboxConfig` now wraps a sentinel `errNoAssetsForArch`, and prewarm treats it as nothing-to-do — logged at debug, pause image still pulled, config `Forget`ten immediately. One deviation from the suggestion: the sentinel is handled inside `prewarm` before the error join rather than in `process`, because `errors.Is` on the joined error would also match when only the pause pull failed transiently, silently dropping its retry.

BenTheElder (approved):

PR #1364, opened 2026-09-01
(atelet): make the restore timing log a joinable per-actor latency record

krisztianfekete:
Restore timing breakdown is the only unsampled per-actor latency record we emit, and since traces are head-sampled at 1% on the data plane, and actor identity is not available from metric labels on purpose. 

This adds this to logs:

 Before:
```json
 {"msg":"Restore timing breakdown",
  "actor":{"type":"*ateapipb.Actor","atespace":"demo","name":"counter-1"},
  "download":310000000,"oci_unpack":50000000,"ateom_restore":60000000,"total":420000000}
```

 After:
```json
 {"msg":"Restore timing breakdown",
  "ate.atespace":"ate-demo-counter-substrate","ate.actor.name":"join-probe",
  "ate.actor.uid":"3a5b4e86-1a77-4da7-84c4-c0993d9c83ec",
  "ate.template.atespace":"ate-demo-counter-substrate","ate.template.name":"counter",
  "ate.snapshot.scope":"full","ate.snapshot.kind":"golden","ate.sandbox.class":"gvisor",
  "ate.actor.restore.duration.volume_mount":0.000004583,
  "ate.actor.restore.duration.manifest_fetch":0.006111636,
  "ate.actor.restore.duration.sandbox_assets":0.000098958,
  "ate.actor.restore.duration.download":0.025386544,
  "ate.actor.restore.duration.oci_unpack":0.000954919,
  "ate.actor.restore.duration.ateom_restore":0.147448298,
  "ate.actor.restore.duration.total":0.181189689,
  "trace_id":"c9d7ea9a5f90a5eb4399bfa0b966d5ac","span_id":"89280ba761137c8e"}
```

The duration keys are the `ate.actor.restore.duration` instrument's name plus an `ate.snapshot.phase` value, so a log key and the histogram it mirrors are literally the same identifier in source, which is also why they're seconds.

Also adds `ateattr.ActorLogAttrs`, the component-log twin of `ActorLogLabels`, held to the same key set by a test. 

 Verified on kind:
 - the ateom Actor restoring record for the same actor carries the identical `ate.actor.uid` and `trace_id`, this was impossible before
 - `ate_actor_restore_duration_seconds_sum{phase="total"} = 0.181189689`, matching the log to the nanosecond
 - induced a download failure by deleting a snapshot object: record sets `ate.failure.reason:` `FAILED_GET_EXTERNAL_OBJECT` and correctly omitted `ateom_restore`; the metric for the same restore shows the same six phases

- [x] Tests pass

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -949,9 +949,9 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
 	actorRef := resources.ActorRef{Atespace: req.GetAtespace(), Name: req.GetActorName()}
 
 	// Per-step timing so we can attribute resume latency between the rustfs
-	// download/decompress, the OCI image unpack, and ateom's own work. Logged at
-	// the end, and recorded per phase on the way out so a failed restore still
-	// reports the phases it completed. Phases left at zero never ran.
+	// download/decompress, the OCI image unpack, and ateom's own work. Reported on
+	// the way out, so a failed restore still accounts for the phases it completed.
+	// Phases left at zero never ran.
 	tStart := time.Now()
 	var dMount, dManifest, dAssets, dDownload, dBundles, dAteom time.Duration
 	op := snapshotOp{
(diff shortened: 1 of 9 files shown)


cmd/atelet/main.go
```diff
@@ -949,25 +949,37 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+			{ateattr.SnapshotPhaseVolumeMount, dMount},
+			{ateattr.SnapshotPhaseManifestFetch, dManifest},
+			{ateattr.SnapshotPhaseSandboxAssets, dAssets},
+			{ateattr.SnapshotPhaseDownload, dDownload},
+			{ateattr.SnapshotPhaseOCIUnpack, dBundles},
+			{ateattr.SnapshotPhaseAteomRestore, dAteom},
+			{ateattr.SnapshotPhaseTotal, time.Since(tStart)},
+		}
+		s.instruments.recordRestore(ctx, op, err, phases...)
+		slog.LogAttrs(ctx, slog.LevelInfo, "Restore timing breakdown",
```
git286, line 988: Now that the timing log fires on failures too (via the defer), the preexisting collateral cleanup can make it lie.

The prep leg has two phases. If ensureSandboxAssets finishes (say 1.2s) and then the download leg fails, the errgroup cancels prepareOCIBundles mid-flight. isCollateral(err, prepErr) is true, so we zero both dAssets and dBundles, but dAssets was a complete, valid measurement. Since the log drops zero phases and the docs say "absent = never ran", the record now claims sandbox assets never ran when it actually completed, and the histogram drops a valid sample.

cmd/atelet/main.go
```diff
@@ -949,25 +949,37 @@ func (s *AteomHerder) Restore(ctx context.Context, req *ateletpb.RestoreRequest)
...
+			snapshotLogAttrs(attribution, op, restoreDurationMetric, err, phases)...)
```
git286: Since the log moved into a defer, a panic mid-restore now produces a success shaped record: no return runs during unwinding, so the named err is still nil and ate.failure.reason is omitted. A crashed restore will get logged as a fast success.

cmd/atelet/main.go
krisztianfekete: Good catch, this has been a problem even before my PR as well for histograms.

cmd/atelet/main.go
krisztianfekete: Fixed.

git286 (approved):

PR #1365, opened 2026-09-01
ateapi: UpdateActor allows updating ActorTemplate.

### LINKED ISSUE #477: Actor Image Upgrade Design

zoez7:
Fixes https://github.com/agent-substrate/substrate/issues/477 Implements the following actor template update flow:

```
UpdateActor(): set .template = templ-v2

ResumeActor()
    resumes using templ-v2
    wait for readyz
    if fail: return failure
        .template = templ-v2; .status.current_template = templ-v1
        (the next resume will attempt templ-v2 again)
    set .status = STATUS_RUNNING
    set .status.current_template = templ-v2
    return success
    .template = templ-v2; .status.current_template = templ-v2
```


internal/e2e/suites/demo/updatetemplate_test.go
```diff
@@ -158,6 +162,15 @@ func runUpdateTemplateTestCase(t *testing.T, onCommit ateapipb.SnapshotContentSc
 		Metadata:      suspended.GetMetadata(),
 		ActorTemplate: refB,
 	}})
+	// gVisor cannot restore data-only from a FULL snapshot yet, so the API
+	// refuses to repoint an actor whose gVisor templates commit FULL scope.
```
nybidari: This should work now, we need to untar the durable volumes (from the snapshot) first and then just cold start the sandbox.

zoez7: I think that was an older revision of the PR. The newest revision does allow cold start.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -277,11 +278,25 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
-			if _, err := resolveActorTemplate(ctx, s.store, newVal); err != nil {
+			if state := oldVal.GetStatus().GetState(); state != ateapipb.ActorState_ACTOR_STATE_SUSPENDED {
+				return status.Errorf(codes.FailedPrecondition,
+					"actor must be %s to change its actor template (got: %s)", ateapipb.ActorState_ACTOR_STATE_SUSPENDED, state)
+			}
+			newTemplate, err := resolveActorTemplate(ctx, s.store, newVal)
+			if err != nil {
+				return err
+			}
+			oldTemplate, err := resolveActorTemplate(ctx, s.store, oldVal)
```
nybidari, line 295: If oldTemplate cannot be resolved and the actor has no snapshot (oldVal.GetStatus().GetLatestSnapshot() == nil), should we bypass volume comparison?

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -311,6 +326,40 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
 }
 
+// validateTemplateVolumesUnchanged rejects a template repoint that changes
+// the template's volumes or any container's volume mounts: an actor's
+// snapshot data is laid out per the volumes and mount paths it was captured
+// with, so a different layout would restore it to the wrong places. The
+// volumes list must be identical, and containers present in both templates
+// must keep identical mounts, order included; containers added or removed by
+// the new template are unconstrained.
+func validateTemplateVolumesUnchanged(oldTemplate, newTemplate *ateapipb.ActorTemplate) error {
```
nybidari, line 355: We should check sandbox class of old template and new template here and fail if it is not same

cmd/ateapi/internal/controlapi/actor.go
zoez7: Done. Skipped validation when oldTemplate cannot be resolved , because the actor will not be able to Resume anywya.

cmd/ateapi/internal/controlapi/actor.go
zoez7: Done.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -277,6 +278,34 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
 
 		// Do any further work on the resource.
 
+		// Update actor template is only allowed while the actor is suspended.
+		// The repointed ref must also resolve, mirroring CreateActor's
+		// check (same non-atomicity caveat; resume re-resolves and fails
+		// cleanly), and the replacement's sandbox class, volumes, and volume
```
dberkov, line 284: I was discussing the statement (volume must be match)  with @msau42 in July , and as I remember, we  was thinking allowing adding new volumes, but disallow remove volumes.

@msau42 - do you remember what you planned to implement?

zoez7: My understanding is volumes are immutable for MVP, and we'll explore adding new volumes post MVP. 
Let me know if I missed something.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -302,6 +331,54 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
 }
 
+// validateTemplateSandboxClassUnchanged rejects a template repoint that
+// changes the sandbox class: snapshots are not portable across sandbox
+// runtime families, so the actor's saved state could not be restored under
+// the new template.
+func validateTemplateSandboxClassUnchanged(oldTemplate, newTemplate *ateapipb.ActorTemplate) error {
+	oldClass := oldTemplate.GetSandboxConfig().GetSandboxClass()
+	newClass := newTemplate.GetSandboxConfig().GetSandboxClass()
+	if oldClass != newClass {
```
dberkov, line 341: sandboxClass is an object , the compare will check references. You might want to use proto compare.

zoez7: Oh, I didn't check the sandboxConifg, it checks the enum of sandboxClass.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -302,6 +331,54 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
 	return storedActor, nil
 }
 
+// validateTemplateSandboxClassUnchanged rejects a template repoint that
+// changes the sandbox class: snapshots are not portable across sandbox
+// runtime families, so the actor's saved state could not be restored under
```
dberkov, line 336: the comment says, not portable across families, however code check references. Does it mean we allow change the sandbox binary from same family?

I think we supposed to allow reference to different sandbox binary from same family. In any case, code does it not match comment. 

The name "validateTemplateSandboxClassUnchanged" is inaccurate, if the intention to check family only.

zoez7: `validateTemplateSandboxClassUnchanged` validates that they are the same class/family (gVisor or micorVM), but updating to a different SandboxConfig of the same class/family is allowed.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -302,6 +331,54 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
+
+// validateTemplateVolumesUnchanged rejects a template repoint that changes
+// the template's volumes or any container's volume mounts: an actor's
+// snapshot data is laid out per the volumes and mount paths it was captured
+// with, so a different layout would restore it to the wrong places. The
+// volumes list must be identical, and containers present in both templates
+// must keep identical mounts, order included; containers added or removed by
+// the new template are unconstrained.
+func validateTemplateVolumesUnchanged(oldTemplate, newTemplate *ateapipb.ActorTemplate) error {
+	if !slices.EqualFunc(oldTemplate.GetVolumes(), newTemplate.GetVolumes(), func(a, b *ateapipb.Volume) bool {
```
dberkov, line 356: slices.EqualFunc assumes the order must be the same, I think your intention was to check the content and ignore the order.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -714,13 +723,16 @@ func (w *ActorWorkflow) ensureAteletRestored(ctx context.Context, actorRef resou
...
-		// Same wire-scope derivation as the local branch above: the snapshot
-		// restores as DATA_ON_GOLDEN when the golden URI was resolved per the
-		// template's onResume configuration.
-		scope := actorSnapshotContentScopeToAtelet(src.Scope)
-		if !src.GoldenSnapshotURI.IsZero() {
+		var scope ateletpb.SnapshotScope
+		var goldenSnapshotURI string
+		switch {
+		case src.TemplateReplaced:
+			scope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
```
dberkov, line 730: can it be DATA+Golden if template replaced and golden exist?

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -277,6 +278,34 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
 
 		// Do any further work on the resource.
 
+		// Update actor template is only allowed while the actor is suspended.
+		// The repointed ref must also resolve, mirroring CreateActor's
+		// check (same non-atomicity caveat; resume re-resolves and fails
+		// cleanly), and the replacement's sandbox class, volumes, and volume
+		// mounts must match the old template's.
+		if !proto.Equal(oldVal.GetActorTemplate(), newVal.GetActorTemplate()) {
```
dberkov, line 286: **Repointing an actor whose template has no durable-dir volumes silently discards all of its FULL-snapshot state.** UpdateActor performs no snapshot-scope/durable-volume check, and the forced-DATA resume cold-boots a fresh guest with no error or event.

Failure scenario: an actor uses a template with `onCommit: FULL` and zero durable-dir volumes (valid config; such templates exist in `demos/`). It suspends, uploading a FULL snapshot (memory+rootfs, no durable tar). The user repoints it at a volume-compatible template — accepted. On resume, `TemplateReplaced` forces `SNAPSHOT_SCOPE_DATA` (workflow_resume.go:729-730); ateom-gvisor (main.go:945-966) and ateom-microvm (restore.go:132-156) skip the untar (no durable volumes) and plain create+start the containers, ignoring the downloaded memory checkpoint. Resume reports success; the actor's entire persisted state is gone silently.

Note: the earlier in-PR guard (db5ec38e) that refused FULL-scope repoints was removed (3800aabd) without a durable-volumes-required condition replacing it.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -188,6 +191,11 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
 			return nil, nil, src, status.Errorf(codes.DataLoss, "ActorSnapshot %s/%s: %v", ref.GetAtespace(), ref.GetName(), err)
 		}
 		src.Scope = snapshot.GetStatus().GetContentScope()
+		// The snapshot records the template it was captured under; a
+		// different UID on the actor's current template means the actor was
+		// repointed since the capture.
+		snapshotTemplateUID := snapshot.GetStatus().GetActorTemplateUid()
+		src.TemplateReplaced = snapshotTemplateUID != "" && snapshotTemplateUID != actorTemplate.GetMetadata().GetUid()
```
dberkov, line 198: **TOCTOU: a repoint racing an in-flight resume is silently absorbed** (actual code path: `assignWorkerAttempt`'s version-conflict retry around workflow_resume.go:346 and 551-567 — outside this diff, so anchoring the comment here where `TemplateReplaced` is computed). The retry refreshes only the actor, not the already-resolved template/snapshot source, and the next suspend then stamps the NEW template UID onto OLD-template guest content, permanently disabling `TemplateReplaced` detection.

Failure scenario: actor SUSPENDED under template A; `ResumeActor` takes the lease and `loadActorForResume` resolves template A and `src` (`TemplateReplaced=false`). `UpdateActor` repoints A→B (the gate at actor.go:287 sees SUSPENDED and commits — it takes no lease). `assignWorkerAttempt`'s UpdateActor hits `ErrVersionConflict`, re-reads (workflow_resume.go:551-567), sees still-SUSPENDED, and retries with the stale `actorTemplate=A` and `src`: atelet restores A's FULL snapshot un-forced while the spec ref says B. The next suspend stamps the new snapshot's `ActorTemplateUid` from the spec-resolved template B (workflow_suspend.go:386), not `Status.CurrentActorTemplateUid` (A), so snapshot-UID==template-UID forever after and the actor runs A's guest state under B's identity indefinitely.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -691,16 +699,17 @@ func (w *ActorWorkflow) ensureAteletRestored(ctx context.Context, actorRef resou
...
-		// pause snapshot restores as DATA_ON_GOLDEN — atelet combines the
-		// golden snapshot's guest state with the actor's data. Otherwise the
-		// scope mirrors what the pause captured.
-		req.Scope = actorSnapshotContentScopeToAtelet(actorTemplate.GetSnapshotsConfig().GetOnPause())
-		if !src.GoldenSnapshotURI.IsZero() {
+		// The wire scope describes the restore OPERATION: DATA_ON_GOLDEN when
+		// loadActorForResume resolved a golden URI per the template's onResume
+		// configuration, else what the pause captured.
+		switch {
+		case src.TemplateReplaced:
```
dberkov, line 706: **The `TemplateReplaced` arm in the LOCAL-checkpoint branch fires spuriously off the stale durable `LatestSnapshot`** and downgrades a resume-from-paused to `SNAPSHOT_SCOPE_DATA`, discarding the paused guest's memory. A local checkpoint is always captured under the current template, so this arm is never correct in this branch.

Failure scenario: suspend under template A (`LatestSnapshot` records A's UID; written only at workflow_suspend.go:400, never cleared by pause or resume) → repoint to B → resume (correct DATA restore) → pause (local FULL checkpoint captured under B, `LocalSnapshotInfo` set, `LatestSnapshot` untouched) → resume: `loadActorForResume` computes `TemplateReplaced = A-UID != B-UID = true` (lines 197-198), the local branch is taken (`LocalSnapshotInfo != nil`, line 682), and the switch sends wire scope DATA for a FULL local checkpoint — atelet either rejects the restore or cold-boots, silently losing the paused session, on every pause/resume cycle until the next suspend rewrites `LatestSnapshot`.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -277,6 +278,34 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
+			}
+			oldTemplate, err := resolveActorTemplate(ctx, s.store, oldVal)
+			if err == nil {
+				if err := validateTemplateSandboxClassUnchanged(oldTemplate, newTemplate); err != nil {
+					return err
+				}
+				if err := validateTemplateVolumesUnchanged(oldTemplate, newTemplate); err != nil {
+					return err
+				}
+			} else if !errors.Is(err, errActorTemplateNotFound) {
```
dberkov, line 303: **Deleting the old template waives ALL repoint compatibility checks** (sandbox class, volumes, mounts) — and `DeleteActorTemplate` has no in-use guard — so a suspended actor can be repointed across sandbox families or volume layouts with zero validation. The snapshot's durably-recorded template (`actor_template_uid`/ObjectRef, and the GCS manifest's `SandboxClass`) could serve as a fallback anchor.

Failure scenario: an operator deletes gVisor template A (unconditional DELETE, atepg.go:598-603; `ServiceImpl.DeleteActorTemplate` is a TODO passthrough) while actor X is suspended on it, then repoints X to microvm template B with different volumes: `resolveActorTemplate(oldVal)` returns `errActorTemplateNotFound` and both validators are skipped. Nothing re-checks downstream (atelet only cross-checks golden-vs-actor manifest class), so the forced-DATA resume extracts A's data into B's mismatched layout under a different sandbox family — failing at restore time or silently misplacing data.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -277,6 +278,34 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
+		if !proto.Equal(oldVal.GetActorTemplate(), newVal.GetActorTemplate()) {
+			if state := oldVal.GetStatus().GetState(); state != ateapipb.ActorState_ACTOR_STATE_SUSPENDED {
+				return status.Errorf(codes.FailedPrecondition,
+					"actor must be %s to change its actor template (got: %s)", ateapipb.ActorState_ACTOR_STATE_SUSPENDED, state)
+			}
+			newTemplate, err := resolveActorTemplate(ctx, s.store, newVal)
+			if err != nil {
+				return err
+			}
+			oldTemplate, err := resolveActorTemplate(ctx, s.store, oldVal)
```
dberkov, line 295: **Chained repoints bypass the mount-layout invariant.** Validation anchors on the old spec ref's template instead of the template the snapshot was captured under, and the per-container mount check skips containers absent from either side — so compatibility is non-transitive.

Failure scenario: snapshot captured under A (container `main` mounts volume `data` at `/data`). While still SUSPENDED: repoint A→B where B drops `main` (passes — removed containers unconstrained, actor.go:368-371 `if !ok { continue }`); repoint B→C where C re-adds `main` mounting `data` at `/other` (passes — `main` absent from B). On resume the forced-DATA restore surfaces the durable volume at `/other` instead of `/data` — exactly the "restore to the wrong places" condition the validator's own comment says it prevents — while a direct A→C repoint would have been rejected.

Anchoring validation on `snapshot.status.actor_template_uid` (which this PR's own resume mechanism uses) closes both this and the deleted-template hole.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -795,6 +808,13 @@ func (w *ActorWorkflow) finalizeRunning(ctx context.Context, actorRef resources.
 
 	storedActor, err := w.store.UpdateActor(ctx, actorRef, store.PreconditionFrom(latestActor), func(toUpdate *ateapipb.Actor) error {
 		toUpdate.Status.State = ateapipb.ActorState_ACTOR_STATE_RUNNING
```
dberkov, line 810: **`finalizeRunning` blindly stamps `State=RUNNING` after a fresh re-read**, so a lease-free CRASHED write landing between restore success and finalize is silently overwritten, producing a RUNNING actor with a nil worker assignment.

Failure scenario: restore succeeds; the worker pod dies and `DeleteWorker`'s `ensureBoundActorReleased` (workflow_worker_delete.go:84-143, takes no actor lease, explicitly handles RESUMING) sets `State=CRASHED` and clears `WorkerAssignment`. `finalizeRunning` then GetActor's the post-crash version, so `PreconditionFrom(latestActor)` matches, and it writes RUNNING (now also stamping `CurrentActorTemplate*`) without checking the current state or restoring the assignment — a zombie RUNNING actor with no worker; the store has no CRASHED→RUNNING transition guard.

Pre-existing, but this PR touches the function, and the delete-worker code's comment assumes the optimistic check protects it.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -714,13 +723,16 @@ func (w *ActorWorkflow) ensureAteletRestored(ctx context.Context, actorRef resou
...
-
-		// Same wire-scope derivation as the local branch above: the snapshot
-		// restores as DATA_ON_GOLDEN when the golden URI was resolved per the
-		// template's onResume configuration.
-		scope := actorSnapshotContentScopeToAtelet(src.Scope)
-		if !src.GoldenSnapshotURI.IsZero() {
+		var scope ateletpb.SnapshotScope
+		var goldenSnapshotURI string
+		switch {
+		case src.TemplateReplaced:
```
dberkov, line 729: **When the new template has `onResume.fromData: Golden`, the `TemplateReplaced` arm outranks the golden arm and drops the NEW template's own golden URI**, sending plain `SNAPSHOT_SCOPE_DATA` — a cold image boot — even though `DATA_ON_GOLDEN` with B's golden is exactly what the policy prescribes and is fully compatible.

Failure scenario: actor holds a durable DATA snapshot recorded under old template A; it is repointed to template B (`fromData=Golden`, golden ready). `loadActorForResume` sets both `TemplateReplaced=true` (line 198) and `GoldenSnapshotURI=B's golden` (lines 224-250 — resolved from the CURRENT template with no `TemplateReplaced` check, so the flags do co-occur). `ensureAteletRestored` emits scope=DATA with empty `GoldenSnapshotUri` in both branches; ateom-microvm's DATA path calls `coldBootActorRetrying` (restore.go:148-156), so every post-repoint resume pays cold-boot latency and loses the golden's warmed guest state — defeating the template class whose design is to never cold-boot — until a new snapshot is taken under B.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -714,13 +723,16 @@ func (w *ActorWorkflow) ensureAteletRestored(ctx context.Context, actorRef resou
...
 			tele.SnapshotKind = ateattr.SnapshotKindLatest
 		}
-
-		// Same wire-scope derivation as the local branch above: the snapshot
-		// restores as DATA_ON_GOLDEN when the golden URI was resolved per the
-		// template's onResume configuration.
-		scope := actorSnapshotContentScopeToAtelet(src.Scope)
-		if !src.GoldenSnapshotURI.IsZero() {
+		var scope ateletpb.SnapshotScope
+		var goldenSnapshotURI string
```
dberkov, line 727: **The scope/golden-URI precedence ladder (`TemplateReplaced` > golden > captured scope) is copy-pasted into both restore branches** of `ensureAteletRestored` with subtly different defaults and shapes, instead of being computed once on `resumeSnapshotSource` in `loadActorForResume` where all three inputs are resolved.

Cost: the next scope-affecting rule (e.g. letting a repointed DATA snapshot still ride the new template's golden, per the adjacent finding) must be inserted at the right precedence position in two switches whose default arms already differ (onPause config vs `src.Scope`) and whose outputs land in different places (req fields vs locals); getting one branch wrong ships divergent wire scopes for local vs durable restores of the same actor. A single helper, e.g. `src.wireScope(fallback) (scope, goldenURI)`, eliminates the bug class.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -277,6 +278,34 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
+			if err == nil {
+				if err := validateTemplateSandboxClassUnchanged(oldTemplate, newTemplate); err != nil {
+					return err
+				}
+				if err := validateTemplateVolumesUnchanged(oldTemplate, newTemplate); err != nil {
+					return err
+				}
+			} else if !errors.Is(err, errActorTemplateNotFound) {
+				// Skip the validation if old template is not found
```
dberkov, line 304: **This comment is attached to the opposite branch it describes.** "Skip the validation if old template is not found" sits inside `else if !errors.Is(err, errActorTemplateNotFound)` directly above `return err` — the path taken when the old template WAS found but failed to resolve; the skip is the invisible fall-through.

A reader will believe `return err` implements the skip, or misread `resolveActorTemplate`'s not-found semantics. Restructuring as `if err != nil && !errors.Is(err, errActorTemplateNotFound) { return err }` with the skip comment on the not-found case is self-documenting.

cmd/ateapi/internal/controlapi/actor_test.go
```diff
@@ -867,6 +867,288 @@ func TestUpdateActor(t *testing.T) {
...
+	}, {
+		name:    "mountless container renamed",
+		oldTmpl: template(oneVolume, container("main", dataMount), container("sidecar")),
+		newTmpl: template(oneVolume, container("main", dataMount), container("helper")),
+	}}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validateTemplateVolumesUnchanged(tt.oldTmpl, tt.newTmpl)
+			if gotErr := err != nil; gotErr != tt.wantErr {
+				t.Fatalf("validateVolumeMountsUnchanged() error = %v, wantErr %v", err, tt.wantErr)
```
dberkov, line 1141: **Test failure message names a nonexistent function**: `t.Fatalf("validateVolumeMountsUnchanged() error = ...")` while the function under test is `validateTemplateVolumesUnchanged` — residue of a pre-rename identifier. A failing run points engineers at a function that doesn't exist.

PR #1369, opened 2026-09-01
fix(ateom-gvisor): declare seccompProfile Unconfined so the worker survives RuntimeDefault

### LINKED ISSUE #1368: ateom gvisor: worker breaks on clusters that default seccompProfile to RuntimeDefault
## Expected Behavior

The gVisor worker should start its `runsc` sandbox on any cluster, including clusters whose admission layer defaults `pod.spec.securityContext.seccompProfile` to `RuntimeDefault`.

## Actual Behavior

On a cluster that force-defaults `RuntimeDefault` (for example via a MutatingAdmissionPolicy), `runsc create` fails: the sandbox child process exits during startup and the parent reports

```
cannot create sandbox: cannot read client sync file: waiting for sandbox to start: EOF
```

The `runsc` gofer/boot child processes call `pivot_root(2)` while setting up the sandbox root filesystem (`runsc/cmd/sandboxsetup/fs.go`, invoked from `gofer_mount.go`). `runsc` spawns these as child processes (`exec.Command` in `runsc/sandbox/sandbox.go`), and seccomp filters are inherited across fork/exec, so the children run under the pod's `RuntimeDefault` profile. `pivot_root` is not in the `RuntimeDefault` allowlist at all, so it is denied regardless of the worker's capabilities, the child dies, and the sandbox never starts.

The microVM worker already declares `seccompProfile: Unconfined` for exactly this reason; `workerpool_apply.go` notes `pivot_root` is "a syscall the profile denies whatever capabilities the worker holds." The gVisor branch leaves the profile unset and assumes the runtime default is sufficient. That assumption only holds on clusters that don't enforce a default seccomp profile; on one that does, the gVisor worker breaks the same way the microVM worker would without its relaxation. This only relaxes the filter on gVisor's own control processes (gofer/boot); the actor workload inside the sandbox is still confined by the sentry's own seccomp.

## Specifications

- Version: 39df566d
- Platform: gVisor (systrap)

yelshall:
Fixes #1368

The gVisor worker leaves `seccompProfile` unset and relies on the runtime default. `runsc`'s gofer/boot child processes `pivot_root()` while setting up the sandbox rootfs, and the default seccomp profile denies `pivot_root` whatever capabilities the worker holds — the same reason the microVM worker already declares `Unconfined`. On a cluster that defaults pod seccomp to `RuntimeDefault` (e.g. via an admission policy), the sandbox child dies during startup and `runsc create` fails with `cannot read client sync file: waiting for sandbox to start: EOF`.

This declares `Unconfined` explicitly for the gVisor worker too, mirroring the microVM branch. It relaxes the filter only on gVisor's own control processes; the actor workload inside the sandbox remains confined by the sentry's own seccomp.

> This change was prepared with AI assistance; I have reviewed and tested it.

- [x] Tests pass

### DIFF
--- a/cmd/atecontroller/internal/controllers/workerpool_apply.go
+++ b/cmd/atecontroller/internal/controllers/workerpool_apply.go
@@ -312,10 +312,10 @@ func ateomSecurityContext(class atev1alpha1.SandboxClass) *corev1ac.SecurityCont
 				WithType(corev1.SeccompProfileTypeUnconfined))
 	}
 	// runsc's sandbox child also pivot_root()s, so it needs the same relaxation:
-	// a cluster that defaults seccomp to RuntimeDefault (an increasingly common
-	// baseline) denies the syscall whatever capabilities the worker holds, and the
-	// sandbox then dies during startup. Declare Unconfined explicitly so the
-	// worker does not depend on the cluster leaving the profile unset.
+	// a cluster that defaults seccomp to RuntimeDefault denies the syscall whatever
+	// capabilities the worker holds, and the sandbox then dies during startup.
+	// Declare Unconfined explicitly so the worker does not depend on the cluster
+	// leaving the profile unset.
 	return sc.
 		WithCapabilities(corev1ac.Capabilities().
 			WithDrop("ALL").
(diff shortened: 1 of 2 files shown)


BenTheElder (approved):
Thanks!

PR #1371, opened 2026-09-01
Run the dev NFS server on a Kubernetes-published image

hajiler:
Currently CSI NFS setup deploys a  copy of the upstream csi-driver-nfs sample, which runs an unsafe personal image with privileged permissions.

This PR replaces it with registry.k8s.io/e2e-test-images/volume/nfs, the NFS server the Kubernetes project builds from test/images/volume/nfs for its own e2e suite. It is signed, published to the same registry as the rest of the CSI bundle, and built for both amd64 and arm64.

- [ X] Tests pass
- [  Appropriate changes to documentation are included in the PR


hack/csi/nfs-server.yaml
```diff
@@ -0,0 +1,68 @@
+# Copyright 2026 Google LLC
```
msau42, line 1: I would still put this under 3rd party

hajiler: DOne.

msau42 (approved):

PR #1374, opened 2026-09-01
ate-setup: mirror the versioned substrate install of the shell installer

HavenXia:
Same logic in #1357 but ported to the `ate-setup` go tool.

- [ ] Tests pass

### DIFF
--- a/cmd/ate-setup/internal/demos/simple.go
+++ b/cmd/ate-setup/internal/demos/simple.go
@@ -48,9 +48,17 @@ var ExternalVolumePlaceholders = []string{
 	"EXTERNAL_VOLUMES",
 }
 
-// Render expands a demo template with the configured bucket name.
+// Render expands a demo template with the configured bucket name and the
+// build version (pool templates pin worker pods to version-labeled nodes).
 func Render(e *steps.Env, relPath string, extraValues map[string]string, drop []string) ([]byte, error) {
-	values := map[string]string{bucketNamePlaceholder: e.Cfg.BucketName}
+	version, _, err := e.SubstrateVersion()
+	if err != nil {
+		return nil, err
+	}
+	values := map[string]string{
+		bucketNamePlaceholder: e.Cfg.BucketName,
+		"SUBSTRATE_VERSION":   version,
+	}
 	for k, v := range extraValues {
 		values[k] = v
 	}
--- a/cmd/ate-setup/internal/steps/version.go
+++ b/cmd/ate-setup/internal/steps/version.go
@@ -1,0 +1,133 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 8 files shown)


cmd/ate-setup/internal/demos/simple.go
```diff
@@ -48,9 +48,17 @@ var ExternalVolumePlaceholders = []string{
...
+// build version (pool templates pin worker pods to version-labeled nodes).
 func Render(e *steps.Env, relPath string, extraValues map[string]string, drop []string) ([]byte, error) {
-	values := map[string]string{bucketNamePlaceholder: e.Cfg.BucketName}
+	version, _, err := e.SubstrateVersion()
+	if err != nil {
+		return nil, err
+	}
+	values := map[string]string{
+		bucketNamePlaceholder: e.Cfg.BucketName,
+		"SUBSTRATE_VERSION":   version,
```
zoez7, line 60: Just so I understand, we are not using this yet, and are transitioning to the go-based installer?

cmd/ate-setup/internal/steps/version.go
```diff
@@ -0,0 +1,133 @@
...
+// that does not carry it yet. Nodes that already carry the label keep their
+// value: during a rolling upgrade the operator owns the per-node value, and
+// an install must not yank nodes across versions behind its back.
+func (e *Env) LabelNodesSubstrateVersion(ctx context.Context) error {
+	version, _, err := e.SubstrateVersion()
+	if err != nil {
+		return err
+	}
+	log.Stepf("label_nodes_substrate_version (%s)", version)
+	nodes, err := e.Kube.Typed.CoreV1().Nodes().List(ctx, metav1.ListOptions{
```
zoez7, line 99: I was using https://github.com/agent-substrate/substrate/pull/1357/ as a reference, do we need to also update the documentation in https://screenshot-v2.corp.google.com/19vo1as0320o8 since the new nodes are labeled?

zoez7 (approved):

cmd/ate-setup/internal/steps/version.go
HavenXia: Not needed — what the screenshot shows are **GKE node pools**. 
The installation labels nodes, on every environment (non-GKE and GKE alike). 

We don't touch GKE node-pool labels for users yet: nodes GKE adds later are born from the pool's labels, so today the pool stamp is a documented manual step (in your screenshot).

I am thinking about make that in the `setup-gcp` tools, so after installation, if autoscaling happens, users will get new nodes with correct version label.

cmd/ate-setup/internal/demos/simple.go
HavenXia: Yes — the value is unused until the demo worker pool templates gain a ${SUBSTRATE_VERSION} nodeSelector placeholder, which comes in the follow-up demos PR.

PR #1377, opened 2026-09-01
atepg(cleanup): update the in-place metadata setter migration

shrutiyam-glitch:
Storage-layer cleanup, follow-up to the DV migration PRs (#1250, #1244, #1303, #1261). No behavior changes visible to API clients.

- Retire `newCreateMetadata` / `newUpdateMetadata`. Both carried TODO: EOL this in favor of `set{Create,Update}Metadata `from the Worker PR's review. The Worker paths were migrated there; this switches the remaining callers — atespace, egress policy, snapshot, and snapshot-tag creates/updates — to the in-place setters (with nil-metadata guards) and deletes the two helpers.

- Drop the clone in `CreateActor`/`CreateActorTemplate`/`CreateWorker`. The only caller (`ServiceImpl.Create<resource>`) builds a dedicated object for the call and never reads it afterward, so the deep copy on every actor creation bought nothing. 

- [ ] Tests pass


cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -271,8 +250,14 @@ func pgErrConstraint(err error) string {
...
 
-	dbAtespace := proto.Clone(atespace).(*ateapipb.Atespace)
-	dbAtespace.Metadata = newCreateMetadata("", name)
+	// The atespace is mutated in place: callers pass a dedicated object.
+	dbAtespace := atespace
+	if dbAtespace.Metadata == nil {
+		dbAtespace.Metadata = &ateapipb.ResourceMetadata{}
+	}
+	// Atespaces are global-scoped, so the atespace is always empty.
+	dbAtespace.Metadata.Atespace = ""
```
thockin: This should have been validated already, right?  If this is set and we get this far, there's a problem.  Doing this here implies that this is acceptable, and I think it's impossible?

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -271,8 +250,14 @@ func pgErrConstraint(err error) string {
 func (p *Persistence) CreateAtespace(ctx context.Context, atespace *ateapipb.Atespace) (*ateapipb.Atespace, error) {
 	name := atespace.GetMetadata().GetName()
 
-	dbAtespace := proto.Clone(atespace).(*ateapipb.Atespace)
-	dbAtespace.Metadata = newCreateMetadata("", name)
+	// The atespace is mutated in place: callers pass a dedicated object.
+	dbAtespace := atespace
+	if dbAtespace.Metadata == nil {
```
thockin: Under what conditions would we end up here with nil metadata?  If that is the case, we don't know the name of the atespace to create, we certainly can't proceed.  This should have been validated already?

I see similar patterns elsewhere -- we should make them all consistent.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -271,8 +250,14 @@ func pgErrConstraint(err error) string {
 func (p *Persistence) CreateAtespace(ctx context.Context, atespace *ateapipb.Atespace) (*ateapipb.Atespace, error) {
 	name := atespace.GetMetadata().GetName()
 
-	dbAtespace := proto.Clone(atespace).(*ateapipb.Atespace)
-	dbAtespace.Metadata = newCreateMetadata("", name)
+	// The atespace is mutated in place: callers pass a dedicated object.
+	dbAtespace := atespace
```
thockin: Why not just get rid of `dbAtespace` and use `atespace`?

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -588,11 +574,12 @@ func (p *Persistence) CreateActor(ctx context.Context, actor *ateapipb.Actor) (*
...
 
-	// TODO: doing a full clone here is wasteful - the caller already has to
-	// make modifications to the actor before passing it in, so we can safely
-	// mutate it in place.  This breaks some of the contract tests, so we can
-	// fix it later.
-	dbActor := proto.Clone(actor).(*ateapipb.Actor)
+	// The actor is mutated in place: the caller already builds a dedicated
+	// object to pass in, so a defensive clone is wasted work.
+	dbActor := actor
+	if dbActor.Metadata == nil {
```
thockin: Same as elsewhere - we really CAN'T be in here without metadata because we don't have a name!

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -383,7 +368,8 @@ func (p *Persistence) DeleteAtespace(ctx context.Context, name string) (*ateapip
 
 func (p *Persistence) CreateActorTemplate(ctx context.Context, template *ateapipb.ActorTemplate) (*ateapipb.ActorTemplate, error) {
 	atespace, name := template.GetMetadata().GetAtespace(), template.GetMetadata().GetName()
-	dbTemplate := proto.Clone(template).(*ateapipb.ActorTemplate)
+	// The template is mutated in place: callers pass a dedicated object.
+	dbTemplate := template
 	if dbTemplate.Metadata == nil {
```
thockin: we really CAN'T be in here without metadata because we don't have a name!

thockin (commented):
I appreciate the cleanup, but I think this is being "too nice" about bad input.

cmd/ateapi/internal/store/atepg/atepg.go
shrutiyam-glitch: makes sense. thanks !

cmd/ateapi/internal/store/atepg/atepg.go
shrutiyam-glitch: yeah, this is already validated.

shrutiyam-glitch (commented):
Thanks for the review ! 
sorry about the delay :( 
lost the notification in the mails over the weekend.

Issue #817, opened 2026-08-10
Suspend-from-PAUSED: support gVisor Full-capture → Data-commit conversion (blocked on #790)


dberkov:
Tracks the gVisor half of the scope-conversion gate introduced in #816 (part of #791).

## Current behavior

When a PAUSED actor is suspended and the pause captured **Full** but the template commits **Data**, atelet must narrow the upload to the durable data (`narrowFullCaptureToData` in `cmd/atelet/main.go`):

- **micro-VM**: implemented — the durable data is a single self-contained `durable-dir.tar` (`ateompath.DurableDirTarFile`), carved out of the full file set and uploaded alone.
- **gVisor**: returns `codes.Unimplemented` ("gVisor cannot extract durable data from a full checkpoint yet") — a full gVisor checkpoint is a monolithic sandbox image, and durable data cannot be separated from it today. The `TODO(#790)` marks the branch.

Control-plane effect: a gVisor actor paused with `onPause: Full` under a template with `onCommit: Data` cannot be suspended from PAUSED — the upload fails `Unimplemented` and the actor stays SUSPENDING under client retries (it remains resumable). All other combinations (Full→Full, Data→Data) work on gVisor.

## Unblocking work

#790 (split-checkpoint runsc / onResume.fromData phase 2) introduces gVisor checkpoints whose filesystem/durable state is separable from the sandbox image. Once a runsc release with split checkpoints ships:

1. Implement the `SandboxClassGvisor` branch of `narrowFullCaptureToData` (route it through the durable-data artifact the split checkpoint produces, analogous to the micro-VM durable tar).
2. Flip the e2e case: `TestDurableDirLifecycle/"onCommit:Data, onPause:Full, suspend from PAUSED"` is `microVMOnly` today — drop the restriction so it covers gVisor.
3. Re-check the pause side: `onPause: Full` + gVisor + `onCommit: Data` could then also be validated at template level if desired.

## Relationship to #791

#791 (suspend a PAUSED actor) is functionally complete for micro-VM and for scope-matched gVisor suspends via #810/#812/#813/#816, but should stay open until this gap is closed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

HavenXia:
Fixed in #1379

PR #1379, opened 2026-09-01
Support durable directories and combined restore for gVisor

### LINKED ISSUE #451: Support resuming actors with golden memory snapshot and DurableDirectory

nybidari:
- Use direct host bind mounts for durable directories in gVisor containers instead of internal overlay filestores.
- Move tarutil into internal/tarutil so it can be shared across ateom packages.
- Archive durable directory bind mounts into durable-dir.tar on full and data checkpoints (replacing runsc fscheckpoint).
- On restore, unpack durable-dir.tar into the host bind mount before starting/restoring containers.
- Support SNAPSHOT_SCOPE_DATA_ON_GOLDEN in ateom-gvisor, combining golden guest state with latest actor durable data.
- Allow atelet to narrow full gVisor captures to durable-dir.tar for data snapshots.
- Remove CEL validation disallowing onResume.fromData: Golden for gVisor, and enable combined restore demo tests for gVisor.

Fixes #451


manifests/ate-install/generated/ate.dev_actortemplates.yaml
zoez7, line 1: This is being deleted in https://github.com/agent-substrate/substrate/pull/1376, please use the substrate proto ActorTemplate instead. https://github.com/agent-substrate/substrate/blob/07ebc56797f912d3e17cf869b5db7e103363b8e7/pkg/proto/ateapipb/ateapi.proto#L707

nybidari: Thanks, updated the changes.

cmd/ateom-gvisor/runsc.go
```diff
@@ -224,6 +225,66 @@ func (r *runsc) cmdFsCheckpoint(ctx context.Context, containerName, checkpointPa
...
+// argument construction can be unit-tested without executing runsc.
+func (r *runsc) pauseArgs(containerName string) []string {
+	return []string{
+		"-log-format", "json",
+		"--alsologtostderr",
+		// "-debug",
+		// "-debug-log", ateompath.RunscDebugLogDir(r.actorUID, containerName) + "/",
+		// "-debug-to-user-log",
+		// "-log-packets",
+		// "-strace",
```
BenTheElder: nit: we should write these to the debug mode for the ateom, that can be a follow-up but generally prefer to do that over leave commented lines :-)

nybidari: Removed the commented debug flags now, will have a follow up PR to wire these flags to ateom's debug mode.

cmd/ateom-gvisor/main.go
```diff
@@ -750,23 +754,29 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
...
+		tarErr := tarDurableVolumes(ctx, ateompath.DurableDirVolumeMountsDir(req.GetActorUid()), checkpointPath)
+		if err := rcmd.cmdResume(ctx, "pause"); err != nil {
+			return nil, fmt.Errorf("while resuming pause container: %w", err)
+		}
+		if tarErr != nil {
+			return nil, fmt.Errorf("while archiving durable-dir volumes: %w", tarErr)
 		}
 	case ateompb.SnapshotScope_SNAPSHOT_SCOPE_FULL:
 		// Checkpoint pause container (root of the sandbox)
 		if err := rcmd.cmdCheckpoint(ctx, "pause", checkpointPath); err != nil {
```
BenTheElder, line 777: Should this be pause -> tar -> checkpoint instead of checkpoint -> tar? Or IOW is this the right failure handling for just resuming if the disk is transiently full or something?

cmd/ateom-gvisor/main.go
```diff
@@ -750,23 +754,29 @@ func (s *AteomService) CheckpointWorkload(ctx context.Context, req *ateompb.Chec
...
 		}
-		if err := rcmd.cmdFsCheckpoint(ctx, "pause", checkpointPath, ddv); err != nil {
-			return nil, fmt.Errorf("while fscheckpointing durable-dir %q: %w", ddv[0], err)
+		if err := rcmd.cmdPause(ctx, "pause"); err != nil {
+			return nil, fmt.Errorf("while pausing pause container: %w", err)
+		}
+		tarErr := tarDurableVolumes(ctx, ateompath.DurableDirVolumeMountsDir(req.GetActorUid()), checkpointPath)
+		if err := rcmd.cmdResume(ctx, "pause"); err != nil {
+			return nil, fmt.Errorf("while resuming pause container: %w", err)
+		}
```
BenTheElder, line 770: claude surfaced:

```suggestion
		tarErr := tarDurableVolumes(ctx, ateompath.DurableDirVolumeMountsDir(req.GetActorUid()), checkpointPath)
		// Undoing our own pause must not depend on the caller's context:
		// tarutil does not check ctx, so a deadline expiring mid-tar would
		// fail the resume instantly and leave the sandbox paused forever.
		resumeCtx, cancelResume := context.WithTimeout(context.WithoutCancel(ctx), resumeTimeout)
		defer cancelResume()
		if err := rcmd.cmdResume(resumeCtx, "pause"); err != nil {
			return nil, fmt.Errorf("while resuming pause container: %w", err)
		}
```

I think that's right, or some variation on it, the alternative is to make tarutil context aware, but that seems like a bigger lift for this PR.

nybidari: Thanks, updated the changes.

cmd/ateom-gvisor/main.go
nybidari: Left a TODO

zoez7:
Needs a rebase. The test failure was fixed on main.

zoez7:
Added e2e test in https://github.com/agent-substrate/substrate/pull/1365 to test downgrading from Full -> DurDir. Test passed.

BenTheElder:
Thanks!

HavenXia:
This also fixes #817

PR #1388, opened 2026-09-02
ate-setup: add flags and functionality to use pre-built images

### LINKED ISSUE #1387: setup-ate should handle prebuilt images
right now ate-setup builds images from scratch. I'm interested in having a workflow installation option that handles deploying prebuilt images.

annapendleton:
Fixes #1387

Also tested manually on GKE cluster and kind cluster both existing build on demand pathways and relying on prebuilt images from my own registry.

- [X] Tests pass

### DIFF
--- a/cmd/ate-setup/internal/images/images.go
+++ b/cmd/ate-setup/internal/images/images.go
@@ -36,7 +36,9 @@ const ModulePath = "github.com/agent-substrate/substrate"
 // The list is closed on purpose. A reference that is not in it names something
 // this build has no published image for, and failing beats guessing at a name
 // that was never pushed. TestComponentsCoverTheManifests keeps it in step with
-// the manifests.
+// the manifests, which is half of it: a new component also has to be published
+// under the same repository and tag as the rest before a pre-built install can
+// use it.
 var Components = []string{
 	"cmd/ateapi",
 	"cmd/atecontroller",
--- a/cmd/ate-setup/internal/images/images.go
+++ b/cmd/ate-setup/internal/images/images.go
@@ -1,0 +1,105 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package images decides where ate-setup gets container images from.
+//
+// The manifests name images as ko:// import paths, which ko builds and
+// publishes on every install. That needs a checkout, a Go toolchain, and a
+// registry to push to. Installing an already-published release instead means
+// mapping those same references onto images that exist, which is what Source
+// describes and Prebuilt performs.
+package images
+
+import (
+	"fmt"
+	"path"
+)
+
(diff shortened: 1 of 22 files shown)


cmd/ate-setup/internal/images/prebuilt.go
```diff
@@ -0,0 +1,122 @@
...
+	// attempt at a time is slow, and the failures are usually related.
+	unknown := make(map[string]bool)
+
+	out := koRef.ReplaceAllFunc(manifest, func(match []byte) []byte {
+		image, ok := imageByRef[string(match)]
+		if !ok {
+			unknown[string(match)] = true
+			return match
+		}
+		return []byte(p.src.Repo + "/" + image + ":" + p.src.Tag)
```
aditya-shantanu: Tag-only references break every demo deploy in prebuilt mode: ActorTemplate images, image volumes, and SandboxConfig pauseImage all carry the CEL rule `self.contains('@')` ("All images must be pinned"), so `--image-repo ... deploy demo counter` is rejected at admission. Resolving each tag to its digest at install time (one HEAD per component, e.g. go-containerregistry) and emitting `repo/name:tag@sha256:...` fixes that, and also restores the digest-pinning ko gave the core deployments — immunity to mutable-tag drift.

cmd/ate-setup/internal/images/images.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Components is every package the installer can deploy a published image for,
+// as an import path within the module.
+//
+// The list is closed on purpose. A reference that is not in it names something
+// this build has no published image for, and failing beats guessing at a name
+// that was never pushed. TestComponentsCoverTheManifests keeps it in step with
+// the manifests.
+var Components = []string{
+	"cmd/ateapi",
```
aditya-shantanu, line 43: Worth one sentence here or in commands.md: with a closed Components list, adding a new deployable package fails prebuilt installs until this list (and the published release set) grows with it — TestComponentsCoverTheManifests catches the manifest side, but a release pipeline has to publish the same set.

cmd/ate-setup/internal/images/prebuilt.go
annapendleton: Good catch — real break, not just lost pinning. The rule is in three places: `actortemplate_types.go:45` and `:301`, `sandboxconfig_types.go:83`. Server dry-run against a live cluster:

```
# tag only (what this emitted before)
spec.containers[0].image: Invalid value: "localhost:5001/counter:v0.0.0-kindtest":
All images must be pinned (changing the image invalidates snapshots)

# tag + digest
actortemplate.ate.dev/pin-check created (server dry run)
```

Fixed as suggested: `Prebuilt` takes an `images.Digester`, one `remote.Head` per image via go-containerregistry (already vendored), cached per install, docker-keychain auth like ko. Output is `<repo>/<image>:<tag>@sha256:...` — the same shape ko produces at `pkg/publish/default.go:273`. A failed lookup is a hard error, never a fallback to the bare tag.

Consequence worth flagging: `--image-repo` now needs read access to the registry from wherever `ate-setup` runs, not just from the cluster. Documented in `commands.md`, `differences.md`, and the README. Teardown is unaffected — delete paths don't resolve images.

Verified on kind: `deploy ate-system --image-repo localhost:5001 --image-tag v0.0.0-kindtest` rolls out fully pinned, e.g. `localhost:5001/ateapi:v0.0.0-kindtest@sha256:f12664d0...`.

cmd/ate-setup/internal/images/images.go
annapendleton: Agreed. Noted on `Components` that the test holding it against the manifests is half of what adding a component takes — the image also has to be published under the same repository and tag as the rest before a pre-built install can use it. Same point in `commands.md` and the `differences.md` closed-list paragraph.

aditya-shantanu (approved):

PR #1389, opened 2026-09-02
temporarily drop device support

BenTheElder:
See: https://github.com/agent-substrate/substrate/issues/752#issuecomment-5501074841

TLDR: I think we need to rethink how we express devices.

This PR drops the user facing support, while retaining some other bits around CDI for-reuse when we're ready.


thockin:
Merging

PR #1391, opened 2026-09-02
ate-setup: widen digest-lookup auth and pass through pinned tags

annapendleton:
Two follow-ups to the pre-built image install.

Digest lookups authenticated with the docker config file alone. Someone
installing a published release onto GKE, whose Artifact Registry credentials are
gcloud's rather than a credential helper wired into `~/.docker/config.json`, got
a working build from source and a 401 from `--image-repo`. The lookup keychain is
now `authn.NewMultiKeychain(authn.DefaultKeychain, google.Keychain)`, the same
relative order ko uses. `DefaultKeychain` stays first, so `~/.docker/config.json`
still wins wherever it has an entry, and `google.Keychain` returns `Anonymous`
immediately for any non-Google registry, so this is a no-op for Docker Hub, ghcr,
ECR, ACR and the Kind local registry. ECR and ACR would need credential-helper
modules this repository does not depend on, so those still need a `docker login`;
the doc comment no longer claims the lookup authenticates the way ko does.

Separately, `pin` appended the looked-up digest unconditionally, so
`--image-tag v1@sha256:...` produced `<repo>/<image>:v1@sha256:...@sha256:...`,
which no registry can parse. A tag that already carries a digest now passes
through as written.

No dependency change: `pkg/v1/google` was already vendored, and `atelet` already
uses the same package for its image pulls.

Verified against Artifact Registry with `DOCKER_CONFIG` pointed at an empty
directory, which reproduces the affected setup — without this change the lookup
fails with `DENIED: Unauthenticated request`, with it the tag resolves. Also
installed on a Kind cluster from a local registry, where all five control-plane
images came up pinned to the digests the registry published.

- [X] Tests pass

### DIFF
--- a/cmd/ate-setup/internal/images/prebuilt_test.go
+++ b/cmd/ate-setup/internal/images/prebuilt_test.go
@@ -94,6 +94,15 @@ func TestPrebuiltResolveBytes(t *testing.T) {
 			want: "spec:\n  workerImage: example.com/substrate/ateom-gvisor:v1.2.3" + digestSuffix + "\n",
 		},
 		{
+			// A tag can carry its own digest, and then there is nothing to look
+			// up. The stub would answer with testDigest a second time, so the
+			// expected output is also what catches a redundant lookup.
+			name: "tag already carries a digest",
+			src:  images.Source{Repo: "example.com/substrate", Tag: "v1.2.3" + digestSuffix},
+			in:   "image: ko://github.com/agent-substrate/substrate/cmd/ateapi\n",
+			want: "image: example.com/substrate/ateapi:v1.2.3" + digestSuffix + "\n",
+		},
+		{
 			name: "nested package maps to its image name",
 			in:   "image: ko://github.com/agent-substrate/substrate/demos/multi-template/fspersist\n",
 			want: "image: example.com/substrate/fspersist:v1.2.3" + digestSuffix + "\n",
--- a/cmd/ate-setup/commands.md
+++ b/cmd/ate-setup/commands.md
@@ -60,6 +60,16 @@ Each reference is then pinned to the digest its tag names, which takes one HEAD
 request per image, so the installer needs read access to `REPO` and not only the
 cluster does.
 
+That read is authenticated with the docker config file and the credential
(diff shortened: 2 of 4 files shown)


cmd/ate-setup/internal/images/prebuilt_test.go
```diff
@@ -93,6 +93,15 @@ func TestPrebuiltResolveBytes(t *testing.T) {
 			in:   "spec:\n  workerImage: ko://github.com/agent-substrate/substrate/cmd/ateom-gvisor\n",
 			want: "spec:\n  workerImage: example.com/substrate/ateom-gvisor:v1.2.3" + digestSuffix + "\n",
 		},
+		{
+			// A tag can carry its own digest, and then there is nothing to look
+			// up. The stub would answer with testDigest a second time, so the
+			// expected output is also what catches a redundant lookup.
+			name: "tag already carries a digest",
```
HavenXia, line 100: With this passthrough, `--image-tag v1@sha256:...` also becomes the substrate version (steps/version.go:51), and  such format fail the label check at version.go:57 because we use such in `atelet` name, so deploy ate-system errors at the atelet step unless VERSION is also set. 

Would you take the tag before @ there (strings.Cut(tag, "@")) so a digest-carrying tag works without extra env? 

In `cmd/ate-setup/internal/steps/version.go` we can add a 
```
if e.Cfg.Images.IsPrebuilt() {
    raw = os.Getenv("VERSION")
    if raw == "" {
        // The tag may carry a digest; the version is the tag alone.
        raw, _, _ = strings.Cut(e.Cfg.Images.Tag, "@")
    }
}
```
So we only add the `v1` before @ as version. Which can satisfy the version format

HavenXia: I did that in #1391

HavenXia (approved):

cmd/ate-setup/internal/images/prebuilt_test.go
annapendleton: #1400 looks good to me, based on chat, let's merge this one first then yours

PR #1392, opened 2026-09-02
docs: manual rolling upgrade runbook (clone workerpool approach)

### LINKED ISSUE #1272: [Upgrade] System upgrade runbook

### LINKED ISSUE #1273: [Upgrade] Rollback support for system upgrade CLI/binary/script

HavenXia:
The runbook for moving a running substrate to a new build without losing actor state. An user runs the roll by hand with `kubectl`, `kubectl ate`, `ate-setup`, `jq`, and `grpcurl`, and every piece of upgrade state lives in cluster objects, so the roll can stop and resume at any point.

The order follows the [upgrade design](https://docs.google.com/document/d/1JduAyGZFyqdNp4UhKiv0EN-is3iW5tf5BwGTWbt2ouI/edit?usp=sharing&resourcekey=0-MXG12QCkleIxhY7cOB7g6A): CRDs, then `ate-controller`, then the dataplane node by node, then `ate-api-server` and `atenet`. 

The dataplane moves by version label. The new atelet DaemonSet sits next to the old one, each serving WorkerPool is cloned with the new worker image and the new version pin, and each node is drained (workers marked `DRAINING` through the `DrainWorker` RPC), emptied by suspending its actors, relabeled, and cleared of old worker pods. The old objects stay untouched until a separate retire step, so rollback is one label flip per node.

Tested end to end locally on a GKE cluster: install at one build, run counter actors, upgrade to a second build following the document, and confirm the actors resume on the new pool with their state intact. Rollback and retire were exercised the same way.

Fix #1272, fix #1273.

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -5,9 +5,11 @@ Agent Substrate install to a new build version. The roll moves one
 node at a time, so no actor loses state and at most one node's worth
 of capacity is out of service while the rest of the fleet keeps
 serving. It needs `kubectl`, `kubectl ate`, `go run ./cmd/ate-setup`,
-`jq`, `grpcurl`, and on GKE `gcloud`. All of its state lives in
-cluster objects, so you can stop, look around, and pick up again at
-any point.
+`jq`, and `grpcurl`; nothing in it is specific to one Kubernetes
+provider. On GKE it also needs `gcloud` for the two node pool steps,
+1 and 8, which other providers have their own equivalents of. All of
+its state lives in cluster objects, so you can stop, look around, and
+pick up again at any point.
 
 The order is `ate-controller` first, then the dataplane, then the rest
 of the control plane. The controller goes first because it manages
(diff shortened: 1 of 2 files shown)


annapendleton (dismissed):

docs/upgrade.md
```diff
@@ -0,0 +1,485 @@
+# Rolling upgrade runbook
+
+This runbook provides step by step guidance to upgrade a running
+Agent Substrate install to a new build version. The roll moves one
+node at a time, so no actor loses state and at most one node's worth
+of capacity is out of service while the rest of the fleet keeps
+serving. It needs `kubectl`, `kubectl ate`, `go run ./cmd/ate-setup`,
+`jq`, `grpcurl`, and on GKE `gcloud`. All of its state lives in
```
annapendleton: "GKE `gcloud`." - should this be more agnostic? does this only work on GKE or does it work for all substrate components?

docs/upgrade.md
```diff
@@ -0,0 +1,487 @@
...
+   pacing.
+
+## Preflight
+
+The roll uses these names throughout. Collect them up front:
+
+| name | what it is | how to get it |
+|---|---|---|
+| `$CLUSTER`, `$ZONE` | the GKE cluster and its location | `gcloud container clusters list` |
+| `$NODEPOOL` | the GKE node pool | `gcloud container node-pools list --cluster $CLUSTER --zone $ZONE` |
```
annapendleton, line 64: these gcloud commands - possibly should be kind commands (eg. assume kind is our base provider in substrate repo)

annapendleton (approved):
not worth blocking review on - but we likely need to revisit some of the wording here so it's more cloud generic

PR #1403, opened 2026-09-02
Update gVisor to the 2026-09-02 nightly and switch to zstd tarballs

dberkov:
## Summary

Improves the cold-node activation performance tracked in #811: the gVisor release extract on a cold node currently exceeds the router resume budget, so the first resume after a release bump always fails.

- Point the default gvisor SandboxConfig at the 2026-09-02 nightly, using the `gvisor.tar.zstd` tarballs the nightlies now publish alongside `gvisor.tar.bz2`. The win is decode speed, not just download size: stdlib bzip2 is a pure-Go, single-threaded decoder, and extracting the ~165 MB release costs a node roughly 20 seconds of CPU on the first actor operation it hosts — the dominant term of a cold-node activation. With zstd the same extraction takes ~2.9 s, measured on a GKE node running this change. The tarball is also ~22% smaller to download (129 MB vs 166 MB for x86_64).
- Teach `extractTarArchive` to decompress `.tar.zst` / `.tar.zstd` archives using `github.com/klauspost/compress/zstd` (already a dependency via ategcs), with test coverage for both suffixes.
- Update the `gvisor.tar.bz2` mentions in docs, manifests comments, and the SandboxConfig API comment (generated CRD refreshed via `hack/update/codegen.sh`).

The bucket only publishes sha512 checksums; the pinned sha256 sums were computed from downloads verified against the published sha512 sums.

## Test plan

- `go test ./cmd/atelet/` (includes new `.tar.zst` / `.tar.zstd` extraction cases)
- `go build ./cmd/atelet/...`, `go vet`, `gofmt` clean
- Verified the real x86_64 tarball lists `runsc` plus the `gvisor-bin/` helpers
- Deployed to a GKE cluster: atelet fetched and extracted the 2026-09-02 x86_64 zstd tarball in ~2.9 s (vs ~20 s for bz2)

### DIFF
--- a/cmd/atelet/sandbox_assets.go
+++ b/cmd/atelet/sandbox_assets.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	gzip "github.com/klauspost/compress/gzip"
+	"github.com/klauspost/compress/zstd"
 	"io"
 	"io/fs"
 	"log/slog"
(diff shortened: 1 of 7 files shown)


nybidari (approved):
LGTM

PR #1405, opened 2026-09-02
setup-gcp: enable artifactregistry.googleapis.com in bootstrap

aditya-shantanu:
Bootstrap grants `roles/artifactregistry.reader` to the GKE node service account and the atelet workload-identity principal, and ko pushes the control-plane images through gcr.io's Artifact Registry backing — but `enableRequiredAPIs` never enabled `artifactregistry.googleapis.com`, so a fresh project failed at image push/pull instead of step 1.


aditya-shantanu:
Consolidated into #1408 per feedback — closing.

PR #1406, opened 2026-09-02
teardown.sh: cover everything bootstrap creates, without a dev-env file

aditya-shantanu:
`hack/teardown.sh --all` left billable/lingering state behind and could not run in common situations:

- **Revoke atelet's project-level bindings** (`roles/storage.objectAdmin`, `roles/artifactregistry.reader` for the workload-identity principal): `grant_atelet_permissions` adds them, nothing removed them.
- **Delete the three Substrate monitoring dashboards**, matched by the display names in `tools/setup-gcp/dashboards/`.
- **Accept configuration from the environment** when `.ate-dev-env.sh` is absent, so installers and one-liners can drive the script; a missing variable is now named specifically.
- **Drop the kubectl cluster-admin precheck**: every step talks to GCP, not the cluster, and the check blocked tearing down a cluster that was already gone or unreachable.
- `--all` now runs in the true reverse of bootstrap's setup order.

### DIFF
--- a/hack/teardown.sh
+++ b/hack/teardown.sh
@@ -16,29 +16,35 @@
 
 set -o errexit -o nounset -o pipefail
 
-# Source the environment variables
+# Source the environment variables. The file is optional: an installer (or a
+# user pasting a one-liner) can pass the same variables through the
+# environment instead, which also makes teardown possible on a machine that
+# never had a dev-env file.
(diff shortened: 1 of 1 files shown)


aditya-shantanu:
Consolidated into #1408 per feedback — closing.

PR #1407, opened 2026-09-02
commands.md: the hack/install-demo-*.sh scripts still exist

aditya-shantanu:
The closing paragraph claimed the per-demo hack scripts are gone, but `hack/install-ate.sh` still sources `hack/install-demo-*.sh` and generates its `--deploy-demo-NAME` / `--delete-demo-NAME` flags from them. Describe the two installers as parallel instead of declaring one dead.

PR #1408, opened 2026-09-02
Address substrate-gke installer feedback: AR API, teardown coverage, doc fix

aditya-shantanu:
Three fixes from user feedback on the substrate-gke installer experience, consolidated per reviewer preference (replaces #1405, #1406, #1407):

**1. setup-gcp: enable `artifactregistry.googleapis.com` in bootstrap.** Bootstrap grants `roles/artifactregistry.reader` to the GKE node service account and the atelet workload-identity principal, and ko pushes the control-plane images through gcr.io's Artifact Registry backing — but `enableRequiredAPIs` never enabled the API, so a fresh project failed at image push/pull instead of step 1.

**2. teardown.sh: cover everything bootstrap creates, without a dev-env file.**
- Revoke atelet's project-level bindings (`roles/storage.objectAdmin`, `roles/artifactregistry.reader`), which `grant_atelet_permissions` adds but nothing removed.
- Delete the three Substrate monitoring dashboards, matched by the display names in `tools/setup-gcp/dashboards/`.
- Accept configuration from the environment when `.ate-dev-env.sh` is absent (missing variables are named specifically), so installers and one-liners can drive the script.
- Drop the kubectl cluster-admin precheck: every step talks to GCP, not the cluster, and it blocked tearing down a cluster that was already gone.
- `--all` now runs in the true reverse of bootstrap's setup order.

**3. commands.md: the `hack/install-demo-*.sh` scripts still exist.** The closing paragraph claimed they are gone, but `hack/install-ate.sh` still sources them and generates its `--deploy-demo-NAME` / `--delete-demo-NAME` flags from them.

### DIFF
--- a/cmd/ate-setup/commands.md
+++ b/cmd/ate-setup/commands.md
@@ -173,6 +173,7 @@ The demo list is not hard-coded here — it is built from the registry in
 `go run ./cmd/ate-setup deploy demo --help` is authoritative for both the list
 and the per-demo flags.
 
-Each demo previously had its own `hack/install-demo-*.sh`, sourced by the
-installer, which registered the flags above. Those scripts are gone; the demos
-now live in [`internal/demos`](internal/demos).
+The demos also each have a `hack/install-demo-*.sh`, sourced by
+`hack/install-ate.sh`, which registers `--deploy-demo-NAME` /
+`--delete-demo-NAME` flags on that installer. ate-setup does not use those
+scripts; its demos live in [`internal/demos`](internal/demos).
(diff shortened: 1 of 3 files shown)


aditya-shantanu:
Pushed one more commit: the env-var guard now lives per step (`require` helper), so each step demands only the variables it uses — deleting a bucket no longer insists on `NODE_POOL_NAME` — and `--all` skips the separate node-pool deletion when `NODE_POOL_NAME` is unset, since deleting the cluster removes its pools. This lets an installer drive the full teardown through the environment without tracking the pool name.

annapendleton (approved):

Issue #1395, opened 2026-09-02
ateapi: a control plane that cannot reach its store serves 200 on /healthz and /readyz, stays Ready, and fails every RPC


mayawang:
### Expected Behavior

Readiness should mean "this process can serve". When `ate-api-server` cannot reach its store it can
serve nothing — every RPC fails — so it should fail `/readyz`, leave the Service endpoints, and
surface a signal that something above it can act on: an alert, a metric, a Condition, anything.

Liveness should keep succeeding throughout. The process is alive and will recover on its own once
the store returns; restarting it would make things worse, not better (see #<O-26>).

### Actual Behavior

The two probes are indistinguishable, and both are green while the control plane serves nothing.

Store scaled to zero, sampled every 15 s for six minutes:

| | |
|---|---|
| Baseline | `healthz=200 readyz=200 Ready=True restarts=0 rpc=OK` |
| T+15s → T+360s — **24 consecutive samples** | `healthz=200 readyz=200 Ready=True restarts=0 rpc=FAIL` |

Six minutes in which the pod is 200 on `/healthz`, 200 on `/readyz`, `Ready=True` to both the kubelet
and its Service, has never restarted, is still receiving traffic — and fails **every** RPC. Nothing
in Kubernetes has any way to know, and nothing in the control plane says so either.

This is not a slow-detection problem. There is no detection: `/readyz` does not consult the store
under any circumstances, so the interval is infinite.

What it costs, concretely: on 27 Aug we spent about three hours on an outage of this shape, most of
it looking in the wrong place, because every signal at the top of the stack was green and the only
component visibly misbehaving was the api-server crash loop from #<O-26> — which is a *symptom* of
the store being gone, not the cause. A store outage presents as an api-server bug.

### Steps to Reproduce the Problem

1. Install on GKE with the PostgreSQL backend and confirm `ate-api-server` is serving:
   `kubectl-ate get atespaces` returns.
1. Take the store away: `kubectl -n ate-system scale sts/postgres --replicas=0`.
1. Poll all four signals every 15 s for six minutes:
   ```
   kubectl -n ate-system get pod <ate-api-server pod> -o jsonpath='{.status.containerStatuses[0].ready} {.status.containerStatuses[0].restartCount}'
   curl -s -o /dev/null -w '%{http_code}' localhost:9090/healthz      # via port-forward
   curl -s -o /dev/null -w '%{http_code}' localhost:9090/readyz
[...]

PR #1413, opened 2026-09-02
ateapi: give the api-server a startup probe covering its store-connect budget

### LINKED ISSUE #1394: ateapi: ate-api-server is SIGTERMed by its own liveness probe halfway through its store-connect retry budget, then exits instead of retrying
### Expected Behavior

A control plane whose store is temporarily unreachable **at startup** should do one of two things:

1. Come up, serve its health surface, report **not ready**, and keep retrying for the life of the
   process. A control plane that cannot reach its store should stay up and say so.
2. At an absolute minimum, be allowed to spend the retry budget it configures for itself.
   `postgresConnectTries = 30` at `postgresConnectPeriod = 2s` is a deliberate 60-second budget.

Either way, a store outage shorter than the budget should not produce a restart.

### Actual Behavior

Neither. The process gets **~30 seconds of its 60-second budget**, is killed by its own liveness
probe, exits, and enters `CrashLoopBackOff` — where exponential backoff then keeps it down well past
the point where the store has come back.

Measured on a Postgres install on GKE, store scaled to zero, one replica deleted to force it onto
the boot path:

```
19:31:06.317  attempt 1  ... lookup postgres.ate-system.svc: no such host
19:31:34.396  attempt 15 ... lookup postgres.ate-system.svc: no such host
19:31:36.192  ERROR Failed to set up persistence backend
              err="setting up PostgreSQL: context canceled"
```

**15 of 30 attempts, and the loop ends in `context canceled` rather than budget exhaustion.** Pod
events:

```
Liveness probe failed: Get "http://10.x.x.x:9090/healthz": connect: connection refused
Killing — Container ate-api-server failed liveness probe, will be restarted   ×6
```

Result: **CrashLoopBackOff, 7 restarts in 7 minutes**, each one `Failed to set up persistence
backend`.

Two properties of this that are worth stating plainly, because both are counter-intuitive:

- **The kill is not caused by an unhealthy process.** It is caused by there being **nothing
  listening** on `:9090` yet. `/healthz` is not returning an error; the port is closed.
- **The fatal is boot-only.** A *running* replica, on the same cluster during the same outage, rode
  out six minutes with `restarts=0` and was serving again **15 s** after the store returned. The
  pgx pool simply starts working. So the dangerous event is precisely the one that puts every
  replica on the boot path at once — a node replacement.

### Steps to Reproduce the Problem

1. Install on GKE with the PostgreSQL backend (`--store-backend=postgres`). Confirm both
   `ate-api-server` replicas are Ready and serving.
1. Take the store away: `kubectl -n ate-system scale sts/postgres --replicas=0`. Note that nothing
   happens — the running replicas stay `Ready=True` with `restarts=0` and no probe fails. (That is
   the companion defect, #<O-27>.)
1. Force one replica onto the boot path while the store is down:
   `kubectl -n ate-system delete pod <one ate-api-server pod>`. This is what a node replacement does
   to every replica simultaneously.
1. Watch the replacement pod: `kubectl -n ate-system get pod -w` and
   `kubectl -n ate-system describe pod <new pod>`. It reaches `CrashLoopBackOff` within ~90 s.
   The events show `Liveness probe failed: ... 9090/healthz: connect: connection refused`, **not** a
   probe returning a failure status.
1. Read the container log. Count the `Failed to connect to PostgreSQL, retrying...` lines: there are
   **15**, not 30, and the loop terminates with `setting up PostgreSQL: context canceled`.
1. Bring the store back: `kubectl -n ate-system scale sts/postgres --replicas=1`. The **surviving**
   replica serves again within ~15 s with `restarts=0`. The **crash-looping** replica comes Ready
   only when its own CrashLoopBackOff timer next fires — for us, at the same second Postgres became
   available, after 7 restarts, but by then the backoff had already reached the tens of seconds and
   is capped at 5 min.

Why CI does not catch it: e2e installs bring Postgres up before or alongside the api-server and
never hold it down across a pod start, so the retry loop is never exercised under a probe.

### Specifications

- **Version:** `4c1b37d` (the base of `release-0.1-rc`) plus three cherry-picks. **Confirmed still
  present on `main` at `ea3bdc32`** (2026-09-02) by reading the code — see line references below.
  Store-agnostic: this survived the valkey → Postgres migration (`60073ecd`, #940) untouched, and
  the same mechanism produced a much longer outage on a valkey install on 27 Aug.
- **Platform:** GKE, `us-central1-c`, cluster provisioned by `setup-gcp bootstrap`,
  `--store-backend=postgres`, `postgres-0` 2/2 with a Bound 500Gi `standard-rwo` PVC. Reproduced
  2026-08-27.
- **Evidence:** container logs and pod events from the run above.

---

### Root cause

Three things on `main` at `ea3bdc32` that are individually reasonable and jointly fatal.

**1. The health surface starts after the store connect.** `cmd/ateapi/main.go`:

- `:127` — `persistence, err := connectStore(shutdownCtx)`
- `:129` — `serverboot.Fatal(ctx, "Failed to set up persistence backend", err)`
- `:236` — `go serverboot.StartMetricsServer(ctx, serverboot.MetricsServerOptions{...})` with
  `EnableHealthz: true` at `:239`

So `:9090` binds roughly 110 lines of boot later than the retry loop begins. During the entire retry
window the port is closed.

**2. The liveness probe's budget is shorter than the retry budget it is racing.**
`manifests/ate-install/ate-api-server.yaml` sets `livenessProbe` on `/healthz:9090` with
`initialDelaySeconds: 10`, `periodSeconds: 10`, and no `failureThreshold` — so the default 3
applies. First SIGTERM lands at **~30 s** from container start. The retry loop wants 60 s.

**3. The retry loop is rooted in the shutdown context, so the SIGTERM cancels it.**
`connectStore(shutdownCtx)`, where `shutdownCtx` comes from `signal.NotifyContext(ctx, SIGTERM,
os.Interrupt)`. The `select` in `connectPostgresWithRetries` (`:330`) returns `ctx.Err()` on
cancellation. Hence `context canceled` at attempt 15 rather than
`connect to PostgreSQL after 30 attempts`.

Then `serverboot.Fatal` at `:129` exits, the kubelet restarts with backoff (10s, 20s, 40s, 80s,
160s, capped at 300s), and each restart repeats the whole thing.

### Why this is worse on `main` than it was

On the valkey install this was one contributor to a long outage. On `main` it is the *primary*
failure path for an ordinary GKE event, because the store got less available at the same time as
this code stayed the same:

`manifests/ate-install/postgres.yaml` is `replicas: 1` on an RWO PVC, with **no
PodDisruptionBudget**, no `topologySpreadConstraints` and no node affinity — while `ate-api-server`
has a PDB and two replicas. #1253 raised the claim to 500Gi and 2–16 CPU, which makes the pod harder
to place on a replacement node. After an ungraceful node loss, GKE's force-detach of the PD alone is
**~6 minutes** before the volume can attach elsewhere, then reschedule and WAL replay.

Six minutes against a thirty-second window. A node-pool upgrade — on by default in every GKE release
channel — therefore restarts every `ate-api-server` replica at the exact moment the store is least
available, and every one of them lands on this path. Whole-pool node replacement under load has not
been tested against the Postgres backend, so the true window length is unknown; it is not plausibly
under 30 s.

### Interaction with #1196, which is landing on this function now

#1196 (*ateapi: add versioned PostgreSQL schema migrations*) edits `connectPostgresWithRetries`
directly. It makes one genuine improvement and one regression to this issue:

- **Better:** it adds `atepg.ErrUnavailable` and returns immediately on any error that is not it, so
  a bad DSN or a failed migration fails in one attempt instead of burning the budget. That is the
  right change — but it is the other branch. The store-unavailable path keeps all 30 attempts, the
  same ordering, and the same 40-second ceiling.
- **Worse:** it moves goose migrations onto the boot critical path, ahead of the health server,
  behind a `pg_advisory_xact_lock`. Boot gets longer and gains a serialization point: with two
  replicas restarting together, one holds the lock while the other waits — both inside the window
  where nothing is listening on `:9090`. On an empty schema that is milliseconds; on a real database
  during an upgrade it is not bounded by anything.

**The ordering fix should therefore be sequenced with #1196, not against today's `main`** — the
health server has to start before `connectStore` *and* before the migration step. It is a handful of
lines in a function that PR already touches.

### Candidate fixes

1. **Start the health surface before `connectStore`.** Move the `StartMetricsServer` goroutine (and
   its `Readiness`) above the store connect, so `/healthz` answers from the first second of the
   process. This alone converts the failure from "killed at 15 of 30 attempts" to "uses its full
   budget", and it is the two-line version.
1. **Do not root the connect retry in the shutdown context** — or, better, keep it rooted there but
   let the process survive a store that is merely slow, per fix 3.
1. **Retry for the life of the process instead of exiting.** Replace `serverboot.Fatal` on store
   failure with a background reconnect loop, and report the condition through readiness (see
   #1395 , which needs a reversible readiness predicate for exactly this). This is the fix that
   matches the observed behaviour of a *running* replica, which already rides out arbitrary
   outages — a booting one should not be strictly less resilient than a running one.
1. **Set an explicit `failureThreshold` and a `startupProbe`** on the manifest, so slow boots are
   distinguished from dead processes. Complementary, not a substitute: it changes the number without
   fixing the ordering.

1 and 3 are the two I would pick. 1 is safe enough to land on its own before the other.

**One design constraint, because the obvious fix for #1395  breaks this one:** liveness must **not**
be gated on store reachability. `/healthz` should mean "the process is alive", start before the store
connect, and never consult the store; `/readyz` is where store reachability belongs. Gating
`/healthz` on the store would make this issue strictly worse — the kubelet would then kill a pod that
is correctly waiting.

Related: #1395 (same health surface, opposite phase), #1196 (lands on this function), #636 (the
valkey instance of the same crash loop, closed by deleting the backend rather than by a fix).

mayawang:
Partially addresses #1394 — the "spend the retry budget it configures for itself" half of that issue's expected behavior. Candidate fix 4 from the issue, done first because it is manifest-only.

## The problem, in one line

`ate-api-server` configures itself a 60s store-connect budget and is killed by its own liveness probe at ~30s, halfway through it.

## Why the kill happens

`:9090` is bound near the end of boot — after `connectStore` and, since #1196, after the goose migrations — so nothing answers `/healthz` for the whole of that window. The liveness probe was the only probe governing it: `initialDelaySeconds: 10` then probes at 10s, 20s and 30s, so the default `failureThreshold: 3` is met at ~30s. The events show `connect: connection refused`, not a probe returning failure — the port is closed, the process is fine.

The kill then makes it worse than a plain restart. `connectStore(shutdownCtx)` roots the retry loop in the signal context, so SIGTERM cancels it mid-budget: the loop reports `context canceled` at attempt 15 of 30 rather than exhausting the budget, `serverboot.Fatal` exits, and CrashLoopBackOff holds the pod down well past the point where the store came back.

A *running* replica rides out the same outage with `restarts=0` and serves again ~15s after the store returns. A booting replica was strictly less resilient than a running one, which is backwards.

## The change

One `startupProbe`, 24 × 5s. Liveness is suspended until `/healthz` answers, so boot gets ~115s before the kubelet intervenes, against a ~60s connect budget.

| | before | after |
|---|---|---|
| kubelet kills at | ~30s | ~115s |
| connect attempts used | 15 of 30 | 30 of 30 |
| failure reported as | `context canceled` | `connect to PostgreSQL after 30 attempts: …` |

A dead process is still restarted, just on a budget sized for boot rather than one that never was. And because a startup probe also suspends the *readiness* probe, the pod now stays out of Service endpoints while booting instead of being latched Ready by the zero-value `Readiness` — a small move in the right direction for #1395 rather than against it, which is why this rather than simply raising `failureThreshold`.

## What this deliberately does not do

- **It does not fix #1394.** The process still `Fatal`s once its own 60s budget is gone. The motivating scenario there is an ungraceful node loss, where GKE's PD force-detach alone is ~6 minutes; this raises tolerance from 30s to 60s, and only "retry for the life of the process, report through readiness" survives six minutes. That fix needs the reversible readiness predicate #1395 describes.
- **It does not bound the migrations.** #1196 put goose on the boot critical path behind a `pg_advisory_xact_lock`, so with two replicas restarting together one waits on the other — inside the window where nothing is listening. On an empty schema that is milliseconds; on a real database during an upgrade it is not bounded by anything, and no probe budget can bound it. Ordering can.
- **It does not change the boot ordering.** A slow boot and a dead process are still indistinguishable while the port is closed. Moving the health surface ahead of `connectStore` is candidate fix 1 and the natural follow-up.

## Testing

`hack/verify-all.sh` passes (`metrics.sh` skipped locally — needs weaver or docker; no metrics change here).

Note for anyone rebasing: `make test` currently fails on `TestExternalVolumeRenders` in `cmd/ate-setup/internal/demos/counter` on a clean `cbae8250`, unrelated to this change.

### DIFF
--- a/manifests/ate-install/ate-api-server.yaml
+++ b/manifests/ate-install/ate-api-server.yaml
@@ -160,6 +160,21 @@ spec:
           initialDelaySeconds: 5
           periodSeconds: 2
           failureThreshold: 3
+        # :9090 is bound near the end of boot, after the store connect and
+        # the schema migrations, so nothing answers here until both finish.
+        # The store connect alone budgets 60s (30 attempts, 2s apart) and the
+        # migrations serialize across replicas on an advisory lock. Without a
+        # startup probe the liveness probe below SIGTERMs the container at
+        # ~30s, halfway through that budget: the retry loop is rooted in the
+        # signal context, so it reports "context canceled" and the process
+        # exits into CrashLoopBackOff instead of waiting out a store that is
+        # merely slow to return. A booting replica is not a dead one.
+        startupProbe:
+          httpGet:
+            path: /healthz
+            port: 9090
+          periodSeconds: 5
+          failureThreshold: 24
         # /healthz stays 200 while a terminating pod drains; /readyz
         # turns 503, so liveness and readiness diverge correctly during
         # shutdown.


mayawang:
@iplay88keys tagging you for context rather than as a formal reviewer — GitHub won't let me request you. This sits directly on top of #1196: the "can't bound the migrations" caveat is about the advisory lock you put on the boot path, and I'd like a sanity check that ~115s is the right budget now that goose runs ahead of the health server.

annapendleton (approved):
having liveness probe makes sense - would like to vet the 120s value we have but could be fixed in follow-up PR if needed as well

iplay88keys:
I'm not sure on the specific timing, but I agree a liveness probe is useful.

annapendleton:
I'm comfortable pushing this through and we can adjust timing in follow-up PRs, merging now.

PR #1416, opened 2026-09-02
atelet: retry transient GCS errors on snapshot uploads

dberkov:
Under a suspend load burst, GCS sheds writes with 429 rateLimitExceeded ("request distribution is too uneven across the key-ranges") while it scales a cold bucket's key ranges up to the offered rate. The storage client's default RetryIdempotent policy never retries an object write that carries no precondition, so a single transient 429 on pages.img propagated straight to FAILED_SAVE_SNAPSHOT and failed the suspend. In a load test this showed up as hundreds of failed checkpoints in the first minute of the run.

Set a RetryAlways policy on the wrapped client and on each client in the part-upload pool, so every operation retries 408/429/5xx with backoff. RetryAlways is safe for this client: every write targets a unique name (snapshot UUID directories, runID-suffixed part names) or is idempotent (compose and copy from fixed sources, delete). Suspend is latency-sensitive, so the worst case is bounded at 5 attempts with under 5s of backoff sleep in total.

Because RetryAlways is only safe for this package's own operations, NewGCSClient no longer accepts a caller-built *storage.Client it would have to mutate. It builds the client itself from ctx and opts, the same way it already builds the pool clients, so the policy cannot leak onto a client shared with other code, and the requirement that the pool's options match the client's holds by construction. atelet's main no longer touches the raw storage client at all.

### DIFF
--- a/cmd/atelet/internal/ategcs/gcs.go
+++ b/cmd/atelet/internal/ategcs/gcs.go
@@ -21,8 +21,10 @@ import (
 	"fmt"
 	"io"
 	"sync"
+	"time"
 
 	"cloud.google.com/go/storage"
+	"github.com/googleapis/gax-go/v2"
 	"google.golang.org/api/option"
 )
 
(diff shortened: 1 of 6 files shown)


maxsmythe (approved):
LGTM

Issue #866, opened 2026-08-11
Add DeleteOptions and honor version/uid preconditions on all Delete RPCs


laoj2:
[docs/api-style-guide.md](https://github.com/agent-substrate/substrate/blob/main/docs/api-style-guide.md#35-delete) §3.5 and §7 require every Delete<Type>Request to carry a shared DeleteOptions options = 2 field holding optional version and uid preconditions. The DeleteOptions message does not exist in pkg/proto/ateapipb/ateapi.proto, and none of the three Delete requests carry it.

PR #1417, opened 2026-09-02
Garbage Collect snapshots and remove the snapshot resource

### LINKED ISSUE #664: Add retention policy and garbage collection for ActorSnapshots

laoj2:
Fixes #664 

This PR implements the idea described in https://github.com/agent-substrate/substrate/issues/664#issuecomment-5499311489

It does more than Garbage Collection of snapshots, because we also got rid of the Snapshot resource (from the DB/API).

Now, an external snapshot is owned by a single resource:

- An Actor owns the snapshot it writes at suspend
- A tag owns a copy taken at tag creation,
- An actor cloned from a tag borrows the tag's snapshot until its own first suspend.

Garbage Collection: whoever created/owns the snapshot is the only one who ever deletes them: 
i.e., if an actor is deleted and it owns a snapshot. The underlying snapshot is deleted with the actor. 

this PR:

- Drops table actor_snapshots
- Keeps table actor_snapshot_tags 
- Adds an object copy at tag creation, and an owned versus borrowed distinction on the Actor
- Adds synchronous external snapshot deletion at actor suspend, at actor delete, and at tag delete 

- [x] Tests pass


cmd/ateapi/internal/store/atepg/migrations/000001_initial.sql
```diff
@@ -52,34 +52,17 @@ CREATE TABLE actor_templates (
...
-CREATE TABLE actor_snapshot_tags (
-    atespace           text NOT NULL,
-    name               text NOT NULL,
-    snapshot_atespace  text NOT NULL,
-    snapshot_name      text NOT NULL,
-    uid                text NOT NULL,
-    version            bigint NOT NULL,
-    proto              bytea NOT NULL,
     PRIMARY KEY (atespace, name),
     CONSTRAINT actor_snapshot_tags_atespace_fk
```
juli4n, line 62: Not for this PR, but I think we should use consistent syntax to define the foreign key on the parent table, as well as use consistent names for the constraint name.

internal/resources/snapshot.go
```diff
@@ -62,6 +62,19 @@ func NewSnapshotURI(location, atespace, name string) (SnapshotURI, error) {
 	return SnapshotURI{uri: uri, location: location, atespace: atespace, name: name}, nil
 }
 
+// tagSnapshotPrefix separates the snapshots ActorSnapshotTags own from the
+// ones actors own. Actor snapshots are named by NewSnapshotName, which returns
+// a UUID, so no actor can ever be handed a name a tag holds.
+const tagSnapshotPrefix = "tag-"
```
juli4n: I wonder if we should model the ownership model in the GCS hierarchy. 

A regular snapshot is owned by an actor, so they could be under:

```
<location>/atespaces/<atespace>/actors/<actor-UID>/snapshots/<snapshot-uid>
```

When you create a tag, the snapshot is owned by the `ActorSnapshotTag` resource, so they could be under:

```
<location>/atespaces/<atespace>/actor-snapshot-tags/<snapshot-uid>
```

I suspect this might simplify the deletion as we can compute the base path for the resource and delete everything under it?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -517,24 +529,20 @@ message ActorStatus {
...
+  // external_snapshot is the Actor's current external snapshot.
+  // If the actor was created from an ActorSnapshotTag:
+  //   * This is the tag's snapshot AND
+  //   * current_snapshot_tag is set
+  // Otherwise: this field is unset until the Actor is first suspended
   //
   // +k8s:optional
-  // +k8s:subfield(atespace)=+k8s:required
-  ObjectRef latest_snapshot = 4;
+  ExternalSnapshot external_snapshot = 4;
```
juli4n, line 544: Should `current_snapshot_tag` be under `external_snapshot`? I know `ExternalSnapshot` is being used in other places, maybe we could have an `ExternalSnapshotInfo` (to match `LocalSnapshotInfo`) that has the fields from ExternalSnapshot and `current_snapshot_tag`?

The `ActorStatus` message is getting a lot of fields related to snapshots so if we can do better grouping, better.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -634,35 +623,47 @@ message WorkerAssignment {
...
+  // UID of the ActorTemplate the snapshot was taken under. Actors can only be
+  // seeded from a tag under the same template.
+  string actor_template_uid = 2;
+
+  // in_progress_snapshot_uri is where CreateActorSnapshotTag is copying the
+  // tag's own snapshot to. It is minted before a single object is written and
+  // cleared once snapshot is set, so a tag left with it set names exactly the
+  // objects its unfinished create stranded, and deleting the tag collects
+  // them.
+  string in_progress_snapshot_uri = 3;
```
EItanya, line 654: This feels a little bit oddly specific. What about modeling this as a state enum?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1402,30 +1402,44 @@ message DeleteActorEgressPolicyRequest {
...
+// the copy for the same Actor; a name held by a finished tag, or by a pending
+// tag of a different Actor, is ALREADY_EXISTS.
 message CreateActorSnapshotTagRequest {
-  // The tag to create.
-  ActorSnapshotTag actor_snapshot_tag = 1;
+  // The suspended Actor whose external snapshot to tag. The tag is created in
+  // this Actor's Atespace.
+  //
+  // +k8s:opaqueType
+  ObjectRef actor = 1;
```
EItanya: I understand this is not for now, but do we want to at least leave room in the API for actors having > 1 snapshot in their histrory

manifests/ate-install/ate-api-server.yaml
```diff
@@ -119,6 +119,11 @@ spec:
               fieldPath: metadata.uid
         - name: OTEL_RESOURCE_ATTRIBUTES
           value: k8s.namespace.name=$(POD_NAMESPACE),k8s.pod.name=$(POD_NAME),k8s.pod.uid=$(POD_UID),service.instance.id=$(POD_UID)
+        # ate-api copies external snapshots for tags and deletes the ones
+        # nothing refers to any more, so it needs the same backend atelet reads
```
EItanya, line 123: This comment is not so much about this line of code, but rather conceptually allowing more component access to storage, is that something we want to do? Can we somehow make ate-api readonly for the actual actor buckets so it can't mess with those

internal/resources/snapshot.go
EItanya: I was thinking the same

EItanya (commented):
So this PR is quite large so gonna leave an initial review. My biggest comment is about the storage owner. There's a lot of added code related specifically to giving ate-api access to the storage buckets but I would like to think through that, it may be necessary, but worth discussing.

Here are some correctness issues codex discovered, seem correct on first review from me:

```
## Review findings

### [P1] Serialize tag deletion with an active copy

`DeleteActorSnapshotTag` can list and delete a partial prefix, then remove the tag row while
`TagActorSnapshot` is still copying objects. Objects written after that listing become permanently
orphaned because the row holding their URI is gone.

Please serialize creation and deletion using a tag-specific lease.

### [P2] Delete the fetched tag instance, not just its name

The handler fetches a tag and cleans its snapshot, but then deletes by `atespace/name` without
checking the fetched UID or version.

With concurrent deletes and name reuse, a stale request can delete the replacement tag while
leaving its newly minted snapshot behind. Please make the store deletion conditional on the
fetched tag UID.

### [P2] Abort multipart uploads when completion fails

`copyMultipart` aborts after a part-copy failure, but returns directly when
`CompleteMultipartUpload` fails. S3 retains those uploaded parts and continues billing for them.

Please route completion failures through the same abort cleanup.
```

manifests/ate-install/ate-api-server.yaml
laoj2: It also means that the var value should match between the components (maybe obvious, though).

>  Can we somehow make ate-api readonly for the actual actor buckets so it can't mess with those

Right now, ate-api needs write access to delete buckets. If we move this functionality to atelet, we'll need to make sure an atelet is called during actor deletion. Today, this is not always the case. 

If a suspended actor is deleted, we skip calling atelet because there's no worker assignment (so the control plane doesn't know which atelet to call and skips that). This is similar to the issue with GCing local snapshots too that I wrote in: https://github.com/agent-substrate/substrate/issues/664#issuecomment-5441337112

So, we can consider moving the external snapshot deletion to atelet after that other problem is fixed, WDYT?

EItanya: > It also means that the var value should match between the components (maybe obvious, though).

Seems like we need a config struct somewhere that's shared at the very least, probably need that anway.

> So, we can consider moving the external snapshot deletion to atelet after that other problem is fixed, WDYT?

I'm not sold either way, but definitely worth bikeshedding @juli4n and @thockin

pkg/proto/ateapipb/ateapi.proto
laoj2: We unfortunately need the in progress uri. The current sequence of tagging a snapshot is:

1. tx1: write in_progress_snapshot_uri to the desired URI
2. Copy snapshot to the URI
3. tx2: unset in_progress_snapshot_uri, and set external_snapshot.

So, if we fail in between 2 and 3 for some reason, we can recover from where it was left (and delete the partially copied files). If we only have a state enum we lose that ability

This idea was borrowed from what actor does too for both external and local snapshots: https://github.com/agent-substrate/substrate/blob/a9c1bd389403af59520e419d3884d447cf40edd2/pkg/proto/ateapipb/ateapi.proto#L513-L518

internal/resources/snapshot.go
laoj2: Done. This also means that ownership can be inferred from the external storage uri, so current_snapshot_tag is no longer needed (it was only used to determine whether a snapshot belongs to a tag or not).

pkg/proto/ateapipb/ateapi.proto
laoj2: current_snapshot_tag is now gone from the api

internal/resources/snapshot.go
```diff
@@ -22,44 +22,151 @@ import (
...
-//	gs://bucket/root                              location
-//	gs://bucket/root/snapshots/team-a/<name>      this URI
-//	gs://bucket/root/snapshots/team-a/<name>/...  an object in the snapshot
-type SnapshotURI struct {
-	uri      string
-	location string
+// StoragePrefix is a validated object-storage prefix: a scheme, a bucket and a
+// path, with no query, fragment or userinfo. Every object below this prefix has a name
+// starting with its path, so deleting the prefix collects all of them.
+type StoragePrefix struct {
```
juli4n, line 46: Any reason for this to not be just a string? Do we need a struct field for it?

laoj2:
> So this PR is quite large so gonna leave an initial review. My biggest comment is about the storage owner. There's a lot of added code related specifically to giving ate-api access to the storage buckets but I would like to think through that, it may be necessary, but worth discussing.
> 
> Here are some correctness issues codex discovered, seem correct on first review from me:
> 
> ```
> ## Review findings
> 
> ### [P1] Serialize tag deletion with an active copy
> 
> `DeleteActorSnapshotTag` can list and delete a partial prefix, then remove the tag row while
> `TagActorSnapshot` is still copying objects. Objects written after that listing become permanently
> orphaned because the row holding their URI is gone.
> 
> Please serialize creation and deletion using a tag-specific lease.
> 
> ### [P2] Delete the fetched tag instance, not just its name
> 
> The handler fetches a tag and cleans its snapshot, but then deletes by `atespace/name` without
> checking the fetched UID or version.
> 
> With concurrent deletes and name reuse, a stale request can delete the replacement tag while
> leaving its newly minted snapshot behind. Please make the store deletion conditional on the
> fetched tag UID.
> 
> ### [P2] Abort multipart uploads when completion fails
> 
> `copyMultipart` aborts after a part-copy failure, but returns directly when
> `CompleteMultipartUpload` fails. S3 retains those uploaded parts and continues billing for them.
> 
> Please route completion failures through the same abort cleanup.
> ```

Thanks! I addressed [P1] Serialize tag deletion with an active copy and [P2] Abort multipart uploads when completion fails in my latest 2 commits.

For ` [P2] Delete the fetched tag instance, not just its name`, we have this problem across all Delete RPCs today. I left this out purposely (to not make this PR even larger), so we can fix that as part of https://github.com/agent-substrate/substrate/issues/866

internal/resources/snapshot.go
laoj2: This could be a string. The nice part about having a type is that it increases the likelihood we'll generate a storagePrefix using the helper functions in this package. But, I dont have a strong opinion here and can revert this back to string

manifests/ate-install/ate-api-server.yaml
BenTheElder: I think we will wind up wanting to do our own central substrate aware object store authz so we can build something scalable but ~vendor neutral for consistency. https://github.com/agent-substrate/substrate/issues/777

Atelet has waaay too much permission right now. It would be better if we could limit it to the actors it hosts.

Central access seems ok.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -129,6 +130,17 @@ service Control {
...
 
+// ExternalSnapshot addresses a snapshot held in object storage.
+message ExternalSnapshot {
+  // snapshot_uri addresses the snapshot's prefix in object storage.
+  // TODO: Add DV (required, maxLength?)
+  string snapshot_uri = 1;
+
+  // content_scope is what the snapshot captured.
+  // TODO: Add DV (optional, minimum, maximum?)
+  SnapshotContentScope content_scope = 2;
```
juli4n, line 144: // +k8s:minimum=1
// +k8s:maximum=2 # keep this in sync with the SnapshotContentScope enum

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1423,30 +1410,44 @@ message DeleteActorEgressPolicyRequest {
...
+// the copy for the same Actor; a name held by a finished tag, or by a pending
+// tag of a different Actor, is ALREADY_EXISTS.
 message CreateActorSnapshotTagRequest {
-  // The tag to create.
-  ActorSnapshotTag actor_snapshot_tag = 1;
+  // The suspended Actor whose external snapshot to tag. The tag is created in
+  // this Actor's Atespace.
+  //
+  // +k8s:opaqueType
+  ObjectRef actor = 1;
```
juli4n: I think we should move this field to be part of the ActorSnapshotTag resource (e.g. `ObjectRef source_actor`). This would be consistent with `CreateActor` where you specify the "source_tag" in the resource.

cmd/ateapi/internal/controlapi/actor_snapshot_tag.go
```diff
@@ -47,40 +48,63 @@ var actorSnapshotTagScopeNames = func() []string {
...
-func (s *ServiceImpl) GetActorSnapshot(ctx context.Context, snapshotRef resources.ActorSnapshotRef) (*ateapipb.ActorSnapshot, error) {
+func (s *ServiceImpl) CreateActorSnapshotTag(ctx context.Context, tag *ateapipb.ActorSnapshotTag) (*ateapipb.ActorSnapshotTag, error) {
 	// TODO: implement this
-	return s.store.GetActorSnapshot(ctx, snapshotRef)
+	return s.store.CreateActorSnapshotTag(ctx, tag)
 }
 
-func validateGetActorSnapshotRequest(req *ateapipb.GetActorSnapshotRequest) field.ErrorList {
+func validateCreateActorSnapshotTagRequest(req *ateapipb.CreateActorSnapshotTagRequest) field.ErrorList {
 	var fldPath *field.Path
```
juli4n, line 78: I know this PR is doing *a lot* so we can this as a follow up PR, but can you add TODO comments to replace all the validations here with DV tags?

cmd/ateapi/internal/controlapi/actor_snapshot_tag.go
```diff
@@ -264,20 +250,80 @@ func validateUpdateActorSnapshotTagRequest(req *ateapipb.UpdateActorSnapshotTagR
...
 }
 
+// releaseTagSnapshot deletes the objects the tag's external snapshot is made
+// of. It tolerates a partly-collected snapshot, so a retry finishes cleanly.
+// It collects the in-progress snapshot too.
+func (s *RPCService) releaseTagSnapshot(ctx context.Context, tag *ateapipb.ActorSnapshotTag) error {
+	if s.objectStore == nil {
+		return nil
+	}
+	atespace, name := tag.GetMetadata().GetAtespace(), tag.GetMetadata().GetName()
```
juli4n, line 336: Nit: you can just do:

```
tagRef := resources.ActorSnapshotTagRefFromActorSnapshotTag(tag)
```

and then use it from the error message.

cmd/ateapi/internal/controlapi/actor_snapshot_tag.go
```diff
@@ -264,20 +250,80 @@ func validateUpdateActorSnapshotTagRequest(req *ateapipb.UpdateActorSnapshotTagR
...
+// and resumes over whatever is left.
+//
+// The tag stays resolvable while its snapshot is being collected, so a
+// CreateActor racing this delete can seed an Actor from content that is going
+// away. That race is accepted for now.
+//
+// Note that this destroys the external snapshot: an Actor created from the tag
+// and never suspended is still borrowing it and becomes unrecoverable. Do not
+// delete a tag while clones of it exist.
 func (s *RPCService) DeleteActorSnapshotTag(ctx context.Context, req *ateapipb.DeleteActorSnapshotTagRequest) (*ateapipb.ActorSnapshotTag, error) {
```
juli4n, line 293: It looks like the delete orchestration should be part of a workflow? I'm ok with a TODO to move it later though.

laoj2: Done

pkg/proto/ateapipb/ateapi.proto
laoj2: Good point. Moved in the latest commit

cmd/ateapi/internal/controlapi/actor_snapshot_tag.go
EItanya: Should an actor reusing a tag mirror it? Definitely not sold on that but thinking out loud.

laoj2: Yes, that's an interesting option to explore. We need to run benchmarks to see if this won't be too slow or if it's acceptable to have N copies of the same snapshot (maybe this is fine? in the next Suspend() the snapshot will be replaced anyway with the actor's own snapshot). Left this decision out of the PR on purpose, given it's gigantic, but we should follow up on this.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -652,36 +632,61 @@ message WorkerAssignment {
...
-  // +k8s:opaqueType
-  ObjectRef snapshot = 2;
+
+  // status is set by the control plane; it is ignored on write.
+  //
+  // +k8s:optional
+  ActorSnapshotTagStatus status = 2;
+
+  // scope controls who may create an Actor from this tag.
   ActorSnapshotTagScope scope = 3;
```
juli4n, line 683: // +k8s:required
// +k8s:minimum=1
// +k8s:maximum=2 # keep this in sync with the ActorSnapshotTagScope enum

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -652,36 +632,61 @@ message WorkerAssignment {
...
-// ActorSnapshotTag is an immutable, Atespace-owned alias and retention pin.
-// Its owning Atespace cannot be deleted until the tag is removed.
+// ActorSnapshotTag is an Atespace-owned, stable name for one external
+// snapshot. The external snapshot pointed by a tag is a copy of an actor snapshot.
+// The snapshot is copied when the tag is
+// created, and deleted when the tag is. Its owning Atespace cannot be deleted
+// until the tag is removed.
 message ActorSnapshotTag {
   // Common resource metadata: name, uid, version, timestamps.
   // +k8s:opaqueType
```
juli4n: // +k8s:required
// +k8s:subfield(atespace)=+k8s:required

cmd/ateapi/internal/controlapi/workflow_snapshot_tag.go
```diff
@@ -0,0 +1,267 @@
...
+		return nil, "", fmt.Errorf("while reserving actor snapshot tag %s: %w", tagRef, err)
+	}
+
+	// The tag already exists. This means one of the two are true:
+	//   * We tried to write this tag before for this actor & snapshot,
+	//      but failed and this is a retry
+	//   * This tag exists for another actor/snapshot. We should fail right away,
+	//     because a tag is immutable.
+	// We read the tag back to figure out which is our situation.
+	existing, err := w.store.GetActorSnapshotTag(ctx, tagRef)
```
juli4n: I'm worried that the logic to make this retriable / recoverable is very brittle. I wonder if we can just say that a failed CreateActorSnapshotTag cannot be retried (at least for now). It failed, you need to delete and create it again. That simplifies this significantly.

laoj2: Removed it. This is way simpler now.

juli4n (approved):

PR #1419, opened 2026-09-02
imagecache: tune pull retry backoff per registry class

dberkov:
## Summary

go-containerregistry retries 429s and 5xxs by default, but with a backoff tuned for a single flaky client: 3 attempts over ~4s with 0.1 jitter. When a whole fleet is throttled at once (registry.k8s.io rate-limited a 1000-node benchmark's pause-image pulls), every node retries nearly in lockstep and gives up before the limiter cools down, so the retries just replay the stampede.

This replaces the default with two profiles, split by **whose rate limit the pull is spending**:

- **Shared (communal) registries** — registry.k8s.io, Docker Hub, quay.io, ghcr.io, and other well-known anonymous-pull hosts. Their rate limits are shared across all clients, so a fleet retrying in lockstep just replays the stampede. These get a longer, fully-jittered backoff (4 attempts, up to ~14s) that outlasts a throttling wave and de-synchronizes the herd. The motivating case is registry.k8s.io, the CNCF-hosted registry that serves the pause image: it throttles aggressively, so a fleet retrying on the default schedule gives up before the rate limiter cools down.
- **Dedicated registries (the default)** — everything else: GCP Artifact Registry, AWS ECR, Harbor, self-hosted, local dev registries. The deployment owns the quota, the server absorbs retry bursts, and pulls sit on the Run/Restore critical path, so retries start at 200ms with 4 attempts capped at 2s — more attempts in less total wall clock, still failing fast enough to surface a real outage to the RPC-level retry.

Defaulting to the dedicated profile (rather than allowlisting private registries) is deliberate: Substrate is open source and customers run many registry vendors, but fleet-scale pulls realistically hit either the pause image or the customer's own registry. The small, stable set of communal hosts is enumerable; the set of private registry vendors is not.

## Test plan

- `TestRetryBackoffFor` pins the split across GCP, ECR, Harbor, local, and communal registry hosts.
- `TestEnsureImage_RetriesRateLimit` proves a pull survives transient 429s end to end through the retry transport.

### DIFF
--- a/internal/imagecache/imagecache.go
+++ b/internal/imagecache/imagecache.go
@@ -96,6 +96,67 @@ const (
 	layerPullConcurrency = 4
 )
 
+// Pull retry backoffs, chosen per registry by retryBackoffFor. Both replace
+// go-containerregistry's default (3 attempts over ~4s, jitter 0.1), which is
+// tuned for a single flaky client, and both apply to every request of a pull,
+// including the /v2/ auth ping. The retryable status codes stay at the
+// library default, which already includes 429. Vars so tests can shrink the
+// waits.
+var (
+	// dedicatedRegistryBackoff covers registries where the deployment has
+	// its own quota (Artifact Registry, ECR, Harbor, self-hosted, …): the
+	// server absorbs retry bursts and nobody else competes for the limit,
+	// and pulls sit on the Run/Restore critical path, so retries start fast
+	// and come often — more attempts in less total wall clock, still quick
+	// enough to surface a real outage to the RPC-level retry.
+	dedicatedRegistryBackoff = remote.Backoff{
+		Duration: 200 * time.Millisecond,
+		Factor:   2.0,
+		Jitter:   1.0,
+		Steps:    4,
+		Cap:      2 * time.Second,
+	}
+	// sharedRegistryBackoff covers communal registries (registry.k8s.io,
+	// Docker Hub, …), where pulls are anonymous and rate limits are shared:
+	// a whole fleet can be throttled at once, so retries must outlast the
+	// throttling wave and full jitter must de-synchronize the herd rather
+	// than replay it.
+	sharedRegistryBackoff = remote.Backoff{
+		Duration: 1 * time.Second,
+		Factor:   2.0,
+		Jitter:   1.0,
+		Steps:    4,
+		Cap:      10 * time.Second,
+	}
+)
+
+// sharedRegistries are the well-known communal registries whose rate limits
+// are shared across all anonymous clients. Everything not listed here is
+// assumed dedicated: fleet-scale pulls realistically hit either the pause
+// image (registry.k8s.io) or the customer's own registry, and enumerating
+// the small stable set of communal hosts beats guessing at every private
+// registry vendor.
+var sharedRegistries = map[string]bool{
+	"registry.k8s.io":      true,
+	"docker.io":            true,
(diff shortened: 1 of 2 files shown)


maxsmythe (approved):
LGTM

internal/imagecache/imagecache_test.go
```diff
@@ -91,6 +94,72 @@ func newTestStore(t *testing.T, opts ...Option) *Store {
...
+		"ghcr.io":                             sharedRegistryBackoff,
+	} {
+		if got := retryBackoffFor(registry); got != want {
+			t.Errorf("retryBackoffFor(%q) = %+v, want %+v", registry, got, want)
+		}
+	}
+}
+
+// TestEnsureImage_RetriesRateLimit proves a pull survives transient 429s.
+// go-containerregistry's retry transport, configured with pullRetryBackoff,
```
igooch, line 122: Stale identifier in comment: `pullRetryBackoff` no longer exists — this PR split it into `dedicatedRegistryBackoff` and `sharedRegistryBackoff`. The comment will mislead the next reader grepping for the name. Suggest referencing the two current vars (or "the per-registry backoff").

internal/imagecache/imagecache.go
```diff
@@ -96,6 +96,67 @@ const (
...
+
+// sharedRegistries are the well-known communal registries whose rate limits
+// are shared across all anonymous clients. Everything not listed here is
+// assumed dedicated: fleet-scale pulls realistically hit either the pause
+// image (registry.k8s.io) or the customer's own registry, and enumerating
+// the small stable set of communal hosts beats guessing at every private
+// registry vendor.
+var sharedRegistries = map[string]bool{
+	"registry.k8s.io":      true,
+	"docker.io":            true,
```
igooch, line 141: This `"docker.io"` entry is unreachable in production: `retryBackoffFor` is only ever called with `parsedRef.Context().Registry.RegistryStr()`, which normalizes `docker.io` (and bare refs like `ubuntu:latest`) to `index.docker.io` — verified against go-containerregistry v0.21.7. The `index.docker.io` and `registry-1.docker.io` entries do the real work; `docker.io` only matters because `TestRetryBackoffFor` calls `retryBackoffFor("docker.io")` directly. Harmless defensiveness, but worth a note or dropping to avoid implying `docker.io` is ever seen here.

internal/imagecache/imagecache.go
```diff
@@ -96,6 +96,67 @@ const (
...
+		Jitter:   1.0,
+		Steps:    4,
+		Cap:      2 * time.Second,
+	}
+	// sharedRegistryBackoff covers communal registries (registry.k8s.io,
+	// Docker Hub, …), where pulls are anonymous and rate limits are shared:
+	// a whole fleet can be throttled at once, so retries must outlast the
+	// throttling wave and full jitter must de-synchronize the herd rather
+	// than replay it.
+	sharedRegistryBackoff = remote.Backoff{
```
igooch, line 124: Worth confirming: `sharedRegistryBackoff` adds up to ~14s (1+2+... jittered, Cap 10s, 4 steps) *per request*, and the retry transport applies it to every request of a pull independently (ping + manifest + each blob). Under a sustained throttling wave a multi-layer shared-registry pull can therefore accumulate far beyond 14s. That is intended for the pause-image warm path, but make sure no shared-registry pull sits behind a bounded RPC/router deadline where this would turn a throttle into a hang rather than a fast surfaced failure.

PR #1421, opened 2026-09-02
podcertcontroller: add flags to raise client-go rate limits

dberkov:
## Summary

The controller built its Kubernetes client from the default rest.Config, whose client-side limiter (QPS=5, Burst=10) is shared by every worker of both signers. Certificate issuance was therefore capped at roughly 5 PodCertificateRequests per second no matter the `--workers-per-signer` value, and pods in a large rollout sat in FailedMount ("credential bundle is not issued yet") for the duration: about 2000 seconds for a 10k-pod worker pool in the 1000-node load test.

This adds `--kube-api-qps` and `--kube-api-burst` and applies them to the config on both the in-cluster and kubeconfig paths. Both default to 0, which leaves client-go's own defaults untouched — behavior only changes where a deployment opts in. A follow-up PR will set values sized for large rollouts in `manifests/ate-install/pod-certificate-controller.yaml`.

## Test plan

- `go build` / `go vet` clean.
- Verified against the 1000-node load test (with raised limits set via the flags): bundle issuance keeps pace with pod creation instead of plateauing at ~4.5/sec.

### DIFF
--- a/cmd/podcertcontroller/main.go
+++ b/cmd/podcertcontroller/main.go
@@ -80,6 +80,17 @@ var (
 		"Number of concurrent worker goroutines per signer.",
 	)
 
+	kubeAPIQPS = pflag.Float32(
+		"kube-api-qps",
+		0,
+		"Sustained queries per second allowed against the Kubernetes API. 0 keeps the client-go default.",
+	)
+	kubeAPIBurst = pflag.Int(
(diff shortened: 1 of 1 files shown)


ahmedtd (approved):

maxsmythe (approved):
LGTM

PR #1423, opened 2026-09-02
Record the SandboxConfigRef in snapshot storage instead of sandbox assets

zoez7:
## Summary

This change makes durable snapshot storage record a reference to the SandboxConfig object instead:

- The snapshot manifest carries only the SandboxConfig reference (name, UID, resourceVersion) — never asset content.
- Checkpoint reports the reference from the actor's on-node sandbox record, and the pause/suspend finalizers stamp it onto `LocalSnapshotInfo` / `ActorSnapshot`.
- Resume resolves restore assets from the reference recorded on the snapshot whose sandbox runs the guest, erroring with `FailedPrecondition` when the SandboxConfig is gone or was re-created under the same name. Older snapshots without a reference fall back to the pool's config. `RestoreRequest` now requires `sandbox_assets`, since the manifest is no longer a source of binaries.
- atelet's node-local flow is otherwise unchanged; Restore cross-checks the manifest's reference against the request's assets as defense in depth.

Also relaxes `UpdateActor` template validation to compare volumes by name rather than list order, since volumes are referenced by name and order carries no meaning.

## Test plan

- Unit tests for the new reference plumbing in ateapi (pause/suspend/resume) and atelet (Run/Restore/Checkpoint record round-trip).
- New `updatesandboxconfig` e2e suite asserting SandboxConfig provenance on the snapshots that suspends produce, including the repointed-actor restore path.


BenTheElder:
> atelet's node-local flow is otherwise unchanged; Restore cross-checks the manifest's reference against the request's assets as defense in depth.

the manifest fetch is a bit of a latency limiter right now, I think we need to start planning to drop this

otherwise agree directionally

PR #1425, opened 2026-09-03
imagecache: clarify pull retry backoff comments

dberkov:
## Summary

Comment-only follow-ups to the review comments on #1419 (https://github.com/agent-substrate/substrate/pull/1419#pullrequestreview-comments):

- **Stale identifier** ([comment](https://github.com/agent-substrate/substrate/pull/1419#discussion_r3919918156)): the retry test's doc comment still referenced `pullRetryBackoff`, which that PR split into `dedicatedRegistryBackoff` and `sharedRegistryBackoff`. It now points at the per-registry backoff picked by `retryBackoffFor`.
- **Unreachable `docker.io` entry** ([comment](https://github.com/agent-substrate/substrate/pull/1419#discussion_r3919918239)): kept the entry (harmless, and a lookup by the canonical name should classify the same way), but the map comment now states plainly that production lookups only ever see `index.docker.io` because `name.ParseReference` normalizes `docker.io` and bare refs before `retryBackoffFor` runs.
- **Backoff accumulation vs. deadlines** ([comment](https://github.com/agent-substrate/substrate/pull/1419#discussion_r3919918313)): confirmed the concern is bounded and documented why. The backoff is per request, so a multi-layer pull under a sustained throttle can accumulate well past one request's ~14s worst case — but every request carries the caller ctx (`remote.WithContext`), go-containerregistry's retry transport never retries a context error, so retrying overshoots a deadline by at most one backoff sleep. Production pulls run under the Run/Restore RPC ctx, which ateapi caps at `maxRPCDeadline` (10m); the router's parking budget deliberately defers to that server-side deadline. A throttled shared-registry pull therefore surfaces as the RPC's deadline error, not a hang.

## Test plan

Comment-only change: `go test ./internal/imagecache/`, `go vet`, and `gofmt` all pass.

### DIFF
--- a/internal/imagecache/imagecache.go
+++ b/internal/imagecache/imagecache.go
@@ -102,6 +102,16 @@ const (
 // including the /v2/ auth ping. The retryable status codes stay at the
 // library default, which already includes 429. Vars so tests can shrink the
 // waits.
+//
+// The backoff applies per request, so under a sustained throttle a
+// multi-layer pull can accumulate far more wall clock than one request's
+// worst case (~14s shared). The caller's ctx still bounds the total: every
+// request carries it (remote.WithContext in remoteOpts), a canceled ctx
+// fails the next attempt immediately and is never retried, so retrying
+// overshoots a deadline by at most one backoff sleep. Production pulls run
+// under the Run/Restore RPC ctx, which ateapi's resume path caps at its
+// server-wide max RPC deadline — a throttled shared-registry pull surfaces
+// as that RPC's deadline error, not a hang.
 var (
 	// dedicatedRegistryBackoff covers registries where the deployment has
 	// its own quota (Artifact Registry, ECR, Harbor, self-hosted, …): the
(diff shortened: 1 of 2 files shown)


internal/imagecache/imagecache.go
```diff
@@ -137,9 +147,13 @@ var (
 // the small stable set of communal hosts beats guessing at every private
 // registry vendor.
 var sharedRegistries = map[string]bool{
-	"registry.k8s.io":      true,
+	"registry.k8s.io": true,
+	// Pulls only ever present index.docker.io here: name.ParseReference
+	// normalizes docker.io (and bare refs like "ubuntu") to it before
```
igooch: Minor: `docker.io` isn't the canonical name here — `RegistryStr()` returns `index.docker.io`, which is the canonical form this map is actually keyed by in production. Consider "kept so a direct lookup by the short name classifies the same way" to avoid implying `docker.io` is canonical. Non-blocking.

igooch (approved):
/lgtm
/approve

PR #1426, opened 2026-09-03
atelet: constrain snapshot files to checkpoint directories

dubin555:
## Summary

- validate snapshot file names from manifests and ateom checkpoint responses before using them as relative paths
- confine checkpoint upload, local snapshot movement, local restore copies, and external restore writes with `os.Root`
- pass already-open files to the compression helpers so path validation cannot be bypassed by reopening a symlink
- preserve optional empty results for DATA checkpoints that do not contain runtime snapshot files

## Testing

- `go test -race ./cmd/atelet ./cmd/atelet/internal/ategcs -count=1`
- focused symlink confinement tests repeated with the race detector
- `hack/verify-all.sh`

### DIFF
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -654,9 +654,9 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 		return nil, fmt.Errorf("while calling ateom.CheckpointWorkload: %w", err)
 	}
 
-	sandboxRec.SnapshotFiles = resp.GetSnapshotFiles()
-	if len(sandboxRec.SnapshotFiles) == 0 && shouldHaveSnapshots(req) {
-		return nil, ateerrors.NewGRPCError(ctx, codes.DataLoss, ateerrors.ReasonInvalidCheckpointResult, ateerrors.ActorCrashedMetadata(), errors.New("ateom reported no snapshot files for checkpoint"))
+	sandboxRec.SnapshotFiles, err = checkpointSnapshotFiles(resp, shouldHaveSnapshots(req))
+	if err != nil {
+		return nil, ateerrors.NewGRPCError(ctx, codes.DataLoss, ateerrors.ReasonInvalidCheckpointResult, ateerrors.ActorCrashedMetadata(), err)
 	}
 	sandboxRec.Atespace = req.GetAtespace()
 	sandboxRec.ActorName = req.GetActorName()
@@ -708,6 +708,17 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 	}
 
 	return &ateletpb.CheckpointResponse{}, nil
+}
+
+func checkpointSnapshotFiles(resp *ateompb.CheckpointWorkloadResponse, required bool) ([]string, error) {
+	files := resp.GetSnapshotFiles()
+	if len(files) == 0 && required {
+		return nil, errors.New("ateom reported no snapshot files for checkpoint")
+	}
+	if err := validateSnapshotFiles(files); err != nil {
+		return nil, fmt.Errorf("ateom reported invalid snapshot files: %w", err)
+	}
+	return files, nil
 }
 
 func toAteomSnapshotScope(scope ateletpb.SnapshotScope) ateompb.SnapshotScope {
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ -555,20 +555,11 @@ func initSnapshotSizeMetric() error {
 // recordSnapshotSize labels each image with the registry's file.name. That
 // label used to be spelled "kind", which means the snapshot's provenance
 // everywhere else in the ate.* namespace, not one of its files.
-func recordSnapshotSize(ctx context.Context, file, path, templateAtespace, templateName string) {
+func recordSnapshotSize(ctx context.Context, file string, size int64, templateAtespace, templateName string) {
 	if snapshotSizeBytes == nil {
 		return
 	}
-	fi, err := os.Stat(path)
-	if errors.Is(err, os.ErrNotExist) {
-		return
-	}
-	if err != nil {
-		slog.WarnContext(ctx, "Failed to stat snapshot image for size metric",
-			slog.String("file", file), slog.String("path", path), slog.Any("err", err))
-		return
-	}
-	snapshotSizeBytes.Record(ctx, fi.Size(), metric.WithAttributes(
+	snapshotSizeBytes.Record(ctx, size, metric.WithAttributes(
 		semconv.FileNameKey.String(file),
 		ateattr.TemplateAtespaceKey.String(templateAtespace),
 		ateattr.TemplateNameKey.String(templateName),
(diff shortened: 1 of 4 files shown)


BenTheElder:
cc @zoez7

BenTheElder (commented):
Does this actually solve the problem? If an ateom is specifying invalid/evil paths, why wouldn't it just do something like `checkpoint.img -> /var/run/secrets/kubernetes.io/serviceaccount/token`

I think we need to be using os.Root like the rest of the defenses.

EItanya:
> Does this actually solve the problem? If an ateom is specifying invalid/evil paths, why wouldn't it just do something like `checkpoint.img -> /var/run/secrets/kubernetes.io/serviceaccount/token`
> 
> I think we need to be using os.Root like the rest of the defenses.

Agree with Ben, I think we can fairly easily add defensive tests for this.

dubin555:
Thanks, agreed. `filepath.IsLocal` only rejects lexical traversal and does not stop a symlink from escaping the checkpoint directory.

I updated the PR so the actual file operations are confined as well:

- checkpoint uploads open source files through `os.Root`
- external restores create destination files through `os.Root`
- local restore copies use separate source and destination roots
- local checkpoint moves stay within the Actor root
- compression helpers now receive already-open files, so they cannot bypass the root by reopening a path

The lexical validation remains because the same names are also used to construct object-storage keys.

The new tests verify that an upload source symlink, an external restore destination symlink, and a local restore source symlink cannot access files outside their respective roots. The package race tests and the full verification suite pass.

cmd/atelet/sandbox_assets.go
```diff
@@ -493,9 +493,24 @@ func unmarshalSandboxRecord(data []byte) (*sandboxAssetsRecord, error) {
...
+	if err := validateSnapshotFiles(rec.SnapshotFiles); err != nil {
+		return nil, fmt.Errorf("%w: sandbox record/manifest has invalid snapshotFiles: %w", ateerrors.ReasonInvalidSandboxAsset, err)
+	}
 	return rec, nil
 }
 
+// validateSnapshotFiles rejects names that are not lexically local to a
+// checkpoint directory. Actual file access must still use os.Root so symlinks
+// cannot escape that directory.
+func validateSnapshotFiles(files []string) error {
```
BenTheElder, line 505: 🤖 **should-fix** 🟡 – `manifest.json` is still an accepted snapshot file name, and `os.Root` cannot help here: the collision is with a legitimate file in the same directory, not an escape from it.

If ateom reports `snapshot_files ["checkpoint.img","manifest.json"]`, `moveLocalCheckpoint` renames it into the local snapshot dir and the `root.WriteFile` below truncates it with the marshaled atelet record. No error. A later LOCAL restore then hands ateom that JSON where its snapshot file should be, and the real file is gone.

Reject `name == sandboxManifestName` here.

cmd/atelet/sandbox_assets.go
```diff
@@ -493,9 +493,24 @@ func unmarshalSandboxRecord(data []byte) (*sandboxAssetsRecord, error) {
...
+	for i, name := range files {
+		if !filepath.IsLocal(name) {
```
BenTheElder, line 507: 🤖 **nit** 🟢 – `filepath.IsLocal` is wider than what either ateom emits (both send `os.ReadDir` base names). It accepts `"."`, `"a/.."`, and nested `"a/b"`.

Less dangerous now that the sinks are rooted, but those names still fail on the external restore path with an error that carries no Reason, so it is retried forever, where `"../x"` crashes the actor. The module idiom is `name == filepath.Base(name) && filepath.IsLocal(name)`, as in `cmd/ateom-gvisor/internal/cdiinject/soname.go`.

cmd/atelet/sandbox_assets.go
```diff
@@ -493,9 +493,24 @@ func unmarshalSandboxRecord(data []byte) (*sandboxAssetsRecord, error) {
...
+			return fmt.Errorf("snapshotFiles[%d] %q is not local to the checkpoint directory", i, name)
+		}
```
BenTheElder, line 509: 🤖 **nit** 🟢 – Names are neither cleaned nor de-duplicated, so `"checkpoint.img"` and `"./checkpoint.img"` both pass and resolve to the same file. `goldenOnlyFiles` compares raw strings, so it does not treat them as shadowing, and a DATA_ON_GOLDEN restore runs two concurrent `O_TRUNC` writes of different payloads into one file with `g.Wait()` returning nil. Rooting the opens does not prevent this.

`filepath.Clean` each name, reject duplicates after cleaning, and compare cleaned names in `goldenOnlyFiles`.

cmd/atelet/main.go
```diff
@@ -654,9 +645,9 @@ func (s *AteomHerder) Checkpoint(ctx context.Context, req *ateletpb.CheckpointRe
 		return nil, fmt.Errorf("while calling ateom.CheckpointWorkload: %w", err)
 	}
 
-	sandboxRec.SnapshotFiles = resp.GetSnapshotFiles()
-	if len(sandboxRec.SnapshotFiles) == 0 && shouldHaveSnapshots(req) {
-		return nil, ateerrors.NewGRPCError(ctx, codes.DataLoss, ateerrors.ReasonInvalidCheckpointResult, ateerrors.ActorCrashedMetadata(), errors.New("ateom reported no snapshot files for checkpoint"))
+	sandboxRec.SnapshotFiles, err = checkpointSnapshotFiles(resp, shouldHaveSnapshots(req))
+	if err != nil {
+		return nil, ateerrors.NewGRPCError(ctx, codes.DataLoss, ateerrors.ReasonInvalidCheckpointResult, ateerrors.ActorCrashedMetadata(), err)
```
BenTheElder, line 650: 🤖 **nit** 🟢 – This failure return does not set `op.failedPhase`, unlike the `ateom.CheckpointWorkload` return just above it. `recordPhases` only tags the total and `op.failedPhase`, so the `ateom_checkpoint` phase sample is recorded as if it succeeded and per-phase dashboards cannot see that the ateom result was what failed.

Set `op.failedPhase = ateattr.SnapshotPhaseAteomCheckpoint` before returning.

cmd/atelet/main_test.go
```diff
@@ -292,6 +293,126 @@ func TestWriteSystemInfoVolume_TrustBundle(t *testing.T) {
...
+func TestCheckpointResponseRejectsNonLocalFile(t *testing.T) {
+	_, err := checkpointSnapshotFiles(&ateompb.CheckpointWorkloadResponse{
+		SnapshotFiles: []string{"../outside"},
+	}, true)
+	if err == nil {
+		t.Fatal("checkpointSnapshotFiles() accepted a path outside the checkpoint directory")
+	}
+}
+
+func TestCheckpointSnapshotFilesAllowsOptionalEmptyResult(t *testing.T) {
```
BenTheElder, line 320: 🤖 **nit** 🟢 – This covers the `required == false` half. The `len(files) == 0 && required` DataLoss branch, and its mapping to `codes.DataLoss` + `ReasonInvalidCheckpointResult` + `ActorCrashedMetadata`, is still asserted by no test.

Dropping `&& required` would let a FULL-scope checkpoint with no files upload as a valid empty snapshot, and the suite would stay green. A table over `(files, required, wantErr)` with rows `{[], true, err}`, `{[], false, ok}`, `{["checkpoint.img"], true, ok}`, `{["../outside"], true, err}` folds this and the neighbouring case into one.

BenTheElder (commented):
AI found a few more things but I think they can be follow-ups.

Sorry for the delay, I've been catching up since being out a bit, this will need a rebase.

dubin555:
Thanks for the review. I've rebased onto the latest main and resolved the conflicts.

If you agree, I can open an issue to track the remaining filename-validation and test-coverage suggestions, clarify the scope there, and follow up with an implementation.

Some of the suggestions seem reasonable, though the additional checks might be a little premature at this stage. For example, the current ateom implementations do not emit `manifest.json` as a checkpoint artifact under normal operation. I'd appreciate your view on whether reserving that name defensively is something we want to enforce now.

BenTheElder:
Agreed, I should've spent more time reviewing the agent suggestions here. We've been hard pressed.

If you're in the CNCF slack, send me a ping next time you rebase so I can try to get this in ASAP.

Github notifications here are high noise : signal.

BenTheElder (approved):

PR #1429, opened 2026-09-03
Avoid Istio port collision for actor egress

hsteude:
Related to #430. (see [discussion](https://github.com/agent-substrate/substrate/issues/430#issuecomment-5515900419))

## Summary

Ateom currently uses port 15001 for intercepted Actor egress, which conflicts with Istio's outbound listener.

This changes the default to 16001 for the WorkerPool controller and both Ateom implementations. Port 16001 is not otherwise used by Substrate and does not conflict with Istio's documented ports (i'm happy to change it to somehing else tho). If useful for other integrations, I would also be happy to follow up with a configurable WorkerPool-level port.

## Testing

- `make verify`
- Deployed an Istio-injected gVisor WorkerPool
- Verified Actor ingress, authenticated egress, pause, suspend, and resume
- Observed no listener conflicts or container restarts

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR (not
      applicable; this internal default was not previously documented)

### DIFF
--- a/cmd/atecontroller/internal/controllers/workerpool_apply.go
+++ b/cmd/atecontroller/internal/controllers/workerpool_apply.go
@@ -96,7 +96,7 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool, otel ateomOTelSettin
 			"--atunnel-connect-listen-address=:8443",
 			"--atunnel-credential-bundle="+atunnelIdentityMountPath+"/credential-bundle.pem",
 			"--atunnel-trust-bundle="+atunnelIdentityMountPath+"/trust-bundle.pem",
-			"--atunnel-egress-listen-address=0.0.0.0:15001",
+			"--atunnel-egress-listen-address=0.0.0.0:16001",
 			"--atunnel-egress-trust-bundle="+atunnelEgressTrustMountPath+"/trust-bundle.pem",
 		).
 		WithPorts(corev1ac.ContainerPort().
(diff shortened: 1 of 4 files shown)


cmd/atecontroller/internal/controllers/workerpool_apply.go
```diff
@@ -96,7 +96,7 @@ func buildDeploymentApplyConfig(wp *atev1alpha1.WorkerPool, otel ateomOTelSettin
 			"--atunnel-connect-listen-address=:8443",
 			"--atunnel-credential-bundle="+atunnelIdentityMountPath+"/credential-bundle.pem",
 			"--atunnel-trust-bundle="+atunnelIdentityMountPath+"/trust-bundle.pem",
-			"--atunnel-egress-listen-address=0.0.0.0:15001",
+			"--atunnel-egress-listen-address=0.0.0.0:16001",
```
keithmattix: nit: should we keep the ate/8 theme and do 18001?

EItanya: Seconded

hsteude: How did I miss that chance :D. Made it more at(e)sque.

hsteude:
@keithmattix @EItanya : While i'd still like to see this merged, i'm not sure if this is still relevant, give this comment: https://github.com/agent-substrate/substrate/issues/430#issuecomment-5577920357

keithmattix:
My POV (maybe @howardjohn disagrees) is that if users want to do workload level mTLS and authz with substrate workers, that's should be possible and the port collision is still useful to remove

howardjohn:
I am not opposed but I don't think we should use this as a slippery slope to "meshes should implement native substrate concepts in their sidecar" (see https://github.com/agent-substrate/substrate/issues/430#issuecomment-5571630750); they should just be standard PEPs. But keeping a port collision is a pretty flimsy way to enforce that so no problem changing

keithmattix:
> I don't think we should use this as a slippery slope to "meshes should implement native substrate concepts in their sidecar" (see https://github.com/agent-substrate/substrate/issues/430#issuecomment-5571630750); they should just be standard PEPs.

+1

PR #1439, opened 2026-09-03
ateapi: validate system-info projection paths at template creation

### LINKED ISSUE #1436: `SystemInfoVolumeSource` accepts path traversal (e.g. `../`)
`CreateActorTemplate` accepts path traversal sequences (e.g. `../../traversal-escape`) and absolute paths (e.g. `/escaped`) in `ActorMetadataItem.path` and `TrustBundleDataSource.path`.

The failure mode is that the ActorTemplate is created but the golden actor is stuck in `RESUMING` state.

```
kubectl ate resume actor af840fb9-b58e-478b-8b87-ef0400d8c29b -a ate-golden

Error: failed to resume actor: rpc error: code = Internal desc = while populating system-info volume "sys-info": invalid system-info path "../../traversal-escape": must not contain empty, '.', or '..' segments
```

Note that we have some defense in depth in atelet to prevent path traversal so this is *not* a security issue. We should add a validation to shift-left the error to the time when the ActorTemplate resource is created though.

### LINKED ISSUE #1438: Duplicate projection paths accepted in `SystemInfoVolumeSource`
`CreateActorTemplate` accepts templates where multiple data source items project to the exact same file `path` within a `SystemInfoVolumeSource` volume.

Note that `atelet` processes items sequentially and writes each file atomically, causing items to silently overwrite and clobber each other without any error or warning.

thompsonmax:
`CreateActorTemplate` accepted `../escape` and `/escaped` in `ActorMetadataItem.path` and `TrustBundleDataSource.path`, and accepted several data sources projecting to the same path. atelet only rejects the traversal at actor start, so the template is created but the golden actor sits in `RESUMING`. atelet never rejects duplicate paths at all: it writes the files in order and the last one wins.

This PR moves both checks to template validation.

- The path rule lives in a new `internal/volumepath` package that both ateapi and atelet call, so the two cannot drift. It splits on `/` and rejects any segment that is empty, `.`, or `..`, caps depth at 16 segments, and rejects NUL bytes. Any other byte is allowed, so Unicode paths pass. `ActorMetadataItem.path` and `TrustBundleDataSource.path` apply it through a `+k8s:customValidation` hook, and atelet's `writeSystemInfoFile` applies it again before touching the host filesystem.
- `SystemInfoVolumeSource.data_sources` gets a hook that tracks projected paths in a set and reports a duplicate on repeats, in the same shape as the Kubernetes projected-volume validation.

The proto doc comment on `data_sources` also says at most one `actor_metadata` entry may appear. Nothing enforces that, and this PR leaves it alone.

New cases in `TestValidateSystemInfoVolumeSource` and `TestValidateTrustBundleDataSource` cover the rejected shapes, and `internal/volumepath` has its own table test. `hack/verify/codegen.sh` and `hack/verify/golangci-lint.sh` pass.

#1231 moves `writeSystemInfoFile` into `systeminfovolume.go` and onto `os.Root`. Whichever PR lands second points that copy at the shared rule.

Fixes #1436
Fixes #1438

### DIFF
--- a/cmd/ateapi/internal/controlapi/actor_template.go
+++ b/cmd/ateapi/internal/controlapi/actor_template.go
@@ -221,6 +221,82 @@ func ValidateCustom_VolumeMount_MountPath(_ context.Context, _ operation.Operati
 	return nil
 }
 
+// validateSystemInfoPath requires a clean relative Unix path for a file
+// projected into a system-info volume: no leading '/', no empty, '.' or
+// '..' segments, and no control characters. Mirrors the defensive check in
+// atelet's writeSystemInfoFile, which is otherwise only hit at actor start.
+func validateSystemInfoPath(fldPath *field.Path, p string) field.ErrorList {
+	bad := strings.HasPrefix(p, "/") || strings.HasSuffix(p, "/") ||
+		strings.Contains(p, "//") || mountPathBadSegmentRE.MatchString(p)
+	if !bad {
+		for _, r := range p {
+			if r < 0x20 || r == 0x7f {
+				bad = true
+				break
+			}
+		}
+	}
+	if bad {
+		return field.ErrorList{field.Invalid(fldPath, p, "must be a clean relative Unix path: must not start with '/', and contain no '..', '.', '//', trailing '/', or control characters")}
+	}
+	return nil
+}
+
+func ValidateCustom_ActorMetadataItem_Path(_ context.Context, _ operation.Operation, fldPath *field.Path, value, _ *string) field.ErrorList {
+	return validateSystemInfoPath(fldPath, *value)
+}
+
+func ValidateCustom_TrustBundleDataSource_Path(_ context.Context, _ operation.Operation, fldPath *field.Path, value, _ *string) field.ErrorList {
+	return validateSystemInfoPath(fldPath, *value)
+}
+
+// ValidateCustom_SystemInfoVolumeSource_DataSources requires every projected
+// file path to be unique across all data sources, and no path to name a
+// directory of another. atelet writes the files in order into one tree, so a
+// repeated path silently clobbers the earlier file and a file/directory
+// clash fails the actor start.
+func ValidateCustom_SystemInfoVolumeSource_DataSources(_ context.Context, _ operation.Operation, fldPath *field.Path, value, _ []*ateapipb.SystemInfoDataSource) field.ErrorList {
+	var errs field.ErrorList
+	type projected struct {
+		path    string
+		fldPath *field.Path
+	}
+	var seen []projected
+	check := func(fldPath *field.Path, path string) {
+		if path == "" {
+			return // required is enforced by tags
+		}
+		for _, prev := range seen {
+			switch {
+			case prev.path == path:
+				errs = append(errs, field.Duplicate(fldPath, path))
+				return
+			case strings.HasPrefix(prev.path, path+"/"), strings.HasPrefix(path, prev.path+"/"):
+				errs = append(errs, field.Invalid(fldPath, path, fmt.Sprintf("must not be a directory of, or nested under, another projected path (%s)", prev.fldPath)))
+				return
+			}
+		}
+		seen = append(seen, projected{path: path, fldPath: fldPath})
+	}
+	for i, ds := range value {
+		if ds == nil {
+			continue
+		}
+		if tb := ds.TrustBundle; tb != nil {
+			check(fldPath.Index(i).Child("trust_bundle", "path"), tb.Path)
+		}
+		if am := ds.ActorMetadata; am != nil {
+			for j, item := range am.Items {
+				if item == nil {
+					continue
+				}
+				check(fldPath.Index(i).Child("actor_metadata", "items").Index(j).Child("path"), item.Path)
+			}
+		}
+	}
+	return errs
+}
+
 // ValidateCustom_ImageVolumeSource_Reference requires image references to
 // be pinned by digest, because changing the image content under a fixed
 // reference invalidates snapshots.
--- a/cmd/ateapi/internal/controlapi/actor_template.go
+++ b/cmd/ateapi/internal/controlapi/actor_template.go
@@ -23,12 +23,14 @@ import (
 
 	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
 	"github.com/agent-substrate/substrate/internal/resources"
+	"github.com/agent-substrate/substrate/internal/volumepath"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	"k8s.io/apimachinery/pkg/api/operation"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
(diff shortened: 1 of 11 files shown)


ahmedtd:
>ActorMetadataItem.path and TrustBundleDataSource.path get a +k8s:customValidation hook requiring a clean relative Unix path: no leading /, no empty, . or .. segments, and no control characters. That is the same segment rule as atelet's writeSystemInfoFile, plus the control-character rule mount_path already has.

This is good validation, but, just so I understand, did our use of os.Root already block any escapes using this?

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -216,6 +216,82 @@ func ValidateCustom_VolumeMount_MountPath(_ context.Context, _ operation.Operati
 	return nil
 }
 
+// validateSystemInfoPath requires a clean relative Unix path for a file
+// projected into a system-info volume: no leading '/', no empty, '.' or
+// '..' segments, and no control characters. Mirrors the defensive check in
+// atelet's writeSystemInfoFile, which is otherwise only hit at actor start.
+func validateSystemInfoPath(fldPath *field.Path, p string) field.ErrorList {
+	bad := strings.HasPrefix(p, "/") || strings.HasSuffix(p, "/") ||
+		strings.Contains(p, "//") || mountPathBadSegmentRE.MatchString(p)
```
ahmedtd: Can you do this in a more principled way?  Split the string into segments, and explicitly look for segments that are "", ".", or "..".  We should also make sure there are fewer than 16 (or something sane) segments.

And this logic should probably exist in a common package, called from both atelet and the API validation, to prevent drift between them.

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -216,6 +216,82 @@ func ValidateCustom_VolumeMount_MountPath(_ context.Context, _ operation.Operati
...
+	if !bad {
+		for _, r := range p {
+			if r < 0x20 || r == 0x7f {
```
ahmedtd: So, you're iterating over runes here, which means you're implicitly decoding the string as UTF-8.  There are many other non-printable or control 

In general, to Linux, paths are sequences of *bytes*, and any byte is valid except:
* 0 (ends the path)
* '/' (directory separator).

There are two self-consistent options for what to do here:
* Substrate accepts any valid Linux path.  In this case, iterate over `p` using a standard for loop, and check that `p[i]` is not 0.
* Substrate only accepts printable ASCII paths --- iterate over `p` using a standard for loop, and check that 32 <= `p[i]` <= 126.

I would lean towards the first choice so that we don't lock users out of using Unicode characters (whether they encode them in UTF-8, or UTF-16, like monsters).

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -216,6 +216,82 @@ func ValidateCustom_VolumeMount_MountPath(_ context.Context, _ operation.Operati
...
+// directory of another. atelet writes the files in order into one tree, so a
+// repeated path silently clobbers the earlier file and a file/directory
+// clash fails the actor start.
+func ValidateCustom_SystemInfoVolumeSource_DataSources(_ context.Context, _ operation.Operation, fldPath *field.Path, value, _ []*ateapipb.SystemInfoDataSource) field.ErrorList {
+	var errs field.ErrorList
+	type projected struct {
+		path    string
+		fldPath *field.Path
+	}
+	var seen []projected
```
ahmedtd: This is quadratic, I think --- we should use a map to track which paths have been seen.

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -216,6 +216,82 @@ func ValidateCustom_VolumeMount_MountPath(_ context.Context, _ operation.Operati
...
+	check := func(fldPath *field.Path, path string) {
```
ahmedtd: Factoring this out into an inline function is making things less readable, I think.

Take a look at what Kubernetes does for the same check: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L1167

```
switch {
  case ds.TrustBundle != nil:
    if seen[ds.TrustBundle.Path] {
      // append error
    }
    seen[ds.TrustBundle.Path] = true
}
```

I also don't think we need to check the prefixes --- Kubernetes doesn't, and it doesn't seem to be a problem

thompsonmax:
> > ActorMetadataItem.path and TrustBundleDataSource.path get a +k8s:customValidation hook requiring a clean relative Unix path: no leading /, no empty, . or .. segments, and no control characters. That is the same segment rule as atelet's writeSystemInfoFile, plus the control-character rule mount_path already has.
> 
> This is good validation, but, just so I understand, did our use of os.Root already block any escapes using this?

Yeah I believe the use of `os.Root` in #1231 should prevent escaping outside the systemInfoVolume root at actor run/restore in the common `write` function. The validation in this PR should catch this sooner during template creation.

cmd/ateapi/internal/controlapi/actor_template.go
thompsonmax: Done in 90a43d7d. The rule now lives in internal/volumepath: split on '/', reject any segment that is empty, '.' or '..', and cap at 16 segments. Both ValidateCustom_*_Path hooks and atelet's writeSystemInfoFile call it, and the package has its own table test. #1231 moves writeSystemInfoFile into systeminfovolume.go, so whichever PR lands second will point that copy at the same function.

cmd/ateapi/internal/controlapi/actor_template.go
thompsonmax: Went with your first option: the check is now a byte scan that rejects only NUL, so Unicode in any encoding passes (there is a UTF-8 case in the table test). One nuance: proto3 strings must be valid UTF-8, so the UTF-16 case is already rejected on the wire before validation sees it, and NUL is the only byte left to check.

cmd/ateapi/internal/controlapi/actor_template.go
thompsonmax: Switched to a sets.Set[string] keyed by path, in 90a43d7d.

cmd/ateapi/internal/controlapi/actor_template.go
thompsonmax: Rewrote it in that shape: a switch on the union member and a seen set, without the closure. I dropped the prefix check as well. For the record, the case it caught is `certs` beside `certs/ca.pem`, which passes creation and then fails at Run/Restore when MkdirAll hits the file, the same shape of symptom as #1436. Kubernetes accepts it too, so I'm fine leaving it out; it's an ancestor-set check if it ever bites.

thompsonmax:
> did our use of os.Root already block any escapes using this?

Not on main today. `writeSystemInfoFile` in cmd/atelet/main.go is a string check (non-empty, no leading `/`, no empty/`.`/`..` segments) followed by plain filepath.Join, MkdirAll and a temp-and-rename write. The error quoted in #1436 is that check firing, so escapes were already blocked, just not by os.Root.

You're probably thinking of #1231, which is still open. That PR moves the writer into systeminfovolume.go, opens an os.Root at the volume root, and routes MkdirAll and the atomic write through it. It keeps the string check in front, and adds a test for a symlink inside the volume pointing outside it, which a string check can't see. Once that lands, os.Root would independently refuse both `/escaped` (a leading separator is errPathEscapes) and `../../x`.

Either way the refusal happens at Run/Restore, which is the stuck-in-RESUMING symptom. This PR is about rejecting the path when the template is created, so both layers should stay. The segment check now lives in internal/volumepath and both atelet and the API call it, per your other comment.

cmd/ateapi/internal/controlapi/actor_template.go
thompsonmax: Split this out to a validation function in a separate `internal/volumepath/volumepath.go` file which just splits into segments, validates there's at most 16, and validates there's no bad segments ('..', '.', or ''). I'll add it to the atelet validation in whichever PR (this or the live refresh) loses the merge race.

cmd/ateapi/internal/controlapi/actor_template.go
thompsonmax: Went with the first option you mentioned, made this part of the common validation function.

cmd/ateapi/internal/controlapi/actor_template.go
thompsonmax: Switched to a set to check for duplicate paths.

cmd/ateapi/internal/controlapi/actor_template.go
thompsonmax: Went with the switch case on each field instead. Sorry for the weird anonymous function setup earlier, that was pretty ugly.

ahmedtd (approved):

PR #1440, opened 2026-09-03
e2e: give TestActorEgress an in-cluster origin

ygao-g:
Part of #246.

TestActorEgress fetched `http://example.com/`, so the test failed whenever that host was unreachable rather than when egress broke. GitHub-hosted runners also have no IPv6, so egress to an external domain needs extra setup on an IPv6-only cluster.

The test now deploys its own testserver origin and fetches that. Since an in-cluster origin would answer a pre-gateway masquerade just as readily, the test asserts the atenet-egress access log recorded the actor's CONNECT, and a subtest fetches by Service DNS name to keep actor-side resolution covered. ServerPod gains `TargetPort` (defaulting to `Port`) so the Service can publish 80 while the container binds an unprivileged listener; the rendering is unit-tested without a cluster.

First of a series retiring the remaining example.com fetches.

Tested: presubmits on the fork.

### DIFF
--- a/internal/e2e/fixtures/serverpod.yaml.tmpl
+++ b/internal/e2e/fixtures/serverpod.yaml.tmpl
@@ -42,7 +42,7 @@ spec:
 ${ARGS}
     ports:
     - name: serve
-      containerPort: ${PORT}
+      containerPort: ${TARGET_PORT}
     # The suite starts dialing as soon as it has an address, so without a
     # readiness gate the first connection can land before the listener exists
     # and come back refused -- which reads as a failure of whatever sits between
@@ -82,4 +82,4 @@ spec:
   ports:
   - name: serve
     port: ${PORT}
-    targetPort: ${PORT}
+    targetPort: ${TARGET_PORT}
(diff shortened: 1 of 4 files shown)


haiyanmeng (approved):

EItanya:
Why was this PR merged with a WIP title?

haiyanmeng:
> Why was this PR merged with a WIP title?

Good catch. I totally missed that. I think the most likely reason is that @ygao-g  forgot to remove `[WIP]` before requesting me and @LiorLieberman  for review.

haiyanmeng:
@ygao-g , can you clarify the intention here?

haiyanmeng:
@EItanya , I created a revert https://github.com/agent-substrate/substrate/pull/1471.

ygao-g:
> Why was this PR merged with a WIP title?

My bad, the PR is ready for review and to be merged, the [WIP] was added by mistakes after I sending it out for review.  

Will double check my PRs with [WIP] in titles and make sure they are in draft status and cannot be merged to avoid similar situations.

PR #1444, opened 2026-09-03
ateapi: Fix various actor workflow races

### LINKED ISSUE #1443: [Bug]: Lifecycle finalizers can overwrite a concurrent worker-deletion crash
### What happened?

  When a worker disappears during ResumeActor or SuspendActor, DeleteWorker marks the actor
  CRASHED and clears its worker assignment. The lifecycle workflow can subsequently re-read that
  updated actor and unconditionally finalize it:

  - Resume changes CRASHED to RUNNING, leaving a running actor without a worker.
  - Suspend changes CRASHED to SUSPENDED, potentially without a valid snapshot.

  Optimistic version checks do not prevent this because the finalizers read and update the new
  CRASHED version.

### Expected Behavior

finalization should require the actor to remain RESUMING or SUSPENDING. If worker deletion has already crashed it, the workflow should preserve CRASHED and return FailedPrecondition.

### Steps to Reproduce

Regression tests can reproduce both cases by deleting the worker from the fake atelet
immediately after restore/checkpoint completes.

### Sandbox Runtime

Both / Runtime Agnostic

main

_No response_

### Host OS & Architecture

_No response_

```shell

```

### Additional Context

_No response_

### Confirmation

EItanya:
Worker deletion can race both lifecycle finalization and the gap between a worker claim and its actor backlink. Preserve `CRASHED` during resume and suspend finalization, returning `FailedPrecondition` instead of reporting an invalid successful transition.

When a drained worker still has a claim whose actor backlink is pending, advance the actor's version before deleting the worker. The pending resume write then loses its version check and retains the actor's suspended or paused state and snapshots for retry. If resume wins the version check, deletion returns `Aborted` and retries against the committed assignment.

Regression tests cover the lifecycle finalization races, both assignment/deletion orderings, snapshot preservation, and avoiding false crash metrics.

Fixes #1443

Validation:

- PostgreSQL regression tests for both assignment/deletion orderings passed with `-race`.
- `env -u NO_COLOR make verify` in a clean worktree (includes `go test -race ./...`).

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_pause.go
+++ b/cmd/ateapi/internal/controlapi/workflow_pause.go
@@ -212,9 +212,9 @@ func (w *ActorWorkflow) ensureAteletPaused(ctx context.Context, actorRef resourc
 // owned by this actor), records where the local snapshot lives, and commits
 // PAUSED with the assignment cleared in a single update — or CRASHED when the
 // worker's node name was lost, since a local snapshot on an unknown node can
-// never be resumed. It re-reads the actor first so an out-of-band transition
-// (e.g. the syncer crashing the actor after its worker died) is not
-// overwritten: with no assignment left there is nothing to finalize.
+// never be resumed. It re-reads the actor after releasing the worker and only
+// finalizes PAUSING, so an out-of-band transition (e.g. worker deletion
+// crashing the actor) is not overwritten.
 func (w *ActorWorkflow) ensurePausedFinalized(ctx context.Context, actorRef resources.ActorRef, actorTemplate *ateapipb.ActorTemplate) (_ *ateapipb.Actor, err error) {
 	ctx, done := stepSpan(ctx, "FinalizePaused")
 	defer func() { err = done(err) }()
@@ -222,6 +222,9 @@ func (w *ActorWorkflow) ensurePausedFinalized(ctx context.Context, actorRef reso
 	latestActor, err := w.store.GetActor(ctx, actorRef)
 	if err != nil {
 		return nil, err
+	}
+	if got := latestActor.GetStatus().GetState(); got != ateapipb.ActorState_ACTOR_STATE_PAUSING {
+		return nil, status.Errorf(codes.FailedPrecondition, "FinalizePaused prerequisite not met for Actor: %s (got: %v, want %s)", actorRef, got, ateapipb.ActorState_ACTOR_STATE_PAUSING)
 	}
 
 	// 1. Free the worker (if it hasn't been freed yet)
@@ -251,7 +254,9 @@ func (w *ActorWorkflow) ensurePausedFinalized(ctx context.Context, actorRef reso
 		if err != nil {
 			return nil, err
 		}
-		wasAlreadyCrashed := latestActor.GetStatus().GetState() == ateapipb.ActorState_ACTOR_STATE_CRASHED
+		if got := latestActor.GetStatus().GetState(); got != ateapipb.ActorState_ACTOR_STATE_PAUSING {
+			return nil, status.Errorf(codes.FailedPrecondition, "FinalizePaused prerequisite not met for Actor: %s (got: %v, want %s)", actorRef, got, ateapipb.ActorState_ACTOR_STATE_PAUSING)
+		}
 		newState := ateapipb.ActorState_ACTOR_STATE_PAUSED
 		if nodeName == "" {
 			// Without a node name we cannot record where the local snapshot lives,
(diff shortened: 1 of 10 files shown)


cmd/ateapi/internal/controlapi/workflow_pause.go
```diff
@@ -288,7 +293,7 @@ func (w *ActorWorkflow) ensurePausedFinalized(ctx context.Context, actorRef reso
 			toUpdate.Status.WorkerAssignment = nil
 			return nil
 		})
-		if err == nil && storedActor.GetStatus().GetState() == ateapipb.ActorState_ACTOR_STATE_CRASHED && !wasAlreadyCrashed {
+		if err == nil && storedActor.GetStatus().GetState() == ateapipb.ActorState_ACTOR_STATE_CRASHED {
```
sfunkenhauser, line 296: This can't be hit anymore can it? But maybe it never could? Wouldn't the state after the update (and no error) always be paused?

cmd/ateapi/internal/controlapi/workflow_suspend.go
sfunkenhauser, line 390: releaseWorker doesn't actually modify the actor, I don't think this re-fetch is necessary. Can we remove it to simply this logic?

cmd/ateapi/internal/controlapi/workflow_worker_delete.go
```diff
@@ -181,14 +181,14 @@ func (w *WorkerWorkflow) releaseBoundActor(ctx context.Context, worker *ateapipb
 		markSkipped(ctx, "assignment names a superseded actor incarnation")
 		return nil
 	}
-	// Skip if a concurrent SuspendActor already cleared the pointer, or if the
-	// actor has since been placed on a different worker.
-	if actor.GetStatus().GetWorkerAssignment().GetWorker().GetName() != name {
+	// An actor placed on a different worker is not ours to release.
+	actorWorker := actor.GetStatus().GetWorkerAssignment().GetWorker().GetName()
+	if actorWorker != "" && actorWorker != name {
```
sfunkenhauser, line 186: If actorWorker == "" it no longer points at this worker, why are we excluding that case now? Is this to cover a race condition with resume? Ideally we would handle that in resume and not here.

I _think_ the BindActorToWorker call will fail since the worker will be marked draining at this point. So I don't think this is necessary? But maybe there is something I'm overlooking.

cmd/ateapi/internal/controlapi/workflow_pause.go
EItanya: Agreed after tracing the full workflow. A valid worker always has a node name, and normal worker deletion crashes the actor before removing the worker, which the new state guard catches. The existing test manufactured an inconsistent assignment rather than reproducing a reachable workflow. I’ll remove that fallback and simplify finalization.

cmd/ateapi/internal/controlapi/workflow_suspend.go
EItanya: Correct that `releaseWorker` itself doesn’t modify the actor. However, concurrent deletion can already have read the assignment before suspend releases it, then crash the actor afterward. I reproduced this through both RPCs. Without the re-fetch, suspend deletes the previous snapshot before its actor update returns `Aborted`, leaving the actor referencing deleted data. I’ll clarify that purpose and add the snapshot-preservation assertion.

cmd/ateapi/internal/controlapi/workflow_worker_delete.go
EItanya: The bind can commit before draining begins, with the actor update still pending. I reproduced that ordering with actual pod deletion observed by the informer. Without this guard, resume commits `RESUMING` pointing at the deleted worker; the request fails, and retrying crashes the actor. Draining blocks subsequent binds, but cannot invalidate this already-committed claim. The version bump prevents the pending actor update and preserves its resumable state.

cmd/ateapi/internal/controlapi/workflow_suspend.go
sfunkenhauser: I don't believe this re-fetch actually solves the underlying issue. The same sequence of events can happen after this re-fetch (concurrent delete). The suspend flow gets the worker from the actor and updates the worker, the delete flow gets the actor from the worker and updates the actor. Neither operation will result in a version conflict, and it's possible to perform these updates based on stale data.

There are a few race conditions around snapshot GC, and they are being tracked by #1527

cmd/ateapi/internal/controlapi/workflow_pause.go
sfunkenhauser: Sounds good. Looks like you need to push these changes still?

sfunkenhauser:
Needs a rebase, but otherwise LGTM.

sfunkenhauser (dismissed):

sfunkenhauser:
Apologies that I missed this, but needs another rebase 😞

PR #1445, opened 2026-09-03
(otel): split infra vs. workload faults

krisztianfekete:
`ate.failure.reason` only ever named infrastructure faults, and the crash log didn't say which actor crashed. 

This PR adds the following:

- **`ate.failure.domain`** (`infrastructure` / `workload` / `unknown`) is available next to every `ate.failure.reason`. It's a strict function of the reason so it costs no series, but it's *emitted* rather than derived downstream, so a component ahead of ateapi can report a reason this build rejects, which collapses to `UNKNOWN`, and a consumer matching on reason names would file every one of those as infra. `UNKNOWN` now reports `unknown`, not `infrastructure`.
- **`Actor crashed`** record from ateapi, with full identity including the uid, via a helper both crash sites share. It sits under the same guard as the counter so the two can't disagree on how many crashes happened. The counter is barred from carrying actor identity, so this record is the only way to attribute a crash to one agent.
- **`WORKLOAD_NOT_READY`** on the readyz deadline, as the first workload-domain reason. Deadline branch only: a cancellation is ateom draining, not the actor failing.

Tested locally:
```
desc = WORKLOAD_NOT_READY: readyz for "counter" never returned 200 within 5s
```

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/crash.go
+++ b/cmd/ateapi/internal/controlapi/crash.go
@@ -37,9 +37,10 @@ func maybeCrashActor(ctx context.Context, st crashActorStore, actorRef resources
 	}
 
 	if ateerrors.ActorCrashRequested(err) {
-		// Extract AIP-193 ErrorInfo reason enum from the RPC error detail. If unclassified
-		// or unlisted, default to ateattr.ReasonUnknown to protect metric low-cardinality.
-		reason := ateerrors.ExtractReason(err)
+		// Extract AIP-193 ErrorInfo reason enum from the RPC error detail. Normalized
+		// here rather than in crashActor alone, so the log and the counter cannot
+		// report a different reason for the same crash.
+		reason := ateattr.FailureReason(err)
 
 		// Only the ref is knowable here; crashActor logs the authoritative record.
 		attrs := ateattr.ActorRefLogAttrs(actorRef)
--- a/cmd/ateapi/internal/controlapi/crash.go
+++ b/cmd/ateapi/internal/controlapi/crash.go
@@ -37,13 +37,21 @@ func maybeCrashActor(ctx context.Context, st crashActorStore, actorRef resources
 	}
 
 	if ateerrors.ActorCrashRequested(err) {
-		slog.ErrorContext(ctx, "Setting Actor to crashed due to error", slog.Any("error", err))
 		// Extract AIP-193 ErrorInfo reason enum from the RPC error detail. If unclassified
 		// or unlisted, default to ateattr.ReasonUnknown to protect metric low-cardinality.
 		reason := ateerrors.ExtractReason(err)
 
+		// Only the ref is knowable here; crashActor logs the authoritative record.
+		attrs := ateattr.ActorRefLogAttrs(actorRef)
+		attrs = append(attrs, ateattr.FailureLogAttrs(reason)...)
+		attrs = append(attrs,
+			slog.String(string(ateattr.ErrorTypeKey), status.Code(err).String()),
+			slog.Any("err", err),
+		)
+		slog.LogAttrs(ctx, slog.LevelError, "Setting Actor to crashed due to error", attrs...)
+
 		if cerr := crashActor(ctx, st, actorRef, opName, reason); cerr != nil {
-			slog.ErrorContext(ctx, "Failed to crash actor", slog.Any("cerr", cerr))
+			slog.ErrorContext(ctx, "Failed to crash actor", slog.Any("err", cerr))
 			return cerr
 		}
 		return status.Errorf(codes.DataLoss, "actor %s crashed", actorRef)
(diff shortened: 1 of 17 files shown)


cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -97,12 +105,23 @@ func crashActor(ctx context.Context, st crashActorStore, actorRef resources.Acto
 
 	// Increment metric only after a successful UpdateActor, and only if the actor was not already crashed.
 	if !wasAlreadyCrashed {
+		logActorCrashed(ctx, actor, opName, reason)
```
JeffLuoo, line 109: Miss logging crash here - https://github.com/agent-substrate/substrate/blob/daf5a13907afdcbe6aaed2528566840ac010c3eb/cmd/ateapi/internal/controlapi/workflow_pause.go#L298

cmd/ateapi/internal/controlapi/crash.go
```diff
@@ -37,13 +37,21 @@ func maybeCrashActor(ctx context.Context, st crashActorStore, actorRef resources
...
 
 	if ateerrors.ActorCrashRequested(err) {
-		slog.ErrorContext(ctx, "Setting Actor to crashed due to error", slog.Any("error", err))
 		// Extract AIP-193 ErrorInfo reason enum from the RPC error detail. If unclassified
 		// or unlisted, default to ateattr.ReasonUnknown to protect metric low-cardinality.
 		reason := ateerrors.ExtractReason(err)
 
+		// Only the ref is knowable here; crashActor logs the authoritative record.
+		attrs := ateattr.ActorRefLogAttrs(actorRef)
+		attrs = append(attrs, ateattr.FailureLogAttrs(reason)...)
```
JeffLuoo, line 47: https://sourcegraph.com/r/github.com/agent-substrate/substrate@daf5a13/-/blob/internal/ateerrors/ateerrors.go?L185

nit - If errors are unknown, empty string will be returned. Can we normalize it to "unknown" to avoid printing logs like "failed with reason  ."

docs/metrics/registry/metrics.yaml
```diff
@@ -790,6 +829,9 @@ groups:
       - ref: ate.failure.reason
         requirement_level:
           conditionally_required: The operation failed. The key occurs only on the phase that failed and on the total.
+      - ref: ate.failure.domain
+        requirement_level:
+          conditionally_required: ate.failure.reason occurs. The two keys always occur together.
```
JeffLuoo, line 838: recordPhases is shared by both histograms metrics, so this change also puts ate.failure.domain on a failed checkpoint. But metric `ate.actor.checkpoint.duration` still declares only ate.failure.reason — that's a counterexample to the failure-keys-paired rule this PR adds.

internal/ateattr/ateattr.go
```diff
@@ -107,9 +107,56 @@ const (
...
+		slog.String(string(FailureDomainKey), FailureDomain(reason)),
+	}
+}
+
+// FailureDomain classifies a reason value. An unrecognized reason reports
+// FailureDomainUnknown rather than infrastructure, so a taxonomy gap stays
+// visible instead of inflating one side.
+func FailureDomain(reason string) string {
+	if slices.Contains(workloadReasons, ateerrors.Reason(reason)) {
+		return FailureDomainWorkload
```
JeffLuoo, line 158: How about reasons from https://github.com/agent-substrate/substrate/blob/daf5a13907afdcbe6aaed2528566840ac010c3eb/internal/ateerrors/ateerrors.go#L46-L68

- INVALID_CONTAINER_CONFIG
- ReasonInvalidObjectURL
- INVALID_SANDBOX_ASSET

They are errors caused by a bad actorTemplate that users misconfigure or for other reasons. Should we also categorize them as workload error since this is not infra error?

JeffLuoo (changes requested):

cmd/ateapi/internal/controlapi/crash.go
krisztianfekete: Good catch!

cmd/ateapi/internal/controlapi/crash.go
krisztianfekete: Fixed, goes through `ateattr.FailureReason` now!

docs/metrics/registry/metrics.yaml
krisztianfekete: Good point!

internal/ateattr/ateattr.go
krisztianfekete: Agreed on `INVALID_CONTAINER_CONFIG` and `INVALID_OBJECT_URL`, both are now workload. Keeping `INVALID_SANDBOX_ASSET` as infra as `SandboxConfig` is `scope: Cluster`, but depending on ownership, I can see this working fine either way.

JeffLuoo (approved):

PR #1446, opened 2026-09-03
Resolve SandboxConfig from the ActorTemplate instead of the WorkerPool

zoez7:
This PR moves sandbox config selection from the WorkerPool to the ActorTemplate.

The existing behavior is preserved while we are designing the upgrade: sandbox config still cannot be updated once set (ActorTemplates are create-only and `sandbox_config` is immutable).

For now the ActorTemplate still *requires* `sandbox_config.config_name` — there is no resolution of the cluster default (`spec.default`). This is temporary while we figure out the defaulting design.

- [ ] Tests pass


pkg/api/v1alpha1/sandboxconfig_types.go
```diff
@@ -51,18 +51,20 @@ type AssetFile struct {
...
-	// SandboxClass. A WorkerPool with no explicit SandboxConfigName resolves to
-	// the default config for its SandboxClass. At most one default is expected
-	// per SandboxClass.
+	// SandboxClass. Not resolved yet: an ActorTemplate's
+	// sandbox_config.config_name is currently required, so nothing falls back
+	// to the default; the field is retained while the defaulting design is
+	// worked out. At most one default is expected per SandboxClass.
 	//
 	// +optional
 	Default bool `json:"default,omitempty"`
```
dberkov: can we remove the default field in this case, otherwise field exists however nobody uses it?

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -293,15 +293,22 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
 					return err
 				}
 				if err := validateTemplateVolumesUnchanged(oldTemplate, newTemplate); err != nil {
 					return err
 				}
-			} else if !errors.Is(err, errActorTemplateNotFound) {
-				// Skip the validation if old template is not found
+			case errors.Is(err, errActorTemplateNotFound):
+				// Only allow updating ActorTemplate to the same SandboxConfig for now.
+				if name := oldVal.GetStatus().GetSandboxConfigName(); name != "" && name != newTemplate.GetSandboxConfig().GetConfigName() {
```
dberkov: This fallback is materially weaker than the `err == nil` path above, in three ways:

1. It is skipped entirely when `status.SandboxConfigName` is empty. An actor created from `source_snapshot_tag` that has never resumed has no recorded name (`finalizeRunning` is the only writer of that field), so once its template is deleted, a repoint to a template of a *different* sandbox class/config is accepted — and the next resume feeds the old-lineage snapshot to the wrong runtime.
2. It compares only the config *name*; `ActorStatus` records no class. A SandboxConfig deleted and recreated under the same name with a different class (the dirty-teardown hazard the microvm manifest comment warns about) passes this check, and the actor then crashes at resume via `maybeCrashActor` instead of failing the repoint here with `FailedPrecondition`.
3. It never calls `validateTemplateVolumesUnchanged`, so a template with different volume mounts is accepted in this branch and the snapshot restores to the wrong mount paths.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -293,15 +293,22 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
 				return err
 			}
 			oldTemplate, err := resolveActorTemplate(ctx, s.store, oldVal)
-			if err == nil {
-				if err := validateTemplateSandboxClassUnchanged(oldTemplate, newTemplate); err != nil {
+			switch {
+			case err == nil:
+				if err := validateTemplateSandboxConfigUnchanged(oldTemplate, newTemplate); err != nil {
```
dberkov: The normal path validates against the old template re-resolved *by name* (`resolveActorTemplate` follows the ObjectRef, not `status.CurrentActorTemplateUid`) and never consults the `status.SandboxConfigName` provenance this PR adds — the stronger source of truth is wired only into the template-deleted branch below.

Concretely: actor boots under tmpl-A naming config X (status records X); tmpl-A is deleted and recreated naming config Z (legal — templates are create-only but deletable); a repoint to tmpl-B naming Z compares Z vs Z and passes, despite the guest state having been built with X — exactly the condition the fallback branch rejects. Severity is tempered by the resume path's `TemplateReplaced` flag forcing a DATA-scope restore, so the residual harm is the silent config divergence the comment above warns about. Checking "the new template must name `status.SandboxConfigName` when non-empty" in *both* branches would close it.

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -61,9 +66,29 @@ func (s *RPCService) CreateActorTemplate(ctx context.Context, req *ateapipb.Crea
 	return stored, nil
 }
 
+// validateTemplateSandboxConfig checks the template's named SandboxConfig
+// against the cluster: it must exist (FailedPrecondition — retryable, the
```
dberkov: The comment calls the not-found case retryable, but no shipped caller retries: ate-setup's `steps/substrate.go` treats non-AlreadyExists as fatal, and kubectl-ate, `hack/install-ate.sh`, and benchmarking's `deploy.sh` fail hard. That introduces a create-time race that didn't exist when resolution was lazy at first boot.

Example: `hack/run-microvm-demo.sh` applies the microvm SandboxConfig as `install-microvm-deps.sh`'s last act and then immediately creates templates naming it; if the ateapi watch lags the script handoff (apiserver load, watch re-establishment), the deploy aborts with `SandboxConfig "microvm" not found`. The functionaltest helper polls the lister for up to 5s before proceeding — the wait production callers lack. Suggest retrying on FailedPrecondition in ate-setup/scripts, or falling back to an apiserver read on cache miss.

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -61,9 +66,29 @@ func (s *RPCService) CreateActorTemplate(ctx context.Context, req *ateapipb.Crea
...
+// lister may briefly lag a just-created config) and its class must match the
+// template's sandbox_class (InvalidArgument). config_name is required; the
+// declarative validation has already rejected an empty one.
+func (s *RPCService) validateTemplateSandboxConfig(sandbox *ateapipb.SandboxConfig) error {
+	name := sandbox.GetConfigName()
+	sc, err := s.sandboxConfigLister.Get(name)
```
dberkov: This hand-rolls the same invariant `resolveSandboxAssets` enforces at cold boot in this package (lister `Get` by config name + class comparison via `sandboxClassString`), with near-identical error strings that are already drifting. The differing gRPC codes justify separate error *mapping*, not a second lookup+match implementation: a future change to the matching rules (new class, alias, relaxed match) applied to one copy would let a template pass create and then fail every cold boot, or vice versa. Consider extracting a shared lookup helper returning typed not-found / class-mismatch errors that each caller maps to its own codes.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -714,7 +724,7 @@ func (w *ActorWorkflow) ensureAteletRestored(ctx context.Context, actorRef resou
 		tele.WireSnapshotScope = ateattr.SnapshotScopeValue(req.Scope)
 
 		_, err = client.Restore(ctx, req)
-		return tele, maybeCrashActor(ctx, w.store, actorRef, err, "while restoring workload", ateattr.OperationResume)
+		return tele, src.SandboxConfigName, maybeCrashActor(ctx, w.store, actorRef, err, "while restoring workload", ateattr.OperationResume)
```
dberkov: `sandboxConfigName` is threaded differently at each of the seven return sites: `src.SandboxConfigName` here and in the durable-snapshot branch, a reassigned intermediate in the cold-boot branch, and `""` at four error returns — even though the function has named returns. Initializing `sandboxConfigName = src.SandboxConfigName` once at the top and overwriting it only in the cold-boot branch is behavior-identical (the caller discards outputs on error) and removes the per-return variation. As written, a future boot-source branch that picks the wrong value (or leaves the default `""`) compiles cleanly and silently records wrong or missing SandboxConfig provenance on the actor and every subsequent suspend snapshot — the exact field `finalizeRunning` was changed to record.

manifests/ate-install/sandboxconfig-gvisor.yaml
```diff
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Cluster-wide default SandboxConfig for the gVisor (runsc) sandbox class. A
-# WorkerPool with sandboxClass gvisor (the default) and no explicit
-# sandboxConfigName resolves to this. atelet fetches the gVisor release tarball
+# Cluster-wide default SandboxConfig for the gVisor (runsc) sandbox class. An
+# ActorTemplate with sandboxClass gvisor (the default) and no explicit
```
dberkov: This still claims class-default resolution works for templates, but this same PR makes `config_name` required at `CreateActorTemplate` (the functionaltest asserts InvalidArgument) and says so in `sandboxconfig_types.go` ("nothing falls back to the default") and `sandbox_assets.go` (default resolution "currently unreachable"). An operator following this comment and omitting `sandboxConfig.configName` gets InvalidArgument and will debug the SandboxConfig install instead of the template. Suggest the "required for now, defaulting planned" phrasing already used in `architecture.md` and `glossary.md`.

hack/install-ate.sh
```diff
@@ -890,10 +890,10 @@ deploy_ate_system() {
   # SandboxConfig so the defaults below are validated too).
   run_kubectl apply -f manifests/ate-install/sandboxconfig-validation.yaml
 
-  # Install the cluster-wide default sandbox config(s). Sandbox binaries live on
-  # cluster-scoped SandboxConfigs resolved via each WorkerPool's SandboxClass
-  # (decoupled from ActorTemplate). gVisor pools resolve to this default unless
-  # they name their own SandboxConfig.
+  # Install the cluster-wide default sandbox config(s). Sandbox binaries live
+  # on cluster-scoped SandboxConfigs resolved via each ActorTemplate's
+  # sandboxClass. gVisor templates resolve to this default unless they name
```
dberkov: Same as the `sandboxconfig-gvisor.yaml` comment: gVisor templates don't resolve to this default — `config_name` is required at template create as of this PR, so a template that omits it is rejected with InvalidArgument rather than falling back. Suggest aligning with the "required for now, defaulting planned" phrasing in `architecture.md` / `glossary.md`.

pkg/api/v1alpha1/sandboxconfig_types.go
zoez7: Done in 92b4519c50d366a41f034157a35b3a2cd4a98e5c.

cmd/ateapi/internal/controlapi/sandbox_assets.go
```diff
@@ -20,69 +20,37 @@ import (
...
+	templateSandbox *ateapipb.SandboxConfig,
 ) (*ateletpb.SandboxAssets, error) {
-	wp, err := workerPoolLister.WorkerPools(poolNamespace).Get(poolName)
-	if err != nil {
-		return nil, fmt.Errorf("while getting WorkerPool %s/%s: %w", poolNamespace, poolName, err)
-	}
-
-	class := wp.Spec.SandboxClass
+	class := atev1alpha1.SandboxClass(sandboxClassString(templateSandbox.GetSandboxClass()))
 	if class == "" {
```
dberkov: This defaulting is a leftover from the deleted WorkerPool path, and it now fails open: declarative validation (`sandbox_class` required) makes an empty class unreachable for any RPC-created template, and `validateTemplateSandboxConfig` treats `""` as a rejected mismatch. If `sandboxClassString` ever returns `""` (a future enum value missing from converter.go's switch), the template silently boots as gvisor and the mismatch error below blames a class the template never declared — while the `name == ""` check a few lines down fails closed. Two opposite philosophies in one function; suggest deleting the default (or erroring like the name check).

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -293,15 +293,23 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
-				if err := validateTemplateSandboxClassUnchanged(oldTemplate, newTemplate); err != nil {
-					return err
+			switch {
+			case err == nil:
+				// Snapshots are not portable across sandbox runtime
+				// families, so the replacement template must name the same
+				// SandboxConfig.
+				if !proto.Equal(oldTemplate.GetSandboxConfig(), newTemplate.GetSandboxConfig()) {
+					return status.Errorf(codes.FailedPrecondition,
+						"the current actor template names SandboxConfig %q but the new one names %q; the sandbox config must be identical to repoint an actor",
```
dberkov: When only the class differs, this message prints two identical strings. Concretely: SandboxConfig `foo` (class microvm) is deleted and recreated as class gvisor, and a second template naming {GVISOR, foo} validly passes create-time validation; repointing between the two templates is rejected by `proto.Equal` (correctly — the class differs), but the error reads `names SandboxConfig "foo" but the new one names "foo"; the sandbox config must be identical`, which gives the operator nothing actionable. Include `sandbox_class` in the message, or compare name and class separately.

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -61,9 +66,29 @@ func (s *RPCService) CreateActorTemplate(ctx context.Context, req *ateapipb.Crea
...
+	if k8serrors.IsNotFound(err) {
+		return status.Errorf(codes.FailedPrecondition, "SandboxConfig %q not found", name)
```
dberkov: The comment above calls this retryable, but no caller anywhere retries it: ate-setup's CreateActorTemplate step tolerates only AlreadyExists, the internal/e2e fixtures `t.Fatalf`, kubectl-ate fails outright, and benchmarking's `deploy.sh` pipes straight into kubectl-ate. Pre-PR, CreateActorTemplate read no cluster state, so an "apply SandboxConfig; create template" sequence couldn't flake; now, if the ateapi SandboxConfig informer hasn't delivered a just-applied config (watch reconnect, apiserver overload, back-to-back apply+create), template creation fails with `SandboxConfig ... not found`. Suggest a short poll here, or retries in the CLI/setup callers.

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -61,9 +66,29 @@ func (s *RPCService) CreateActorTemplate(ctx context.Context, req *ateapipb.Crea
 	return stored, nil
 }
 
+// validateTemplateSandboxConfig checks the template's named SandboxConfig
+// against the cluster: it must exist (FailedPrecondition — retryable, the
+// lister may briefly lag a just-created config) and its class must match the
+// template's sandbox_class (InvalidArgument). config_name is required; the
+// declarative validation has already rejected an empty one.
+func (s *RPCService) validateTemplateSandboxConfig(sandbox *ateapipb.SandboxConfig) error {
```
dberkov: This duplicates `resolveSandboxAssets`' lister.Get + class-match logic in the same package, and the two copies have already drifted: the error strings differ (`...but sandbox_config.sandbox_class is %q` vs `...but the ActorTemplate's sandbox_config.sandbox_class is %q`), and this copy rejects an empty class as a mismatch while the other silently defaults it to gvisor. A future change to the matching rule (aliases, a new class, defaulting) has to be made twice, or create-time and boot-time will enforce different rules. The differing gRPC codes justify separate error mapping, not a second implementation — a shared lookup helper with sentinel not-found/class-mismatch errors would let each caller map to its own codes.

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -61,9 +66,29 @@ func (s *RPCService) CreateActorTemplate(ctx context.Context, req *ateapipb.Crea
...
+	name := sandbox.GetConfigName()
+	sc, err := s.sandboxConfigLister.Get(name)
+	if k8serrors.IsNotFound(err) {
+		return status.Errorf(codes.FailedPrecondition, "SandboxConfig %q not found", name)
+	}
+	if err != nil {
+		return fmt.Errorf("while getting SandboxConfig %q: %w", name, err)
+	}
+	if class := sandboxClassString(sandbox.GetSandboxClass()); string(sc.Spec.SandboxClass) != class {
+		return status.Errorf(codes.InvalidArgument,
```
dberkov: Minor: the class mismatch returns permanent `InvalidArgument`, but the outcome depends on the same lagging lister the not-found branch above explicitly treats as retryable (`FailedPrecondition`). Delete SandboxConfig `foo` (microvm), recreate it as gvisor, and immediately create a template with {GVISOR, foo}: the stale lister row yields `InvalidArgument` (don't-retry per AIP-193) for a request that would succeed a second later. Defensible when the mismatch is a caller bug, but the state-dependent case argues for `FailedPrecondition` here too.

cmd/ateapi/internal/controlapi/sandbox_assets.go
zoez7: Done in 070fc5031480a31a65082532899be9227293dd5d. Returns an error instead of defaulting the "".

cmd/ateapi/internal/controlapi/actor.go
zoez7: Done in 070fc5031480a31a65082532899be9227293dd5d. restructured the error message.

cmd/ateapi/internal/controlapi/actor_template.go
zoez7: This is retriable in the sense that the client can retry creating the actortemplate if they encounter the FailedPrecondition errors. I don't think the server should poll here, the client should add a retry if they see this error.

cmd/ateapi/internal/controlapi/actor_template.go
zoez7: Done in 070fc5031480a31a65082532899be9227293dd5d. extracted common logic.

cmd/ateapi/internal/controlapi/actor_template.go
zoez7: Done in https://github.com/agent-substrate/substrate/commit/070fc5031480a31a65082532899be9227293dd5d.

PR #1448, opened 2026-09-03
actorTemplate(fix): update volume mount validation

### LINKED ISSUE #1434: Container `volume_mounts` can reference undefined template volumes
Creating an `ActorTemplate` with container volume mounts that reference undeclared volumes (e.g. `ghost-vol` not present in `ActorTemplate.volumes`) succeeds with `codes.OK`.

The failure mode is that the golden actor is created and is left permanently stuck in `ACTOR_STATE_RESUMING`.

We should validate that each `mount.Name` corresponds to an entry in `ActorTemplate.Volumes`.

### LINKED ISSUE #1435: Duplicate `mount_path` for distinct volumes accepted within a container
Creating an `ActorTemplate` where a container mounts two different volumes (e.g., `vol-a` and `vol-b`) to the exact same path (`/data`) succeeds with `codes.OK`.

The failure mode seems to be different in gvisor and GHV. In gvisor, it looks like both mounts will work, one will shadow the other. In GHV, the container will crash.

We should add a validation ensuring all `VolumeMount.MountPath` strings within each container are distinct.

shrutiyam-glitch:
Fixes #1435 #1434 

Two `ActorTemplate` validation gaps that pass creation today and fail at runtime, both closed with `customValidation` hooks (neither rule is expressible in tags):

  - Fixes #1434 — a container `volume_mount` naming a volume the template never declares was accepted, leaving the golden actor permanently stuck in `ACTOR_STATE_RESUMING`. The rule spans two subtrees (containers and volumes), so the hook lives on `CreateActorTemplateRequest.actor_template`, where it sees the whole template. 
  - Fixes #1435 — two different volumes mounted at the same `mount_path` were accepted (the list-map key is name, so path collisions slipped through); The hook on `Container.volume_mounts` enforces path uniqueness within each container. This resolves the todo partially, and also discussed here - https://github.com/agent-substrate/substrate/pull/1303#discussion_r3905902589 

- [ ] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/actor_template_test.go
+++ b/cmd/ateapi/internal/controlapi/actor_template_test.go
@@ -122,6 +122,19 @@ func TestValidateCreateActorTemplateRequest(t *testing.T) {
 		})},
 		field.ErrorList{field.Required(field.NewPath("actor_template", "containers").Index(0).Child("image"), "")},
 	}, {
+		"volume mount referencing a declared volume",
+		&ateapipb.CreateActorTemplateRequest{ActorTemplate: validActorTemplate(func(tmpl *ateapipb.ActorTemplate) {
+			tmpl.Volumes = []*ateapipb.Volume{{Name: "data", DurableDir: &ateapipb.DurableDirVolumeSource{}}}
+			tmpl.Containers[0].VolumeMounts = []*ateapipb.VolumeMount{{Name: "data", MountPath: "/var/data"}}
+		})},
+		nil,
+	}, {
+		"volume mount referencing an undeclared volume",
+		&ateapipb.CreateActorTemplateRequest{ActorTemplate: validActorTemplate(func(tmpl *ateapipb.ActorTemplate) {
+			tmpl.Containers[0].VolumeMounts = []*ateapipb.VolumeMount{{Name: "ghost-vol", MountPath: "/var/data"}}
+		})},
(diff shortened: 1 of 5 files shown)


cmd/ateapi/internal/controlapi/actor_template_test.go
```diff
@@ -487,6 +500,32 @@ func TestValidateActorTemplate(t *testing.T) {
...
+				{Name: "other", MountPath: "/var/data"},
+			}
+		},
+		want: field.ErrorList{field.Duplicate(field.NewPath("containers").Index(0).Child("volume_mounts").Index(1).Child("mount_path"), nil)},
+	}, {
+		name: "nested mount paths are rejected",
+		mutate: func(tmpl *ateapipb.ActorTemplate) {
+			tmpl.Containers[0].VolumeMounts = []*ateapipb.VolumeMount{
+				{Name: "data", MountPath: "/data"},
+				{Name: "config", MountPath: "/data/config"},
```
msau42, line 517: Can you also test a nested path more than a level deep? Like /data and /data/a/b/c

shrutiyam-glitch: Done in [7085bfb](https://github.com/agent-substrate/substrate/pull/1448/commits/7085bfbf73e5a8355e4a9a309ff54030e2da827c)

msau42 (approved):

PR #1460, opened 2026-09-03
manifests: wire OTLP exporter settings into the egress ext-proc

### LINKED ISSUE #1459: atenet-egress ext-proc has no OTLP wiring - all egress metrics and spans are silently dropped
## Summary

The egress gateway's ext-proc container is deployed without the shared OTLP exporter settings: unlike every other control-plane component, the `atenet-egress` Deployment does not consume the `ate-otel-config` ConfigMap via `envFrom`. `OTEL_EXPORTER_OTLP_ENDPOINT` is therefore unset, the OTel SDK falls back to `localhost:4317` (where nothing listens), and **every egress metric and trace span is silently dropped** — while the SDK logs an export-failure warning every ~20 s.

## Observed (live GKE install, substrate rev 0d85e8572161)

Ext-proc logs, repeating forever (~210 warnings per 2 h — the single largest warn/error source in the whole cluster):

```
"OpenTelemetry SDK error" err="traces export: exporter export timeout: rpc error: code = Unavailable
  desc = ... dial tcp [::1]:4317: connect: cannot assign requested address"
"OpenTelemetry SDK error" err="failed to upload metrics: exporter export timeout: ..."
```

Consequences verified in the managed-Prometheus / Cloud Trace data:

- `target_info` in GMP lists ateapi, atecontroller, atelet, atenet-router, and ateom workers — **no entry from the egress deployment**.
- The registry-listed ext_proc instruments (`docs/metrics/registry/metrics.yaml`) do not exist in GMP for the egress gateway: no per-actor CONNECT/authz metrics at all.
- Egress spans are absent from every distributed trace (suspend/resume traces jump from atelet straight to GCS).
- The container's local `:9090/metrics` endpoint serves the data fine (HTTP 200) — it is generated, then dropped at export.

Note this is **not** one of the documented telemetry gaps: `docs/metrics/substrate.yaml` `blind_spots` lists dns, the store, podcertcontroller, etc., but not egress.

## Root cause

`manifests/ate-install/atenet-egress.yaml` (and the sdsmint variant) give the ext-proc container `POD_NAME`/`POD_NAMESPACE` env but no `envFrom: ate-otel-config`, no `POD_UID`, and no `OTEL_RESOURCE_ATTRIBUTES` — all of which the ingress router container has. (The explicitly empty `--otlp-collector-address=` in the args is a red herring: that flag only controls Envoy-side tracing; the ext-proc's own SDK reads `$OTEL_EXPORTER_OTLP_ENDPOINT`.) The `ate-otel-config` header comment even says it is "shared by every control plane component" — egress just isn't in the consumer list.

## Fix

Add the same `envFrom` + `POD_UID` + `OTEL_RESOURCE_ATTRIBUTES` wiring the ingress router has, in both egress manifests. Change is ready on a fork branch and verified on a live install (SDK warnings drop to zero; `target_info` with `k8s.deployment.name=atenet-egress` appears in GMP within ~a minute of rollout):

https://github.com/git286/substrate/tree/fix-egress-otlp-export

Happy to send the PR.

One labeling quirk reviewers may want to weigh in on: because ingress and egress run the same binary, the egress telemetry lands under `job="atenet-router"` (service.name), distinguishable only by `k8s.deployment.name`. If that's considered confusing, a follow-up could set a distinct `service.name` for `--mode=egress`.

git286:
The egress ext-proc container never received the shared OTLP exporter settings: unlike every other control plane component, its Deployment did not consume ate-otel-config via envFrom, so OTEL_EXPORTER_OTLP_ENDPOINT was unset and the OTel SDK fell back to localhost:4317, where nothing listens. Every egress metric and span was silently dropped, with the SDK logging an export-failure warning every ~20s — on a live GKE install this was the largest warn/error source in the cluster, and the registry-listed ext_proc instruments simply did not exist in GMP for the egress gateway.

Give the ext-proc the same envFrom, POD_UID, and OTEL_RESOURCE_ATTRIBUTES wiring the ingress router has, in both the plain and sdsmint egress manifests. --otlp-collector-address stays explicitly empty: that flag only controls Envoy-side tracing, and a span per CONNECT tunnel is dataplane-volume noise; the comment now says so instead of leaving the empty value looking accidental. Also add atenet-egress to the consumer list in the ate-otel-config header comment.

Verified on a live install: after applying the env change the SDK export-failure warnings stop and target_info for the egress deployment appears in Managed Prometheus.

Fixes #1459

- [ ] Tests pass

### DIFF
--- a/manifests/ate-install/ate-otel-config.yaml
+++ b/manifests/ate-install/ate-otel-config.yaml
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 # Single source of truth for the OTLP exporter settings shared by every control
-# plane component (ate-api-server, ate-controller, atelet, atenet-router), which
-# consume it via `envFrom`. ate-controller additionally propagates these values
+# plane component (ate-api-server, ate-controller, atelet, atenet-router,
+# atenet-egress), which consume it via `envFrom`. ate-controller additionally propagates these values
 # to the ateom worker pods it creates, which have no other way to receive them,
 # and atenet-router's --otlp-collector-address defaults to the endpoint below,
 # so Envoy's own spans follow it too.
--- a/manifests/ate-install/atenet-egress-with-sdsmint.yaml
+++ b/manifests/ate-install/atenet-egress-with-sdsmint.yaml
@@ -958,6 +958,9 @@ spec:
         # of it; without this flag the router has no actor-identity roots and
         # denies every egress CONNECT with a 503.
         - --actor-identity-ca-file=/run/actor-id-ca-certs/ca.crt
+        # Disables Envoy-side tracing only (a span per CONNECT tunnel is
+        # dataplane-volume noise). The ext-proc's own OTel SDK exports to
+        # $OTEL_EXPORTER_OTLP_ENDPOINT from ate-otel-config below.
         - --otlp-collector-address=
         # The egress Envoy's admin listener is on 15000, not the 9901 the flag
         # defaults to (that is the ingress gateway's port). Without this the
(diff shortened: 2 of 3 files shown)


haiyanmeng (approved):

git286:
cc @annapendleton : would like to get this PR in before the cut

PR #1461, opened 2026-09-03
[Based on 1492] Resume with Golden+DurDir when ActorTemplate updated.

### LINKED ISSUE #477: Actor Image Upgrade Design

zoez7:
Before this PR, if actor's template has changed, the next resume cold boot the actor with DATA only. After this PR, the next resume will do:
* If the Actor was using FULL memory snapshot, the resume will use Golden+Data.
* If the Actor was using Golden+Data, the resume will still use Golden+Data.
* If the Actor was using Data only, the resume will still Use Data only.

If the DurDir does not exist, after the ActorTemplate update, the next Resume will use the golden snapshot of the new ActorTemplate.

Tested: Added a e2e test where the counter demo now reports a boot uuid that lives only in guest memory (restore from golden preserves it). The `TestUpdateTemplateLifecycle` e2e test uses it to verify the repointed actor restores template B's golden memory image.

Fixes #477 
- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_resume.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume.go
@@ -43,10 +43,12 @@ type resumeSnapshotSource struct {
 	// the actor holds a local snapshot, which takes precedence at restore).
 	SnapshotURI resources.SnapshotURI
 	// GoldenForDataSnapshotURI is the storage location of the ActorTemplate's
-	// golden snapshot. Populated only when the template's onResume
-	// configuration selects the golden snapshot for a pending data-only
-	// restore: the restore then combines the golden snapshot with the
-	// actor's data.
+	// golden snapshot. Populated only when the pending restore must ride the
+	// golden:
+	//   - the actor was repointed to a replacement template, so its snapshot's
+	//     guest state was captured under the replaced one, or
+	//   - the template's onResume configuration selects the golden for a
+	//     data-only restore.
 	GoldenForDataSnapshotURI resources.SnapshotURI
 	// WireScope is the scope of the restore operation, not of any stored
 	// snapshot:
@@ -166,13 +168,14 @@ func validateGoldenSnapshotScope(snapshot *ateapipb.ExternalSnapshot) error {
 //  1. An explicit boot request restores data-only from whatever snapshot
 //     exists (none: cold boot from the spec), never riding a golden.
 //  2. A repointed actor's snapshot holds guest state from the replaced
-//     template, so the restore carries the actor's durable data alone.
+//     template, so only its durable data survives; see
+//     resolveResumeScopeAfterUpdatedTemplate.
 //  3. A paused actor restores its local pause checkpoint at the template's
-//     onPause scope; a Data-scoped pause rides the golden when the onResume
-//     policy selects it.
-//  4. An actor on its own template restores its snapshot at the captured
-//     scope; a Data-scoped capture rides the golden when the onResume
-//     policy selects it.
+//     onPause scope; a Data-scoped checkpoint rides the golden when the
+//     onResume policy selects it.
+//  4. An actor with a durable snapshot restores it at the captured scope;
+//     a Data-scoped capture rides the golden when the onResume policy
+//     selects it.
 //  5. An actor with no snapshot restores the template's golden when one
 //     exists, and cold boots from the spec otherwise.
 func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resources.ActorRef, boot bool) (_ *ateapipb.Actor, _ *ateapipb.ActorTemplate, _ resumeSnapshotSource, err error) {
@@ -199,12 +202,11 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
 	if err != nil {
 		return nil, nil, src, err
 	}
-	external := actor.GetStatus().GetExternalSnapshot()
-	if uri := external.GetSnapshotUri(); uri != "" {
-		if src.SnapshotURI, err = resources.ParseSnapshotURI(uri); err != nil {
-			return nil, nil, src, status.Errorf(codes.DataLoss, "Actor %s external snapshot: %v", actorRef, err)
-		}
+	lastDurableSnapshot, err := resolveActorLatestDurableSnapshot(actor)
+	if err != nil {
+		return nil, nil, src, err
 	}
+	src.SnapshotURI = lastDurableSnapshot.uri
 
 	// Rule 1: an explicit boot request discards any captured guest state and
 	// carries the actor's durable data alone (none: cold boot from the spec).
@@ -214,11 +216,11 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
 	}
 
 	onResumeGolden := actorTemplate.GetSnapshotsConfig().GetOnResume().GetFromData() == ateapipb.ResumeSource_RESUME_SOURCE_GOLDEN
-	// Rule 2: An actor whose template was updated will cold boot.
-	builtOnTemplateUID := actor.GetStatus().GetCurrentActorTemplateUid()
-	if external.GetSnapshotUri() != "" && builtOnTemplateUID != "" && builtOnTemplateUID != actorTemplate.GetMetadata().GetUid() {
-		src.WireScope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
-		return actor, actorTemplate, src, nil
+
+	// Rule 2: the actor was repointed, so its snapshot's guest state belongs
+	// to the replaced template.
+	if isActorTemplateReplaced(actorTemplate, actor) {
+		return resolveResumeScopeAfterUpdatedTemplate(actor, actorTemplate, lastDurableSnapshot)
 	}
 
 	// Rule 3: a paused actor restores its local pause checkpoint, which the
@@ -226,9 +228,15 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
 	if actor.GetStatus().GetLocalSnapshotInfo() != nil {
 		if actorTemplate.GetSnapshotsConfig().GetOnPause() == ateapipb.SnapshotContentScope_SNAPSHOT_CONTENT_SCOPE_DATA {
 			if onResumeGolden {
-				if src.GoldenForDataSnapshotURI, err = resolveGoldenForDataResume(actorTemplate); err != nil {
+				golden, err := resolveGoldenSnapshot(actorTemplate)
+				if errors.Is(err, errNoGoldenSnapshot) {
+					return nil, nil, src, status.Error(codes.FailedPrecondition,
+						"a Golden data resume requires the ActorTemplate golden snapshot, which is not available")
+				}
+				if err != nil {
 					return nil, nil, src, err
 				}
+				src.GoldenForDataSnapshotURI = golden.uri
 				src.WireScope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA_ON_GOLDEN
 				return actor, actorTemplate, src, nil
 			}
@@ -242,12 +250,18 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
 	// Rule 4: the actor restores its own durable snapshot at the captured
 	// scope; the onResume policy only governs data-only restores, so a valid
 	// Full (or legacy Unspecified) capture restores from its own content.
-	if external.GetSnapshotUri() != "" {
-		if external.GetContentScope() == ateapipb.SnapshotContentScope_SNAPSHOT_CONTENT_SCOPE_DATA {
+	if lastDurableSnapshot.exists {
+		if lastDurableSnapshot.capturedScope == ateapipb.SnapshotContentScope_SNAPSHOT_CONTENT_SCOPE_DATA {
 			if onResumeGolden {
-				if src.GoldenForDataSnapshotURI, err = resolveGoldenForDataResume(actorTemplate); err != nil {
+				golden, err := resolveGoldenSnapshot(actorTemplate)
+				if errors.Is(err, errNoGoldenSnapshot) {
+					return nil, nil, src, status.Error(codes.FailedPrecondition,
+						"a Golden data resume requires the ActorTemplate golden snapshot, which is not available")
+				}
+				if err != nil {
 					return nil, nil, src, err
 				}
+				src.GoldenForDataSnapshotURI = golden.uri
 				src.WireScope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA_ON_GOLDEN
 				return actor, actorTemplate, src, nil
 			}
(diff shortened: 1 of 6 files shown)


cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -344,11 +345,19 @@ func (s *ServiceImpl) UpdateActor(ctx context.Context, actorRef resources.ActorR
...
 // snapshot data is laid out per the volumes and mount paths it was captured
 // with, so a different layout would restore it to the wrong places. The
-// volumes list must be identical, and containers present in both templates
+// volumes must be identical up to ordering (they are referenced by name, so
+// list order carries no meaning), and containers present in both templates
 // must keep identical mounts, order included; containers added or removed by
 // the new template are unconstrained.
 func validateTemplateVolumesUnchanged(oldTemplate, newTemplate *ateapipb.ActorTemplate) error {
-	if !slices.EqualFunc(oldTemplate.GetVolumes(), newTemplate.GetVolumes(), func(a, b *ateapipb.Volume) bool {
+	volumesByName := func(volumes []*ateapipb.Volume) []*ateapipb.Volume {
```
dberkov: this change seems unrelated to the description of the PR. Can it be moved to separate PR?

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -42,15 +42,22 @@ type resumeSnapshotSource struct {
...
+	// actor was repointed to a replacement template (its snapshot's guest
+	// state was captured under the replaced one), or the template's onResume
+	// configuration selects the golden for a data-only restore. Either way
+	// the restore combines the golden snapshot with the actor's durable data.
 	GoldenSnapshotURI resources.SnapshotURI
-	// TemplateReplaced is true when the snapshot's recorded template UID
-	// differs from the actor's current template.
-	TemplateReplaced bool
+	// WireScope is the scope of the restore OPERATION, not of any stored
+	// snapshot: DATA_ON_GOLDEN when the restore rides the golden snapshot,
```
dberkov: nit: could you please update the comment, so it will be easy to read the behavior per value, like:
* DATA_ON_GOLDEN  - bla bla bla
* DATA - bla bla bla
* etc..

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -178,6 +185,15 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+	targetTemplateUID := actorTemplate.GetMetadata().GetUid()
+	currActorTemplateUID := actor.GetStatus().GetCurrentActorTemplateUid()
+	if currActorTemplateUID != "" {
+		templateReplaced = currActorTemplateUID != targetTemplateUID
+	}
+	// capturedScope is the content scope the durable snapshot resolved below
+	// was taken with — the wire-scope fallback when no local snapshot exists.
+	var capturedScope ateapipb.SnapshotContentScope
 	if ref := actor.GetStatus().GetLatestSnapshot(); ref != nil {
 		snapshot, err := w.store.GetActorSnapshot(ctx, resources.ActorSnapshotRefFromObjectRef(ref))
```
dberkov: heads up - this code will need to be changed in future rebase, after redesign of the snapshots and tag will land today.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -209,48 +225,88 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+	// template, so regardless of scope the snapshot cannot restore its own
+	// guest state: the restore rides the new template's golden snapshot plus
+	// the actor's durable data — unless the new template's onResume policy
+	// selects cold boot, in which case the restore carries the durable data
+	// alone and needs no golden; an explicit boot request forces the same
+	// data-only path. For an actor on its own template, onResume
+	// applies only to data-only restores (Data pause scope, or a durable
+	// Data snapshot); valid Full snapshots restore from their own content.
+	dataOnly := false
+	if templateReplaced {
```
dberkov: nit: it is very difficult to read the code, could you please try avoiding elseif?

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -189,12 +205,12 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
-		snapshotTemplateUID := snapshot.GetStatus().GetActorTemplateUid()
-		src.TemplateReplaced = snapshotTemplateUID != "" && snapshotTemplateUID != actorTemplate.GetMetadata().GetUid()
+		capturedScope = snapshot.GetStatus().GetContentScope()
+		if currActorTemplateUID == "" {
+			// This is the codepath for actors that were created from a snapshot, and hasn't been resumed yet.
+			snapshotTemplateUID := snapshot.GetStatus().GetActorTemplateUid()
+			templateReplaced = snapshotTemplateUID != "" && snapshotTemplateUID != targetTemplateUID
+		}
 	} else if goldenRef := actorTemplate.GetStatus().GetGoldenSnapshotStatus().GetGoldenSnapshot(); goldenRef != nil && !boot {
 		snapshot, err := w.store.GetActorSnapshot(ctx, resources.ActorSnapshotRefFromObjectRef(goldenRef))
```
dberkov: If a stamped actor has no durable snapshot (`LatestSnapshot == nil` — reachable when a worker delete lands mid-suspend and `ensureSuspendedFinalized` commits SUSPENDED with no snapshot record), this fallback sets `src.SnapshotURI` to the template's *golden*. When that actor is also repointed and the new template's policy is `COLD_BOOT`, `dataOnly` becomes true below and the restore goes out as `Scope=DATA` carrying the golden's URI — the golden actor's durable-dir contents get restored as this actor's own data (pre-PR this state produced a FULL golden restore). With the policy unset, the same state emits a degenerate DATA_ON_GOLDEN with `SnapshotUri == GoldenSnapshotUri` (double manifest fetch). The repoint branch should probably not treat a golden-sourced `src.SnapshotURI` as actor data.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -189,12 +205,12 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
 		}
-		src.Scope = snapshot.GetStatus().GetContentScope()
-		// The snapshot records the template it was captured under; a
-		// different UID on the actor's current template means the actor was
-		// repointed since the capture.
-		snapshotTemplateUID := snapshot.GetStatus().GetActorTemplateUid()
-		src.TemplateReplaced = snapshotTemplateUID != "" && snapshotTemplateUID != actorTemplate.GetMetadata().GetUid()
+		capturedScope = snapshot.GetStatus().GetContentScope()
+		if currActorTemplateUID == "" {
+			// This is the codepath for actors that were created from a snapshot, and hasn't been resumed yet.
```
dberkov: This fallback trusts the snapshot's recorded `ActorTemplateUid`, but `workflow_suspend.go` (~line 410) stamps that field from the freshly-resolved template rather than `actor.Status.CurrentActorTemplateUid`. Suspend a paused actor across a template delete/recreate and the snapshot records the NEW UID over the OLD template's guest bytes; an actor created from that snapshot has no stamp, lands here, sees the UIDs equal, and full-restores the replaced template's guest state — the exact bug class this PR fixes. Stamping the snapshot from `CurrentActorTemplateUid` (when set) would close this and also let the clone-creation guard in actor.go reject it.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -209,48 +225,88 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+	// alone and needs no golden; an explicit boot request forces the same
+	// data-only path. For an actor on its own template, onResume
+	// applies only to data-only restores (Data pause scope, or a durable
+	// Data snapshot); valid Full snapshots restore from their own content.
+	dataOnly := false
+	if templateReplaced {
+		if boot || actorTemplate.GetSnapshotsConfig().GetOnResume().GetFromData() == ateapipb.ResumeSource_RESUME_SOURCE_COLD_BOOT {
+			dataOnly = true
+		} else if src.GoldenSnapshotURI, err = w.resolveGoldenSnapshotURI(ctx, actorTemplate,
+			"a repointed actor's resume requires the new ActorTemplate's golden snapshot, which is not available"); err != nil {
```
dberkov: A terminally failed golden capture makes this a permanent failure: the template reconciler treats a non-empty `ErrorMessage` as terminal and never retries, so every default resume of a repointed (or same-name-recreated) template's actors returns FailedPrecondition — including router-driven resumes on live traffic — until an operator passes `boot=true` per-resume, edits the onResume policy, or recreates the template. It also fires transiently on every repoint-then-resume during the async golden-capture window. Consider checking golden status at repoint time in UpdateActor to surface this early, retrying failed goldens, or at least documenting the escape hatch.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -209,48 +225,88 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+	// the actor's durable data — unless the new template's onResume policy
+	// selects cold boot, in which case the restore carries the durable data
+	// alone and needs no golden; an explicit boot request forces the same
+	// data-only path. For an actor on its own template, onResume
+	// applies only to data-only restores (Data pause scope, or a durable
+	// Data snapshot); valid Full snapshots restore from their own content.
+	dataOnly := false
+	if templateReplaced {
+		if boot || actorTemplate.GetSnapshotsConfig().GetOnResume().GetFromData() == ateapipb.ResumeSource_RESUME_SOURCE_COLD_BOOT {
+			dataOnly = true
```
dberkov: Unset `fromData` and explicit `COLD_BOOT` diverge here — the first and only place they differ. The proto contract says UNSPECIFIED selects the documented default, and docs/glossary.md documents ColdBoot as that default; yet after a repoint the unset template hard-fails without a golden while the explicit-ColdBoot one resumes data-only, and when a golden exists they restore different guest state (golden memory vs cold boot). Either treat UNSPECIFIED as ColdBoot in this branch, or update the `from_data` proto comment and the glossary to document the new repoint default.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -178,6 +185,15 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
 	if err != nil {
 		return nil, nil, src, err
 	}
+	templateReplaced := false
+	targetTemplateUID := actorTemplate.GetMetadata().GetUid()
+	currActorTemplateUID := actor.GetStatus().GetCurrentActorTemplateUid()
+	if currActorTemplateUID != "" {
+		templateReplaced = currActorTemplateUID != targetTemplateUID
```
dberkov: Behavioral note: stamp-based detection marks a paused actor repointed even when its template is deleted and re-applied byte-identically (delete+create is the only way to edit an immutable template). Previously the local FULL pause checkpoint restored with memory intact; now the paused guest memory is silently discarded in favor of the recreated template's golden — or resume fails FailedPrecondition until the fresh golden capture completes. If intended (it is test-encoded), this deserves a release note; otherwise consider a spec-equivalence escape hatch.

internal/e2e/suites/demo/updatetemplate_test.go
```diff
@@ -190,8 +210,56 @@ func runUpdateTemplateTestCase(t *testing.T, onCommit ateapipb.SnapshotContentSc
...
+	//
+	refActorID := "update-ref-" + nsObj.Name
+	t.Logf("Creating reference Actor %q from template B to observe B's golden boot uuid...", refActorID)
+	if _, err := clients.SubstrateAPI.CreateActor(ctx, &ateapipb.CreateActorRequest{Actor: &ateapipb.Actor{
+		Metadata:      &ateapipb.ResourceMetadata{Atespace: demoAtespace, Name: refActorID},
+		ActorTemplate: refB,
+	}}); err != nil {
+		t.Fatalf("failed to create reference Actor: %v", err)
+	}
+	defer func() {
```
dberkov, line 237: Both deferred DeleteActor calls (this one and the main actor's at ~line 101) lack `AnyState` and discard the error, but the test now ends with both actors RUNNING (Step 9 resumes the main actor; the reference actor is never suspended). `workflow_delete.go` only permits deletion from SUSPENDED/CRASHED without `AnyState`, so cleanup silently fails with FailedPrecondition and leaks two actor records per subtest into the shared 'demo' atespace (`actors.atespace` is ON DELETE RESTRICT, so leaked rows block any future DeleteAtespace). Use the suspend-then-delete pattern from demo_test.go:100 or `AnyState: true`, and log the error.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -209,48 +225,88 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+
 	return actor, actorTemplate, src, nil
 }
 
+// resolveGoldenSnapshotURI resolves the storage location of the template's
+// golden snapshot, validating that it can serve as the guest half of a
+// combined restore. missingMsg is the FailedPrecondition message when the
+// template records no golden snapshot.
+func (w *ActorWorkflow) resolveGoldenSnapshotURI(ctx context.Context, actorTemplate *ateapipb.ActorTemplate, missingMsg string) (resources.SnapshotURI, error) {
+	goldenRef := actorTemplate.GetStatus().GetGoldenSnapshotStatus().GetGoldenSnapshot()
```
dberkov: The golden-fallback branch above (~lines 215-228) still inlines this exact sequence (GetActorSnapshot, ErrNotFound→DataLoss, error wrap, validateGoldenSnapshotScope, ParseSnapshotURI), and on the fallback+repointed path this helper then re-fetches, re-validates, and re-parses the same golden ActorSnapshot — a redundant second store round trip on the resume path. Having the fallback branch call the helper (returning the snapshot or its ContentScope) and the repoint branch reuse the already-resolved URI would remove both the duplication and the double fetch.

cmd/ateapi/internal/controlapi/actor.go
zoez7: Removed it.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Done.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Thanks, I resolved the merge conflicts.

PR #1462, opened 2026-09-03
ateapi: trace PostgreSQL statements in the store

### LINKED ISSUE #1455: ateapi: postgres store has no trace instrumentation - DB time is invisible in every trace
## Summary

The ateapi postgres store (`cmd/ateapi/internal/store/atepg/`) creates its `pgxpool` without any tracer hook, so no span is ever emitted for database work. Every RPC trace ends at the gRPC server span produced by the `otelgrpc` stats handler; the time spent in PostgreSQL is indistinguishable from handler/auth overhead.

## Observed

On a live GKE install with tracing enabled end to end:

- `kubectl ate get actor <name> -a <atespace> --trace` produces a trace with **exactly one span**: `ateapi.Control/GetActor` (~97 ms in our run). For a single-row read that is slow enough to want a breakdown, and the trace offers none.
- This is not specific to reads: rich mutation traces (25-span suspend, 33-span resume) contain `step.*`, atelet, ateom, and GCS client spans — but zero postgres spans, despite every one of those flows doing multiple DB round-trips (actor load, state transitions, outbox).

## Root cause

- `cmd/ateapi/internal/store/atepg/atepg.go` has no OTel/tracing references at all; the pool is built from `pgxpool.ParseConfig` with `ConnConfig.Tracer` unset.
- The only vendored OTel instrumentation libraries are `otelgrpc` and `net/http` — there is no pgx instrumentation in the dependency tree.

## Suggested fix

Attach a tracer via pgx v5's native hook: set `config.ConnConfig.Tracer` on the pool config, either with [`otelpgx`](https://github.com/exaring/otelpgx) or a small in-tree `pgx.QueryTracer` that opens spans from the request context. Spans should parent under the active RPC span so `GetActor` shows `db.query` children with statement names.

## Related test-documentation nit

A bug-bash test we ran expects `kubectl ate get actor --trace` to also capture a "worker handshake". That expectation is wrong by design: `GetActor` is a pure store read (`cmd/ateapi/internal/controlapi/actor.go` → `s.store.GetActor`), and worker spans (`step.AssignWorker`, `atelet.AteomHerder/*`) only appear on mutation traces. Worth correcting wherever that test recipe lives, but the actionable gap in this issue is the missing DB spans.

git286:
The store's pgx pools carried no tracer, so no span was ever emitted for database work: a traced GetActor arrived as a single gRPC server span with the PostgreSQL time indistinguishable from handler overhead, and even the multi-span suspend/resume traces showed step, atelet and snapshot spans but nothing for the many store round-trips between them. The store blind spot in docs/metrics/substrate.yaml describes the metrics half of this; the trace half is what this change closes.

Attach a QueryTracer to the pool config (the watch pool inherits it via Copy). Each statement becomes a client span named after its leading keyword (db.SELECT, db.UPDATE, ...) with the parameterized statement text as db.query.text — arguments never appear in it. Spans only join an existing trace: statements from background work (outbox polling, lease maintenance) carry no surrounding span, and opening a root span for each would flood the backend with single-span traces. pgx.ErrNoRows leaves the span unmarked, since the store maps it to NotFound as an expected lookup outcome.

Fixes #1455

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/store/atepg/atepg.go
+++ b/cmd/ateapi/internal/store/atepg/atepg.go
@@ -144,6 +144,8 @@ func poolConfig(dsn string) (*pgxpool.Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("parsing PostgreSQL connection string: %w", err)
 	}
+	// Per-statement trace spans; the watch pool inherits this through Copy().
+	cfg.ConnConfig.Tracer = queryTracer{}
 	usesTLS := cfg.ConnConfig.TLSConfig != nil
 	for _, fallback := range cfg.ConnConfig.Fallbacks {
 		usesTLS = usesTLS || fallback.TLSConfig != nil
--- a/cmd/ateapi/internal/store/atepg/atepg.go
+++ b/cmd/ateapi/internal/store/atepg/atepg.go
@@ -35,6 +35,7 @@ import (
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/jackc/pgx/v5/pgxpool"
+	"go.opentelemetry.io/otel"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
@@ -145,7 +146,7 @@ func poolConfig(dsn string) (*pgxpool.Config, error) {
 		return nil, fmt.Errorf("parsing PostgreSQL connection string: %w", err)
 	}
 	// Per-statement trace spans; the watch pool inherits this through Copy().
-	cfg.ConnConfig.Tracer = queryTracer{}
+	cfg.ConnConfig.Tracer = newQueryTracer(otel.GetTracerProvider(), cfg.ConnConfig)
 	usesTLS := cfg.ConnConfig.TLSConfig != nil
 	for _, fallback := range cfg.ConnConfig.Fallbacks {
 		usesTLS = usesTLS || fallback.TLSConfig != nil
--- a/cmd/ateapi/internal/store/atepg/tracing.go
+++ b/cmd/ateapi/internal/store/atepg/tracing.go
@@ -1,0 +1,72 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package atepg
+
+import (
+	"context"
+	"errors"
+	"strings"
+
+	"github.com/jackc/pgx/v5"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// queryTracer is a pgx QueryTracer that opens one client span per statement,
+// so an RPC trace shows where its time went inside PostgreSQL. Statements are
+// parameterized ($1, $2, ...), so db.query.text carries no argument values.
+type queryTracer struct{}
+
+var _ pgx.QueryTracer = queryTracer{}
+
+func (queryTracer) TraceQueryStart(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryStartData) context.Context {
+	// Join existing traces only. Statements issued from background work
+	// (outbox polling, lease maintenance) carry no span, and opening a root
+	// span for each would flood the backend with single-span traces.
+	if !trace.SpanContextFromContext(ctx).IsValid() {
+		return ctx
+	}
+	ctx, _ = otel.Tracer("atepg").Start(ctx, querySpanName(data.SQL),
+		trace.WithSpanKind(trace.SpanKindClient),
+		trace.WithAttributes(
+			attribute.String("db.system.name", "postgresql"),
+			attribute.String("db.query.text", data.SQL),
+		))
+	return ctx
+}
+
+func (queryTracer) TraceQueryEnd(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryEndData) {
+	span := trace.SpanFromContext(ctx)
+	// pgx.ErrNoRows is an expected lookup outcome (the store maps it to
+	// NotFound), not a query failure.
+	if data.Err != nil && !errors.Is(data.Err, pgx.ErrNoRows) {
+		span.RecordError(data.Err)
+		span.SetStatus(codes.Error, data.Err.Error())
+	}
+	span.End()
+}
+
+// querySpanName is "db." plus the statement's leading keyword ("db.SELECT",
+// "db.INSERT", ...): stable low-cardinality names that group by statement
+// kind, with the full text in db.query.text.
+func querySpanName(sql string) string {
+	fields := strings.Fields(sql)
+	if len(fields) == 0 {
+		return "db.query"
+	}
+	return "db." + strings.ToUpper(fields[0])
+}
--- a/cmd/ateapi/internal/store/atepg/tracing.go
+++ b/cmd/ateapi/internal/store/atepg/tracing.go
@@ -16,57 +16,153 @@ package atepg
 
 import (
 	"context"
-	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/jackc/pgx/v5"
-	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	semconv "go.opentelemetry.io/otel/semconv/v1.40.0"
 	"go.opentelemetry.io/otel/trace"
 )
 
 // queryTracer is a pgx QueryTracer that opens one client span per statement,
 // so an RPC trace shows where its time went inside PostgreSQL. Statements are
 // parameterized ($1, $2, ...), so db.query.text carries no argument values.
-type queryTracer struct{}
-
-var _ pgx.QueryTracer = queryTracer{}
-
-func (queryTracer) TraceQueryStart(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryStartData) context.Context {
-	// Join existing traces only. Statements issued from background work
-	// (outbox polling, lease maintenance) carry no span, and opening a root
-	// span for each would flood the backend with single-span traces.
-	if !trace.SpanContextFromContext(ctx).IsValid() {
-		return ctx
-	}
-	ctx, _ = otel.Tracer("atepg").Start(ctx, querySpanName(data.SQL),
-		trace.WithSpanKind(trace.SpanKindClient),
-		trace.WithAttributes(
-			attribute.String("db.system.name", "postgresql"),
-			attribute.String("db.query.text", data.SQL),
-		))
-	return ctx
+//
+// Spans follow the OpenTelemetry database semantic conventions: the span name
+// is the query summary ("SELECT actors", "UPDATE workers", "commit"), and the
+// connection-level attributes are computed once from the pool configuration.
+type queryTracer struct {
+	tracer trace.Tracer
+	// attrs holds db.system.name, server.address, server.port and
+	// db.namespace, which are the same for every statement on the pool.
+	attrs []attribute.KeyValue
 }
 
-func (queryTracer) TraceQueryEnd(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryEndData) {
-	span := trace.SpanFromContext(ctx)
-	// pgx.ErrNoRows is an expected lookup outcome (the store maps it to
-	// NotFound), not a query failure.
-	if data.Err != nil && !errors.Is(data.Err, pgx.ErrNoRows) {
+var _ pgx.QueryTracer = (*queryTracer)(nil)
+
+// newQueryTracer builds the tracer for pools opened from cc. The tracer is
+// resolved once here rather than per statement.
+func newQueryTracer(tp trace.TracerProvider, cc *pgx.ConnConfig) *queryTracer {
+	return &queryTracer{
+		tracer: tp.Tracer("atepg"),
+		attrs: []attribute.KeyValue{
+			semconv.DBSystemNamePostgreSQL,
+			semconv.ServerAddress(cc.Host),
+			semconv.ServerPort(int(cc.Port)),
+			semconv.DBNamespace(cc.Database),
+		},
+	}
+}
+
+// querySpanKey marks a context whose statement span was opened by
+// TraceQueryStart, so TraceQueryEnd never ends a span it did not start.
+type querySpanKey struct{}
+
+func (t *queryTracer) TraceQueryStart(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryStartData) context.Context {
+	// Join sampled traces only. Statements issued from background work
+	// (outbox polling, lease maintenance) carry no span, and opening a root
+	// span for each would flood the backend with single-span traces. An
+	// unsampled parent is skipped too: every sampler serverboot installs is
+	// ParentBased, so its children could never be sampled, and skipping
+	// them saves the non-recording span and context wrap per statement.
+	if !trace.SpanContextFromContext(ctx).IsSampled() {
+		return ctx
+	}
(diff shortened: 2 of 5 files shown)


cmd/ateapi/internal/store/atepg/tracing.go
```diff
@@ -0,0 +1,72 @@
...
+var _ pgx.QueryTracer = queryTracer{}
+
+func (queryTracer) TraceQueryStart(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryStartData) context.Context {
+	// Join existing traces only. Statements issued from background work
+	// (outbox polling, lease maintenance) carry no span, and opening a root
+	// span for each would flood the backend with single-span traces.
+	if !trace.SpanContextFromContext(ctx).IsValid() {
+		return ctx
+	}
+	ctx, _ = otel.Tracer("atepg").Start(ctx, querySpanName(data.SQL),
```
juli4n: Is an error here an impossible condition? Are we ok swallowing the error?

git286: The discarded value is the span, not an error. Start cannot fail. Its signature is Start(ctx, name, ...) (context.Context, Span). The OpenTelemetry API is built so tracing never breaks the app: if sampling is off you just get an inert span back, never an error.

We discard the span because Start also stores a copy inside the returned context, and that copy is the one we need. The span gets closed in a different function: pgx carries our returned context through the query and passes it to TraceQueryEnd, which retrieves the span with trace.SpanFromContext and ends it. A local span variable would go out of scope as soon as TraceQueryStart returns.

cmd/ateapi/internal/store/atepg/tracing.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (queryTracer) TraceQueryEnd(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryEndData) {
+	span := trace.SpanFromContext(ctx)
+	// pgx.ErrNoRows is an expected lookup outcome (the store maps it to
+	// NotFound), not a query failure.
+	if data.Err != nil && !errors.Is(data.Err, pgx.ErrNoRows) {
+		span.RecordError(data.Err)
+		span.SetStatus(codes.Error, data.Err.Error())
+	}
```
krisztianfekete: Is it possible that this is dead code?

cmd/ateapi/internal/store/atepg/tracing_test.go
```diff
@@ -0,0 +1,135 @@
...
+	ctx, parent := otel.Tracer("test").Start(context.Background(), "parent")
+	qt := queryTracer{}
+
+	// A real failure marks the span.
+	qctx := qt.TraceQueryStart(ctx, nil, pgx.TraceQueryStartData{SQL: "UPDATE actors SET state = $1"})
+	qt.TraceQueryEnd(qctx, nil, pgx.TraceQueryEndData{Err: errors.New("deadlock detected")})
+
+	// ErrNoRows is an expected lookup outcome and must not mark the span.
+	qctx = qt.TraceQueryStart(ctx, nil, pgx.TraceQueryStartData{SQL: "SELECT id FROM actors"})
+	qt.TraceQueryEnd(qctx, nil, pgx.TraceQueryEndData{Err: pgx.ErrNoRows})
```
krisztianfekete: See https://github.com/agent-substrate/substrate/pull/1462/changes#r3949759419 also

cmd/ateapi/internal/store/atepg/tracing_test.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// withRecordingTracer installs a recording global tracer provider for the
+// test and returns the recorder; the previous provider is restored on cleanup.
+func withRecordingTracer(t *testing.T) *tracetest.SpanRecorder {
+	t.Helper()
+	sr := tracetest.NewSpanRecorder()
+	tp := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(sr))
+	prev := otel.GetTracerProvider()
+	otel.SetTracerProvider(tp)
```
krisztianfekete: We have a convention of not swapping the global provider in our tests.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -144,6 +144,8 @@ func poolConfig(dsn string) (*pgxpool.Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("parsing PostgreSQL connection string: %w", err)
 	}
+	// Per-statement trace spans; the watch pool inherits this through Copy().
+	cfg.ConnConfig.Tracer = queryTracer{}
```
krisztianfekete: Can we add a simple test to assert what the comment states?

cmd/ateapi/internal/store/atepg/tracing.go
```diff
@@ -0,0 +1,72 @@
...
+
+// querySpanName is "db." plus the statement's leading keyword ("db.SELECT",
+// "db.INSERT", ...): stable low-cardinality names that group by statement
+// kind, with the full text in db.query.text.
+func querySpanName(sql string) string {
+	fields := strings.Fields(sql)
+	if len(fields) == 0 {
+		return "db.query"
+	}
+	return "db." + strings.ToUpper(fields[0])
```
krisztianfekete: Can we have `db.query.summary`, i.e. `SELECT actors` to match semconv's preferences?

See https://opentelemetry.io/docs/specs/semconv/db/database-spans/

cmd/ateapi/internal/store/atepg/tracing.go
```diff
@@ -0,0 +1,72 @@
...
+	return ctx
+}
+
+func (queryTracer) TraceQueryEnd(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryEndData) {
+	span := trace.SpanFromContext(ctx)
+	// pgx.ErrNoRows is an expected lookup outcome (the store maps it to
+	// NotFound), not a query failure.
+	if data.Err != nil && !errors.Is(data.Err, pgx.ErrNoRows) {
+		span.RecordError(data.Err)
+		span.SetStatus(codes.Error, data.Err.Error())
```
krisztianfekete, line 102: I think we should have access to `SQLSTATE` here, if so, can we pull it into `db.response.status_code` + `error.type`?

cmd/ateapi/internal/store/atepg/tracing.go
```diff
@@ -0,0 +1,72 @@
...
+	// Join existing traces only. Statements issued from background work
+	// (outbox polling, lease maintenance) carry no span, and opening a root
+	// span for each would flood the backend with single-span traces.
+	if !trace.SpanContextFromContext(ctx).IsValid() {
+		return ctx
+	}
+	ctx, _ = otel.Tracer("atepg").Start(ctx, querySpanName(data.SQL),
+		trace.WithSpanKind(trace.SpanKindClient),
+		trace.WithAttributes(
+			attribute.String("db.system.name", "postgresql"),
```
krisztianfekete: Can we use `semconv.DBSystemNamePostgreSQL` / `semconv.DBQueryTextKey` instead? Also missing while we're in here: 
 - `server.address` (Required),
 - `db.namespace`,
 - `db.operation.name`,
 - `db.collection.name`

cmd/ateapi/internal/store/atepg/tracing.go
```diff
@@ -0,0 +1,72 @@
...
+
+var _ pgx.QueryTracer = queryTracer{}
+
+func (queryTracer) TraceQueryStart(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryStartData) context.Context {
+	// Join existing traces only. Statements issued from background work
+	// (outbox polling, lease maintenance) carry no span, and opening a root
+	// span for each would flood the backend with single-span traces.
+	if !trace.SpanContextFromContext(ctx).IsValid() {
+		return ctx
+	}
```
krisztianfekete: Can we use ` IsSampled()` here? Should be identical but cheaper, right?

cmd/ateapi/internal/store/atepg/tracing.go
krisztianfekete: Can we cache the tracer in a struct field instead of resolving it per statement?

cmd/ateapi/internal/store/atepg/tracing.go
```diff
@@ -0,0 +1,72 @@
...
+// querySpanName is "db." plus the statement's leading keyword ("db.SELECT",
+// "db.INSERT", ...): stable low-cardinality names that group by statement
+// kind, with the full text in db.query.text.
+func querySpanName(sql string) string {
+	fields := strings.Fields(sql)
+	if len(fields) == 0 {
+		return "db.query"
+	}
+	return "db." + strings.ToUpper(fields[0])
+}
```
krisztianfekete: What do you think about skipping tx control statements, or wrap the tx in one INTERNAL span and nest the statements? Otherwise this might get a bit noisy as we are getting regular spans for  BEGIN/INSERT/COMMIT/etc.

krisztianfekete (commented):
Thanks, added some comments around mainly semconv and performance!

cmd/ateapi/internal/store/atepg/tracing.go
git286: You are right, it is dead code. pgx never passes ErrNoRows to TraceQueryEnd. It creates that error for the caller after the rows are already closed, so the tracer only sees the real error from the result reader. I removed the check and the test for it. I also added a test against a real PostgreSQL that does a lookup miss and checks the span stays unmarked.

cmd/ateapi/internal/store/atepg/tracing_test.go
git286: Removed together with the ErrNoRows branch.

cmd/ateapi/internal/store/atepg/tracing_test.go
git286: Fixed. The tracer now takes a TracerProvider in its constructor. Connect passes the global one and the tests pass a local recording one, same as the controlapi tests. All tracer tests run with t.Parallel now.

cmd/ateapi/internal/store/atepg/atepg.go
git286: Added TestPoolConfigTracerSharedWithWatchPool. It calls poolConfig, copies the config and checks both point to the same tracer.

cmd/ateapi/internal/store/atepg/tracing.go
git286: Done. Span name is now the summary like "SELECT actors" or "UPDATE workers". The store uses one hand written statement per call so a small keyword scan is enough. It takes the first keyword and the table after FROM, INTO, UPDATE or TABLE. If the statement has a JOIN, a subquery or no table, only the keyword is used, as the spec asks. I kept the original casing so BEGIN and COMMIT from pgx show as "begin" and "commit". The test runs the scan over the real statements from the store.

cmd/ateapi/internal/store/atepg/tracing.go
git286: Yes, we have it. For a PgError I now set db.response.status_code and error.type to the SQLSTATE. For other errors like a canceled context, error.type is the Go type name. The PostgreSQL test checks a missing table gives 42P01.

cmd/ateapi/internal/store/atepg/tracing.go
git286: Switched to the semconv v1.40.0 constants. Added server.address, server.port, db.namespace, db.operation.name, db.collection.name and db.query.summary. The connection level ones are computed once from the pool config, not per statement, because conn.Config() copies the whole config on every call. Small note, the spec lists server.address as Recommended, not Required, but it is cheap so I added it.

cmd/ateapi/internal/store/atepg/tracing.go
git286: Good idea. Changed to IsSampled. It is not exactly the same check but it is safe here because every sampler in serverboot is ParentBased, so a child of an unsampled parent can never be sampled. It saves the non recording span and the context wrap on every statement. I left a comment in the code about the ParentBased assumption. I also made TraceQueryEnd only end a span that TraceQueryStart opened, so an unsampled parent is never touched.

cmd/ateapi/internal/store/atepg/tracing.go
git286: Done. The tracer is resolved once in the constructor and kept in a struct field. This also gives the tests a place to inject their provider.

cmd/ateapi/internal/store/atepg/tracing.go
git286: I would keep them. In pgx, Begin and Commit are normal Exec calls, so each one is a real round trip to the server. COMMIT is where the WAL flush waits, so a lot of the PostgreSQL time we want to see in the trace lands in that span. If we skip it, the gap shows up as unexplained time in the parent span again.

The noise is also smaller than it looks. Only three transactions in atepg.go run on the request path. The outbox ones run in background work with no parent span, so they never produce spans.

I like the idea of one INTERNAL span per transaction with the statements nested under it. That needs a small helper around pool.Begin in the store, so I would do it in a follow up PR to keep this one focused on the pgx tracer. Is that ok with you?

cmd/ateapi/internal/store/atepg/tracing.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	at := -1
+	for i, f := range fields[1:] {
+		switch {
+		case strings.EqualFold(f, "JOIN"):
+			return operation, ""
+		case at < 0 && strings.EqualFold(f, marker):
+			at = i + 2
+		}
+	}
```
krisztianfekete, line 156: This takes the first `FROM`, which is the one inside the subquery when the outer `SELECT` has no table.

If there is a nested `SELECT` after the first word, can we return no collection? Then both become keyword only.

krisztianfekete (commented):
Can you please add a few lines to observability docs and maybe docs/dev/best-practices/tracing.md? E.g. saying that the store now emits one span per statement, that it follows the component sampler and has no separate on/off switch, and that begin and commit appear as spans.

cmd/ateapi/internal/store/atepg/tracing.go
krisztianfekete: Sure, would you mind opening an issue to track this?

cmd/ateapi/internal/store/atepg/tracing.go
git286: Good catch, fixed. Now a SELECT that comes before the FROM means the FROM belongs to a subquery, so we return only the keyword. SELECT EXISTS(SELECT 1 FROM t) and SELECT (SELECT xid FROM t) become just SELECT. I check only before the FROM, so SELECT proto FROM actors WHERE name IN (SELECT ...) still gives actors and INSERT INTO t SELECT ... still gives t. Added these cases to the test.

cmd/ateapi/internal/store/atepg/tracing.go
git286: Sure, opened #1537.

git286:
Added the docs. One paragraph in the Tracing section of docs/observability.md and one in the sampling section of docs/dev/best-practices/tracing.md. They say the store emits one client span per statement named by the query summary, that begin and commit show up as spans, that the spans only join a trace that is already sampled so they follow the ateapi sampler and have no switch of their own, and that background work like outbox polling produces no spans.

The subquery case is fixed and the follow up for transaction spans is #1537.

The last run-tests job failed on TestActorLifecycle_WithExternalVolumes, which is the flaky test in #1533. This push runs it again.

cmd/ateapi/internal/store/atepg/tracing.go
```diff
@@ -0,0 +1,180 @@
...
+		trace.WithSpanKind(trace.SpanKindClient),
+		trace.WithAttributes(attrs...))
+	return context.WithValue(ctx, querySpanKey{}, span)
+}
+
+func (*queryTracer) TraceQueryEnd(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryEndData) {
+	span, ok := ctx.Value(querySpanKey{}).(trace.Span)
+	if !ok {
+		return
+	}
```
krisztianfekete, line 97: This guard has no test, I deleted it, replaced it with `trace.SpanFromContext(ctx)`, and every test still passed. Can you add the `RecordOnly` test so it is covered?

krisztianfekete (commented):
One small thing, otherwise LGTM!

cmd/ateapi/internal/store/atepg/tracing.go
git286: Added TestQueryTracerLeavesRecordOnlyParentOpen. It uses a small sampler that returns RecordOnly, so the parent is recording but not sampled. The statement is skipped, and the test checks the parent is still recording after TraceQueryEnd and that nothing was ended. I tried your change with SpanFromContext and the test fails on both checks, so the guard is covered now.

krisztianfekete (approved):

PR #1465, opened 2026-09-04
ate-api-server: Tell the caller about issuer problems

ahmedtd:
If someone calls us with a JWT, and we don't trust it because the issuer is not configured in the authentication configuration file, it's fine (and very helpful for installers) if we tell them a little more detail.


maxsmythe (approved):
LGTM

PR #1467, opened 2026-09-04
atepg: Guardrail to disallow changes to atespace tables that limit future partitioning

jpbetz:
This adds a test that serves as a guardrail to prevent changes to the atespace-scoped tables that would make it difficult or impossible to partition it in the future.

- [x] Tests pass

cc @bowei @BenTheElder @juli4n @shrutiyam-glitch @thockin

### DIFF
--- a/cmd/ateapi/internal/store/atepg/partition_test.go
+++ b/cmd/ateapi/internal/store/atepg/partition_test.go
@@ -1,0 +1,273 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package atepg
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"slices"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store"
+	"github.com/agent-substrate/substrate/cmd/ateapi/internal/store/storecontract"
+)
+
+// TestActorsTablePartitionable exists to keep it possible to partition the
+// actors table by atespace or by name later, with the other atespace-scoped
+// tables partitioned alongside it by atespace. It runs the store contract
+// suite against a copy of the schema partitioned that way, and fails on any
+// schema change or query that would not work in that layout.
+func TestActorsTablePartitionable(t *testing.T) {
+	// exemptions are statements allowed to read every partition even though
+	// their result lives in one. Do not add one without discussion and
+	// agreement in the community.
+	var exemptions []string
+
+	// spansPartitions are the statements whose result covers every partition
+	// by definition, under each partition key.
+	spansPartitions := map[string][]string{
+		// A global list walks every atespace.
+		"atespace": {globalList("actors"), globalList("actor_templates"), globalList("actor_snapshots")},
+		// A global list walks every name, and a list within one atespace
+		// spans every name hash.
+		"name": {globalList("actors"), scopedActorList},
+	}
+
+	t.Run("by atespace", func(t *testing.T) {
+		runContractSuitePartitioned(t, "atespace", atespaceScopedTables, slices.Concat(spansPartitions["atespace"], exemptions))
+	})
+	t.Run("by name", func(t *testing.T) {
+		runContractSuitePartitioned(t, "name", []string{"actors"}, slices.Concat(spansPartitions["name"], exemptions))
+	})
+
+	t.Run("rejects a unique index that omits the key", func(t *testing.T) {
+		pool := migratedPool(t, "partitioned-unique")
+		// This also rules out foreign keys onto uid, which need this index.
+		if _, err := pool.Exec(t.Context(), `CREATE UNIQUE INDEX actors_uid_key ON actors (uid)`); err != nil {
+			t.Fatal(err)
+		}
+		err := partitionTable(t.Context(), pool, "actors", "atespace")
+		if err == nil || !strings.Contains(err.Error(), "must include all partitioning columns") {
+			t.Fatalf("partitionTable error = %v, want unique index rejection", err)
+		}
+		t.Log(err)
+	})
+	t.Run("rejects a query that omits the key", func(t *testing.T) {
+		var got string
+		pool := partitionedPool(t, "partitioned-fanout", "atespace", []string{"actors"})
+		check := newFanOutCheck("atespace", []string{"actors"}, nil, pool, func(format string, args ...any) { got = fmt.Sprintf(format, args...) })
+		var n int
+		if err := openPool(t, "partitioned-fanout", check).QueryRow(t.Context(), `SELECT count(*) FROM actors WHERE uid = $1`, "u1").Scan(&n); err != nil {
+			t.Fatal(err)
+		}
+		if !strings.Contains(got, "[actors_p0 actors_p1]") {
+			t.Fatalf("fan-out check reported %q, want the uid lookup reading both partitions", got)
+		}
+		t.Log(got)
+	})
+}
+
+// atespaceScopedTables hold one atespace's resources and partition together.
+var atespaceScopedTables = []string{"actors", "actor_egress_policies", "actor_templates", "actor_snapshots", "actor_snapshot_tags"}
+
+// globalList is the statement that lists table across every atespace.
+func globalList(table string) string {
+	return normalizeSQL(`
+		SELECT atespace, name, proto FROM ` + table + `
+		WHERE $1::text IS NULL OR (atespace, name) > ($1, $2)
+		ORDER BY atespace, name
+		LIMIT $3`)
+}
+
+// scopedActorList is the statement that lists one atespace's actors.
+var scopedActorList = normalizeSQL(`
+	SELECT name, proto FROM actors
+	WHERE atespace = $1 AND ($2::text IS NULL OR name > $2)
+	ORDER BY name
+	LIMIT $3`)
+
+// runContractSuitePartitioned runs the store contract suite with tables
+// partitioned on key and fails on any statement, other than the allowed
+// ones, whose plan reads more than one partition of a table.
+func runContractSuitePartitioned(t *testing.T, key string, tables []string, allowed []string) {
+	schema := "partitioned-by-" + key
+	pool := partitionedPool(t, schema, key, tables)
+	check := newFanOutCheck(key, tables, allowed, pool, t.Errorf)
+	traced := openPool(t, schema, check)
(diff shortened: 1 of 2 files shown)


EItanya:
@iplay88keys as well

cmd/ateapi/internal/store/atepg/partition_test.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+		if !strings.Contains(got, "[actors_p0 actors_p1]") {
+			t.Fatalf("fan-out check reported %q, want the uid lookup reading both partitions", got)
+		}
+		t.Log(got)
+	})
+}
+
+// atespaceScopedTables hold one atespace's resources and partition together.
+var atespaceScopedTables = []string{"actors", "actor_egress_policies", "actor_templates", "actor_snapshots", "actor_snapshot_tags"}
```
BenTheElder: 🤖 **should-fix** 🟡 – Coverage is opt-in. This is the fourth hand-maintained copy of the table list (with `clearAll`, `rejectUnversionedSubstrateSchema`, and the doc bullet). A future migration adding an atespace-scoped table with a `uid`-only unique index passes silently unless the author also edits this var.

The catalog already knows the set: `SELECT table_name FROM information_schema.columns WHERE table_schema = current_schema() AND column_name = 'atespace'` returns exactly these five today. Derive the list from it, or assert the var equals it so drift fails loudly.

cmd/ateapi/internal/store/atepg/partition_test.go
```diff
@@ -0,0 +1,263 @@
...
+	storecontract.RunContractTests(t, func(t *testing.T) store.Interface {
+		p, err := NewPersistence(t.Context(), traced)
+		if err != nil {
+			t.Fatal(err)
+		}
+		t.Cleanup(p.Close)
+		clearAll(t, p)
+		return p
+	})
+	if len(check.seen) == 0 {
```
BenTheElder: 🤖 **should-fix** 🟡 – This guard can never fire. Every `NewPersistence` runs `rejectUnversionedSubstrateSchema`, whose `information_schema.tables` query has the literal `'actors'` in its `IN (...)` list. It passes the substring prefilter, gets explained harmlessly, and populates `seen`. So the check is satisfied even if the contract suite ran zero statements on atespace tables.

Count only statements whose plan named a partition of a traced table, or skip `information_schema`/`pg_catalog` queries in the tracer.

cmd/ateapi/internal/store/atepg/partition_test.go
```diff
@@ -0,0 +1,263 @@
...
+	verb, _, _ := strings.Cut(strings.ToUpper(sql), " ")
+	if !strings.Contains("SELECT INSERT UPDATE DELETE", verb) || !slices.ContainsFunc(c.tables, func(t string) bool { return strings.Contains(sql, t) }) {
+		return ctx
+	}
+	if c.allowed[sql] {
+		return ctx
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.seen[sql] {
```
BenTheElder: 🤖 **nit** 🟢 – Keyed on SQL text alone, so only the first-seen arguments are ever explained. A predicate like `($1::text IS NULL OR atespace = $1)` plans as one partition with a value and as an `Append` over both with NULL (checked on PG 18). No production query on a partitioned table has that shape today, so this is latent, but call order would decide the verdict when one arrives. Worth a comment here, or keying `seen` on the nullness pattern of the args as well.

cmd/ateapi/internal/store/atepg/partition_test.go
```diff
@@ -0,0 +1,263 @@
...
+		explain:   explain,
+		fail:      fail,
+		seen:      map[string]bool{},
+	}
+}
+
+func (c *fanOutCheck) TraceQueryStart(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryStartData) context.Context {
+	sql := normalizeSQL(data.SQL)
+	verb, _, _ := strings.Cut(strings.ToUpper(sql), " ")
+	if !strings.Contains("SELECT INSERT UPDATE DELETE", verb) || !slices.ContainsFunc(c.tables, func(t string) bool { return strings.Contains(sql, t) }) {
```
BenTheElder: 🤖 **nit** 🟢 – Substring membership silently skips `WITH`, `MERGE`, and `(SELECT ...)`. A future `WITH d AS (DELETE FROM actors WHERE uid = $1 RETURNING ...)` fans out unreported, which is the class of statement this test exists to catch. A `switch` with an explicit `default: return ctx` makes the whitelist visible; inverting it to a small blocklist (`TRUNCATE`, `BEGIN`, `COMMIT`) would explain anything new by default.

Related: the tracer implements only `QueryTracer`, so `SendBatch` bypasses it. `outbox.go` already batches; a batched read on `actors` would be invisible. `TraceBatchQuery` is cheap insurance.

cmd/ateapi/internal/store/atepg/partition_test.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return err
+	}
+	foreignKeys, err := pgx.CollectRows(rows, pgx.RowTo[string])
+	if err != nil {
+		return err
+	}
+	if _, err := pool.Exec(ctx, fmt.Sprintf(`
+		CREATE TABLE %[1]s_partitioned (LIKE %[1]s INCLUDING ALL) PARTITION BY HASH (%[2]s);
+		CREATE TABLE %[1]s_p0 PARTITION OF %[1]s_partitioned FOR VALUES WITH (MODULUS 2, REMAINDER 0);
```
BenTheElder: 🤖 **nit** 🟢 – Two partitions is a weak detector for multi-value predicates. With `MODULUS 2`, the values `a`, `b`, `c`, `d` all hash to `actors_p0`, so a future `WHERE atespace = ANY($1)` plans as one partition whenever the fixture names collide, a coin flip for two names. `MODULUS 8` or `16` costs nothing here.

cmd/ateapi/internal/store/atepg/partition_test.go
```diff
@@ -0,0 +1,263 @@
...
+func (c *fanOutCheck) TraceQueryStart(ctx context.Context, _ *pgx.Conn, data pgx.TraceQueryStartData) context.Context {
+	sql := normalizeSQL(data.SQL)
+	verb, _, _ := strings.Cut(strings.ToUpper(sql), " ")
+	if !strings.Contains("SELECT INSERT UPDATE DELETE", verb) || !slices.ContainsFunc(c.tables, func(t string) bool { return strings.Contains(sql, t) }) {
+		return ctx
+	}
+	if c.allowed[sql] {
+		return ctx
+	}
+	c.mu.Lock()
```
BenTheElder: 🤖 **nit** 🟢 – The lock is held across the EXPLAIN round trip, serializing every first-seen statement across connections. Harmless now because the only concurrent contract test is on `workers`, which is filtered before the lock, but it becomes a choke point the day an actor-side concurrency test lands. Mark `seen` under the lock, then explain outside it.

cmd/ateapi/internal/store/atepg/partition_test.go
```diff
@@ -0,0 +1,263 @@
...
+			ORDER BY atespace, name
+			LIMIT $3`)
+	}
+	scopedActorList := normalizeSQL(`
+		SELECT name, proto FROM actors
+		WHERE atespace = $1 AND ($2::text IS NULL OR name > $2)
+		ORDER BY name
+		LIMIT $3`)
+
+	t.Run("by atespace", func(t *testing.T) {
```
BenTheElder, line 54: 🤖 **nit** 🟢 – Both layouts rerun all 70 contract subtests, and 34 of them touch only worker, atespace, and lease tables that are filtered out before any check. Measured under `-race`: the package goes from ~6.7s to ~12.4s. Partitioning `actors` by the composite `(atespace, name)` enforces the union of both layouts in one run, and an exported actor-scoped subset of the contract suite would halve it again. Optional; the two literal layouts do read more clearly.

docs/dev/postgresql-schema-evolution.md
```diff
@@ -27,6 +27,14 @@ If two schema structures hold the same data, keep them consistent while both bin
 
 Do not run a large data backfill during startup. Propose a separate migration process before you add such a change.
 
+## Keep the actors table partitionable
+
+To preserve the option to partition the `actors` table by `atespace` or by `name`, and the other atespace-scoped tables (`actor_egress_policies`, `actor_templates`, `actor_snapshots`, `actor_snapshot_tags`) by `atespace`, do not add a schema change or a query that introduces:
+
+- A unique index or constraint on one of these tables that omits `atespace`, or on `actors` that omits `name`.
+- A foreign key that references one of these tables by columns that omit `atespace`, or `actors` by anything other than `(atespace, name)`.
+- A query on one of these tables that does not filter on `atespace`. A list that must read every partition needs an explicit entry in `TestActorsTablePartitionable`.
```
BenTheElder: 🤖 **nit** 🟢 – This bullet contradicts the `by name` subtest. Under that layout `actors` is partitioned on `name`, so a query filtering on `atespace` alone fans out, which is why `scopedActorList` had to be allowlisted. A developer who follows this rule and writes `WHERE atespace = $1 AND uid = $2` on `actors` gets a CI failure. Say that queries on `actors` must also filter on `name`.

BenTheElder (dismissed):
bandwidth is low today but I do want to keep rolling on this, I let fable 5.1 rip:

cmd/ateapi/internal/store/atepg/partition_test.go
jpbetz: Fixed. This is derived now.

cmd/ateapi/internal/store/atepg/partition_test.go
jpbetz: Good catch. Replaced this with a per-table check that is stricter.

cmd/ateapi/internal/store/atepg/partition_test.go
jpbetz: Done

cmd/ateapi/internal/store/atepg/partition_test.go
jpbetz: Done

cmd/ateapi/internal/store/atepg/partition_test.go
jpbetz: Done

docs/dev/postgresql-schema-evolution.md
jpbetz: Good catch, fixed.

cmd/ateapi/internal/store/atepg/partition_test.go
jpbetz: I also tried running the subtests in parallel in their own schemas and it flakes. Let's keep this.

cmd/ateapi/internal/store/atepg/partition_test.go
jpbetz: It's a bit slower, but I'll add it.

jpbetz:
Thanks @BenTheElder (and thanks Fable 5.1). Feedback applied

BenTheElder (dismissed):
Thanks Joe!

jpbetz:
Rebased to sync up with table rename

BenTheElder (dismissed):

jpbetz:
Any interest in merging this one?

EItanya:
> Any interest in merging this one?

Seems to be failing CI? Happy to once it's passing

jpbetz:
Rebased

jpbetz:
Thanks @EItanya 

This is green now.

EItanya (approved):

PR #1469, opened 2026-09-04
Followups to 1283: comments and field moves

thockin:
This turned out to be more minor than I imagined for comments and more significant for field move.

I'm not sure the field move is better, but I really don't like the nesting-naming.


BenTheElder (commented):
Added commits LGTM, needs rebase.

juli4n:
It looks like it needs to regenerate the python proto code (./hack/update/python-codegen.sh)

thockin:
Any objections to folding python-codegen into the main codegen script?

BenTheElder:
> Any objections to folding python-codegen into the main codegen script?

no.

BenTheElder (approved):

PR #1479, opened 2026-09-04
atelet: cache pool resolutions so a list flap cannot split the CPU counter

### LINKED ISSUE #1478: [Bug]: one failed pod list re-homes ate.actor.stats.cpu.time increments onto a permanent pool-less series
### What happened?

atelet's stats sweep resolves worker-pool labels with one fresh pod LIST per sweep (`nodeWorkerPools` in `cmd/atelet/statspoller.go`), and any failure — a timed-out or errored list — makes that tick's samples group **without** the `ate.workerpool.*` labels.

For the gauges that is a one-tick blink. For `ate.actor.stats.cpu.time` it is permanent damage: an OTel counter binds every increment to its attribute set, so one flapped sweep re-homes that tick's CPU increase onto a newly-born pool-less series. Concretely, with sweeps Δ1, Δ2 (list fails), Δ3:

- `{...,pool=P}` holds Δ1+Δ3, `{...}` (no pool) holds Δ2 — forever; counter increments cannot be re-attributed.
- `rate(ate_actor_stats_cpu_time{workerpool="P"})` dips during every flap, indistinguishable from the workload going quiet.
- A phantom pool-less series bursts precisely during apiserver trouble — the same incidents the telemetry exists to debug.
- Cardinality doubles for every group that ever flapped, and the stranded series persists for the process lifetime.

A partial miss (one pod absent from the list while its ateom directory still exists) flaps just that actor's contributions — same mechanism, smaller blast radius.

### Expected Behavior

A pod's pool is immutable for the pod's lifetime (the `ate.dev/worker-pool` label is stamped at creation; pods never move between pools), so resolutions can be cached per pod UID with zero correctness risk. A failed list should answer from the cache and keep the increments on the pooled series. The residual unlabeled case shrinks to "pod first seen while the apiserver is unreachable", which heals on the next successful list.

### Steps to Reproduce

1. Run an actor and let the stats poller sweep it with a working pod list (samples group under `ate.workerpool.*`).
2. Make the next sweep's pod list fail (block the apiserver for >10s, or in a unit test return nil from the resolver).
3. Observe that sweep's samples — including the CPU delta added to `ate.actor.stats.cpu.time` — land on a new label set without the pool keys, and remain there after the list recovers.

### Sandbox Runtime

Both / Runtime Agnostic

main

### Confirmation

baizhenyu:
Fixes #1478.

One failed pod list per sweep grouped that tick's samples without `ate.workerpool.*` labels. Gauges blink and heal; the CPU counter does not: an OTel counter binds increments to their attribute set, so a single flap re-homes that sweep's CPU increase onto a newly-born pool-less series that coexists with the pooled one forever — pool-filtered `rate()` dips on every flap, a phantom unlabeled series bursts exactly during apiserver trouble, and cardinality doubles per flapped group for the process lifetime.

## Fix

A pod's pool is **immutable for the pod's lifetime** (the `ate.dev/worker-pool` label is stamped at creation; pods never move between pools), so caching resolutions per pod UID carries zero correctness risk. `resolveWorkerPools`:

* lets fresh resolutions win (harmless either way, given immutability),
* falls back to `cachedPools` when the list fails outright or misses one pod,
* rebuilds the cache restricted to the pods whose ateom directories exist this sweep — pruning departed pods and bounding memory with the same prune-to-the-living pattern `lastCPU` uses.

The residual unlabeled case shrinks to "pod first seen while the apiserver is unreachable", which heals on the next good list — and is pinned by a test as the documented behavior.

## Testing

* List flap: seed the cache, fail the next list, assert the samples stay on the pooled label set with no pool-less split.
* Prune: an entry whose ateom directory is gone leaves the cache; the live pod's entry stays.
* Residual case: a pod first seen during an outage groups without pool labels (one group, not an error).
* `hack/verify-all.sh` green.

Part of #896, toward #550.

## Note: cache lifetime is tied to ateom directories, which nothing cleans up

`cachedPools` evicts an entry on the first sweep after its ateom directory disappears — but today **nothing removes `ateoms/<podUID>/` when a worker pod is deleted** (the only cleanup anywhere is an ateom clearing its own stale socket at boot; directories accumulate until node replacement). Repo archaeology found no decision behind this: the directory set had no reader until the stats sweep (#961) promoted it into a discovery registry, so GC was never needed before.

Consequences for this PR, sized honestly:

* **Correctness is unaffected.** Pod UIDs are never reused, and a retained entry is only consulted when a sample arrives from that UID — a dead socket produces none.
* **Memory is a rounding error on a pre-existing cost.** The cache is bounded by the directory count (~150 bytes per departed pod, ~1.5 MB per ten thousand); each stale directory already costs more than its cache entry does — a dial attempt every sweep plus debug-log noise, on main today, independent of this PR.
* Eviction is deliberately **not** tied to the pod list — "the list stopped returning this pod" is exactly the flap signal this fix exists to distrust. The filesystem stays the retention authority, consistent with being the discovery authority.

The durable fix is directory GC (atelet holds both halves of the evidence each sweep: the dir set and the node's pod list; "absent from the pod list AND socket dead for N consecutive sweeps → remove" is safe against half-born ateoms and transient list failures). That pre-existing issue is now tracked as #1677; when it lands, cache eviction follows it automatically with zero changes here.

### DIFF
--- a/cmd/atelet/statspoller.go
+++ b/cmd/atelet/statspoller.go
@@ -436,7 +436,7 @@ func newWorkerPoolFetcher(client kubernetes.Interface, nodeName string) func(ctx
 			LabelSelector: workerPoolLabel,
 		})
 		if err != nil {
-			slog.DebugContext(ctx, "Actor stats sweep: worker pool resolution failed; samples group without pool labels", slog.Any("err", err))
+			slog.DebugContext(ctx, "Actor stats sweep: worker pool list failed; answering from cached resolutions", slog.Any("err", err))
 			return nil
 		}
 		pools := make(map[string]workerPoolRef, len(pods.Items))
--- a/cmd/atelet/statspoller.go
+++ b/cmd/atelet/statspoller.go
@@ -68,9 +68,9 @@ const statsRPCTimeout = 55 * time.Second
 // ceil(n/statsSweepConcurrency) timeouts instead of n.
 const statsSweepConcurrency = 8
 
-// workerPoolListTimeout bounds the per-sweep pod list that resolves worker
-// pools. Pool labels are enrichment: better one unlabeled tick than a sweep
-// blocked on the apiserver.
+// workerPoolListTimeout bounds the per-sweep pod list that fetches worker
+// pools: pool labels are enrichment, so a slow apiserver must not stall the
+// sweep it is merely enriching.
 const workerPoolListTimeout = 10 * time.Second
 
 // clampActorStatsPollInterval enforces the floor on a nonzero configured
@@ -118,13 +118,13 @@ type statsPoller struct {
 	// connections the lifecycle RPCs are using.
 	dial func(ctx context.Context, podUID string) (activeStatsClient, io.Closer, error)
 
-	// workerPools resolves this node's worker pod UIDs to the pool that owns
-	// them, called once per sweep. Nil (or a nil map, or a missing entry)
-	// degrades to samples grouped without pool labels rather than dropped:
-	// the pool is enrichment, the sample is the point. The real resolver
-	// lists the node's pods by workerPoolLabel; the ateom directory name IS
-	// the worker pod UID, which is the join key.
-	workerPools func(ctx context.Context) map[string]workerPoolRef
+	// fetchWorkerPools is the raw fetch: one apiserver list mapping this
+	// node's worker pod UIDs to the pool that owns them, called once per
+	// sweep by resolveWorkerPools, which layers cachedPools over it. A nil
+	// return means the fetch failed. The real fetcher lists the node's pods
+	// by workerPoolLabel; the ateom directory name IS the worker pod UID,
+	// which is the join key.
+	fetchWorkerPools func(ctx context.Context) map[string]workerPoolRef
 
 	inst *statsInstruments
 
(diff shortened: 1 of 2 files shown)


cmd/atelet/statspoller_test.go
```diff
@@ -490,10 +490,121 @@ func TestNodeWorkerPools(t *testing.T) {
...
 	if diff := cmp.Diff(want, got, cmp.AllowUnexported(workerPoolRef{})); diff != "" {
-		t.Errorf("nodeWorkerPools mismatch (-want +got):\n%s", diff)
+		t.Errorf("newWorkerPoolFetcher mismatch (-want +got):\n%s", diff)
+	}
+}
+
+// TestStatsPollerPoolCacheSurvivesListFlap pins the fix for the label-set
+// split: a failed list must answer from the cache and keep that tick's
+// samples on the pooled label set.
+func TestStatsPollerPoolCacheSurvivesListFlap(t *testing.T) {
```
JeffLuoo, line 506: The tests never exercise the CPU counter — the thing the PR exists to fix.

executingResponse leaves ActorUid/CpuUsageUsec zero, so every delta is 0 and the label-set check is only a proxy. Can you update to test to test it out?

cmd/atelet/statspoller.go
JeffLuoo: "samples group without pool labels" is what you are fixing in this PR. Can you update the error log?

cmd/atelet/statspoller_test.go
baizhenyu: Right — the label-set check was only a transitive guard, and with the counter at zero the tests never showed an increment surviving a flap. Both tests now baseline a real cpu_usage_usec and assert the delta computed *during* the flap (600µs) / partial list (250µs) lands on the pooled key — the value that feeds the monotonic counter and can never be re-attributed (6308fb93).

cmd/atelet/statspoller.go
baizhenyu: Fixed in 6308fb93 — it now says "worker pool list failed; answering from cached resolutions", which is what actually happens.

JeffLuoo (approved):

PR #1480, opened 2026-09-04
Drop k8s third_party deps now that HEAD is updated

thockin:
The changes we need are live on HEAD of the k8s published repos.  It was a bit of a journey, since Go fought me all the way.

---

### Drop our third_party fork of k8s deps in tools
    
    The changes we need are released now (sort of - on HEAD anyway).

---

### Bump k8s.io/streaming to v0.37.0 (not rc) in tools

---

### Bump codegen deps to HEAD in tools
    
    GOPROXY=direct go get \
        k8s.io/apimachinery@master \
        k8s.io/code-generator@master
    
    This pins them to HEAD of master.  Go is terrible here: The HEAD is not
    actually tagged, so Go just uses the next "reachable" tag which is
    v0.36.0-alpha.  The datestamp is correct, though.

---

### Drop our third_party fork of k8s deps in root
    
    The changes we need are released now (sort of - on HEAD anyway).

---

### Bump k8s deps to v0.37.0 (not rc) in root

---

### Bump apimachinery dep in root to HEAD in root
    
    GOPROXY=direct go mod edit -replace k8s.io/apimachinery=k8s.io/apimachinery@master
    GOPROXY=direct go mod tidy
    GOPROXY=direct go mod vendor
    GOPROXY=direct go mod tidy
    
    This approach (-replace) is needed because Go is horrible here.
    
    The master branch of k8s.io/apimachinery is not tagged, per se, but
    there is an OLDER tag which is "reachable" from HEAD.  So go helpfully
    decides to use that (v0.36.0-alpha.2).  If we just `go get ... @master`
    it works for that dep (pinned to the right date) but then it looks at
    transitive deps.  Because the tag seems to be 0.36 (older), it
    recalculates all the OTHER dependencies and downgrades a whole tangle of
    things to versions that match 0.36, but we are ACTUALLY on 0.37+.
    
    This was the only approach that I (and Gemini) could find.  Blech.

---

### Run updated codegens

---

### Use DV's new maxBytes capability for `[]byte`

    Removes 1 custom.


BenTheElder:
I think you need to regenerate the license folder / manually fix it, to drop the dead third_party licenses

hack/tools/code-generator/go.mod
```diff
@@ -35,8 +35,8 @@ require (
 	golang.org/x/tools v0.49.0 // indirect
 	google.golang.org/protobuf v1.36.12 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
-	k8s.io/apimachinery v0.37.0-rc.0 // indirect
-	k8s.io/code-generator v0.37.0-rc.0 // indirect
+	k8s.io/apimachinery v0.36.0-alpha.2.0.20260904054801-5a60cf1c6b81 // indirect
+	k8s.io/code-generator v0.30.0-alpha.3.0.20260904060942-0cddac4383e4 // indirect
```
BenTheElder, line 39: that's weird, we should have at least v0.37.0-alpha ?? I wonder if the generators are failing to sync. not super important at the moment but ...

we can check with @dims  later

thockin: The datestamps are right but we don't tag on main and git's notion of "reachable" is not human-oriented

_LICENSES/github.com/stretchr/testify/internal/spew/LICENSE
```diff
@@ -0,0 +1,15 @@
+ISC License
```
BenTheElder, line 1: we need to start enforcing a license allowlist at some point

I double checked, ISC is on the list (also I assume k8s is already using it anyhow, but k8s also has some exceptions I think)

https://github.com/cncf/foundation/blob/main/policies-guidance/allowed-third-party-license-policy.md

hack/tools/code-generator/go.mod
dims: I see https://github.com/kubernetes/code-generator/tree/v0.37.0 for sure

thockin:
green

BenTheElder (approved):

PR #1482, opened 2026-09-04
atenet: report "unknown" when a resume does not complete

### LINKED ISSUE #1474: [Bug]: atenet.router.route.duration mislabels failed cold resumes as "none" and emits empty template attributes on routing failures
### What happened?

During worker pool saturation, capacity exhaustion (`no_capacity`), or routing errors, `atenet.router.route.duration` mislabels all failed activations as steady-state traffic (`ate.router.resume="none"`) and emits empty strings for template dimensions (`ate.template.atespace=""`, `ate.template.name=""`)

### Expected Behavior

Failed cold activations must not be tagged with ate.router.resume="none" (which represents steady-state routes on running actors). They should preserve the cold activation intent (e.g. triggered or a dedicated failure state).

### Steps to Reproduce

1. Deploy Agent Substrate on Kubernetes with atenet-router.
2. Send an HTTP request targeting an actor that cannot be resolved or whose activation fails (e.g., worker pool scaled to 0 or nonexistent authority):
```
curl -s -H "Host: nonexistent-actor.default.ate.dev" http://<router-ip>/
```
3. Scrape metrics from the router pod on port 9090 (or 9096 via port-forward):
```
curl -s http://localhost:9090/metrics | grep "atenet_router_route_duration_seconds_count"
```
4. Observe emitted metric line:
```
atenet_router_route_duration_seconds_count{ate_router_outcome="not_found",ate_router_resume="none",ate_template_atespace="",ate_template_name="",otel_scope_name="atenet-router"} 1
```

### Sandbox Runtime

Both / Runtime Agnostic

main

GKE

### Host OS & Architecture

_No response_

```shell

```

### Additional Context

_No response_

### Confirmation

JeffLuoo:
Fixes #1474

`atenet.router.route.duration` carries two labels that the router filled in wrongly on a failure path.

## `ate.router.resume` said "none" for a failed resume

`none` means "the resume found the actor already running" — the warm route. The router also gave `none` to every failed resume, every canceled caller, and every caller shed by a full parking lot. Those failures landed in the warm-route series, so they did not show in the cold-start rate and they polluted the warm-route latency.

A gRPC code cannot recover the fact. A canceled leader's flight outlives its request and keeps restoring the actor, and a `DeadlineExceeded` can land in the middle of a restore. So the PR does not classify the error. It reserves `none`, `triggered` and `joined` for a resume that completed, and adds a fourth value, `unknown`, for every resume that did not:

| Case | Before | After |
|---|---|---|
| Actor already running | `none` | `none` |
| Leader completed a cold activation | `triggered` | `triggered` |
| Joiner waited on a completed cold activation | `joined` | `joined` |
| Resume failed (leader and joiners) | `none` | `unknown` |
| Caller canceled before the flight finished | `none` | `unknown` |
| Caller shed by a full parking lot | `none` | `unknown` |
| Egress, which never resumes an actor | `none` | `unknown` |

## `ate.template.*` was the empty string on a failure

The registry marks `ate.template.atespace` and `ate.template.name` required on this metric, but the router sent an empty string when it failed before it resolved a template. `ateattr.NormalizeTemplateDimension` now sends the constant `"unknown"` instead. The constant does not come from the request, thus it adds no caller-controlled label value — the cardinality rule in `docs/metrics/substrate.yaml` still holds, and the PR records the exception there.

## How to read the change on a dashboard

A query that already splits by `ate.router.resume` gains an `unknown` series and loses the failures that used to hide inside `none`. A `none` rate reads lower after the change and a cold-start rate is unaffected. Do not read the time in the `unknown` series as an activation time.

## Documentation

- `docs/metrics/registry/metrics.yaml`: the `unknown` member of `ate.router.resume`, the `"unknown"` fallback on the two template labels, and a note on the metric.
- `docs/metrics/substrate.yaml`: the cardinality-rule exception.
- `docs/observability.md`: the label list for the metric.

## Tests

- `resumer_test.go`: a failed flight gives `unknown` to the leader and to all nine joiners; a canceled caller gives `unknown`; a shed caller and a shed joiner give `unknown`.
- `metrics_test.go`: empty template dimensions record as `"unknown"`; `Result.resume()` defaults to `unknown`.
- `ateattr_test.go`: `NormalizeTemplateDimension` and the four `RouterResume*` values.

- [x] Tests pass


cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -255,25 +255,23 @@ func (r *ActorResumer) ResumeActor(ctx context.Context, actorRef resources.Actor
...
 		// Disambiguate singleflight resume outcome:
-		// - ResumeOutcomeNone ("none"): resumed == false, actor was already active/running.
-		// - ResumeOutcomeTriggered ("triggered"): Cold activation leader (resumed == true, caller's reqID == leaderID).
-		// - ResumeOutcomeJoined ("joined"): Cold activation joiner (resumed == true, caller's reqID != leaderID).
+		// - ResumeOutcomeNone ("none"): resumed == false and err == nil, actor was already active/running.
+		// - ResumeOutcomeTriggered ("triggered"): Cold activation leader (resumed == true or attempt errored, caller's reqID == leaderID).
+		// - ResumeOutcomeJoined ("joined"): Cold activation joiner (resumed == true or attempt errored, caller's reqID != leaderID).
 		outcome := ResumeOutcomeNone
-		if callRes.resumed {
+		if callRes.resumed || callRes.err != nil {
```
krisztianfekete: A `NotFound` never triggered anything, but it now gets `resume="triggered"`. Can we add a new value like unattempted for definitive errors, and keep triggered/joined for the capacity ones where a resume really was in flight?

cmd/atenet/internal/router/ingress/resumer.go
krisztianfekete: The PR says it reserves `none` for warm hits, but here and other places it still returns `none` on a cancel.

internal/ateattr/ateattr.go
```diff
@@ -56,6 +56,12 @@ const (
 	ActorVersionKey       = attribute.Key("ate.actor.version")
 )
 
+// TemplateUnknown is the fallback for unresolvable or missing template dimensions.
+// Note: "unknown" is syntactically a legal atespace/template name; like
+// ate.sandbox.class="unknown", this trades potential collision with a real
+// object named "unknown" for maintaining a bounded, non-empty metric dimension.
+const TemplateUnknown = "unknown"
```
krisztianfekete, line 63: Can we have a Normalize for this, like for the rest of unknowns?

cmd/atenet/internal/router/ingress/resumer.go
krisztianfekete: A `NotFound` never triggered anything, but it now gets `resume="triggered"`. Can we add a new value like unattempted for definitive errors, and keep triggered/joined for the capacity ones where a resume really was in flight?

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -255,25 +255,23 @@ func (r *ActorResumer) ResumeActor(ctx context.Context, actorRef resources.Actor
...
-		// under the 'outcome' label rather than misreported as an activation.
-		if callRes.err != nil {
-			return nil, ResumeOutcomeNone, callRes.err
-		}
-
 		// Disambiguate singleflight resume outcome:
-		// - ResumeOutcomeNone ("none"): resumed == false, actor was already active/running.
-		// - ResumeOutcomeTriggered ("triggered"): Cold activation leader (resumed == true, caller's reqID == leaderID).
-		// - ResumeOutcomeJoined ("joined"): Cold activation joiner (resumed == true, caller's reqID != leaderID).
+		// - ResumeOutcomeNone ("none"): resumed == false and err == nil, actor was already active/running.
```
krisztianfekete: Here, + the PR says it reserves `none` for warm hits, but we have multiple places it still returns `none` on a cancel.

docs/metrics/registry/metrics.yaml
```diff
@@ -75,13 +75,17 @@ groups:
       - id: ate.template.atespace
         stability: development
         type: string
-        brief: The atespace of the ActorTemplate of the actor.
-        examples: [ate-demo-counter]
+        brief: >
+          The atespace of the ActorTemplate of the actor, or "unknown" if the
+          request failed before template resolution.
```
krisztianfekete: Two things:
 - "if the request failed" misses egress. Egress succeeds and still has no template, so it gets unknown too. Result's comment in extproc/handler.go phrases it right, can we mirror that?
- 7 metrics share this brief but only the router sends unknown. And line 72 plus bounded-or-catalog-scoped in docs/metrics/substrate.yaml both say these labels only ever name something an operator made. unknown doesn't, so it needs writing down there.

cmd/atenet/internal/router/ingress/resumer_test.go
```diff
@@ -146,8 +147,67 @@ func TestActorResumer_ResumeActor(t *testing.T) {
 		if got := status.Code(err); got != codes.NotFound {
 			t.Errorf("expected gRPC code NotFound, got %v (err=%v)", got, err)
 		}
-		if outcome != ResumeOutcomeNone {
-			t.Errorf("expected outcome %q on error, got %q", ResumeOutcomeNone, outcome)
+		if outcome != ResumeOutcomeTriggered {
```
krisztianfekete: Does this expect a nonexistent actor trigering a cold activation basically?

internal/ateattr/ateattr.go
JeffLuoo: Added normalization in `ateattr`, following the pattern of `NormalizeSandboxClass` and `NormalizeOperationName`, and switched `recordRouteDuration` to use it. Added unit tests for it in `ateattr_test.go` as well.

cmd/atenet/internal/router/ingress/resumer.go
JeffLuoo: Great catch. Added `RouterResumeUnattempted = "unattempted"` to `ateattr` and registered it in `metrics.yaml`.

cmd/atenet/internal/router/ingress/resumer.go
JeffLuoo: Updated. Set `none` only when actual warm hits where the actor was already running.

docs/metrics/registry/metrics.yaml
JeffLuoo: Updated both.

cmd/atenet/internal/router/ingress/resumer_test.go
JeffLuoo: Fixed. With `unattempted` introduced, `ActorNotFound` now expects `ResumeOutcomeUnattempted` (since no cold activation was ever attempted). Also added a test case for context cancellation verifying it returns `unattempted`.

internal/ateattr/ateattr.go
```diff
@@ -56,6 +56,21 @@ const (
...
+
+// NormalizeTemplateDimension ensures a template dimension (atespace or name) is
+// non-empty, falling back to TemplateUnknown if unset.
+func NormalizeTemplateDimension(dim string) string {
```
krisztianfekete, line 67: Nice, but it's still only called from the router though, and the group brief now promises unknown for all 7 metrics that ref these attrs.

cmd/atenet/internal/router/ingress/resumer.go
```diff
@@ -160,6 +161,18 @@ func (r *ActorResumer) retryable(err error) bool {
...
+// activation could be attempted (e.g. the actor does not exist, bad request, or
+// permission denied), as opposed to in-flight capacity or transient failures.
+func isDefinitiveResumeError(err error) bool {
+	switch status.Code(err) {
+	case codes.NotFound, codes.InvalidArgument, codes.PermissionDenied, codes.Unauthenticated:
+		return true
+	default:
+		return false
+	}
+}
```
krisztianfekete, line 174: Can we flip the default? Right now `Internal`, `Unimplemented`, `Unknown` etc. fall through to triggered, so an unrecognized code is as an activation in the latency series.
`Unknown` means we don't know, so unattempted seems safer.

Also wondering about `FailedPrecondition`, its registry brief is "the state of the actor did not permit a route" and with parking off it fails immediately, which sounds definitive. And retryable's comment just above already groups `DeadlineExceeded` with `NotFound` and `PermissionDenied`, so it's a bit weird to have two classifiers for the same codes I think.

Wdyt?

internal/ateattr/ateattr.go
```diff
@@ -197,6 +212,9 @@ const (
 	RouterResumeTriggered = "triggered"
```
krisztianfekete, line 212: > initiated cold activation

a request that is `ResourceExhausted` gets triggered without activating anything. Same wording in the registry brief.

docs/metrics/registry/metrics.yaml
```diff
@@ -422,6 +427,13 @@ groups:
                 A resume was already in operation. This request waited for it.
                 This value is separate because it is not a correct sample. One
                 cold start with 50 requests is one slow activation and not 51.
+            - id: unattempted
+              stability: development
+              value: unattempted
+              brief: >
+                No cold activation was in flight. The request was invalid, the
+                actor did not exist, the direction does not resume, or the
+                request canceled before activation was attempted.
```
krisztianfekete, line 436: "No cold activation was in flight" and "request canceled" contradict each other, a canceled joiner was waiting on someone else's activation. Maybe "this request neither attempted nor observed an activation" is more accurate?

internal/ateattr/ateattr.go
JeffLuoo: Thanks for catching it. This change would touch 7 metrics. I would prefer to do it in a follow-up PR for a cleaner chagne.

cmd/atenet/internal/router/ingress/resumer.go
JeffLuoo: Agreed on it. But I want to discuss more on the flipping bit.

If I flip it, I still need a list of **codes** that mean "an activation was in flight". This is not quite maintainable by looking at the code now.

My proposed change will be:

1. Delete `isDefinitiveResumeError`. This removes the classifier.
2. Make err == nil a condition for triggered and joined. A resume that fails reports the fourth value (added in this PR). `FailedPrecondition` and `ResourceExhausted` both go there. It also corrects the words "initiated cold activation" that you found.
3. Change the name `unattempted` to `unknown`. The value includes requests where an activation did operate. One example is a **canceled leader**. Its flight is separate from the request context, and it continues to restore the actor. A second example is `DeadlineExceeded` during a restore. Thus `unattempted` is not true for these requests, and new words in the brief cannot correct this.

WDYT?

krisztianfekete: Yes, to all of these,  you can just make the brief say "resume didn't complete, can't say if an activation ran" or similar.

cmd/atenet/internal/router/ingress/resumer_test.go
```diff
@@ -146,8 +147,87 @@ func TestActorResumer_ResumeActor(t *testing.T) {
...
+	t.Run("SingleflightDeduplication_ErrorDisambiguation", func(t *testing.T) {
+		var resumeCalled int
+		var mu sync.Mutex
+		const concurrentRequests = 10
+		var callersStarted atomic.Int32
+
+		mock := &resumerMockClient{
+			resumeFn: func(ctx context.Context, in *ateapipb.ResumeActorRequest, opts ...grpc.CallOption) (*ateapipb.ResumeActorResponse, error) {
+				mu.Lock()
+				resumeCalled++
```
krisztianfekete, line 184: Should we assert this?

cmd/atenet/internal/router/ingress/resumer.go
JeffLuoo: Thanks. Updated all three I mentioned above.

cmd/atenet/internal/router/ingress/resumer_test.go
JeffLuoo: Updated

docs/metrics/registry/metrics.yaml
JeffLuoo: Agreed. The two statements cannot both be true.

The value has the new name `unknown`, and the brief no longer makes a statement about the activation.

internal/ateattr/ateattr.go
JeffLuoo: Updated.

A request with ResourceExhausted now gets unknown, not triggered. Only a resume that completed can get `trigger`.

JeffLuoo:
@krisztianfekete gentle ping

cmd/atenet/internal/router/ingress/resumer_test.go
```diff
@@ -149,8 +150,85 @@ func TestActorResumer_ResumeActor(t *testing.T) {
 		if got := status.Code(err); got != codes.NotFound {
 			t.Errorf("expected gRPC code NotFound, got %v (err=%v)", got, err)
 		}
-		if outcome != ResumeOutcomeNone {
-			t.Errorf("expected outcome %q on error, got %q", ResumeOutcomeNone, outcome)
+		if outcome != ResumeOutcomeUnknown {
+			t.Errorf("expected outcome %q on a failed resume, got %q", ResumeOutcomeUnknown, outcome)
+		}
```
krisztianfekete, line 155: This can flake, can we make it more robust by following the approach of https://github.com/agent-substrate/substrate/pull/1732 that got merged last week?

krisztianfekete (commented):
Added one comment, the rest looks good to me now, thanks!

cmd/atenet/internal/router/ingress/resumer_test.go
JeffLuoo: Good catch, thank you — updated

JeffLuoo: Updated

krisztianfekete: Thanks, that's the right fix for `FailedFlight`, but I think it landed on the wrong test though, I meant `CallerContextCanceled` just above, on L158 now.

JeffLuoo: Done

krisztianfekete (approved):

PR #1484, opened 2026-09-04
metrics: bound ate.sandbox.class with one rule for every emitter

### LINKED ISSUE #1475: ate.workerpool.workers emits a empty string series with ate_sandbox_class=""
### What happened?

The gauge `ate.workerpool.workers` emits a series with `ate_sandbox_class=""`.

The empty value comes from the worker record, not from the WorkerPool CR. `Control.CreateWorker` is a public RPC and it does not validate `sandbox_class`, so a client can register a worker with an empty one. Worker records written before this keep the empty value.

The pool is seeded at 0 under its own class, so the ghost series sits next to a correct series for the same pool. This makes the pool look like it has two populations of workers.

### Expected Behavior

The gauge must never emit `ate_sandbox_class=""`.

A worker with no sandbox class cannot run an actor. The scheduler matches the class exactly, so it places nothing on that worker. Therefore the worker must not be counted as free capacity of its pool. It must report as `unknown`.

A pool of 3 workers with no sandbox class must emit:

```
ate_workerpool_workers{ate_sandbox_class="gvisor",ate_worker_state="idle",...}     0
ate_workerpool_workers{ate_sandbox_class="gvisor",ate_worker_state="assigned",...} 0
ate_workerpool_workers{ate_sandbox_class="unknown",ate_worker_state="idle",...}    3
```

The pool keeps its own series at 0, so an `idle == 0` alert still fires. The broken workers stay visible under `unknown`.

### Steps to Reproduce

1. Register a worker with an empty sandbox class. Call `Control.CreateWorker`
   with `sandbox_class: ""`. The RPC accepts it.
2. Scrape the Prometheus endpoint on ate-api-server:

```bash
curl -s http://localhost:9090/metrics | grep "ate_workerpool_workers"
```

3. The pool reports 0 under its own class, and a second series with
   `ate_sandbox_class=""` reports the worker.

### Sandbox Runtime

Both / Runtime Agnostic

main

_No response_

### Host OS & Architecture

_No response_

```shell

```

### Additional Context

_No response_

### Confirmation

JeffLuoo:
Fixes #1475

`Control.CreateWorker` does not validate `Worker.sandbox_class`, so a client can register a worker with an empty one (and any record written before this keeps it). `RegisterWorkerCount` tallied that raw value, so `ate.workerpool.workers` emitted `ate_sandbox_class=""` — not a member of the `ate.sandbox.class` registry vocabulary — alongside the pool's seeded series at 0.

This change normalizes the worker's class with the existing `ateattr.NormalizeSandboxClass`, so an empty class reports as `unknown`.

New `ateattr.SandboxClassAttribute`, used by every emitter. This closes the same gap on `ate.actor.crashes` and the `ate.actor.lifecycle.*`.

- [x] Tests pass

### DIFF
--- a/cmd/ateapi/internal/controlapi/metrics_test.go
+++ b/cmd/ateapi/internal/controlapi/metrics_test.go
@@ -400,3 +400,102 @@ func TestWorkerCountSeedsZeroForKnownPools(t *testing.T) {
 		}
 	}
 }
+
+// TestWorkerCountReportsEmptyClassWorkerAsUnknown asserts that workers registered
+// through CreateWorker with no sandbox class report as unknown, not as an
+// empty-string class and not as capacity of the pool they name.
+func TestWorkerCountReportsEmptyClassWorkerAsUnknown(t *testing.T) {
+	pools := func(labels.Selector) ([]*atev1alpha1.WorkerPool, error) {
+		return []*atev1alpha1.WorkerPool{
+			workerPool("ns-1", "pool-empty", ""),
+		}, nil
+	}
+	workers := func() ([]*ateapipb.Worker, error) {
+		return []*ateapipb.Worker{
+			worker("ns-1", "pool-empty", "", false),
+			worker("ns-1", "pool-empty", "", false),
+			worker("ns-1", "pool-empty", "", false),
+		}, nil
+	}
+	reader := newWorkerCountReader(t, workers, pools)
+
+	sum := mustMetric(t, reader, workerpoolWorkersMetric).Data.(metricdata.Sum[int64])
+	got := seriesCounts(sum)
+	// The pool's own series stays at 0: none of the three can be scheduled, so
+	// an idle==0 alert on the gvisor capacity still fires, and the three stay
+	// visible under unknown.
+	want := map[series]int64{
+		{"ns-1", "pool-empty", ateattr.WorkerStateIdle, "gvisor"}:                    0,
+		{"ns-1", "pool-empty", ateattr.WorkerStateAssigned, "gvisor"}:                0,
+		{"ns-1", "pool-empty", ateattr.WorkerStateIdle, ateattr.SandboxClassUnknown}: 3,
+	}
+	if len(got) != len(want) {
+		t.Fatalf("got %d series, want %d: %v", len(got), len(want), got)
+	}
+	for k, v := range want {
+		if gv, ok := got[k]; !ok || gv != v {
+			t.Errorf("series %v = %d (present=%v), want %d", k, gv, ok, v)
+		}
+	}
+}
+
+// TestWorkerCountMicroVMPoolWithEmptyClassWorker asserts that the same holds for
+// a microvm pool: the empty-class worker reports as unknown, and no gvisor series
+// appears for a pool that runs no gvisor.
+func TestWorkerCountMicroVMPoolWithEmptyClassWorker(t *testing.T) {
+	pools := func(labels.Selector) ([]*atev1alpha1.WorkerPool, error) {
+		return []*atev1alpha1.WorkerPool{
+			workerPool("ns-1", "pool-micro", atev1alpha1.SandboxClassMicroVM),
+		}, nil
+	}
+	workers := func() ([]*ateapipb.Worker, error) {
+		return []*ateapipb.Worker{
+			worker("ns-1", "pool-micro", "", false),
+		}, nil
+	}
+	reader := newWorkerCountReader(t, workers, pools)
+
+	sum := mustMetric(t, reader, workerpoolWorkersMetric).Data.(metricdata.Sum[int64])
+	got := seriesCounts(sum)
+	want := map[series]int64{
+		{"ns-1", "pool-micro", ateattr.WorkerStateIdle, "microvm"}:                   0,
+		{"ns-1", "pool-micro", ateattr.WorkerStateAssigned, "microvm"}:               0,
+		{"ns-1", "pool-micro", ateattr.WorkerStateIdle, ateattr.SandboxClassUnknown}: 1,
+	}
+	if len(got) != len(want) {
+		t.Fatalf("got %d series, want %d: %v", len(got), len(want), got)
+	}
+	for k, v := range want {
+		if gv, ok := got[k]; !ok || gv != v {
+			t.Errorf("series %v = %d (present=%v), want %d", k, gv, ok, v)
+		}
+	}
+}
+
+// TestWorkerCountOrphanWorkerWithEmptyClass asserts that an empty-class worker belonging
+// to no known pool falls back to unknown rather than emitting an empty-string class.
+func TestWorkerCountOrphanWorkerWithEmptyClass(t *testing.T) {
+	pools := func(labels.Selector) ([]*atev1alpha1.WorkerPool, error) {
+		return nil, nil
+	}
+	workers := func() ([]*ateapipb.Worker, error) {
+		return []*ateapipb.Worker{
+			worker("ns-1", "pool-orphan", "", false),
+		}, nil
+	}
+	reader := newWorkerCountReader(t, workers, pools)
+
+	sum := mustMetric(t, reader, workerpoolWorkersMetric).Data.(metricdata.Sum[int64])
+	got := seriesCounts(sum)
+	want := map[series]int64{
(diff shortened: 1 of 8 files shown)


JeffLuoo:
Actually, before making this change I have a question about the scheduling - @juli4n 

Worker.sandbox_class is optional: https://github.com/agent-substrate/substrate/blob/a2233676bcb22f33c64a01c79ec04159b111613e/pkg/proto/ateapipb/ateapi.proto#L1688-L1693

and it doesn't have a default value. But this field will be used by scheduler to schedule the worker. CreateWorker accepts "" and nothing fills it in. Should we validate it from the API level?

BenTheElder:
So, I also want to mention that I think workers will have multiple sandbox classes in the future. Actors will be on a specific sandbox class but the idea that we reserve compute only for gvisor or microVM is an artificial distinction that probably won't hold up as we incrementally improve the dataplane.

would that suggest a different approach here?

JeffLuoo:
> So, I also want to mention that I think workers will have multiple sandbox classes in the future. Actors will be on a specific sandbox class but the idea that we reserve compute only for gvisor or microVM is an artificial distinction that probably won't hold up as we incrementally improve the dataplane.
> 
> would that suggest a different approach here?

Thank you for the context. I patch the PR to drop the change in earlier revision had an empty-class worker inherit its pool's class. This PR only normalizes empty string sandbox class to unknown like other metrics. 

For the impact of "workers will have multiple sandbox classes in the future" to the metric `ate.workerpool.workers`, the sum across sandbox classes stops equaling the fleet size. My guess is that worker counts lose the class label?

cc @krisztianfekete we may want to revisit some metrics when this happens.

krisztianfekete (commented):
Approach looks right to me, two things:

1. The same empty class is still on `ate.actor.crashes`.
Same bug as #1475, on an attribute the registry marks required. Maybe we should have a tiny `SandboxClassAttribute` helper in
ateattr so every emitter shares one rule. Fine as a follow up, but please file it
from this PR so it doesn't close along with #1475.

2. `sum by (pool) (ate_workerpool_workers{state="idle"})` still returns 3 for that
pool, because the unknown series keeps the pool labels. Not a regression, as `""`
counted the same way, but as per #1475 these should not not count as pool capacity, and that only happoens if the query includes `class`.

krisztianfekete:
> So, I also want to mention that I think workers will have multiple sandbox classes in the future. Actors will be on a specific sandbox class but the idea that we reserve compute only for gvisor or microVM is an artificial distinction that probably won't hold up as we incrementally improve the dataplane.
> 
> would that suggest a different approach here?

I don't think it changes this PR. We should still do this regardless of how many classes a worker ends up having.


> For the impact of "workers will have multiple sandbox classes in the future" to the metric ate.workerpool.workers, the sum across sandbox classes stops equaling the fleet size. My guess is that worker counts lose the class label?

Today, the instrument's whole contract is that the counts are summable, so `sum by (pool)` = pool size and sum over pools = fleet size, which is why it's an UpDownCounter and not a gauge. Once a worker has a classes, a per-class dimension will double count it. If we drop the class label all is correct, but we cannot answer the "capacity per kind" question, which is something we'd like to answer for sure.

 
So I'd expect it to split, so worker counts stay class-less and summable, and maybe eligibility-by-class becomes its own instrument, where overlaps are fine, and no one wants to sum them. That's ~what `ate.scheduler.eligible_workers` already is.

Not worth designing now while the dataplane is still evolving, and I'd land the label fix  first and revisit when the multi-class model becomes stable. Do you mind opening on issue for this?

JeffLuoo:
> Approach looks right to me, two things:
> 
> 1. The same empty class is still on `ate.actor.crashes`.
>    Same bug as #1475, on an attribute the registry marks required. Maybe we should have a tiny `SandboxClassAttribute` helper in
>    ateattr so every emitter shares one rule. Fine as a follow up, but please file it
>    from this PR so it doesn't close along with #1475.
> 2. `sum by (pool) (ate_workerpool_workers{state="idle"})` still returns 3 for that
>    pool, because the unknown series keeps the pool labels. Not a regression, as `""`
>    counted the same way, but as per #1475 these should not not count as pool capacity, and that only happoens if the query includes `class`.

For point 1, the change is not too long so I just do it in this PR.

internal/ateattr/ateattr.go
```diff
@@ -343,6 +343,14 @@ func NormalizeSandboxClass(class string) string {
 	}
 }
 
+// SandboxClassAttribute sets ate.sandbox.class. Use it in place of
+// SandboxClassKey: no source of the class is validated, thus all emitters must
+// apply the same limit. recordSchedulerAssignment is the one exception. It
```
krisztianfekete: Not the only exception as `cmd/atelet/metrics.go:100` omits when empty as well.

cmd/ateapi/internal/controlapi/metrics_test.go
```diff
@@ -400,3 +400,102 @@ func TestWorkerCountSeedsZeroForKnownPools(t *testing.T) {
 		}
 	}
 }
+
+// TestWorkerCountReportsEmptyClassWorkerAsUnknown asserts that workers registered
+// through CreateWorker with no sandbox class report as unknown, not as an
+// empty-string class and not as capacity of the pool they name.
+func TestWorkerCountReportsEmptyClassWorkerAsUnknown(t *testing.T) {
```
krisztianfekete: Can we turn this, `TestWorkerCountMicroVMPoolWithEmptyClassWorker`, and `TestWorkerCountOrphanWorkerWithEmptyClass` a table?

cmd/ateapi/internal/controlapi/metrics_test.go
```diff
@@ -400,3 +400,102 @@ func TestWorkerCountSeedsZeroForKnownPools(t *testing.T) {
...
+		return []*ateapipb.Worker{
+			worker("ns-1", "pool-micro", "", false),
+		}, nil
+	}
+	reader := newWorkerCountReader(t, workers, pools)
+
+	sum := mustMetric(t, reader, workerpoolWorkersMetric).Data.(metricdata.Sum[int64])
+	got := seriesCounts(sum)
+	want := map[series]int64{
+		{"ns-1", "pool-micro", ateattr.WorkerStateIdle, "microvm"}:                   0,
```
krisztianfekete: Nit: use `atev1alpha1.SandboxClassMicroVM` instead of hardcoding `microvm`?

internal/ateattr/ateattr.go
```diff
@@ -343,6 +343,14 @@ func NormalizeSandboxClass(class string) string {
...
+// omits the attribute when the class is unknown.
+func SandboxClassAttribute(class string) attribute.KeyValue {
```
krisztianfekete, line 352: One more caller for this is `cmd/ateapi/internal/scheduling/metrics.go:95` that still uses the raw key, and the class is required on `ate.scheduler.eligible_workers`.

cmd/ateapi/internal/controlapi/metrics_test.go
JeffLuoo: updated

internal/ateattr/ateattr.go
JeffLuoo: updated

cmd/ateapi/internal/controlapi/metrics_test.go
JeffLuoo: updated

internal/ateattr/ateattr.go
JeffLuoo: Updated that file as well.

krisztianfekete (commented):
Latest looks good!

Only thing left is docs/, still untouched in the PR. Three instruments produce
unknown now instead of two, so metrics.yaml also need to be updated, + the two lines on the workers note about unknown not being capacity.

Could you also open the multi-class followup so for Ben's thread?

JeffLuoo:
Created https://github.com/agent-substrate/substrate/issues/1557

> Only thing left is docs/, still untouched in the PR. Three instruments produce
unknown now instead of two, so metrics.yaml also need to be updated, + the two lines on the workers note about unknown not being capacity.

Updated

krisztianfekete (approved):

JeffLuoo:
@krisztianfekete can you help merge? Not suppose to be merged by author : )

PR #1492, opened 2026-09-04
ateapi: resolve the resume boot source and wire scope once in loadActorForResume

zoez7:
Refactored `loadActorForResume` to resolve the boot source and wire scope once, in 4 rules:

1. If the ActorTemplate was updated since the capture, restore data only from the existing snapshot or checkpoint. #1461 PR will only change this code path to use golden+data.
2. If the actor was resuming from Paused state, use the Local checkpoint at the scope recorded on the checkpoint, falling back to the template's onPause scope for legacy checkpoints that predate the recorded scope. If that scope is Data and onResume.FromData is golden, use golden+data.
3. If the actor was resuming from Suspended state, use the External snapshot at its captured scope. If that scope is Data and onResume.FromData is golden, use golden+data.
4. An actor with no snapshot restores the template's golden when one exists, and cold boots from the spec otherwise. An explicit boot request skips only the golden fallback.

The restore step now dispatches on the resolved kind and fails with Internal on an unhandled kind instead of silently cold booting.

**Behavior changes**
- A resume after an ActorTemplate update now restores data-only from the existing snapshot or checkpoint — including a Paused actor holding only a local checkpoint — and no longer consults the Golden policy, so an absent or invalid golden no longer fails it with FailedPrecondition. This is a temporary placeholder until #1461 switches this path to golden+data.
- A Paused resume no longer validates inputs the local restore never consumes: an unusable golden no longer fails it (intentional bugfix), and the shadowed durable `ExternalSnapshot.SnapshotUri` is no longer parsed, so an unparseable record no longer fails it with DataLoss.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR - Not required

### DIFF
--- a/cmd/ateapi/internal/controlapi/workflow_resume.go
+++ b/cmd/ateapi/internal/controlapi/workflow_resume.go
@@ -34,23 +34,71 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+// restoreKind identifies which boot source the restore step drives. Resolved
+// once by loadActorForResume so the restore step and its telemetry dispatch
+// on it instead of re-deriving the precedence from raw actor fields.
+type restoreKind int
+
+const (
+	// restoreColdBoot boots from the template spec: no snapshot anywhere.
+	// The zero value, so a zero resumeSnapshotSource means cold boot.
+	restoreColdBoot restoreKind = iota
+	// restoreLocalSnapshot restores the actor's local pause checkpoint on
+	// the worker node. Outranks any durable snapshot the actor also holds.
+	restoreLocalSnapshot
+	// restoreDurableSnapshot restores the actor's own durable snapshot at
+	// SnapshotURI.
+	restoreDurableSnapshot
+	// restoreGoldenSnapshot restores the template's golden snapshot at
+	// SnapshotURI as the actor's own content.
+	restoreGoldenSnapshot
+)
+
+// telemetrySnapshotKind maps the restore kind to the snapshot-kind label of
+// the resume lifecycle metric.
+func (k restoreKind) telemetrySnapshotKind() string {
+	switch k {
+	case restoreLocalSnapshot:
+		return ateattr.SnapshotKindLocal
+	case restoreDurableSnapshot:
+		return ateattr.SnapshotKindLatest
+	case restoreGoldenSnapshot:
+		return ateattr.SnapshotKindGolden
+	case restoreColdBoot:
+		return ateattr.SnapshotKindBoot
+	default:
+		// An unhandled kind fails ensureAteletRestored; stay within the
+		// registered label values here.
+		return ateattr.SnapshotKindBoot
+	}
+}
+
 // resumeSnapshotSource is the boot source resolved once by loadActorForResume
 // and passed by value to the restore step — never mutated after resolution.
 type resumeSnapshotSource struct {
-	// SnapshotURI is the storage location of the durable snapshot to restore
-	// from: the actor's own latest snapshot when one exists, the template's
-	// golden snapshot otherwise. Zero means cold boot from the spec (unless
-	// the actor holds a local snapshot, which takes precedence at restore).
+	// Kind selects the restore path; the remaining fields only carry what
+	// that path consumes.
+	Kind restoreKind
+	// SnapshotURI is the storage location of the external snapshot a
+	// restoreDurableSnapshot or restoreGoldenSnapshot restore boots from;
+	// zero for every other kind. A paused actor's shadowed durable snapshot
+	// is never parsed or carried: the local checkpoint outranks it.
 	SnapshotURI resources.SnapshotURI
-	Scope       ateapipb.SnapshotContentScope
-	// GoldenSnapshotURI is the storage location of the ActorTemplate's golden
-	// snapshot. Populated only when the template's onResume configuration
-	// selects the golden snapshot as the boot source for the pending restore:
-	// restore then combines the golden snapshot with the actor's data.
-	GoldenSnapshotURI resources.SnapshotURI
-	// TemplateReplaced is true when the snapshot's recorded template UID
-	// differs from the actor's current template.
-	TemplateReplaced bool
+	// GoldenForDataSnapshotURI is the storage location of the ActorTemplate's
+	// golden snapshot. Populated only when the template's onResume
+	// configuration selects the golden snapshot for a pending data-only
+	// restore: the restore then combines the golden snapshot with the
+	// actor's data.
+	GoldenForDataSnapshotURI resources.SnapshotURI
+	// WireScope is the scope of the restore operation, not of any stored
+	// snapshot:
+	//   - DATA_ON_GOLDEN: the restore rides the golden snapshot.
+	//   - DATA: the restore discards the snapshot's guest state and carries
+	//     the actor's durable data alone.
+	//   - FULL or unspecified: the restore uses a FULL snapshot.
+	//
+	// Meaningless for a cold boot: RunRequest carries no scope.
+	WireScope ateletpb.SnapshotScope
 }
 
 // restoreTelemetry labels the restore operation for the resume lifecycle
@@ -154,7 +202,22 @@ func validateGoldenSnapshotScope(snapshot *ateapipb.ExternalSnapshot) error {
 }
 
 // loadActorForResume fetches the current actor record and its template, and
-// resolves the boot source for the pending restore.
+// resolves the boot source for the pending restore. The source is decided by
+// the first matching rule:
+//
+//  1. The snapshot of an actor whose ActorTemplate was updated holds guest
+//     state from the previous template, so the restore carries the actor's
+//     durable data alone.
+//  2. A paused actor restores its local pause checkpoint at the scope the
+//     checkpoint was captured with; a Data-scoped pause rides the golden
+//     when the onResume policy selects it.
+//  3. An actor on its own template restores its snapshot at the captured
+//     scope; a Data-scoped capture rides the golden when the onResume
+//     policy selects it.
+//  4. An actor with no snapshot restores the template's golden when one
+//     exists, and cold boots from the spec otherwise. An explicit boot
+//     request skips the golden fallback and always cold boots; it has no
+//     effect on an actor with a snapshot of its own.
 func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resources.ActorRef, boot bool) (_ *ateapipb.Actor, _ *ateapipb.ActorTemplate, _ resumeSnapshotSource, err error) {
 	ctx, done := stepSpan(ctx, "LoadActorForResume")
 	defer func() { err = done(err) }()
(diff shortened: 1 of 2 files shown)


cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +199,117 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
-			if goldenURI == "" {
-				return nil, nil, src, status.Error(codes.FailedPrecondition, "a Golden data resume requires the ActorTemplate golden snapshot, which is not available")
-			}
-			if err := validateGoldenSnapshotScope(goldenSnapshotStatus.GetGoldenSnapshot()); err != nil {
-				return nil, nil, src, err
+	}
+
+	// Rule 1: an explicit boot request discards any captured guest state and
+	// carries the actor's durable data alone (none: cold boot from the spec).
+	if boot {
```
dberkov: Rule 1 changes the `boot=true` semantics from "skip the golden fallback" to "force a data-only restore": a suspended actor with a FULL durable snapshot (or a paused actor with an onPause=FULL checkpoint) previously restored at FULL scope with guest memory intact, and now silently drops all guest state. The tests assert the new behavior deliberately, but the public contract still describes the old one — `pkg/proto/ateapipb/ateapi.proto` ("If true, skip golden snapshot and boot the workload from scratch.") and the `kubectl-ate resume-actor --boot` help text are unchanged. If the new semantics are intended, those docs need updating and the change deserves a call-out beyond "refactor"; external callers get state-destroying behavior under an unchanged wire contract otherwise.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +199,117 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+		src.WireScope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
+		return actor, actorTemplate, src, nil
+	}
+
+	onResumeGolden := actorTemplate.GetSnapshotsConfig().GetOnResume().GetFromData() == ateapipb.ResumeSource_RESUME_SOURCE_GOLDEN
+	// Rule 2: An actor whose template was updated will cold boot.
+	builtOnTemplateUID := actor.GetStatus().GetCurrentActorTemplateUid()
+	if external.GetSnapshotUri() != "" && builtOnTemplateUID != "" && builtOnTemplateUID != actorTemplate.GetMetadata().GetUid() {
```
dberkov: This guard keys on `external.GetSnapshotUri() != ""`, so a repointed PAUSED actor with only a local checkpoint skips Rule 2 entirely: run (stamps `CurrentActorTemplateUid`) → pause (writes only `LocalSnapshotInfo`, no external snapshot) → template deleted and recreated under the same name with a new UID → resume falls through to Rule 3 and restores the replaced template's guest state at FULL scope. That's exactly what the new doc says must not happen ("a mismatch ... means the snapshot's guest state must not be restored"), and it's asymmetric — the same actor *with* a leftover durable snapshot correctly gets DATA. The gap predates this PR, but since this rewrites the exact condition and codifies the invariant, the guard should key on any snapshot (local or external), not the external URI.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +199,117 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+	goldenURI, err := resolveGoldenSnapshot(actorTemplate)
+	if errors.Is(err, errNoGoldenSnapshot) {
+		return actor, actorTemplate, src, nil
+	}
+	if err != nil {
+		return nil, nil, src, err
+	}
+	src.SnapshotURI = goldenURI
+	// A golden snapshot is always FULL scope.
+	src.WireScope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_FULL
```
dberkov, line 305: Two new decision branches have no test coverage: this Rule 5 success path (golden becomes `src.SnapshotURI` with WireScope hardcoded FULL — `TestLoadActorForResume_GoldenFallbackRejectsNonFullGolden` only covers the rejection path), and Rule 3's paused + onPause=DATA + non-Golden-policy DATA branch (line 235 — every paused test case uses the GOLDEN policy, seeds a durable snapshot, or passes boot=true). A regression in either (e.g. the golden fallback losing the FULL wire scope) would pass the suite unnoticed.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +199,117 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+	// Rule 2: An actor whose template was updated will cold boot.
+	builtOnTemplateUID := actor.GetStatus().GetCurrentActorTemplateUid()
+	if external.GetSnapshotUri() != "" && builtOnTemplateUID != "" && builtOnTemplateUID != actorTemplate.GetMetadata().GetUid() {
+		src.WireScope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
+		return actor, actorTemplate, src, nil
+	}
+
+	// Rule 3: a paused actor restores its local pause checkpoint, which the
+	// template's onPause scope describes.
+	if actor.GetStatus().GetLocalSnapshotInfo() != nil {
```
dberkov: Rules 3 and 4 are ~15-line copy-paste twins differing only in the scope source (template onPause vs `external.GetContentScope()`), and both hand-write the DATA→SNAPSHOT_SCOPE_DATA-else-FULL mapping that `actorSnapshotContentScopeToAtelet` (converter.go) already provides. Rule 3's raw `GetOnPause() == DATA` also diverges from the `effectiveContentScope(...)` convention used in workflow_pause.go and workflow_suspend.go (equivalent today, drifts if normalization grows). A shared helper like `func(scope, onResumeGolden, tmpl) (goldenURI, wireScope, error)` collapses both blocks to ~3 lines and keeps the next golden-ride policy change from having to land twice.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +199,117 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+
+	// Rule 1: an explicit boot request discards any captured guest state and
+	// carries the actor's durable data alone (none: cold boot from the spec).
+	if boot {
+		src.WireScope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
+		return actor, actorTemplate, src, nil
+	}
+
+	onResumeGolden := actorTemplate.GetSnapshotsConfig().GetOnResume().GetFromData() == ateapipb.ResumeSource_RESUME_SOURCE_GOLDEN
+	// Rule 2: An actor whose template was updated will cold boot.
```
dberkov: "will cold boot" misdescribes this branch: it keeps `src.SnapshotURI` and sets WireScope=DATA — a data-only restore of the durable snapshot, not a cold boot (which in this file means a RunRequest with no snapshot). It also contradicts the function doc three lines up ("the restore carries the actor's durable data alone").

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -690,19 +768,11 @@ func (w *ActorWorkflow) ensureAteletRestored(ctx context.Context, actorRef resou
...
-		// configuration, else what the pause captured.
-		switch {
-		case src.TemplateReplaced:
-			req.Scope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
-		case !src.GoldenSnapshotURI.IsZero():
-			req.Scope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA_ON_GOLDEN
-			req.GoldenSnapshotUri = src.GoldenSnapshotURI.String()
-		default:
-			req.Scope = actorSnapshotContentScopeToAtelet(actorTemplate.GetSnapshotsConfig().GetOnPause())
+		req.Scope = src.WireScope
```
dberkov, line 826: The PR centralizes WireScope/golden resolution in `resumeSnapshotSource`, but the restore *kind* is still re-derived three separate ways in `ensureAteletRestored`: local-vs-external from `actor.GetStatus().GetLocalSnapshotInfo()` here, cold-boot from `!src.SnapshotURI.IsZero()` below, and a third derivation for the telemetry SnapshotKind — coherent with the rule list only by hand-mirrored ordering (Rules 1–2 return with `src.SnapshotURI` populated for paused actors even though this branch ignores it). Carrying a resolved Kind (local/external/coldBoot) in `resumeSnapshotSource` would collapse all three and would have made the Rule 2/Rule 3 guard gap structurally impossible.

cmd/ateapi/internal/controlapi/workflow_resume_test.go
```diff
@@ -1142,8 +1152,121 @@ func TestLoadActorForResume_TemplateReplaced(t *testing.T) {
...
+			goldenURI:  goldenSnapshotURI,
+		},
+		{
+			name:   "paused actor restores its checkpoint data-only",
+			paused: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
dberkov, line 1467: This reproduces ~40 lines of the `TestLoadActorForResume_OnGoldenDataResume` seeding harness nearly verbatim (seedOpts branches, actorState toggle, seedWorkflowActor, MustCreateAtespace, the template literal with OnResumeConfig and conditional GoldenSnapshotStatus), with only the OnPause field and golden scope differing — and the template is hand-built a third time in `TestLoadActorForResume_TemplateReplaced`. A shared seeding helper, or a `boot` column in the existing table, would absorb these cases and keep the copies from drifting when the seeding shape changes.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Added test cases in fb186be75118fea9576bde69978a072907eaf47b.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Done in f98e91e44646f8cecedd1f2e795907cc463c699e. Added a src.kind.

cmd/ateapi/internal/controlapi/workflow_resume_test.go
zoez7: Done in fb186be75118fea9576bde69978a072907eaf47b. Added shared helper.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Done in f98e91e44646f8cecedd1f2e795907cc463c699e. Extracted shared method.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Done in c593c18f8d2b79039dce2d67ab2dfefdd2ace91d. Reserved the current semantic.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Done, updated the comment.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +237,121 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
 		}
+		return actor, actorTemplate, src, nil
 	}
 
+	// Rule 4: an actor with no snapshot of its own restores the template's
+	// golden snapshot as its own content when one exists, and cold boots
+	// from the spec otherwise (the wire scope is inert then: RunRequest
+	// carries no scope). An explicit boot request skips the golden fallback
+	// without resolving it, so an unusable golden cannot block the boot.
+	if boot {
```
dberkov, line 292: The PR description's step 1 ("If boot is true, boot from data-only from whatever snapshot exists, never use golden") is not what this implements: boot only skips the Rule-4 golden *fallback* and has no effect on an actor with its own snapshot. boot=true on an actor with a Data durable snapshot under onResume.fromData=GOLDEN still resolves and rides the golden (pinned by the new test "Data durable snapshot still rides the golden under the Golden policy"), and still fails FailedPrecondition when the golden is missing — the opposite of "never use golden". The code matches the proto contract ("skip golden snapshot and boot the workload from scratch") and its own doc comment, so either the description is stale and will mislead the follow-up, or the intended semantics weren't implemented; worth resolving before merge.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +237,121 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+	case actor.GetStatus().GetLocalSnapshotInfo() != nil:
+		src.Kind = restoreLocalSnapshot
+	case !src.SnapshotURI.IsZero():
+		src.Kind = restoreDurableSnapshot
+	}
+
+	// Rule 1: an actor's template has been updated, so the restore carries
+	// the actor's durable data alone.
+	builtOnTemplateUID := actor.GetStatus().GetCurrentActorTemplateUid()
+	if src.Kind != restoreColdBoot && builtOnTemplateUID != "" && builtOnTemplateUID != actorTemplate.GetMetadata().GetUid() {
```
dberkov, line 262: Behavior change under a "refactor only" checklist: a repointed PAUSED actor holding only a local checkpoint (no durable snapshot) now restores at wire scope DATA. The old code never computed TemplateReplaced without an external snapshot URI, so it restored the checkpoint at the onPause scope (FULL, or DATA_ON_GOLDEN under the Golden policy) — i.e. it restored guest state from the replaced template. Forcing DATA is arguably the correct fix, and the new tests pin it, but it's an undisclosed semantic change for in-flight paused actors and deserves a call-out in the description/commit message.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +237,121 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+		src.WireScope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
```
dberkov, line 263: Error-path removal: a repointed actor with a Data snapshot under onResume.fromData=GOLDEN and a missing or non-Full golden no longer fails FailedPrecondition — the resume now silently proceeds data-only. The old loadActorForResume resolved and validated the golden even when TemplateReplaced made the restore ignore it, so a broken/absent golden surfaced loudly at resume time; Rule 1 now returns before consulting the policy. The new tests pin this deliberately and the restore request is unchanged, but anyone relying on that failure to surface broken goldens loses the signal — another behavior change to disclose.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -749,9 +851,8 @@ func (w *ActorWorkflow) ensureAteletRestored(ctx context.Context, actorRef resou
 		}
 		_, err = client.Restore(ctx, req)
 		return tele, maybeCrashActor(ctx, w.store, actorRef, err, "while restoring durable snapshot", ateattr.OperationResume)
-	} else {
+	default: // restoreColdBoot
 		slog.InfoContext(ctx, "Actor has no snapshot; ActorTemplate has no golden snapshot; Booting from ActorTemplate spec")
```
dberkov: This branch unconditionally logs "ActorTemplate has no golden snapshot", which is false when boot=true deliberately skipped an existing golden: Rule 4 skips it by design, then this tells the operator debugging why the golden wasn't used that it doesn't exist. Pre-existing message, but the rewritten dispatch now has the information (boot vs. genuinely-absent golden) to log the true reason.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +237,121 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
...
+	builtOnTemplateUID := actor.GetStatus().GetCurrentActorTemplateUid()
+	if src.Kind != restoreColdBoot && builtOnTemplateUID != "" && builtOnTemplateUID != actorTemplate.GetMetadata().GetUid() {
+		src.WireScope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
+		return actor, actorTemplate, src, nil
+	}
+
+	// Rule 2: a paused actor restores its local pause checkpoint, which the
+	// template's onPause scope describes.
+	if src.Kind == restoreLocalSnapshot {
+		if src.GoldenForDataSnapshotURI, src.WireScope, err = resolveRestoreScope(actorTemplate.GetSnapshotsConfig().GetOnPause(), actorTemplate); err != nil {
```
dberkov: Rule 2 derives the local-restore scope from the template's current onPause instead of calling the existing `pausedContentScope` helper (workflow_suspend.go), which prefers the scope recorded on `LocalSnapshotInfo.ContentScope` and whose doc claims it matches resume's derivation. Equivalent today only because ActorTemplates are immutable in place (spec changes force a new UID, caught by Rule 1) — but suspend and resume now disagree on which source is authoritative, and if template specs ever become mutable, resume would restore a checkpoint at a scope it wasn't captured with. Passing `pausedContentScope(actor.GetStatus().GetLocalSnapshotInfo(), actorTemplate)` into `resolveRestoreScope` removes the divergence at no cost.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -688,44 +811,23 @@ func (w *ActorWorkflow) ensureAteletRestored(ctx context.Context, actorRef resou
...
-			req.Scope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA
-		case !src.GoldenSnapshotURI.IsZero():
-			req.Scope = ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA_ON_GOLDEN
-			req.GoldenSnapshotUri = src.GoldenSnapshotURI.String()
-		default:
-			req.Scope = actorSnapshotContentScopeToAtelet(actorTemplate.GetSnapshotsConfig().GetOnPause())
+			LocalConfig: &ateletpb.LocalCheckpointConfiguration{SnapshotName: actor.GetStatus().GetLocalSnapshotInfo().GetSnapshotName()},
+		}
+		req.Scope = src.WireScope
+		if src.WireScope == ateletpb.SnapshotScope_SNAPSHOT_SCOPE_DATA_ON_GOLDEN {
```
dberkov, line 828: The local and durable/golden restore branches duplicate both the DATA_ON_GOLDEN→GoldenSnapshotUri conditional and the `tele.WireSnapshotScope` assignment, so two copies must stay in sync on every scope change. loadActorForResume guarantees `GoldenForDataSnapshotURI` is non-zero only for DATA_ON_GOLDEN (and a zero URI's `String()` is ""), so both branches can set `GoldenSnapshotUri: src.GoldenForDataSnapshotURI.String()` unconditionally, and `tele.WireSnapshotScope` can be hoisted above the switch next to `tele.SnapshotKind`.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -179,59 +237,121 @@ func (w *ActorWorkflow) loadActorForResume(ctx context.Context, actorRef resourc
 	if err != nil {
 		return nil, nil, src, err
 	}
-	goldenSnapshotStatus := actorTemplate.GetStatus().GetGoldenSnapshotStatus()
-	if uri := actor.GetStatus().GetExternalSnapshot().GetSnapshotUri(); uri != "" {
+	external := actor.GetStatus().GetExternalSnapshot()
+	if uri := external.GetSnapshotUri(); uri != "" {
```
dberkov: `src.SnapshotURI` is parsed and populated even when the resolved Kind is restoreLocalSnapshot, leaving the documented populated-but-unused field ("May be populated but unused") — and a paused actor whose shadowed durable snapshot URI is unparseable gets DataLoss even though the local restore never consumes it. Pre-existing behavior, but the refactor documents the trap instead of removing it: parsing the external URI inside Rule 3 would make SnapshotURI non-zero exactly when it is the boot source, matching the struct's own field contract.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -639,9 +759,10 @@ func (w *ActorWorkflow) ensureVolumesAttached(ctx context.Context, actor *ateapi
 	return nil
 }
 
-// ensureAteletRestored brings the workload up on the assigned worker:
-// restoring the actor's local snapshot when one exists, else its durable (or
-// golden) snapshot, else cold-booting from the template spec. This is the
+// ensureAteletRestored brings the workload up on the assigned worker,
+// dispatching on the resolved src.Kind: the actor's local pause checkpoint,
+// an external (durable or golden) snapshot, or a cold boot from the template
+// spec. This is the
```
dberkov: Edit residue: the rewritten first sentence wasn't reflowed with the rest of the paragraph, leaving an orphaned mid-sentence fragment ("...or a cold boot from the template spec. This is the") on its own line before "atelet reentrancy seam".

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Done in 1d76e7c2b8e930be9bf86c6c18e96318f6ac5fc4

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Done. Now we only populates this when a non-local snapshot is used.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Updating the PR description, this is a bug fix.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Updating the PR description, this is a bug fix. The behaviour will change after #1461 - when actortemplate will start using Data+Golden to restore.,

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Updating the PR description.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Fixed the comment.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Done in bce36dbf4460acecd526bac67a059a07b4245efa, using the `pausedContentScope`.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Updated the log message.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: I disagree with this Claude finding. Guarding the `GoldenForDataSnapshotURI` by checking the wireScope first is more defensive in case there is a future bug in setting the GoldenForDataSnapshotURI string.

cmd/ateapi/internal/controlapi/workflow_resume.go
zoez7: Done, Using pausedContentScope to resolve the template's content scope.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -34,23 +34,71 @@ import (
...
 
+// restoreKind identifies which boot source the restore step drives. Resolved
+// once by loadActorForResume so the restore step and its telemetry dispatch
+// on it instead of re-deriving the precedence from raw actor fields.
+type restoreKind int
+
+const (
+	// restoreColdBoot boots from the template spec: no snapshot anywhere.
+	// The zero value, so a zero resumeSnapshotSource means cold boot.
+	restoreColdBoot restoreKind = iota
```
zoez7: restoreColdBoot is not the right word. It should be restoreFromTemplate, as "cold boot" means booting from data only snapshot.

PR #1496, opened 2026-09-04
Name the gvisor root container as "_pause" so no actor container can collide with it

### LINKED ISSUE #1495: ActorTemplate's container should not use name "pause"
Because ateom creates a container named "pause" for every actor. We should add validation for that.

HavenXia:
Because ateom creates a container named "pause" for every actor.

Fixes #1495

- [ ] Tests pass

### DIFF
--- a/cmd/ateom-gvisor/main.go
+++ b/cmd/ateom-gvisor/main.go
@@ -45,6 +45,7 @@ import (
 	"github.com/agent-substrate/substrate/internal/childreap"
 	"github.com/agent-substrate/substrate/internal/contextlogging"
 	"github.com/agent-substrate/substrate/internal/imagecache"
+	"github.com/agent-substrate/substrate/internal/ocispec"
 	"github.com/agent-substrate/substrate/internal/otlprelay"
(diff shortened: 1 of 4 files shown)


sfunkenhauser (dismissed):

cmd/ateapi/internal/controlapi/actor_template.go
```diff
@@ -196,6 +196,17 @@ func ValidateCustom_HTTPGetAction_Path(_ context.Context, _ operation.Operation,
 	return nil
 }
 
+// pauseContainerName is the container the ateom creates for each actor,
+// an application container using the same will would collide with it.
+const pauseContainerName = "pause"
```
BenTheElder: we should consider using a namespacing solution or something instead.

IIRC Kubernetes doesn't prohibit this.

HavenXia: https://github.com/agent-substrate/substrate/issues/1495#issuecomment-5576057722
Julian has a better suggestion IMO, can you take a look at that? I can update this one to such

BenTheElder (approved):

PR #1501, opened 2026-09-04
Smoosh CredentialBroker and WorkerCapacity into AteomSupport

ahmedtd:
We shouldn't have a bunch of different gRPC services for atelet to provide services to ateoms.  Combine the two that currently exist into one (AteomSupport).


BenTheElder (commented):
second commit LGTM

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -57,6 +57,19 @@ service Control {
...
+  rpc MintActorJWT(MintActorJWTRequest) returns (MintActorJWTResponse) {}
+
+  // Create a Substrate-issued SPIFFE certificate asserting the actor identity.
+  //
+  // * Called by atelet to provision an atunnel with a certificate for
+  //   communication with the egress gateway.  TODO(ahmedtd): Migrate this use
+  //   case to a distinct certificate to prevent actor/atunnel confusion.
+  // * Called by the egress gateway when actor client certificate injection is
+  //   configured for outbound requests.
+  rpc MintActorCertificate(MintActorCertificateRequest) returns (MintActorCertificateResponse) {}
```
EItanya, line 72: I'm not so sure I agree with this change. From a logical/security perspective these seem fairly different than the rest of these services. I understand that from an RBAC perspective we _can_ treat them differently, but it can help to logically separate things which have different responsibilities.

ahmedtd:
This is now unstacked.

pkg/proto/ateapipb/ateapi.proto
ahmedtd: I think the likely end state here is:

* MintActorJWT/Certificate in Control API.  These are called by the egress gateway, as well as customer harnesses in front of substrate, to get credentials for actions that should be undertaken as "the actor".

* MintAteomActorCertificate in a new, atelet/ateom-focused API also served by ateapi.  This would vend certs for atunnel to connect to the egress gateway.  There's still some active debate on whether or not the API actually needs to be split in this way.

ahmedtd: I think the specifics of what the support API look like also depend on whether atelet will still be around at GA, or if multi-actor ateoms will be connected directly to ateapi.

EItanya: Strictly based on the comments above it sounds like there are open questions here which should be resolved before this PR?

ahmedtd: No, I think the only question is which gRPC "service" the RPCs will be in on ateapi.  The actual form and purpose of the RPCs is pretty clear.

PR #1504, opened 2026-09-04
Don't run PR tests on docs-only changes

thockin:
I've seen a few PRs which are just wasting CI time.

I don't know if this actually WORKS, of course...

### DIFF
--- a/.github/workflows/pr-workflow.yaml
+++ b/.github/workflows/pr-workflow.yaml
@@ -17,6 +17,9 @@ on:
   pull_request:
   push:
     branches: [main]
+    paths-ignore:
+      - 'docs/**'
+      - '**.md'
   schedule:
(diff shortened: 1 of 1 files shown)


BenTheElder:
You can leave CI broken this way, if you're not careful.

e2e: sure

verify? ... one spelling linter and then CI is wedged on the next PR because we skipped it.

thockin:
I'm just looking for a simple path to not waste time.

BenTheElder:
Can we configure this per job? so we skip e2e and run the linters or something?

thockin:
Seems like no, only the inverse (run job when **/*.go is touched) - that seems even more brittle

BenTheElder:
> Seems like no, only the inverse (run job when **/*.go is touched) - that seems even more brittle

Yes. strongly discourage that even more.

What if we split the workflow into two?

PR #1505, opened 2026-09-04
demos/counter: introduce server and improve code readability

rakyll:
While I reorganized the structure and naming, the file is still heavily overloaded with features, impacting overall readability. In this change:

- Encapsulate mutable state into a `server` struct that implements `http.Handler`, eliminating package-level global variables. Move `main()` to the top of the file for immediate entrypoint readability and clean up naming across struct fields and helper methods.

- Also improve listener initialization by binding sockets synchronously before spawning background worker goroutines, and replace the leaking `time.Tick` with `time.NewTicker`.

- Follow up: We are testing some features through this demo but we should remove them and keep this as a simple "increment" demo. When others read this demo, they should be able to focus on file based counter and how it's persisted.

### DIFF
--- a/demos/counter/counter.go
+++ b/demos/counter/counter.go
@@ -38,185 +38,218 @@ import (
 	"github.com/spf13/pflag"
 )
 
-var (
-	requestCount             uint64
-	ready                    atomic.Bool
-	fileMutex                sync.Mutex
-	sigtermSleepDurationSecs atomic.Int64
-)
-
-func incrementFileCounter(filePath string) int {
-	fileMutex.Lock()
-	defer fileMutex.Unlock()
-	counter := 0
-	data, err := os.ReadFile(filePath)
-	if err == nil {
-		if i, err := strconv.Atoi(string(data)); err == nil {
-			counter = i
-		}
-	}
-	counter++
-	err = os.WriteFile(filePath, []byte(strconv.Itoa(counter)), 0o644)
-	if err != nil {
-		return -1
-	}
-	return counter
-}
-
 func main() {
-	sigtermSleepDurationSecs.Store(15)
-	fileCounterDirectory := pflag.String("file-counter-directory", "/home/counter", "Directory for file counter")
-	secondFileCounterDirectory := pflag.String("second-file-counter-directory", "", "Directory for a second file counter; empty disables it. Used to exercise an Actor with more than one durable volume")
-	validateExistingFilePath := pflag.String("validate-existing-file-path", "", "Path to existing file to validate reading")
-	extraPort := pflag.Int("extra-port", 0, "Additional port to listen on, for exercising atenet-router's arbitrary-port ingress support; 0 disables it")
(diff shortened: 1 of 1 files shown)


rakyll:
I'm actually going to close this for now. We should revisit this demo and simplify it. I'll send a follow up change for the ticker leak.

demos/counter/counter.go
```diff
@@ -38,185 +38,218 @@ import (
...
+	tcpPort := pflag.Int("tcp-port", 0, "Port for TCP echo to test atunnel CONNECT ingress; 0 disables it")
+	pflag.Parse()
+
+	ctx := context.Background()
+	slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stdout, nil)))
+
+	srv := newServer(*counterDir, *secondCounterDir, *validateExistingFilePath)
+	srv.handleSignals(ctx)
+
+	go startHTTPServer(ctx, ":80", srv, "Starting counter server on port 80")
```
dberkov: The PR description says listener sockets are now "bound synchronously before spawning background worker goroutines", but all three listeners still bind asynchronously inside the goroutines (`http.ListenAndServe` at line 202, `net.Listen` at line 222), and `srv.setReady()` at line 72 runs with no guarantee any socket is bound.

If port 80 is already taken, main still flips readiness and logs "Readyz now reports OK" before the goroutine discovers the bind error and exits — a "ready" process with no listening socket, and a checkpoint taken in that window snapshots `ready=true` for a server that never served.

Suggested shape: `net.Listen` in `main` (fail fast on bind errors), `http.Serve(ln, handler)` in the goroutine, and `setReady()` only after all binds succeed.

demos/counter/counter.go
```diff
@@ -38,185 +38,218 @@ import (
...
+	fmt.Fprint(w, "ok\n")
+}
+
+func (s *server) storeShutdownDelay(w http.ResponseWriter, r *http.Request) {
+	raw := r.URL.Query().Get("duration")
+	if raw == "" {
+		http.Error(w, "missing duration parameter", http.StatusBadRequest)
+		return
+	}
+	secs, err := strconv.ParseInt(raw, 10, 64)
```
dberkov: This validation accepts any non-negative int64, but `handleSignals` computes `time.Duration(secs) * time.Second` (line 140), which overflows for `secs` > ~9.2e9. For example `?duration=9223372036854775807` passes validation and returns 200, yet the product wraps to a negative duration, so `time.Sleep` returns immediately and the process exits on SIGTERM with ~0s grace — the opposite of what the caller asked for (`termination_test.go` relies on this endpoint to keep pods alive during shutdown).

Pre-existing with `Atoi` on 64-bit, but since this rewrite touches the exact validation, it's worth adding the upper bound, e.g. `secs > math.MaxInt64/int64(time.Second)`.

rakyll:
See https://github.com/agent-substrate/substrate/pull/1506 instead.

PR #1506, opened 2026-09-04
demos/counter: improve atomic use and parse duration as int64

rakyll:
Changes
- **Use `atomic.Uint64`**: Change `requestCount` from `uint64` to `atomic.Uint64`.
- **Direct `int64` parsing**: Use `strconv.ParseInt directly without intermediary `int` casts.

### DIFF
--- a/demos/counter/counter.go
+++ b/demos/counter/counter.go
@@ -39,7 +39,7 @@ import (
 )
 
 var (
-	requestCount             uint64
+	requestCount             atomic.Uint64
 	ready                    atomic.Bool
 	fileMutex                sync.Mutex
 	sigtermSleepDurationSecs atomic.Int64
@@ -89,7 +89,7 @@ func main() {
 	defaultMux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
 		fileCounter := incrementFileCounter(filepath.Join(*fileCounterDirectory, "a.txt"))
-		memoryCounter := atomic.AddUint64(&requestCount, 1)
+		memoryCounter := requestCount.Add(1)
 		currentIP := getCurrentIP()
 
 		fileContentStr := ""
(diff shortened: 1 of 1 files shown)


demos/counter/counter.go
```diff
@@ -215,7 +215,10 @@ func main() {
 	slog.InfoContext(ctx, "Count", slog.Int("count", count), slog.String("fshash", hashRandomFile()))
 	count++
 
-	for range time.Tick(10 * time.Second) {
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
```
dberkov: This rewrite fixes no leak, and the `defer ticker.Stop()` is unreachable: since Go 1.23 (go.mod is on 1.27.0), unreferenced `time.Tick` tickers are GC-collectible — and this ticker lives for the whole process anyway. The `for range ticker.C` loop is the final statement of `main` with no `break`/`return`, and every exit path goes through `os.Exit` (SIGTERM goroutine, listener-failure paths), which skips defers, so `Stop()` can never execute.

Suggest either keeping the one-line `for range time.Tick(10 * time.Second)` (the Go 1.23+ docs explicitly say this remains fine), or dropping the dead `defer` and the "leak fix" framing from the commit message.

dberkov:
The title, body, and commit message all describe a third change — "parse duration as int64" via `strconv.ParseInt` — but it is not in the diff: `/set-sigterm-sleep` still uses `strconv.Atoi` plus an `int64(d)` cast, byte-identical to `main`.

As is, the commit message would land on `main` describing a change that never happened (per `AGENTS.md`, commit messages should describe the change and stand on their own). Please either implement the `strconv.ParseInt` change or strip that claim from the title/body/commit message.

demos/counter/counter.go
rakyll: We don't need this change. It's picked up automatically by linting tools that are not yet up to date. Reverted.

rakyll (commented):
PTAL

mayawang (approved):
LGTM. The ParseInt change is in the diff at head now, so dberkov's point is covered.

`atomic.Uint64 `is worth more than tidiness here, it makes any future non-atomic access a compile error rather than a silent race, and it confirms line 92 is the only reader.

One optional nit, pre-existing, so ignore if you'd rather keep the diff tight. ParseInt with `bitSize 64 accepts up to ~9.2e18, and time.Duration(d) * time.Second` on line 81 overflows int64 nanoseconds above roughly 9.2e9 seconds. So ?duration=10000000000 wraps negative and time.Sleep returns straight away, which is the opposite of what an e2e test asserting graceful shutdown would expect. Atoi behaved the same way on 64-bit so it's not a regression, but an upper bound alongside the existing d < 0 check would close it.

mayawang:
One more thing, spotted after I approved. The commit message still lists a change that isn't in the diff, same shape as Dima's point:

▎ Fix time.Tick leak: Replace time.Tick with time.NewTicker and ensure it is released via defer ticker.Stop()

Line 218 is unchanged, still for range time.Tick(10 * time.Second). The PR body only lists the two changes, so I think it's just the commit message that drifted.

I'd strip the bullet rather than implement it. That loop is the last statement in main() and never exits, so it's the case time.Tick's docs call fine, and a deferred Stop() wouldn't run anyway given the SIGTERM path goes through os.Exit.

Other two look good.

rakyll:
PTAL, commit message is fixed.

PR #1509, opened 2026-09-05
imagecache: Re-unpack a layer retired while joining its flight

### LINKED ISSUE #1076: TestConcurrentEnsureImageAndEvict flakes

igooch:
Fixes #1076. Overlaps #1079 — both fix the same flake, only one should merge.

## The bug

`ensureLayer` and `retireLayer` shared one singleflight key per layer. That sharing *was* the retire/reuse interlock, but it made the pull path depend on something singleflight cannot express: a caller whose `Do` **joins** a flight rather than leading one learns nothing about what it joined. When the flight was a retirement, `ensureLayer` returned the layer path anyway — and the rename had just moved it aside. `pull` recorded a layer with nothing behind it, and its re-verify failed a healthy pull with `layer dir vanished during pull (evicted?)`.

#1076 attributes this to an eviction cutoff postdating `ensureLayer`'s mtime refresh. That can't fire at the default `minAge = 2m`: `cutoff = now - 2m` while a refreshed dir has mtime ≈ `now`, so `retireLayer` always vetoes. That window is real only at `minAge ≈ 0`.

## The fix

Use a lock for the interlock, and leave the singleflight to do only what it's good at.

- **Interlock**: a per-layer `semaphore.Weighted`, held by `ensureLayer` across its stat and refresh-or-unpack, and by `retireLayer` across its stat and rename. The two still can't interleave, but a waiter now *waits* and then re-reads the pool, instead of inheriting a verdict it can't interpret. Retirement uses `TryAcquire`, so a GC pass never blocks behind a download — same effect as the old join-is-a-veto rule.
- **Singleflight**: deduplication only, where joining is unambiguous. The leader's outcome is the joiner's — one download per herd, failure included, as it was before eviction existed.

The interlock is per layer, never shared between them: it is held for the length of an unpack, so two layers on one lock would serialize unrelated downloads and let a retirement veto — and so re-date for a whole min-age window — a layer it could have taken. `semaphore` (already a dep, same module as `singleflight`) rather than `sync.Mutex` because retirement needs `TryAcquire` and the pull path gets a context-aware `Acquire`; `k8s.io/utils/keymutex` has no try-variant. The `Acquire` falls back to `TryAcquire` on a context error, serving only the cache hit: the semaphore prefers the context error over a free slot, and a cancelled pull leading the flight for an already-cached layer must not fail the lookup for healthy joiners — but with the layer absent, the context error is returned rather than starting a download the layer streams (bound to the pull's parent context) would never cancel. Entries are created on first use and never removed — the one piece of per-layer bookkeeping eviction does not reclaim — because releasing one safely while a caller waits on it needs refcounting.

No behavior change outside the interlock: the layer-presence check keeps the semantics it always had (any stat failure counts the layer absent), now shared by its three call sites as `layerFSPresent`.

## Testing

The flake never reproduced naturally (~49,000 iterations, zero failures). With a throwaway instrumented build that widened the retire window and forced the evictor to win the race, joins-that-found-no-dir matched failures 1:1 in every run. Against this branch the same harness gives **0 failures at every head start**, versus 66/182/186 before.

Five committed tests pin the interlock:

- retirement vetoed while the interlock is held
- ensure waiting out a retirement, then repacking
- ensure waiting out a *failed* retirement, then reusing
- joined ensure sharing its leader's failure
- distinct layers never sharing an interlock
- a cancelled context not failing a lookup whose layer is already cached

The three that race release the held lock only once the call under test is observably blocked, so they cannot pass without exercising the contended path.

Full package `-race`: 74 pass, 0 fail. `TestConcurrentEnsureImageAndEvict` 300× race-clean, `TestRetireLayerVsEnsureImageRace` 200× race-clean. `make verify` clean except `metrics.sh` (needs Docker), `proto-fmt.sh` (needs `clang-format`), `shellcheck.sh` — the latter two fail identically on `main`, and this touches no metrics, protos, or shell scripts.

## Deliberately not fixed here

A joiner inherits a cancelled leader's `context.Canceled`, so pull B can fail on pull A's caller going away. That predates this PR — `ensureLayer` on `main` already returns the leader's error unconditionally — and the interlock neither causes nor worsens it. Undoing it means deciding whether shared work should follow any one caller's context at all, which is wider than this change; `ensureLayer`'s doc comment now records the decision.

One caveat on the usual justification: I could not find a server-side retry on the Run/Restore path that would absorb it. `EnsureImage` failures reach `maybeCrashActor`, are not crash-tagged, and come back as a plain error to the Resume RPC caller — there is no requeue. So the self-healing argument rests on the RPC client, which I have not traced.

## Follow-up (separate issue)

A pull whose last layer downloads for longer than `minAge` with no sibling completions stops touching its record, so eviction can legitimately remove its completed layers and the pull fails into a retry with the same `vanished during pull` string. That is the documented wedge-detection design, not the race this PR fixes; a time-based record heartbeat during long downloads would be its own change.

`TestConcurrentEnsureImageAndEvict` tests less than it appears: over 5,000 iterations it produced 33 eviction candidates and removed **zero** records, so the eviction-wins branch is almost never taken.

### DIFF
--- a/internal/imagecache/README.md
+++ b/internal/imagecache/README.md
@@ -244,13 +244,14 @@ skips the scan entirely, conservatively, if any record or bundle spec
 fails to read. There is no online whole-pool scan (ext4's split: bounded
 recovery at mount, fsck offline).
 
-**Deletion is two-phase.** A layer is atomically renamed to `.rm-*`
-inside the layer's singleflight (one `rename(2)` — eviction can never
-stall a pull), then removed asynchronously; a crash in between leaves
-the dir for the startup sweep. This matters because the kernel offers no
-protection here: deleting a directory that is a live overlay lowerdir in
-another mount namespace succeeds silently, leaves the overlay's behavior
-undefined, and doesn't even free the space until the mount goes away.
+**Deletion is two-phase.** A layer is atomically renamed to `.rm-*` under
+the layer's interlock (one `rename(2)`, taken only if uncontended — so
+eviction can never stall a pull), then removed asynchronously; a crash in
+between leaves the dir for the startup sweep. This matters because the
+kernel offers no protection here: deleting a directory that is a live
+overlay lowerdir in another mount namespace succeeds silently, leaves the
+overlay's behavior undefined, and doesn't even free the space until the
+mount goes away.
 
 Deleting the cache root by hand (while no actors are starting) remains
 safe — the store re-pulls whatever is missing.
--- a/internal/imagecache/gc.go
+++ b/internal/imagecache/gc.go
@@ -38,7 +38,7 @@ package imagecache
 //
 // Deletion is two-phase: the only steps that contend with the pull path
 // are one os.Remove of a record and one rename of a layer dir to a ".rm-*"
-// name inside the layer's singleflight (see retireLayer); the slow
+// name under the layer's interlock (see retireLayer); the slow
 // RemoveAll of multi-GB trees happens afterwards, on dirs nothing can
 // reach by diffID. A crash in between leaves a ".rm-*" dir for the
 // startup sweep.
--- a/internal/imagecache/imagecache.go
+++ b/internal/imagecache/imagecache.go
@@ -215,6 +215,11 @@ type Store struct {
 	imageSF singleflight.Group
 	layerSF singleflight.Group
 
+	// layerLocks holds the retire/reuse interlock per layer (see layerLock).
+	// Entries are added on first use and never removed, so this is the one
+	// piece of per-layer bookkeeping eviction does not reclaim.
+	layerLocks sync.Map
+
 	// evictMu serializes EvictUnused passes (concurrent passes would fight
 	// over the same candidates for no benefit).
 	evictMu sync.Mutex
@@ -517,7 +522,7 @@ func (s *Store) cachedImage(digest v1.Hash) (*Image, error) {
 			return nil, fmt.Errorf("invalid diffID %q in image record for %s: %w", d, digest, err)
 		}
 		dir := s.layerDir(diffID)
-		if _, err := os.Stat(filepath.Join(dir, layerFSDirName)); err != nil {
+		if !layerFSPresent(dir) {
 			return nil, nil
 		}
 		layerDirs[i] = dir
@@ -635,6 +640,7 @@ func (s *Store) pull(ctx context.Context, parsedRef name.Reference, digest v1.Ha
 	// naming a missing lowerdir. Ordered after the rewrite so even this
 	// failure path leaves the surviving layers referenced.
 	for _, dir := range layerDirs {
+		// Not layerFSPresent: the errno is part of the report.
 		if _, err := os.Stat(filepath.Join(dir, layerFSDirName)); err != nil {
 			return nil, fmt.Errorf("layer dir vanished during pull (evicted?): %w", err)
 		}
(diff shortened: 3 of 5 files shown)


ahmedtd:
Can you clean up the PR description?  It's very difficult to follow.

PR #1512, opened 2026-09-05
atelet: add filecache store skeleton

dberkov:
First slice of the node-local artifact cache proposed in #690 (see the design comment there): a generic `cmd/atelet/internal/filecache` package that will back golden-snapshot restores (today re-downloaded per actor) and sandbox-asset fetches (today raced by concurrent misses, never evicted).

This PR is the concurrency-free skeleton:

- **Keys** (`key.go`): constructor-only `Key` with two forms — `SHA256Key` for content-addressed artifacts and `URIKey` for immutable-URI sources. Canonical forms are prefix-disjoint, part joins are unambiguous (NUL separator), and the entry dir name is `sha256(canonical)` so a future GC root set can match entries by hashing keys.
- **Store layout** (`filecache.go`): `New` + `WithMinAge`/`WithFetchTimeout` (defaults mirror imagecache's 2m/10m), the `entries/` + `tmp/` + `.rm-*` layout, and the debug-only `meta.json` sidecar helpers.
- **`SweepDebris`**: startup reaper for crash debris (unfinished fetches under `tmp/`, interrupted evictions as `.rm-*`); never touches published entries; joins removal errors instead of stopping at the first.
- **`TotalBytes`**: usage measure for the future GC driver's byte budget; tolerates entries vanishing mid-walk.

Follow-ups (stacked): `GetFileTo` (singleflight fetch + atomic publication + hardlink-out), then `EvictUnused` (LRU + min-age + two-phase retire).

Tested with `go test -race ./cmd/atelet/internal/filecache/`; `make verify` clean except the pre-existing `clang-format`-missing failure in `proto-fmt.sh` (no protos touched).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #690, opened 2026-08-02
atelet: node-local cache for golden snapshots (OnGolden data resumes re-download the golden every time)


dberkov:
Follow-up from the #683 review (https://github.com/agent-substrate/substrate/pull/683): with `onResume.fromData: Golden`, every resume of a Data snapshot downloads the template's golden snapshot from object storage — including **un-pause**, which is meant to be the fast, node-local path. The golden's memory image is the dominant transfer, so this adds significant latency per un-pause. The pre-existing golden-fallback boot (fresh actor, no snapshot) has the same characteristic today.

## Proposal

Cache golden snapshot files node-locally in atelet, content-addressed by the golden snapshot URI prefix (a golden snapshot is immutable once written, and one golden serves every actor of the template on the node):

- On a `DATA_ON_GOLDEN` restore (or golden-fallback boot), download the golden's files once into a shared cache dir (alongside the existing static-files/sandbox-asset cache), then hardlink or copy them into the per-actor restore dir.
- Subsequent restores on the node skip the download entirely.
- Eviction: goldens are per-template-version; an LRU or size cap over the cache dir handles template churn. Invalidation is not needed for a given URI (immutable), only cleanup of unused goldens.

Care points: the restore dir must remain private per actor (cloud-hypervisor demand-pages from it for the VM's lifetime, so hardlinks must not allow cross-actor mutation; and the cache write must be atomic (temp + rename) like the asset cache.

🤖 Generated with [Claude Code](https://claude.com/claude-code)
EOF
)

dberkov:
## Design proposal: `filecache` — a generic node-local artifact cache in atelet

Full design written up as `docs/dev/filecache-design.md` (PR to follow); summary below. It covers this issue's golden-snapshot case and, in a later milestone, folds the sandbox-asset cache (`fetchAsset`/`fetchGVisorRelease`) onto the same library — that path has the same check-then-download race and never deletes anything.

### Shape

A small library at `cmd/atelet/internal/filecache`, porting the mechanics `internal/imagecache` already proved for OCI layers (singleflight dedup, mtime-LRU, two-phase retire) to a generic key→file cache. Two independent store instances: a **golden store** and (M3) an **assets store**, each with its own root and byte budget.

```go
func (s *Store) GetFileTo(ctx, key Key, dst string, fetch FileFetcher) error  // golden files
func (s *Store) GetFile(ctx, key Key, fetch FileFetcher) (Entry, error)       // in-place single files
[...]

PR #1513, opened 2026-09-05
atelet: add filecache singleflight retrieval (GetFileTo)

dberkov:
Second slice of the node-local artifact cache (#690), **stacked on #1512** — fork branches can't serve as PR bases, so this targets `main` and includes the skeleton commit underneath; **review only the top commit** (`atelet: add filecache singleflight retrieval`). I'll rebase after #1512 merges, which collapses the diff to just this change.

Adds the retrieval path the golden-snapshot Restore integration will call:

- **`FileFetcher`** takes a destination *path* (not an `io.Writer`) so `ategcs.FetchLocalFileFromGCSWithZstd`'s sparse-aware output plugs in unchanged; fetcher errors are wrapped with `%w` so `ateerrors.CrashIfReason` classification sees through the store.
- **`GetFileTo(ctx, key, dst, fetch)`** — atomic get-and-link. Miss: singleflight per key, flight on `context.WithoutCancel` + fetch timeout (waiters select on their own `ctx.Done()`), fetch into `tmp/`, validate, chmod `0444`, publish via one atomic rename. Hit: hard link to `dst` + last-use mtime touch under `hitMu.RLock` (eviction will take the exclusive side). No negative caching.
- Validation up front: constructor-built key, absolute not-yet-existing `dst` on the cache's mount (same *mount*, not just same disk — bind mounts fail `link(2)` with `EXDEV`), non-nil fetcher. `EXDEV` and `EEXIST` get dedicated errors.
- Scope cut, deliberate: no cross-filesystem copy fallback. Cache and consumers share `BasePath` today; if a future layout splits them (e.g. a cache-only local-SSD mount), a per-store copy mode is a backward-compatible `Option`.

Tests: 16 concurrent callers → one fetch; canceled waiter returns while the detached flight completes and warms the cache; fetch failure reaches all waiters with zero debris and no published entry, next call refetches; one-inode verification across destinations and cache copy; `0444` mutation tripwire; argument rejection; hit refreshes the LRU clock.

### DIFF
--- a/cmd/atelet/internal/filecache/filecache.go
+++ b/cmd/atelet/internal/filecache/filecache.go
@@ -45,7 +45,10 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
+
+	"golang.org/x/sync/singleflight"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


dberkov:
Closing in favor of a single consolidated PR carrying the whole M1 library as reviewable commits — the fork-based stacking made per-PR diffs and CI awkward. Replacement PR link to follow.

dberkov:
Consolidated into #1517.

PR #1515, opened 2026-09-05
atelet: add filecache eviction (EvictUnused)

dberkov:
Third slice of the node-local artifact cache (#690), **stacked on #1512 and #1513** — fork branches can't serve as PR bases, so this targets `main`; **review only the top commit** (`atelet: add filecache eviction`). I'll rebase as the stack merges.

Completes the M1 library core with pressure-driven eviction:

- **`EvictUnused(ctx, targetBytes, dryRun)`** — lock-free candidate snapshot, min-age gate, ordering unlinked-first then LRU, and a stop-at-target victim loop. Selection tracks expected bytes; **`FreedBytes` is credited only after each `RemoveAll` succeeds**, so the stats never overreport reclaimed space (a failed removal leaves a `.rm-*` dir for `SweepDebris` and shows up in the returned error instead).
- **Two-phase retire** (`retireEntry`): inside the key's singleflight + `hitMu` exclusive — re-stat, veto if the last-use clock moved (a hit landed) — then rename to a uniquely-suffixed `.rm-*`. Sharing the singleflight group with fetches means a retire that overlaps an in-flight fetch joins it and skips harmlessly, and a `GetFileTo` that joins a retire just loops and refetches (the `linkRetries` loop from #1513).
- **Stats vocabulary**: `Retired` counts namespace removals (irreversible at rename time), `FreedBytes` physical reclamation, `PendingBytes` retired-but-still-consumer-linked bytes the kernel frees later, plus per-reason skip counts.
- Physical deletion runs after all retires, outside `hitMu` and the singleflight — hits and fetches never wait on a multi-GB `RemoveAll`; `evictMu` (which serializes passes) intentionally stays held.

Tests: LRU order + stop-at-target; min-age veto; unlinked preferred over older linked; the core safety property (evicting a linked entry leaves the consumer's file readable, next get refetches); dry run touches nothing; removal-failure accounting (`Retired=1, FreedBytes=0, err != nil`); direct `retireEntry` veto and retire→sweep handoff.

### DIFF
--- a/cmd/atelet/internal/filecache/evict.go
+++ b/cmd/atelet/internal/filecache/evict.go
@@ -1,0 +1,262 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package filecache
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"syscall"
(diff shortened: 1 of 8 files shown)


dberkov:
Closing in favor of a single consolidated PR carrying the whole M1 library as reviewable commits — the fork-based stacking made per-PR diffs and CI awkward. Replacement PR link to follow.

dberkov:
Consolidated into #1517.

PR #1521, opened 2026-09-06
Use Tag UIDs in snapshot storage paths

### LINKED ISSUE #1508: Use Tag.metadata.uid as part of the snapshot URI
Snapshots in object storage can be owned by either `Actor` or `Tag` resources. Who owns the resource dictates the base path under which they are stored.

* Actor: `<base-location>/atespaces/<actor-atespace>/actors/<actor-uid>/snapshots/<snapshot-name>`
* Tag: `<base-location>/atespaces/<tag-atespace>/tags/<tag-name>`

(both `snapshot-name` and `tag-name` are random UUID generated at snapshot creation time)

I think we should change the Tag case to be:

`<base-location>/atespaces/<tag-atespace>/tags/<tag-uid>`

This has a much nicer property, which for both resources that can own snapshots, you can find all the blobs they ever referenced under a given path.

I think using deterministic naming like this will also allow us to remove the `Tag.status.in_progress_snapshot_uri` field.

@laoj2

EItanya:
Fixes #1508

Store tag snapshots at `<base>/atespaces/<atespace>/tags/<tag-uid>`. Replace `in_progress_snapshot_uri` with immutable `storage_location`, so pending and completed tags share UID-based cleanup independent of the source actor or template.

- [x] Tests pass: race-enabled control API tests and PostgreSQL tag contract tests.
- [x] Documentation updated.

Lint and code-generation verification also passed.


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -645,11 +645,11 @@ message TagStatus {
   // seeded from a tag under the same template.
   string actor_template_uid = 2;
 
-  // in_progress_snapshot_uri is where CreateTag is copying the tag's own
-  // snapshot to. It is minted before a single object is written and cleared
-  // once snapshot is set, so a tag left with it set names exactly the objects
-  // its unfinished create stranded, and deleting the tag collects them.
-  string in_progress_snapshot_uri = 3;
+  // Immutable base storage location captured when the tag is reserved. Its
```
juli4n: Can we make this documentation a bit better? Seems very focus on implementation details, but not meant by API consumers.

juli4n (approved):

PR #1523, opened 2026-09-06
Publish ActorTemplate golden snapshots as tags

### LINKED ISSUE #1507: Make ActorTemplate controller generate a golden tag
In #1417 we added support for garbage collecting snapshots from the object storage. Snapshots in the storage can be owned by either an `Actor` or a `Tag` resource. 

When it's owned by an `Actor`, during `SuspendActor` we store the new snapshot, and delete the old one. Deleting an actor also deletes the actor owned snapshot.

When it's owned by a `Tag`, the snapshot referenced is immutable. Deleting the `Tag` resource deletes the snapshot from the storage.

There is a gap today in the way we manage "golden actors". The ActorTemplate controller creates a golden actor, waits for it to pass the readiness probe, suspends it, and stores the snapshot (owned by the actor) in the `ActorTemplate` status. This is problematic, as the snapshot is owned by the golden actor, and any subsequent resume/suspend cycle will delete it.

The proposed fix is to make the controller create the golden actor, resume, wait for readiness probe, suspend, and tag it, then delete the golden actor. The tagging process is what moves the ownership of the golden snapshot to a tag.

When an Actor is created, if an ObjectRef is provided in `Actor.source_tag`, the actor uses that snapshot as a starting point. If it's not provided, the ate-apiserver resolves the golden tag referenced in the ActorTemplate status, and uses it. The golden snapshot becomes nothing more than a "smart default" during `CreateActor`.

EItanya:
Fixes #1507

Golden snapshots currently remain owned by the temporary golden actor, so another resume/suspend cycle or actor deletion can collect a snapshot still referenced by its template. The controller now copies the warmed snapshot into a published tag, deletes the golden actor, and records the tag reference on the template. Interrupted tag creation and cleanup remain retryable; template deletion cleans up both resources.

`CreateActor` resolves an explicit `sourceTag` or the template's golden tag into the actor's initial snapshot. Actors created before the golden tag is ready retain their cold-boot behavior. The golden tag uses the template UID as its name in `ate-golden`. The proto replaces `golden_snapshot` with `golden_tag` at field 1, without backward compatibility.

This PR is based directly on `main` and does not depend on #1521.

Follow-up recommendation: move the create → resume → wait → suspend → tag → delete sequence into a golden-template workflow using the existing workflow conventions. The reconciler now coordinates multiple recoverable steps; it could retain scheduling and retries while delegating that sequence to the workflow. This refactor is outside this PR.

- [x] Tests pass

Validation: full `env -u NO_COLOR make verify` passed after rebasing onto `main`. After the final proto field-number change, bindings were regenerated and the control API unit/functional tests plus proto-format and Go-format checks passed.


cmd/ateapi/internal/controlapi/template_reconciler.go
```diff
@@ -200,11 +203,29 @@ func (r *ActorTemplateReconciler) reconcileOne(ctx context.Context, ref resource
...
 		}
 
+		// A completed tag survives a crash during actor deletion or checkpointing.
+		tag, err := r.control.GetTag(ctx, &ateapipb.GetTagRequest{Tag: goldenActorRef})
+		if err != nil && status.Code(err) != codes.NotFound {
+			return 0, fmt.Errorf("while getting golden tag: %w", err)
+		}
+		if err == nil {
+			if tag.GetStatus().GetActorTemplateUid() != tmpl.GetMetadata().GetUid() || resources.ActorRefFromObjectRef(tag.GetSourceActor()) != resources.ActorRefFromObjectRef(goldenActorRef) {
+				return 0, fmt.Errorf("golden tag belongs to another actor or template")
```
laoj2: what about failing this with r.fail, so we don't retry forever if the existing golden tag matches another actor?

cmd/ateapi/internal/controlapi/functionaltest/actor_template_test.go
```diff
@@ -134,6 +139,98 @@ func TestActorTemplateCRUD(t *testing.T) {
...
+	}
+	if _, err := tc.client.SuspendActor(ctx, &ateapipb.SuspendActorRequest{Actor: resources.ActorRefFromActor(late).ToObjectRef()}); err != nil {
+		t.Fatal(err)
+	}
+	assertSnapshotPresent(t, tc, uri)
+	for _, actor := range []*ateapipb.Actor{early, late} {
+		if _, err := tc.client.DeleteActor(ctx, &ateapipb.DeleteActorRequest{Actor: resources.ActorRefFromActor(actor).ToObjectRef(), AnyState: true}); err != nil {
+			t.Fatal(err)
+		}
+	}
```
laoj2: can we check `assertSnapshotPresent(t, tc, uri)` again here to make sure we didn't delete the golden snapshot when an actor is deleted?

cmd/ateapi/internal/controlapi/actor_template.go
laoj2, line 152: can we add a unit test for DeleteActorTemplate?

cmd/ateapi/internal/controlapi/template_reconciler_test.go
```diff
@@ -714,3 +763,115 @@ func TestResync_FollowsPagination(t *testing.T) {
 		t.Errorf("queued %d templates, want 3 (all pages walked)", got)
 	}
 }
+
+func TestReconcileOne_GoldenTagRecovery(t *testing.T) {
```
laoj2, line 767: can we make these tests table-driven? It's a bit hard to read right now the per-case inputs/expectations. For example, we could have a struct like this:

```
tests := []struct {
		name string
		// tag is the golden tag an earlier pass left behind, if any.
		tag *ateapipb.Tag
		// actorDeleted seeds a pass that died after deleting the golden actor.
		actorDeleted bool
		// createTagErr, deleteActorErr and deleteTagErr fail one step of the
		// first pass. The retry runs with every step working again.
		createTagErr   error
		deleteActorErr error
		deleteTagErr   error
		// wantCreateTags is how many copies the whole scenario must cost.
		wantCreateTags int
	}{
```

laoj2 (dismissed):

laoj2 (approved):

PR #1524, opened 2026-09-07
ateom-microvm: reseed the guest CRNG on restore

### LINKED ISSUE #1449: Golden snapshot state pollution - general problem
## Summary

While working on #853 (telemetry relay) we found that the metric counter problem is actually a special case of a much more general issue: any value computed before the snapshot is frozen into every actor restored from it. If that value is supposed to be fresh, unique, or current per actor, it is silently wrong in every actor, and nothing errors.

#802 already states this well: "Anything living in checkpointed process memory (env vars) or in the container image is frozen into snapshots... every actor restored from it would observe the source actor's values."

We have several issues today that are each one instance of this class (#761, #802, #1284, #853), but no issue for the class itself. I want to use this issue to name the general problem, collect the instances, and agree on the general solution strategies, so we stop solving it one case at a time in different shapes.

## The general problem

The golden snapshot is taken after the template app finishes init. So everything the app computed during init is shared by all actors of the template, and everything it accumulated is inherited as a baseline. The pollution is not a one time event at actor creation: the polluted values sit in process memory, so a suspend snapshot carries  them too, and the actor keeps them for its whole life.

I think the instances fall into 4 types:
  
 1. Uniqueness broken. Values generated once at startup are identical in every clone: otel resource / instance id (#761), hostname (hardcoded to runsc), app generated worker ids, session ids. The worst member here is RNG state: if the PRNG or crypto DRBG state is in frozen memory, all clones produce the same "random" sequence. Duplicate UUIDs, colliding tokens, repeated nonces. The snapshot/restore world outside (AWS Lambda SnapStart, CRIU, Java CRaC) treats this as a security problem and reseeds entropy on restore (VmGenID etc). I did not find any substrate issue about guest RNG reseed on restore, for either runtime. I think this is a real gap and deserves its own child issue.

2. Currency broken. Values that were true at build time but not now: cached credentials and certs (#802 covers this, with short TTL + activation binding as mitigations), authorization decisions (#1284: "The actor comes back with the authority it had when the snapshot was taken"), cached config, feature flags, DNS caches, cached wall clock values and everything derived from them.
 
3.  Zero point broken. Accumulated quantities that should start from the actor's birth but actually start from the golden build: metric counters (#761, #853), uptime,  rate limiter buckets, circuit breaker state. This is the case we hit in the telemetry design.

4. Phase alignment. All clones share the same timer offsets and the same jitter sequence (jitter from a frozen RNG is identical jitter). A thousand actors from one  golden can retry and poll in lockstep. A thundering herd is built into the template. We never diagnosed this but I believe it exists.

## Solution strategies

From the existing work I see 5 strategies. They differ on one important axis: does the fix need the workload to do something, or can the platform do it alone. This axis matters because actor code is untrusted and many authors will not read our docs. So the rule I propose: any guarantee the platform makes (trust, attribution, security) must only depend on strategies that need zero workload cooperation. Workload side strategies can reduce errors but never carry a guarantee.

### Platform alone:

1. Correct in flight at platform boundaries. When polluted data transits a channel the platform owns, the platform fixes it without asking anyone. Telemetry is the flagship case: ateom stamps identity and clamps timestamps using activation time (#853). Note this does not generalize: most polluted state never leaves the sandbox, it only works because telemetry happens to pass through ateom.
  
2. Inject freshness at restore, platform side. Things the platform can reset before the actor becomes reachable: RNG reseed on restore (the unfiled gap above), #1284 policy reconciliation. 

3. Make stale copies harmless. Design the value so a frozen copy is useless: #802 does this for credentials with short TTLs and activation binding. The workload cannot opt out of this protection, so it counts as platform side.

### Needs workload cooperation (optimizations and best practices, never guarantees):

4. Out of snapshot channels. #802 SystemInfo volume: atelet regenerates per actor files on every Run/Restore. The platform provides the channel, but it only helps if the workload reads the files fresh instead of caching at startup. So this one is mixed: platform mechanism plus workload discipline.
 
5. Workload authoring rules. Documented rules for template authors: only compute constants at init (constant endpoints, constant addresses); keep golden build warmup off instrumented paths or reset accumulators (metrics SDK etc) as the last build step; defer init of anything per actor until after activation; when #450 lands, register a restore hook that refreshes state (reseed app level RNG, recreate the otel MeterProvider). Following these makes a workload fully clean. Not following them means bounded, documented errors, never wrong attribution.

eliranw:
## What

On the microVM restore path, ateom reseeds the guest kernel CRNG with fresh entropy through the kata-agent's ReseedRandomDev RPC, right after the guest resumes.

This implements the guest RNG reseed on restore that #1449 calls out as an unfiled gap (its Strategy 2, inject freshness at restore). Scoped to the kernel CRNG on the microVM runtime, so it does not close #1449.

## Why

Two actors restored from the same snapshot start with the same frozen CRNG state. The kernel reseeds from ambient entropy on its own, but not immediately. Measured on a live restore, two clones kept returning identical /proc/sys/kernel/random/uuid values for up to about 36ms after resume before that happened. A workload reading randomness in its first moments after resume can land in that window and get duplicate values across clones.

Cloud Hypervisor has no VmGenID device to signal the guest, so ateom reseeds it directly. On each restore it hands a fresh 32-byte nonce to the kata-agent, which mixes it into the guest CRNG. The nonce differs per restore, so clones diverge regardless of the kernel's own timing. The kata-agent already exposes this RPC, so the change is host-side in ateom.

## Behavior

Fires on every restore (golden cold-start and resume alike). Cold boot does not need it. Best-effort: a failure is logged, not fatal, so a transient agent error never fails an otherwise good restore.

## Scope and limits

- Covers the kernel CRNG only. A workload's own userspace PRNG seed lives in the checkpointed app memory and stays a workload concern (Strategy 5 in #1449).
- The reseed runs just after Resume, through the kata-agent, so it lands a few milliseconds after the vCPUs resume. In the live test the reseeded window shrank from about 36ms to about 6ms but did not reach zero. The first read or two after resume can still be frozen before the reseed lands. Fully closing it needs freezing the workload cgroup across the reseed, or a VMM VmGenID that acts before the vCPUs resume. A TODO in the code points at the VmGenID path once Cloud Hypervisor gains the device. Left as a follow-up.
- microVM runtime only. gVisor reads getrandom and urandom live from the host with no checkpointed RNG state, so restored gVisor sandboxes already get fresh entropy.

## Testing

- Unit test on the nonce generator (length, and that two nonces differ).
- Verified live on a GKE microVM cluster with a workload that continuously records /proc/sys/kernel/random/uuid, so the restore instant is observable. Two clones from one golden snapshot were compared. Stock ateom returned identical CRNG output for up to about 36ms (7 consecutive reads) after resume, varying run to run. With this change the shared window shrank to about 6ms (1 read), then diverged.

### DIFF
--- a/cmd/ateom-microvm/restore.go
+++ b/cmd/ateom-microvm/restore.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"log/slog"
@@ -62,6 +63,37 @@ func restoreMemMode(ctx context.Context, info ch.VMMInfo) string {
 			slog.String("mode", ch.MemRestoreEager))
 	}
 	return ch.MemRestoreEager
+}
+
+// reseedGuestCRNG mixes fresh, per-restore entropy into the restored guest's kernel
+// CRNG through the kata-agent (see the call site in restoreFullScope for why a restore
+// needs this). The nonce is a throwaway; its only job is to differ between restores so
+// that clones of one snapshot diverge instead of producing identical randomness.
+func reseedGuestCRNG(ctx context.Context, actorUID string) error {
+	nonce, err := newReseedNonce()
+	if err != nil {
+		return err
+	}
+	dctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	ac, err := dialAgentRetry(dctx, kata.VsockSocketPath(actorUID), 5*time.Second)
+	if err != nil {
+		return fmt.Errorf("dialing kata-agent for reseed: %w", err)
+	}
+	defer ac.Close()
+	if err := ac.ReseedRandomDev(dctx, nonce); err != nil {
+		return fmt.Errorf("reseeding guest CRNG: %w", err)
+	}
+	return nil
+}
+
+// newReseedNonce returns 32 bytes of fresh entropy to mix into the guest CRNG.
+func newReseedNonce() ([]byte, error) {
+	nonce := make([]byte, 32)
+	if _, err := rand.Read(nonce); err != nil {
+		return nil, fmt.Errorf("generating reseed nonce: %w", err)
+	}
+	return nonce, nil
 }
 
 // RestoreWorkload brings the actor back from a snapshot, on a possibly different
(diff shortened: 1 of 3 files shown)


eliranw:
@BenTheElder would love a look at this

cmd/ateom-microvm/restore.go
```diff
@@ -353,6 +385,30 @@ func (s *AteomService) restoreFullScope(ctx context.Context, p actorBootParams,
...
+	// same "random" values. Cloud Hypervisor has no VmGenID device to signal the guest,
+	// so we feed fresh per-restore entropy through the kata-agent's ReseedRandomDev, which
+	// mixes it into /dev/random and reseeds. Best-effort: a failure leaves the actor
+	// running rather than failing the restore, but it is logged because the result is
+	// silently non-unique randomness.
+	// ponytail: this runs just after Resume, so a workload that reads randomness in the
+	// first instants after resume can outrun the reseed. Fully closing that needs a
+	// freeze/thaw around the reseed, or a VMM VmGenID that acts before the vCPUs resume.
+	//
+	// TODO: switch to a Cloud Hypervisor VmGenID device once it exists. clh has no such
```
BenTheElder, line 400: is there an upstream bug?

eliranw: https://github.com/cloud-hypervisor/cloud-hypervisor/pull/8907 - Seems like they are working on a PR upstream

Issue #1526, opened 2026-09-07
[Bug]: An actor whose worker pod goes away is permanently unrecoverable, and its clean snapshot cannot be used to bring it back


mayawang:
### What happened?

Any actor that is not already `SUSPENDED` when its worker pod disappears is moved to
`ACTOR_STATE_CRASHED` with its worker assignment cleared
(`cmd/ateapi/internal/controlapi/workflow_worker_delete.go`, `releaseBoundActor`).

**`CRASHED` is terminal.** Every verb that could bring the actor back is refused:

- `resume actor` → `FailedPrecondition: AssignWorker prerequisite not met (got: ACTOR_STATE_CRASHED, want ACTOR_STATE_SUSPENDED or ACTOR_STATE_PAUSED)`
- `suspend actor` → `FailedPrecondition: MarkSuspending prerequisite not met (want ACTOR_STATE_RUNNING or ACTOR_STATE_PAUSED)`
- there is no `recover`, `restart` or `reset` verb

**And the actor still holds a clean snapshot that no API can start it from.** `status.latestSnapshot`
points at a complete snapshot in GCS, taken before the crash. The state is intact, addressable and
unusable. The only way out is to delete the actor and create a new one, losing everything it held.
Every subsequent request to it returns a 503 the caller cannot recover from.

We hit this by scaling a `WorkerPool` from 12 replicas back to 9. Two of ten actors happened to be
awake on removed pods. Both died in the same instant, 50 minutes before anyone looked at the
cluster.

**The documented Substrate operations are safe, and this report is scoped accordingly.**
`docs/upgrade.md` ground rule 2 says not to edit a serving worker pool, and runbook step c requires
every actor on a node to be suspended by hand before the node moves. A version roll done by the book
loses nothing, and our scale-down was the thing the ground rules tell you not to do. This is not a
report about the upgrade path.

**What is uncovered is infrastructure-level.** GKE node auto-upgrade, spot and preemptible nodes,
OOM kills and node auto-repair all reach `releaseBoundActor`, none is a Substrate verb, and none is
mentioned in any install prerequisite. Auto-upgrade is the one that matters: it is **on by default**
and it fires on the cloud provider's schedule rather than the operator's.

**`DrainWorker` is not a mitigation, which is worth stating explicitly.** It deliberately leaves the
resident actor in place; `TestDrainWorker_KeepsAssignment` asserts "the 1 left in place". Drain gates
[...]

PR #1528, opened 2026-09-07
docs: warn that worker node pools must not auto-upgrade or use spot nodes

mayawang:
Documentation half of #1526.

## What

Two additions, no code:

- `tools/setup-gcp/README.md`, Create Cluster: a prerequisite saying node auto-upgrade must be off on any pool that runs workers, and that those nodes must not be spot or preemptible. It sits next to the beta-API warning, which is the block the quickstart already points bring-your-own-cluster users at.
- `docs/upgrade.md`: the runbook assumptions now state the same thing, and ground rule 2 says what actually happens to an actor when a serving worker pool is edited, rather than only that you should not edit one. Scaling a serving pool down is called out explicitly, since that is not obviously "editing" it.

## Why

An actor that is awake when its worker pod goes away moves to `ACTOR_STATE_CRASHED`. That state is terminal: `resume` and `suspend` are both refused, there is no recover verb, and the `latestSnapshot` the actor still holds cannot be used to start it. The actor has to be deleted and recreated.

`buildCreateClusterRequest` sets no `Management` on the node pool, so GKE's default applies and auto-upgrade is on. Nothing in the install documentation mentions it. A cluster built exactly as documented can therefore lose actors on Google's maintenance schedule, and the operator has no way to know that in advance.

Found during acceptance testing on `release-0.1` at `c48b3a3c`, where a `WorkerPool` scaled from 12 replicas to 9 destroyed two of ten actors permanently. That scale-down was our mistake and ground rule 2 already covered it, which is why this PR is scoped to documentation. What the docs did not cover is that the consequence is permanent, or that a trigger exists which fires without an operator doing anything.

## Scope

This does not fix the underlying behaviour and is not a substitute for #1526. Auto-repair, preemption and OOM kills reach the same path and no prerequisite can close them. Telling operators to disable a standard Kubernetes safety feature is a reasonable answer for this release and not a reasonable one for 1.0, so the precondition relaxation in #1526 is still owed.

## Testing

Docs only. `make test` passes. `make verify` does not complete in my environment: `hack/verify/codegen.sh` fails creating the Locust codegen virtualenv because `python3-venv` is not installed locally, which is unrelated to this change.

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -27,6 +27,13 @@ a version suffix, and the installed ate-api-server serves
 install instead, because its DaemonSet selector cannot be changed in
 place.
 
+Node auto-upgrade is off on every node pool that runs workers, and
+none of those nodes is spot or preemptible. The roll paces node moves
+one at a time so that no actor loses state; a node the cloud provider
+moves on its own schedule ignores that pacing, and any actor awake on
+it is destroyed. See the
+[Create Cluster warning](../tools/setup-gcp/README.md#2-create-cluster).
+
 Actor snapshots are readable by both the old and the new build. An
 actor can therefore suspend on one version and resume on the other in
 either direction, which is what lets the two versions serve side by
@@ -48,7 +55,14 @@ Three things break an upgrade.
    same node, and old workers end up next to the new atelet: exactly
    the version skew the roll exists to prevent.
 2. **Do not edit a serving worker pool.** The controller would roll
-   the pool's Deployment straight through live actors.
+   the pool's Deployment straight through live actors, and an actor
+   that is awake when its worker pod goes away does not survive it.
+   It moves to `ACTOR_STATE_CRASHED`, which is terminal: `resume` and
+   `suspend` are both refused, there is no recover verb, and the
+   snapshot the actor still holds cannot be used to start it. The
+   actor has to be deleted and recreated, losing its state. The same
+   applies to scaling a serving pool down, which removes pods without
+   suspending the actors on them.
 3. **(If on GKE) Do not touch the node pool's label until every node
    is rolled.** A pool label update applies in place to every node in
    the pool, so the whole fleet flips at once, with no drain and no
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -27,12 +27,11 @@ a version suffix, and the installed ate-api-server serves
 install instead, because its DaemonSet selector cannot be changed in
 place.
 
-Node auto-upgrade is off on every node pool that runs workers, and
-none of those nodes is spot or preemptible. The roll paces node moves
-one at a time so that no actor loses state; a node the cloud provider
-moves on its own schedule ignores that pacing, and any actor awake on
-it is destroyed. See the
-[Create Cluster warning](../tools/setup-gcp/README.md#2-create-cluster).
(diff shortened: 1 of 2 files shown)


docs/upgrade.md
```diff
@@ -48,7 +55,14 @@ Three things break an upgrade.
    same node, and old workers end up next to the new atelet: exactly
    the version skew the roll exists to prevent.
 2. **Do not edit a serving worker pool.** The controller would roll
-   the pool's Deployment straight through live actors.
+   the pool's Deployment straight through live actors, and an actor
```
HavenXia: Suggest some wording update, since we do have eviction path implemented by @sfunkenhauser .

>  **Do not edit or scale down a serving worker pool.** Either one deletes worker pods under live actors. A deleted pod goes through the eviction path: `SIGTERM` is forwarded into actors and Substrate keeps accepting a suspend for 60 seconds, so an actor whose harness handling SIGTERM gracefully survives. Any other actor is marked `CRASHED`, which cannot be resumed; it has to be deleted and recreated.

tools/setup-gcp/README.md
```diff
@@ -95,6 +95,32 @@ Filestore CSI driver disabled).
 > podcertificate ClusterTrustBundles to be ready" and `kubectl get
 > clustertrustbundles` reports the resource type is not served.
 
+> [!WARNING]
```
HavenXia, line 98: Wording for the first paragraph

> [!WARNING]
> **Do not let GKE drain worker nodes on its own schedule, and do not run
> workers on spot or preemptible nodes.** When a worker pod is deleted,
> `SIGTERM` is forwarded into actors and substrate keeps accepting a suspend for
> 60 seconds. An actor whose harness handling SIGTERM gracefully
> survives; any other actor is marked `CRASHED`, which cannot be resumed.

tools/setup-gcp/README.md
```diff
@@ -95,6 +95,32 @@ Filestore CSI driver disabled).
...
+>   --cluster "${CLUSTER_NAME}" --location "${CLUSTER_LOCATION}" \
+>   --no-enable-autoupgrade
+> ```
+>
+> This is a management setting, so it takes effect without recreating nodes and
+> is safe to apply to a serving cluster. Node auto-repair, preemption and OOM
+> kills reach the same path and cannot be configured away, so treat the setting
+> as removing the scheduled risk rather than all of it. Change versions through
+> the [rolling upgrade runbook](../../docs/upgrade.md), which has you suspend
+> every actor on a node before the node moves.
```
HavenXia: Suggestions (key feature of our current upgrade):

has as you suspend every actor **at your own pace**

docs/upgrade.md
```diff
@@ -27,6 +27,13 @@ a version suffix, and the installed ate-api-server serves
 install instead, because its DaemonSet selector cannot be changed in
 place.
 
+Node auto-upgrade is off on every node pool that runs workers, and
```
HavenXia: We can shorten this to be 

> Nothing drains worker nodes on its own: node auto-upgrade is off on every pool that runs workers and none of them is spot or preemptible, as the [Create Cluster warning](../tools/setup-gcp/README.md#2-create-cluster)
requires.

HavenXia (approved):

Issue #1525, opened 2026-09-07
[Bug]: atenet-router's 10s default route timeout cuts off long actor turns, and the 504 cannot be safely retried


mayawang:
### What happened?

`atenet-router` ships with `--route-timeout` unset, so Envoy's route to the actor falls back to
`defaultRouteTimeout = 10s`, with a further 5s of request parking in front of it. Any actor turn
that legitimately runs longer than that is cut off mid-flight and the caller gets
`504 upstream request timeout`.

This is what a fresh install does out of the box, with no unusual configuration. For an LLM agent
harness that relays a model completion, a turn running past 10s is the normal case rather than the
exception, so the effective ceiling on a stock cluster is 15s of work.

**The 504 cannot be safely retried.** It does not say whether the actor received the turn, and it
did: the actor is `RUNNING` and still working. Retrying runs the whole turn a second time. A correct
client therefore has to stop and report a failure, and the work still in progress is discarded.
From outside, it reads as the agent hanging, and nothing in the response points at a timeout an
operator could raise.

**The worse half is where the abandoned answer goes.** The actor finishes the work the 504 gave up
on, the response has nowhere to return to, and it is delivered to whoever speaks to that actor
next, as a fluent `HTTP 200` answering somebody else's question. In our run, 6 of 10 actors were
found holding a stale answer afterwards, and one of them surfaced internal harness framing into
user-visible text. Neither end observes an error, so a consumer cannot detect this without a turn
identifier the transport does not currently carry.

Observed as **10 failures out of 10** on a burst-wake test, each at 15.9–18.4s, which is the 5s
parking budget plus the 10s route timeout plus harness overhead.

### Expected Behavior

Two things, and the first is the release-relevant one.

1. **A stock install should not cap an actor turn at 10 seconds.** The default should be a value
   that suits the workloads the platform is built for, or the install should refuse to come up
   without an explicit choice.
2. **A response whose caller has gone away must never be delivered to a different caller.** A timed-out
   turn should either be cancelled with the actor, or its response discarded. Handing it to the next
   requester is a cross-request data leak, not just a lost turn.

Item 1 is a manifest line and a constant. The value to pick is already written down in the shipped
manifest, commented out, with the drain settings that move with it named in the same comment. See
Additional Context. Item 2 is the one that needs a design decision, and it should not hold up item 1.

### Steps to Reproduce

```
1. Install Substrate from the stock manifests, unmodified:
     hack/install-ate.sh
[...]

PR #1529, opened 2026-09-07
atenet: raise the default route timeout, and size shutdown separately

mayawang:
Addresses the timeout half of #1525. Deliberately not a closing keyword: see Scope.

## The problem

`atenet-router` shipped with the workload route capped at 10s, plus 5s of parking in front of it. That is below the normal duration of the turns this platform exists to serve, since an agent relaying a model completion holds the request open for the whole generation. A turn past the ceiling returns `504 upstream request timeout`, and the caller cannot safely retry it: the actor did receive the turn and is still working, so a retry runs it twice.

Observed as 10 failures out of 10 on a burst-wake test during acceptance, each at 15.9s to 18.4s.

The same ceiling bounds gRPC server-streaming and bidi RPCs, which is what the `TODO(liorlieberman)` above the constant is about. This does not resolve that TODO. It asks for streaming to get its ceiling without imposing one on every workload, and a raised default is the blunt version it warns against, so the TODO stays in place pointing at #1291.

## Why it was not a one-line change

`defaultRouteTimeout` was doing two jobs. It was the route default, and it was the anchor the drain sequence sized its Envoy-drain window and `--drain-timeout` from.

The `--route-timeout` flag never fed the drain, and `drainTimeout()` carries a comment saying so. But the two constants were one, so raising it for the route would have taken the derived drain from ~20s to over 5m against a 60s grace period, and the kubelet would SIGKILL mid-drain. This splits them so the route number can move without the drain number following.

## What this does

Splits the constant.

- `defaultRouteTimeout` is now 5m and governs the route only.
- `drainRouteBudget` is new, stays at 10s, and is what shutdown plans for. All the drain arithmetic is unchanged, in `drainTimeout()` and in `dataplaneWindow`.

The behaviour that does not change: a turn still running after the drain budget does not survive a shutdown. Operators who need one to raise `--drain-timeout` and the grace period together, as before, and the manifest still says so.

One thing to flag, since it is the weak point. Envoy's stream idle default is also 5m and we never set it, so raising the route timeout alone would not have worked: a turn that sends nothing while the actor thinks is idle by that measure and would be reset before the route timeout was ever reached. `routeIdleTimeout()` takes the larger of the two, which leaves everything below 5m exactly as it was.

At exactly 5m, which is now the default, the two deadlines coincide and either may fire first. An idle-triggered end reaches the client as a stream reset rather than a 504, which is the worse of the two. Making the idle timer strictly later would settle it, but that changes behavior for anyone already running `--route-timeout` above 5m, so I left it alone. Happy to take the other call.

The commented-out `--route-timeout=5m` comes out of the router manifest, since it is now the default, and the comment above it is rewritten to describe lowering rather than raising.

## Scope

**This does not fix the other half of #1525 and that issue should stay open.** When a turn times out, the actor keeps working, finishes, and the response has nowhere to return to. It is then delivered to whoever speaks to that actor next, as a fluent `HTTP 200` answering somebody else's question. We saw that on 6 of 10 actors. A longer ceiling makes it rarer, not impossible, and anyone whose turns exceed 5m still reaches it.

## Testing

`go test ./cmd/... ./internal/...` passes.

- `TestRouterConfigDrainTimeoutIndependentOfRouteTimeout` is new. It pins the property this change turns on rather than the arithmetic: the derived drain must not track the route timeout, and the full sequence must fit inside the 60s grace period in the router manifest. It fails on the naive version of this change.
- `TestXdsServer_RouteTimeout/IdleTimeoutAtTheDefaultRouteTimeout` covers the equality case above, with `IdleTimeoutTracksLongerRouteTimeout` and `IdleTimeoutKeepsEnvoyDefaultWhenRouteTimeoutIsShorter` either side of it.
- `SetterOverrides` now sets 30s instead of 5m, which was silently passing against the new default whether or not the setter worked, and lowering is the direction an operator capping turn length actually goes.
- The drain derivation cases and the idle-timeout cases now use values that stay meaningful with the new default.

`make verify` does not complete in my environment: `hack/verify/codegen.sh` fails creating the Locust codegen virtualenv because `python3-venv` is not installed locally. `hack/verify/boilerplate.sh`, `gofmt` and `go vet` are clean.


LiorLieberman (commented):
Whats the best way to reproduce it?

mayawang:
> Whats the best way to reproduce it?

Easiest way is to look at what Envoy actually got, no workload needed:

```
kubectl port-forward -n ate-system deploy/atenet-router 9901:9901
curl -s 'localhost:9901/config_dump?resource=dynamic_route_configs' \
  | jq '..|.route?|select(.)|.timeout'
```

A stock install gives 10s. The same command against a cluster where I've added --route-timeout=5m gives 300s. That's the whole thing: nothing in manifests/ate-install/atenet-router.yaml sets the flag, so the route lands on the constant.

For the failure itself, any actor that takes longer than ~15s to answer gets 504 upstream request timeout, the 10s route plus the 5s parking budget in front of it. I saw it 10 out of 10 on burst-wake during acceptance, between 15.9s and 18.4s. Happy to grab you a trace off that cluster if you'd rather see it than the config.

The response-reuse half I can't hand you a recipe for. It showed up on 6 of 10 actors and I haven't pinned the trigger, which is why #1525 stays open after this merges.

LiorLieberman:
I am concerned that we seprate timeout configs more to even more constants. I already filled a previously https://github.com/agent-substrate/substrate/issues/1291. 

Ideally we have a way to control this) 

Promised to review tonight.

LiorLieberman:
I've asked @botengyao and @yanavlasov to take a look as envoy experts

cmd/atenet/internal/router/config_test.go
```diff
@@ -189,6 +189,33 @@ func TestRouterConfigExtProcMaxRequests(t *testing.T) {
...
 
+// The whole shutdown sequence has to fit inside the router pod's
+// terminationGracePeriodSeconds (60s, manifests/ate-install/atenet-router.yaml)
+// or the kubelet SIGKILLs mid-drain. That is why the derivation uses
+// drainRouteBudget and not the route timeout: with a route ceiling sized for a
+// full model generation, deriving from it would put the drain alone past five
+// minutes. This pins the independence, which the arithmetic cases below do not.
+func TestRouterConfigDrainTimeoutIndependentOfRouteTimeout(t *testing.T) {
+	const (
+		drainDelay  = 13 * time.Second // --drain-delay in the router manifest
```
botengyao: how is 13 picked?

mayawang:
Fair on the constants. This started as a default bump for something acceptance hit and then grew the shutdown half, so it's sitting right in the middle of what you want to consolidate. Happy for you to take it over or fold it into #1291, whichever is less work for you. 

One thing to keep separate: #1525 stays open whatever the timeout ends up being. The half that bit us is a timed-out turn not being abandoned, and no route value fixes that.

> I am concerned that we seprate timeout configs more to even more constants. I already filled a previously #1291.
> 
> Ideally we have a way to control this)
> 
> Promised to review tonight.

botengyao (commented):
In general, I agree we can split the drain timeout and route timeout, and raise the default workload route timeout. The route timeout is the inner end-to-end, so 10s is too low for some LLM requests.

Apart from the above, it is totally okay to split the idle_timeout, but the rule is the idle timeout is distinct to [timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-timeout), which provides an upper bound on the upstream response time; [idle_timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-idle-timeout) instead bounds the amount of time the request’s stream may be idle.

mayawang:
Thanks, and agreed on the distinction. The version you looked at didn't respect it: `routeIdleTimeout` took the max of the two, so at the new 5m default they land on the same value and either can fire. An idle-triggered end reaches the client as a stream reset where the route timeout reaches it as a 504, and only one of those is diagnosable.

Changed so they stay distinct and ordered. The idle timeout is now held 30s past the route timeout, still never tightened below Envoy's 5m. `IdleTimeoutNeverFiresBeforeTheRouteTimeout` pins it across the range rather than at one value.

Why set it at all rather than leave Envoy's default alone: a turn that emits nothing while the actor thinks is idle by that measure even though it's progressing, so raising the route timeout past 5m without touching idle wouldn't raise anything.

@LiorLieberman  this adds one more constant, which is the direction you don't want. It's a fixed 30s margin, nothing configurable, and happy to inline it if you'd rather not have the name.

> In general, I agree we can split the drain timeout and route timeout, and raise the default workload route timeout. The route timeout is the inner end-to-end, so 10s is too low for some LLM requests.
> 
> Apart from the above, it is totally okay to split the idle_timeout, but the rule is the idle timeout is distinct to [timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-timeout), which provides an upper bound on the upstream response time; [idle_timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-idle-timeout) instead bounds the amount of time the request’s stream may be idle.

cmd/atenet/internal/router/config_test.go
mayawang: It isn't picked, it's copied. 13s is the `--drain-delay` already in the router manifest, from the graceful-termination change, and I'd duplicated it here as a literal.

Now the test reads it. -`-drain-delay `and terminationGracePeriodSeconds both come out of `manifests/ate-install/atenet-router.yaml`, with the container args parsed back through the router command's own flag set, so retuning either one fails here instead of drifting away from the Go side quietly.

LiorLieberman (dismissed):

mayawang:
Hi @LiorLieberman, your approval came off this one and that's on my rebase and conflict resolution, not on anything changing. Sorry for the re-ping.

The conflict itself was one hunk in cmd.go. The dataplane flag rename that landed on main touched the --envoy-admin-address help string, and this PR had reworded --drain-timeout on the line right above it. Took both, no logic change either side.

CI's green. Mind re-stamping & merging when you get a minute?

LiorLieberman (approved):

PR #1542, opened 2026-09-08
atenet: enable the 5m route timeout in the router manifest

### LINKED ISSUE #1525: [Bug]: atenet-router's 10s default route timeout cuts off long actor turns, and the 504 cannot be safely retried

mayawang:
release-0.1 ships the workload route at the 10s default, so any turn slower than about 15s (10s route plus the 5s parking budget in front of it) returns `504 upstream request timeout`. The caller cannot safely retry: the actor received the turn and is still working, so a retry runs it twice.

This uncomments the `--route-timeout=5m` line already present in the manifest, directly under the comment describing the trap.

**Not a cherry-pick of #1529.** That PR raises the constant on `main` and removes this line as redundant. Here the constant stays at 10s, so the manifest line is the fix.

**No other timing moves.** On this branch `drainTimeout()` derives from the constant and deliberately ignores `--route-timeout`, so the drain sequence and `terminationGracePeriodSeconds` are unchanged. The route idle timeout resolves to 5m either way.

**Does not close #1525.** The other half, where an abandoned answer can surface to the next caller, is unaffected by a longer ceiling.

### DIFF
--- a/manifests/ate-install/atenet-router.yaml
+++ b/manifests/ate-install/atenet-router.yaml
@@ -180,7 +180,7 @@ spec:
         # deliberately does NOT scale with this; if you raise it and want long
         # turns to survive a shutdown, raise --drain-timeout and
         # terminationGracePeriodSeconds alongside it.
-        # - "--route-timeout=5m"
+        - "--route-timeout=5m"
         env:
         - name: POD_NAME
           valueFrom:


annapendleton:
Please make sure this is submitted into main branch prior to submission in release-0.1

we want to avoid divergence.

mayawang:
Closing this. Main first is right, and if the mitigation for release-0.1 is documenting the manifest line rather than moving the default here, this PR is the wrong shape for it.

#1529 carries the main-side question. If we do end up backporting after that lands, the cherry-pick applies clean onto this branch and the router tests pass, I checked already.

PR #1545, opened 2026-09-08
Egresspolicy impl

LiorLieberman:
Enforces an Actor's `EgressPolicy` at the egress gateway. Eeally encouraged to review commit by commit.

- `internal/egresspolicy`: one parser and evaluator for hostname patterns and CIDRs, used by ateapi to validate a policy and by atenet to enforce it, so what the API accepts is what the gateway can match.
- A per-actor policy cache in atenet (`--egress-policy-cache-ttl`, 10s default). The TTL is how stale a decision can be. A missing policy is cached as a deny; a failed fetch is not cached.
- Enforcement in the egress ext_proc handler, on the leg that knows the destination, and the Envoy manifests that give it those legs. The inner listener is now also added to the non-sdsmint gateway (already existed in the sdsmint one). (yes we dont plan to have two gateways in the future but this is a separate workstream)
- E2E coverage +  README updates.

Request Flow
- Every CONNECT is authenticated from the actor certificate. The tunnel opens only if some rule could match its destination, and the dialed `IP:port` is passed to the legs inside as filter state. An Actor with no policy gets no tunnel.
- Every request the gateway can read (cleartext HTTP on both gateways, HTTPS the sdsmint gateway terminates) is evaluated once, as the API describes: rules in order, over the request's `Host` and the address the actor dialed. The first match authorizes it; no match is a 403 `egress denied`.
- The request is sent to what the matching rule checked: a hostname match is resolved and dialed by name, an address or `all` match goes to the dialed address. On sdsmint both re-originate TLS with the SNI and certificate check taken from the `Host`. The bytes never go anywhere the rule did not look at.
- Traffic the gateway cannot read (TLS on the plain gateway, opaque TCP) is decided at the CONNECT by address rules alone, and closed when none allows it.

Follow-ups
- Hostname rules for TLS the plain gateway passes through. That needs the SNI (with a hostname) at the CONNECT.
- Credential injection (`inject_static_headers`). A matched rule that declares one is refused with 501 rather than forwarded without it.
- Metrics for policy decisions, which follow in their own PR.

- [x] Tests pass


cmd/atenet/internal/router/egress/credentials.go
```diff
@@ -0,0 +1,46 @@
...
+// applyEffects applies a matched hostname rule's effects to the request.
+//
+// TODO(liorlieberman): credential injection. For each
+// CredentialHeaderInjection, resolve credential_uri through the Credential
+// Provider RPC (a service the gateway will call when a request needs a
+// secret; the URI names the provider class and name), then set the header
+// to prefix + value with OVERWRITE_IF_EXISTS_OR_ADD so an actor-supplied
+// header never wins. Until that lands, a rule that promises an injection
+// denies: the policy author expects the secret to be attached, and forwarding
+// without it would be the API saying one thing and the gateway doing another.
```
LiorLieberman: be less precise on this comment - just have a "TODO(yufan-su) implement credential injection"

internal/egresspolicy/egresspolicy.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+)
+
+// Destination is what a request or connection is going to, as far as the leg
+// evaluating it can tell. A request names a host in :authority, so Hostname
+// or IP is set, never both; a connection has only IP and Port. A rule that
+// needs a field the leg cannot supply never matches there.
+type Destination struct {
+	// Hostname is the normalized DNS name: lowercase ASCII, no trailing dot,
+	// no port. Empty when the destination was not named by a hostname.
```
bowei, line 41: hostname never has a port (remove ", no port"), it will confuse people who will then have to check that there are hostnames that have ports somewhere in the repo.

internal/egresspolicy/egresspolicy.go
```diff
@@ -0,0 +1,282 @@
...
+type compiledRule struct {
+	hostnames []HostnamePattern
+	effects   *ateapipb.EgressRuleEffects
+	prefixes  []netip.Prefix
+	all       bool
+}
+
+// Compile parses every pattern and CIDR in policy. An entry that does not
+// parse is dropped from its rule and reported; ateapi validated it on write,
+// so this only happens on drift, and dropping fails closed. The returned
```
bowei, line 76: "so this should not happen, but fail close in this case"?

Is this actually expected for version drift?

internal/egresspolicy/egresspolicy.go
```diff
@@ -0,0 +1,282 @@
...
+// DNS-1123 subdomain, optionally prefixed with "*." to match exactly one
+// non-empty leftmost label. IP literals, ports, URLs, trailing dots, and any
+// other placement of "*" are rejected.
+func ParseHostnamePattern(raw string) (HostnamePattern, error) {
+	if raw == "" {
+		return HostnamePattern{}, errors.New("pattern is empty")
+	}
+	name, wildcard := strings.CutPrefix(raw, "*.")
+	if !isHostname(name) {
+		return HostnamePattern{}, fmt.Errorf("pattern %q is not a DNS hostname, optionally with a complete leftmost-label wildcard", raw)
```
bowei, line 179: "%q is not a valid HostnamePattern"

the extra stuff about leftmost label is confusing.

internal/egresspolicy/egresspolicy.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// HostnamePattern is one parsed HostnameRule pattern: an exact name, or a
+// wildcard standing in for the whole leftmost label.
+type HostnamePattern struct {
+	// name is the exact name, or the suffix after "*." for a wildcard.
+	name     string
+	wildcard bool
+	// dotSuffix is "." + name, precomputed: a policy may carry 256 rules of
+	// 256 patterns, and building it per match dominated evaluation.
```
bowei, line 165: comment is a bit weird after the ":"

internal/egresspolicy/egresspolicy.go
```diff
@@ -0,0 +1,282 @@
...
+		return hostname == p.name
+	}
+	label, found := strings.CutSuffix(hostname, p.dotSuffix)
+	return found && label != "" && !strings.Contains(label, ".")
+}
+
+// ParsePrefix parses an IPBlockRule CIDR. Only the canonical form is accepted:
+// no leading zeros, no bits set beyond the prefix length, IPv6 in RFC 5952
+// lowercase compressed notation, and no IPv4-mapped IPv6 addresses.
+func ParsePrefix(raw string) (netip.Prefix, error) {
```
bowei, line 210: ParseIPBlock 

(try to keep the terms consistent in the code base, prefix is not used in other places).

internal/egresspolicy/egresspolicy.go
```diff
@@ -0,0 +1,282 @@
...
+	prefix, err := netip.ParsePrefix(raw)
+	if err != nil {
+		return netip.Prefix{}, fmt.Errorf("%q is not a CIDR prefix", raw)
+	}
+	if prefix.Addr().Is4In6() {
```
bowei, line 215: why does this matter?

internal/egresspolicy/egresspolicy.go
```diff
@@ -0,0 +1,282 @@
...
+// not an IP literal. DNS-1123 labels may be all digits, so the last label is
+// checked (RFC 1123 section 2.1: a host name never has dotted-decimal form),
+// which also rejects spellings like "01.2.3.4". IPv6 literals fail the
+// subdomain check on their own.
+func isHostname(name string) bool {
+	if name == "" || len(content.IsDNS1123Subdomain(name)) != 0 {
+		return false
+	}
+	last := name[strings.LastIndexByte(name, '.')+1:]
+	return strings.Trim(last, "0123456789") != ""
```
bowei, line 268: This line is kind of strange. Seems more complex than I was thinking.

cmd/atenet/internal/router/egress/policycache.go
```diff
@@ -0,0 +1,160 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/agent-substrate/substrate/internal/egresspolicy"
+	"github.com/agent-substrate/substrate/internal/resources"
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+)
+
+// DefaultPolicyCacheTTL is how long a fetched policy stays usable, and so how
+// stale a decision can be.
```
bowei, line 35: do we cache decisions or the policy itself?

bowei: we really need the API to tell us what validity is ok. Add a TODO here.

manifests/ate-install/atenet-egress-with-sdsmint.yaml
```diff
@@ -348,12 +381,43 @@ data:
...
                     route:
                       cluster: egress_forward_proxy
-                      # The route timeout runs until the response is completely
-                      # processed, which for a stream is never, so 0 is what an
-                      # SSE or long-poll response needs to survive at all.
+                      # The route timeout runs until the response completes, which
+                      # for SSE or long-poll is never, so 0.
                       timeout: 0s
               http_filters:
               #ATE_MITM_EXTPROC_FILTER -- see the mitm_listener comment above.
```
haiyanmeng, line 388: Adding the new Substrate default ext_proc after `#ATE_MITM_EXTPROC_FILTER` means that if both the Substrate default ext_proc  and a user-owned ext_proc are used together, the user-owned ext_proc will execute first. 

Is this intended?

manifests/ate-install/atenet-egress.yaml
```diff
@@ -160,7 +361,114 @@ data:
               - name: envoy.filters.http.router
                 typed_config:
                   "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
+
+        # The passthrough chains: TLS this gateway does not terminate, and
```
haiyanmeng, line 365: Allowing HTTPS traffic to passthrough means that the egress gateway cannot enforce hostnameRules. If this is what we want to ship, we need to have proper warning to users. 

IMO, we should not ship this, since it is not a complete solution.

internal/e2e/suites/networking/networking_test.go
```diff
@@ -348,20 +357,30 @@ func accessLogField(line, key string) (string, bool) {
 }
 
 func createAndResumeActor(t *testing.T, ctx context.Context, prefix string, template e2e.Fixture) (string, *ateapipb.Actor) {
```
haiyanmeng, line 359: We can remove this function and let callers call `createAndResumeActorWithEgress` directly

internal/e2e/suites/networking/egresspolicy_test.go
```diff
@@ -0,0 +1,246 @@
...
+		t.Fatalf("fetch of the same origin by name %s returned HTTP %d, want 403 egress denied; body: %s", url, status, body)
+	}
+	t.Logf("fetch by name was denied as expected: %s", body)
+}
+
+// TestActorEgressHTTPSByHostname: the policy names example.com. sdsmint
+// terminates the TLS and polices by name: example.com 200, example.org 403.
+// The plain gateway cannot read the request and no address rule allows either
+// destination, so both connections are closed before a byte reaches an origin.
+func TestActorEgressHTTPSByHostname(t *testing.T) {
```
haiyanmeng, line 138: If we decide to have two gateways, let us split this test into two: one for the MITM gw, and the other for the gw without MITM.

internal/e2e/suites/networking/egresspolicy_test.go
```diff
@@ -0,0 +1,246 @@
...
+		t.Fatalf("fetch of a host outside the policy returned HTTP %d, want a failed fetch (502) through the plain gateway; body: %s", status, body)
+	}
+	t.Logf("closed before the handshake (plain gateway): %s", body)
+}
+
+// TestActorEgressHTTPSByAddress: the policy names the addresses example.com
+// resolves to and no hostname. The plain gateway decides HTTPS by address at
+// the CONNECT, so the fetch succeeds. sdsmint polices the request by name,
+// which no rule allows.
+func TestActorEgressHTTPSByAddress(t *testing.T) {
```
haiyanmeng, line 177: If we decide to have two gateways, let us split this test into two: one for the MITM gw, and the other for the gw without MITM.

manifests/ate-install/atenet-egress-with-sdsmint.yaml
```diff
@@ -348,12 +381,43 @@ data:
...
+              # Every request on this chain is authorized against the actor's
+              # EgressPolicy by the same sidecar; the actor comes from the filter
+              # state the CONNECT leg set, never from a header.
+              - name: envoy.filters.http.ext_proc
+                typed_config:
+                  "@type": type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExternalProcessor
+                  stat_prefix: egress_policy
```
haiyanmeng, line 395: The same `stat_prefix` is used for both ext_proc added in this PR. Let us use different prefixes.

cmd/atenet/internal/router/extproc/dispatch.go
```diff
@@ -30,10 +30,30 @@ const (
...
-// CONNECTs, and so the one that selects the egress handler. It must stay in sync
-// with the filter chain name in manifests/ate-install/atenet-egress.yaml.
-const EgressFilterChainName = "egress"
+// Filter chains of the egress gateway that call ext_proc; each is a leg of the
+// egress handler. Must match the chain names in
+// manifests/ate-install/atenet-egress.yaml and atenet-egress-with-sdsmint.yaml.
+const (
+	// EgressFilterChainName terminates the actor's outer mTLS CONNECT: the
+	// certificate is authenticated and the address rules decided here.
+	EgressFilterChainName = "egress"
```
haiyanmeng, line 39: Should we rename this to sth like `connect`?

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -67,30 +70,61 @@ const (
...
 func (h *Handler) Direction() extproc.Direction { return extproc.DirectionEgress }
 
-// HandleRequestHeaders authenticates the actor behind an egress CONNECT before
-// the gateway tunnels it out, using the actor certificate atunnel presented in
-// the mTLS handshake. Nothing the actor can write — no CONNECT header, no
-// request metadata — contributes to the identity; the only inputs are the
-// certificate the actor-identity CA signed and the control plane's own view of
-// that actor.
+// HandleRequestHeaders dispatches on the filter chain the request arrived on.
+// An empty chain name is a non-Envoy dataplane, which calls out for the
```
haiyanmeng, line 103: Is it possible that a non-Envoy dataplane lands here?

haiyanmeng:
For the egress gateway with MITM, if an actor's egress policy does not include any `hostnameRule`, the https traffic seems still getting to the inner listener. I think in this case, the CONNECT listener can make a decisive decision.

haiyanmeng:
Can you update the PR description with more details on how traffic is handled in both gateways?

manifests/ate-install/atenet-egress-with-sdsmint.yaml
LiorLieberman: done

internal/e2e/suites/networking/networking_test.go
LiorLieberman: done

internal/egresspolicy/egresspolicy.go
LiorLieberman: replaced

internal/e2e/suites/networking/egresspolicy_test.go
LiorLieberman: we wont eventually, as discussed, but splitting.

internal/e2e/suites/networking/egresspolicy_test.go
LiorLieberman: same as above.

internal/egresspolicy/egresspolicy.go
LiorLieberman: changed to ParseCIDR (given the latest rename of ipblock)

internal/egresspolicy/egresspolicy.go
LiorLieberman: edge case but if someone added a rule on an IPv4 thats written in IPv6 form (with ::fffff). The gateway calls Unmap on every address before matching, so that rule would never match, so we reject it.

internal/egresspolicy/egresspolicy.go
LiorLieberman: hmm, 

only when ateapi and atenet at different versions mid-rollout or something. And to mitigate if Parse fails we dont include it.

manifests/ate-install/atenet-egress-with-sdsmint.yaml
LiorLieberman: Thats intended for this change. The idea is that the user's ext proc does not see the injected secrets for example. I am trying to think about what are the usecases for a user extproc to want to run **after** substrate's extProc. I am not sure. I can see some cases but we would need more feedback.

manifests/ate-install/atenet-egress.yaml
LiorLieberman: We can warn users but there is no point adding warnings now. ALL traffic was permitted until now so no regression. We would need to (a) document that MITM is required for hostname rules (b) add support for hostname rules **without** MITM (based on SNI) - this will be a follow up once we have hostname in the connect (and not ip).

It is fine to say to users, if you need hostnameRules you have to MITM.

cmd/atenet/internal/router/extproc/dispatch.go
LiorLieberman: egress name already existed. We need to do a full cosmetic follow up on the codebase once we deprecate/remove the non-MITM-capable gateway.  Which will be we once a full MITM gateway working safely with knobs to enable/disable MITM

manifests/ate-install/atenet-egress-with-sdsmint.yaml
LiorLieberman: That was intended for this change. The idea is that the user's ext proc does not see the injected secrets for example. But thats a good question, leaving this comment opened. I can see cases for both.

Question comes down to:

User first, as today. The operator's extProc sees every request exactly as the actor sent it, including ones we then deny, and it never sees anything we add. In headers-only mode an injected Authorization header is precisely what a processor placed after ours would receive. The ~downside is that in the deny case by **our policy** we pay one extra round trip to user extproc service for requests we deny anyway.

Our policy first -- this has cleaner semantics. Our policy is the first gate, the plug-in only ever sees allowed traffic, and every policy denial reads egress denied unified response. The ~downside is that the plug-in would see injected credentials once they exist.


/cc @bowei as well

cmd/atenet/internal/router/egress/egress.go
LiorLieberman: depends if those non-envoy dataplanes call this extProc handler. HandleRequestHeaders is common and passing that chain name is something we only need to the envoy impl currently. I am leaving a way to support something that does not need this chain name.

cmd/ateapi/internal/controlapi/actor.go
```diff
@@ -668,17 +667,10 @@ func (s *RPCService) MintActorCertificate(ctx context.Context, req *ateapipb.Min
...
 		template = &x509.Certificate{
-			URIs: []*url.URL{
-				{
-					Scheme: "spiffe",
-					// TODO(identity): Must be configurable per-install, so that each install can set it to a unique value.
-					Host: "substrate-actor.local",
-					// TODO(identity): Prefix with "atunnel" to prevent
-					// confusion between atunnel and an actor pretending to be
-					// an atunnel.
-					Path: path.Join("atespace", dbActor.GetMetadata().GetAtespace(), "actor", dbActor.GetMetadata().GetName()),
```
LiorLieberman, line 679: TODOs were moved into `internal/resources/spiffe.go`

manifests/ate-install/atenet-egress-with-sdsmint.yaml
LiorLieberman: flipped the order to have substrate policy first.

haiyanmeng (approved):

haiyanmeng:
> * Enforcement in the egress ext_proc handler, on the leg that knows the destination, and the Envoy manifests that give it those legs. The inner listener is not also added to the non-sdsmint gateway (already existed in the sdsmint one). (yes we dont plan to have two gateways in the future but this is a separate workstream)

`The inner listener is not` -> ` The inner listener is now`

haiyanmeng:
> For the egress gateway with MITM, if an actor's egress policy does not include any `hostnameRule`, the https traffic seems still getting to the inner listener. I think in this case, the CONNECT listener can make a decisive decision.

@bowei , @LiorLieberman , we can address this in a followup PR, potentially bundled with the other changes to the EgressPolicy API. However, this specific enhancement is not blocked by any API change.

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -67,31 +72,62 @@ const (
 	xfccChainKey = "chain"
 )
 
-// Handler authenticates the actor behind each egress CONNECT.
+// deniedBody is the body of every policy denial. The reason goes to the log,
+// not to the actor.
+const deniedBody = "egress denied"
```
bowei, line 77: Probably want to put this in a bunch of functions that create the egress denied error(s) rather than have the bare const here.

cmd/atenet/internal/router/egress/egress.go
```diff
@@ -124,20 +160,94 @@ func (h *Handler) HandleRequestHeaders(ctx context.Context, md *extproc.RequestM
...
+	case decision.Allowed:
+		slog.InfoContext(ctx, "egress tunnel opened: an address rule allows the destination", attrs...)
+		res := allow()
+		res.DynamicMetadata = passthroughDestination(dest)
+		return res, nil
+	case leg == extproc.EgressFilterChainName && policy.HasHostnameRules():
+		// Only the Envoy gateway has request legs behind this one. A dataplane
+		// that calls out for the CONNECT alone sends no chain name and is
+		// refused below.
+		slog.InfoContext(ctx, "egress tunnel opened: no address rule allows the destination, requests inside it are decided one by one", attrs...)
```
bowei, line 197: "requests inside it are decided one by one" do we need to clarify this with the fact that we check the :authority: that it matches the connection-based allow?

bowei (commented):
Comments are for follow ups

bowei (approved):

PR #1548, opened 2026-09-09
Remove the top level --boot flag from ResumeActorRequest

### LINKED ISSUE #1566: ResumeActor: the semantics of the --boot flag is not exactly "boot from scratch"
The semantics of the --boot flag is not exactly "boot from scratch", during ResumeActor:

- Actor has its own external snapshot and boot=true: flag ignored, restores the actor's snapshot.
- Actor has a local snapshot and boot=true: flag ignored, restores the local snapshot.
- Actor has no snapshot, template has no golden snapshot: cold boot from the spec regardless of the boot flag.
- Actor has no snapshot, template has a golden snapshot: boot=false restores the golden, boot=true cold boots from the spec. This is the only case where the flag is used.

zoez7:
Fixes https://github.com/agent-substrate/substrate/issues/1566

Today the Resume workflow resolves its restore source in the following order
1. First check if the actor has node-local snapshot, 
2. then its own durable external snapshot, 
3. then the template's golden snapshot. 
 
The boot flag was consulted at exactly one point in that chain, where it suppressed using the golden-snapshot, which made its behavior much narrower than "boot from scratch" suggests:

- Actor has its own external snapshot and boot=true: flag ignored, restores the actor's snapshot.
- Actor has a local snapshot and boot=true: flag ignored, restores the local snapshot.
- Actor has no snapshot, template has no golden snapshot: cold boot from the spec regardless of the boot flag.
- Actor has no snapshot, template has a golden snapshot: boot=false restores the golden, boot=true cold boots from the spec. This is the only case where the flag is used.

The glutton benchmark was the only caller that set boot=true, on each actor's first resume, to report true cold-start latency as a separate ResumeActorColdStart stats row. @maxsmythe let me know if this is required.

The proto field number and name were reserved.

- [x] Tests pass


maxsmythe:
This seems like it may actually be a different kind of bug... what happens when a snapshot with durdir gets into a permanently bad state (crashed)? What are the options for reading only the persisted data?

Otherwise, if this is never an expected action, I see no reason to benchmark it other than catching the edge case where there is no golden snapshot (which could still be interesting).

zoez7:
> what happens when a snapshot with durdir gets into a permanently bad state (crashed)? What are the options for reading only the persisted data?

This is the API being discussed in https://github.com/agent-substrate/substrate/issues/1556

juli4n (dismissed):
Needs rebase.

PR #1549, opened 2026-09-09
ateom-gvisor: drop -direct from runsc restore

### LINKED ISSUE #1530: ateom-gvisor: `runsc restore -direct` serializes concurrent restores on disk latency
**What happens**

[cmd/ateom-gvisor/runsc.go](https://github.com/agent-substrate/substrate/blob/fda35d0c04e4d55c5ddb9e1ff50ed4e1ceafbf5f/cmd/ateom-gvisor/runsc.go#L287) passes `-direct` to `runsc restore`, so gVisor reads the checkpoint's memory pages files with O_DIRECT. The atelet has just written those files to `<actor>/restore-state/`, so they are already in the page cache; O_DIRECT bypasses it, and every read goes to the disk and waits on its latency. The effect is a fixed restore budget per node, about 130 MB/s of snapshot data, shared by all restores running at the same time. [gVisor's own comment on the flag](https://github.com/google/gvisor/blob/93ed0053d1ee9501766a6159d60a520c1c005430/runsc/cmd/restore.go#L47-L52) says it helps only when the files are not already cached:

In substrate the file is always warm: the atelet writes it locally right before the restore.

**Measured** (OpenClaw actors, ~765 MB snapshots, n2-standard-48, 1 TB pd-ssd)

The atelet logs each restore as phases in its "Restore timing breakdown" line; `ateom_restore` is the phase that runs `runsc restore` and reads the pages files.

| | with `-direct` | without |
|---|---|---|
| 1 restore, total | 3.5 to 4 s | 3.8 to 4 s |
| 4 at once, total each | 8 to 19 s | 4.4 to 4.8 s |
| 4 at once, `ateom_restore` phase each | 5 to 15 s | 1.0 to 1.8 s |
| 88 at once, total p50 / max | 190 s / ~200 s | 14.7 s / 22.6 s |
| Node CPU peak during the 88 | 39 to 48 cores | 17 cores |

Without the flag, the only phase that still grows with concurrency is `download` (GCS fetch and decompress), from about 1.2 s alone to 2.2 s at eight at once.

**Impact**

Any burst of concurrent resumes on one node takes minutes, and adding cores does not help. Node CPU, disk throughput, and network were all far from their limits while this happened.

**Proposed fix**

Drop `-direct`, keep `-background`. Trade-off: the pages files then occupy page cache while being read, up to about 134 GB per node at 176 workers × 765 MB, which fits on 192 GB nodes.

**To reproduce**

Create 8 actors from a large-memory template, resume them, leave them idle for 2 minutes, suspend. Resume 1, then 4, then 8 at once and time each `ResumeActor` call; watch the `ateom_restore` phase in the atelet log grow with concurrency. Rebuild `ateom-gvisor` without the flag and repeat.

chuangw6:
Fixes https://github.com/agent-substrate/substrate/issues/1530

O_DIRECT bypasses the page cache and disables readahead, so concurrent
restores on a node serialize on disk latency (about 130 MB/s per node).
The atelet has just written the pages file, so it is already cached;
reading it from there lets 88 concurrent restores finish in 15 s
instead of 190 s.

### DIFF
--- a/cmd/ateom-gvisor/runsc.go
+++ b/cmd/ateom-gvisor/runsc.go
@@ -284,7 +284,6 @@ func (r *runsc) cmdRestore(ctx context.Context, out io.Writer, containerName, ch
 		"-image-path", checkpointPath,
 		"-pid-file", ateompath.PIDFilePath(r.actorUID, containerName),
 		"-background",
-		"-direct",
 		"-detach",
 		containerName,
 	)


cmd/ateom-gvisor/runsc.go
```diff
@@ -284,7 +284,7 @@ func (r *runsc) cmdRestore(ctx context.Context, out io.Writer, containerName, ch
 		"-image-path", checkpointPath,
 		"-pid-file", ateompath.PIDFilePath(r.actorUID, containerName),
 		"-background",
-		"-direct",
+		// No "-direct": O_DIRECT serializes restores on disk latency, see https://github.com/agent-substrate/substrate/issues/1530
```
nybidari: We need not add this line. Just removing direct should be fine.

chuangw6: Done, removed the comment - I was thinking as a reminder so that people don't add it back in future.

nybidari (approved):

PR #1553, opened 2026-09-09
docs: add a metrics best-practices guide for component authors

### LINKED ISSUE #1552: docs: add a guide on how to add metrics
We have the metric registry (`docs/metrics/`), the cardinality rules in `substrate.yaml`, and the label keys in `internal/ateattr`. What we don't have is a doc that tells a component author how to actually add a metric. Today the conventions live as patterns in the existing `metrics.go` files, and people have to reverse-engineer them.

This came up twice recently:

- The last bug bash found three label bugs in instruments written by different people (#1474, #1475, #1478). Each broke a rule nobody had written down.
- The new snapshot caching layer in atelet wants to emit metrics instead of logs, and the author asked where the guidance is.

We have `tracing.md` and `otel-collector.md` under `docs/dev/best-practices/`. Metrics has the most rules and no guide.

**Proposal:** add `docs/dev/best-practices/metrics.md` covering:

- when to use a metric vs a per-actor log record
- which instrument to pick, naming, units, buckets
- label rules (no actor identity, bounded values, normalize at the producer, keys in `ateattr`)
- how to report failures on the same instrument
- the Go shape (injected meter, nil-safe instruments) and `ManualReader` tests
- how to register it in the registry and `observability.md`
- a worked example

Link it from `observability.md` and `AGENTS.md`.

Out of scope: enforcing the rules with Weaver policies, and a stability/deprecation policy for metric names. Both are separate work.

Draft PR: `git286:docs-metrics-best-practices`.

git286:
Component authors had the metric registry and the cardinality rules but no single place that says how to add an instrument: which kind to pick, how to name it, which labels are allowed, how the Go is shaped, how to test it with a ManualReader, and how to register it. Each of those conventions existed only as a pattern in one of the metrics.go files.

The guide sits beside tracing.md and otel-collector.md. It uses the instruments already in the tree as the examples, ends with a worked example of a node-local snapshot cache (a counter by outcome, a fill histogram, a size gauge and an eviction counter, and what is deliberately not a label). The registry snippets in the guide pass `weaver registry check` when appended to the real registry.

observability.md and AGENTS.md now point at the guide from the place that already told contributors to update the registry.

Fixes #1552

- [x] Tests pass

### DIFF
--- a/docs/dev/best-practices/metrics.md
+++ b/docs/dev/best-practices/metrics.md
@@ -32,7 +32,9 @@ Every metric PR has these parts. Reviewers check for each one.
 1. The instrument is defined in Go against an injected `metric.Meter`, nil-safe,
    with a `metric.WithUnit` and a `metric.WithDescription`.
 2. Every label key is a constant in `internal/ateattr`, and every label value
-   comes from a bounded set defined there.
+   either comes from a bounded set defined there or names an operator-created
+   object (a template, a pool), as the [label rules](#labels-and-cardinality)
+   allow.
 3. A unit test collects the instrument through a `ManualReader` and asserts the
    name, the unit, the instrument kind, and the label set of each series.
 4. The instrument and any new attribute are in
@@ -54,8 +56,12 @@ Rules of thumb:
 
 * **Prefer the observable form for "how many exist".** An observable callback
   reads the truth (a cache, a map) at collection time, so a missed decrement
-  cannot drift the value. Use a synchronous UpDownCounter only when there is no
-  state to enumerate at collection time.
+  cannot drift the value, and a group that disappears from the source
+  disappears from the export. Use a synchronous UpDownCounter when the change
+  is in hand on the request path and a series that keeps its last value after
+  the final decrement is acceptable, as the router's parked-request count
+  does; use the observable form when you would otherwise be mirroring a data
+  structure you already own.
 * **UpDownCounter or gauge: ask who owns the total.** The test is not whether
   the value can be summed. Ask whether the labels split a total you own and
   account for, or only tag separate readings you sampled from elsewhere. Idle
@@ -81,8 +87,12 @@ Rules of thumb:
   binary, and the same measurement from a second component then needs a second
   name.
 * Name the thing measured, not the aggregation. `duration`, `size`, `requests`,
-  `workers`. The exporter adds `_seconds`, `_bytes`, `_total`, `_bucket` from
-  the unit and the instrument kind; do not put them in the name.
+  `workers`. The unit lives in the instrument's unit field and each exporter
+  renders it its own way: the Prometheus exporter on kind appends `_seconds`,
+  `_bytes` and `_total`, while Cloud Monitoring keeps the OpenTelemetry name
+  and appends only `_bucket`, `_count` and `_sum` (the shipped dashboards
+  query `atelet.snapshot.size_bucket`). Put no unit or aggregation in the
+  name, or one backend will show it twice.
 * Use the upstream semantic-convention metric when one exists, with its name
   (`rpc.server.call.duration` comes from `otelgrpc` as is). When upstream has
   the shape but not the concept, mirror the shape under an `ate.*` name:
@@ -276,22 +286,25 @@ inside an existing component adds nothing there.
 ### Observable instruments
 
 For a value you can enumerate at collection time, register a callback rather
-than tracking increments:
+than tracking increments. This is the size instrument from the
+[worked example](#worked-example-a-node-local-snapshot-cache):
 
 ```go
-// The cache owns its index and the counts per kind partition it, so this is
+const sizeMetric = "ate.snapshotcache.size"
+
+// The cache owns its index and the bytes per kind partition it, so this is
 // an UpDownCounter and not a gauge; observable, because the index is the
 // truth to read at collection time.
-entries, err := meter.Int64ObservableUpDownCounter(entriesMetric,
-	metric.WithUnit("{snapshot}"),
-	metric.WithDescription("Number of snapshots held in the node-local cache."))
+size, err := meter.Int64ObservableUpDownCounter(sizeMetric,
+	metric.WithUnit("By"),
+	metric.WithDescription("Bytes of snapshots held in the node-local cache, by snapshot kind."))
 ...
 _, err = meter.RegisterCallback(func(_ context.Context, o metric.Observer) error {
-	for kind, n := range cache.countByKind() {
-		o.ObserveInt64(entries, n, metric.WithAttributes(ateattr.SnapshotKindKey.String(kind)))
+	for kind, n := range cache.bytesByKind() {
+		o.ObserveInt64(size, n, metric.WithAttributes(ateattr.SnapshotKindKey.String(kind)))
 	}
 	return nil
-}, entries)
+}, size)
 ```
 
 Two habits from `RegisterWorkerCount` in ateapi: if the source is unavailable,
--- a/docs/dev/best-practices/metrics.md
+++ b/docs/dev/best-practices/metrics.md
@@ -47,7 +47,7 @@ Every metric PR has these parts. Reviewers check for each one.
 |---|---|---|---|
 | How many times something happened, and its rate | Counter | `Int64Counter` | `ate.imagecache.requests` (`internal/imagecache/metrics.go`) |
 | How long or how big each occurrence was, as a distribution | Histogram | `Float64Histogram` (seconds), `Int64Histogram` (bytes) | `ate.actor.restore.duration` (`cmd/atelet/metrics.go`), `atelet.snapshot.size` (`cmd/atelet/main.go`) |
-| How many things exist right now, where the sum across labels is meaningful | UpDownCounter | `Int64UpDownCounter` (you own the increments), `Int64ObservableUpDownCounter` (you can enumerate them on demand) | `atenet.router.parking.active` (`cmd/atenet/internal/router/ingress/metrics.go`), `ate.workerpool.workers` (`cmd/ateapi/internal/controlapi/metrics.go`) |
+| How many things exist right now, where the sum across labels is meaningful | UpDownCounter | `Int64ObservableUpDownCounter` (you can enumerate them at collection time), `Int64UpDownCounter` (you own the increments; rare) | `ate.workerpool.workers` (`cmd/ateapi/internal/controlapi/metrics.go`) |
 | A level that is not summable across labels | Gauge | `Int64ObservableGauge` / `Float64ObservableGauge` | `ate.actor.stats.memory.working_set` (`cmd/atelet/statspoller.go`) |
 
 Rules of thumb:
@@ -55,7 +55,7 @@ Rules of thumb:
 * **Prefer the observable form for "how many exist".** An observable callback
   reads the truth (a cache, a map) at collection time, so a missed decrement
   cannot drift the value. Use a synchronous UpDownCounter only when there is no
-  state to enumerate, as the parking lot does.
+  state to enumerate at collection time.
 * **Do not use a gauge for something that is a sum.** Worker counts are an
   UpDownCounter because idle plus assigned is the pool and the pools sum to the
   fleet. Memory working set is a gauge because summing it across templates is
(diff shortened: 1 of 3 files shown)


docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,464 @@
...
+		return
+	}
+	if err != nil {
+		outcome = failureOutcome(err)
+	}
+	attrs := []attribute.KeyValue{ateattr.SnapshotCacheOutcomeKey.String(outcome)}
+	if outcome == ateattr.SnapshotCacheOutcomeError {
+		attrs = append(attrs, ateattr.ErrorTypeKey.String(errorType(err)))
+	}
+	i.requests.Add(context.WithoutCancel(ctx), 1, metric.WithAttributes(attrs...))
```
JeffLuoo: ate.snapshot.kind is marked required in the Weaver Registry definition below but the code doesn't say so.

Use
```
	attrs := []attribute.KeyValue{
		ateattr.SnapshotCacheOutcomeKey.String(outcome),
		ateattr.SnapshotKindKey.String(kind),
	}
```

docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,464 @@
...
+        requirement_level: required
+      - ref: error.type
+        requirement_level:
+          conditionally_required: The outcome is error.
+```
+
+The `annotations.substrate` block is substrate's own and Weaver passes it
+through: `emitted_by` names the binaries, `code_anchor` the file that creates
+the instrument, `cuj` the question an operator answers with it, and `buckets`
+the histogram boundaries. Fill all of them; a reader of the registry should not
```
JeffLuoo: Make it more specific and clear

Fill all of them -> Fill all applicable fields (including buckets for histograms); a reader of the registry should not need the code.

JeffLuoo (changes requested):

docs/dev/best-practices/metrics.md
git286: Good catch. I added kind as a parameter of recordRequest and set ate.snapshot.kind next to the outcome, so every path carries it and the code matches the registry. I also added a short note after the snippet on why a required label must be known on every path.

docs/dev/best-practices/metrics.md
git286: Done. Changed it to: Fill all applicable fields, including buckets for a histogram.

JeffLuoo (dismissed):

docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,477 @@
...
+  is one of substrate's own reasons (`internal/ateerrors`). Use
+  `ateattr.FailureAttributes(reason)`; never set the two keys by hand. This is
+  the right choice inside atelet and ateom handlers, where the gRPC status is
+  assigned only after the handler returns and would read `Unknown`.
+
+Separate a caller that gave up from a failure: `context.Canceled` and
+`context.DeadlineExceeded` are their own outcomes (`cancelled`, `timeout`) on
+the outcome label, not `error`. And record them: a cancelled pull still ran.
+Use `context.WithoutCancel(ctx)` on the `Record` or `Add` call for that case,
+so the SDK does not drop the measurement along with the request.
```
krisztianfekete: `Add` never checks `ctx.Err()`,, so nothing gets dropped. Real reason is exemplars, `WithoutCancel` keeps the span.

docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,477 @@
...
+
+A cache in front of snapshot downloads, in atelet, wants to answer: is the cache
+helping, how much does a miss cost, and how much disk does it hold. That is
+three instruments, not one:
+
+| Question | Instrument | Labels |
+|---|---|---|
+| Is it helping? | `ate.snapshotcache.requests` counter, by outcome | `ate.snapshotcache.outcome`, `ate.snapshot.kind`, `error.type` on error |
+| What does a miss cost? | `ate.snapshotcache.fill.duration` histogram, seconds, `snapshotPhaseBuckets` | `ate.snapshot.kind`, `ate.template.atespace`, `ate.template.name`, failure pair on failure |
+| How much does it hold? | `ate.snapshotcache.size` observable gauge, bytes, plus an `ate.snapshotcache.evictions` counter | `ate.snapshot.kind`; evictions also carry an `ate.snapshotcache.eviction.reason` enum (`capacity`, `ttl`, `explicit`) |
```
krisztianfekete: This should be an `UpDownCounter`.

docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,477 @@
...
+
+Every label key is a constant in `internal/ateattr/ateattr.go`, and every
+bounded value set is a group of constants there beside it. A metric never
+declares `attribute.Key("...")` locally. Add the key and its values to
+`ateattr` first; the registry entry and the code then agree by construction.
+
+Before adding a label, ask what the dashboard groups by. A label nobody will
+group or filter by multiplies the series count for nothing. Three or four
+labels is typical; the restore histogram, with the most, has eight, and several
+are conditional.
```
krisztianfekete, line 181: Can we mention the SDK's hard limit of 2000 series per instrument here?

docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,477 @@
...
+   `docs/metrics/substrate.yaml` is updated.
+6. The metrics table in `docs/observability.md` has a row for it.
+
+## Choosing the instrument
+
+| You want to know | Instrument | Go constructor | Example in tree |
+|---|---|---|---|
+| How many times something happened, and its rate | Counter | `Int64Counter` | `ate.imagecache.requests` (`internal/imagecache/metrics.go`) |
+| How long or how big each occurrence was, as a distribution | Histogram | `Float64Histogram` (seconds), `Int64Histogram` (bytes) | `ate.actor.restore.duration` (`cmd/atelet/metrics.go`), `atelet.snapshot.size` (`cmd/atelet/main.go`) |
+| How many things exist right now, where the sum across labels is meaningful | UpDownCounter | `Int64UpDownCounter` (you own the increments), `Int64ObservableUpDownCounter` (you can enumerate them on demand) | `atenet.router.parking.active` (`cmd/atenet/internal/router/ingress/metrics.go`), `ate.workerpool.workers` (`cmd/ateapi/internal/controlapi/metrics.go`) |
```
krisztianfekete: Can we use the examples at `internal/imagecache` and `cmd/ateapi/internal/controlapi`, which do follow the rules instead of ingress?

docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,477 @@
...
+        requirement_level:
+          conditionally_required: The outcome is error.
+```
+
+The `annotations.substrate` block is substrate's own and Weaver passes it
+through: `emitted_by` names the binaries, `code_anchor` the file that creates
+the instrument, `cuj` the question an operator answers with it, and `buckets`
+the histogram boundaries. Fill all applicable fields, including `buckets` for a
+histogram; a reader of the registry should not
+need the code.
```
krisztianfekete: Should we explain `golden_signals` as well?

docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,477 @@
...
+  name.
+* Name the thing measured, not the aggregation. `duration`, `size`, `requests`,
+  `workers`. The exporter adds `_seconds`, `_bytes`, `_total`, `_bucket` from
+  the unit and the instrument kind; do not put them in the name.
+* Follow the upstream pattern when one exists: `*.operation.name`,
+  `*.duration`, `error.type`. Reuse an upstream attribute verbatim rather than
+  aliasing it into `ate.*` (`error.type` and `file.name` are used as is).
+* A label key that belongs to a subsystem is rooted at the subsystem
+  (`ate.imagecache.outcome`), not under `actor`, when every actor shares the
+  thing it describes.
```
krisztianfekete: Can we add a line mentioned to use upstream metrics, if available? Also  `ate.workerpool.desired_workers`/`ready_workers` mirroring `k8s.deployment.desired_pods`/`available_pods` is a good example for the case where semconv has the shape but not the name.

docs/dev/best-practices/metrics.md
git286: You are right. I checked the SDK, Add and Record never look at ctx.Err(), and the only thing they read from the context is the span for exemplars. I removed the WithoutCancel advice and now say to pass the request context as is. The snippet is updated too. The same wrong comment is in internal/imagecache/metrics.go, I will fix that in a separate small PR.

docs/dev/best-practices/metrics.md
git286: Agreed. Bytes per kind sum to the total cache size on the node, so by our own rule it is an UpDownCounter. Changed it to an observable UpDownCounter in the example.

docs/dev/best-practices/metrics.md
git286: Good idea. Added a paragraph on the 2000 attribute sets per instrument limit, that extra series fold into one with otel.metric.overflow=true, and that the limit is per process.

docs/dev/best-practices/metrics.md
git286: Done. The table now points at ate.workerpool.workers for UpDownCounter and no longer mentions the parking instruments. The synchronous form has no clean example in the tree, so the table just notes it is rare.

docs/dev/best-practices/metrics.md
git286: Yes. Added a sentence on golden_signals, the four values, and that you list every one that applies so dashboard authors can find instruments by signal.

docs/dev/best-practices/metrics.md
git286: Added. The naming section now says to use the upstream semconv metric when one exists, and to mirror its shape under an ate name when only the shape fits, with desired_workers and ready_workers mirroring k8s.deployment.desired_pods and available_pods as the example.

docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,494 @@
...
+A new component calls one of these and defers `ShutdownProvider`; a new package
+inside an existing component adds nothing there.
+
+### Observable instruments
+
+For a value you can enumerate at collection time, register a callback rather
+than tracking increments:
+
+```go
+gauge, err := meter.Int64ObservableGauge(entriesMetric,
```
krisztianfekete: Should be an observable UpDownCounter, right?

git286: Yes, same mistake as the size row. Fixed it to an observable UpDownCounter. I also went through the whole doc once more against its own rules and fixed a few more things: the worked example said three instruments but listed four, the package path should be cmd/atelet/internal/snapshotcache per the code layout rules, the gauge rule of thumb had a reason for the memory gauges that the code does not actually give, and the 2000 limit is per reader. All in the latest commit.

docs/dev/best-practices/metrics.md
```diff
@@ -0,0 +1,501 @@
...
+
+* **Prefer the observable form for "how many exist".** An observable callback
+  reads the truth (a cache, a map) at collection time, so a missed decrement
+  cannot drift the value. Use a synchronous UpDownCounter only when there is no
+  state to enumerate at collection time.
+* **Do not use a gauge for something that is a sum.** Worker counts are an
+  UpDownCounter because idle plus assigned is the pool and the pools sum to the
+  fleet. Reserve a gauge for a level whose sum across labels means nothing: a
+  ratio, an age, a temperature. If you find yourself writing `sum by` over it
+  in a query, it was an UpDownCounter.
```
krisztianfekete: This makes your gauge example below wrong, since summing `working_set` per node is the normal query, and upstream sums `k8s.pod.memory.usage` across pods and still calls it a gauge. Better question is whether the labels split a total you own (idle plus assigned is the pool, so `UpDownCounter`) or just tag separate readings you sampled from elsewhere (gauge).

git286: Agreed and TBH this is confusing. The rule is now: did you get the number by adding and subtracting (UpDownCounter) or by reading it (gauge)? Fixed the rule, the table rows, and the two spots in the example.

krisztianfekete (approved):

PR #1558, opened 2026-09-09
Support default egress policies in ActorTemplates

### LINKED ISSUE #1324: Support a default egress policy in ActorTemplate
## Summary

Allow an `ActorTemplate` to define the initial egress policy for Actors created from it.

Today, every Actor starts without an egress policy and callers must create the policy separately. Templates should be able to provide a default when most Actors created from the template need the same access.

## Proposed behavior

Add an optional policy block to `ActorTemplate`. The block contains egress rules but no resource metadata.

When an Actor is created from the template:

- If the block is present, create the Actor's `default` egress policy in the same operation.
- Set the policy Atespace from the Actor and its name to `default`.
- Generate the policy UID, version, and timestamps normally.
- If the block is absent, preserve the current behavior and create no policy.
- Validate the template block using the same validation as an `EgressPolicy`.

The policy is copied at Actor creation time. It is not inherited dynamically: later template changes do not affect existing Actors, and callers may update or delete the Actor's policy normally.

Actor and policy creation should be atomic so an Actor is not left without its configured default policy after a partial failure.

## API sketch

```proto
message ActorTemplate {
  // Existing fields...

  // Initial egress policy for Actors created from this template.
  EgressPolicyTemplate default_egress_policy = 9;
}

message EgressPolicyTemplate {
  repeated EgressRule rules = 1;
}
```

The separate template type avoids embedding `EgressPolicy` resource metadata where it has no useful meaning.

## Acceptance criteria

- Creating an Actor from a template with a policy creates its `default` policy.
- Creating an Actor from a template without one creates no policy.
- Invalid policy rules cause ActorTemplate validation to fail.
- Updating or deleting the Actor's policy does not affect the template.
- Actor creation cannot partially succeed without its configured policy.

EItanya:
Allow ActorTemplates to define a default egress policy, validated like an EgressPolicy and copied atomically when an Actor is created. Actors without a configured default retain the current behavior; copied policies can be updated or deleted independently.

Validation: `make verify` passed, including rollback and policy lifecycle tests.

Fixes https://github.com/agent-substrate/substrate/issues/1324


haiyanmeng:
cc @bowei  @LiorLieberman

LiorLieberman:
hmmm - plus one on having support for actor default egress policies. But I think our controller / egress policy implementation should just be deny-by-default regardless if the api has the policy within the actor. Its kinda orthogonal but wanted to make sure we are on the same page that we dont want to couple having that default empty egress policy in order to trigger deny by default

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -767,6 +779,14 @@ message ActorTemplate {
 
   // +k8s:optional
   ActorTemplateStatus status = 8;
+
+  // Initial egress policy, copied atomically when an Actor is created from
+  // this template. The policy is named "default" in the Actor's atespace.
+  // Later template changes do not affect existing Actors' policies, which
```
haiyanmeng, line 799: `Later template changes do not affect existing Actors' policies`: should we make this configurable?

I imagine there are cases where users would like existing actors to pick up a EgressPolicyTemplate change automatically.

cmd/ateapi/internal/controlapi/functionaltest/egress_policy_test.go
```diff
@@ -54,6 +56,95 @@ func functionalEgressPolicy() *ateapipb.EgressPolicy {
...
+				t.Fatal(err)
+			}
+			md := policy.GetMetadata()
+			if md.GetAtespace() != testAtespace || md.GetName() != "default" || md.GetVersion() != 1 || md.GetUid() == "" || md.GetUid() == actor.GetMetadata().GetUid() || md.GetCreateTime() == nil || md.GetUpdateTime() == nil {
+				t.Errorf("unexpected policy metadata: %v", md)
+			}
+			if diff := cmp.Diff(tt.policy.Rules, policy.Rules, protocmp.Transform()); diff != "" {
+				t.Errorf("copied rules mismatch (-want +got):\n%s", diff)
+			}
+
```
haiyanmeng, line 106: The tests below this line does not seem to be related to the test table defined above. Please move them to a separate test.

EItanya:
>  hmmm - plus one on having support for actor default egress policies. But I think our controller / egress policy implementation should just be deny-by-default regardless if the api has the policy within the actor. Its kinda orthogonal but wanted to make sure we are on the same page that we dont want to couple having that default empty egress policy in order to trigger deny by default

100%, this doesn't change that. The implementation will always be deny by default, this is just a way to add default policy for all actors in a template.

PR #1559, opened 2026-09-09
docs: document the per-actor usage events channel

baizhenyu:
Documents the per-actor usage events channel that #1206 added, closing the docs request from that review.

New `Per-Actor Usage Events` section in the logging guide:
- an example record and the consumer contract (filter on `msg` + `kind`; identity rides the same label group as lifecycle events and container logs, so the guide's existing query dimensions apply unchanged, and one `labels."ate.actor.uid"` filter returns an actor's output, transitions, and usage interleaved);
- per-field semantics, including the part consumers must not get wrong: `memory_current_bytes`/`memory_working_set_bytes` are point-in-time, while `memory_peak_bytes`/`cpu_usage_usec` accumulate within an epoch whose boundary depends on `source` (cgroup restarts on restore, guest-agent survives it) — so window CPU is the increase between samples, never a sum;
- the sampling knob (`--actor-stats-poll-interval`) and the delivery contract (best-effort behind a bounded queue, independent of `--log-level`);
- the cardinality rule in bold: log-based metrics over these events must never label by actor identity.

The metrics section now names the `ate.actor.stats.*` instruments in its registry pointer and cross-links here for per-actor detail.

Every technical claim is checked against the code and the `WorkloadStatsSample` proto contract (epoch scoping, the `memory.peak`/Linux 5.19 caveat, trace-context absence, drop-warning text, flag semantics).

### DIFF
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -154,6 +154,36 @@ ateapi's `Actor crashed` is the other one. It is written once per committed tran
 
 The counter carries the same reason but no actor identity, so this record is the only way to attribute a crash to one agent. The decision-point line that precedes it (`Setting Actor to crashed due to error`) carries only `ate.atespace` and `ate.actor.name`: it is written before the Actor is loaded, so no uid exists yet.
 
+### Per-Actor Usage Events
+
+atelet emits one usage record per **executing** actor per sampling tick, from the same sweep that feeds the [`ate.actor.stats.*` metrics](#the-metric-registry). The two are the halves of one split: the metrics aggregate to the bounded template/pool label set a TSDB can hold, and everything carrying actor identity travels here, on the log stream, where cardinality is free. An idle fleet is silent by design — a worker with no executing actor emits nothing.
+
+```json
+{"time":"…","level":"INFO","msg":"Actor usage sample",
+ "logging.googleapis.com/labels":{
+   "ate.atespace":"ate-demo-counter","ate.actor.name":"counter-1","ate.actor.uid":"8f2a…",
+   "ate.template.atespace":"ate-demo-counter","ate.template.name":"counter",
+   "ate.workerpool.namespace":"ate-demo-counter","ate.workerpool.name":"counter-pool"},
+ "kind":"periodic","sandbox_class":"gvisor","source":"cgroup",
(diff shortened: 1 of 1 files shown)


docs/observability.md
```diff
@@ -154,6 +154,36 @@ ateapi's `Actor crashed` is the other one. It is written once per committed tran
 
 The counter carries the same reason but no actor identity, so this record is the only way to attribute a crash to one agent. The decision-point line that precedes it (`Setting Actor to crashed due to error`) carries only `ate.atespace` and `ate.actor.name`: it is written before the Actor is loaded, so no uid exists yet.
 
+### Per-Actor Usage Events
+
+atelet emits one usage record per **executing** actor per sampling tick, from the same sweep that feeds the [`ate.actor.stats.*` metrics](#the-metric-registry). The two are the halves of one split: the metrics aggregate to the bounded template/pool label set a TSDB can hold, and everything carrying actor identity travels here, on the log stream, where cardinality is free. An idle fleet is silent by design — a worker with no executing actor emits nothing.
+
+```json
+{"time":"…","level":"INFO","msg":"Actor usage sample",
+ "logging.googleapis.com/labels":{
```
JeffLuoo, line 163: nit - not related to this PR. But why do we tag `logging.googleapis.com/labels` here? Is it because this log is copied from cloud logging? My concern is the source of logs have it.

JeffLuoo (approved):

docs/observability.md
baizhenyu: The source really does write it — that's the mechanism, not a paste artifact. `logging.googleapis.com/labels` is one of GKE structured logging's special fields: the logging agent promotes its value into the indexed `LogEntry.labels` (what the `labels."…"` queries in this guide hit) and strips it from `jsonPayload`; a plain `labels` key would stay unindexed inside the payload. It's conditional — `actorlog.LabelsKey(metadata.OnGCE())` picks plain `labels` off GCE, so non-GCP deployments never emit the vendor key — and it's the same convention actor container logs and lifecycle events have used since actorlog introduced it; the usage emitter reuses the same function so all three promote identically (and going vendor-neutral later is a one-function change).

JeffLuoo: ack'd ty

PR #1565, opened 2026-09-09
Increase actor termination grace period to 30m

### LINKED ISSUE #1560: Consider a longer workerGracePeriod
When a worker pod is deleted, ateom traps SIGTERM and propagates it to the actor containers. Then it gives them 1 minute of grace period before hard-terminating them (SIGKILL). Note that the worker pod itself has a grace period if 1 hour.

This is problematic as today, the likelihood of an Actor entering crashed during worker pod eviction is too high. Practically any rolling update of the worker pod deployment (e.g. a WorkerPool Update) will crash the actors running on the pod. 

A few things we should consider to make this better:

* Bump the workloadGracePeriod to something like 30 minutes.
* Document that Actors that run uninterrupted for more than 30 minutes can suffer data loss.
* Add a way to recover / revert an actor from CRASHED state to the last saved snapshot (see #1556).

sfunkenhauser:
A one-minute grace period is nowhere near enough for an actor to suspend normally, so evictions were very likely to leave actors CRASHED instead of cleanly suspended.

Previously there were two 1 minute timeouts. Waiting for in-flight RPCs to finish, and then another for SIGTERM. With such a short termination grace period the in-flight RPC could have eaten a significant portion of the timeout leaving very little for the actor.

With a longer timeout this is less of a concern, and has now been collapsed into a single deadline for both operations. 

Fixes #1560

- [ x ] Tests pass


cmd/ateom-gvisor/main.go
```diff
@@ -93,9 +93,11 @@ var (
...
 const actorHTTPUpstream = "http://" + ateomnet.ActorVethIP + ":80"
 
-// Workers get a conservative shutdown period. This needs to be significantly less than the K8s
-// termination grace period for the ateom.
-const workloadGracePeriod = 1 * time.Minute
+// workloadGracePeriod is the whole budget for draining the worker on shutdown.
+// It needs to stay significantly less than the K8s termination grace period
+// for the ateom, so the escalation to SIGKILL happens here rather than as a
+// kubelet SIGKILL of ateom itself. A var so tests can shorten it.
+var workloadGracePeriod = 30 * time.Minute
```
juli4n: Why switching to a var? 

>> A var so tests can shorten it.

I don't see it being set in tests.

sfunkenhauser: Good catch, these were left over from a previous iteration and are no longer needed. The workload / containerKillTimeout are still var so they can be overridden in the tests and save ~10s.

juli4n (approved):

PR #1569, opened 2026-09-09
Log the build version when each component starts

HavenXia:
Emit one "<component> starting" line right after the logger is up, carrying the stamped version.
atecontroller also gains the --version flag it was missing.

- [x] Tests pass

### DIFF
--- a/cmd/atecontroller/main.go
+++ b/cmd/atecontroller/main.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"log/slog"
 	"os"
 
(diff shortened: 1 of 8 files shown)


annapendleton (approved):
tysm!

cmd/atecontroller/main.go
```diff
@@ -53,6 +55,7 @@ var (
 
 	ateAPIConnSpec = pflag.String("ateapi-conn-spec", "k8s:///api.ate-system.svc:443", "")
 
+	showVersion  = pflag.Bool("version", false, "Print version and exit.")
```
JeffLuoo, line 58: What is the use case of just logging the version and then terminate the controller?

HavenXia: It's the exact same `--version` flag the other components already have (ateapi, atelet, podcertcontroller .etc). the flag was provided to allow print the version of the component. 
Normal startup go through the new slog line below.

JeffLuoo: ty

JeffLuoo (approved):

PR #1577, opened 2026-09-09
A few improvements to `kubectl ate get workers`.

### LINKED ISSUE #1570: Improve `kubectl ate get workers`
I found this as UX friction during the bug bash. The `kubectl ate get workers` sub-command works different than other `get` sub-commands, for really no reason.

* It always lists workers. There is no way to get a single worker like in `kubectl ate get {actor, actor-template}`.
* The state field is not shown, so I cannot see if a worker is being, for example, being drained.
* There is no `Age` column like all other resources.
* There is no information about allocated / capacity for CPU and memory.

juli4n:
Fixes #1570

Allow getting the state of a single worker, plus aligning the columns better to the fields in the resource.

### DIFF
--- a/cmd/kubectl-ate/internal/printer/printer_test.go
+++ b/cmd/kubectl-ate/internal/printer/printer_test.go
@@ -28,9 +28,9 @@ import (
 // age rendering is deterministic, restoring it on cleanup.
 func pinNow(t *testing.T, now time.Time) {
 	t.Helper()
-	prev := timeNow
-	timeNow = func() time.Time { return now }
-	t.Cleanup(func() { timeNow = prev })
+	prev := TimeNow
+	TimeNow = func() time.Time { return now }
+	t.Cleanup(func() { TimeNow = prev })
 }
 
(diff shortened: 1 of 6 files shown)


cmd/kubectl-ate/internal/printer/printer_test.go
```diff
@@ -276,28 +293,34 @@ func TestPrintWorkersTo_Table_Free(t *testing.T) {
 	}
 	output := buf.String()
 
-	expected := `NAMESPACE   POOL     CLASS   POD     STATUS
-default     pool-1           pod-1   FREE
+	expected := `NAME   POOL     STATE                      ACTORS   CPU   MEMORY   POD             AGE
+       pool-1   WORKER_STATE_UNSPECIFIED   0/0      -     -        default/pod-1   56y
```
zoez7, line 297: I guess the `56y` here was because the CreateTime was nil?

zoez7 (approved):

Issue #1077, opened 2026-08-19
CSI volume support: follow-up improvements from PR #552


hajiler:
PR #552 ("CSI volume support") introduced the initial CSI integration—a `CSIDriverConfig` CRD, an `internal/volume/csi` plugin, and control-plane/worker-plane wiring. It landed with seven `TODO` comments in non-vendored code. This issue tracks them so they aren't lost.

### 1. Correctness / driver compatibility

- [ ] **Propagate `PublishContext` from `ControllerPublishVolume` to the node mount path**
  — https://github.com/agent-substrate/substrate/blob/d7629ed42eef637bacb9ff29c75428e3ec9904b7/internal/volume/csi/plugin.go#L125
  Current behavior: `Plugin.AttachVolume` discards the response (`_ = resp.GetPublishContext()`) because the Substrate `VolumePlugin` interface has nowhere to put it.
  Proposed fix: extend `VolumePlugin` so `AttachVolume` returns the publish context, persist it alongside the volume's assignment state, and pass it into `NodeStageVolume`/`NodePublishVolume` on the worker.
  Why it matters: drivers that require it (e.g. AWS EBS) cannot successfully mount today, so this gates any real cloud driver beyond the hostpath test driver.

- [ ] **Support configurable volume access modes instead of hardcoding `SINGLE_NODE_WRITER`**
  — https://github.com/agent-substrate/substrate/blob/d7629ed42eef637bacb9ff29c75428e3ec9904b7/internal/volume/csi/plugin.go#L236
  Current behavior: `getStandardCapabilities()` returns one hardcoded `SINGLE_NODE_WRITER` + `Mount` capability, used for every create, attach, and mount call.
  Proposed fix: add an access-mode field to the volume API type, map it to the corresponding `csi.VolumeCapability_AccessMode` enum, and validate the requested mode against the driver's reported capabilities. Unblocks `ReadWriteMany` / `ReadOnlyMany` volumes shared across actors.

### 2. Efficiency and log noise

- [ ] **Query CSI driver capabilities at plugin initialization**
  — https://github.com/agent-substrate/substrate/blob/d7629ed42eef637bacb9ff29c75428e3ec9904b7/internal/volume/csi/plugin.go#L113
[...]

PR #1578, opened 2026-09-09
csi: add publishContext to actor volumes

hajiler:
#1077 

When publishing volumes, some CSI drivers require additional metadata (e.g. device) path for attachment. This PR adds a publishContext to actor external volumes and integrates into the actor life cycle.

- [X ] Tests pass


cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -613,30 +616,52 @@ func schedulingConstraints(actor *ateapipb.Actor, tmpl *ateapipb.ActorTemplate)
...
+		resp, err := plugin.AttachVolume(ctx, volume.AttachVolumeRequest{VolumeID: vol.GetStorageVolumeId(), Node: node})
+		if err != nil {
+			return nil, fmt.Errorf("failed to attach volume %q to node %q: %w", vol.GetStorageVolumeId(), node, err)
+		}
+		vol.PublishContext = resp.PublishContext
+		vol.PublishContextNode = node
+		attached[vol.GetVolumeName()] = vol
+
+	}
+
```
Sneha-at, line 648: Should we skip the DB update if there are no external volumes to be attached? i.e `len(attached) == 0`

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -613,30 +616,52 @@ func schedulingConstraints(actor *ateapipb.Actor, tmpl *ateapipb.ActorTemplate)
...
+	storedActor, updateErr := w.store.UpdateActor(ctx, actorRef, store.PreconditionFrom(actor), func(toUpdate *ateapipb.Actor) error {
```
Sneha-at, line 649: How do we use this publish context information stored for a volume? I do not see a DB update on volume detach, if we plan to use the DB details to get the current publish context for the actor volume then it might not have the right details. Though, if the details are checked only for a running actor then implicitly the actor volumes are already attached and thus the DB has correct up-to-date information.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -330,6 +330,27 @@ message ExternalVolume {
...
+  // volume was attached to a node, which the node plugin needs to complete the
+  // mount (e.g. the device path for AWS EBS). It is empty for drivers that do
+  // not implement attachment, and is only meaningful for
+  // publish_context_node. Keys and values are the driver's own, so they are
+  // only bounded, not validated.
+  //
+  // +k8s:optional
+  // +k8s:maxProperties=32
+  // +k8s:eachKey=+k8s:maxLength=128
+  // +k8s:eachVal=+k8s:maxLength=256
```
Sneha-at, line 344: Is this a standard limitation? Can publish context frequently cross this limit, what happens if publish context > 256 characters?

cmd/ateapi/internal/controlapi/workflow_resume.go
anishgang: I am currently doing this. In ensuredVolumesAttached, if len(mountedVols) == 0 { return actor, nil } to immediately bypass the store update. Also, I am checking !maps.Equal(vol.GetPublishContext(), pubCtx) so we also skip the DB write if the volume is already attached to that node. Answered in #1729

pkg/proto/ateapipb/ateapi.proto
anishgang: While standard device paths are generally pretty short ( <50 chars), certain drivers can obviously pass longer connection strings. To provide more headroom, I did increase it from 256 to 1024 (which should be like 10x headroom for any driver). Answered in #1729

cmd/ateapi/internal/controlapi/workflow_resume.go
anishgang: We use PublishContext to tell the worker where the attached disk is located (i.e /dev/xvdf) so it can mount it. On detach, we now explicitly wipe it from the database (PublishContext = nil). When an actor resumes, we always attach the disk and refresh this in the DB before launching the container. Answered in #1729

hajiler: Yeah as Anish said, publish context is used during attach. Refering to Anish's PR to handle this.

pkg/proto/ateapipb/ateapi.proto
hajiler: CSI spec's only limit is that publishContext must be less than 4KiB. Substrate requires that we have limits for number of entries, key and value though. 

Right now Filestore has no publish context which is the primary driver that we are supporting. For other CSI drivers I think it's better to have a restrictive policy that we can easily loosen later if need be than having a policy that is too flexible.

cmd/ateapi/internal/controlapi/workflow_resume.go
hajiler: Done, this will likely require a rebase.

cmd/ateapi/internal/controlapi/workflow_resume.go
hajiler: PublishContext is only referred to after an actor resume (when mounting a volume). So whenever a volume is mounted we can assume that first, the the publish context was updated for the volume. 

I think clearing this is an unnecessary write.

Sneha-at: Do we ever refer to the publish context to get current details for the volume, this is only relevant if we check publish context even once the volume is unmounted.

pkg/proto/ateapipb/ateapi.proto
Sneha-at: Do we have any publish context limit set for PDCSI driver? I agree with tighten now and loosen later but 256 characters seems too short, 1024 might be better to start off.

cmd/ateapi/internal/controlapi/workflow_resume.go
hajiler: Hmm from a CSI perspective no, publish context isn't used at between when the volume is detached and when the volume gets attached again (and the publish context gets updated). 

I don't see a scenario where they would want to, but I guess a user could take a look at the actor object and see the publishcontext.

pkg/proto/ateapipb/ateapi.proto
hajiler: Done.

PR #1580, opened 2026-09-09
demos/sandbox: use ateclient.NewClient for authenticated ate-api access

### LINKED ISSUE #1579: [Bug]: demos/sandbox client fails with missing bearer token on actor resume
## Summary
Running `demos/sandbox/client` fails when invoking `ResumeActor` against `ate-api` because the client establishes an unauthenticated gRPC connection without attaching a Bearer token.

## Steps to Reproduce
1. Port-forward the Substrate API server (`kubectl port-forward -n ate-system svc/api 8080:443`).
2. Run the sandbox client:
   ```bash
   go run ./demos/sandbox/client --atespace=demo --name=sb1
   ```

## Actual Behavior
```text
2026/09/09 21:51:13 Connecting to ateapi at localhost:8080...
2026/09/09 21:51:13 Resuming actor sb1...
2026/09/09 21:51:13 Failed to resume actor: rpc error: code = Unauthenticated desc = missing bearer token
```

## Root Cause
`demos/sandbox/client/main.go` uses a custom `dialAteAPI()` helper that calls `grpc.NewClient` with TLS but omits `PerRPCCredentials` (Bearer JWT / ServiceAccount token). As a result, `ate-api`'s auth interceptor rejects the unauthenticated request.

### Expected Behavior

The client should authenticate with `ate-api` using the standard `internal/ateclient.NewClient`, which automatically:
- Obtains an `ate-client` ServiceAccount bearer token via the k8s TokenRequest API (or uses `--token-file`).
- Validates the serving certificate against `ClusterTrustBundle`.
- Injects bearer credentials into RPC requests.

### Steps to Reproduce

1. Deploy the sandbox demo on an Agent Substrate cluster:
`       ./hack/install-ate.sh --deploy-demo-sandbox`

2. Create a sandbox actor:
`       kubectl ate create actor sb1 -a ate-demo-sandbox --template sandbox-template`

3. Port-forward the API and router:
```       kubectl port-forward -n ate-system svc/api 8080:443 &
       kubectl port-forward -n ate-system svc/atenet-router 8000:80 &
```

4. Run the sandbox client:
`       go run ./demos/sandbox/client --atespace=ate-demo-sandbox --name=sb1`

5. Error:
`       Failed to resume actor: rpc error: code = Unauthenticated desc = missing bearer token`

### Sandbox Runtime

gVisor (runsc)

0b3d2d078

_No response_

### Host OS & Architecture

_No response_

```shell

```

### Additional Context

_No response_

### Confirmation

ajaysundark:
Fixes unauthenticated gRPC calls by replacing the custom dialAteAPI helper with internal/ateclient.NewClient, which automatically manages bearer tokens and ClusterTrustBundle TLS verification.

Fixes #1579

- [X] Tests pass

### DIFF
--- a/demos/sandbox/client/main.go
+++ b/demos/sandbox/client/main.go
@@ -17,7 +17,6 @@ import (
 	"bufio"
 	"bytes"
 	"context"
-	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -28,12 +27,11 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/agent-substrate/substrate/internal/ateclient"
 	"github.com/agent-substrate/substrate/internal/atenet"
 	"github.com/agent-substrate/substrate/internal/resources"
 	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
 	"github.com/spf13/pflag"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
 )
 
 type ProcessRequest struct {
@@ -49,22 +47,14 @@ type ProcessResponse struct {
 	Error  string `json:"error,omitempty"`
 }
 
-func dialAteAPI(endpoint string) (ateapipb.ControlClient, *grpc.ClientConn, error) {
-	creds := credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})
-
-	conn, err := grpc.NewClient(endpoint, grpc.WithTransportCredentials(creds))
-	if err != nil {
-		return nil, nil, err
-	}
-
-	return ateapipb.NewControlClient(conn), conn, nil
-}
-
 func main() {
 	actorName := pflag.String("name", "", "Name of the sandbox actor (required)")
 	atespace := pflag.String("atespace", "", "Atespace the actor lives in (required)")
-	ateapiAddr := pflag.String("ateapi", "localhost:8080", "Address of the ateapi gRPC server")
+	ateapiAddr := pflag.String("ateapi", "", "Address of the ateapi gRPC server (e.g. localhost:8080). If omitted, automatically port-forwards.")
 	atenetAddr := pflag.String("atenet", "localhost:8000", "Address of the atenet HTTP router")
+	tokenFile := pflag.String("token-file", "", "Path to a bearer token for ate-api authentication, or - to read it from stdin. Defaults to a Kubernetes ServiceAccount token.")
+	kubeconfig := pflag.String("kubeconfig", "", "Path to the kubeconfig file")
+	k8sContext := pflag.String("context", "", "The name of the kubeconfig context to use")
 	pflag.Parse()
 
 	if *actorName == "" {
@@ -89,11 +79,11 @@ func main() {
 
 	// 1. Connect to ateapi and Resume Actor
 	log.Printf("Connecting to ateapi at %s...", *ateapiAddr)
-	cli, conn, err := dialAteAPI(*ateapiAddr)
+	cli, err := ateclient.NewClient(ctx, *kubeconfig, *k8sContext, *ateapiAddr, *tokenFile, false)
 	if err != nil {
-		log.Fatalf("Failed to dial ateapi: %v", err)
+		log.Fatalf("Failed to connect to ateapi: %v", err)
 	}
-	defer conn.Close()
+	defer cli.Close()
 
 	log.Printf("Resuming actor %s...", actorRef.Name)
 	_, err = cli.ResumeActor(ctx, &ateapipb.ResumeActorRequest{Actor: actorRef.ToObjectRef()})


ajaysundark:
cc @aditya-shantanu @zlammerts-svg

demos/sandbox/client/main.go
```diff
@@ -49,22 +47,14 @@ type ProcessResponse struct {
...
-	return ateapipb.NewControlClient(conn), conn, nil
-}
-
 func main() {
 	actorName := pflag.String("name", "", "Name of the sandbox actor (required)")
 	atespace := pflag.String("atespace", "", "Atespace the actor lives in (required)")
-	ateapiAddr := pflag.String("ateapi", "localhost:8080", "Address of the ateapi gRPC server")
+	ateapiAddr := pflag.String("ateapi", "", "Address of the ateapi gRPC server (e.g. localhost:8080). If omitted, automatically port-forwards.")
 	atenetAddr := pflag.String("atenet", "localhost:8000", "Address of the atenet HTTP router")
+	tokenFile := pflag.String("token-file", "", "Path to a bearer token for ate-api authentication, or - to read it from stdin. Defaults to a Kubernetes ServiceAccount token.")
```
aditya-shantanu, line 55: `--token-file=-` is incompatible with this program: ateclient reads the stdin token via `io.ReadAll(os.Stdin)` (`internal/ateclient/builder.go`, `readBearerToken`), which consumes stdin until EOF — but this client is an interactive stdin REPL (`bufio.NewScanner(os.Stdin)` below). Interactively the client hangs at connect until Ctrl-D, and either way the REPL scanner then gets immediate EOF, so the session exits right after resume + deferred suspend. Piped usage (`echo ls | sandbox-client --token-file=-`) sends the command text as the bearer token. Suggest rejecting `-` here (e.g. `log.Fatal("--token-file=- is not supported by the REPL client; pass a file path")`) or reading the token before entering the REPL from a different fd.

demos/sandbox/client/main.go
```diff
@@ -89,11 +79,11 @@ func main() {
 
 	// 1. Connect to ateapi and Resume Actor
 	log.Printf("Connecting to ateapi at %s...", *ateapiAddr)
-	cli, conn, err := dialAteAPI(*ateapiAddr)
+	cli, err := ateclient.NewClient(ctx, *kubeconfig, *k8sContext, *ateapiAddr, *tokenFile, false)
```
aditya-shantanu, line 82: Long sessions leak the actor: with `--token-file` empty, `ateclient` mints a single ServiceAccount token with `ExpirationSeconds: 3600` and wraps it in `bearerTokenCreds` — a static string with no refresh (`internal/ateclient/builder.go`). The REPL's only later control-plane RPC is the deferred `SuspendActor` at exit, so a sandbox session kept open for more than an hour sends an expired token on `exit`, gets Unauthenticated, and the actor is silently left running (only a `log.Printf`). Unlike `fileBearerTokenCreds`, which re-reads the file per RPC, the minted token can never be refreshed. Consider minting the token lazily per RPC (or re-minting on expiry) in ateclient, or at least surfacing this limit — an interactive sandbox REPL is exactly the long-lived client that hits it.

aditya-shantanu, line 82: Capability regression for tunnel-only users: even with an explicit `--ateapi` endpoint (and even with `--token-file` supplied), `dialDirect` unconditionally loads a kubeconfig, builds a clientset, and lists `certificates.k8s.io/v1beta1` ClusterTrustBundles for the TLS roots; with no `--token-file` it additionally needs TokenRequest RBAC on the `ate-client` SA in `ate-system`. The old `dialAteAPI` needed no cluster access at all, so anyone who ran the demo through a shared tunnel/bastion or in CI without a kubeconfig now fails at `failed to load kubeconfig` (or at the CTB list) before ever dialing. That may be an acceptable trade for authenticated TLS, but it changes the meaning of `--ateapi` from "just dial this address" to "dial this address, but still require cluster RBAC" — worth a note in the flag help and README, or an escape hatch (e.g. `--ca-file`) in ateclient.

demos/sandbox/client/main.go
```diff
@@ -49,22 +47,14 @@ type ProcessResponse struct {
...
-	}
-
-	return ateapipb.NewControlClient(conn), conn, nil
-}
-
 func main() {
 	actorName := pflag.String("name", "", "Name of the sandbox actor (required)")
 	atespace := pflag.String("atespace", "", "Atespace the actor lives in (required)")
-	ateapiAddr := pflag.String("ateapi", "localhost:8080", "Address of the ateapi gRPC server")
+	ateapiAddr := pflag.String("ateapi", "", "Address of the ateapi gRPC server (e.g. localhost:8080). If omitted, automatically port-forwards.")
```
aditya-shantanu, line 53: Silent behavior change for existing invocations: anyone following the current README (manual `kubectl port-forward svc/api 8080:443` in Terminal 1) who omits `--ateapi` used to hit the manual tunnel via the old `localhost:8080` default. Now the empty default ignores that tunnel and opens its own port-forward using the kubeconfig's *current context* — which can point at a different cluster than the manual tunnel — and resumes/suspends that cluster's actor with no indication (the only log prints an empty address). Worth calling out in the README, and logging which mode/cluster is used (see the next comment on line 81).

aditya-shantanu, line 53: `demos/sandbox/README.md` was not updated even though the PR checklist marks documentation as included: it still mandates "Terminal 1: API Server — `kubectl port-forward -n ate-system svc/api 8080:443`" and shows `--ateapi=localhost:8080` as the canonical invocation, while the auto-port-forward default and the new `--token-file`/`--kubeconfig`/`--context` flags appear nowhere. The manual API forward is now redundant, and the README also doesn't mention that even the explicit-endpoint path now needs kubeconfig + ClusterTrustBundle/TokenRequest access. Please update section 3 and the usage example to the new workflow.

demos/sandbox/client/main.go
```diff
@@ -89,11 +79,11 @@ func main() {
 
 	// 1. Connect to ateapi and Resume Actor
 	log.Printf("Connecting to ateapi at %s...", *ateapiAddr)
```
aditya-shantanu, line 81: In the new default mode (`--ateapi=""`) this prints `Connecting to ateapi at ...` with an empty address, and it misdescribes what happens (an automatic port-forward to `svc/api`). The multiplex demo handled the same migration explicitly (`demos/claude-code-multiplex/ui/server.go`: `"[ui] connecting to ateapi (endpoint=%q; empty = auto port-forward to svc/api)"`).

```suggestion
	if *ateapiAddr == "" {
		log.Print("Connecting to ateapi (auto port-forward to svc/api)...")
	} else {
		log.Printf("Connecting to ateapi at %s...", *ateapiAddr)
	}
```

demos/sandbox/client/main.go
aditya-shantanu, line 82: Minor, pre-existing hazard made the default: the auto port-forward pins the tunnel to a single api pod and "does not fail over if that pod later dies" (`internal/portforward/portforward.go`). For a long-lived REPL, if that pod is rescheduled mid-session the tunnel dies silently (commands keep flowing via atenet), and the deferred `SuspendActor` on `exit` fails, leaving the actor unsuspended. The manual `kubectl port-forward` workflow had the same weakness but at least failed visibly in its own terminal. Not blocking — just noting the failure mode this demo now inherits by default.

demos/sandbox/client/main.go
```diff
@@ -49,22 +47,14 @@ type ProcessResponse struct {
...
-}
-
 func main() {
 	actorName := pflag.String("name", "", "Name of the sandbox actor (required)")
 	atespace := pflag.String("atespace", "", "Atespace the actor lives in (required)")
-	ateapiAddr := pflag.String("ateapi", "localhost:8080", "Address of the ateapi gRPC server")
+	ateapiAddr := pflag.String("ateapi", "", "Address of the ateapi gRPC server (e.g. localhost:8080). If omitted, automatically port-forwards.")
 	atenetAddr := pflag.String("atenet", "localhost:8000", "Address of the atenet HTTP router")
+	tokenFile := pflag.String("token-file", "", "Path to a bearer token for ate-api authentication, or - to read it from stdin. Defaults to a Kubernetes ServiceAccount token.")
+	kubeconfig := pflag.String("kubeconfig", "", "Path to the kubeconfig file")
```
aditya-shantanu, line 56: Cleanup/altitude: this is now the third copy of the ateclient connection flag set — `--kubeconfig`/`--context`/`--token-file` help strings are verbatim from `cmd/kubectl-ate/internal/cmd/root.go:57-60`, the endpoint flag is `--ateapi` here vs `--endpoint` in kubectl-ate for the same `NewClient` parameter, and `demos/claude-code-multiplex/ui/server.go` has a third ad-hoc variant (env var). A small shared registration helper in `internal/ateclient` (or at least reusing kubectl-ate's flag names) would keep the consumers from drifting as the client contract evolves.

PR #1581, opened 2026-09-10
enhance tests to validate volume lifecycle for actor state changes

Sneha-at:
NA

This PR enhances the tests to validate volume lifecycle during state transition failures

 • **TestDeleteActor_VolumeDeletionFailure_RetrySuccess**- Verifies that when volume deletion fails during DeleteActor, the actor transitions to ACTOR_STATE_DELETING and its volumes to STATUS_DELETING. A     
  subsequent retry successfully finalizes volume deletion and deletes the actor from the store.                                                                                                               
  • **TestResumeActor_VolumeAttachFailureAndRetry** - Verifies that when volume attachment fails during ResumeActor, the actor remains in ACTOR_STATE_RESUMING with worker assignment and provisioned volumes     
  intact. Retrying ResumeActor re-attempts volume attachment and successfully transitions the actor to ACTOR_STATE_RUNNING.                                                                                   
  • **TestResumeActor_VolumeAttachFailure_DeleteActor** - Verifies that an actor stuck in ACTOR_STATE_RESUMING after an attach failure rejects standard DeleteActor requests, but calling DeleteActor with        
  AnyState=true successfully cleans up worker assignments and provisioned volumes.                                                                                                                            
  • **TestResumeActor_MultiVolumePartialAttachFailure_Retry** - Verifies that when attaching multiple volumes during ResumeActor and one volume fails, previously attached volumes remain attached and the actor  
  stays in ACTOR_STATE_RESUMING. A subsequent retry attaches only the remaining unattached volumes and transitions the actor to ACTOR_STATE_RUNNING.                                                          
  • **TestSuspendActor_VolumeDetachFailure_RetrySuccess** - Verifies that volume detach failures during SuspendActor leave the actor in ACTOR_STATE_SUSPENDING with its worker assignment held. A subsequent retry
  succeeds in detaching the volume, releasing the worker, and transitioning the actor to ACTOR_STATE_SUSPENDED.                                                                                               
  • **TestSuspendActor_VolumeDetachFailure_DeleteActorAnyState** - Verifies that an actor stuck in ACTOR_STATE_SUSPENDING due to a volume detach failure rejects standard deletion, but cleanly detaches volumes  
  and deletes the actor when called with AnyState=true.                                                                                                                                                       
  • **TestPauseActor_VolumeLifecycle_DetachAndResumeAttach** - Validates the volume lifecycle across pause and unpause, confirming external volumes are detached when transitioning from RUNNING to PAUSED and re-
  attached when resumed back to RUNNING.                                                                                                                                                                      
  • **TestPauseActor_VolumeDetachFailure_RetrySuccess** - Verifies that volume detachment failure during PauseActor leaves the actor in ACTOR_STATE_PAUSING. A retry of PauseActor re-attempts volume detachment, 
  succeeds, and advances the actor state to ACTOR_STATE_PAUSED.                                                                                                                                               
  • **TestResumeActor_PausedLocalSnapshotMissing_Crashes** - Verifies that if local snapshot files on the worker are missing when resuming a PAUSED actor, the control plane marks the actor ACTOR_STATE_CRASHED  
  and frees the assigned worker pod.                                                                                                                                                                          
  • **TestDetachActorVolumes** - Tests control plane volume detachment logic across multiple scenarios, including container mount filtering, partial detach errors, unassigned workers, and idempotent not-found  
  responses from plugins.                                                                                                                                                                                     
  • **TestCreateActorVolumes (Storage class parameters case)** - Verifies that parameters configured on a StorageClass (such as disk and filesystem types) are correctly propagated into the volume context of    
  newly created external volumes.                                                                                                                                                                             
  • **TestEnsureExternalSnapshotsReleased_DeletePrefixFailure** - Verifies that transient object store errors during actor deletion preserve un-deleted snapshot objects and return an error, which cleanly       
  completes deletion on a retry.                                                                                                                                                                              
  • **TestMountExternalVolumes** - Tests worker-side volume mounting, verifying host mount directory creation, handling pre-existing paths, plugin errors, and ensuring that partial failures during multi-volume 
  mounts do not unmount already mounted volumes.                                                                                                                                                              
  • **TestVolumeHostDirectoryCleanup** - Verifies worker-side host directory management, ensuring volume unmounting preserves mount directories while directory reset removes empty directories and prevents data 
  loss on non-empty ones.                                                                                                                                                                                     
  • **TestExternalVolume_NodeMigration** - Verifies cross-node actor migration with external volumes by suspending an actor, evicting its worker pod, and resuming it on a new worker node while ensuring         
  application state is preserved

- [ x] Tests pass

### DIFF
--- a/cmd/atelet/volumes.go
+++ b/cmd/atelet/volumes.go
@@ -29,13 +29,17 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// volumeHostPath resolves an actor UID and volume name to its mount path on the host.
+// A variable so tests can direct mounts into a temporary directory.
+var volumeHostPath = ateompath.VolumeHostPath
+
 func (s *AteomHerder) mountExternalVolumes(ctx context.Context, actorUID string, volumes []*ateletpb.Volume) error {
 	for _, vol := range volumes {
 		ext := vol.GetExternal()
 		if ext == nil {
 			continue
 		}
-		hostPath := ateompath.VolumeHostPath(actorUID, vol.GetName())
+		hostPath := volumeHostPath(actorUID, vol.GetName())
 		if err := os.MkdirAll(hostPath, 0o750); err != nil {
 			return fmt.Errorf("failed to create mount point %q: %w", hostPath, err)
 		}
@@ -58,7 +62,7 @@ func (s *AteomHerder) unmountExternalVolumes(ctx context.Context, actorUID strin
 		if ext == nil {
 			continue
 		}
-		hostPath := ateompath.VolumeHostPath(actorUID, vol.GetName())
+		hostPath := volumeHostPath(actorUID, vol.GetName())
 		slog.InfoContext(ctx, "Unmounting volume", slog.String("volume_id", ext.GetStorageVolumeId()), slog.String("host_path", hostPath), slog.String("volume_type", ext.GetVolumeType()))
 		// TODO: Standardize volume plugin lookup and error handling across control plane
 		// and worker plane (e.g. via a shared helper).
--- a/internal/e2e/suites/demo/demo_test.go
+++ b/internal/e2e/suites/demo/demo_test.go
@@ -488,6 +488,92 @@ func TestDeleteActorAnyStateWithExternalVolume(t *testing.T) {
 	}
 }
 
+func TestExternalVolume_NodeMigration(t *testing.T) {
+	if e2e.IsMicroVM() {
+		t.Skip("Skipping TestExternalVolume_NodeMigration for microVM environment")
+	}
+
+	ctx := context.Background()
+	clients := e2e.GetClients()
+	nsObj := e2e.CreateNamespace(t)
+
+	at, err := createActorTemplateWithExternalVolume(ctx, t, clients, nsObj, ateapipb.SnapshotContentScope_SNAPSHOT_CONTENT_SCOPE_DATA, ateapipb.SnapshotContentScope_SNAPSHOT_CONTENT_SCOPE_DATA, ateapipb.ResumeSource_RESUME_SOURCE_COLD_BOOT)
+	if err != nil {
+		t.Fatalf("failed to initialize ActorTemplate: %v", err)
+	}
+
+	actorName := "extvol-migration-" + nsObj.Name
+
+	t.Logf("Creating Actor %q...", actorName)
+	if _, err := clients.SubstrateAPI.CreateActor(ctx, &ateapipb.CreateActorRequest{Actor: &ateapipb.Actor{
+		Metadata:      &ateapipb.ResourceMetadata{Atespace: demoAtespace, Name: actorName},
+		ActorTemplate: e2e.TemplateRef(at),
+	}}); err != nil {
+		t.Fatalf("failed to create Actor: %v", err)
+	}
+	waitForActorState(ctx, t, clients, actorName, ateapipb.ActorState_ACTOR_STATE_SUSPENDED)
+
+	t.Logf("Resuming Actor %q...", actorName)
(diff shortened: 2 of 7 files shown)


cmd/atelet/volumes.go
```diff
@@ -29,13 +29,17 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// volumeHostPath resolves an actor UID and volume name to its mount path on the host.
+// A variable so tests can direct mounts into a temporary directory.
+var volumeHostPath = ateompath.VolumeHostPath
```
hajiler, line 34: Instead of making this a global variable, would it be better add it to ateomherder?

internal/e2e/suites/demo/demo_test.go
```diff
@@ -489,6 +488,92 @@ func TestDeleteActorAnyStateWithExternalVolume(t *testing.T) {
...
+		Actor: &ateapipb.ObjectRef{Atespace: demoAtespace, Name: actorName},
+	}); err != nil {
+		t.Fatalf("failed to suspend Actor: %v", err)
+	}
+	waitForActorState(ctx, t, clients, actorName, ateapipb.ActorState_ACTOR_STATE_SUSPENDED)
+
+	// Delete initial worker pod to force migration onto another worker upon resume
+	if initialWorkerPod != "" && initialWorkerNS != "" {
+		t.Logf("Evicting initial worker pod %s/%s to force migration...", initialWorkerNS, initialWorkerPod)
+		_ = clients.K8s.CoreV1().Pods(initialWorkerNS).Delete(ctx, initialWorkerPod, metav1.DeleteOptions{})
```
hajiler, line 555: Should we add some wait to verify that the pod is deleted?

cmd/atelet/volumes.go
```diff
@@ -29,13 +29,20 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+func (s *AteomHerder) volumeHostPath(actorUID, volumeName string) string {
+	if s.volumeHostPathFn != nil {
```
hajiler: cmd/atelet/main.go:447 adds a volumeHostPathFn seam to AteomHerder purely for tests, with the default encoded in both NewService and volumes.go:32. T

The package var ateompath.ActorsDir already provides this hook, and the same PR's resetActorDirs subtests use it. Half of volumes_test.go injects the function, the other half swaps the var.

I'm not a fan of global variables, but if ateompath.ActorsDir already exists and is used should we default to that?

internal/e2e/suites/demo/demo_test.go
Sneha-at: The workload pod is no longer considered for scheduling once it is [marked for deletion](https://github.com/agent-substrate/substrate/blob/d0d85c397db77dee84f3c87f1b23cb10a714f3b5/cmd/atecontroller/internal/workersync/syncer.go#L264) but we can add a wait time to give the DrainWorker some time to update the status. But as the pod need not be deleted successfully for migration we don't have to check the pod status pos  wait

Sneha-at: Added a 5 second wait to allow to be marked for deletion

cmd/atelet/volumes.go
Sneha-at: Agreed — dropped the seam. `volumeHostPathFn` is gone from `AteomHerder` (and from `NewService` in main.go), and `volumes.go` calls `ateompath.VolumeHostPath` directly. All of `volumes_test.go` now goes through a single `withTempActorsDir` helper that swaps `ateompath.ActorsDir`, so the volume subtests and the `resetActorDirs` subtests use the same mechanism and every path derived from `ActorsDir` (including the ones the OCI spec builder computes) moves together. Done in c31cc71e.

msau42 (dismissed):

msau42 (approved):

PR #1583, opened 2026-09-10
ateapi: apply a replica's own worker writes to its cache at commit

### LINKED ISSUE #1370: Make store publish worker-watch updates locally
When a `Worker` resource is created / updated / deleted we publish a `WorkerEvent` which has the operation type and the serialized resource. All replicas, including the one that performed the update consume the events and reply them into the worker cache the same way, from the DB. There is a potential performance optimization by making the replica that performed the update publish the notification locally (and faster).

The event from the DB will arrive later and be ignored (because the version is not newer). This way the staleness of the workercache on each replica gets reduced.

shrutiyam-glitch:
Fixes #1370

**Summary**
`atepg` keeps a registry of the live `WatchWorkers` channels and publishes each committed worker event onto them at commit, about one poll interval ahead of the outbox row carrying the same event. Replicas see their own writes sooner without a second notification API.

**Why**
Replicas previously learned of their own writes only from the outbox watch, ~50ms later. In that window the scheduler re-picks a worker it just assigned and retries on the precondition check, and capacity a replica just freed stays invisible to its own placement. Cross-replica propagation is unchanged.

**Key changes**
- `WatchWorkers` enrolls its channel; the poller deregisters before closing it, which is what makes a send on a closed channel impossible.
- `writeAndAppendEvent` publishes only after `tx.Commit` returns.
- The commit-time copy is decoded from the same payload the outbox stores, so both deliveries are identical, including unknown-field pruning.
- Sends are non-blocking and never advance the poll cursor, so a full buffer costs a watcher nothing and gap-free delivery is unchanged.
- Consumers now also see their own writes out of xid order and again on the poll. `workercache` already fences creates and updates on version, so duplicates and stale events are no-ops.

**Testing**
`TestLocalPublishReachesWatchers`, `TestLocalPublishSurvivesWatchClose`, and `TestCache_WatchEventsAreFenced` are new.
`TestWatchWorkers_DeliveryFencedByOldestTransaction` now consumes the commit-time copy first and asserts the xmin fence on the outbox copy; the fence governs the poll, and a local publish bypasses it by construction.

Full `atepg` and `workercache` suites pass under `-race`.

- [X] Tests pass

### DIFF
--- a/cmd/ateapi/internal/store/atepg/atepg.go
+++ b/cmd/ateapi/internal/store/atepg/atepg.go
@@ -58,6 +58,15 @@ type Persistence struct {
 	pollFailureCloseAfter time.Duration
 	stopMaintenance       context.CancelFunc
 	maintenanceDone       chan struct{}
+	localEvents           func(store.WorkerEvent)
+}
+
+// PublishEventsLocally registers a fast-path sink for committed create/update
+// events ahead of outbox delivery. Later outbox duplicates are deduplicated
+// by version. Deletes are excluded to prevent resurrection races.
+// sink runs on the write path, so it must be fast and non-blocking.
+func (p *Persistence) PublishEventsLocally(sink func(store.WorkerEvent)) {
+	p.localEvents = sink
 }
 
 var _ store.Interface = (*Persistence)(nil)
--- a/cmd/ateapi/internal/store/atepg/atepg.go
+++ b/cmd/ateapi/internal/store/atepg/atepg.go
@@ -61,10 +61,9 @@ type Persistence struct {
 	localEvents           func(store.WorkerEvent)
 }
 
-// PublishEventsLocally registers a fast-path sink for committed create/update
-// events ahead of outbox delivery. Later outbox duplicates are deduplicated
-// by version. Deletes are excluded to prevent resurrection races.
-// sink runs on the write path, so it must be fast and non-blocking.
+// PublishEventsLocally registers a sink for committed create/update/delete
+// worker events ahead of outbox delivery. Later outbox duplicates are deduplicated
+// by version.
 func (p *Persistence) PublishEventsLocally(sink func(store.WorkerEvent)) {
 	p.localEvents = sink
 }
--- a/cmd/ateapi/internal/store/atepg/outbox.go
+++ b/cmd/ateapi/internal/store/atepg/outbox.go
@@ -101,6 +101,11 @@ func (p *Persistence) writeAndAppendEvent(ctx context.Context, eventType store.W
 	if err := tx.Commit(ctx); err != nil {
 		return nil, fmt.Errorf("committing transaction: %w", err)
 	}
+	if worker != nil && p.localEvents != nil && eventType != store.WorkerEventDeleted {
+		// Cloned because the returned worker also goes back to the API
+		// caller, while the sink (the worker cache) retains its copy.
+		p.localEvents(store.WorkerEvent{Type: eventType, Worker: proto.Clone(worker).(*ateapipb.Worker)})
+	}
 	return worker, nil
 }
 
--- a/cmd/ateapi/internal/store/atepg/outbox.go
+++ b/cmd/ateapi/internal/store/atepg/outbox.go
@@ -101,7 +101,7 @@ func (p *Persistence) writeAndAppendEvent(ctx context.Context, eventType store.W
 	if err := tx.Commit(ctx); err != nil {
 		return nil, fmt.Errorf("committing transaction: %w", err)
 	}
-	if worker != nil && p.localEvents != nil && eventType != store.WorkerEventDeleted {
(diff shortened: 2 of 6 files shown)


cmd/ateapi/internal/workercache/workercache.go
```diff
@@ -210,6 +210,16 @@ func (c *Cache) resync(ctx context.Context) *store.WorkerWatch {
 	return watch
 }
 
+// ApplyLocal eagerly updates the cache on commit, ahead of the watch.
+// Version fencing deduplicates later watch events. Deletes are ignored
+// to prevent out-of-order watch replays from resurrecting workers.
+func (c *Cache) ApplyLocal(event store.WorkerEvent) {
+	if event.Type == store.WorkerEventDeleted {
```
juli4n: I don't think we should ignore worker deletion? The worker name is the Pod UID, a worker resurrecting is not something that can realistically happen.

cmd/ateapi/main.go
```diff
@@ -156,6 +156,13 @@ func main() {
 	if err := workerCache.Start(ctx); err != nil {
 		serverboot.Fatal(ctx, "Failed to seed worker cache", err)
 	}
+	// Apply local writes directly to the cache on commit to eliminate outbox
+	// polling delay; later watch events are deduplicated by version.
+	if publisher, ok := persistence.(interface {
```
juli4n: Should we do this registration in `WorkerCache.Start`?

cmd/ateapi/internal/store/atepg/outbox.go
```diff
@@ -101,6 +101,11 @@ func (p *Persistence) writeAndAppendEvent(ctx context.Context, eventType store.W
 	if err := tx.Commit(ctx); err != nil {
 		return nil, fmt.Errorf("committing transaction: %w", err)
 	}
+	if worker != nil && p.localEvents != nil && eventType != store.WorkerEventDeleted {
```
juli4n: I think we should always publish events, including delete.

cmd/ateapi/internal/workercache/workercache.go
shrutiyam-glitch: got it, this makes sense. 
even if the scheduler picks a non-existent worker, the bind would eventually fail.

cmd/ateapi/internal/workercache/workercache.go
```diff
@@ -69,11 +69,19 @@ type workerListWatcher interface {
...
 // Returns as soon as the initial sync succeeds.
+//
+// Stores that publish their committed writes in-process
+// are wired to the cache here.
 func (c *Cache) Start(ctx context.Context) error {
 	watch, err := c.sync(ctx)
 	if err != nil {
 		return err
 	}
+	if publisher, ok := c.store.(interface {
```
juli4n, line 80: I feel this can be brittle. If the type assertion fails (e.g. the store gets wrapped into some sort of delegate pattern), the type assertion will silently fail and the performance optimization just gets disabled. I could be hard to troubleshoot.

Have you considered a model where we don't need the downstream consumers of the store API to register? The store itself could keep a registry of the watches it has registered, and publish `WorkerEvent` on them. This would need zero changes outside the `store/atepg` package.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -58,6 +58,14 @@ type Persistence struct {
...
 	stopMaintenance       context.CancelFunc
 	maintenanceDone       chan struct{}
+	localEvents           func(store.WorkerEvent)
+}
+
+// PublishEventsLocally registers a sink for committed create/update/delete
+// worker events ahead of outbox delivery. Later outbox duplicates are deduplicated
+// by version.
+func (p *Persistence) PublishEventsLocally(sink func(store.WorkerEvent)) {
+	p.localEvents = sink
```
juli4n, line 68: This can be brittle, if you have multiple callers registering sinks, only the last one will get the events. See my other comment about whether we can come up with a model where clients of the store don't need to register themselves.

shrutiyam-glitch: Originally I had registered the sink directly so the cache updated synchronously — `ApplyEvent` called inline at commit — since `workercache` is the only watcher.

Switched to the registry you suggested: `Persistence` keeps its own set of live `WatchWorkers` channels and publishes committed events onto them, so nothing downstream registers and there's no capability to probe for. `workercache` reads them off the **same channel** the outbox poller feeds. This trades away the synchronous guarantee. 
Locally published events also arrive ahead of their xid position and again when the poll reaches them, so the channel now carries duplicates. Gap-free delivery is unaffected — the cursor only advances on the poller's send — and `workercache` version-fences creates and updates, so both are no-ops there.

juli4n (approved):

BenTheElder:
Thanks!

Issue #1458, opened 2026-09-03
ate-api-server: add /statusz (feature parity with atenet router, KEP-4827 convention)


git286:
## Summary

Feature request: give `ate-api-server` a `/statusz` endpoint following the pattern the atenet router already ships (`--status-port`, default 4040, see `cmd/atenet/internal/router/status.go`). Today the two control-plane components are inconsistent: the router serves a rich HTML status dashboard, while ateapi has nothing on 4040 (connection refused) and its only HTTP port (9090) serves just `/metrics` and `/healthz`.

This is also the direction upstream Kubernetes is taking for its own control-plane components via [KEP-4827 component statusz](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/4827-component-statusz) (with `/flagz` in KEP-4828), so following it here means adopting an emerging convention rather than inventing one.

## Motivation — three real debugging sessions that needed it

During a bug bash on a live GKE install (substrate rev 0d85e8572161), every ateapi-side incident required log spelunking that a status page would have shortcut:

1. **Wedged actors after template deletion.** Deleting an ActorTemplate with live actors left them unable to suspend ("actor template not found") or delete (`FailedPrecondition`). ateapi's own metrics counted these failures (`ate_actor_lifecycle_operation_duration_seconds{error_type="FailedPrecondition"}`), but there was no place to *see* recent failing RPCs. The router's "Recent 100 Processing Inquiries" table is exactly the right pattern; ateapi needs the control-plane equivalent.
2. **Silent golden-snapshot failure.** A template with an unpullable image (MANIFEST_UNKNOWN) retried `AteomHerder/Run` every ~20s for 45 minutes while `kubectl ate get actor-template` showed an empty ERROR column and `status: {}`. A statusz section for in-flight template/golden work would have surfaced the retry loop and its last error immediately.
3. **Postgres invisibility.** DB behavior is currently invisible in traces (#1455) *and* there is no page showing pool utilization, connectivity, or outbox depth — the two gaps compound.

The router page also proved its worth in the same bash: an operator flag change (`--route-timeout=180s`) was immediately visible under "Global Arguments & Configuration".

## Suggested content

- Build revision, uptime, resolved flags (KEP-4827/4828 style)
- PostgreSQL: pool stats (in-use/idle/max), connectivity health check with ok/err counters + last-known-good, outbox depth/lag
- Worker cache: known workers per pool, eligible-worker counts
[...]

PR #1596, opened 2026-09-10
Add an internal ateapi statusz page

swiftdiaries:
### Description

- Added a `statusz` page for the `ateapi`
- Used the existing ate-router page as reference

Addresses #1458 

### Review Notes

Things I'm not sure about:
- I'm not aggregating metrics across replicas, is this needed?
- On the metrics themselves, could I please get a review on what's needed/how they're represented?

### Screenshot

Screenshot 2026-09-14 at 17 00 42

--- 

### AI-generated PR description

Adds an internal ateapi `/statusz` page for inspecting the serving replica. HTML and JSON expose build/version, uptime, live readiness, listener/drain settings, and safely redacted startup configuration, alongside known workers grouped by namespace/pool, PostgreSQL client-pool occupancy, and recent failed authenticated unary Control RPCs.

The dedicated `--status-port` listener defaults to 4040, can be disabled with a nonpositive value, and stays available through gRPC drain. It is exposed through the internal API Service. Configuration uses a closed display policy to redact credentials, environment-resolved values, and sensitive paths.

Each request reads diagnostics directly from existing in-memory readers; there is no diagnostic response cache or background sampler. The browser offers Manual / 5s / 10s default / 30s full-page refresh, matching the router. The selection resets to 10 seconds on each page load.

Worker counts reflect this replica's cached view of shared inventory and must not be summed across replicas; the page displays at most 100 groups with totals and truncation information. Pool statistics perform no SQL, connection acquisition, or connectivity probe. Failure history retains the newest 100 matching completions with bounded method/principal text, status, and elapsed time, without payloads, credentials, or raw errors. It resets on restart and is not an error-rate metric.

Includes both the endpoint foundation and runtime diagnostics.

Validation: focused ateapi/statusz race tests passed on the combined head `a1b96d9f302766f4f744693e06bd442a2c57da3f`; diff checks passed. Earlier validation covered PostgreSQL-backed pool tests, browser refresh behavior, GolangCI-Lint, kube-api-linter, and applicable static checks. The full `make verify` run encountered inherited color/telemetry environment failures; those affected packages passed when rerun without the conflicting variables. Some static wrappers required a clean worktree and were not run during the uncommitted refactor. No single full-verification pass, live-cluster deployment, or E2E validation is claimed.


swiftdiaries:
cc @git286

bowei:
+1 I love statusz pages :-)

swiftdiaries:
@bowei @haiyanmeng added a `statusz` page based on the issue description. Generated a screenshot with some synthetic seed data to visualize what it'd look like.

Could I please a get review on the representation of the metrics and if replica-wide aggregation is needed here? Tried to keep it simple for the first pass.

cmd/ateapi/status_rpc.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		elapsed := completedAt.Sub(startedAt)
+		if elapsed < 0 {
+			elapsed = 0
+		}
+		r.add(completedAt, statusz.RPCFailure{
+			CompletedAt:   completedAt.UTC().Format(time.RFC3339Nano),
+			Method:        boundedUTF8(info.FullMethod),
+			PrincipalKind: boundedUTF8(p.Kind),
+			PrincipalID:   boundedUTF8(p.ID),
```
git286: I think this will put the raw caller identity (JWT subjects and SPIFFE URIs) on an unauthenticated page, which is a leak of sensitive data ("recent failed RPCs" section).

git286 (changes requested):

cmd/ateapi/status_rpc.go
swiftdiaries: thanks, my bad - totally missed that! fixed!

git286:
The page is served by a new HTTP listener on 4040. It is separate from the gRPC server, so the auth interceptors on 443 do not apply to it. It binds to all interfaces and the manifest also puts 4040 on the `api` Service, so any pod in any namespace can open `http://api.ate-system.svc:4040/statusz?format=json`.

Even with redaction, the page still shows build version and commit, listener addresses, flag names, PG pool counts, and the last 100 failed Control RPCs. For a platform that runs untrusted agent workloads, I think this is too much to expose with no check.

I suggest closing the network path instead of adding more redaction:
  
- Bind the status listener to `127.0.0.1:4040`. `kubectl port-forward` still works. The router already does this for ext_proc
- Remove the `status` port from the `api` Service. The documented workflow only uses pod port-forward, so nothing needs it.
- Update docs/observability.md. "Only exposed through the cluster-internal Service" includes every tenant pod, so it reads safer than it is.

With this, access control becomes Kubernetes RBAC (`pods/portforward`), which is the same model kube-apiserver uses for its own `/statusz` and `/flagz` (KEP-4827, KEP-4828). If we later need an in-cluster reader, that should be a separate change behind the existing authenticated listener, not 4040 on the Service.

Please keep the redaction anyway. It is still useful when people paste the page into tickets or chat.

swiftdiaries:
addressed review comments and rebased, could you please re-review? @git286

manifests/ate-install/ate-api-server.yaml
```diff
@@ -286,4 +290,4 @@ spec:
   - name: grpc
     protocol: TCP
     port: 443
-    targetPort: 443
+    targetPort: grpc
```
git286: Why change this?

Now that the status port is loopback-only and not on the Service, nothing in this PR needs the name.

cmd/ateapi/main.go
```diff
@@ -308,19 +312,72 @@ func main() {
...
+		},
+		Drain: statusz.Drain{
+			Delay:   drainDelay.String(),
+			Timeout: drainTimeout.String(),
+		},
+		Flags: projectStatusFlags(pflag.CommandLine, statusEnvSources),
+	}, workerCache.Workers, readiness.Ready, poolReader, rpcFailures.Failures, time.Now))
+	statusHTTP, err := startStatusHTTPServer(*statusPort, statusMux, net.Listen)
+	if err != nil {
+		serverboot.Fatal(ctx, "Failed to start status HTTP server", err)
```
git286: a failed bind of the status port should not take down ateapi.                                       

If `127.0.0.1:<status-port>` cannot be opened, this calls `serverboot.Fatal` and the process exits. At this point Postgres pools, informers, and the gRPC listener are already up, so Kubernetes will restart it into CrashLoopBackOff. 

The metrics listener on line 315 handles the same failure by logging and continuing. The status page is less important than metrics, so I think it should do the same: log the error and keep serving gRPC.

PR #1598, opened 2026-09-10
Add agentgateway to CI & rename dataplane flag

keithmattix:
Run agentgateway data plane tests as a part of substrate CI (non-blocking to start so we can confirm it's not flaky). Also, change the `--atenet-router` flag to `--atenet-dataplane` to make it clearer that the flag controls ingress and egress.

I've run the e2es locally across gVisor and microVM plus the MITM variants for both. The only skip we do for agentgateway is `TestIngressProtocolDowngrade` because 1. the behavior its testing only exists on the non-CONNECT atunnel ingress path and agentgateway only sends CONNECT to atunnel and 2. I'm not sure that we want this to be a part of the contract that substrate is bound by (e.g. do we really want to commit to atunnel always parsing HTTP?).

My goal with getting both dataplanes into CI is to start taking steps to codify the proxy (router + egress PEP) contract for substrate. The telemetry they emit, atunnel expectations, etc. are all important contracts to  explicitly call out so that they don't become too coupled to a single dataplane implementation.

- [X] Tests pass


keithmattix:
Current failure is a postgres timeout. I think CI just needs to be re-kicked

internal/e2e/suites/networking/grpcingress_test.go
```diff
@@ -57,6 +58,9 @@ var grpcEchoFixtureManifests = e2e.SubstrateFixtureManifests{
 // TestIngressGRPC below is the positive counterpart: the same path, against an
 // actor that really does speak gRPC.
 func TestIngressProtocolDowngrade(t *testing.T) {
+	if os.Getenv("E2E_DATAPLANE") == "agentgateway" {
+		t.Skip("TODO: is HTTP/2-to-HTTP/1 downgrade, and rejecting gRPC for HTTP/1-only actors, an AgentGateway ingress contract?")
```
EItanya, line 61: Moved from [my comment on #1628](https://github.com/agent-substrate/substrate/pull/1628#discussion_r4005663030); this feedback belongs to #1598.

Is that question here whether this is a general ingress contract?

Keith already [answered “Yes” on the original thread](https://github.com/agent-substrate/substrate/pull/1628#discussion_r4006441827).

internal/e2e/suites/parking/parking_test.go
```diff
@@ -62,11 +64,14 @@ func TestRequestParking(t *testing.T) {
 		t.Fatalf("creating router client: %v", err)
 	}
 	defer router.Close()
-	statusz, err := e2e.NewStatuszClient(ctx)
-	if err != nil {
-		t.Fatalf("creating statusz client: %v", err)
+	var statusz *e2e.StatuszClient
+	if os.Getenv("E2E_DATAPLANE") != "agentgateway" {
+		statusz, err = e2e.NewStatuszClient(ctx)
```
EItanya: Moved from [my comment on #1628](https://github.com/agent-substrate/substrate/pull/1628#discussion_r4005667725); this feedback belongs to #1598.

Can we use a more generic solution here?

EItanya (commented):
🤖 Relocated with AI assistance; the original inline comment wording is preserved.

Moved from [my review on #1628](https://github.com/agent-substrate/substrate/pull/1628#pullrequestreview-5198203845), which is stacked on this PR. The three inline comments and the ingress-contract portion of that review belong here:

> I have some clarifying questions about what the actual ingress contract is since there are definitely dataplane specific checks in the tests

The benchmark-specific questions remain on #1628. The CI matrix suggestion is already addressed here.

internal/e2e/suites/parking/parking_test.go
```diff
@@ -273,14 +284,39 @@ func waitForParkedCount(ctx context.Context, t *testing.T, statusz *e2e.StatuszC
...
+			if cond(active) {
 				return
 			}
 		}
 		time.Sleep(150 * time.Millisecond)
 	}
 	t.Fatalf("timed out waiting for the parking gauge to satisfy the condition (last active=%d)", last)
 }
+
+func agentGatewayParkingCount(ctx context.Context) (int, bool) {
```
EItanya: Moved from [my comment on #1628](https://github.com/agent-substrate/substrate/pull/1628#discussion_r4005672113); this feedback belongs to #1598.

Same question about unifying the data gathering mechanism here

Keith’s [reply on the original thread](https://github.com/agent-substrate/substrate/pull/1628#discussion_r4006449447) points to this PR’s description and its goal of defining a shared proxy contract.

internal/e2e/suites/parking/parking_test.go
keithmattix: My plan is to just get both data planes to a working state in CI then iterate on a contract that both data planes implement. New contracts need would need to be persisted there instead of one off. So agree with the comment but will do it in a follow up

internal/e2e/suites/parking/parking_test.go
keithmattix: Yeah +1 to my comment above; we'll define it in a shared contract

internal/e2e/suites/networking/grpcingress_test.go
keithmattix: Answering again here for posterity: yes that's the question. I also bring it up in the PR description

benchmarking/automation/tests.yaml
```diff
@@ -43,7 +43,7 @@
 #                  outrun the default.
 #   ateArgs        Optional. List of strings appended verbatim to
 #                  `hack/install-ate.sh --deploy-ate-system` when this
-#                  test's substrate is deployed (e.g. ["--atenet-router=agentgateway"]
+#                  test's substrate is deployed (e.g. ["--dataplane=agentgateway"]
```
LiorLieberman: maybe atenet-dataplane? easier to connect that this change happens in atenets

internal/e2e/suites/networking/networking_test.go
```diff
@@ -258,38 +258,46 @@ func postThroughEgressActor(t *testing.T, ctx context.Context, router *e2e.Route
...
-			if !strings.Contains(line, "/actor/"+actorName) {
-				continue
+			switch line.container {
+			case "envoy":
+				authority, ok := accessLogField(line.text, "authority")
+				if ok && strings.HasSuffix(authority, ":"+port) && strings.Contains(line.text, "/actor/"+actorName) {
+					t.Logf("egress gateway tunneled the request: %s", line.text)
+					return true
+				}
+			case "agentgateway":
```
LiorLieberman, line 276: is there a way we can wrap it somehow to avoid many switch cases

internal/e2e/suites/networking/networking_test.go
```diff
@@ -258,38 +258,46 @@ func postThroughEgressActor(t *testing.T, ctx context.Context, router *e2e.Route
...
 // replica, until predicate accepts the lines written since.
-func waitForAccessLog(t *testing.T, ctx context.Context, since metav1.Time, want string, predicate func(lines []string) (bool, error)) {
+func waitForAccessLog(t *testing.T, ctx context.Context, since metav1.Time, want string, predicate func(lines []gatewayAccessLogLine) bool) {
 	t.Helper()
 	const (
 		gatewayNamespace = "ate-system"
 		gatewaySelector  = "app=atenet-egress"
-		gatewayContainer = "envoy"
-		// The access log's line prefix, from the HttpConnectionManager
-		// text_format_source in manifests/ate-install/atenet-egress.yaml.
```
LiorLieberman, line 291: where did these go to

internal/e2e/suites/parking/parking_test.go
```diff
@@ -62,11 +64,14 @@ func TestRequestParking(t *testing.T) {
 		t.Fatalf("creating router client: %v", err)
 	}
 	defer router.Close()
-	statusz, err := e2e.NewStatuszClient(ctx)
-	if err != nil {
-		t.Fatalf("creating statusz client: %v", err)
+	var statusz *e2e.StatuszClient
+	if os.Getenv("E2E_DATAPLANE") != "agentgateway" {
```
LiorLieberman: We really need to avoid if envoy if agentgateway everywhere? is there a way we can initialize it in the suit once and avoid switch cases everywhere?

internal/e2e/collector_metrics.go
```diff
@@ -31,6 +31,7 @@ const (
 	collectorNamespace = "otel-system"
 	collectorService   = "opentelemetry-collector"
 	collectorPromPort  = 8889
+	routerStatsPort    = 15020
```
LiorLieberman: what is this port for?

LiorLieberman (commented):
Thanks Keith, LGTM overall a few questions

internal/e2e/suites/networking/networking_test.go
keithmattix: Abstracted into `accessLogField` and the caller `assertEgressGatewayConnect`

internal/e2e/collector_metrics.go
keithmattix: Will rename to prefix with agentgateway. The current metrics reader in our test code relies on OTLP exported metrics while agentgateway just has native prom metrics

internal/e2e/suites/networking/networking_test.go
keithmattix: I can make an interface in the testbed to encode all the dataplane specific behavior. FWIW, my plan is to create a dataplane contract so some of these nuances/features get described in a central place, but we'll have the test wrapper to make this cleaner in the mentime

keithmattix: That being said it's a bigger change than the few conditionals we have today, so I'll push it in a new commit to get some feedback

internal/e2e/suites/parking/parking_test.go
keithmattix: Implemented something in [25c30d6](https://github.com/agent-substrate/substrate/pull/1598/commits/25c30d6eda8cfae7cdeb87b55d8c55b0a1ae6b24)

benchmarking/automation/tests.yaml
keithmattix: Done

LiorLieberman (approved):
LGTM
Thanks @keithmattix

PR #1602, opened 2026-09-10
ateletpb: split the Restore snapshot source from the checkpoint destination

dberkov:
Pure refactoring of the atelet Restore API, preparing for the node-local snapshot cache (#690, #1551) without any behavior change: `RestoreRequest` gets its own external snapshot *source* type, so read-side attributes of a restore source have a home — today the URI, next (in the M2 cache work) a sharing/cacheability property set by the control plane.

Three commits, each buildable and tested:

1. **`ateletpb: give Restore its own external snapshot source message`** — `ExternalRestoreConfiguration` replaces `ExternalCheckpointConfiguration` in `RestoreRequest`'s config oneof. The type change is wire-compatible: `snapshot_uri` keeps field number 1, so old peers decode the message exactly as before (the constraint is documented on the field). `base_config` (field 16) supersedes the top-level `golden_snapshot_uri` string, typing the DATA_ON_GOLDEN base as a restore source too — a `string → message` change is not wire-compatible, hence the new field number and the transition below. `CheckpointRequest` is untouched.
2. **`atelet: read the restore base snapshot from base_config`** — dual-read: one helper prefers `base_config` and falls back to the superseded field, so an old ateapi keeps working. Validation requires a transitional caller's two fields to agree and rejects a base snapshot outside `SNAPSHOT_SCOPE_DATA_ON_GOLDEN`.
3. **`ateapi: send the restore base snapshot as base_config`** — dual-write on both DATA_ON_GOLDEN resume paths, so an old atelet keeps working. A new functional test drives the golden-data resume end to end (onCommit DATA + onResume.fromData GOLDEN → suspend → resume) and pins both fields set and agreeing on the wire.

**Rollout**: safe in either order — new ateapi → old atelet ignores the unknown fields and reads `golden_snapshot_uri`; old ateapi → new atelet falls back to it. A follow-up PR removes (and reserves) field 12 and drops the dual read/write once both components have rolled.

Tested: full ateapi + atelet suites including the controlapi functional tests; gofmt, golangci-lint, boilerplate clean.

### DIFF
--- a/internal/proto/ateletpb/atelet.proto
+++ b/internal/proto/ateletpb/atelet.proto
@@ -360,6 +360,17 @@ enum SnapshotScope {
   SNAPSHOT_SCOPE_DATA_ON_GOLDEN = 3;
 }
 
+// ExternalRestoreConfiguration is an external snapshot a restore reads.
+// Split from the checkpoint-side ExternalCheckpointConfiguration (a write
+// destination) so read-side attributes of a restore source have a home.
+message ExternalRestoreConfiguration {
+  // The object storage URI of the snapshot to read. Object names are
+  // appended to it, so it addresses the snapshot as a whole rather than any
+  // one object. Must stay field 1: old peers decode this message as
+  // ExternalCheckpointConfiguration.
+  string snapshot_uri = 1;
+}
+
 message CheckpointRequest {
   string target_ateom_uid = 1;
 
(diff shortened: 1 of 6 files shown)


internal/proto/ateletpb/atelet.proto
```diff
@@ -433,18 +444,16 @@ message RestoreRequest {
...
-  // The object storage URI of the ActorTemplate's golden snapshot.
-  // Set only when scope is SNAPSHOT_SCOPE_DATA_ON_GOLDEN: restore combines
-  // the golden snapshot (memory + full fs delta) with the durable data in
-  // the snapshot referenced by `config`. A top-level field rather than part
-  // of the `config` oneof: the actor's snapshot may be local (a pause
-  // checkpoint) while the golden snapshot is always external.
+  // Superseded by base_config. During the transition callers set both and
+  // atelet prefers base_config, so ateapi and atelet can roll in either
+  // order; removed (and reserved) in a follow-up once both sides have
+  // rolled.
```
BenTheElder: 🤖 **should-fix** 🟡 – The transition isn't needed. Before 1.0 we don't keep cross-version wire compatibility: a release deploys as a unit (versioned atelet DaemonSet, version-labeled nodes), so a new ateapi never talks to an old atelet, or the other way round.

That means this PR can replace `golden_snapshot_uri` with `base_config` outright, and a lot goes away:

- the dual-write on both resume paths in `workflow_resume.go`
- `restoreBaseConfig`'s fallback and the must-agree check in `validateRestoreRequest` (`cmd/atelet/main.go:1735-1760`)
- the dual-field assertions in the new functional test
- the follow-up PR to remove and reserve field 12
- the "Must stay field 1" constraint on `ExternalRestoreConfiguration.snapshot_uri`

The refactor you're after, a restore-source type with a home for read-side attributes, stays exactly as it is.

BenTheElder (commented):
I think I agree with the agent here, this could be a lot simpler if we leaned into the fact that currently we don't support rolling upgrades on a node.

PR #1601, opened 2026-09-10
benchmarking: expose raw atelet and ate metrics via Prometheus export…

Nishanth29:
Fixes #1589

### What this PR does
This PR routes incoming raw OTLP metrics from `atelet` and `ateapi` to the Prometheus pull exporter on port 8889 in `benchmarking/telemetry/meter.yaml`.

### Problem
Currently, `telemetry-meter` only routes aggregated count metrics (`substrate_spans` and `substrate_datapoints`) to the Prometheus exporter on port 8889. Raw OTLP metrics (including snapshot sizes, restore durations, and worker pool states) are only forwarded upstream to managed cloud collectors.

In local or standalone benchmarking environments where cloud monitoring backends are not configured, in-cluster Prometheus cannot scrape these raw histograms and gauges.

### Proposed Changes
In `benchmarking/telemetry/meter.yaml`:
- Add a `metrics/raw` pipeline under `service.pipelines` routing incoming `otlp` metrics through the `memory_limiter` processor to the `prometheus` exporter on port 8889.
- Leaves existing forwarding and counting pipelines intact.

### How this was tested
- Deployed on a benchmark cluster and ran the `glutton` user workload.
- Verified that `telemetry-meter:8889` exposes `atelet_snapshot_size_bytes` and `ate_workerpool_workers`.
- Verified that in-cluster Prometheus scrapes these metrics via the default pod annotations without error.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

PR #1599, opened 2026-09-10
benchmarking: expand cAdvisor scrape allowlist for kernel PSI, networ…

Nishanth29:
Fixes #1588

### What this PR does
This PR updates the Prometheus cAdvisor scrape configuration in `benchmarking/monitoring.yaml` to retain Linux kernel Pressure Stall Information (PSI), network bandwidth, disk I/O throughput, and host-level node resource utilization.

### Problem
Currently, `monitoring.yaml` restricts cAdvisor scrapes to only:
`container_(memory_working_set_bytes|cpu_usage_seconds_total|spec_memory_limit_bytes)`
and unconditionally drops series with `container=""`.

Because cAdvisor reports the host root cgroup (`id="/"`) with an empty container label, this configuration strips:
1. All host-level kernel PSI stalls (`cpu.pressure`, `io.pressure`, `memory.pressure`).
2. Host system daemon resource overhead (`containerd`, `kubelet`, virtualization runtime).
3. Container network bandwidth (`container_network_*`) and disk I/O (`container_fs_*`).

Under high-density multi-tenant agent benchmarks, diagnosing whether tail latency spikes stem from CPU scheduling contention, disk storage bottlenecks, or memory pressure is difficult without kernel PSI.

### Proposed Changes
In `benchmarking/monitoring.yaml`:
1. **Relabel Host Root cgroup**: Relabel `id="/"` to `container="node"` before dropping empty container labels, preserving host-level node rollups.
2. **Expand Allowlist Regex**: Keep `container_pressure_*` (Linux PSI), `container_network_*`, `container_fs_*`, `container_cpu_cfs_*`, and `container_memory_rss`.
3. **Prevent Pod Double-Counting**: Drop empty container labels selectively *only* for cumulative memory working set and CPU usage to prevent 2x double-counting between pod slices and container scopes.

### How this was tested
Tested on a benchmark cluster running the `glutton` user workload:
- Verified that `container_pressure_*` metrics (CPU, memory, and I/O) are actively collected and queryable in Prometheus.
- Verified that host-level root cgroup metrics appear under `container="node"`.
- Verified that container network and filesystem I/O metrics are retained, while pod-level duplicate CPU and memory working-set series are dropped as intended.

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR

PR #1603, opened 2026-09-10
benchmarking/locust: capture cluster hardware facts and density frontiers.

### LINKED ISSUE #1590: [benchmarking] Capture cluster hardware density frontiers and Prometheus server telemetry in Locust runner
### Background & Problem
According to the [Actor Density Benchmark Specs](https://docs.google.com/document/d/1sv5aBXvGOQ69iaqFxdPZ-d6tNbh5AMuwKODDyYjzYQw/edit?tab=t.0), key evaluation targets are active actors per node, per vCPU, per GB RAM, and the **actor-to-pod (A/P) bin-packing ratio** (defined as **active concurrent actors / worker pod count**).

Currently:
1. `benchmarking/locust/runner.py` only records client-side request latencies and does not discover cluster hardware capacity (nodes, vCPUs, RAM allocatable).
2. During high-density benchmark runs, client requests may appear successful while server-side actors are queued or experiencing memory pressure. We need ground-truth telemetry from Prometheus (concurrent active/suspended actors, PSI pressure stalls, snapshot sizes, and ateapi throughput) captured alongside client metrics.

### Proposed Changes
1. **Cluster Hardware Discovery:** Add RBAC in `locust.yaml` allowing the runner to query `nodes` and compute cluster density frontiers (`actors/node`, `actors/vCPU`, `actors/GB RAM`, and the steady-state **A/P ratio**: `active_actors / worker_pods`) written into `stats.jsonl`.
2. **Prometheus Telemetry Harvester (`server_telemetry.py`):** Query in-cluster Prometheus at the end of benchmark trials for ground-truth actor packing, PSI kernel pressure, and snapshot performance, outputting to `server_summary.json` and `stats.jsonl`.
3. Preserve `status.json` contract (`{"locust_exit_code": 0, "stats_generated": true}`) to maintain compatibility with test harnesses and CI orchestrators.

### References
- [Agent Substrate: Actor Density Benchmark Specs](https://docs.google.com/document/d/1sv5aBXvGOQ69iaqFxdPZ-d6tNbh5AMuwKODDyYjzYQw/edit?tab=t.0)
- [Substrate Benchmark Telemetry & Metrics Enhancement](https://docs.google.com/document/d/1f9b1mTyxICeLZPojrxCujZ16Elh4IFo3_WmBeNCvk1c/edit?pli=1&tab=t.0)

cc @maxsmythe @roycaihw @aditya-shantanu

Nishanth29:
Fixes [#1590](https://github.com/agent-substrate/substrate/issues/1590)

## What this PR does

In alignment with the [Actor Density Benchmark Specs](https://docs.google.com/document/d/1sv5aBXvGOQ69iaqFxdPZ-d6tNbh5AMuwKODDyYjzYQw/edit?tab=t.0), this PR teaches the Locust runner to discover cluster capacity, record actor density frontiers, and harvest server-side Prometheus metrics into `stats.jsonl` and `server_summary.json`. Discovery lives in its own module, `benchmarking/locust/cluster_facts.py`, rather than inside `runner.py`.

## Proposed Changes

**Cluster hardware discovery (`cluster_facts.py`, `locust.yaml`)**

- Reads allocatable cores, RAM, node count, machine type and worker pod count through the official Kubernetes Python client, so in-cluster and local auth both work without a `kubectl` subprocess.
- Adds a ClusterRole with `list` on nodes and pods. `--no-cluster-facts` skips discovery, since listing is expensive on a large cluster.
- Anything unreadable stays null. Nothing is guessed or defaulted, so a real reading is always distinguishable from a missing one.

**Density frontiers (`trial_summary` in `stats.jsonl`)**

- `actors_per_node`, `actors_per_vcpu`, `actors_per_gb_ram`.
- Actors per pod as `ap_ratio_p50` / `p90` / `p99` over the steady-state samples rather than one average, so it is visible whether the system actually pegs at 1.
- Raw readings persist beside the derived ones in `raw_configuration`, so ratios can be re-derived after the fact.

**Server ground truth (`server_telemetry.py`, `server_summary.json`)**

- Physical worker assignment from `ate_workerpool_workers`, as packing percentiles plus the underlying timeseries.
- Host Linux kernel PSI stalls for CPU, memory and IO, and CFS throttling.
- Snapshot sizes as P50, P90, P95 and mean, checkpoint counts, restore and checkpoint latencies, and throughput. Counts, mean and throughput cover the steady-state window; the quantiles are a 5m rate at window end.
- `server_telemetry.py` uses only the standard library, every call with a timeout. An unreachable Prometheus leaves nulls rather than failing the run, and `--prometheus-url` retargets it. `status.json` keeps its existing minimal schema.
- Depends on #1599 and #1601 landing first: the PSI queries need the `container="node"` relabel and the snapshot queries need the `metrics/raw` pipeline. Merged ahead of those, the PSI and snapshot fields come back `null`.

**Docs**

- `benchmarking/README.md` covers the new flags and every field in the output files.

## How this was tested

- 15 unit tests across `test_cluster_facts.py` and `test_server_telemetry.py`, covering percentile edges, steady-state detection, counter resets across an atelet restart, and the null-versus-zero rules.
- Multi-user runs on a benchmark cluster. Frontier fields land in `stats.jsonl`, `server_summary.json` is populated, and `status.json` is unchanged.
- Emitted values were re-derived by hand from raw Prometheus and matched.

## References

[Agent Substrate: Actor Density Benchmark Specs](https://docs.google.com/document/d/1sv5aBXvGOQ69iaqFxdPZ-d6tNbh5AMuwKODDyYjzYQw/edit?tab=t.0)

- [x] Tests pass

### DIFF
--- a/benchmarking/automation/manifests/runner-job.yaml.tmpl
+++ b/benchmarking/automation/manifests/runner-job.yaml.tmpl
@@ -41,6 +41,36 @@ roleRef:
   name: atelet-endpointslices
   apiGroup: rbac.authorization.k8s.io
 ---
+# Hardware discovery for the runner: runner.py reads nodes and worker pods to
+# derive cluster capacity frontiers. Same ClusterRole as the one in
+# benchmarking/locust/manifests/locust.yaml, repeated here so this manifest
+# stands alone -- the locust Deployment is not part of an automated run.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: locust-hardware-discovery
+rules:
+- apiGroups: [""]
+  resources: ["nodes", "pods"]
+  verbs: ["get", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  # Intentionally not named locust-hardware-discovery: that binding belongs to
+  # locust.yaml. Two bindings on one ClusterRole add up, but two bindings
+  # sharing a name overwrite each other, silently revoking whichever subject
+  # the losing copy listed.
+  name: benchmark-runner-hardware-discovery
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: locust-hardware-discovery
+subjects:
+- kind: ServiceAccount
+  name: benchmark-runner
+  namespace: benchmarking
+---
 apiVersion: batch/v1
 kind: Job
 metadata:
--- a/benchmarking/locust/cluster_facts.py
+++ b/benchmarking/locust/cluster_facts.py
@@ -1,0 +1,267 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Discovers cluster hardware capacity and records per-trial density frontiers.
+
+Reads allocatable CPU/RAM, node count and worker pod count from the Kubernetes
+API, then derives the actor-density frontiers (actors per node / vCPU / GB RAM
+and the A/P bin-packing percentiles) for a completed trial.
+"""
+
+from __future__ import annotations
+
+import argparse
+import csv
+import json
+from pathlib import Path
+from typing import Any, TextIO
+
+from kubernetes import client, config
+from kubernetes.client.rest import ApiException
+from kubernetes.utils import parse_quantity
+
+API_TIMEOUT_SECONDS = 5
+WORKER_POOL_NAMESPACE = "benchmark-workloads"
+WORKER_POOL_LABEL = "ate.dev/worker-pool"
+LIVE_POD_PHASES = ("Running", "Pending")
+MACHINE_TYPE_LABEL = "node.kubernetes.io/instance-type"
+
+# Shape returned when the cluster cannot be read, or when discovery is skipped
+# with --no-cluster-facts. Keeping one definition means a trial_summary row has
+# the same keys either way, so consumers never have to special-case it.
+EMPTY_FACTS: dict[str, Any] = {
+    "machine_type": None,
+    "node_count": None,
+    "allocatable_cores": None,
+    "allocatable_ram_gb": None,
+    "worker_pod_count": None,
+}
+
+
+def _log(logs: TextIO | None, msg: str) -> None:
+    """Mirrors runner.tee without importing it, to avoid a circular import."""
+    print(msg, flush=True)
+    if logs is not None:
+        logs.write(msg + "\n")
+        logs.flush()
+
+
+def _load_kube_config(logs: TextIO | None = None) -> bool:
+    """Loads in-cluster credentials, falling back to a local kubeconfig."""
+    try:
+        config.load_incluster_config()
+        return True
+    except config.ConfigException:
+        pass
+    try:
+        config.load_kube_config()
+        return True
+    except config.ConfigException as e:
+        _log(logs, f"Notice: no Kubernetes credentials available: {e}")
+        return False
+
+
+def _count_worker_pods(v1: client.CoreV1Api, logs: TextIO | None = None) -> int | None:
+    """Counts live pods in the worker pool.
+
+    Prefers the dedicated worker namespace and falls back to a cluster-wide
+    lookup for clusters that place the pool elsewhere. Both listings are
+    filtered server-side by label so the apiserver never streams us the full
+    pod inventory.
+    """
+    try:
+        pods = v1.list_namespaced_pod(
+            namespace=WORKER_POOL_NAMESPACE,
+            label_selector=WORKER_POOL_LABEL,
+            resource_version="0",
+            _request_timeout=API_TIMEOUT_SECONDS,
+        ).items
+    except ApiException as e:
+        _log(logs, f"Notice: could not list pods in {WORKER_POOL_NAMESPACE}: {e.reason}")
+        pods = []
+
+    if not pods:
+        pods = v1.list_pod_for_all_namespaces(
+            label_selector=WORKER_POOL_LABEL,
+            resource_version="0",
+            _request_timeout=API_TIMEOUT_SECONDS,
+        ).items
+
+    live = [p for p in pods if p.status.phase in LIVE_POD_PHASES]
+    return len(live) or None
+
+
+def get_cluster_hardware_facts(logs: TextIO | None = None) -> dict[str, Any]:
+    """Reads allocatable node capacity and worker pod count from the cluster.
+
+    Never raises: a trial must still publish its results when the cluster is
+    unreadable, so any failure leaves the affected facts as None.
+    """
+    facts: dict[str, Any] = dict(EMPTY_FACTS)
+    if not _load_kube_config(logs):
+        return facts
+
+    v1 = client.CoreV1Api()
+
+    # resource_version="0" is served from the apiserver's watch cache rather
+    # than etcd, which keeps this cheap on large clusters.
+    try:
+        nodes = v1.list_node(
+            resource_version="0", _request_timeout=API_TIMEOUT_SECONDS
+        ).items
+        total_cores = 0.0
+        total_ram_bytes = 0
+        machine_types = set()
+        for node in nodes:
+            allocatable = node.status.allocatable or {}
+            if "cpu" in allocatable:
+                total_cores += float(parse_quantity(allocatable["cpu"]))
+            if "memory" in allocatable:
+                total_ram_bytes += int(parse_quantity(allocatable["memory"]))
+            labels = (node.metadata.labels or {}) if node.metadata else {}
+            machine_type = labels.get(MACHINE_TYPE_LABEL)
+            if machine_type:
+                machine_types.add(machine_type)
+        facts["node_count"] = len(nodes)
+        facts["allocatable_cores"] = round(total_cores, 2)
+        facts["allocatable_ram_gb"] = round(total_ram_bytes / (1024**3), 2)
+        # Recorded so results stay comparable across hardware changes. A
+        # heterogeneous pool is reported as a sorted comma-joined list rather
+        # than picking one node arbitrarily.
+        facts["machine_type"] = ",".join(sorted(machine_types)) or None
+    except Exception as e:
+        _log(logs, f"Notice: could not read node capacity: {e}")
+
+    try:
+        facts["worker_pod_count"] = _count_worker_pods(v1, logs)
+    except Exception as e:
+        _log(logs, f"Notice: could not count worker pods: {e}")
+
+    return facts
+
+
+def append_trial_summary(
+    jsonl_path: Path,
+    stats_csv: Path,
+    stats_history_csv: Path,
+    args: argparse.Namespace,
+    data_ts: str,
+    facts: dict[str, Any],
+    logs: TextIO | None = None,
+) -> None:
+    active_users = args.users
+    machine_type = facts.get("machine_type")
+    node_count = facts.get("node_count")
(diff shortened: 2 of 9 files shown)


benchmarking/locust/runner.py
```diff
@@ -289,7 +298,16 @@ def run_test(args: argparse.Namespace, csv_prefix: Path, logs: TextIO, traces: T
 
     boomer_proc = None
     if with_boomer:
-        boomer_cmd = [BOOMER_BINARY, "--user-class", Path(test_file(args.file)).stem]
+        boomer_cmd = [BOOMER_BINARY]
+        if "boomer-glutton" not in os.path.basename(BOOMER_BINARY):
```
maxsmythe: Is all this switching to avoid breaking changes? Shouldn't the binary be bundled in the container? May be able to just straight rename the binary.

benchmarking/locust/runner.py
```diff
@@ -371,6 +389,304 @@ def stats_to_jsonl(stats_csv: Path, jsonl_path: Path, timestamp: str, tag: str,
...
+        "E": 1000**6,
+    }
+    for suffix, mult in multipliers.items():
+        if mem_str.endswith(suffix):
+            num = float(mem_str[: -len(suffix)].strip())
+            return int(num * mult)
+    return int(float(mem_str))
+
+
+def get_cluster_hardware_facts(logs: TextIO | None = None) -> dict[str, any]:
```
maxsmythe: I think this batch of code is large enough and focused enough to warrant its own file.

benchmarking/locust/runner.py
```diff
@@ -371,6 +389,304 @@ def stats_to_jsonl(stats_csv: Path, jsonl_path: Path, timestamp: str, tag: str,
...
+    facts: dict[str, any] = {
+        "node_count": None,
+        "allocatable_cores": None,
+        "allocatable_ram_gb": None,
+        "worker_pod_count": None,
+    }
+    # 1. Environment variables override
```
maxsmythe: Why do we need overrides?

benchmarking/locust/runner.py
```diff
@@ -371,6 +389,304 @@ def stats_to_jsonl(stats_csv: Path, jsonl_path: Path, timestamp: str, tag: str,
...
+        except ValueError:
+            pass
+
+    if all(
+        facts[k] is not None
+        for k in ("node_count", "allocatable_cores", "allocatable_ram_gb", "worker_pod_count")
+    ):
+        return facts
+
+    # 2. In-cluster HTTP API or local kubectl discovery
```
maxsmythe: I think K8s has a Python client we can use for this?

benchmarking/locust/runner.py
```diff
@@ -371,6 +389,304 @@ def stats_to_jsonl(stats_csv: Path, jsonl_path: Path, timestamp: str, tag: str,
...
+                                p for p in p_data.get("items", [])
+                                if p.get("status", {}).get("phase") in ("Running", "Pending")
+                            ]
+                            if items:
+                                facts["worker_pod_count"] = len(items)
+                                break
+                    except Exception:
+                        pass
+        elif shutil.which("kubectl"):
+            # Local workstation fallback via kubectl
```
maxsmythe: Why do we need this? Also I think we may get standard auth for free w/a client

benchmarking/locust/runner.py
```diff
@@ -437,6 +757,49 @@ def main() -> None:
                         jsonl_path.unlink()
                 else:
                     stats_generated = jsonl_path.exists()
+                    if stats_generated:
+                        facts = get_cluster_hardware_facts(logs)
```
maxsmythe: for truly large clusters, we may want the ability to disable this (hitting every node may take a long time)

maxsmythe (changes requested):
Thanks for this! Left a few initial comments.

benchmarking/locust/runner.py
Nishanth29: good call, will be adding a --cluster-facts / --no-cluster-facts flag (default on) so it can be skipped
entirely.

benchmarking/locust/runner.py
Nishanth29: yeah, my benchmark cluster was still running an older image from before the #1295 rename, so I needed
these to get my trials running against it. None of it applies to a normal run since the Dockerfile
bundles /app/boomer-worker. Reverting to that and dropping the env knobs.

benchmarking/locust/runner.py
Nishanth29: I added them as a fallback in case the runner couldn't read the cluster, but this PR adds the RBAC for
that anyway so they aren't really needed. Removing them.

benchmarking/locust/runner.py
Nishanth29: oh yeah... I went with urllib because the locust image is distroless and I was trying to avoid
adding a dependency, but we already install pip packages in there anyway, so that isn't really a
concern. will be switching to the client.

benchmarking/locust/runner.py
Nishanth29: that was just so I could run runner.py from my workstation, but you're right that the client handles
that for us. Dropping the kubectl path.

benchmarking/locust/runner.py
Nishanth29: will move it into its own file

Nishanth29:
Thanks Max, good catches. All six are fixed: reverted the boomer switching,
moved discovery into `cluster_facts.py`, dropped the env overrides and the
kubectl fallback, switched to the K8s client, and added `--cluster-facts` /
`--no-cluster-facts` so the listing can be skipped on big clusters.

Also fixed some telemetry bugs, rewrote the tests and documented the flags and output fields in the README.

benchmarking/automation/manifests/runner-job.yaml.tmpl
```diff
@@ -41,6 +41,37 @@ roleRef:
...
   apiGroup: rbac.authorization.k8s.io
 ---
+# Hardware discovery for the runner: runner.py reads nodes and worker pods to
+# derive cluster capacity frontiers. Same ClusterRole as the one in
+# benchmarking/locust/manifests/locust.yaml, repeated here so this manifest
+# stands alone -- the locust Deployment is not part of an automated run.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: locust-hardware-discovery
```
maxsmythe: nit: benchmark-runner-hardware-discovery to avoid name collision here too

benchmarking/locust/unit_tests/test_cluster_facts.py
```diff
@@ -0,0 +1,253 @@
...
+from pathlib import Path
+import sys
+import tempfile
+from types import SimpleNamespace
+import unittest
+from unittest import mock
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+# Probe only the third-party dep, so a bad import in our own modules fails the
```
maxsmythe: This comment doesn't sound right... doesn't this avoid failures do to bad import? Shouldn't we just make sure required modules are installed?

benchmarking/locust/unit_tests/test_cluster_facts.py
```diff
@@ -0,0 +1,253 @@
...
+    HAS_KUBERNETES = True
+except ImportError:  # pragma: no cover - depends on the local environment
+    HAS_KUBERNETES = False
+
+if HAS_KUBERNETES:
+    import cluster_facts
+    import runner
+    from kubernetes.client.rest import ApiException
+
+needs_kubernetes = unittest.skipUnless(
```
maxsmythe: Avoiding conditional imports would avoid complexities like this.

benchmarking/locust/cluster_facts.py
```diff
@@ -0,0 +1,272 @@
...
+        pods = []
+        namespaced_failed = True
+
+    if not pods:
+        reason = (
+            "the namespaced lookup failed"
+            if namespaced_failed
+            else f"no {WORKER_POOL_LABEL} pods in {WORKER_POOL_NAMESPACE}"
+        )
+        _log(logs, f"Notice: {reason}; scanning all namespaces for the worker pool")
```
maxsmythe: Why are we assuming the worker pool may be in some unknown namespace?

What happens if there are multiple namespaces with worker pools?

We probably want to set solid conventions here.

benchmarking/locust/cluster_facts.py
```diff
@@ -0,0 +1,272 @@
...
+    node_count = facts.get("node_count")
+    cores = facts.get("allocatable_cores")
+    ram_gb = facts.get("allocatable_ram_gb")
+    pod_count = facts.get("worker_pod_count")
+
+    actors_per_node = round(active_users / node_count, 2) if node_count else None
+    actors_per_vcpu = round(active_users / cores, 2) if cores else None
+    actors_per_gb_ram = round(active_users / ram_gb, 2) if ram_gb else None
+
+    # A/P bin-packing percentiles over the steady-state samples. None when
```
maxsmythe: I know you said what A/P stands for, but I can't remember. Is this a standard metric? If not, we probably want to be more verbose.

benchmarking/locust/cluster_facts.py
```diff
@@ -0,0 +1,272 @@
...
+                        continue
+                    try:
+                        u = float(row.get("User Count", ""))
+                    except (TypeError, ValueError):
+                        continue
+                    if u > 0:
+                        observed.append(u)
+            # Steady state is every sample at or above 90% of the target. A run
+            # that never got there falls back to every non-zero sample.
+            steady = [u for u in observed if u >= active_users * 0.9] or observed
```
maxsmythe: active users may not be a constant if we are using load shapes:

https://docs.locust.io/en/2.16.1/custom-load-shape.html

Not sure how you'd want to handle that scenario. Easiest would probably be to just drop the calculation threshold.

benchmarking/locust/cluster_facts.py
```diff
@@ -0,0 +1,272 @@
...
+    stats_parsed = False
+    if stats_csv.exists():
+        try:
+            with open(stats_csv) as f:
+                reader = csv.DictReader(f)
+                for row in reader:
+                    name = row.get("Name", "")
+                    reqs = int(row.get("Request Count", 0) or 0)
+                    fails = int(row.get("Failure Count", 0) or 0)
+                    if name == "Aggregated":
```
maxsmythe: Is "Aggregated" only there sometimes? I'm confused why we bother to sum if this row exists.

benchmarking/locust/server_telemetry.py
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+
+def get_steady_state_window(
+    stats_history_csv: Path,
+    active_users: int,
+    start_ts: int,
+    end_ts: int,
+) -> tuple[int, int]:
+    """Derives steady-state [T_steady, T_end] where User Count >= 0.9 * active_users."""
```
maxsmythe: Same comment here about active users not always being constant if custom load shapes are used.

maxsmythe (changes requested):
Some minor questions about structure. Biggest questions are around some odd things that look like anticipated edge cases (such as worker pods in an unexpected namespace), and a question about how custom load shapes interact with the idea of a steady state.

benchmarking/locust/cluster_facts.py
Nishanth29: an empty namespaced list is indistinguishable from the pool sitting elsewhere, since listing a missing namespace isn't an error, so the scan was the guardrail for that. agreed on the convention though, and the scan has no rule for picking between pools if it finds several. the pool is declared in benchmark-workloads by workloads.yaml.tmpl and the WorkerPool CRD is namespaced, so i'll rely on that and let the count be null if the read fails.

benchmarking/locust/unit_tests/test_cluster_facts.py
Nishanth29: the try only wraps kubernetes, ours are imported outside it, so a bad import there still fails. but agreed on the second half, it's a required dep, so dropping the conditional import.

benchmarking/locust/unit_tests/test_cluster_facts.py
Nishanth29: yep, goes away with it.

benchmarking/locust/cluster_facts.py
Nishanth29: not standard, it's actors per pod. renaming to actors_per_pod_p50/p90/p99.

Nishanth29:
superseded by #1723 and #1725, which split this into the kubernetes side and the prometheus side. closing this one.

PR #1605, opened 2026-09-11
atelet: simplify the file cache internals per review

dberkov:
Follow-up to the readability review on #1517, kept as its own PR so that branch stays exactly as reviewed. **Stacked on #1517: review only the top commit** (`atelet: simplify the file cache internals per review`).

What it does, mapped to the review comments:

- **`createDestFile` seam — removed entirely**, along with the `sparseDest` interface and the `Fd()` type-assertion it forced: `CopyFile` opens with `os.Create`, `copySparse` writes real `*os.File`s. The userspace fallback is now tested through the public API: `CopyFile` from disk to `/dev/shm`, where `copy_file_range` genuinely fails with EXDEV (hard-link probe verifies the mounts differ; skips where no second filesystem exists — platforms without a kernel copy path exercise userspace in the plain holes test already). The close-error test is deleted with the seam; it only existed because the seam made it possible.
- **`getTo`'s parameters** — clarified rather than restructured, because neither is a test seam: `fetch` is the caller's downloader (the production API; M2 binds the GCS fetcher through it), and the second parameter is the link-vs-copy serving strategy shared by the two public methods. It now has a named, documented type (`serveHit`) so the loop reads as that split.
- **Debris schedule** — eviction passes now retry `.rm-*` leftovers a previous pass failed to remove (race-free under `evictMu`; skipped in dry-run; test included), so stranded bytes no longer wait for a restart. `tmp/` stays startup-swept only — a later sweep would delete in-flight fetches' working directories — and `SweepDebris`'s doc now says both things explicitly.
- **File naming** — `filecache.go` split into `doc.go` (package contract), `store.go` (store lifecycle), `sweep.go`; the sizing helpers moved next to their consumer in `evict.go`.
- **Comment language** — trimmed to shorter declarative sentences throughout; same technical content.

Kept as-is: `FileFetcher` remains a func type — it is the production injection point (call sites bind a GCS client and object URI via closure), and single-method func types are the stdlib idiom for exactly this (`http.HandlerFunc`, `filepath.WalkFunc`).

Tested: full atelet suite; filecache and sparsefile under `-race -count=3`; gofmt, golangci-lint, boilerplate clean.


dberkov:
Closing: per discussion the cleanup lands as the top commit on #1517 directly (same commit, 5a6b6ba1) instead of a stacked PR.

PR #1616, opened 2026-09-11
Enhancements to setup-gcp

msau42:
Multiple enhancements to setup-gcp

- Add boot disk customizations. Since snapshots are written to boot disk, being able to increase the size will mitigate IO throttling.
- Require only one of project-id or project-number to be specified
- Verify region and cluster-location are compatible

- [x] Tests pass

### DIFF
--- a/tools/setup-gcp/cmd/cluster.go
+++ b/tools/setup-gcp/cmd/cluster.go
@@ -16,7 +16,6 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log/slog"
 	"slices"
@@ -333,10 +332,11 @@ var clusterCmd = &cobra.Command{
 	Use:   "cluster",
 	Short: "Create GKE cluster",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if cfg.ProjectID == "" {
-			return errors.New("--project-id is required")
+		ctx := cmd.Context()
+		if err := resolveProjectID(ctx, &cfg); err != nil {
+			return err
 		}
-		return createClusterIdempotent(cmd.Context(), &cfg)
+		return createClusterIdempotent(ctx, &cfg)
 	},
 }
 
--- a/tools/setup-gcp/cmd/cluster.go
+++ b/tools/setup-gcp/cmd/cluster.go
@@ -58,6 +58,15 @@ func buildCreateClusterRequest(parent string, cfg *Config) *containerpb.CreateCl
 			DatapathProvider: containerpb.DatapathProvider_ADVANCED_DATAPATH,
 		}
 	}
+	nodeConfig := &containerpb.NodeConfig{
+		MachineType: cfg.MachineType,
+	}
+	if cfg.BootDiskSizeGB > 0 {
+		nodeConfig.DiskSizeGb = cfg.BootDiskSizeGB
+	}
+	if cfg.BootDiskType != "" {
+		nodeConfig.DiskType = cfg.BootDiskType
+	}
 	return &containerpb.CreateClusterRequest{
 		Parent: parent,
 		Cluster: &containerpb.Cluster{
@@ -67,9 +76,7 @@ func buildCreateClusterRequest(parent string, cfg *Config) *containerpb.CreateCl
 				{
 					Name:             "substrate-node-pool",
 					InitialNodeCount: 2,
-					Config: &containerpb.NodeConfig{
-						MachineType: cfg.MachineType,
-					},
+					Config:           nodeConfig,
 				},
 			},
 			EnableK8SBetaApis: &containerpb.K8SBetaAPIConfig{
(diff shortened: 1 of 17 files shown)


tools/setup-gcp/cmd/enable.go
```diff
@@ -32,10 +30,11 @@ var enableApisCmd = &cobra.Command{
 	Use:   "apis",
 	Short: "Enable required GCP APIs",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if cfg.ProjectID == "" {
-			return errors.New("--project-id is required")
+		ctx := cmd.Context()
+		if err := resolveProject(ctx, &cfg); err != nil {
```
aditya-shantanu: Chicken-and-egg: `resolveProject` calls the Resource Manager API, but `cloudresourcemanager.googleapis.com` is one of the services this command is about to enable. On a fresh project, `enable apis --project-id=X` (and bootstrap step 1) now fails at resolution, where it previously worked with just `--project-id`. Consider resolving lazily — skip the GetProject call when `ProjectID` is set and the command doesn't need `ProjectNumber` (only the IAM path does).

tools/setup-gcp/cmd/cluster.go
```diff
@@ -106,6 +113,9 @@ func createClusterInternal(ctx context.Context, cfg *Config, client *container.C
 }
 
 func createClusterIdempotent(ctx context.Context, cfg *Config) error {
+	if err := validateClusterLocation(cfg); err != nil {
```
aditya-shantanu, line 116: Both `RunE`s (cluster, bootstrap) already call `validateClusterLocation` before reaching here, so this triple-validates on every path. Keeping only this one (closest to use) seems simplest.

tools/setup-gcp/cmd/cluster.go
```diff
@@ -58,6 +58,15 @@ func buildCreateClusterRequest(parent string, cfg *Config) *containerpb.CreateCl
 			DatapathProvider: containerpb.DatapathProvider_ADVANCED_DATAPATH,
 		}
 	}
+	nodeConfig := &containerpb.NodeConfig{
+		MachineType: cfg.MachineType,
+	}
+	if cfg.BootDiskSizeGB > 0 {
+		nodeConfig.DiskSizeGb = int32(cfg.BootDiskSizeGB)
```
aditya-shantanu: Silent `int64`→`int32` truncation; a typo'd `--boot-disk-size 5000000000` becomes garbage. Cheap to reject out-of-range values in `validateClusterLocation`-style upfront validation (or just make the flag/field `int32`).

tools/setup-gcp/cmd/cluster.go
```diff
@@ -332,4 +365,6 @@ func init() {
 	clusterCmd.Flags().StringVar(&cfg.Subnetwork, "subnetwork", getEnv("SUBNETWORK", "default"), "VPC subnetwork name [env: SUBNETWORK]")
 	clusterCmd.Flags().StringVar(&cfg.MachineType, "machine-type", getEnv("GVISOR_NODE_MACHINE_TYPE", "c3-standard-4"), "Machine type for the gVisor node pool [env: GVISOR_NODE_MACHINE_TYPE]")
 	clusterCmd.Flags().BoolVar(&cfg.EnableDataplaneV2, "enable-dataplane-v2", getEnv("ENABLE_DATAPLANE_V2", true), "Enable Dataplane V2 [env: ENABLE_DATAPLANE_V2]")
+	clusterCmd.Flags().Int64Var(&cfg.BootDiskSizeGB, "boot-disk-size", getEnv("BOOT_DISK_SIZE_GB", int64(500)), "Boot disk size in GB for the node pool [env: BOOT_DISK_SIZE_GB]")
```
aditya-shantanu: `hyperdisk-balanced` isn't supported on all machine families (e2/n1/n2 etc.), so the new default breaks `--machine-type e2-standard-4` where the old GKE-default disk worked. Worth a note in the README (or validating the combination) since machine-type is user-settable right above.

tools/setup-gcp/cmd/enable.go
msau42: Made project-id required and made project-number optional. project-number is lazily resolved during bucket and iam steps.

tools/setup-gcp/cmd/cluster.go
msau42: done

tools/setup-gcp/cmd/cluster.go
msau42: making it int32 to match the proto

tools/setup-gcp/cmd/cluster.go
msau42: reverted the defaults to empty so it'll use GKE's defaults if unset

aditya-shantanu (approved):

ahmedtd (approved):

Issue #1197, opened 2026-08-25
[Upgrade] Implement Substrate System Upgrade


HavenXia:
Implement the substrate system upgrade that can do upgrade from end to end.

Approved Design doc(#473): https://docs.google.com/document/d/1JduAyGZFyqdNp4UhKiv0EN-is3iW5tf5BwGTWbt2ouI/edit?usp=sharing&resourcekey=0-MXG12QCkleIxhY7cOB7g6A

## Task list

- [ ] #861 — `workerImage` optional (+ rename from `ateomImage`): PR #1210
- [ ] Leader election for ate-controller, so exactly one controller acts during the handover window: PR #1212
- [ ] #1268 — compiled-in versioned default worker image per sandbox class
- [ ] #1269 — version-keyed worker sets: the controller only creates and mutates sets of its own build version, pod nodeSelector pins workers to same-version nodes
- [ ] #1270 — versioned atelet DaemonSets selecting nodes by the `ate.dev/substrate-version` label; install-time node labeling

### Upgrade tooling and ops

- [ ] #1272 — upgrade driver (run / rollback / status / cleanup)
- [ ] #1273 — rollback support
- [ ] `install-ate.sh --upgrade-ate-system`: orchestration, GKE node pool handling 
- [ ] Runbook + demo

### Docs

- [ ] #1185 — api-changes.md: rules for changing protos and CRDs
- [ ] version-skew-policy.md, the supported component version combinations

HavenXia:
System upgrade is implementated and documented.

https://github.com/agent-substrate/substrate/blob/main/docs/upgrade.md

PR #1624, opened 2026-09-11
docs: reorganize the upgrade runbook for better UX

HavenXia:
Part of #1197

- [x] Tests pass


Raji14:
A few suggestions before merge:
1. **Warning 1 under "Three things that break an upgrade"**: 
   The heading is "Three things that break an upgrade", but Warning 1 is phrased as the correct action: *"Flip the node's version label before deleting its old worker pods."*
   Consider rephrasing it to state the anti-pattern (like warnings 2 & 3 do):
   > **Do not delete old worker pods before flipping the node's version label.** Otherwise the old pool reschedules replacements onto the same node next to the new atelet...
2. **Pre-flight check for multi-node clusters**:
   In the previous version, line 199 noted: `# At least two nodes; on a single node the roll is a full stop.` This was omitted from the new checklist. We should add a checklist item confirming the cluster has at least 2 worker nodes, since a rolling zero-downtime upgrade cannot work on a single node.
3. **Rollback list completeness (Steps 3 & 4)**:
   In `## Something's wrong, or I need to undo`, the list jumps from "Past step 5" to "Past step 2". For an operator who reached Step 3 (deployed new atelet) or Step 4 (cloned workerpool) but hasn't flipped nodes yet, it would be helpful to include:
   - `- Past step 4: delete the cloned WorkerPool (kubectl -n $NS delete workerpool $NEW_WORKERPOOL)`
   - `- Past step 3: delete the new DaemonSet (kubectl delete daemonset -n ate-system -l app=atelet,ate.dev/substrate-version=$NEW_VERSION)`
4. **"Check your progress" note phrasing**:
   *"Steps 1, 2 and 6 leave no mark you can read back from the cluster."*
   Those steps do update cluster objects (e.g. `ate-controller` / `api` deployments, CRDs); it's just that the 3 quick-check commands above don't query them. Clarifying: *"The three quick-check commands above do not report whether steps 1, 2, or 6 have run..."* would avoid confusion.
5. **Step 2 readiness check**:
   In Step 2: *"wait for READY to equal DESIRED again on every serving pool... before step 5."* 
   Because Step 4 clones the serving pool, we should probably advise waiting before Step 4 so operators don't clone a pool whose Deployment is actively rolling.
6. **Doc consistency (60s vs 30m)**:
   Good catch updating the suspend window from 60s to 30 minutes to match `workloadGracePeriod`. As a follow-up (or in this PR), `tools/setup-gcp/README.md#2-create-cluster` still mentions "about 60 seconds" and should be updated as well.

Raji14 (commented):
A few suggestions before merge:
1. **Warning 1 under "Three things that break an upgrade"**: 
   The heading is "Three things that break an upgrade", but Warning 1 is phrased as the correct action: *"Flip the node's version label before deleting its old worker pods."*
   Consider rephrasing it to state the anti-pattern (like warnings 2 & 3 do):
   > **Do not delete old worker pods before flipping the node's version label.** Otherwise the old pool reschedules replacements onto the same node next to the new atelet...
2. **Pre-flight check for multi-node clusters**:
   In the previous version, line 199 noted: `# At least two nodes; on a single node the roll is a full stop.` This was omitted from the new checklist. We should add a checklist item confirming the cluster has at least 2 worker nodes, since a rolling zero-downtime upgrade cannot work on a single node.
3. **Rollback list completeness (Steps 3 & 4)**:
   In `## Something's wrong, or I need to undo`, the list jumps from "Past step 5" to "Past step 2". For an operator who reached Step 3 (deployed new atelet) or Step 4 (cloned workerpool) but hasn't flipped nodes yet, it would be helpful to include:
   - `- Past step 4: delete the cloned WorkerPool (kubectl -n $NS delete workerpool $NEW_WORKERPOOL)`
   - `- Past step 3: delete the new DaemonSet (kubectl delete daemonset -n ate-system -l app=atelet,ate.dev/substrate-version=$NEW_VERSION)`
4. **"Check your progress" note phrasing**:
   *"Steps 1, 2 and 6 leave no mark you can read back from the cluster."*
   Those steps do update cluster objects (e.g. `ate-controller` / `api` deployments, CRDs); it's just that the 3 quick-check commands above don't query them. Clarifying: *"The three quick-check commands above do not report whether steps 1, 2, or 6 have run..."* would avoid confusion.
5. **Step 2 readiness check**:
   In Step 2: *"wait for READY to equal DESIRED again on every serving pool... before step 5."* 
   Because Step 4 clones the serving pool, we should probably advise waiting before Step 4 so operators don't clone a pool whose Deployment is actively rolling.
6. **Doc consistency (60s vs 30m)**:
   Good catch updating the suspend window from 60s to 30 minutes to match `workloadGracePeriod`. As a follow-up (or in this PR), `tools/setup-gcp/README.md#2-create-cluster` still mentions "about 60 seconds" and should be updated as well.

HavenXia:
Thanks, pushed second commit. 

On point 2:
I remove "# At least two nodes; on a single node the roll is a full stop." because a single node does not stop the roll, it just means every actor is suspended at once during step 5, so I put that as a note at the top of step 5 rather than a go/no-go item in the checklist.

Raji14:
@HavenXia 

Two open questions for you:
Do you want the dropped assumptions restored in this PR, or filed as follow-up? I'd push for DrainWorker at minimum since it's cheap and prevents a late failure.
The intro still promises "at most one node's worth of capacity is out of service", which contradicts the new single-node caveat in step 5 — worth reconciling given that was your point 2.

HavenXia:
> Haven Xia (@HavenXia)
> 
> Two open questions for you: Do you want the dropped assumptions restored in this PR, or filed as follow-up? I'd push for DrainWorker at minimum since it's cheap and prevents a late failure. The intro still promises "at most one node's worth of capacity is out of service", which contradicts the new single-node caveat in step 5 — worth reconciling given that was your point 2.

Talked offline and add both points.

Raji14:
LGTM, thanks!
Non-blocking note for the future:
$TOKEN across terminal sessions: The checklist sets TOKEN=$(kubectl ...) in the operator's current subshell. If an operator opens a new terminal window to resume at Step 5, $TOKEN may be unset. Adding $TOKEN to the Names used throughout table is a minor nice-to-have for a future pass.

Raji14 (approved):
LGTM

HavenXia:
ci is broken, #1643 should fix it

Issue #1563, opened 2026-09-09
OpenFGA-based authorization - basic


ahmedtd:
Land the basic OpenFGA authorization described in [Authorization for Substrate](https://docs.google.com/document/d/1LYjBlXWLJ18beLDw74Rs_Tq38wFkCTeau4_7_98Op3c/edit?resourcekey=0-zY5W-u0AlAT_GIU6i394CQ&tab=t.6uuc4ftvayew#heading=h.tywbz0y55dxk)

PR #1625, opened 2026-09-11
Introduce Policy Governance APIs in ate api. 

SHRUTI6991:

## Summary

Replaces the generic `GetPolicy` and `SetPolicy` RPCs in `ateapi.proto` with explicit per-resource access policy RPCs (`GetGlobalAccessPolicy`, `UpdateGlobalAccessPolicy`, `GetAtespaceAccessPolicy`, and `UpdateAtespaceAccessPolicy`).

## Changes Made

1. **RPC Declarations (`Control` Service)**:
   - Added `GetGlobalAccessPolicy` and `UpdateGlobalAccessPolicy` for deployment-wide global access control.
   - Added `GetAtespaceAccessPolicy` and `UpdateAtespaceAccessPolicy` for Atespace-level access control.
   - Removed generic `GetPolicy` and `SetPolicy` RPC methods.

2. **Message Schema (`ateapi.proto`)**:
   - Introduced `AccessPolicy`: Describes a policy subresource nested under a parent resource with `metadata` (where `metadata.name` is always `"default"`) and role `bindings`.
   - Introduced `Binding`: Binds principal members (`user:...`, `group:...`, `atespace:...`) to a role defined in `model.fga`.
   - Introduced typed request messages: `GetGlobalAccessPolicyRequest`, `UpdateGlobalAccessPolicyRequest`, `GetAtespaceAccessPolicyRequest`, and `UpdateAtespaceAccessPolicyRequest`.
   - Removed legacy `GetPolicyRequest`, `SetPolicyRequest`, `Policy`, and `Condition` messages.

## Key Design Considerations

- **Singleton Subresource Pattern**: Every parent resource (Global singleton or Atespace) has at most one access policy subresource named `"default"`.
- **Concurrency Control & ABA Protection**: `Update*AccessPolicy` requests enforce `metadata.uid` and `metadata.version` as required preconditions to guard against ABA delete/re-creation and concurrent overwrites.
- **Global Scope Matching**: `metadata.atespace` for Global and Atespace access policies is empty (`""`), as both parent resource types operate at global scope.


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1273,6 +1272,76 @@ message DeleteAtespaceRequest {
...
+  string resource = 1;
+
+  // The policy to set.
+  //
+  // +k8s:required
+  Policy policy = 2;
+}
+
+// Policy represents the access control policy for a resource.
+message Policy {
```
juli4n, line 1297: I think Policy is a very generic name. We already have EgressPolicy, and we will for sure have other policies. Can we think a more concrete name? AuthPolicy?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1273,6 +1272,76 @@ message DeleteAtespaceRequest {
   ObjectRef atespace = 1;
 }
 
+// GetPolicyRequest requests the access control policy for a resource.
+message GetPolicyRequest {
+  // The resource for which the policy is being requested (e.g. "atespaces/team-a", "clusters/root").
+  //
+  // +k8s:required
+  string resource = 1;
```
juli4n, line 1280: Let's discuss the resource model for auth policies. Do we want it to fit into the existing substrate resource model? Are they going to be resources (or sub-resources) themselves? Will each resource have a single policy attached to it? What represents the primary key of the policy. (see docs/api-style-guide.md).

SHRUTI6991: >  Do we want it to fit into the existing substrate resource model?

I actually tried this before. My primary concern is around retrofitting `ObjectRef` for cluster permissions which is built around atespace being the root:`(atespace, name)`.  

For auth, it starts from cluster: https://github.com/agent-substrate/substrate/blob/main/internal/authz/model.fga. Each resource has `owner`, `editor`, `viewer` permission. A cluster admin has the permission to create atespace. It can also give permissions to view atespaces to cluster viewer.

I actually prefer keeping the `resource` as string.  Canonical strings like `atespaces/team-a` map directly 1:1 to OpenFGA too. 

We can introduce `GetClusterPolicy`, `GetAtespacePolicy` but this kind of RPCs leads to resource explosion and isn't very scalable (imo). 

> Are they going to be resources (or sub-resources) themselves?

It should be a sub resource. The existence of an `AccessPolicy` is tied to the presence of parent resource like `Atespace` or `Actor`. Independent resource has their own lifecycles (similar to K8 roles) and I don't think this is necessary for substrate since our backing store is OpenFGA. 

If the resource is deleted, the policies (openfga tuples) should be cascade deleted.  cc: @ahmedtd wdyt? 

> Will each resource have a single policy attached to it?

Yes this is strictly 1:1. Each resource has at most one policy attached to it which is why I used `etag` for guarding the policy state. 

> What represents the primary key of the policy?

We are going to store these as OpenFGA tuples which is defined as `(user, relation, object`. Relationship tuples are stored under OpenFGA's composite primary key.

```
PRIMARY KEY (store, object_type, object_id, relation, user_type, user_id, user_relation)
```

where `object_type: cluster, atespace` and `object_id: root, team-a`.

Design for full context: go/authorization-for-substrate

pkg/proto/ateapipb/ateapi.proto
SHRUTI6991: May be `AccessPolicy` is a better name.

pkg/proto/ateapipb/ateapi.proto
SHRUTI6991: Started a one pager to discuss: go/substrate-policy-governance-api. Let's take it there.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1273,6 +1272,76 @@ message DeleteAtespaceRequest {
...
+
+  // The list of role bindings that grant permissions on the resource.
+  //
+  // +k8s:optional
+  repeated Binding bindings = 3;
+}
+
+// Binding binds a role to a set of principal members.
+message Binding {
+  // The role name (e.g. "roles/owner", "roles/editor", "roles/viewer").
```
thockin, line 1316: Is there any way to fetch (via API) a list of roles and permissions?

I am assuming we do not want to offer custom roles at this point, right?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1273,6 +1272,76 @@ message DeleteAtespaceRequest {
...
+  repeated string members = 2;
+
+  // Optional condition under which this binding is active.
+  //
+  // +k8s:optional
+  Condition condition = 3;
+}
+
+// Condition represents an attribute-based condition expression.
+message Condition {
```
thockin, line 1333: Do we need this in v1?  If we REALLY need it, we need way better docs on what inputs the CEL expression has access to

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1273,6 +1272,76 @@ message DeleteAtespaceRequest {
...
+message Binding {
+  // The role name (e.g. "roles/owner", "roles/editor", "roles/viewer").
+  //
+  // +k8s:required
+  string role = 1;
+
+  // The list of subjects (e.g. "user:alice@example.com", "serviceAccount:...").
+  //
+  // +k8s:optional
+  repeated string members = 2;
```
thockin, line 1324: commented in doc - this needs to be specified more fully

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1273,6 +1272,76 @@ message DeleteAtespaceRequest {
...
+}
+
+// Condition represents an attribute-based condition expression.
+message Condition {
+  // Short title or identifier for the condition.
+  //
+  // +k8s:optional
+  string title = 1;
+
+  // Textual expression (e.g. CEL expression).
```
thockin, line 1339: Is it "e.g. CEL" as in "CEL or something like it" or is it literally CEL ?  We need to document the capabilities if we do this feature

pkg/proto/ateapipb/ateapi.proto
SHRUTI6991: OpenFGA provides an API to provide `AuthorizationModel` which basically gives you the schema. But if we want, we can expose `ListRoles`.

Yes no custom roles at this point. But it can be supported: https://openfga.dev/docs/modeling/custom-roles.

pkg/proto/ateapipb/ateapi.proto
SHRUTI6991: Copying from the doc here:

I actually started with a very simple schema which has "user" (that encompasses everything, human user, K8 service account). 

Future evolution of schema can have something like:

```
type user
type service_account
type group
relations
define member: [user, service_account]
```

1. user:<email-or-sub> (Human User): Authenticated via OIDC token (e.g., user:alice@google.com).

2. service_account:<k8s-namespace>/<k8s-sa-name> (Kubernetes Workload): Maps directly to the Kubernetes Namespace and ServiceAccountName extracted from K8s TokenReview or our substratex509.PodIdentity mTLS certificate extension (e.g., service_account:ate-system/atenet-router).

3. group:<group-id>#member (Group): Maps to OIDC group claims / Google Groups.

4. We can also extend this to actor identity for actor-to-actor authorization.

pkg/proto/ateapipb/ateapi.proto
SHRUTI6991: we don't. I will remove this for now.

pkg/proto/ateapipb/ateapi.proto
SHRUTI6991: we can do this but not for `v1`. Let's take the discussion when we need it.

PR #1626, opened 2026-09-12
Separate ateom certificates vs direct actor certificates

ahmedtd:
We need to draw a strong distinction between credentials that will be wielded directly by actor logic, and credentials used by system components on behalf of an actor.  This will help us avoid attacks where an actor pretends to be a system component.
    
Ateom certificates are requested by ateom/atelet and used for the atunnel connection to the egress gateway.  They have a SPIFFE URI like `spiffe://${trustdomain}/ateom-for-actor/${atespace}/${actor}`.
    
Actor certificates are requested by the egress gateway and used for opening outbound requests on behalf of the actor.  They have a SPIFFE URI like `spiffe://${trustdomain}/actor/${atespace}/${actor}`. Note, this use case is not actually implemented yet.


LiorLieberman:
whats your view on adding another part "/x/" vs prefixing the existing part already - like `spiffe://${trustdomain}/ateom-actor/${atespace}/${actor}`

Wouldn't it ease regexes / validation in different component that uses that peer identity?

cmd/atelet/ateomsupport.go
```diff
@@ -28,12 +29,12 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-type credentialBroker struct {
-	ateletpb.UnimplementedCredentialBrokerServer
-	controlClient ateapipb.ControlClient
+type ateomSupportServer struct {
```
LiorLieberman, line 32: whats the rationale behind "ateomSupportServer" name? whats ateomSupport?

EItanya:
> Stacked over https://github.com/agent-substrate/substrate/pull/1501

Stacking over a PR which doesn't feel ready to go is confusing, can this work stand on its own?

Or even better can we get to the bottom of the questions in https://github.com/agent-substrate/substrate/pull/1501#discussion_r3992057235 before marking these ready for review?

ahmedtd:
> whats your view on adding another part "/x/" vs prefixing the existing part already - like `spiffe://${trustdomain}/ateom-actor/${atespace}/${actor}`
> 
> Wouldn't it ease regexes / validation in different component that uses that peer identity?

Yes, that's a good idea.

cmd/atelet/ateomsupport.go
ahmedtd: It's a server to support ateoms by providing RPCs they need.  I am not attached to the name.  I just want to avoid having $n$ different services between atelet and ateom.

ahmedtd:
> > whats your view on adding another part "/x/" vs prefixing the existing part already - like `spiffe://${trustdomain}/ateom-actor/${atespace}/${actor}`
> > Wouldn't it ease regexes / validation in different component that uses that peer identity?
> 
> Yes, that's a good idea.

Done, I used `ateom-for-actor`

ahmedtd:
This PR grew pretty large, so I have split it into three.

* This PR now only contains the definition and implementation of the new MintAteomActorCertificate RPC.
* #1809 makes atelet/ateom/atunnel use the new RPC
* #1810 cleans up the now-unneeded purpose field.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -2038,3 +2041,36 @@ message SetWorkerCapacityResponse {
...
 }
+
+// Request message for MintAteomActorCertificate.
+message MintAteomActorCertificateRequest {
+  // The actor for which the certificate should be issued.
+  //
+  // Must be a valid actor that currently exists according to the actor store.
+  //
+  // +k8s:required
+  ObjectRef actor = 6;
```
thompsonmax: Why are we starting field numbers at 6 for a brand new message?

cmd/ateapi/internal/workerservice/certificate.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, fmt.Errorf("while retrieving actor: %w", err)
+	}
+	if dbActor.GetMetadata().GetUid() != req.GetActorUid() {
+		return nil, status.Error(codes.Aborted, "conflict; actor has been deleted and recreated")
+	}
+
+	// Parse the CSR.
+	csr, err := x509.ParseCertificateRequest(req.GetCertificateSigningRequest())
+	if err != nil {
+		return nil, fmt.Errorf("while parsing CSR: %w", err)
```
thompsonmax: Should this return `InvalidArgument`?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -2038,3 +2041,36 @@ message SetWorkerCapacityResponse {
...
   Worker worker = 1;
 }
+
+// Request message for MintAteomActorCertificate.
+message MintAteomActorCertificateRequest {
+  // The actor for which the certificate should be issued.
+  //
+  // Must be a valid actor that currently exists according to the actor store.
+  //
+  // +k8s:required
```
thompsonmax, line 2056: Other actor references include "+k8s:subfield(atespace)=+k8s:required," to ensure an atespace is specified, should we require that here as well?

cmd/ateapi/internal/workerservice/certificate.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	if err := csr.CheckSignature(); err != nil {
+		slog.ErrorContext(ctx, "Failed to verify CSR signature", slog.Any("err", err))
```
thompsonmax: [nit] This can be triggered by the client providing a bad signature on a CSR, so should we log at WARN rather than ERROR?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -2038,3 +2041,36 @@ message SetWorkerCapacityResponse {
...
+  bytes certificate_signing_request = 2;
+}
+
+// Response message for MintAteomActorCertificate.
+message MintAteomActorCertificateResponse {
+  // Response contains a list of DER encoded certificates. The first entry is the
+  // leaf certificate, and any remaining entries are intermediates in
+  // leaf-to-root order.
+  repeated bytes actor_certificates = 1;
+}
```
thompsonmax: [nit] dropped the trailing newline here

ahmedtd: Done.

pkg/proto/ateapipb/ateapi.proto
ahmedtd: Done.

pkg/proto/ateapipb/ateapi.proto
ahmedtd: Done.

cmd/ateapi/internal/workerservice/certificate.go
ahmedtd: I don't think we need to log this, actually, I just switched it up so the error is returned.  This doesn't seem like a sensitive error that we need to hide, it's post-authn/z

cmd/ateapi/internal/workerservice/certificate.go
ahmedtd: Done.

LiorLieberman (approved):

HavenXia (approved):

LiorLieberman:
are you planning to update whatever is using the purpose field on the cert to use the new format?

ahmedtd:
> are you planning to update whatever is using the purpose field on the cert to use the new format?

Yes, there are other stacked PRs over this, #1809 and #1810

Issue #1593, opened 2026-09-10
Standardize how ate-apiserver sets resource defaults


juli4n:
There is no way to set field defaults today. We should have some common mechanism to back-fill default values during create RPC calls so that resources stored in the DB always have their default values persisted.

zlammerts-svg:
Julian, is this one done at this point?

PR #1630, opened 2026-09-12
Add support for field defaults

juli4n:
Populate default fields during both Create and Update RPC calls so that resources are stored with defaults instead of zero values. This means that a) consumers don't need to be aware of the default value and b) default values can potentially change.

Only `ActorTemplate` had defaults, but added default functions for all resources so that all RPC handlers are wired up

Note that optional fields with defaults need to be tagged as `+k8s:required`. This is because they are required when they are stored in the DB. This is different than required from the client perspective.

Part of #1593


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -847,26 +847,29 @@ message SandboxConfig {
 
 // +k8s:customValidation # on_commit must be a subset of on_pause
 message SnapshotsConfig {
-  // on_pause selects what is captured during pause actor. UNSPECIFIED is
-  // tolerated for compatibility and reads as FULL.
+  // on_pause defines the scope of the snapshot captured when an actor
+  // is paused. Defaults to FULL when unset.
   //
-  // +k8s:optional
+  // +k8s:required
```
thockin, line 853: In fact we have logic for this in DV: 

```
   // All of our tags are expressed from the perspective of a client of the    
    // API, but the code we generate is for the server. Optional is tricky.    
    //    
    // A field which is marked as optional and does not have a default is    
    // strictly optional. A client is allowed to not set it and the server will    
    // not give it a default value. Code which consumes it must handle that it    
    // might not have any value at all.    
    //    
    // A field which is marked as optional but has a default is optional to    
    // clients, but required to the server. A client is allowed to not set it    
    // but the server will give it a default value. Code which consumes it can    
    // assume that it always has a value.    
    //    
    // One special case must be handled: optional non-pointer fields with    
    // default values. If the default is not the zero value for the type, then    
    // the zero value is used to decide whether to assign the default value,    
    // and so must be out of bounds; we can proceed as above.    
    //    
    // But if the default is the zero value, then the zero value is obviously    
    // valid, and the fact that the field is optional is meaningless - there is    
    // no way to tell the difference between a client not setting it (yielding    
    // the zero value) and a client setting it to the zero value.    
    //    
    // TODO: handle default=ref(...)    
    // TODO: handle manual defaulting    
    if hasDefault, zeroDefault, err := rtv.hasZeroDefault(context); err != nil {                                                                      
        return Validations{}, err    
    } else if hasDefault {    
        if !util.IsNilableType(context.Type) && zeroDefault {    
            return Validations{Comments: []string{"optional value-type fields with zero-value defaults are purely documentation"}}, nil    
        }    
        validations, err := rtv.doRequired(context)    
        if err != nil {    
            return Validations{}, err    
        }    
        for i, fn := range validations.Functions {    
            validations.Functions[i] = fn.WithComment("optional fields with default values are effectively required")    
        }    
        return validations, nil    
    }                                                                                               
```

@yongruilin @jpbetz 2 things go wrong for us to use DV here:

1) Defaults is a totally different tool and different tag style.
2) We really want to be able to say `+default=SNAPSHOT_CONTENT_SCOPE_FULL` but validation-gen barfs: "failed to parse default value "SNAPSHOT_CONTENT_SCOPE_FULL": invalid character 'S' looking for beginning of value".  We could fix that to just take any such value as a Go symbol name (if JSON parse fails with this specific error)

We could do `+default=1` but eww.  Here's what we get if we do that:

```diff
diff --git cmd/ateapi/internal/controlapi/zz_generated.validation.go cmd/ateapi/internal/controlapi/zz_generated.validation.go
index 94c2e534..93047826 100644
--- cmd/ateapi/internal/controlapi/zz_generated.validation.go
+++ cmd/ateapi/internal/controlapi/zz_generated.validation.go
@@ -5658,6 +5658,7 @@ func Validate_SnapshotsConfig(
 			}
 			// call field-attached validations
 			earlyReturn := false
+			// optional fields with default values are effectively required
 			if e := validate.RequiredValue(ctx, op, fldPath, obj, oldObj).MarkShortCircuit(); len(e) != 0 {
 				errs = append(errs, e...)
 				earlyReturn = true
diff --git pkg/proto/ateapipb/ateapi.pb.go pkg/proto/ateapipb/ateapi.pb.go
index 0317b0ed..b4c344d5 100644
--- pkg/proto/ateapipb/ateapi.pb.go
+++ pkg/proto/ateapipb/ateapi.pb.go
@@ -2452,7 +2452,8 @@ type SnapshotsConfig struct {
 	// on_pause defines the scope of the snapshot captured when an actor
 	// is paused. Defaults to FULL when unset.
 	//
-	// +k8s:required
+	// +k8s:optional
+	// +default=1
 	// +k8s:minimum=1
 	// +k8s:maximum=2 # keep this in sync with the SnapshotContentScope enum
 	OnPause SnapshotContentScope `protobuf:"varint,1,opt,name=on_pause,json=onPause,proto3,enum=ateapi.SnapshotContentScope" json:"on_pause,omitempty"`
diff --git pkg/proto/ateapipb/ateapi.proto pkg/proto/ateapipb/ateapi.proto
index 41b76370..d61f8a9a 100644
--- pkg/proto/ateapipb/ateapi.proto
+++ pkg/proto/ateapipb/ateapi.proto
@@ -850,7 +850,8 @@ message SnapshotsConfig {
   // on_pause defines the scope of the snapshot captured when an actor
   // is paused. Defaults to FULL when unset.
   //
-  // +k8s:required
+  // +k8s:optional
+  // +default=1
   // +k8s:minimum=1
   // +k8s:maximum=2 # keep this in sync with the SnapshotContentScope enum
   SnapshotContentScope on_pause = 1;
```

The question is whether we think `optional` or `required` is actually correct for documentation and tools.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -1036,20 +1039,21 @@ message ContainerReadyz {
...
   int32 timeout_seconds = 2;
 }
 
 // HTTPGetAction describes an HTTP GET against the container's interior IP.
 message HTTPGetAction {
-  // path defaults to "/readyz". Must be a URL path starting with "/", using
-  // only RFC 3986 path-segment characters, without query or fragment.
+  // path must be a URL path starting with "/", using only RFC 3986
+  // path-segment characters, without query or fragment.
+  // Defaults to "/readyz" when unset.
```
thockin, line 1054: This is a really bad default - we should fix it to "/" and make "/readyz" explicit.

cmd/ateapi/internal/controlapi/defaults.go
```diff
@@ -0,0 +1,89 @@
...
+// limitations under the License.
+
+package controlapi
+
+import (
+	"github.com/agent-substrate/substrate/pkg/proto/ateapipb"
+)
+
+// defaultActorTemplate applies the ActorTemplate defaults in place.
+func defaultActorTemplate(t *ateapipb.ActorTemplate) {
```
thockin, line 22: Curious why not put these in their respective resource-files?

pkg/proto/ateapipb/ateapi.proto
yongruilin: defaulter-gen supports `+default=ref(xxx)`. There was an effort to support the same in DV a while back but I lost the track. https://github.com/kubernetes/kubernetes/pull/136361

pkg/proto/ateapipb/ateapi.proto
jpbetz: +1.  Defaults should be used sparingly.  Having "port" default to "type: TCP" is arguably sane if that's what the vast majority of users will want.  Having a highly opinionated defaults, defaults that are computed dynamically, or are expected to change across releases tend to be trouble.

cmd/ateapi/internal/controlapi/defaults.go
juli4n: No particular reason. I don't have a strong opinion.

pkg/proto/ateapipb/ateapi.proto
juli4n: I was scoping this PR to introduce the machinery and not change the current defaults. I would suggest we do a follow-up and change this?

pkg/proto/ateapipb/ateapi.proto
thockin: We use `=ref()` exactly twice in all of Kubernetes.  We should probably just EOL that feature.

My original idea above (if the value of the default tag does not parse as JSON treat it as a Go symbol) fails because we would want to say `+default=PROTO_ENUM_VAL` but that is encoded as `SnapshotContentScope_SNAPSHOT_CONTENT_SCOPE_FULL` in Go (which is stupid, IMO, but that's how it is).

I'm not sure what the right answer is but we should think about how to do better.  Not blocking.

cmd/ateapi/internal/controlapi/defaults.go
thockin: The main argument for putting it in the per-resource file is (IMO) that this is where I would expect to find it.  Having it in a different file is different from all the other trends towards per-resource files.  Not blocking, can followup

thockin:
FTR: Julian and I spoke off-bug.  The preference is to frame `required` / `optional` in terms of the stored data (a field which has a default value is required in the database, though it is optional to clients.

We do not have a DV tag to express that and `+default` is not usable without some work so for now we are marking such fields a `required` and (in English) documenting the default value.  In the future I would love to move defaulting to a declarative tag where we can, but it's low ROI for the moment.

There is an issue where primitive fields may be logically "required" but their zero-value is valid.  If you mark those as `required` DV will check that the value != zero_value, which is wrong.  DV special-cases primitive fields which have both `+k8s:optional` and `+default=<zeroval-for-type>` and says "there's nothing I can do to validate this - it's just for documentation" (there's no way to tell if a user specified 0 or did not specify anything at all).  But we do not handle `required` + `default=<zero>` specially.  Protobuf elides those from the wire (and storage).  We could mark such fields as `optional` in protobuf, which turns them into pointers, but that seems awkward.  Or we could fix DV to handle required+default the same way as optional. @juli4n

PR #1631, opened 2026-09-13
Glutton Benchmarking Enhancements

maxsmythe:
Benchmarking: multi-actor glutton VUs, live window, resume retries, client-side latency
    
    Glutton actors per VU. Each VU creates --actors-per-user actors on
    startup and cycles through them round-robin, so one goroutine can drive
    many mostly-idle actors. runner.py forwards the flag to boomer-glutton.
    A crashed actor stays crashed for the run (ateapi never rehabilitates
    it), so it is marked on the first Aborted "crashed" error, skipped from
    then on, and counted in a CrashCount stat.
    
    Live window. --min/--max-live-time (default 0-0) set how long an actor
    stays resumed between its first ping and the suspend. Up to
    --max-pings-per-wake pings (default 1) run inside it, spaced 0.2-1.0s
    apart. The wait window (--min/--max-wait-time) remains the gap between
    one actor's suspend and the VU's next resume, and every return path
    sleeps it, so a failing startUser, resume, or crashed actor does not
    spin on boomer's zero-delay re-entry. With the defaults the cycle is
    resume, ping, suspend, wait: the same shape as before.
    
    Resume retries. ResumeActor retries ateapi's transient "concurrent
    update conflict" Aborted up to five times with a 50ms backoff, inside
    the timed call, so the conflict no longer shows up as a failure.
    
    Client-side latency. Every gRPC row in the locust stats now reports
    client wall clock, which covers retries, queueing, and the network. The
    server's elapsed-time trailer stays on the trace span only.
    
    Worker robustness. The router HTTP client keeps up to 10000 idle
    connections per host so each VU reuses its connection across wakes. A
    failed dynconfig fetch after the first successful one keeps the last
    fetched values instead of exiting the worker.
    
    Orchestrator. util.run logs the duration of each shell command.

- [ ] Tests pass

### DIFF
--- a/internal/benchmarking/boomer/dynconfig/dynconfig.go
+++ b/internal/benchmarking/boomer/dynconfig/dynconfig.go
@@ -47,8 +47,10 @@ const (
 // Config is the dynamic-mutable subset of boomer's behavior. Holder swaps
 // it atomically so task goroutines read a consistent snapshot.
 type Config struct {
-	MinWait          time.Duration
-	MaxWait          time.Duration
+	MinWait          time.Duration // gap between one actor's suspend and the VU's next resume, lower bound
+	MaxWait          time.Duration // upper bound of the same gap
+	MinLive          time.Duration // time a GluttonUser actor stays resumed between its first ping and suspend, lower bound
+	MaxLive          time.Duration // upper bound of the live window; zero (the default) suspends right after the ping
 	TraceProbability float64
 	DurDirFileSize   int64  // bytes
 	ResumeMode       string // ResumeModeExplicit | ResumeModeImplicit
@@ -57,6 +59,7 @@ type Config struct {
 	MemTarget        string // resident RAM the GluttonUser fills via WriteRAM, suffixed (e.g. "2Gi"); "" disables
 	MemChurn         string // RAM re-randomized in place each cycle via WriteRAM rotate, suffixed (e.g. "64Mi"); "" disables
 	MemRead          string // RAM walked (one byte per page) via ReadRAM after each resume, suffixed (e.g. "1Gi") or "all"; "" disables
+	MaxPingsPerWake  int    // cap on pings a GluttonUser sends during one resume/suspend cycle; values < 1 read as 1
 }
 
 // Holder lets readers Load() the current Config and writers Store() a new
@@ -90,6 +93,8 @@ type payload struct {
 	TraceProbability *float64 `json:"trace_probability"`
 	MinWaitTime      *float64 `json:"min_wait_time"`
 	MaxWaitTime      *float64 `json:"max_wait_time"`
+	MinLiveTime      *float64 `json:"min_live_time"`
+	MaxLiveTime      *float64 `json:"max_live_time"`
 	DurDirFileSize   *float64 `json:"durdir_file_size_bytes"`
 	ResumeMode       *string  `json:"resume_mode"`
 	DurDirReadMode   *string  `json:"durdir_read_mode"`
@@ -97,6 +102,7 @@ type payload struct {
 	MemTarget        *string  `json:"mem_target"`
 	MemChurn         *string  `json:"mem_churn"`
 	MemRead          *string  `json:"mem_read"`
+	MaxPingsPerWake  *float64 `json:"max_pings_per_wake"`
 }
 
 // Parse decodes a JSON blob (typically from a CLI flag) and merges its
@@ -154,6 +160,15 @@ func (c Config) Validate() error {
 	if c.MaxWait < c.MinWait {
 		return fmt.Errorf("max_wait_time (%v) cannot be less than min_wait_time (%v)", c.MaxWait, c.MinWait)
 	}
+	if c.MinLive < 0 {
+		return fmt.Errorf("min_live_time cannot be negative: %v", c.MinLive)
+	}
+	if c.MaxLive < 0 {
+		return fmt.Errorf("max_live_time cannot be negative: %v", c.MaxLive)
+	}
+	if c.MaxLive < c.MinLive {
+		return fmt.Errorf("max_live_time (%v) cannot be less than min_live_time (%v)", c.MaxLive, c.MinLive)
+	}
 	if c.TraceProbability < 0 || c.TraceProbability > 1 {
 		return fmt.Errorf("trace_probability must be between 0.0 and 1.0, got: %f", c.TraceProbability)
 	}
@@ -169,6 +184,9 @@ func (c Config) Validate() error {
 	if c.DurDirReadMode != "" && c.DurDirReadMode != ReadModeData && c.DurDirReadMode != ReadModeDigest {
 		return fmt.Errorf("invalid durdir_read_mode %q: must be %q or %q", c.DurDirReadMode, ReadModeData, ReadModeDigest)
 	}
+	// MaxPingsPerWake < 1 is treated as 1 at read time (see iterate() in
+	// glutton/lifecycle.go), so Config's zero value stays usable — no
+	// validate rejection here.
 	// MemTarget, MemChurn, and MemRead are passed to glutton verbatim
 	// (MemRead's "all" excepted, which the driver maps to an empty
 	// whole-array walk), which owns the parse; invalid values fail loudly
@@ -190,6 +208,12 @@ func (p payload) merge(current Config) Config {
 	if p.MaxWaitTime != nil {
 		out.MaxWait = time.Duration(*p.MaxWaitTime * float64(time.Second))
 	}
+	if p.MinLiveTime != nil {
+		out.MinLive = time.Duration(*p.MinLiveTime * float64(time.Second))
+	}
+	if p.MaxLiveTime != nil {
+		out.MaxLive = time.Duration(*p.MaxLiveTime * float64(time.Second))
+	}
 	if p.DurDirFileSize != nil {
 		out.DurDirFileSize = int64(*p.DurDirFileSize)
 	}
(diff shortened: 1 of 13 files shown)


internal/benchmarking/boomer/glutton/lifecycle.go
```diff
@@ -215,26 +338,69 @@ func (u *gluttonUser) create(ctx context.Context) error {
...
+			backoff = resumeMaxBackoff
+		}
+		return lastErr
 	})
 	if err != nil {
+		// ateapi reports a crashed actor as codes.Aborted with "crashed" in
+		// the message (see workflow_resume.go). Mark the user so iterate()
+		// stops touching it, and surface a CrashCount tick so operators can
+		// see the crash total in the locust stats table.
+		if s, ok := status.FromError(err); ok && s.Code() == codes.Aborted && strings.Contains(s.Message(), "crashed") {
```
sairajp-rewind, line 385: Can we make this a bit more robust? This part classifies every ResumeActor failure.  Could we check the state directly instead: on any resume failure, GetActor and test Status.State == ACTOR_STATE_CRASHED? One extra RPC on an already-failing path, and it survives changes to ateapi's error text.

internal/benchmarking/boomer/dynconfig/dynconfig.go
```diff
@@ -272,49 +299,58 @@ func StartPoll(
...
+// sends a spawn message for every ramp step, so a long ramp fetches once per
+// second. `onError` is invoked when a fetch fails, with `fetched` true if an
+// earlier spawn fetch succeeded: the holder then still has a value from the
+// master, and the caller can keep running on it. With `fetched` false the
+// worker has only its command-line defaults, and callers typically exit.
 // Returns an error if the event subscription itself fails (handler signature
 // mismatch), which is a programmer error and should be treated as fatal too.
-func SubscribeSpawn(url string, holder *Holder, sampler ProbabilityUpdater, fetchTimeout time.Duration, onError func(error)) error {
+func SubscribeSpawn(url string, holder *Holder, sampler ProbabilityUpdater, fetchTimeout time.Duration, onError func(err error, fetched bool)) error {
+	var fetched atomic.Bool
```
sairajp-rewind, line 340: fetched only records spawn-path successes, but StartPoll writes to the same Holder. If a poll lands first and the first spawn fetch then fails, we exit at saying the worker has only its command-line defaults — when it actually has the master's values. Could this bit live on the Holder instead, so either fetcher sets it?

internal/benchmarking/boomer/glutton/lifecycle.go
```diff
@@ -150,41 +230,84 @@ func (r *taskRuntime) startUser(ctx context.Context) (*gluttonUser, error) {
 func (r *taskRuntime) shutdown(ctx context.Context) {
 	r.users.Range(func(_, val any) bool {
 		u := val.(*gluttonUser)
-		if u.actorRunning {
-			u.suspend(ctx)
+		for _, a := range u.actors {
+			if a.actorRunning {
+				a.suspend(ctx)
+			}
+			a.delete(ctx)
```
sairajp-rewind, line 241: This is linear on both ends, VUs one at a time, actors within each VU one at a time. With --actors-per-user that's #VUs × N × 2 RPCs against the fixed 60s budget in boomer-worker/main.go (and runner.py hard-kills at 90s). The actors we don't delete keep holding workers. Could we fan out over actors with bounded concurrency, and scale the timeout with the actor count?

internal/benchmarking/boomer/glutton/lifecycle.go
```diff
@@ -215,26 +338,69 @@ func (u *gluttonUser) create(ctx context.Context) error {
...
+		var backoff time.Duration // 0 → first retry runs immediately
+		var lastErr error
+		for range resumeMaxAttempts {
+			_, lastErr = u.cfg.APIStub.ResumeActor(callCtx, &ateapipb.ResumeActorRequest{
+				Actor: u.ref(),
+			}, grpc.Trailer(tr))
+			if lastErr == nil {
+				return nil
+			}
+			if !isConcurrentUpdateConflict(lastErr) {
```
sairajp-rewind, line 365: ateapi already retries these server-side (ensureWorkerAssigned runs a 12-step backoff capped at ~3s (workflow_resume.go)) before it surfaces Aborted at all. So by the time we see a conflict here that budget is already spent, and retrying up to 5 more times inside the timed call can stack ~15s into a single ResumeActor sample. 
Won't that bloat the latency we're trying to measure? Should the retried resumes be reported separately, so the conflict cost stays visible instead of folding into p99?

internal/benchmarking/boomer/glutton/lifecycle.go
```diff
@@ -215,26 +338,69 @@ func (u *gluttonUser) create(ctx context.Context) error {
...
+				return lastErr
+			}
+			if backoff > 0 {
+				select {
+				case <-time.After(backoff):
```
sairajp-rewind: Should we jitter this? It's a flat 50ms for every attempt, so VUs contending on the same worker retry in lockstep rather than spreading out.

internal/benchmarking/boomer/glutton/lifecycle.go
```diff
@@ -215,26 +338,69 @@ func (u *gluttonUser) create(ctx context.Context) error {
...
+		return lastErr
 	})
 	if err != nil {
+		// ateapi reports a crashed actor as codes.Aborted with "crashed" in
+		// the message (see workflow_resume.go). Mark the user so iterate()
+		// stops touching it, and surface a CrashCount tick so operators can
+		// see the crash total in the locust stats table.
+		if s, ok := status.FromError(err); ok && s.Code() == codes.Aborted && strings.Contains(s.Message(), "crashed") {
+			u.crashed = true
+			bmetrics.RecordFailure("actor", "CrashCount", userClass, 0, "actor entered ACTOR_STATE_CRASHED")
```
sairajp-rewind, line 387: tracedCall has already recorded this error as a ResumeActor failure by the time we get here, so a crash reports twice — once as the RPC failure, once as CrashCount.

internal/benchmarking/boomer/dynconfig/dynconfig.go
maxsmythe: Not sure there is a clear best answer to this question.

In the case of a headless run, the flags are pre-provided and what you are saying makes sense.

In the case of a UI-driven run, the worker will be live long before the test parameters are provided, so SubscribeSpawn() is the only way to ensure the intended values are set.

In the case of a run where the values change over time (A second run from a UI-driven run, or a non-standard load shape), then these values may change over time and the test should fail because the parameters are inaccurate.

I don't think there is a "best" behavior.

The main thing I want to avoid is an un-impactful, transient error causing a long test run to fail.

Probably the best thing to do is to eventually revisit dynamic config (which needs overhauling anyway just to avoid proliferation of parameters), and try to find a more holistic approach (e.g. some way of identifying if a config is stale).

In the interim, this minimizes the risk of starting a test with a stale config while avoiding unnecessary crashes (with known edge cases).

internal/benchmarking/boomer/glutton/lifecycle.go
maxsmythe: There is a more fundamental issue here -- shutdown is actually never called until the process is literally torn down. We cannot solve that issue until boomer is modified to support the concept of a user.

The original code also had this edge case (you could hit the same problem by creating large numbers of users, this just decouples actor count from user count slightly).

For truly large tests (which are run by benchmarking/automation), the leak is harmless because substrate is torn down between each test.

Finally, trying to be "smart" and guess a timeout is probably not a great idea, since scaling depends on specific hardware. General pattern is to let users define a custom timeout, which would be beyond the scope of this PR (but not a bad idea).

internal/benchmarking/boomer/glutton/lifecycle.go
maxsmythe: I don't think this is true? Usually this points to an error resulting from write conflicts due to conflicting writes from different callers, meaning this caller is acting on stale state and it's up to caller to retry (drawing this from an K8s `resourceVersion` analogy).

https://github.com/agent-substrate/substrate/blob/94f2285f923f25e9ebd5fae47fbfe58af77fc788/cmd/ateapi/internal/controlapi/actor.go#L344-L357

internal/benchmarking/boomer/glutton/lifecycle.go
maxsmythe: Makes sense. I debated to jitter or not to jitter, but generally better to jitter. Only argument against it is that generally concurrent update conflicts are not about a rush of requests happening in lockstep, but different systems trying to update the same value at the same time. One request will always win, then the remaining systems (N - 1 of them) retry until all systems have succeeded, whereas jitter is usually about retry storms.

No harm in adding a bit of jitter though.

internal/benchmarking/boomer/glutton/lifecycle.go
maxsmythe: That is fine. The idea is to figure out how many actors have been removed from the roster.

internal/benchmarking/boomer/glutton/lifecycle.go
maxsmythe: Unfortunately I don't think we can without a change to ATE API.

"Aborted" state covers happy-path errors too, like "concurrent update conflict". We would not be calling this on just crashed actors, we would be calling it for ALL request failures, which could cause write amplification problems.

Ideally there would be some kind of stable error identifier that could be used to programatically identify a crashed actor (hey @juli4n LMK what you think about error classification) .

Another option could be to use a constant, but there is string substitution here.

Searching for "crashed" in the string is the least-bad option.

internal/benchmarking/boomer/glutton/lifecycle.go
maxsmythe: Added jitter

sairajp-rewind:
lgtm

sairajp-rewind (approved):

PR #1632, opened 2026-09-13
Add the option to install a large-cluster manifest and cordon control plane to dedicated machines

maxsmythe:
This change allows users to install Substrate on larger clusters. It adds a `--cluster-size` flag to enable more t-shirt-style sizing in the future to accommodate clusters of different sizes.

It also adds a --cordon-control-plane flag that allows taints/tolerances/antiaffinity/node labels to have each control plane element run on its own dedicated machine.

- [ ] Tests pass

### DIFF
--- a/hack/install-ate.sh
+++ b/hack/install-ate.sh
@@ -209,6 +209,31 @@ wait_for_pool_rollout_fatal() {
   fi
 }
 
+# kubectl_delete_tolerant runs `kubectl delete "$@"` and swallows the "no
+# matches for kind ..." error class — teardown often runs against a cluster
+# where an earlier upgrade already removed a CRD that manifests/ate-install
+# still references (e.g. SandboxConfig), and errexit would otherwise abort
+# the rest of the sweep. Real errors (any `error:`/`Error` line whose
+# message is NOT the missing-kind one) still fail.
+kubectl_delete_tolerant() {
+  local err_tmp status
+  err_tmp=$(mktemp)
+  status=0
+  run_kubectl delete "$@" 2>"${err_tmp}" || status=$?
+  cat "${err_tmp}" >&2
+  if (( status == 0 )); then
+    rm -f "${err_tmp}"
+    return 0
+  fi
+  local other
+  other=$(grep -E '^(error|Error)' "${err_tmp}" | grep -v 'no matches for kind' || true)
+  rm -f "${err_tmp}"
+  if [[ -n "${other}" ]]; then
+    return "${status}"
+  fi
+  return 0
+}
+
 run_kubectl_ate() {
   local _start_ns
   _start_ns=$(date +%s%N)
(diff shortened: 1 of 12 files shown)


annapendleton (approved):

hack/install-ate.sh
```diff
@@ -79,6 +94,10 @@ function usage() {
   echo "  --delete-all                           Delete core system and all registered demos"
   echo "  --atenet-router=envoy|agentgateway     Select the ingress and egress dataplane (default: envoy)"
   echo "  --podcert-workers-per-signer N         Concurrent workers per podcertificate-controller signer (default: 1)"
+  echo "  --cluster-size size0|size10            Cluster size profile (default: size0). \"size10\" assumes a dedicated postgres node"
+  echo "  --cordon-control-plane                 Pin each control plane pod to its own node: assumes a pool labeled and tainted"
+  echo "                                         ate.dev/workloadType=ate-control-plane:NoSchedule with one node per pod (7 at the"
```
BenTheElder, line 103: nit: we need to get this into the docs somewhere, I nearly had https://github.com/agent-substrate/substrate/pull/1651/changes tolerate all taints for atelet.

hack/install-ate.sh
```diff
@@ -79,6 +94,10 @@ function usage() {
   echo "  --delete-all                           Delete core system and all registered demos"
   echo "  --atenet-router=envoy|agentgateway     Select the ingress and egress dataplane (default: envoy)"
   echo "  --podcert-workers-per-signer N         Concurrent workers per podcertificate-controller signer (default: 1)"
+  echo "  --cluster-size size0|size10            Cluster size profile (default: size0). \"size10\" assumes a dedicated postgres node"
```
BenTheElder, line 101: what do these names mean? this doesn't seem intuitive?

maxsmythe: I was discouraged from "small" and "large", so "size0" is default (smallest) size. "size10" is larger. There is room for "size5", etc...

hack/install-ate.sh
maxsmythe: added to achitecture.md

hack/install-ate.sh
BenTheElder: I see. This feels unintuitive, and just "small" / "large" by another name ...

Maybe we should just require you to pass custom resizing patches ...?

Can defer to a follow-up to unblock work

manifests/ate-install/postgres/postgres.yaml
```diff
@@ -207,7 +207,7 @@ spec:
         # down to fit a CI runner.
         resources:
           requests:
-            cpu: "2"
+            cpu: "4"
             memory: "1Gi"
```
BenTheElder, line 211: is 1Gi request / 2GI limit memory reasonable with 4 CPU?

BenTheElder:
cc @bowei

hack/install-ate.sh
maxsmythe: Yeah, it's  essentially "small" and "large", but with the ability for more gradations later if necessary.

It looks like most of it could be a kustomize file, with the Postgres connection string being the one exception.

One good thing about "sizing"-type flags is that they should make it transparent when/if manifests change later.

Ultimately I'm guessing we'll need to have some generic way of customizing the footprint during install for production use cases?

manifests/ate-install/postgres/postgres.yaml
maxsmythe: Ooops, leftover from experimentation. Reverted this file.

maxsmythe:
@bowei re-worked it to work with the golang-only install process

msau42 (dismissed):
This is ok from a testing/dev perspective, but I do anticipate we'll need something more customizable for production.

maxsmythe:
100% WRT needing something more customizable for production.

msau42 (approved):

PR #1638, opened 2026-09-14
Record actor state changes from ateapi

krisztianfekete:
# Record actor state changes from ateapi

ateapi now writes a log record every time an actor changes state.

## Before

Nothing tells you what state an actor is in. Metrics can't carry actor identity, and the control plane samples traces at 10%, so most transitions leave nothing behind at all. If you want to know whether an agent is running, paused or suspended, there's nowhere to look.

## After

One filter:

```text
ate.actor.uid="8f2a1c4e6b0d47f1" AND ate.actor.state!=""
```

Last record wins. That's the state it's in now, and the timestamp is when it got there.

The record:

```json
{
  "time": "…",
  "level": "INFO",
  "msg": "Actor state changed",
  "ate.atespace": "ate-demo-counter",
  "ate.actor.name": "counter-1",
  "ate.actor.uid": "8f2a…",
  "ate.template.atespace": "ate-demo-counter",
  "ate.template.name": "counter",
  "ate.actor.operation.name": "suspend",
  "ate.actor.state": "suspended"
}
```

## The two new keys

`ate.actor.state` is the `ateapipb.ActorState` values lowercased, so the log wording and the state machine can't drift apart. 

`ate.actor.operation.name` says what caused the change. The state on its own doesn't tell you: an actor lands in `suspended` from a suspend and `paused` from a pause, and only one of those gives the worker back.

`Actor crashed` picks up the same state key. A crash is a state change too, and it's the one an actor can reach without any operation finishing.

## Notes for review

**Why ateapi.** It owns the state machine. It also sees transitions that never reach a worker, like deleting an actor that was already suspended, or a resume that fails in the scheduler.

**It can't log a state the store never had.** The record goes out after the store commit, and every state commit has a version precondition, so a losing writer in a concurrent update writes nothing.

- [x] Tests pass


JeffLuoo (commented):
Have you looked at [OTel entity events](https://opentelemetry.io/docs/specs/otel/entities/entity-events/)? This record is effectively an `entity.state` stream for `ate.actor` — same identity/description/relationship split, same last-write-wins contract. **Not a blocker for this PR** given entity event is still in development stage.

Two comments - 

- There's no terminal record. `deleting` is the last state a deleted actor ever reports, so "gone" and "stuck deleting" look identical to a consumer. Can we log the actual delete as well?

- Transition-only emission has no cold-start or expiry story: an actor that hasn't transitioned inside the retention window has no state, and a missed record is indistinguishable from steady state. Just document it somewhere.

krisztianfekete:
> Have you looked at [OTel entity events](https://opentelemetry.io/docs/specs/otel/entities/entity-events/)? This record is effectively an `entity.state` stream for `ate.actor` — same identity/description/relationship split, same last-write-wins contract. **Not a blocker for this PR** given entity event is still in development stage.

Yes, that's exactly the what I am designing this around. `ate.actor.uid` identifies, everything else describes. So we can add `entity.state` in later as second record with `entity.id` being the uid and the rest as description. I checked their state after their recent blog a couple of week ago, but I think it's still early to go all in.

> * There's no terminal record. `deleting` is the last state a deleted actor ever reports, so "gone" and "stuck deleting" look identical to a consumer. Can we log the actual delete as well?
> * Transition-only emission has no cold-start or expiry story: an actor that hasn't transitioned inside the retention window has no state, and a missed record is indistinguishable from steady state. Just document it somewhere.

Thanks, added these!

cmd/ateapi/internal/controlapi/workflow.go
```diff
@@ -67,6 +69,23 @@ func markSkipped(ctx context.Context, reason string) {
...
+// logActorStateChanged records an actor state change. The last record for an
+// actor's uid is the state it is in now.
+//
+// Call it after UpdateActor returns, not inside the mutate closure, which can be
+// retried. Every state commit carries a version precondition, so a call that
+// returns is the one that made the change.
+//
+// Crashes go through logActorCrashed instead, so read the state off
+// ate.actor.state rather than off the message.
+func logActorStateChanged(ctx context.Context, actor *ateapipb.Actor, opName, state string) {
```
JeffLuoo: I think "creation" should also be covered. When an actor is created, it holds a suspended stage: https://github.com/agent-substrate/substrate/blob/main/cmd/ateapi/internal/controlapi/actor.go#L112

An actor created and never resumed has no state at all in the stream — not "unknown since retention", genuinely never emitted, even with infinite retention. If we have creation event, it can at least record something in logs for this actor.

My understanding of this logs like "logs for all lifecycle of an actor".

JeffLuoo: Regarding the state field in the parameters.

https://github.com/agent-substrate/substrate/blob/44f4000c8ddf1bc2b5e69138c2658c27616a8cec/internal/ateattr/ateattr.go#L285-L300

The function takes the state as a literal, but `storedActor` is the committed record and already carries it. That means each of the eight call sites hand-pairs a constant with whatever the mutate closure assigned, and nothing catches a mismatch. A wrong literal produces a record that confidently reports a state the store never held.

ateattr already has the pattern in `SnapshotScopeValue` as attached above. Can you update the function to grep the state from `actor *ateapipb.Actor`?

JeffLuoo (changes requested):
Thanks for working on it, just a few more comments.

cmd/ateapi/internal/controlapi/workflow.go
krisztianfekete: Added. `CreateActor` logs now with `operation=create`, reporting suspended since that's what a new actor actually is.

krisztianfekete: Added `ActorStateValue` so the helper reads state off the committed record. Only the delete-terminal one still names it by hand, since the row is gone by then.

JeffLuoo (approved):

Issue #1640, opened 2026-09-14
Actor resource telemetry: adapt to multi-actor workers


baizhenyu:
## What

Adapt the actor resource-telemetry stack (#550: `GetWorkloadStats` / `GetActiveWorkloadStats` on both ateom runtimes, the atelet stats sweep with template-level metrics and per-actor usage events) to multi-actor workers (#1266).

The stack was built on the one-actor-per-worker state machine, and it says so out loud: the discovery response's sample slot is singular because "an ateom serves one actor at a time" (`internal/proto/ateompb/ateom.proto`, `GetActiveWorkloadStatsResponse`). #1266 phase 3 retires that premise.

It also raises the stakes: once workers host N actors, worker-pod-level metrics (metrics-server, `kubectl ate top workers`) stop being a usable proxy for any actor's usage at all. This pipeline becomes the only actor-granular source, so it should be multi-actor-clean before workers report capacity > 1.

## What breaks, and what deliberately does not

**The one wire break: `GetActiveWorkloadStatsResponse` arity.** `oneof result { sample | no_sample_reason }` cannot answer for N actors. The sweep (#961) is a merged caller and atelet (DaemonSet) skews against `ateomImage`s baked per pool — but the stance here is to **accept the skew window and break the shape in place**: telemetry degrades during a mixed-version rollout and recovers when it converges; that is not worth a dual-population protocol.

**Everything else survives by design, with small adaptations:**
- The keyed `GetWorkloadStats(actor_uid)` is untouched by pluralization — uid-addressed, per-actor codes stay correct — and its cost advantage becomes real (one actor vs. sweeping the whole worker).
- The lifecycle first/final sample bracket (the open #896 follow-up; #1206 merged periodic-only) is unaffected either way: its agreed direction — ateom takes the final sample inside CheckpointWorkload and echoes it in the response — is per-actor by construction.
- The atelet sweep's discovery (one socket dir per worker pod) and attribution (echoed identity only) are unchanged; CPU-delta state is already keyed by actor uid, so it is multi-actor-correct as-is.
[...]

PR #1644, opened 2026-09-14
ateom: pluralize the discovery read for multi-actor workers

baizhenyu:
Step 1 of #1640 (adapting the actor resource-telemetry stack to multi-actor workers, #1266): the discovery read's wire shape breaks in place, ahead of the runtime work, while the capacity cap still holds every list to at most one entry.

## Contract

`GetActiveWorkloadStatsResponse` becomes:

```proto
message GetActiveWorkloadStatsResponse {
  repeated WorkloadStatsSample samples = 1;
}
```

The `oneof {sample | no_sample_reason}` and the `NoSampleReason` enum are deleted outright — no `reserved` tombstones, per the accepted-skew stance below.

* **Available** = the empty list. "Available" stopped being a *reason*; it is the absence of workloads.
* **Not measurable yet** (boot, restore, teardown, a lifecycle transition underneath the lock-free read) = a **pending entry**: full attribution + sandbox class with `source = STATS_SOURCE_UNSPECIFIED`, whose "not measured" meaning `WorkloadStatsSample` already defines. It became a per-workload statement because with N workloads it can't be a response-level one — and it closes an old gap: a workload that dies during boot is attributable to a blind caller instead of anonymous.
* The keyed `GetWorkloadStats` is untouched: uid-addressed, per-actor codes correct at any occupancy.

## Consumers

The atelet sweep folds every entry of every response; nothing in it assumes at most one. Pending entries contribute nothing anywhere — not to the aggregates (`sampled_actors` keeps meaning "measured"), not to the CPU delta baselines, and no usage event — exactly matching the old no-sample behavior, so metrics and events are byte-identical before and after this change on a capacity-1 fleet.

## Version skew: broken

There is no dual population, no fallback decode, no rollout gate (per the decision recorded on #1640). Given that this is prior to GA, I recommend making the breaking change in favor of code simplicity..

## Testing

* Both runtimes: available → empty list; mid-boot → one pending entry asserted in full (attribution, class, `observed_at` set); transition-mid-read → pending entry for the new occupant / empty when the slot cleared; the discovery sample stays byte-identical to the keyed read's against the same fixture.
* atelet: pending entries skipped for aggregates, baselines, and events; multi-entry folding exercised by the loop shape.
* `verify-all.sh` clean end to end (includes proto regen check and the metrics registry).

Part of #1640, toward #1266.


git286:
> so an old atelet sees a multi-actor worker as its last sample

Is it true? When the same message field shows up twice, protobuf merges the second into the first instead of replacing it and so the data can be mixed actually

baizhenyu:
> > so an old atelet sees a multi-actor worker as its last sample
> 
> Is it true? When the same message field shows up twice, protobuf merges the second into the first instead of replacing it and so the data can be mixed actually

I don't think so. I would just assume version skew would break the gRPC. Since this is before public GA, so I did not really handle the breaking changes in a graceful way.

git286 (approved):

PR #1652, opened 2026-09-15
Add minimal prow-github-actions setup

ahmedtd:
DO_NOT_MERGE until we figure out a good way forward with Zizmor flagging pull_request_target.

Add a minimal prow-github-actions setup.

It should support labeling, as well as automatic PR submission.  This will be pretty janky and slow until we have

I believe the automatic merging setup will be secure, since without OWNERS files, only org / repo members should be able to /lgtm or /approve


BenTheElder:
We could consider using the github merge queues for merge, but I think we really want the ability to `/retest`.

I only see zizmor failing on the unpinned action digests?

EItanya:
> We could consider using the github merge queues for merge

FWIW I've really liked this in kagent so far, but I agree that not being able to re-run is a seriously terrible UX. GH has no answer to that?

swiftdiaries:
Heavy +1, being able to do `/retest` and `/ok-to-test` is something that'd be really, really helpful

.github/workflows/pr-workflow.yaml
```diff
@@ -1,180 +1,54 @@
...
-  contents: read
 jobs:
   run-tests:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
-      with:
-        fetch-depth: 0
+      uses: actions/checkout@v3
```
stshrive, line 25: You might want something like this:

```
actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```

.github/workflows/pr-workflow.yaml
```diff
@@ -1,180 +1,54 @@
...
     - name: Setup Go
-      uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5.6.0
+      uses: actions/setup-go@v5
```
stshrive, line 27: `actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # v5.4.0`

.github/workflows/pr-workflow.yaml
```diff
@@ -1,180 +1,54 @@
...
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+# Adapted from https://github.com/cncf/prow-github-actions/blob/main/templates/workflow-templates/prow.yml
```
stshrive, line 16: This is missing something like this:

```
permissions:
  contents: read
  packages: read
  pull-requests: write
  issues: write
  checks: write
  statuses: write
```

Not sure these are the optimal configurations. Consider reviewing these for the needs here.

BenTheElder:
> FWIW I've really liked this in kagent so far, but I agree that not being able to re-run is a seriously terrible UX. GH has no answer to that?

Grant write access :(

Obviously we will have _more_ people with that over time but ... it's an extremely broad permission. A lot of contributors should ideally have permission to launch / rerun tests without having permission to merge yet ...

BenTheElder:
> Heavy +1, being able to do /retest and /ok-to-test is something that'd be really, really helpful

To be clear: I don't think this action supports that yet.

PR #1654, opened 2026-09-15
Remove an actor's directory when atelet terminates it

chw120:
Terminate ended by resetting the actor's directories: it empties the subdirectories it knows and recreates them for a next activation. But nothing recreates the actor — Terminate is the control plane's word that it no longer exists — and nothing else on the node reclaims the per-actor root. The root, the sandbox assets record and the runsc debug logs under it survived every actor that ever ran on the node.

`removeActorDirs` takes the root as well. It goes through `resetActorDirs` first for the care a blanket `RemoveAll` lacks: it refuses to proceed while a volume directory is still populated, so a failed unmount cannot become a deletion of the mount's contents, and it can remove a bundle upper dir carrying an image's read-only modes.

### Next

Terminate is almost never reached. `ensureAteletTerminated` returns early when the actor has no worker assignment, and every actor `DeleteActor` accepts has none: deletion is allowed only from SUSPENDED or CRASHED, and both clear the
assignment in the commit that enters them. On current main this changes the outcome only for `--any-state` deletes of a live actor.

It still has to land first — the node must be able to reclaim the directory before the control plane asks it to — and it needs no API change, so it need not wait on the other half's design. Teaching the node to notice actors the control plane no longer knows about is a follow-up.

- [x] Tests pass


cmd/atelet/main.go
```diff
@@ -1944,6 +1946,23 @@ func resetActorDirs(actorUID string) error {
...
+//
+// resetActorDirs runs first for the care a blanket RemoveAll lacks. It refuses
+// to proceed while a volume directory is still populated, so a failed unmount
+// cannot become a deletion of the mount's contents, and it can remove a bundle
+// upper dir carrying an image's read-only modes.
+func removeActorDirs(actorUID string) error {
+	if err := resetActorDirs(actorUID); err != nil {
+		return err
+	}
+	if err := os.RemoveAll(ateompath.ActorPath(actorUID)); err != nil {
```
zoez7, line 1960: Discussed offline. `resetActorDirs` is used by both suspend and resume to clean up the actor directory, where as the additional `os.RemoveAll(ateompath.ActorPath(actorUID))` here is to clean up the leaked directories when force deleting an actor.

PR #1658, opened 2026-09-15
(feat): Actor lifecycle events over otlp

krisztianfekete:
## What this does

ateapi writes a record every time an actor changes state. Until now those records only went to the pod's stdout, and nothing reads stdout. This sends the same records to a collector as OTLP log events.

Actor name and uid cannot be metric labels (too many values), and traces are sampled at 1%. So these records are the only way to answer "what state is this actor in, and since when".

## Changes

- `serverboot.InitLogging` sets up a LoggerProvider, next to the existing tracer and meter ones.
- New `internal/actorevent` package builds the log records.
- ateapi emits at the two places that already write the stdout records.
- Two event names: `ate.actor.state_changed` and `ate.actor.crashed`.
- Both names are registered in `docs/metrics/registry/events.yaml`, so `make verify` checks them.
- kind gets a logs pipeline and a count connector. The e2e suite reads the counts back.
- Docs updated. `otel-collector.md` said substrate has no LoggerProvider, which is no longer true.

## Opt-in

`OTEL_LOGS_EXPORTER` defaults to `none`. Only the kind overlay sets it to `otlp`. The base ConfigMap is untouched, so no deployed environment changes when this merges. 

## Notes on the design

- **No slog bridge.** Only two call sites emit these records, so emitting twice costs two lines. A bridge would also send every ateapi log over the wire, could not set the event name, and would loop, because SDK export errors are logged through slog.
- **Batching processor, not the simple one.** These records sit on the actor resume path. A processor that exports inside the emit call would add a blocking gRPC call there, so a slow collector would become control plane latency.
- **Two event names, not one per state.** `ate.actor.state` already says which transition happened. A crash gets its own name because it carries two extra attributes and a higher severity.
- **Both copies are kept on purpose.** No collector in this repo reads pod stdout, so nothing is duplicated today. `kubectl logs` keeps working. If a filelog agent is ever added, drop one of the two. The escape hatch is written down in `docs/metrics/substrate.yaml`.

## Dependencies

Adds `otel/log`, `otel/sdk/log` and `otlploggrpc`, all pinned at v0.20.0. That is the release that matches the pinned `otel v1.44.0`. v0.21.0 would pull the core modules to v1.45.0, which this change does not need. The logs API has a
v1.47.0 release candidate upstream, so it is on its way to stable.

## Testing

- Unit tests for the exporter resolver, the record builder, and both ateapi emit sites.
- The record builder test checks the attribute set matches what the event name declares, in both directions.
- The ateapi tests check the OTLP record carries the same attributes as the stdout record.
- Ran end to end on kind. Records arrive with the right event name, severity, attributes, and with trace context on the record's own fields rather than as attributes.
- Checked the off state too. With `OTEL_LOGS_EXPORTER` removed, the collector receives no log records and stdout is unchanged.

- [x] Tests pass


internal/actorevent/actorevent.go
```diff
@@ -0,0 +1,143 @@
...
+	"time"
+
+	"go.opentelemetry.io/otel/log"
+	"go.opentelemetry.io/otel/log/global"
+
+	"github.com/agent-substrate/substrate/internal/ateattr"
+)
+
+// ScopeName is how a consumer selects this stream.
+const ScopeName = "github.com/agent-substrate/substrate/internal/actorevent"
```
EItanya: Is this typically the repo, or the dns style name like `ate.dev`?

krisztianfekete:
@JeffLuoo, can you please take a look at this when you have some free cycles?

JeffLuoo:
Sure!

JeffLuoo:
Can you first create a bug to track the work and general discussion?

krisztianfekete:
> Can you first create a bug to track the work and general discussion?

Sure, wanted to open one for related topics, but I think we can keep these reviewed in parallel as this PR
- adds no new records as ateapi has written both to stdout for a while now, so this is only about the OTLP support, and
- is off everywhere by default as only the kind overlay sets OTEL_LOGS_EXPORTER

Opened the discussion here: https://github.com/agent-substrate/substrate/issues/1683

internal/actorevent/actorevent.go
```diff
@@ -0,0 +1,143 @@
...
+func BuildRecord(ev Event, t time.Time, attrs []slog.Attr) log.Record {
+	var rec log.Record
+	rec.SetEventName(ev.Name)
+	rec.SetTimestamp(t)
+	rec.SetSeverity(ev.Severity)
+	rec.SetBody(log.StringValue(ev.Body))
+
+	kvs := make([]log.KeyValue, 0, len(attrs))
+	for _, a := range attrs {
+		kvs = append(kvs, log.String(a.Key, a.Value.String()))
```
JeffLuoo: `log.String(a.Key, a.Value.String())` is lossless today, because ActorLogAttrs and FailureLogAttrs build only slog.String. The Keys test pins the key names, not the value kinds. So the first slog.Int or slog.Duration attribute gives a number on stdout and a string over OTLP, and no test fails. The package doc says the two copies cannot drift, and this is the one way they can. A type switch, or a test that asserts every attribute is string-kind, would help it.

manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -94,15 +108,28 @@ data:
           processors: [batch]
           exporters: [prometheus, debug]
 
-        # The count pipelines read the same receiver as the two pipelines
-        # above, and they only count. The connector declares MutatesData
-        # false, thus it changes nothing that reaches Jaeger or Prometheus.
+        # The actor lifecycle events. debug is the only exporter: kind has no
```
JeffLuoo: nit - "actor lifecycle events" is what we export today but we expect to have more events coming from the system. Let's make it general like "ate events" or something similar.

manifests/ate-install/kind/otel-collector.yaml
```diff
@@ -74,6 +74,20 @@ data:
...
+        # useful of the two anyway.
+        logs:
+          substrate_actor_state_changes:
+            description: Actor state change events received, by state.
+            attributes:
+              - key: service.name
+                default_value: unknown
+              - key: ate.actor.state
+                default_value: unknown
+
```
JeffLuoo, line 90: This reads the bare otlp receiver with no filter, and the connector at line 88 defaults ate.actor.state to unknown. The same collector doc says the logs pipeline also serves user workloads, so a demo actor that emits OTLP logs lands in `substrate_actor_state_changes{ate_actor_state="unknown"}`. The e2e is unaffected, because it reads named states. Worth a filter on the scope, or a description at line 84 that says what is actually counted. If we want to just track state change count, add a filter here.

internal/serverboot/logging.go
```diff
@@ -0,0 +1,135 @@
...
+// through the OTel logs API are affected; slog to stdout is always on.
+type LogsExporter string
+
+const (
+	// LogsExporterNone drops the records. It is the component default, so an
+	// environment that has not opted in changes nothing.
+	LogsExporterNone LogsExporter = "none"
+	// LogsExporterOTLP sends to OTEL_EXPORTER_OTLP_ENDPOINT.
+	LogsExporterOTLP LogsExporter = "otlp"
+)
```
JeffLuoo, line 41: Could we accept console here as a third value? Just easier to test change.. Or are you intentional not adding it?

JeffLuoo (changes requested):

internal/serverboot/logging.go
krisztianfekete: I intentionally left it out. We already write the same record to stdout so the console version would print the same thing in slightly different format which I think is more confusing than useful here.

As per seeing the OTLP format, the collector's debug exporter already dumps the full record, so unit tests can assert that directly.

internal/actorevent/actorevent.go
krisztianfekete: Thanks, good catch, this is now addressed.

manifests/ate-install/kind/otel-collector.yaml
krisztianfekete: Done.

manifests/ate-install/kind/otel-collector.yaml
krisztianfekete: This is now fixed by just dropping `default_value` on `ate.actor.state`. The connector drops items missing a key that has no default, so we don't even need a processor here. A real event with state unknown still has the key so that still counts.

internal/serverboot/logging.go
JeffLuoo: > As per seeing the OTLP format, the collector's debug exporter already dumps the full record, so unit tests can assert that directly.

I want to break it down into 2 parts:

1. Instrumentation using OTel SDK for logs / events - Yes. And we should make it freely configurable, including options to write logs to console (stdout).

2. Collection part: We shouldn't assume a collector will be deployed. Users can use any collection agent.

Hence my point is we shouldn't limit the configuration of OTel SDK, and provide users guidance to configure.

krisztianfekete: `stdout` already has the same information in `ate.*` format in JSON that is readable and parsable by any agent/collector. Adding console will result in the exact same data, duplicated in a slightly different format.

JeffLuoo: Thanks, I agree that a console exporter is not needed here: the existing structured stdout record remains available to kubectl logs and to a future filelog collector, while OTLP provides the schema-specific event form.

I do want to make one long-term invariant explicit: one lifecycle event should be constructed once, then projected to stdout and OTLP, rather than maintained as two parallel emissions at each call site. Reusing []slog.Attr is a good start, but slog.LogAttrs(...) and actorevent.Emit(...) still leave separate body/severity/emission paths that can drift as the event evolves.

Would it make sense to have actorevent own a single higher-level emission helper or canonical event record and derive both outputs from it? This is not a blocker for this PR, but I would like us to capture the invariant in the design or a follow-up issue so it does not get lost.

krisztianfekete: Sure, opened it here, can pick it up as a follow-up once this one is merged: https://github.com/agent-substrate/substrate/issues/1744

JeffLuoo (approved):

Issue #1550, opened 2026-09-09
kubectl ate for egresspolicy


LiorLieberman:
kubectl ate for egress policy

/cc @EItanya @bowei

PR #1659, opened 2026-09-15
kubectl-ate: add get and create egress-policy

ygao-g:
Part of #1550

This PR adds `kubectl ate get` and `kubectl ate create` support for `egress-policy`.
- `get` prints a table by default, or a bare JSON/YAML document with `-o`. 
- `create` consumes that same document. `-f -` reads stdin.
- Omitted `metadata` is filled from `--atespace` and the fixed name `default`; a manifest naming another atespace is rejected before any RPC. 
- `get` accepts exactly one actor for now; a follow-up adds several actors and a list document.

Recommend reviewing the four commits one at a time:
- printer and manifest decoder with a round-trip test, 
- `get` command, 
- `create` command, 
- then the README updates on their own.

- [x] Tests pass:
  - unit tests and `make verify`
  - https://github.com/ygao-g/substrate/pull/33 against this head on a local kind cluster;
  - Also tested the new `test-egress.sh` step on a local kind cluster; 
- [x] Appropriate changes to documentation are included in the PR.


cmd/kubectl-ate/internal/printer/printer.go
```diff
@@ -373,6 +373,25 @@ func PrintTagTo(out io.Writer, tag *ateapipb.Tag, format string) error {
 	return PrintTagsTo(out, []*ateapipb.Tag{tag}, format)
 }
 
+// PrintEgressPolicyTo prints one actor's egress policy. json and yaml emit the
```
bowei: would prefer printer_egress_policy.go instead of putting it in one giant printer.go

cmd/kubectl-ate/internal/cmd/create_egress_policy.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+var createEgressPolicyCmd = &cobra.Command{
+	Use:     "egress-policy <actor-name> -f <manifest>",
+	Aliases: []string{"egress-policies"},
+	Short:   "Create the egress policy of an actor",
+	Long: `Create the egress policy of an actor from a manifest file.
+
+The manifest is a single YAML (or JSON) document holding one ateapipb.EgressPolicy
+message in its protojson form, exactly as printed by
```
bowei: This text is weird and also not correct. It says the manifest is in YAML or JSON and then says it's protojson. 

"manifest is a YAML or JSON egress policy. metadata is inferred from the actor name and atespace. The Actor must already exist and have no existing policy object."

cmd/kubectl-ate/internal/cmd/create_egress_policy.go
```diff
@@ -0,0 +1,104 @@
...
+	actor     *ateapipb.ObjectRef
+	policy    *ateapipb.EgressPolicy
+	outputFmt string
+	out       io.Writer
+}
+
+func (r *CreateEgressPolicyRunner) Run(ctx context.Context) error {
+	created, err := r.creator.CreateActorEgressPolicy(ctx, &ateapipb.CreateActorEgressPolicyRequest{Actor: r.actor, EgressPolicy: r.policy})
+	if err != nil {
+		return fmt.Errorf("failed to create egress policy for actor %q: %w", r.actor.GetName(), err)
```
bowei: when we report the error we should always report both atespace and actor

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// policyName is the fixed name of an actor's egress policy.
+const policyName = "default"
+
+// egressPolicyFromManifest parses a single protojson-shaped YAML or JSON
+// document. Parsing is strict: unknown fields are an error. metadata.atespace
+// and metadata.name default to atespace and policyName.
+func egressPolicyFromManifest(data []byte, atespace string) (*ateapipb.EgressPolicy, error) {
+	jsonData, err := yaml.YAMLToJSON(data)
```
bowei: add comment // YAML is a superset of JSON so we can use the same parser.

cmd/kubectl-ate/internal/cmd/get_egress_policy.go
```diff
@@ -0,0 +1,94 @@
...
+	"google.golang.org/grpc/status"
+)
+
+var getEgressPolicyAtespaceFlag string
+
+var getEgressPolicyCmd = &cobra.Command{
+	Use:     "egress-policy <actor-name>",
+	Aliases: []string{"egress-policies"},
+	Short:   "Get the egress policy of an actor",
+	Long: "Get the egress policy of an actor. An actor has at most one egress policy; " +
```
bowei: Delete text starting with "An actor has at most..." this is not relevant.

bowei (commented):
Check that the help text makes sense and is relevant to the command. The help text should not talk about what Egress policy does, constraints etc.

cmd/kubectl-ate/internal/cmd/create_egress_policy.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2026 Google LLC
```
juli4n: Can you merge all commands for the EgressPolicy resource in a singel egress_policy.go file? See tag.go as an example. 

We need to consolidate other resources (actor, worker, etc), but this directory is getting fairly polluted and merging into a single file per resource is the direction we are going.

cmd/kubectl-ate/internal/printer/printer.go
ygao-g: Done, moved to printer_egress_policy.go.

cmd/kubectl-ate/internal/cmd/create_egress_policy.go
ygao-g: Reworded along the lines you suggested, plus a pointer to get -o yaml as the source of a valid manifest, similar to 'create actor-template' command.

cmd/kubectl-ate/internal/cmd/create_egress_policy.go
ygao-g: Done for create and get.

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Added.

cmd/kubectl-ate/internal/cmd/get_egress_policy.go
ygao-g: Removed. get now has only the short description, like the other get commands.

cmd/kubectl-ate/internal/cmd/create_egress_policy.go
ygao-g: Done. get and create now live in egress_policy.go with one test file.

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, fmt.Errorf("invalid YAML: %w", err)
+	}
+	if string(jsonData) == "null" {
+		return nil, fmt.Errorf("manifest is empty")
+	}
+	policy := &ateapipb.EgressPolicy{}
+	if err := protojson.Unmarshal(jsonData, policy); err != nil {
+		return nil, err
+	}
+	if policy.Metadata == nil {
```
juli4n, line 120: The logic to override metadata looks like belong to the caller of this function, not here. I think this should be just the parser? That would be consistent with `actorTemplateFromManifest`.

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,207 @@
...
+	return runner.Run(ctx)
+}
+
+var (
+	createEgressPolicyAtespaceFlag string
+	createEgressPolicyFilenameFlag string
+)
+
+var createEgressPolicyCmd = &cobra.Command{
+	Use:     "egress-policy <actor-name> -f <manifest>",
```
juli4n: The ability to pass a manifest and provide (atespace, name) to override it lhis looks handy for the flow where you want to copy the policy of an actor to another (otherwise you will need to store in a file, edit, call kubectl ate create). It's not consistent with ActorTemplates though, which have a similar feature. Maybe add a TODO in create_actor_template.go to add support for this as well?

juli4n:
I guess update will come in a follow up PR?

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Small correction to the reading: the flags do not override the manifest. The store looks up an egress policy  by ActorRef, the table's primary key is (atespace, actor_name), and the server requires metadata.name to be 'default'. On the client, --atespace and 'default' are filled in only when the manifest omits metadata, and a manifest that names a different atespace is rejected before dialing.

That means a same-atespace copy already works as a pipe: `get egress-policy src -o yaml | create egress-policy dst -f -`. Only a cross-atespace copy needs the manifest edited first. 

I opened issue #1680 for the retarget flow, since it needs more deiscussion and design calls (silent override vs. an explicit flag, and how create actor-template should take a name). 

WDYT?

ygao-g:
> I guess update will come in a follow up PR?

Yes. update and delete will come in following PRs

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done. egressPolicyFromManifest is now parse-only (YAML to JSON, reject an empty document, strict protojson), and the create runner then calls a new fillEgressPolicyMetadata for the defaulting logic.

juli4n: Do we really need that defaulting logic? Why not just let it fail if the metadata is not provided in the manifest?

cmd/kubectl-ate/README.md
```diff
@@ -232,6 +232,49 @@ kubectl ate create actor <actor-name> -a <atespace> --template <template-name> -
...
+kubectl ate get egress-policy <actor-name> -a <atespace>
+kubectl ate get egress-policy <actor-name> -a <atespace> -o yaml
+
+# Create an actor's egress policy from a manifest, or clone another actor's.
+kubectl ate create egress-policy <actor-name> -a <atespace> -f policy.yaml
+kubectl ate get egress-policy <src-actor> -a <atespace> -o yaml | \
+  kubectl ate create egress-policy <actor-name> -a <atespace> -f -
+
+# Delete an actor's egress policy; the actor keeps running with all egress denied.
+kubectl ate delete egress-policy <actor-name> -a <atespace>
```
haiyanmeng: This PR does not support `delete` command, right?

cmd/kubectl-ate/README.md
```diff
@@ -232,6 +232,49 @@ kubectl ate create actor <actor-name> -a <atespace> --template <template-name> -
...
+The manifest is one `EgressPolicy` in YAML or JSON; `metadata` may be omitted
+and server-managed fields are ignored, so the output of `get -o yaml` is a valid
+manifest as is.
+
+```yaml
+# policy.yaml: rules are evaluated in order; the first match wins.
+rules:
+- hostnames: {patterns: ["api.example.com"]}
+- cidrs: {cidrs: ["10.64.0.0/16"]}
+```
```
haiyanmeng: https://github.com/agent-substrate/substrate/pull/1656 adds a doc for Egress Policy rules and evaluation order. We could refer to that doc added in that PR once it is submitted.

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,217 @@
...
+	getter    EgressPolicyGetter
+	actor     *ateapipb.ObjectRef
+	outputFmt string
+	out       io.Writer
+	errOut    io.Writer
+}
+
+func (r *GetEgressPolicyRunner) Run(ctx context.Context) error {
+	policy, err := r.getter.GetActorEgressPolicy(ctx, &ateapipb.GetActorEgressPolicyRequest{Actor: r.actor})
+	if status.Code(err) == codes.NotFound {
```
haiyanmeng, line 156: Should we check whether the actor exists? Otherwise, we won't be able to differentiate `when an actor does not exist` from `an actor exists without an egress policy`.

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,217 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// policyName is the fixed name of an actor's egress policy.
+const policyName = "default"
+
+// egressPolicyFromManifest parses a single protojson-shaped YAML or JSON
+// document into an EgressPolicy. Parsing is strict: unknown fields are an
+// error. Metadata is returned as written; see fillEgressPolicyMetadata.
+func egressPolicyFromManifest(data []byte) (*ateapipb.EgressPolicy, error) {
```
haiyanmeng, line 68: what happens if `data` includes multiple egress policies separated by `---` lines?

cmd/kubectl-ate/README.md
ygao-g: Correct, and good catch. I have taken the `delete` lines out of the README

cmd/kubectl-ate/README.md
ygao-g: Agreed,  I have
trimmed the paragraph here down to what a CLI reader needs and left a
`TODO(#1550)` comment above the section to add the link to `docs/egress-policy.md`
once #1656 merges.

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Fixed. Now it errors when there are multiple non-empty yaml. PTAL

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Good catch!  The two are indistinguishable today, and the CLI could only tell them
apart by making a second `GetActor` RPC per actor, which I would rather not add to a
read path that is meant to loop over many actors.

I filed issue #1703 and am preparing a fix this on the server-side as a separate PR.
More specifically, make `GetActorEgressPolicy` return either `Actor <atespace>/<name> not found` or `EgressPolicy not found`,
with a `google.rpc.ResourceInfo` detail to name the missing resource. The CLI then fails when the actor is
missing and keeps the deny-all note when the actor exists without a policy.

WDYT?

haiyanmeng: Sounds good. Then let us add a comment pointing to the issue you filed, and update the error message to indicate the actor may not exist.

cmd/kubectl-ate/internal/cmd/egress_policy.go
haiyanmeng: Can you add test coverage for the case when `data` includes multiple egress policies separated by `---` lines?

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done, add a comment and changed error messages. An example: actor \"c1\" in atespace \"team-a\" has no egress policy or does not exist; an actor without a policy has all egress denied

cmd/kubectl-ate/internal/cmd/egress_policy.go
haiyanmeng: Please address this one

haiyanmeng (changes requested):

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Closed this by mistakes and thanks for opening it! The table in TestEgressPolicyFromManifest now has rows for
- two and three complete egress policies split by ---, 
- a case with an empty document between two policies,
- two JSON documents split the same way.

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,254 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/protojson"
+	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/yaml"
+)
+
+// policyName is the fixed name of an actor's egress policy.
+const policyName = "default"
```
bowei: move const to where this is used

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,254 @@
...
+
+// egressPolicyFromManifest parses a single protojson-shaped YAML or JSON
+// document into an EgressPolicy. Parsing is strict: unknown fields are an
+// error. Metadata is returned as written; fillEgressPolicyMetadata defaults it.
```
bowei: Remove text starting with "Metadata is returned..."

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,254 @@
...
+	if err := protojson.Unmarshal(jsonData, policy); err != nil {
+		return nil, fmt.Errorf("invalid EgressPolicy: %w", err)
+	}
+	return policy, nil
+}
+
+// manifestToJSON converts a YAML or JSON manifest to JSON. YAML is a superset
+// of JSON, so one reader parses both. The manifest must hold exactly one
+// non-empty document; the empty documents a leading or trailing "---" leaves
+// behind are skipped.
```
bowei, line 83: Returns a slice of the documents in the YAML file, in JSON format.

Why return slice if we require that there is only one?

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,254 @@
...
+			return nil, fmt.Errorf("invalid YAML: %w", err)
+		}
+		j, err := yaml.YAMLToJSON(doc)
+		if err != nil {
+			return nil, fmt.Errorf("invalid YAML: %w", err)
+		}
+		if string(j) == "null" {
+			continue
+		}
+		documents++
```
bowei: Do we want to parse all of the docs and then return an error, or just return an error if the parse contains more than one document?

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,254 @@
...
+		return jsonData, nil
+	default:
+		return nil, fmt.Errorf("manifest holds %d documents, expected one", documents)
+	}
+}
+
+// fillEgressPolicyMetadata sets metadata.atespace to atespace and
+// metadata.name to policyName when the manifest omits them. A manifest that
+// names another atespace or name is rejected, not retargeted.
+func fillEgressPolicyMetadata(policy *ateapipb.EgressPolicy, atespace string) error {
```
bowei: fill -> override

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,254 @@
...
+	case "":
+		policy.Metadata.Name = policyName
+	case policyName:
+	default:
+		return fmt.Errorf("manifest metadata.name %q must be %q", policy.Metadata.Name, policyName)
+	}
+	return nil
+}
+
+var getEgressPolicyAtespaceFlag string
```
bowei: can we consolidate all of the flag vars and policyCmd decls into one block. Or you can put them into egress_policy_cmd.go

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,254 @@
...
+type EgressPolicyGetter interface {
+	GetActorEgressPolicy(ctx context.Context, req *ateapipb.GetActorEgressPolicyRequest, opts ...grpc.CallOption) (*ateapipb.EgressPolicy, error)
+}
+
+// GetEgressPolicyRunner executes the get egress-policy command logic.
+type GetEgressPolicyRunner struct {
+	getter    EgressPolicyGetter
+	actor     *ateapipb.ObjectRef
+	outputFmt string
+	out       io.Writer
```
bowei: [nit] name this `stdout` and `stderr` if that is what these correspond to

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,254 @@
...
+
+The manifest is a YAML or JSON EgressPolicy, as printed by
+"kubectl ate get egress-policy <actor-name> -a <atespace> -o yaml".
+Its metadata may be omitted.`,
+	Args: cobra.ExactArgs(1),
+	RunE: runCreateEgressPolicy,
+}
+
+// EgressPolicyCreator abstracts CreateActorEgressPolicy RPC calls.
+type EgressPolicyCreator interface {
```
bowei: check all of your exported types and names to make sure they actually need to be exported.

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,254 @@
...
+	CreateActorEgressPolicy(ctx context.Context, req *ateapipb.CreateActorEgressPolicyRequest, opts ...grpc.CallOption) (*ateapipb.EgressPolicy, error)
+}
+
+// CreateEgressPolicyRunner executes the create egress-policy command logic.
+type CreateEgressPolicyRunner struct {
```
bowei: check this needs to be exported

cmd/kubectl-ate/internal/printer/egress_policy.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2026 Google LLC
```
bowei, line 1: I just noticed that the file name is `printer_...` and the package is also named printer. Delete the `printer_` prefix => `egress_policy.go`, `egress_policy_test.go`

cmd/kubectl-ate/README.md
```diff
@@ -232,6 +232,48 @@ kubectl ate create actor <actor-name> -a <atespace> --template <template-name> -
 kubectl ate delete tag <tag-name> -a <atespace>
 ```
 
+### Egress Policies
+
+<!-- TODO(#1550): link docs/egress-policy.md for the rule types and their evaluation order once that page exists. -->
+
+An actor has at most one egress policy, and one without a policy has all
```
bowei: Remove text from ", and one without..." to the end of the paragraph"

cmd/kubectl-ate/README.md
```diff
@@ -232,6 +232,48 @@ kubectl ate create actor <actor-name> -a <atespace> --template <template-name> -
...
+egress denied. `get` prints that no-policy state as a note on stderr and
+exits 0, so a loop over actors does not stop at the first one without a
+policy. The server does not yet tell a missing actor from an actor without
+a policy, so a mistyped actor name gets the same note.
+
+```bash
+# Show an actor's egress policy, or dump it as a document.
```
bowei: Change to 

```
# Get an actor's egress policy.
```

cmd/kubectl-ate/README.md
```diff
@@ -232,6 +232,48 @@ kubectl ate create actor <actor-name> -a <atespace> --template <template-name> -
...
+kubectl ate get egress-policy <actor-name> -a <atespace>
+kubectl ate get egress-policy <actor-name> -a <atespace> -o yaml
+
+# Create an actor's egress policy from a manifest, or clone another actor's.
```
bowei:
```
# Create an egress policy
```

cmd/kubectl-ate/README.md
```diff
@@ -232,6 +232,48 @@ kubectl ate create actor <actor-name> -a <atespace> --template <template-name> -
...
+kubectl ate create egress-policy <actor-name> -a <atespace> -f policy.yaml
+kubectl ate get egress-policy <src-actor> -a <atespace> -o yaml | \
```
bowei, line 250:
```
# Copy egress policy from another Actor
```

cmd/kubectl-ate/README.md
```diff
@@ -232,6 +232,48 @@ kubectl ate create actor <actor-name> -a <atespace> --template <template-name> -
...
+  kubectl ate create egress-policy <actor-name> -a <atespace> -f -
+```
+
+The manifest is one `EgressPolicy` in YAML or JSON; `metadata` may be omitted
+and server-managed fields are ignored, so the output of `get -o yaml` is a valid
+manifest as is.
+
+```yaml
```
bowei: Delete this snippet...

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: To clarify, the function errors out if there are multiple documents, otherwise it returns a single JSON document. Does this answer your question?

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Good catch! Now it returns an error as soon as a second non-empty document is seen

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done.

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done. The flag vars are in one block at the beginning, followed by the two command declarations.

cmd/kubectl-ate/README.md
ygao-g: Done. Thx for suggestions!

cmd/kubectl-ate/README.md
ygao-g: Done

cmd/kubectl-ate/README.md
ygao-g: Done

cmd/kubectl-ate/README.md
ygao-g: Done

cmd/kubectl-ate/README.md
ygao-g: done

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: done

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done. Unexported following types: 
- EgressPolicyGetter
- GetEgressPolicyRunner
- EgressPolicyCreator
- CreateEgressPolicyRunner

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done. Unexported following types: 
- EgressPolicyGetter
- GetEgressPolicyRunner
- EgressPolicyCreator
- CreateEgressPolicyRunner

cmd/kubectl-ate/internal/printer/egress_policy.go
ygao-g: Done

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,250 @@
...
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("invalid YAML: %w", err)
+		}
+		j, err := yaml.YAMLToJSON(doc)
+		if err != nil {
+			return nil, fmt.Errorf("invalid YAML: %w", err)
+		}
+		if string(j) == "null" {
```
bowei: This loop is kind of confusing to me.

Why not do this:

```golang
doc, err := reader.Read()
if err != nil {
  return fmt.Errorf("manifestToJSON: %w", err)
}
// Check that YAML has only one document by reading again.
_, err = reader.Read()
switch {
case errors.Is(err, io.EOF):
  // This is expected, we got one document.
case err == nil:
  return nil, errors.New("manifestToJSON: invalid, multidocument YAML")
default:
  return nil, err
}
jsondata, err := yaml.YAMLToJson(doc)
if err != nil {
  return nil, fmt.Errorf("manifestToJSON: could not convert to JSON: %w", err)
}
return jsondata, nil
```

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,250 @@
...
+		jsonData = j
+	}
+	if jsonData == nil {
+		return nil, fmt.Errorf("manifest is empty")
+	}
+	return jsonData, nil
+}
+
+// policyName is the fixed name of an actor's egress policy.
+const policyName = "default"
```
bowei: move this inside func where it is used

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,250 @@
...
+
+// overrideEgressPolicyMetadata sets metadata.atespace to atespace and
+// metadata.name to policyName when the manifest omits them. A manifest that
+// names another atespace or name is rejected, not retargeted.
+func overrideEgressPolicyMetadata(policy *ateapipb.EgressPolicy, atespace string) error {
+	if policy.Metadata == nil {
+		policy.Metadata = &ateapipb.ResourceMetadata{}
+	}
+	switch policy.Metadata.Atespace {
```
bowei: This is correct but is the intent to default if metadata is empty or to default each field if it's empty. Suggest for more clarity:

```golang
const defaultPolicyName = "default"
if policy.Metadata == nil {  
	policy.Metadata = &ateapipb.ResourceMetadata{}
}
// Default if empty
if policy.Metadata.Atespace == "" {
  policy.Metadata.Atespace = atespace
}
if policy.Metadata.Name == "" {
  policy.Metadata.Name = defaultPolicyName
}
// Check metadata values match
if policy.Metadata.Atespace != atespace || policy.Metadata.Name != defaultPolicyName {
  return // error
}

return nil
```

cmd/kubectl-ate/internal/cmd/egress_policy.go
```diff
@@ -0,0 +1,250 @@
...
+	stderr    io.Writer
+}
+
+func (r *getEgressPolicyRunner) Run(ctx context.Context) error {
+	policy, err := r.getter.GetActorEgressPolicy(ctx, &ateapipb.GetActorEgressPolicyRequest{Actor: r.actor})
+	if status.Code(err) == codes.NotFound {
+		// No policy is a valid state, not a failure: the gateway denies all egress.
+		// TODO(#1703): the server answers NotFound for a missing actor too, so this
+		// note cannot tell a mistyped name from an actor without a policy.
+		fmt.Fprintf(r.stderr, "actor %q in atespace %q has no egress policy or does not exist; an actor without a policy has all egress denied\n", r.actor.GetName(), r.actor.GetAtespace())
```
bowei: delete text after "; an actor without.."

We do not want to put semantic behavior into the CLI text as it's going to be hard to keep consistent.

bowei: We should fetch the Actor to make this unambiguous. This is annoying for the user to be handed this error message.

bowei (commented):
Looks ok, can you look at the suggestions?

cmd/kubectl-ate/internal/cmd/egress_policy.go
bowei: OR we should fix what the server is returning so the error is very clear.

bowei (dismissed):

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: IMO, defaulting logic is helpful so a hand-written manifest doesn't need to repeat the fixed policy name `default` the atespace that the command line already gives, and a manifest naming a different atespace is rejected rather than retargeted.

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Took the structure with a helper `nextDocument` helper to skip the empty documents or those only have comments. PTAL

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done

cmd/kubectl-ate/internal/cmd/egress_policy.go
ygao-g: Done with a CLI-side fix. On a NotFound the command now reads the actor. File a github issue #1703 to track the server-side fix.

ygao-g: Done

bowei (approved):

Issue #1665, opened 2026-09-15
[Bug]: An actor whose resume, suspend or delete fails gets stuck in RESUMING / SUSPENDING / DELETING; it cannot be deleted, its worker stays allocated, and no API call recovers it


ronlv10:
### What happened?
When a lifecycle operation fails on the node, the actor is left in the transitional state it was already committed to and never leaves it:

- a failed resume leaves it `ACTOR_STATE_RESUMING`
- a failed suspend leaves it `ACTOR_STATE_SUSPENDING`
- a failed delete leaves it `ACTOR_STATE_DELETING`

In all three the worker keeps `allocated.actors = 1`. Nothing in the API gets it out. Re-issuing the same operation returns the same error every time, and `DeleteActor(any_state=true)` fails on the node teardown and leaves the actor `DELETING`, still allocated. The only recovery today is deleting the worker pod, which moves the actor to `CRASHED` and frees the worker.
### How we saw it happen
A few ways we saw to reproduce it (and happened in benchmarks):

1. **The request is cancelled or times out mid-teardown.** ateom runs `runsc kill/wait/state/delete` under the request context; a cancellation landing mid-`runsc delete` SIGKILLs runsc partway through destroying the container.
2. `runsc delete` **fails.** `runsc delete -force` can remove the container's state record and still exit 128 (e.g. 5s timeout). Every later `runsc state`/`runsc delete` on that container then fatal-exits 128 (`loading container: file does not exist`), so the teardown can never complete on retry.
3. **The sandbox is lost.** Observed as the ateom container being OOM-killed, for example. `runsc checkpoint` then fails (`cannot checkpoint container "_pause" in state stopped`) and the delete hits (2).
### Expected Behavior
An actor stuck in a transitional state after a node-side failure must be recoverable through the API:

- `DeleteActor(any_state=true)` should succeed and free the worker. A container runsc no longer has a record of is already gone and teardown should treat it as done.
- Better:  the actor should be moved to `CRASHED` and its worker released, the way a missing worker pod already is.
### Steps to Reproduce
Scripts (vibe coded): [hack/repro/stuck-actor](https://github.com/ronlv10/substrate/tree/stuck-actor-repro/hack/repro/stuck-actor) (talk to ateapi and the router directly; setup in the README there).

```
python3 repro_wedge.py oomhog --template <template>
# -> VERDICT[oomhog]: PINNED (actor=SUSPENDING, workers_allocated=[...])
python3 recover_probe.py <actor>
[...]

PR #1666, opened 2026-09-15
ateom-gvisor:  treat a container runsc has no record of as already destroyed

ronlv10:
Issue : #1665
## Problem
Once a container doesn't exists on worker anymore, ateom can never tear the actor down again. `runsc state` and `runsc delete -force` both fatal-exit 128 on a container runsc has no record of (`loading container: file does not exist`; the `-force` not-found tolerance sits after the spec fetch that fails first). 
`cleanupContainers` returned that error, `TerminateWorkload` failed, and ateapi skipped releasing the worker. The actor stayed `DELETING` and its share of the worker's capacity stayed booked; every retry hit the same 128 at the `state` check.

The record goes missing in two observed ways:

- The first `runsc delete` after the sandbox is lost removes the record and still exits 128, because gVisor's `Destroy` runs every step and reports errors at the end. Seen when the ateom container was OOM-killed and restarted: the sandbox and its rootfs mounts were gone, the records were not, and the delete failed on the missing filestore after erasing the record.
- A request deadline or cancellation landing mid-`runsc delete` SIGKILLs runsc partway through destroying the container.
## Change
`cleanupContainers` now treats "runsc has no record of it" as already destroyed:

- When `runsc state` errors, it runs `runsc list` and skips the container if it is not listed. `runsc list` loads no container, so it never fatal-exits on a missing one, which distinguishes "already gone" from a real failure without matching on runsc's error text. It runs only on the error path; the success path is unchanged.
- `stopContainers` and `cleanupContainers` run detached from the request context with their own 30s timeout, as the post-checkpoint resume already does, so a caller deadline can no longer SIGKILL runsc mid-delete. Both `TerminateWorkload` and `CheckpointWorkload`'s post-checkpoint cleanup go through this path.


cmd/ateom-gvisor/cleanup_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/agent-substrate/substrate/internal/proto/ateompb"
+)
+
+// fakeRunsc mocks the runsc binary. It lets a test mock the state, list and
```
ronlv10: hopefully it makes sense to create such fake

cmd/ateom-gvisor/main.go
```diff
@@ -885,31 +885,51 @@ func (r *runsc) stopContainers(ctx context.Context, containers []*ateompb.Contai
...
-	for _, ctr := range containers {
-		if err := r.cmdDelete(ctx, ctr.GetName()); err != nil {
-			return fmt.Errorf("while deleting %q application container: %w", ctr.GetName(), err)
+	// Check state of all containers to mimic containerd.
+	// Without this, `runsc delete` occasionally throws an error.
+	present := make([]string, 0, len(names))
+	for _, name := range names {
+		if err := r.cmdState(ctx, name); err != nil {
+			err = fmt.Errorf("while checking state of %q container: %w", name, err)
+			gone, listErr := r.isContainerAlreadyGone(ctx, name)
```
juli4n: Instead of listing every time, why not listing at the beginning of this function and then only trying to delete the requested containers (plus pause) which are in the list response?

ronlv10: As it's the path of suspend/checkpoint, I didn't want to add an extra step for the happy path.
I was ok with running it multiple times only in the edge case where the container is already deleted.
I did consider having something like a closure with a cache that will be called in the path of failure, something like

```
var listed []string
var listOnce sync.Once
alreadyGone := func(name string) (bool, error) {
        listOnce.Do(func() { listed, listErr = r.cmdList(ctx) })
        if listErr != nil {
                return false, listErr
        }
        return !slices.Contains(listed, name), nil
}
```

WDYT?

juli4n: Ah, good point. Probably not worth the complexity. What you have SGTM. I think we need to track why the runsc state call is needed in the first place, as it sounds like a bug in the CLI. This is a good workaround though.

FYI @fvoznika

cmd/ateom-gvisor/cleanup_test.go
juli4n: Can we somehow avoid the need for fakeRunsc?

cmd/ateom-gvisor/main.go
fvoznika: I'm fixing `runsc delete` so the call to `runsc state` isn't needed anymore. IIRC. it had to do with a zombie sandbox process not being reaped timely.

ronlv10: @fvoznika 
If you're already touching those areas, could we make the `runsc delete` command return a specific exit code for 'not found'? This way, we wouldn't need to list the containers first just to see if that was the reason.

cmd/ateom-gvisor/cleanup_test.go
ronlv10: Updated to use the existing containerRuntime interface - expand it to include other functions  - PTAL, WDYT ?

cmd/ateom-gvisor/main.go
juli4n: I think the are a couple of things happening here:

a) There is a bug in the runsc command that requires that we run runsc state before runsc delete, otherwise runsc delete fails under some conditions. This is clearly a bug in runsc. It should require a specific sequence of calls to succeed the deletion. Once we fix this bug, we should be able to delete the calls to `runsc state` completely.

b) The `--force` flag in `runsc delete` has unclear semantics. It's documented as a way to delete a container irrespective of it's running state, but it also means "ignore not found", although it's not documented. There seems to be also a bug in it if you pass multiple containers and some of them are not found (the executing halts on the first not found)[1]. I filed https://github.com/google/gvisor/issues/14806 for this.

It looks like If we fix both, this should just be a single CLI execution: `runsc delete --force [container names]`.

@fvoznika thoughts?

[1]: https://github.com/google/gvisor/blob/262f28ae6dee59b2b593342f00a8b9001de45763/runsc/cmd/delete.go#L100

juli4n (approved):

cmd/ateom-gvisor/main.go
fvoznika: a) It's being addressed in https://github.com/google/gvisor/pull/14797

b) `delete --force` is idempotent, so it still succeeds after the sandbox is deleted and no longer exists. This is dictated by the OCI (implicitly by following what `runc` does).

If you need something that kills the sandbox on deletion but fails if the sandbox doesn't exist anymore, we can add a different flag for this. PLMK.

PR #1668, opened 2026-09-15
e2e: poll for the rotated trust bundle after suspend/resume

thompsonmax:
## Why

`TestActorIdentity_AfterRestore_IsOwnID_NotGolden` flaked twice on microVM, once on `main` (run 35011781948) and once on #1667 (run 35023632017, attempt 1):

```
identity_test.go:197: after suspend/resume: /run/ate/trust-bundle.pem = "", want the rotated sanitized bundle ... (probe read error: "open /run/ate/trust-bundle.pem: input/output error; ")
```

## Root cause

Race condition when the trust bundle is rotated in the test:

1. atelet rewrites the bundle by rename. The rotation replaces the file at /run/ate/trust-bundle.pem with a new inode. The old inode is unlinked but the guest still references it from its earlier reads.
2. The rename can land during the checkpoint. atelet only stops the refresher after the ateom checkpoint returns, so nothing prevents the rewrite from happening between the guest's last lookup and the snapshot.
3. virtiofsd cannot migrate an unlinked inode. In find-paths mode it has to find a path for every inode the guest holds. The old bundle no longer has one, so guest-error mode marks it faulty on restore, which means EIO instead of a failed restore.
4. The guest still trusts its cached dentry. Guest jiffies do not advance while the VM is paused, so the dentry cached just before suspend is still within its timeout right after resume. The first open goes to the faulty inode and gets EIO. Once the dentry expires, a fresh lookup finds the regenerated file and reads work again.

The other files under `/run/ate` weren't rewritten in that window, which is why only the trust assertion fails.

## Fix

The live-refresh phase of the same test already polls with `waitForTrust` for up to two minutes. Do the same after resume. The test still asserts that the resumed actor sees the rotated bundle whichever side of the suspend the rewrite lands on. It no longer requires that on the first read.

## Testing

`go vet ./internal/e2e/suites/identity/` passes.

Ran this 30 times on my own machine to verify it de-flakes the test using:

```
E2E_SANDBOX_CLASS=microvm hack/run-e2e-kind.sh ./internal/e2e/suites/identity \              
  -count=30 -timeout 3h -run 'TestActorIdentity_AfterRestore_IsOwnID_NotGolden'
```


BenTheElder:
Thanks, Is 2m a reasonable wait time for resume?

Should we consider avoiding a concurrent rotation during the resume?

thompsonmax:
> Thanks, Is 2m a reasonable wait time for resume?
> 
> Should we consider avoiding a concurrent rotation during the resume?

The concurrent rotation is intentional, actors should be able to get the updated trust bundle even if the rotation occurs concurrently with a suspend + resume.

IIUC, the cached dentry for the trust bundle lasts 1s after the actor is resumed (this condition waits for the actor to be running before starting the timeout) so 2 minutes is probably a bit overkill. Dropped it down to 10s.

BenTheElder:
> The concurrent rotation is intentional, actors should be able to get the updated trust bundle even if the rotation occurs concurrently with a suspend + resume.

Yeah, I'm just wondering if we should do something like taking a lock so the actual rotation blocks resume mid-rotation or something. That might be worse but OTOH our own tests were ~crashing on startup.

thompsonmax:
> > The concurrent rotation is intentional, actors should be able to get the updated trust bundle even if the rotation occurs concurrently with a suspend + resume.
> 
> Yeah, I'm just wondering if we should do something like taking a lock so the actual rotation blocks resume mid-rotation or something. That might be worse but OTOH our own tests were ~crashing on startup.

Hmm, the crash you're referring to here is just the EIO on reading the trust bundle after the resume, right? e.g.

```
    identity_test.go:197: after suspend/resume: /run/ate/trust-bundle.pem = "", want the rotated sanitized bundle "-----BEGIN CERTIFICATE-----\nMIIBjjCCATSgAwIBAgIURLlT9x0Dd8euPAOxoNIa3sRsaokwCgYIKoZIzj0EAwIw\nJTEjMCEGA1UEAxMaQVRSQlhBN0RGWlBMTFpZNEpFNVNQUzc0QkIwHhcNMjYwOTE1\nMjEyMTE4WhcNMjcwOTE1MjEyMTE4WjAlMSMwIQYDVQQDExpBVFJCWEE3REZaUExM\nWlk0SkU1U1BTNzRCQjBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABIV4++3onzQI\nFn3z44KPGbg4HpjUkMopnA2mZmHj0eAi6mE4xMGeFsZqj7aI67LBZ8/6OaQlD2fw\nvjhH4+FFk7ajQjBAMA4GA1UdDwEB/wQEAwIChDAPBgNVHRMBAf8EBTADAQH/MB0G\nA1UdDgQWBBThDTiGwUl3fXGMkaOT75Eq+EmY3zAKBggqhkjOPQQDAgNIADBFAiEA\nlZ3TbBQrid8tl/fYwfSMjfGaHOJd2oh+oEUnIyBBxWECIDZ/JX1EYQFBDp4fLNfR\niTTJFsJbbRUK+iM3Lb2YxsGC\n-----END CERTIFICATE-----\n" (probe read error: "open /run/ate/trust-bundle.pem: input/output error; ")
    fixture.go:137: Running command: kubectl --context=kind-kind delete --ignore-not-found -f /tmp/TestActorIdentity_AfterRestore_IsOwnID_NotGolden1414252523/001/probe.yaml
        namespace "ate-e2e-probe-microvm-identity" deleted
        workerpool.ate.dev "probe" deleted from ate-e2e-probe-microvm-identity namespace
--- FAIL: TestActorIdentity_AfterRestore_IsOwnID_NotGolden (39.35s)
FAIL
FAIL	github.com/agent-substrate/substrate/internal/e2e/suites/identity	39.913s
```

So, this comes from the guest's cached dentry for the old bundle inode, which virtiofsd can't migrate after a rename. The guest caches entries for 1s, and that clock freezes while suspended, so a rotation in the second before checkpoint lands in the snapshot as a dead entry, on resume returning EIO until the cached dentry expires.

One possible fix is to stop the live refresher during suspend, check if a re-write happened in the last second, and wait it out so the stale entry expires.

This would cost up to 2s in extra suspend latency, only if a rotation comes in during this window. But would eliminate this up to 1s window after resume where the trust bundle file gives EIO if you try to read it (which also only occurs if you have a rotation come in during this window).

If that sounds reasonable, I'm happy to merge this fix first to stop the test flakes, then add that fix in a follow up.

One thing to mention though is that trust bundle rotation should be very rare in practice (`kubectl ate admin make-ca-pool` currently sets a 365 day validity), so this might not be the highest priority thing to fix. But happy to add an issue tracking it.

BenTheElder (approved):

PR #1669, opened 2026-09-15
Add `--enable-nested-virtualization` to `setup-gcp`

hdp617:
Nested virtualization exposes `/dev/kvm`, which microVMs need. The new `--enable-nested-virtualization` flag turns it on for the node pool `setup-gcp` creates. It defaults to on, so pass `--enable-nested-virtualization=false` for a cluster that does not need it.

Also, rename the env var `GVISOR_NODE_MACHINE_TYPE` to `NODE_MACHINE_TYPE` to be consistent with other env vars and to avoid confusion for microVMs. The old env var still works with a warning.

- [x] Tests pass


hdp617:
cc @cheftako @BenTheElder

hack/ate-dev-env.sh.example
```diff
@@ -34,7 +34,9 @@ export CLUSTER_NAME=substrate-poc
 export CLUSTER_VERSION=1.36.3-gke.1767000
 export NODE_POOL_NAME=gvisor-pool
 export NODE_POOL_VERSION=1.36.3-gke.1767000
-export GVISOR_NODE_MACHINE_TYPE=c3-standard-4
+export NODE_MACHINE_TYPE=c3-standard-4
+# Enable nested virtualization (required for microVMs)
+# export ENABLE_NESTED_VIRTUALIZATION=true
```
BenTheElder: we should default to doing this.

hack/ate-dev-env.sh.example
```diff
@@ -34,7 +34,9 @@ export CLUSTER_NAME=substrate-poc
 export CLUSTER_VERSION=1.36.3-gke.1767000
 export NODE_POOL_NAME=gvisor-pool
 export NODE_POOL_VERSION=1.36.3-gke.1767000
-export GVISOR_NODE_MACHINE_TYPE=c3-standard-4
```
BenTheElder, line 37: this is a breaking change given the workflow is to create a copy of dev-env.sh when you first setup substrate and modify it.

most of the team will not have this new environment variable

hack/ate-dev-env.sh.example
hdp617: Done.

hack/ate-dev-env.sh.example
hdp617: I added a fallback to the old env var with a warning. We can remove it later; I don't want to break anyone :)

tools/setup-gcp/cmd/bootstrap.go
```diff
@@ -86,7 +86,8 @@ func init() {
 	bootstrapCmd.Flags().StringVar(&cfg.ClusterVersion, "cluster-version", getEnv("CLUSTER_VERSION", ""), "Kubernetes version [env: CLUSTER_VERSION]")
 	bootstrapCmd.Flags().StringVar(&cfg.Network, "network", getEnv("NETWORK", "default"), "VPC network name [env: NETWORK]")
 	bootstrapCmd.Flags().StringVar(&cfg.Subnetwork, "subnetwork", getEnv("SUBNETWORK", "default"), "VPC subnetwork name [env: SUBNETWORK]")
-	bootstrapCmd.Flags().StringVar(&cfg.MachineType, "machine-type", getEnv("GVISOR_NODE_MACHINE_TYPE", "c3-standard-4"), "Machine type for the gVisor node pool [env: GVISOR_NODE_MACHINE_TYPE]")
+	bootstrapCmd.Flags().StringVar(&cfg.MachineType, "machine-type", defaultMachineType, "Machine type for the node pool [env: NODE_MACHINE_TYPE]")
+	bootstrapCmd.Flags().BoolVar(&cfg.EnableNestedVirtualization, "enable-nested-virtualization", getEnv("ENABLE_NESTED_VIRTUALIZATION", false), "Create the node pool with nested virtualization, exposing /dev/kvm for micro-VM workers; needs a machine type that supports it [env: ENABLE_NESTED_VIRTUALIZATION]")
```
BenTheElder: make this true?

BenTheElder: (and the other references)

hdp617: Done, in both commands and the README tables.

Two consequences, both of which seem right: reconciling a cluster created before this flag now logs the drift warning, since GKE cannot turn nested virtualization on for an existing node pool; and cluster creation now fails outright for the machine types GCE excludes from the feature (E2, memory-optimized, Arm, and AMD other than N4D) instead of quietly creating nodes without `/dev/kvm`.

tools/setup-gcp/cmd/cluster.go
```diff
@@ -371,8 +398,9 @@ func init() {
...
 	clusterCmd.Flags().StringVar(&cfg.Network, "network", getEnv("NETWORK", "default"), "VPC network name [env: NETWORK]")
 	clusterCmd.Flags().StringVar(&cfg.Subnetwork, "subnetwork", getEnv("SUBNETWORK", "default"), "VPC subnetwork name [env: SUBNETWORK]")
-	clusterCmd.Flags().StringVar(&cfg.MachineType, "machine-type", getEnv("GVISOR_NODE_MACHINE_TYPE", "c3-standard-4"), "Machine type for the gVisor node pool [env: GVISOR_NODE_MACHINE_TYPE]")
+	clusterCmd.Flags().StringVar(&cfg.MachineType, "machine-type", defaultMachineType, "Machine type for the node pool [env: NODE_MACHINE_TYPE]")
 	clusterCmd.Flags().BoolVar(&cfg.EnableDataplaneV2, "enable-dataplane-v2", getEnv("ENABLE_DATAPLANE_V2", true), "Enable Dataplane V2 [env: ENABLE_DATAPLANE_V2]")
+	clusterCmd.Flags().BoolVar(&cfg.EnableNestedVirtualization, "enable-nested-virtualization", getEnv("ENABLE_NESTED_VIRTUALIZATION", true), "Create the node pool with nested virtualization, exposing /dev/kvm for micro-VM workers; needs a machine type that supports it. Turn off with --enable-nested-virtualization=false [env: ENABLE_NESTED_VIRTUALIZATION]")
```
BenTheElder: 🤖 **should-fix** 🟡 – Defaulting this to true makes `bootstrap` and `create cluster` hard-fail on configurations GCE does not allow nested virtualization for: E2, Arm, memory-optimized, and every AMD type except N4D, plus any `CLUSTER_VERSION` pinned below 1.28.4-gke.1083000.

There is no local check, so the failure surfaces as an opaque `INVALID_ARGUMENT` from GKE after the request goes out. The tool already validates boot disk and location locally, so a guard on machine-type family (and version, if pinned) would fit the existing shape and give the operator an actionable message.

BenTheElder, line 404: 🤖 **nit** 🟢 – `getEnv` parses with `strconv.ParseBool`, which rejects `off` and `no`. Those fall through to the `true` fallback, so `ENABLE_NESTED_VIRTUALIZATION=off` silently turns the feature **on**.

This is pre-existing `getEnv` behavior, but it matters more here: this is the first default-on knob where the env var is the documented way to turn something off, and the failure is silent in the dangerous direction. Either accept the common falsey spellings in `getEnv`, or log when a set value fails to parse instead of falling back.

tools/setup-gcp/cmd/common.go
```diff
@@ -49,6 +51,18 @@ type Config struct {
 	DashboardDir string
 }
 
+var defaultMachineType = resolveMachineTypeDefault()
+
+func resolveMachineTypeDefault() string {
+	if _, ok := os.LookupEnv("NODE_MACHINE_TYPE"); !ok {
```
BenTheElder: 🤖 **nit** 🟢 – Keying off `LookupEnv` presence rather than a non-empty value means an exported-but-empty `NODE_MACHINE_TYPE` does two wrong things at once: it suppresses the `GVISOR_NODE_MACHINE_TYPE` fallback, and it returns `""`.

An empty `MachineType` lets GKE pick `e2-medium`, which cannot do nested virtualization, so the run then fails the request added in this PR. `if v := os.Getenv("NODE_MACHINE_TYPE"); v == ""` covers both.

tools/setup-gcp/cmd/common.go
```diff
@@ -49,6 +51,18 @@ type Config struct {
 	DashboardDir string
 }
 
+var defaultMachineType = resolveMachineTypeDefault()
```
BenTheElder: 🤖 **nit** 🟢 – As a package-level var this runs at init, so the `GVISOR_NODE_MACHINE_TYPE is deprecated` warning fires on every invocation of the binary — `--help`, `create bucket`, and runs that pass `--machine-type` explicitly — for anyone who still has the old variable exported. Resolving it inside `init()` where the flag is registered, or at first use, keeps the warning tied to the command that actually reads it.

BenTheElder (commented):
AI review I agree with (dismissed irrelevant comments):

tools/setup-gcp/cmd/cluster.go
hdp617: Add a warning for such values.

tools/setup-gcp/cmd/common.go
hdp617: Done.

tools/setup-gcp/cmd/common.go
hdp617: Done.

tools/setup-gcp/cmd/common.go
```diff
@@ -77,9 +107,12 @@ func getEnv[T getEnvType](key string, fallback T) T {
 		ret = int32(v)
 	}
 
-	if err == nil {
-		return ret.(T)
+	if err != nil {
+		if _, warned := warnedEnvKeys.LoadOrStore(key, struct{}{}); !warned {
+			slog.Warn("Ignoring unparsable environment variable", "key", key, "value", val, "using", fallback, "error", err)
```
BenTheElder, line 112: 🤖 **nit** 🟢 – `using=fallback` can be wrong. This runs at flag-registration time, so an explicit CLI flag still wins afterwards. `ENABLE_NESTED_VIRTUALIZATION=off ... --enable-nested-virtualization=false` logs `using=true` while the cluster is created with nested virtualization off.

Dropping the `using` field, or wording it as the default rather than the effective value, would avoid the contradiction.

tools/setup-gcp/README.md
```diff
@@ -135,10 +135,14 @@ go run ./tools/setup-gcp create cluster [flags]
 | `--version` | Kubernetes version. | `CLUSTER_VERSION` | None |
 | `--network` | VPC network name. | `NETWORK` | `default` |
 | `--subnetwork` | VPC subnetwork name. | `SUBNETWORK` | `default` |
-| `--machine-type` | Machine type for the gVisor node pool. | `GVISOR_NODE_MACHINE_TYPE` | `c3-standard-4` |
+| `--machine-type` | Machine type for the node pool. | `NODE_MACHINE_TYPE` | `c3-standard-4` |
+| `--enable-nested-virtualization` | Create the node pool with nested virtualization, exposing `/dev/kvm` for micro-VM workers. | `ENABLE_NESTED_VIRTUALIZATION` | `true` |
```
BenTheElder, line 139: 🤖 **should-fix** 🟡 – Worth a note here that this applies only to the pool `setup-gcp` creates. GKE does not propagate nested virtualization to pools created later, and it cannot be enabled on an existing pool.

The "create additional pools" recipe further down (lines 150-154) only mentions `--node-labels`, so a pool made by following it has no `/dev/kvm` and micro-VM workers scheduled there stay Pending with no obvious cause. Adding the nested-virtualization flag to that recipe would close the gap.

BenTheElder: [Or we can consider a follow-up to support reconciling, but again, this code should really be moved out to the GKE installer repo anyhow, this tool does not belong in the main repo]

BenTheElder (approved):
I'm ok to handle further nits in follow-up and land and iterate

PR #1670, opened 2026-09-16
Initialize OpenFGA server on ATE API Server bootstrap.

SHRUTI6991:

## Summary
Initializes an embedded OpenFGA server within `ateapi` backed by PostgreSQL and updates the top-level authorization scope.

## Key Changes
- **Rename scope to `global`**: Renamed `type cluster` and `parent_cluster` to `type global` and `parent_global`, adding `can_set_policy` and `can_get_policy` to `global`.
- **Embedded OpenFGA server (`internal/authz/server.go`)**:
  - Compiles `model.fga` into OpenFGA's protobuf representation.
  - Runs OpenFGA PostgreSQL schema migrations (`goose_db_version`).
  - Serializes startup via `pg_advisory_lock` to avoid multi-replica races.
  - Connects OpenFGA's PostgreSQL adapter with a new dedicated `*pgxpool.Pool`.
  - Idempotently creates or reuses the `"substrate"` store and authorization model.
- **Service wiring (`cmd/ateapi`)**: Exposes `Pool()` on `*atepg.Persistence` and initializes `authz.NewServer` on bootstrap.

## Verification
- `openfga model test`: 103/103 checks passing (`model_test.fga.yaml`).
- `go test -mod=mod ./internal/authz/...`: Passes against PostgreSQL (migrations, tuple writes, permission checks, and restart idempotency).


SHRUTI6991:
Please review first commit. The other commit is vendor dependencies.

internal/authz/server.go
```diff
@@ -0,0 +1,291 @@
...
+	datastore, err := postgres.NewWithDB(pool, nil, cfg)
+	if err != nil {
+		return nil, fmt.Errorf("creating OpenFGA postgres adapter: %w", err)
+	}
+
+	fgaServer, err := server.NewServerWithOpts(
+		server.WithDatastore(datastore),
+	)
+	if err != nil {
+		datastore.Close()
```
thompsonmax, line 93: IIUC, this also closes the provided `*pgxpool.Pool`. Maybe worth mentioning that in the NewServer() function comment and double checking to make sure that a dedicated pool is always provided here so this doesn't break anything (e.g. if it was a shared pool).

internal/authz/server_test.go
```diff
@@ -0,0 +1,193 @@
...
+		_ = os.Setenv("TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE", socket)
+	}
+	return nil
+}
+
+func startPostgres(t *testing.T) *pgxpool.Pool {
+	t.Helper()
+	ctx := context.Background()
+	if err := configureDockerEnv(ctx); err != nil {
+		t.Skipf("Docker not available for testcontainers: %v", err)
```
thompsonmax, line 59: I think this only catches if the docker CLI is not installed, but doesn't check whether the docker daemon is healthy.

I believe there's an existing `testcontainers.SkipIfProviderIsNotHealthy(t)` function we can add a call to here to validate the daemon is healthy.

internal/authz/server.go
SHRUTI6991: Good catch, I actually created a dedicated pool for openfga connections.

internal/authz/server_test.go
SHRUTI6991: Done.

internal/authz/server.go
SHRUTI6991: This is also because, my implementation for create atespace + auth check happens in 2 different transactions since openFGA always executes in it's own transaction.

cmd/ateapi/main.go
```diff
@@ -149,6 +143,14 @@ func main() {
 		defer closer.Close()
 	}
 
+	if poolProvider, ok := persistence.(interface{ Pool() *pgxpool.Pool }); ok {
```
ahmedtd: I don't think we should be creating a new connection pool for OpenFGA --- we should have one connection pool for both OpenFGA and the store (another connection pool just increases connection pressure on Postgres for no gain).

This looks like it might be a larger refactor, since we should probably be dialing  the connection pools in main and injecting them into the store and OpenFGA modules.  It can be done as a follow-up.

@juli4n

internal/authz/server.go
```diff
@@ -0,0 +1,291 @@
...
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/jackc/pgx/v5/stdlib"
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"github.com/openfga/language/pkg/go/transformer"
+	"github.com/openfga/openfga/assets"
+	"github.com/openfga/openfga/pkg/server"
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/storage/postgres"
+	"github.com/openfga/openfga/pkg/storage/sqlcommon"
+	"github.com/pressly/goose/v3"
```
ahmedtd, line 34: Just to clarify, this is the same solution we are using for applying migrations for the main store?

internal/authz/server.go
```diff
@@ -0,0 +1,291 @@
...
+	// Substrate's own schema_migrations table.
+	migrationTableName = "goose_db_version"
+)
+
+//go:embed model.fga
+var modelDSL string
+
+// Server wraps an embedded OpenFGA server backed by PostgreSQL and
+// initialized with Substrate's authorization model.
+type Server struct {
```
ahmedtd, line 54: I am not sure on whether or not we should have this wrapper struct or instead just return the *server.Server from NewServer.

Let's see some of the API call integrations, and we can revisit.

internal/authz/server.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		fgaServer.Close()
+		datastore.Close()
+		return nil, err
+	}
+
+	storeID, modelID, err := ensureStoreAndModel(ctx, fgaServer)
+	unlock()
+	if err != nil {
+		fgaServer.Close()
```
ahmedtd, line 107: Needing to write these lines points sort of points to that we should be creating the fgaServer and datastore in main (where they can have an unconditional defer close() after they are created).

The risk of doing this without defer is that a panic will leave them open.  But obviously, it's complicated to write the defer() logic in the scope of NewServer.  That's why constructor functions should usually accept their stateful dependencies as injected arguments, not try to create them inline.

I would reconsider whether this package should have a wrapper struct, or just provide utility functions to initialize the FGA server, datastore, and apply migrations.

ahmedtd (dismissed):

thompsonmax (approved):

internal/authz/server.go
SHRUTI6991: Yes its the same: https://github.com/agent-substrate/substrate/blob/5f7c6901089c9c1e6acce624330f149b81a9cfc4/cmd/ateapi/internal/store/atepg/schema.go#L28

cmd/ateapi/main.go
SHRUTI6991: working on this as followup task: https://github.com/agent-substrate/substrate/issues/1755.

BenTheElder (approved):

PR #1672, opened 2026-09-16
Declare runtime environment variables and generate their reference

### LINKED ISSUE #1662: Make Go environment variables discoverable through a documented registry
Substrate reads environment variables through scattered `os.Getenv` and `os.LookupEnv` calls. Discovering supported variables, defaults, and behavior currently requires searching code and individual guides.

We could introduce a lightweight way to declare environment variables in Go with their name, type, default, description, and consuming component, then generate a searchable reference from those declarations.

Both projects have useful examples:

- [Kagent’s env package](https://github.com/kagent-dev/kagent/blob/main/go/core/pkg/env/var.go) provides typed registration, component metadata, and Markdown/JSON export.
- [Istio’s env package](https://github.com/istio/istio/blob/master/pkg/env/var.go) provides typed registration and `VarDescriptions()` to support documentation generation.

A similar approach could use a small `internal/env` package, with declarations near the code that owns each setting and a generated Markdown reference linked from the docs. The reference should explain component-specific defaults, accepted values, and precedence relative to CLI flags.

Migration should preserve existing configuration behavior. Generated documentation should use declared metadata, never live environment values or secrets.

EItanya:
Fixes #1662

This PR does not fix test/setup code, it's meant to start the process for the production code paths and get the conversation going.

Adds generic environment declarations with name, default, and description for production components and shared runtime packages. Generates a Markdown reference with separate operator/system sections and checks for documentation drift.


cmd/ateapi/main.go
```diff
@@ -66,6 +67,35 @@ const maxRPCDeadline = 10 * time.Minute
 
 const minResyncInterval = 250 * time.Millisecond
 
+var storageBackendEnv = env.Var[string]{
+	Name:    "ATE_STORAGE_BACKEND",
+	Default: "",
+	Description: `Selects the snapshot storage backend. The exact, case-sensitive value s3 selects S3.
+Unset, empty, and all other values select GCS.`,
```
LiorLieberman, line 74: >Unset, empty, and all other values select GCS.`,

I read this as "GCS is the default" why not setting it as the deafult. Relying on empty to select GCS will make it less clean to change in the future if ever needed

cmd/atecontroller/main.go
```diff
@@ -49,6 +50,40 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
+var otelEndpointEnv = env.Var[string]{
+	Name:        "OTEL_EXPORTER_OTLP_ENDPOINT",
+	Default:     "",
+	Description: "Default for --otel-exporter-otlp-endpoint. Empty disables injection of worker telemetry settings; a flag overrides the environment.",
+}
```
LiorLieberman, line 57: should these all be in a separate env file or something?

cmd/atelet/main.go
```diff
@@ -84,6 +85,21 @@ import (
 	"k8s.io/utils/lru"
 )
 
+var storageBackendEnv = env.Var[string]{
+	Name:    "ATE_STORAGE_BACKEND",
+	Default: "",
+	Description: `Selects the snapshot storage backend. The exact, case-sensitive value s3 selects S3.
+Unset, empty, and all other values select GCS.`,
```
LiorLieberman, line 92: same comment on default, plus why are we repeating it, can this be shared?

LiorLieberman (commented):
we are buliding a whole env package, is there no existing packages for what you are trying to do?

Issue #641, opened 2026-07-30
DeleteActor never cleans the actor's node directory


igooch:
## Expected Behavior

Deleting an actor releases everything the actor held, including its node-local state: after
`DeleteActor`, `/var/lib/ateom-gvisor/actors/<uid>/` should be removed (or at minimum contain
nothing that marks the actor as still placed on the node).

This matters doubly once #463 Phase 2 lands: the image-cache GC's root set is the per-bundle
`rootfs-overlay.json` specs — "spec on disk" means "actor placed here; do not evict its image."
A deleted actor should not root an image.

## Actual Behavior

`DeleteActor` is a pure control-plane operation: it deletes the store record and cleans up volumes
(`cmd/ateapi/internal/controlapi/delete_actor.go`), but nothing ever tells the node. atelet's RPC
surface is only `Run`/`Checkpoint`/`Restore`, and `resetActorDirs` runs lazily — at the start of the
next `Run`/`Restore` or the end of a `Checkpoint` for the **same actorUID**. A deleted actor's UID
never comes back, so `/var/lib/ateom-gvisor/actors/<uid>/` persists until the node is replaced.

What leaks depends on the state the actor was deleted from (the store allows SUSPENDED or CRASHED,
`cmd/ateapi/internal/store/ateredis/ateredis.go`):

- **Deleted from CRASHED**: a crash means no `Checkpoint` ran after the last `Run`/`Restore`, so the
  **entire bundle dir survives** — the actor's writable `upper/` layer and the
  `bundles/<container>/rootfs-overlay.json` spec. Post-#463-Phase-2, that leaked spec is
  indistinguishable from a live one, so the dead actor's image is **rooted forever**: GC logs it as
  in-use on every pass and it can never be evicted, regardless of disk pressure. Hit twice during
  Phase 2 GC validation. The GC is behaving as designed (ambiguity must resolve toward retention —
  evicting a mounted lowerdir is catastrophic); the defect is that deletion leaves the "I am placed
  here" marker behind. Filed separately from #463 so this over-retention isn't mistaken for a GC bug.
- **Deleted after a clean suspend** (the common path): the bundle was wiped at `Checkpoint`, but the
  `ActorPath` skeleton, `sandbox-assets.json`, and any `local-checkpoint/` contents are left
  behind.

## Steps to Reproduce the Problem

1. On a Kind cluster with substrate installed, run an actor, then crash it (e.g. kill its
   workload) so it lands in CRASHED.
[...]

PR #1675, opened 2026-09-16
feat(api): introduce RevertActor lifecycle RPC

### LINKED ISSUE #1556: Need a way to recover an actor from CRASHED state
Today an Actor that gets into CRASHED state cannot be recovered. We should have a way to "rollback" to the latest saved snapshot and resume. There are different ways of encoding this into the API. A few options I can think of are a) an explicit `RollbackActor` (or `RevertActor`) method which flips the state back to `SUSPENDED` b) allow `ResumeActor` to just resume (this works today if we remove the precondition check for `CRASHED`) c) allow `SuspendActor` to switch from `CRASHED` to `SUSPENDED`. Others?

Related to #119

shrutiyam-glitch:
#### Summary
This PR introduces the `RevertActor` RPC for actor lifecycle management. It includes the API definition, the corresponding workflow execution logic, observability metrics, and updates to the authorization model to support reverting actors. 

Fixes #1556
Docs updated in #1711 

#### Commit-wise Changes

**1. Add RevertActor RPC (`1145d99`)**
* Introduces the new `RevertActor` RPC to the API definitions.
* Updates the corresponding protobuf bindings (affecting files like `ateapi.pb.go` and `ateapi_pb2.py`).

**2. Add the revertActor workflow and observability metrics (`dcb1fae`)**
* Implements the core `revertActor` workflow logic, designed to be idempotent and re-enterable. It progresses through the following steps:
  * **Mark Reverting:** Validates that the actor is in a revertable state (`RUNNING`, `PAUSED`, or `CRASHED`) and transitions its state to `REVERTING`.
  * **Discard Worker:** Safely tears down the execution environment by terminating the workload, detaching volumes, and releasing the assigned worker.
  * **Collect In-Progress Snapshot:** Cleans up external object storage by deleting any objects a previous suspend operation was partway through writing.
  * **Finalize:** Commits the actor to `SUSPENDED` and strips all node-local and in-progress state pointers (clearing `WorkerAssignment`, `LocalSnapshotInfo`, etc.), returning the actor to its untouched external snapshot.
* Instruments the workflow with lifecycle operation metrics (e.g., updating `ate.actor.lifecycle.operation.duration` to track `revert` operations).
* *Note/TODO:* Currently, when reverting a paused actor, the workflow drops the pointer to the node-local state but does *not* actually prune the local checkpoint bytes from the node (this is tracked in #641).

**3. Add `can_revert` to the authorization model (`2fc53ae`)**
* Adds the `can_revert` permission to the auth model, mirroring the shape of `can_suspend` (editor tier of the parent atespace, plus a direct grant so a machine identity can revert the actor it drives without holding an atespace role).

**4. Serve RevertActor and add the CLI verb (`f04fab4`)**
* Wires the `Control.RevertActor` service method to the workflow (replacing the generated stub that previously answered `Unimplemented`).
* Adds the `"ate revert actor"` CLI command, making the feature usable end-to-end.
* Implements `Terminate` for the fake atelet. This was necessary because reverting an actor from the `RUNNING` state is the first path to reach this call in functional tests (previously, delete tests skipped this step as they ran against actors with no worker assignment).

**5. Add a manual verify script for RevertActor (`8022db3`)**
* Adds a script to manually exercise `RevertActor` against a real control plane, since unit and functional tests only run against a fake atelet.
* Tests reverting from `CRASHED`, `RUNNING`, and `PAUSED` states, and verifies that attempting to revert a `SUSPENDED` actor is properly rejected.
* Simulates a crash by deleting the worker pod the actor runs on to verify the workflow can handle the absence of a worker to terminate.

- [ ] Tests pass


cmd/ateapi/internal/controlapi/functionaltest/actor_test.go
```diff
@@ -3516,3 +3516,112 @@ func TestMintActorJWT_Success(t *testing.T) {
 		t.Fatalf("Error while calling MintActorJWT: %v", err)
 	}
 }
+
+// TestRevertActor returns a running actor to the snapshot its last suspend
+// wrote, without taking a new one.
+func TestRevertActor(t *testing.T) {
```
zoez7, line 3523: We also need a test case for reverting from CRASHED?

hack/verify-revert-actor.sh
zoez7, line 1: Is this file for manual testing only? Did you mean to commit it?

cmd/ateapi/internal/controlapi/workflow_revert.go
```diff
@@ -0,0 +1,255 @@
...
+			return fmt.Errorf("while releasing worker: %w", err)
+		}
+	}
+
+	return w.releaseAssignmentWithoutBacklink(ctx, actor)
+}
+
+// ensureInProgressSnapshotCollected deletes the objects a suspend was partway
+// through writing when the actor was reverted.
+func (w *ActorWorkflow) ensureInProgressSnapshotCollected(ctx context.Context, actor *ateapipb.Actor, actorTemplate *ateapipb.ActorTemplate) (err error) {
```
zoez7: Nit: ensureInProgressSnapshotCollected -> ensureInProgressSnapshotDiscarded?

cmd/ateapi/internal/controlapi/workflow_revert.go
```diff
@@ -0,0 +1,255 @@
...
+	if err = w.ensureWorkerDiscarded(leaseCtx, actorRef, actor, actorTemplate); err != nil {
+		return nil, err
+	}
+
+	if err = w.ensureInProgressSnapshotCollected(leaseCtx, actor, actorTemplate); err != nil {
+		return nil, err
+	}
+
+	// TODO: Check if local checkpoints need to be pruned #641
+	// w.ensureLocalCheckpointsPruned(leaseCtx, actor)
```
zoez7, line 92: The local checkpoints should be pruned by ensureWorkerDiscarded -> ensureAteletTerminated above. is that not the case?

hack/verify-revert-actor.sh
shrutiyam-glitch: Yeah, it is just for manual testing. But, since it is under `hack/` folder, I pushed it as well.

cmd/ateapi/internal/controlapi/workflow_revert.go
```diff
@@ -0,0 +1,255 @@
...
+
+	actor, err := w.store.GetActor(ctx, actorRef)
+	if err != nil {
+		if errors.Is(err, store.ErrNotFound) {
+			return nil, nil, status.Errorf(codes.NotFound, "Actor %s not found", actorRef)
+		}
+		return nil, nil, fmt.Errorf("while fetching actor: %w", err)
+	}
+
+	actorTemplate, err := resolveActorTemplate(ctx, w.store, actor)
```
zoez7, line 117: I think there is an issue with the sequence:

1. suspended actor on actor template v1 , take a full snapshot, actor.status.current_actor_template_uid == v1
3. UpdateActor to v2, ResumeActor, now  current_actor_template_uid updated to v2
4. Call RevertActor, now the snapshot is reverted to the one taken on v1, but the actor's current_actor_template_uid is still v2.

Can we move the actor_template_uid  as a field of ExternalSnapshot, so the revert will take effect together?

cmd/ateapi/internal/controlapi/workflow_revert.go
shrutiyam-glitch: Currently, `ensureAteletTerminated` runs pruneLocalCheckpoints(actorUID) only for RUNNING actors 

For PAUSED actors, worker_assignment is nil (https://github.com/agent-substrate/substrate/blob/85ce8ed5313f64aa4d8d5fb616ce450abf1e95e0/cmd/ateapi/internal/controlapi/workflow_delete.go#L122-L125
), so ensureAteletTerminated skips. 
so, I guessed it will be a separate step itself and get called from `deleteActor` flow and `revertActor` flow.

cmd/ateapi/internal/controlapi/functionaltest/actor_test.go
shrutiyam-glitch: Done in https://github.com/agent-substrate/substrate/pull/1675/commits/a66fe23d9145109345420e407e40dd447697cf5a
thanks !

cmd/ateapi/internal/controlapi/workflow_revert.go
shrutiyam-glitch: Done in https://github.com/agent-substrate/substrate/pull/1675/commits/e0571bd30e24642e17a85b198f53af4d6e0835d4

shrutiyam-glitch:
The tests are failing because of atelet panic - documented here - #1710. 
Temporary fix for this - #1684

SHRUTI6991:
The FGA part looks good to me.

cmd/ateapi/internal/controlapi/workflow_revert.go
shrutiyam-glitch: handled this case in a follow up PR - https://github.com/agent-substrate/substrate/pull/1713

cmd/ateapi/internal/controlapi/workflow_revert.go
zoez7: I see, we should be able to call pruneLocalCheckpoints for paused actors as well, after https://github.com/agent-substrate/substrate/pull/1742 is submitted, we can do it in a follow up PR.

cmd/ateapi/internal/controlapi/workflow_revert.go
```diff
@@ -0,0 +1,256 @@
...
+
+// ensureRevertedFinalized commits SUSPENDED and drops every pointer to the
+// execution that was discarded, in a single update.
+// It re-reads the actor first so an out-of-band transition is not overwritten,
+// and so the version precondition guards what was actually observed.
+func (w *ActorWorkflow) ensureRevertedFinalized(ctx context.Context, actorRef resources.ActorRef) (_ *ateapipb.Actor, err error) {
+	ctx, done := stepSpan(ctx, "FinalizeReverted")
+	defer func() { err = done(err) }()
+
+	latestActor, err := w.store.GetActor(ctx, actorRef)
```
zoez7, line 241: check the latestActor is in Reverting state first?

cmd/ateapi/internal/controlapi/workflow_revert.go
```diff
@@ -0,0 +1,256 @@
...
+//
+// The sweep afterwards runs on every origin, not just when the record names no
+// worker. An assignment commits before the actor is updated to point at it, so
+// a crash in between leaves a row nothing references — and a CRASHED actor is
+// the likeliest holder of one. Absence is the ordinary case, not an error.
+func (w *ActorWorkflow) ensureWorkerDiscarded(ctx context.Context, actorRef resources.ActorRef, actor *ateapipb.Actor, actorTemplate *ateapipb.ActorTemplate) (err error) {
+	ctx, done := stepSpan(ctx, "DiscardWorker")
+	defer func() { err = done(err) }()
+
+	if actor.GetStatus().GetWorkerAssignment() != nil {
```
zoez7: We should check if the worker is still assigned to the same Actor.

cmd/ateapi/internal/controlapi/workflow_revert.go
```diff
@@ -0,0 +1,256 @@
...
+		return nil
+	}
+
+	uri, err := resources.ParseSnapshotURI(inProgress)
+	if err != nil {
+		return fmt.Errorf("while parsing the in-progress snapshot %q: %w", inProgress, err)
+	}
+	// A suspend records the in-progress URI under the actor's own prefix
+	// before atelet writes the first object, so a URI owned by anything else
+	// is a corrupted record: deleting it would collect another actor's data.
```
zoez7:
```suggestion
	// is a corrupted record.
```

It shouldn't be possible that another actor owns this snapshot.

cmd/ateapi/internal/controlapi/workflow_revert.go
```diff
@@ -0,0 +1,256 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// revertableStates are the states a revert is accepted from.
+var revertableStates = []ateapipb.ActorState{
+	ateapipb.ActorState_ACTOR_STATE_RUNNING,
+	ateapipb.ActorState_ACTOR_STATE_PAUSED,
+	ateapipb.ActorState_ACTOR_STATE_CRASHED,
```
zoez7, line 39: Reverting should also be allowed for retrying the request.

hack/verify-revert-actor.sh
shrutiyam-glitch: removed the file

cmd/ateapi/internal/controlapi/workflow_revert.go
shrutiyam-glitch: it is being checked explicitly in the `ensureMarkedReverting` step before the `revertableStates` are checked - https://github.com/agent-substrate/substrate/pull/1675/changes#diff-f3a280897c9511be4309ea3ad2e0c9acd8d6584f11038738d5b30f4a4e3c302cR139-R141
to mark it as `skipped`
 should that be fine ?

zoez7: 👍🏼  sounds good.

cmd/ateapi/internal/controlapi/workflow_revert.go
shrutiyam-glitch: done in https://github.com/agent-substrate/substrate/pull/1675/commits/30870b7d352d50625b9d88d22989b3175cbeb9ac

cmd/ateapi/internal/controlapi/workflow_revert.go
shrutiyam-glitch: done in https://github.com/agent-substrate/substrate/pull/1675/commits/eaf34c58f732ab183fe9ce26aa54680f6f522098

zoez7 (approved):

PR #1676, opened 2026-09-16
ateapi: guard egress policy delete with uid and version

ygao-g:
Part of #1550.

### Summary

A client that reads an actor's egress policy and later deletes it has no way to say which policy it meant. If the policy was updated or recreated in between, the delete removes the newer one. `DeleteActorEgressPolicy` now accepts the optional `DeleteOptions` (`uid`, `version`) that `DeleteWorkerRequest` already takes; when they are set, the store deletes only if the stored policy still matches and the client gets `ABORTED` otherwise. Omitting them keeps the unguarded delete.

### Commits

1. `ateapi: let egress policy deletes carry uid and version preconditions` — proto change; `ateapi.pb.go`, `ateapi_pb2.py` and `zz_generated.validation.go` are regenerated with `hack/update/codegen.sh`.
2. `ateapi/store: refuse egress policy deletes with a stale uid or version` — `store.Interface.DeleteEgressPolicy` takes `store.DeletePreconditions`; the row is locked `FOR UPDATE` and checked before deletion, returning `ErrUIDConflict` / `ErrVersionConflict`; zero values waive the check.
3. `ateapi: reject stale egress policy deletes with Aborted` — the handler forwards `req.Options`; the existing mapper returns `ABORTED` on conflict and `NOT_FOUND` for a missing policy; validation returns `INVALID_ARGUMENT` for a malformed uid or a version below 1.

Actor deletion still cascades to the policy without precondition checks (by design). No in-tree caller deletes egress policies today other than `kubectl ate delete egress-policy`; its `--uid`/`--version` flags come in a follow-up.

### Testing

On a kind cluster running this server, driven by a CLI that fills `DeleteOptions`:
- stale `version` or wrong `uid` → `ABORTED` (`EgressPolicy version conflict` / `EgressPolicy UID conflict`), policy kept
- `uid x` or `version -1` → `INVALID_ARGUMENT` from request validation
- matching uid and version → deleted; repeated → `NOT_FOUND`; no options → deleted as before
- the same guarded request against the previous server build → `INVALID_ARGUMENT: unknown field with protobuf tag 2`

- [x] Tests pass:
  - unit tests, store contract tests, in-process gRPC functional tests, `make verify`
  - the guarded delete matrix above on a local kind cluster running this server
- [x] Appropriate changes to documentation are included in the PR: the new request field is documented in `ateapi.proto`; no user-facing docs change until the CLI flags land.


cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -994,21 +994,36 @@ func (p *Persistence) UpdateEgressPolicy(ctx context.Context, actorRef resources
...
+func (p *Persistence) DeleteEgressPolicy(ctx context.Context, actorRef resources.ActorRef, pre store.DeletePreconditions) (*ateapipb.EgressPolicy, error) {
+	tx, err := p.pool.Begin(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("beginning egress policy delete: %w", err)
+	}
+	defer tx.Rollback(ctx) //nolint:errcheck // no-op once committed
+
+	// Lock and read the current policy, refuse the delete if the caller
+	// pinned a different uid or version, then delete that same row.
+	deleted, err := getEgressPolicyRow(ctx, tx, `
```
haiyanmeng: `deleted` is very confusing since this line does not delete anything. Let us rename it `current`. 

We should also rename the `deleted` variable in `DeleteWorker`. 

@juli4n , WDYT?

ygao-g: Done, renamed to current in both places., and put The DeleteWorker rename in a separate last commit, so it is easy to drop if @juli4n would rather leave the worker path alone.

haiyanmeng (approved):

cmd/ateapi/internal/store/atepg/worker.go
```diff
@@ -147,11 +147,11 @@ func (p *Persistence) DeleteWorker(ctx context.Context, name string, pre store.D
 	return p.writeAndAppendEvent(ctx, store.WorkerEventDeleted, func(ctx context.Context, tx pgx.Tx) (*ateapipb.Worker, error) {
 		// Locked rather than plainly read so the incarnation pre was evaluated
 		// against is the one the DELETE removes.
-		deleted, err := getWorkerRowForUpdate(ctx, tx, name)
+		current, err := getWorkerRowForUpdate(ctx, tx, name)
```
bowei: The change in this file doesn't actually do anything as far as I can tell. Can you remove it.

ygao-g: Reverted the changes, thx!

cmd/ateapi/internal/store/atepg/atepg.go
ygao-g: Left the var names in `DeleteWorker` unchanged.

bowei (dismissed):
There are merge conflicts

ygao-g:
> There are merge conflicts

Resolved it.

bowei:
merge conflicts?

bowei (approved):

ygao-g:
Heard from @juli4n that have a PR already under review in https://github.com/agent-substrate/substrate/pull/1807, aiming to support this for all resources. 

Will close this one and wait for theirs.

Issue #1677, opened 2026-09-16
atelet: stale ateom directories are never garbage collected; every deleted worker pod leaks its socket dir onto the discovery scan


baizhenyu:
### What happened?

Nothing ever removes `/var/lib/ateom-gvisor/ateoms/<worker-pod-UID>/` when a worker pod is deleted. Every ateom creates its directory (holding exactly one file, `ateom.sock`) at boot; the only cleanup anywhere in the tree is an ateom `RemoveAll`-ing its own stale socket when the *same* pod restarts. Directories accumulate until node replacement, and every pool rollout replaces all worker pods with fresh UIDs — so growth is rollout-driven and linear in node lifetime.

This was rational neglect, not a decision: repo archaeology (issues, PRs, commit history) found no discussion, because the directory set had no reader — atelet dialed sockets by exact UID from control-plane state, and a dead pod's leftover directory cost a few idle bytes. That changed when the stats sweep (#961) promoted the directory listing into its discovery registry. Since then every stale directory costs a dial+probe per sweep per node, debug-log noise, and (since #1479) a `cachedPools` entry. The only written acknowledgment is the comment that shipped with the sweep: "stale directories left by deleted worker pods (nothing garbage-collects them eagerly)". #1479's description references this issue as the durable fix.

### Expected Behavior

Stale ateom directories are garbage collected. Two complementary mechanisms:

**1. atelet janitor (the necessary mechanism — covers SIGKILL, OOM, node-crash leftovers).** New `cmd/atelet/ateomgc.go` modeled on the `imagegc.go` loop/flags/seams pattern. The evidence rule must make false positives structurally impossible, because the blast radius of wrongly deleting a live ateom's directory is severe and self-hiding: the bound unix listener keeps serving established connections on the unlinked inode with no error, but every NEW connect fails ENOENT forever (nothing re-binds the socket), while the worker keeps advertising capacity — a semi-permanent worker blackhole with loss of checkpoint/stop control over its actor. Per pass, per directory entry:

1. Names must parse as UUIDs; skip and warn anything else (operator debris is not ours to delete).
2. The janitor takes its OWN node-scoped pod LIST (`spec.nodeName` field selector, NO label selector — a pod that lost its pool label must not become deletion-eligible). A list ERROR aborts the whole pass: no strikes advance, nothing is deleted. This is deliberately not the stats sweep's `fetchWorkerPools`, whose nil-means-failed collapse is right for label enrichment and disqualifying for deletion evidence.
[...]

PR #1678, opened 2026-09-16
ateom: remove the ateom directory on graceful shutdown

baizhenyu:
Part of #1677 — the graceful-shutdown half; the atelet-side janitor for ungraceful exits is tracked there.

Nothing removes `ateoms/<podUID>/` when a worker pod goes away, so every pool rollout leaks one directory (holding a dead `ateom.sock`) per replaced worker — and since the stats sweep enumerates that directory as its discovery registry (#961), each leak costs a dial+probe per sweep, forever.

## Change

A `defer os.RemoveAll(ateomDir)` in both runtime mains, registered right after the `MkdirAll` that creates the directory. ~7 lines each plus the comment carrying the safety argument.

**Why this is safe:**
* **Same-pod container restart:** kubelet serializes containers — the replacement is not started until this process, and therefore this defer, has exited. The remove cannot race the successor's `MkdirAll`.
* **Shutdown ordering:** the SIGTERM path drains, `GracefulStop`s, `Serve` returns, `do` returns — the defer runs after the listener is already closed, so nothing can be dialing the socket it removes.
* **Boot-failure paths:** error returns before `Listen` also run the defer; the pod restarts and re-creates the directory. Harmless.

**What it deliberately does not cover:** SIGKILL after the grace period, OOM kills, node crashes — defers don't run there by nature. That residue is the janitor's job (#1677); this change shrinks the janitor's caseload to exactly those, since rollouts — the dominant growth driver — terminate gracefully.

## Testing

Both runtime mains are untested boot plumbing (no unit seam exists for `do()`); verified by linux builds and test-compiles of both packages.

**Live-verified on the ate-dev GKE cluster** (gVisor pool; the micro-VM change is the identical lines):

* Baseline census of one node found **11 ateom directories with exactly 1 live pod** — ten stale leftovers from prior rollouts, the leak in the wild.
* Rolled the pool onto an image built from this branch. The outgoing pre-change pod (`445af5aa…`) **leaked its directory** on termination, reproducing the bug side-by-side with the fix.
* Gracefully deleted a pod running this branch (`5618282c…`): its directory **was removed** on shutdown, while the pre-change pod's leaked directory remained in the same census — before/after behavior on the same node, same listing.


cmd/ateom-gvisor/main.go
```diff
@@ -174,6 +174,12 @@ func do(ctx context.Context) error {
 	if err := os.MkdirAll(ateomDir, 0o700); err != nil {
 		return fmt.Errorf("in os.MkdirAll(%q): %w", ateomDir, err)
 	}
```
JeffLuoo, line 179: `*podUID` is never validated, and this defer turns an invalid value destructive. `ateompath.AteomPath("")` returns `AteomsDir()` itself, so a process that starts without the flag deletes `/var/lib/ateom-gvisor/ateoms` on exit, together with the socket directory of every other ateom on the node. Before this change an empty UID was harmless: `MkdirAll` just created `ateoms/` and the socket landed at `ateoms/ateom.sock`.

```suggestion
	if err := resources.ValidateAteomUID(*podUID); err != nil {
		return fmt.Errorf("in resources.ValidateAteomUID: %w", err)
	}
	if err := os.MkdirAll(ateomDir, 0o700); err != nil {
		return fmt.Errorf("in os.MkdirAll(%q): %w", ateomDir, err)
	}
```

Same applies to `cmd/ateom-microvm/main.go:160`. Move the `ateomDir :=` line below the check if you prefer validation first; the suggestion above keeps the diff minimal.

Alternatively, add a check in defer to so we do safe removal of directory on shutdown.

JeffLuoo (changes requested):

cmd/ateom-gvisor/main.go
baizhenyu: Right — the defer turned a harmless misconfiguration destructive; empty was the worst case but the validator also closes the path-shaped ones. Applied your suggestion verbatim in both mains (e935a49f), failing at boot with the same check the atelet RPC boundary uses — a bad UID means atelet can't dial this ateom anyway, so fail-fast beats a guarded remove.

JeffLuoo (dismissed):

cmd/ateom-gvisor/main.go
```diff
@@ -171,9 +171,18 @@ func do(ctx context.Context) error {
...
+	}
 	if err := os.MkdirAll(ateomDir, 0o700); err != nil {
 		return fmt.Errorf("in os.MkdirAll(%q): %w", ateomDir, err)
 	}
+	// Clean up the ateom directory during graceful shutdown (#1677).
+	defer func() {
+		if err := os.RemoveAll(ateomDir); err != nil {
+			slog.ErrorContext(ctx, "Failed to remove the ateom directory on shutdown", slog.Any("err", err))
+		}
+	}()
```
BenTheElder, line 185: 🤖 **nit** 🟢 – One exit path skips this defer. At `main.go:272` a `svr.Serve` error calls `os.Exit(1)`, which runs no defers, so the directory is left behind along with the tracer and meter shutdowns. `cmd/ateom-microvm/main.go` returns the error on the same path and does clean up.

Pre-existing and only reachable on a permanent listener or accept failure, but it is the one case the PR description's "error returns also run the defer" does not cover. Returning the error instead of exiting would make the two binaries match.

baizhenyu: Good catch — that was the one exit that skipped the defer, and it also skipped the tracer/meter shutdowns. Returned the error instead, matching ateom-microvm (6f6d21e0).

BenTheElder (approved):

PR #1679, opened 2026-09-16
Add build-essential and use newer git in lima Linux VM

hdp617:
* Add the missing `build-essential` pkg.
* Use newer git version in lima Linux VM.

The bundled version in the guest image doesn't support ref storage backend. As a result, `./hack/microvm-assets/assemble.sh` fails with the following error.

```
fatal: unknown repository extension found:
	refstorage
```

- [ ] Tests pass


hdp617:
cc @BenTheElder

BenTheElder:
We should probably run this in a container instead FWIW.

hdp617:
> We should probably run this in a container instead FWIW.

Just to confirm, are you suggesting building the microvm assets in a Docker container?

BenTheElder:
> Just to confirm, are you suggesting building the microvm assets in a Docker container?

Yes, that seems better than depending on installing in the host. We can do that in a follow-up though.

BenTheElder (approved):

hdp617:
> > Just to confirm, are you suggesting building the microvm assets in a Docker container?
> 
> Yes, that seems better than depending on installing in the host. We can do that in a follow-up though.

I like it for hermetic builds. Will do in a follow-up.

hdp617:
I think this is no longer needed with https://github.com/agent-substrate/substrate/pull/1708. Will close once confirmed.

PR #1682, opened 2026-09-16
ateomnet: replace stale named namespaces

BenTheElder:
Remove mounted and unmounted leftovers before namespace creation. Reject invalid names and avoid following symlinks during unmount. Existing namespace handles remain valid.

Tiny pre-req for #1266

Also considering anonymous namespaces or per-activation IDs but eh we can revisit that later.


BenTheElder:
cc @EItanya @bowei @LiorLieberman

EItanya (commented):
🤖 AI-generated review.

No correctness issues found in `06089cb5`. Two optional improvements:

1. **Clarify namespace ownership** — [net.go:387–393](https://github.com/agent-substrate/substrate/blob/06089cb5f9628de7ada20f6d30ebbb7f0e06d7a6/internal/ateomnet/net.go#L387-L393) describes per-actor activation, but both current callers create the namespace at worker startup using the pod UID. Document that lifecycle and the caller’s responsibility to own the name exclusively.

2. **Test retained handles** — [net_linux_test.go:482–486](https://github.com/agent-substrate/substrate/blob/06089cb5f9628de7ada20f6d30ebbb7f0e06d7a6/internal/ateomnet/net_linux_test.go#L482-L486) closes the original handle before replacement. Keeping it open and verifying it remains usable would cover the PR’s stated guarantee.

All eight package tests passed locally under root. An additional probe confirmed retained handles remain usable, the replacement is distinct, and the caller’s namespace is restored. CI is green; cluster E2E was not rerun locally.

BenTheElder:
Good points, updated.

EItanya (approved):

PR #1684, opened 2026-09-16
fix(atelet): Supersede stale system-info volume registrations in atelet

shrutiyam-glitch:
When a worker pod crashes or is deleted out-of-band, `ateapi` transitions the actor to `CRASHED` and clears `worker_assignment` without calling `atelet.Terminate`. If the actor is later recovered and scheduled onto a new worker pod on the same node, `systemInfoVolumeRefresher.Register` panicked on the leftover entry in r.actors.

Supersede any existing registration for the same actor UID by marking the previous entry stale under its mutex and replacing it in r.actors.

Temporary fix for  #1710 

- [X] Tests pass


cmd/atelet/systeminfovolume.go
```diff
@@ -105,22 +105,23 @@ func newSystemInfoVolumeRefresher(lister certlisters.ClusterTrustBundleLister, i
...
-		r.actors[actorUID] = actor
-	}
+	prev := r.actors[actorUID]
+	r.actors[actorUID] = actor
 	r.mu.Unlock()
-	if dup {
-		panic(fmt.Sprintf("system-info volumes: actor %s registered twice", actorUID))
+	if prev != nil {
+		prev.mu.Lock()
+		prev.stale = true
```
thompsonmax, line 123: Can we add an info log for when we hit this? This should be fairly rare, it would be good to track how often this occurs in practice

cmd/atelet/systeminfovolume.go
```diff
@@ -105,22 +105,23 @@ func newSystemInfoVolumeRefresher(lister certlisters.ClusterTrustBundleLister, i
...
 	r.mu.Lock()
-	_, dup := r.actors[actorUID]
-	if !dup {
-		r.actors[actorUID] = actor
-	}
+	prev := r.actors[actorUID]
+	r.actors[actorUID] = actor
 	r.mu.Unlock()
-	if dup {
-		panic(fmt.Sprintf("system-info volumes: actor %s registered twice", actorUID))
```
thompsonmax, line 123: I'm a bit worried that with this change, a valid Deregister call (such as those in cmd/atelet/main.go:509 and :1119) would race with a second registration and cause the new registration to get erased.

I think the solution to this is to track the "registration" somehow after its created, so Deregister knows not to delete any second registration.

Example diff:

```
+// registration is the entry Register created. Undo removes that entry and
+// nothing else, so a caller that fails after registering cannot drop a
+// newer registration for the same actor.
+type registration struct {
+     r     *systemInfoVolumeRefresher
+     actor *registeredActor
+}
+
+// Undo drops the registration unless a later Register superseded it.
+// Safe on a nil handle so callers can defer it before Register runs.
+func (h *registration) Undo() {
+     if h == nil {
+             return
+     }
+     h.r.drop(h.actor)
+     h.actor.mu.Lock()
+     h.actor.stale = true
+     h.actor.mu.Unlock()
+}
+
+// drop removes actor from the registry if it is still the live entry.
+func (r *systemInfoVolumeRefresher) drop(actor *registeredActor) {
+     r.mu.Lock()
+     if r.actors[actor.uid] == actor {
+             delete(r.actors, actor.uid)
+     }
+     r.mu.Unlock()
+}
```

```
 // Register records actorUID's system-info volumes and writes their contents
-// from current cluster state. If actorUID is already registered (for example
-// after a worker pod crash left a stale entry without Terminate), the previous
-// registration is superseded.
-func (r *systemInfoVolumeRefresher) Register(actorUID string, ref resources.ActorRef, volumes []*systemInfoVolume) error {
+// from current cluster state. A second Register for the same actorUID
+// supersedes the first, which stops receiving refreshes. A failed Register
+// leaves no entry behind.
+func (r *systemInfoVolumeRefresher) Register(actorUID string, ref resources.ActorRef, volumes []*systemInfoVolume) (*registration, error) {
      actor := &registeredActor{uid: actorUID, ref: ref, volumes: volumes}
      // Held until the initial write finishes so a refresh cannot interleave.
      actor.mu.Lock()
      defer actor.mu.Unlock()
@@
      if prev != nil {
              prev.mu.Lock()
              prev.stale = true
              prev.mu.Unlock()
      }

      for _, v := range volumes {
              if err := r.write(ref, actorUID, v); err != nil {
-                     return fmt.Errorf("while populating system-info volume %q: %w", v.Name, err)
+                     // actor.mu is already held here, so this cannot go through Undo.
+                     r.drop(actor)
+                     actor.stale = true
+                     return nil, fmt.Errorf("while populating system-info volume %q: %w", v.Name, err)
              }
      }
-     return nil
+     return &registration{r: r, actor: actor}, nil
```

```
--- a/cmd/atelet/main.go
+++ b/cmd/atelet/main.go
@@ func (s *AteomHerder) Run(
-     defer func() {
-             if err != nil {
-                     s.systemInfoVolumes.Deregister(actorUID)
-             }
-     }()
-     if err := s.systemInfoVolumes.Register(actorUID, actorRef, systemInfoVolumesFor(actorUID, req.GetSpec())); err != nil {
+     reg, err := s.systemInfoVolumes.Register(actorUID, actorRef, systemInfoVolumesFor(actorUID, req.GetSpec()))
+     if err != nil {
              return nil, err
      }
+     defer func() {
+             if err != nil {
+                     reg.Undo()
+             }
+     }()
      if err := s.prepareOCIBundles(ctx, actorUID, actorRef,
@@ func (s *AteomHerder) Restore(
-     // Undo the Register if the restore fails.
+     // Undo the Register if the restore fails. The prep leg writes reg and
+     // nothing reads it before g.Wait returns.
+     var reg *registration
      defer func() {
              if err != nil {
-                     s.systemInfoVolumes.Deregister(actorUID)
+                     reg.Undo()
              }
      }()
@@
-             if err = s.systemInfoVolumes.Register(actorUID, actorRef, systemInfoVolumesFor(actorUID, req.GetSpec())); err != nil {
+             if reg, err = s.systemInfoVolumes.Register(actorUID, actorRef, systemInfoVolumesFor(actorUID, req.GetSpec())); err != nil {
```

thompsonmax: I'm happy to tackle this myself in a follow-up too if you'd prefer that. I don't think it necessarily needs to block this fix.

shrutiyam-glitch: Thanks !

cmd/atelet/systeminfovolume.go
```diff
@@ -104,23 +105,38 @@ func newSystemInfoVolumeRefresher(lister certlisters.ClusterTrustBundleLister, i
...
-	_, dup := r.actors[actorUID]
+	prev := r.actors[actorUID]
+	dup := prev != nil && prev.ateomUID == ateomUID
 	if !dup {
 		r.actors[actorUID] = actor
 	}
 	r.mu.Unlock()
 	if dup {
-		panic(fmt.Sprintf("system-info volumes: actor %s registered twice", actorUID))
+		panic(fmt.Sprintf("system-info volumes: actor %s registered twice on worker %s", actorUID, ateomUID))
```
thompsonmax: Hmm did you mean to reintroduce the panic on duplicates here?

thompsonmax: IIUC I think this can still cause these cases to panic:

- The RESUMING retry on main. When ateapi retries a resume for an actor still RESUMING, ensureWorkerAssigned reuses the persisted assignment, so the retry goes to the same worker pod and carries the same ateom UID. A version conflict in finalizeRunning after a successful Restore is enough to trigger it, and routed requests retry ResumeActor on their own. Under the current PR that retry kills atelet for the whole node.
- A crash that leaves the worker pod alive. crashActor releases only this actor's assignment through ReleaseActorFromWorker, so the worker stays in the pool. A Checkpoint that fails with a crash directive leaves the registry entry behind, because Checkpoint only deregisters after ateom succeeds. After RevertActor, the resume can be scheduled back onto that same worker. Same actor UID, same ateom UID, panic. Once PR 1675 lands this is a user-reachable atelet crash.

Let's stick with just the log line on duplicate registrations? The ateom UID seems useful to have in the log, it seems it being the same would mean that ateapi asked a worker to run an actor it had already started.

shrutiyam-glitch: Okay, sure 
so, i added ateomUID to distinguish legitimate recovery on a new worker pod (prev.ateomUID != ateomUID, where the old pod vanished without `Terminate`) from double-registering on the exact same worker pod (prev.ateomUID == ateomUID), was trying to preserve the original invariant check for the latter.

thompsonmax (approved):

PR #1687, opened 2026-09-16
CI: Emit JUnit from test runners, pin their exec bounds

### LINKED ISSUE #1685: Test Runners should always Emit JUnit
Test runners fial to emit JUnit causing all `exit 0` pathways to appear "green". Emitting JUnit will enable skip/fail/pass differentiation and can be consumed by downstream gate more predictably.

stshrive:
Today CI depends on process exit codes. This leads to a few issues with reporting results. For example, a green build cannot be distinguished from one where no tests ran at all; a `-run` filter that matches nothing; an e2e suite short-circuiting without `--e2e`; or a testcontainer skip: all currently exit 0.
    
This change adds gotestsum as a pinned tool module allowing opt-in for JUnit reporting in both test runners by using `E2E_JUNIT_FILE` and `ROOT_JUNIT_FILE` respectively. When left unset, the runners exec `go test` directly and gotestsum is never built leaving local runs unchanged for now.
    
Additionally, this change pins two execution bounds that `go test` would otherwise infer. Both remain overridable via environment variables as listed below.
    
- `-p`: system defaults didn't align with cluster capacity so   pinning the value avoids over allocation. (override with `E2E_PARALLELISM`).
    
- `-timeout`: Suite defaults conflicted with some E2E test timings   causing some tests that are expected to timeout at the same time to abort rather than fail as expected. Setting an 30m default avoids the conlficting timeouts producing clean failure signals (override with `E2E_TIMEOUT`).

Fixes #1685


hack/run-e2e.sh
```diff
@@ -104,5 +119,45 @@ if [[ -n "${KUBECTL_CONTEXT:-}" ]]; then
...
 fi
 
-exec go test -v "$target_path" ${go_test_args[@]+"${go_test_args[@]}"} -args --e2e ${extra_e2e_args[@]+"${extra_e2e_args[@]}"} ${e2e_args[@]+"${e2e_args[@]}"}
+# Pin the two bounds go test would otherwise infer from the machine.
+#
+# -p: the default is GOMAXPROCS, so suite concurrency tracks the runner's CPU
+# count rather than what the single-node kind cluster can absorb. The target
+# expands to 12 suite packages; on a larger runner they would all start at once.
+#
+# -timeout: the default is 10m, which is exactly TemplateReadyTimeout for the
```
annapendleton: default says 10m here, but it actually looks like 30m in the code - we should correct the documentation

Timeout looks like new behavior being introduced which could impact dev velocity if it's too low. 30m I'm less concerned about than the 10m as it's less aggressive

stshrive: I see the confusion related to this comment. This code is correcting for not having previously specified anything. The default, _when unspecified_, is 10m leading to the issue outlined here. I will update the language here to make sure that aligns.

stshrive: I can also lower it to something more like `12m` to `15m`. The gist of this change is to ensure that the two timeouts, `go test` and `TemplateReadyTimeout`, aren't exactly the same and that the latter is strictly less than the former.

stshrive: Updated the comments.

annapendleton (approved):

maxsmythe (approved):
LGTM

PR #1689, opened 2026-09-16
per sandbox networking

BenTheElder:
This is part of #1266 , opening now for discussion.

Stacked on https://github.com/agent-substrate/substrate/pull/1682 which was slightly orthogonal.

This is loosely based on the mini proposal by @howardjohn  as discussed in the community meeting, and feedback from @bowei @EItanya @LiorLieberman @aojea.

https://docs.google.com/document/d/1TycfQ3iiEpbI3rveMIj0S2PpPuLecb8I5R--yTpt9Ig/edit?resourcekey=0-kJbtEZ-KGzuL5eCjHDvBhg&tab=t.0#heading=h.ga9bfaf55ptk

Roughly:
1. Actor sandboxes each get their own netns. 
    - gVisor grabs all interfaces in the netns, and tap currently requires running something like slipr, so for now we do two netns + a veth when gVisor.
3. In the netns we directly intercept TCP => atunnel for general traffic.
4. In the netns we serve a trivial TCP+UDP DNS relay to the pod resolution. 
    - In the future we can insert policy here.
6. We consistently inject a modified resolv.conf instead of bind-mounting it (gVisor) across both runtimes.
7. Readiness probe dialing happens in the actor netns.

Every actor gets the same fixed guest IP as before, which is only visible to the actor.
All inbound/outbound traffic comes from atunnel / the DNS relay.
The actor no longer has any direct use of the pod interface, so we can begin to consider ateom using the network itself.
When we add the rest of multi-actor changes, this greatly simplifies thing.

Full multi-actor requires further changes, but this diff is already large (suggest reading commit by commit) and can stand-alone. I'll file more stacked changes when we've got consensus on this one.


EItanya (commented):
🤖 AI-generated review.

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,569 @@
...
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	return listeners, nil
+}
+
+// EgressServer serves one actor's captured connections. Satisfied by
+// atunnel.Egress; an interface so this package does not depend on it.
+type EgressServer interface {
```
EItanya: Could these interfaces be unexported (`egressServer` and `dnsServer`)? Neither is referenced by name outside `ateomnet`, and Go's implicit interface satisfaction lets callers pass `*atunnel.Egress` and `*atunnel.DNSRelay` to `ServeSandbox` without naming the interfaces. Their methods should stay exported so implementations in other packages can satisfy them.

`ServeSandboxEgress` and `ServeSandboxDNS` also have no callers outside this package, so those helpers could be unexported too, leaving `ServeSandbox` as the exported entry point.

cmd/ateom-gvisor/sandboxnet.go
```diff
@@ -0,0 +1,135 @@
...
+
+// prepareSandboxNetwork builds the actor's network and starts serving it.
+func (s *AteomService) prepareSandboxNetwork(ctx context.Context, actorUID string) error {
+	// A previous actor's session can still be here if its teardown never ran,
+	// and overwriting it would leak both namespaces, their /run/netns mounts,
+	// the egress listener and the DNS goroutines.
+	if err := s.releaseSandboxNetwork(ctx); err != nil {
+		return fmt.Errorf("while releasing the previous sandbox network: %w", err)
+	}
+	session, err := ateomnet.ServeSandbox(ctx, ateomnet.SandboxNetworkConfig{
```
EItanya, line 38: **Preserve the documented behavior when no egress gateway is configured.**

[gVisor setup](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/cmd/ateom-gvisor/sandboxnet.go#L43) and [microVM setup](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/cmd/ateom-microvm/sandboxnet.go#L41) now always redirect TCP into atunnel, but `prepareActorEgress` still returns nil for an absent gateway and `activateActorNetworking` then skips activating egress. The inactive handler closes every intercepted connection. The API server still defaults the gateway address to empty, and the [protocol contract](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/internal/proto/ateompb/ateom.proto#L137) explicitly promises direct egress in that case.

Could we preserve this supported mode, or explicitly require a gateway and update the configuration contract? At present an actor can start successfully with all external TCP connectivity broken. This was traced through both runtimes; I did not run a complete boot without a gateway.

internal/atunnel/dns.go
```diff
@@ -0,0 +1,302 @@
...
+				return nil, err
+			}
+			buf := make([]byte, maxDNSDatagram)
+			n, err := conn.Read(buf)
+			if err != nil {
+				return nil, err
+			}
+			return buf[:n], nil
+		}()
+		if err != nil {
```
EItanya, line 237: **Preserve DNS failover on recoverable server errors.**

[exchangeUDP](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/internal/atunnel/dns.go#L236) returns the first received response, including `SERVFAIL`. Since the guest now sees only the gateway as its nameserver, its resolver can no longer try the pod's backup nameserver itself.

I reproduced this on the reviewed head using a real Go resolver and two loopback DNS servers: the first returned `SERVFAIL`, the second served a valid A record. Lookup failed; the first server received two queries and the healthy backup received zero. A diagnostic retry change made the same probe succeed. Could the relay try the next upstream on recoverable DNS errors before returning the failure?

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,569 @@
...
+		if err != nil {
+			return nil, err
+		}
+		session.sockets = append(session.sockets, served...)
+	}
+	return session, nil
+}
+
+// Close stops serving the sandbox and removes its namespaces.
+func (s *SandboxSession) Close(ctx context.Context) error {
```
EItanya, line 532: **Make the session shutdown contract explicit.**

[SandboxSession.Close](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/internal/ateomnet/sandbox.go#L536) accepts a context but ignores it, discards socket-close errors, and does not wait for serving goroutines. [DNS cancellation](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/internal/ateomnet/sandbox.go#L390) is represented by the first element of an ordered `[]io.Closer`, while egress draining is handled separately by runtime calls to `atunnel.Deactivate`.

The new wait group inside `DNSRelay.Serve` does not give `SandboxSession.Close` a completion guarantee because its caller launches it in an untracked goroutine. Could the session retain explicit cancellation and completion state and document the teardown order? Either use `Close(ctx)` for a bounded drain, or remove the unused context and document that shutdown is asynchronous. This is a lifecycle-contract concern; I have not reproduced a production deadlock.

internal/ateomnet/sandbox_linux_test.go
```diff
@@ -0,0 +1,504 @@
...
+				t.Errorf("unexpected interface %q in the actor namespace", l.Attrs().Name)
+			}
+		}
+		return nil
+	}); err != nil {
+		t.Fatalf("listing links: %v", err)
+	}
+}
+
+// A micro-VM snapshot freezes the guest's ARP entry for its gateway, so the
```
EItanya: **Add focused tests for the actual TAP and DNS protocol paths.**

The [fixed-MAC test](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/internal/ateomnet/sandbox_linux_test.go#L340) explains the microVM restore requirement but uses `Veth=true` and examines `atside`. Production microVM uses [setupActorTap](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/cmd/ateom-microvm/net.go#L54), which has no focused test. The no-veth test only verifies that the namespace contains loopback. Could we test TAP recreation, its fixed MAC/gateway address, and returned queue descriptors directly?

The [DNS forwarding tests](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/internal/atunnel/dns_test.go#L120) mostly exchange incomplete DNS headers or arbitrary strings such as `query` and `answered`. TCP tests cover connection limits and cancellation without successfully forwarding a DNS query. Valid exchanges through a real resolver client, including UDP truncation/TCP fallback and SERVFAIL retry, would protect the protocol behavior. CI covers both runtimes and is green; these are targeted regression-coverage requests.

cmd/ateom-gvisor/sandboxnet.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	s.netMu.Lock()
+	s.network = session
+	s.netMu.Unlock()
+	return nil
+}
+
+// releaseSandboxNetwork stops serving the actor and takes its network down.
+func (s *AteomService) releaseSandboxNetwork(ctx context.Context) error {
```
EItanya, line 59: **Share the common session-management glue.**

The [gVisor](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/cmd/ateom-gvisor/sandboxnet.go#L66) and [microVM](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/cmd/ateom-microvm/sandboxnet.go#L58) adapters repeat session replacement, locking, dialer closures, nil-session handling, readiness aliases, and port parsing. The latest stale-session cleanup fix already required identical edits in both prepare methods.

Optional cleanup: share the small runtime-independent session switching/dialing portion and port parser, keeping resolver-file handling and TAP/veth setup local. The `readyzDialer` wrappers merely return `sandboxDialer` and can be removed.

cmd/ateom-microvm/sandboxnet.go
```diff
@@ -0,0 +1,127 @@
...
+	// and overwriting it would leak both namespaces, their /run/netns mounts,
+	// the egress listener and the DNS goroutines.
+	if err := s.releaseSandboxNetwork(ctx); err != nil {
+		return fmt.Errorf("while releasing the previous sandbox network: %w", err)
+	}
+	session, err := ateomnet.ServeSandbox(ctx, ateomnet.SandboxNetworkConfig{
+		ActorUID:      actorUID,
+		EgressPort:    s.atunnelEgressPort,
+		DNSPort:       atunnel.DNSPort,
+		GatewayHWAddr: gatewayHWAddr,
```
EItanya: **Remove ineffective configuration and misleading state.**

MicroVM passes [GatewayHWAddr](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/cmd/ateom-microvm/sandboxnet.go#L45), but `SetupSandboxNetwork` consumes it only when `Veth=true`; microVM leaves that false. The actual MAC assignment is inside `setupActorTap`, while production gVisor leaves the option unset. The option therefore has no effect in current production callers despite appearing to configure a restore-critical property.

[PodSideIP](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/internal/ateomnet/sandbox.go#L56) is always the actor IP, is inaccessible from the pod namespace, and has no production reader. Optional cleanup: remove the ineffective option and unused field, leaving MAC ownership in the code that actually sets it.

internal/atunnel/ingress.go
```diff
@@ -417,14 +434,41 @@ func (s *Server) Activate(atespace, actorName string) error {
 		return fmt.Errorf("atunnel: actor %s is already active", s.active.ref)
 	}
 	ctx, cancel := context.WithCancel(context.Background())
+	dial := activationDialer(ctx, s.dial)
+	proxy := *s.proxy
+	if _, ok := proxy.Transport.(protocolMirrorTransport); ok {
```
EItanya: **Make per-activation transport construction explicit.**

[Activate](https://github.com/agent-substrate/substrate/blob/554332a0e1d49212c6ce7635c888e2dc22a9d416/internal/atunnel/ingress.go#L439) creates a fresh transport and activation-bound dialer only when the template transport's concrete type is exactly `protocolMirrorTransport`. The alternate branch accommodates transports installed directly by tests.

Current production construction satisfies the condition, so this is not a demonstrated isolation failure. As an optional cleanup, could we construct the per-activation transport explicitly and use a small internal factory if tests need substitution? That would make the activation invariant clearer than the concrete-type branch.

internal/atunnel/dns.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	for _, u := range upstreams {
+		if _, _, err := net.SplitHostPort(u); err != nil {
+			return nil, fmt.Errorf("atunnel: invalid upstream resolver %q: %w", u, err)
+		}
+	}
+	d := &net.Dialer{Timeout: dnsExchangeTimeout}
+	return &DNSRelay{
+		upstreams:   upstreams,
+		dial:        d.DialContext,
```
EItanya: Claude loves callback test seams, any reason not to just put the `net.Dialer` on here for clarity?

EItanya (commented):
🤖 mostly AI generated, reviewed by me :)

Sorry for the double review, the agent initially didn't understand that I wanted to append the first review comment to these...

internal/atunnel/dns.go
BenTheElder: Yes. Good catch.

cmd/ateom-gvisor/sandboxnet.go
BenTheElder: After some side discussions: I think we want to change this to clearly state that egress is not supported without a gateway, but a gateway is not strictly required (perhaps your actor does not need egress).

If we really need it later we can add a no-capture mode, but it greatly simplifies things to always use atunnel, and atunnel will need an egress gateway for egress.

internal/ateomnet/net.go
```diff
@@ -170,6 +170,36 @@ func CleanupActorNetwork(ctx context.Context, interiorNetNS netns.NsHandle) erro
...
+// namespace, remounting /proc/sys read-write when the runtime bind-mounted it
+// read-only. A no-op when it already reads that way.
+func setNetSysctl(key, value string) error {
+	path := "/proc/sys/" + key
+	if b, err := os.ReadFile(path); err == nil && strings.TrimSpace(string(b)) == value {
+		return nil
+	}
+	if err := os.WriteFile(path, []byte(value+"\n"), 0o644); err == nil {
+		return nil
+	}
```
bowei, line 101: you should check EROFS explicitly?

internal/ateomnet/resolvconf.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	root, err := os.OpenRoot(rootfs)
+	if err != nil {
+		return fmt.Errorf("opening rootfs %q: %w", rootfs, err)
+	}
+	defer root.Close()
+	if err := root.Mkdir("etc", 0o755); err != nil && !errors.Is(err, fs.ErrExist) {
+		return fmt.Errorf("creating %q: %w", filepath.Join(rootfs, "etc"), err)
+	}
+	if err := root.Remove("etc/resolv.conf"); err != nil && !errors.Is(err, fs.ErrNotExist) {
```
bowei, line 60: Prefer write out file and then use atomic rename to replace file contents.

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+	"github.com/google/nftables/binaryutil"
+	"github.com/google/nftables/expr"
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
+	"golang.org/x/sys/unix"
+)
+
+// SandboxNetwork holds a sandbox's runtime and gateway namespaces.
+type SandboxNetwork struct {
+	// ActorUID names the namespaces, so a teardown can find them again.
```
bowei: Can we be specific -- this is the UID from the Actor resource stored in the DB

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+	ActorUID string
+	// RuntimeNetNS is what the sandbox runs in. gVisor claims every interface
```
bowei: Is this specific to gVisor, maybe call it the gVisor netNS

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+	RuntimeNetNS netns.NsHandle
+	// GatewayNetNS holds atunnel's sockets and the kernel end of the link.
+	// For microVMs it shares RuntimeNetNS.
+	//
+	// TODO: we hope gVisor can take that same single-namespace shape soon,
+	// once runsc can be given one interface rather than claiming every
+	// interface in the namespace it runs in.
+	GatewayNetNS netns.NsHandle
+	// PodSideIP is identical across sandboxes, isolated by namespace.
+	PodSideIP net.IP
```
bowei: This is == PodIP?

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+	Veth bool
+
+	// EgressPort is where atunnel serves this actor. Every TCP connection the
+	// actor makes is redirected to it, whatever port it was aimed at.
+	EgressPort uint16
+
+	// GatewayHWAddr fixes the gateway's MAC, which a micro-VM snapshot freezes
+	// into the guest's ARP cache. gVisor re-ARPs and can leave it unset.
+	//
+	// Applies to the veth path only. The tap path sets its own MAC in
```
bowei: This comment is confusing. Is this only for gvisor path or both (text says basically both at the same time).

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+	// setupActorTap, after LinkAdd, because tuntap creation ignores the
+	// hardware address in the link attributes. Both belong here once the two
+	// runtimes share one shape.
+	GatewayHWAddr net.HardwareAddr
+}
+
+// SetupSandboxNetwork creates isolated networking with fixed sandbox addresses.
+// gVisor uses a veth pair across runtime and gateway namespaces because it takes
+// over every interface in its namespace. MicroVMs use a tap in one namespace.
+// Gateway nftables redirect TCP egress to atunnel, preserving SO_ORIGINAL_DST.
```
bowei: we may want to have separate sock for DNS (tcp) so the DNS vs non-DNS can be separated

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+	defer func() {
+		if retErr != nil {
+			actorNS.Close()
+			_ = removeNamedNetNS(actorName)
+		}
+	}()
+
+	// Without a veth, atunnel shares the namespace with the runtime's tap.
+	atunnelNS := actorNS
+	if cfg.Veth {
```
bowei, line 104: can we factor the if Veth leg into its own func

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+	c.AddRule(&nftables.Rule{Table: table, Chain: prerouting, Exprs: exprs})
+
+	if err := c.Flush(); err != nil {
+		return fmt.Errorf("while installing the actor egress redirect: %w", err)
+	}
+	return nil
+}
+
+// ipv4HeaderDst is the offset of the destination address in an IPv4 header.
+const ipv4HeaderDst = 16
```
bowei: move this into the func where it is used

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+
+func actorSubnet() *net.IPNet {
+	_, subnet, err := net.ParseCIDR(ActorVethSubnet)
```
bowei, line 265: Parse in init() and make this a var? We don't need to keep calling a path with panic() at runtime.

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+			if err := e.ServeFor(context.Background(), actorKey, l); err != nil {
+				slog.WarnContext(ctx, "Actor egress listener stopped",
+					slog.String("actorUID", actorKey), slog.Any("err", err))
+			}
+		}(l)
+	}
+	return listeners, nil
+}
+
+// DNSServer answers an actor's DNS. Satisfied by atunnel.DNSRelay; an interface
```
bowei: recommend put DNS stuff in dns.go

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+		case "tcp", "tcp4", "tcp6", "udp", "udp4", "udp6":
+		default:
+			return nil, net.UnknownNetworkError(network)
+		}
+		hostname, _, err := net.SplitHostPort(addr)
+		if err != nil {
+			return nil, err
+		}
+		if _, err := netip.ParseAddr(hostname); err != nil {
+			return nil, fmt.Errorf("sandbox dial requires an IP literal: %w", err)
```
bowei: sandbox dial => "NetNSDialer supports only IP literals (got %q): %w"

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+		}
+		type result struct {
+			conn net.Conn
+			err  error
+		}
+		done := make(chan result, 1)
```
bowei: minor: using WaitGroup does not result in heap allocation for the channel.

internal/ateomnet/sandbox_linux_test.go
BenTheElder: done

cmd/ateom-microvm/sandboxnet.go
BenTheElder: done

cmd/ateom-gvisor/sandboxnet.go
BenTheElder: done

internal/atunnel/dns.go
BenTheElder: done

internal/atunnel/ingress.go
BenTheElder: done

internal/atunnel/dns.go
BenTheElder: agreed, done

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+			host, err := netns.Get()
+			if err != nil {
+				// Nothing has switched yet, so the thread is still in the
+				// worker's namespace and is safe to hand back.
+				runtime.UnlockOSThread()
+				done <- result{err: fmt.Errorf("while reading the current netns: %w", err)}
+				return
+			}
+			defer host.Close()
+			if err := netns.Set(ns); err != nil {
```
bowei: Can we have netns.MustGet() and netns.MustSet() that panics on error? I'm assuming we won't hit the panics in normal operation unless there is a bug somewhere.

It would simplify the logic here.

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,496 @@
...
+		}
+		if _, err := netip.ParseAddr(hostname); err != nil {
+			return nil, fmt.Errorf("sandbox dial requires an IP literal: %w", err)
+		}
+		type result struct {
+			conn net.Conn
+			err  error
+		}
+		done := make(chan result, 1)
+		go func() {
```
bowei: This code is bit confusing to me -- why do we have to open and close the NS? Isn't the only that we are doing switching between them? I would have expected something like:

func doInNetNS(netns, do func()) error {
   runtime.LockOSThread()
   defer runtime.UnlockOSThread()

   oldNS := setNetNS(netns)
   defer setNetNS(oldNS)

   do()
}

internal/ateomnet/resolvconf.go
BenTheElder: We're doing this while nothing is reading it? I don't see the benefit. This only happens when the actor is not yet running.

internal/proto/ateletpb/atelet.proto
```diff
@@ -125,7 +125,7 @@ message RunRequest {
   // so a later Checkpoint can pin the same version into the snapshot manifest.
   SandboxAssets sandbox_assets = 8;
 
-  // When absent, actor traffic uses direct egress instead of atunnel.
+  // When absent the actor has no egress: its TCP is captured and refused.
   optional EgressGateway egress_gateway = 9;
```
haiyanmeng, line 129: Is it safe to drop the `optional` keyword?

BenTheElder: It's still optional?

internal/ateomnet/sandbox.go
BenTheElder: done

internal/proto/ateletpb/atelet.proto
BenTheElder: https://github.com/agent-substrate/substrate/pull/1689#discussion_r4041441624

internal/ateomnet/sandbox.go
BenTheElder: done

internal/ateomnet/sandbox.go
BenTheElder: The microVM tap lives here too, reworded.

internal/ateomnet/net.go
BenTheElder: done

internal/ateomnet/sandbox.go
BenTheElder: cleaned up, this is dead now after other refactoring.

internal/ateomnet/sandbox.go
BenTheElder: cleaned up / refactored out.

internal/ateomnet/sandbox.go
BenTheElder: agreed, done.

internal/ateomnet/sandbox.go
BenTheElder: done in var assignment at the package level

internal/ateomnet/sandbox.go
BenTheElder: done.

internal/ateomnet/sandbox.go
BenTheElder: this is now `setupVethPair`

internal/ateomnet/sandbox.go
BenTheElder: done, verbatim.

internal/ateomnet/sandbox.go
BenTheElder: It's also a little bit nicer API. Done.

internal/ateomnet/sandbox.go
BenTheElder: DNS has both TCP and UDP sockets which are distinct from the atunnel socket. Clarifying this comment.

internal/proto/ateletpb/atelet.proto
haiyanmeng: https://github.com/agent-substrate/substrate/pull/1717 requires the `--egress-gateway-address` flag of `ate-api-server` to be non-empty. If the flag is always non-empty, the `egress_gateway` field should be non-empty. 

@bowei , WDYT?

internal/ateomnet/sandbox.go
BenTheElder: refactored, should be a little more readable.

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,623 @@
...
+// Gateway nftables redirect the actor's outbound TCP to atunnel, preserving
+// SO_ORIGINAL_DST. Traffic to the gateway address is not redirected, so DNS
+// over UDP and TCP reaches the relay's own sockets.
+func SetupSandboxNetwork(ctx context.Context, cfg SandboxNetworkConfig) (_ *SandboxNetwork, retErr error) {
+	actorUID := cfg.ActorUID
+	if actorUID == "" {
+		return nil, fmt.Errorf("actornet: actor UID is required")
+	}
+
+	actorName := ateompath.ActorNetNSName(actorUID)
```
haiyanmeng: `actorName` -> `actorNSName`

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,623 @@
...
+	// micro-VM's tap lives here, and gVisor claims every interface here and
+	// moves their addresses into its own stack.
+	RuntimeNetNS netns.NsHandle
+	// GatewayNetNS holds atunnel's sockets and the kernel end of the link.
+	// For microVMs it shares RuntimeNetNS.
+	//
+	// TODO: we hope gVisor can take that same single-namespace shape soon,
+	// once runsc can be given one interface rather than claiming every
+	// interface in the namespace it runs in.
+	GatewayNetNS netns.NsHandle
```
haiyanmeng, line 57: Naming this `GatewayNetNS` is very confusing, since this is for atunnel. How about this?

| Current | Proposed |
|---|---|
| `SandboxNetwork.GatewayNetNS` | `SandboxNetwork.AtunnelNetNS` |
| `SandboxGatewayNetNSName` | `SandboxAtunnelNetNSName` |
| `setupGatewaySide` | `setupAtunnelSide` |
| `gatewayVethName` | `atunnelVethName` |

internal/ateomnet/sandbox.go
BenTheElder: Good point, I think there are one or two more similar spots, making these more obvious.

internal/proto/ateletpb/atelet.proto
BenTheElder: Given 1717 isn't merged, I don't think we should change this here.

internal/ateomnet/sandbox.go
BenTheElder: It's not just atunnel, so I'm not sure that's clearer. We're also doing the DNS listener, and possibly soon an "upward API" socket.

I do hope we can align gVisor soon so there's just one netns and no veth.

internal/proto/ateletpb/atelet.proto
haiyanmeng: SG.

internal/ateomnet/sandbox.go
LiorLieberman: Point is that its not "Gateway" and using the word "gateway" is confusing? What is it beyond Atunnel?

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,530 @@
...
+
+	// Veth separates gVisor's interfaces from the kernel-owned gateway.
+	// MicroVMs use a tap in a single namespace instead.
+	Veth bool
+
+	// EgressPort is where atunnel serves this actor. Every TCP connection the
+	// actor makes is redirected to it, whatever port it was aimed at.
+	EgressPort uint16
+
+	// GatewayHWAddr fixes the gateway's MAC, which a micro-VM snapshot freezes
```
LiorLieberman: which Gateway?

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,530 @@
...
+	// setupActorTap, after LinkAdd, because tuntap creation ignores the
+	// hardware address in the link attributes. Both belong here once the two
+	// runtimes share one shape.
+	GatewayHWAddr net.HardwareAddr
+}
+
+// SetupSandboxNetwork creates isolated networking with fixed sandbox addresses.
+// gVisor uses a veth pair across runtime and gateway namespaces because it takes
+// over every interface in its namespace. MicroVMs use a tap in one namespace.
+// Gateway nftables redirect the actor's outbound TCP to atunnel, preserving
```
LiorLieberman: same here, what Gateway? the default-gateway?

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,530 @@
...
+}
+
+// CleanupSandboxNetwork closes namespace handles and removes their names.
+func CleanupSandboxNetwork(network *SandboxNetwork) error {
+	if network == nil {
+		return nil
+	}
+	var errs error
+	// Compare before Close sets the handle to -1; microVMs share one descriptor.
+	separate := network.GatewayNetNS != network.RuntimeNetNS
```
LiorLieberman: separateNs?

internal/ateomnet/sandbox.go
```diff
@@ -0,0 +1,530 @@
...
+func NetNSDialer(ns netns.NsHandle) func(context.Context, string, string) (net.Conn, error) {
+	return func(ctx context.Context, network, addr string) (net.Conn, error) {
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+		switch network {
+		case "tcp", "tcp4", "tcp6", "udp", "udp4", "udp6":
+		default:
+			return nil, net.UnknownNetworkError(network)
+		}
```
LiorLieberman: nit: could probably be abstracted to a func

internal/ateomnet/sandbox_atunnel_linux_test.go
LiorLieberman, line 1: whats the diff between sandbox_atunnel_linux vs sandbox_linux?

internal/ateompath/ateompath.go
```diff
@@ -82,6 +82,18 @@ func AteletOTLPSocketPath() string {
 
 // AteomsDir is the parent of every per-ateom directory. Each ateom creates
 // AteomPath(podUID) under it when it boots, so listing this directory is how a
+// ActorNetNSName is the named network namespace one actor's sandbox runs in.
+// Per actor rather than per pod, so a worker can hold several sandboxes whose
+// networks cannot see each other.
+func ActorNetNSName(actorUID string) string {
+	return "ateom-actor:" + actorUID
+}
```
LiorLieberman: enforce a limit + trim? I think its 255 chars max

internal/atunnel/dns.go
LiorLieberman, line 1: No way to block DNS traffic, correct? we should document it

cmd/ateom-gvisor/sandboxnet.go
LiorLieberman: plus 1 to ben

internal/ateomnet/dns.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// dnsServer answers an actor's DNS. Satisfied by atunnel.DNSRelay; an interface
+// so this package does not depend on it.
+type dnsServer interface {
+	ServePacket(ctx context.Context, pc net.PacketConn) error
+	Serve(ctx context.Context, listener net.Listener) error
+}
+
+// ServeSandboxDNS serves UDP and TCP DNS in the gateway namespace.
```
LiorLieberman: we need something more clear than "gateway" here imo

internal/ateompath/ateompath.go
```diff
@@ -110,15 +98,19 @@ func AteomSocketPath(podUID string) string {
 	)
 }
 
-func AteomNetNSName(podUID string) string {
-	return "ateom:" + podUID
+// ActorNetNSName names an actor's sandbox network namespace.
+func ActorNetNSName(actorUID string) string {
+	return "ateom-actor:" + actorUID
```
LiorLieberman, line 103: enforce a limit + trim? I think its 255 chars max

LiorLieberman (commented):
LGTM overall

internal/ateomnet/sandbox.go
haiyanmeng: @bowei , WDYT?

BenTheElder: keep in mind we will not have both of these once gvisor can adapt. `RuntimeNetNS` will be the only one. `GatewayNetNS` is only if we have to split them across the veth, we drop this one later.

I don't think `Atunnel` vs `Runtime` is clearer personally but I'll rename to a consensus.

internal/ateomnet/sandbox.go
BenTheElder: clarified.

internal/ateomnet/sandbox.go
BenTheElder: clarified

internal/ateomnet/sandbox_atunnel_linux_test.go
BenTheElder: avoids an import cycle, clarified.

internal/atunnel/dns.go
BenTheElder: documented

internal/ateomnet/dns.go
BenTheElder: clarified, these are the sandbox's default gateway, not the "egress gateway"

BenTheElder:
One naming bikeshed outstanding. We can revisit that monday.

Resolved the others and did some more passes for bugs / minor nits.

internal/ateompath/ateompath.go
LiorLieberman: un-resolve - so we dont need to that?

internal/ateompath/ateompath.go
BenTheElder: we depend on the actorUID length being capped low enough for unix socket (which is 103 byte) already. If we want to start defensively validating that inline I'd take that as a follow-up.

internal/ateomnet/sandbox.go
EItanya: > keep in mind we will not have both of these once gvisor can adapt

I agree with @BenTheElder here that there isn't really a great name for this thing since it's a temporary solution. FWIW `gateway` means something different in this context, and is correct, we're just thinking of it as an L7 gateway.

Can we just leave it for now since it's an internal impl detail that will be removed?

BenTheElder:
[trivial rebase on main]

EItanya (commented):
I'm giving this my LGTM, holding off on the official approval because there are open comments from other folks.

haiyanmeng (approved):

internal/ateomnet/net.go
```diff
@@ -21,30 +21,24 @@ import (
...
-	HostVethName      = "ateom0"
-	ActorVethName     = "eth0"
-	HostVethCIDR      = "169.254.17.1/30"
-	ActorVethCIDR     = "169.254.17.2/30"
-	ActorVethGateway  = "169.254.17.1"
-	ActorVethIP       = "169.254.17.2"
-	ActorNftTableName = "ateom_actor"
-	dnsPort           = 53
+	ActorVethName    = "eth0"
+	HostVethCIDR     = "169.254.17.1/30"
```
bowei: These CIDRs area bit funky -- once you apply the bitmask, this is just 

169.254.17.1/30 == 169.254.17.0/30

169.254.17.2/30 == 169.254.17.0/30

Did you want them to be disjoint?

internal/ateomnet/resolvconf.go
bowei: [nit] It's more of a "common pattern for writing files in Unix" kind of thing.

internal/ateomnet/net.go
BenTheElder: These are interface addresses with prefix length. The "CIDR" name might be confusing. The bitmask is for routing tables for ARP but .1 is the sandbox gateway and .2 is the actor, "out of a /30".

BenTheElder: IOW these values should be correct but they might need a better name.

bowei: Suggest name change and comment:

```
// hostVethLocalAddress passed to `ip address add`. Note: this is not an
// actual CIDR but the <ip address>/<netmask>.
hostVethLocalAddress = ".."
actorVethLocalAddress = ".."
```

internal/ateomnet/sandbox_linux_test.go
```diff
@@ -0,0 +1,571 @@
...
+		{"unix", "/tmp/socket"},
+	} {
+		if conn, err := NetNSDialer(-1)(context.Background(), target.network, target.address); err == nil {
+			_ = conn.Close()
+			t.Errorf("accepted %s %s", target.network, target.address)
+		}
+	}
+}
+
+func TestNetNSDialerDoesNotPinPendingThreads(t *testing.T) {
```
bowei: This one might be overkill and potentially flaky.

internal/atunnel/dns.go
```diff
@@ -0,0 +1,348 @@
...
+	// dialer reaches upstream resolvers from the worker namespace.
+	dialer *net.Dialer
+
+	// Limits are shared across all sandboxes using this relay.
+	inFlight    chan struct{}
+	connections chan struct{}
+}
+
+// NewDNSRelay forwards to upstreams, each "host:port".
+func NewDNSRelay(upstreams []string) (*DNSRelay, error) {
```
bowei, line 65: We should file a TODO to analyze this and make the code more robust. I am ok with getting this in for the moment.

bowei (dismissed):

internal/ateomnet/resolvconf.go
BenTheElder: I know the common pattern and I've implemented it many times but ...

1. We're only writing it when the actor is offline and nothing is reading this filesystem.
2. We have to be really careful with how we write into the container provided root filesystem to avoid symlink security bugs.

2\) Makes me disinclined to rewrite it just for the sake of the common pattern. We have no need for it here and the current implementation leveraging `os.Root` helps us avoid security bugs. That abstraction doesn't provide a `mv` equivilant.

internal/atunnel/dns.go
BenTheElder: Agreed. I found a bug or two in previous iterations, I expect this needs further hardening.

internal/ateomnet/sandbox_linux_test.go
BenTheElder: Yeah, this caught a regression during development but keeping it is wild overkill. Removing.

internal/atunnel/dns.go
BenTheElder: Tracking in https://github.com/agent-substrate/substrate/issues/1816

internal/ateomnet/resolvconf.go
BenTheElder: https://go.dev/blog/osroot

https://pkg.go.dev/os#Root

EItanya (approved):
Received 👍 from @bowei and @haiyanmeng so putting final stamp, 🎉

PR #1694, opened 2026-09-17
benchmarking/locust: Add sweperf trajectory-based benchmarking workload

### LINKED ISSUE #1692: benchmarking: add a workload that simulates a real agent trajectory
### Problem

The workloads in the benchmarking suite today are:
| Workload | What it tests |
|---|---|
| `glutton` | Memory: fills/churns/reads RAM via `WriteRAM`/`ReadRAM` |
| `durdir` | Durable directory: file write, digest read, restore after resume |
| `sleep` | Idle actor lifecycle |
| `usermem` / `kernelmem` | Targeted memory accounting |

Each isolates one variable, which is exactly what you want for
attributing a regression. But it means we currently have no measurement of
suspend/resume against a workload that looks like what Substrate actually runs:
performing a trajectory of tasks with a real dependency tree.

### Proposal

Add a `sweperf` user class that runs a recorded agent trajectory against a real
SWE-bench task image, suspending and resuming the actor between slices of the
trace.

Shape:

- Partition a recorded trajectory of N steps into C contiguous cycles.
- Per cycle: resume the actor, replay that cycle's slice through the router,
  suspend it again.
- Report per-cycle latency alongside the existing `SuspendActor` /
  `ResumeActor` metrics, so lifecycle cost can be read against a realistic
  working set.

Proposed knobs, following the existing `--durdir-*` convention:

| Flag | Meaning |
|---|---|
| `--sweperf-template` | ActorTemplate name |
| `--sweperf-total-steps` | Steps in the trace |
| `--sweperf-num-cycles` | Cycles to partition steps into |

Reported metrics: `Workload_Cycle_<n>` per cycle, plus the existing
`CreateActor` / `SuspendActor` / `ResumeActor`.

### Status

I have an implementation working end-to-end on a GKE dev cluster (gVisor) and
will open a PR!

pmandewalkar:
Fixes #1692 

Adds a SWE-Perf benchmarking workload to the benchmarking suite, exercising
actor suspend/resume against a real SWE-bench task image (`astropy-7336`)
rather than a synthetic load.

The workload runs a recorded agent trajectory in configurable cycles,
suspending and resuming the actor between each, so the benchmark measures
lifecycle cost under a realistic in-sandbox server.

### What's here

| Area | Change |
|---|---|
| `internal/benchmarking/boomer/sweperf/` | New `SweperfUser` boomer user class (+ tests) |
| `internal/benchmarking/boomer/dynconfig/` | `sweperf_template`, `sweperf_total_steps`, `sweperf_num_cycles` knobs |
| `benchmarking/locust/common/sweperf_config.py` | `--sweperf-*` Locust flags |
| `benchmarking/locust/tests/sweperf.py` | Stub user class so the master attributes boomer's stats rows |
| `benchmarking/workloads/manifests/` | `swebench-astropy-7336` ActorTemplate |

Defaults: 21 trace steps partitioned into 4 cycles.

### Verification

- `hack/verify-all.sh` — pass
- `go test -race ./internal/benchmarking/... ./cmd/benchmarking/...` — pass
- End-to-end on a GKE dev cluster (gVisor), 1 user:

```
state: running | users: 1 | fail_ratio: 0.0

NAME                     REQ  FAIL   AVG_ms
CreateActor                1     0        1
CreateAtespace             1     0        1
ResumeActor               20     0      495
SuspendActor              20     0      947
Workload_Cycle_1           5     0     1996
Workload_Cycle_2           5     0      964
Workload_Cycle_3           5     0     4877
Workload_Cycle_4           5     0     3022
```

164 requests total, 0 failures, 0 errors.

The ActorTemplate image is currently pinned to a personal Artifact Registry
repo; a shared public registry for these SWE-Perf task images is planned.

- [x] Tests pass


benchmarking/locust/tests/sweperf.py
```diff
@@ -0,0 +1,46 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Stub SweperfUser declaration.
+
+The real load implementation lives in the boomer-Go worker at
+cmd/benchmarking/boomer-worker/; this Python class is declared only so the
```
maxsmythe: nit: sweperf, not boomer-worker

benchmarking/workloads/manifests/swebench-astropy-7336-template.yaml.tmpl
```diff
@@ -0,0 +1,42 @@
...
+  matchLabels:
+    workload: benchmark-ateom
+containers:
+- name: astropy
+  # Fill this in before deploying. The task image is built from the SWE-Perf
+  # repo, which supplies both the server entrypoint and the trajectory under
+  # test; the command below runs that entrypoint against that trace, so the
+  # image has to be the one built for this task. A shared public registry of 
+  # these images is planned; for now this is our working version. 
+  image: us-central1-docker.pkg.dev/nishanth-gke-dev/agent-sandbox/swe-astropy-7336@sha256:944924a9a4571c0fac72ac5d1c9be71ff5e98ef2cdcd1034ceaa48d47f755ced
```
maxsmythe, line 28: for security, this should point to an official repo

internal/benchmarking/boomer/sweperf/sweperf.go
```diff
@@ -0,0 +1,699 @@
...
+// chunk partitions totalSteps instructions into numCycles slices.
+type chunk struct {
+	start int
+	end   int
+}
+
+// generateDynamicChunks splits totalSteps into numCycles contiguous chunks.
+// The remainder is spread over the earliest cycles, so chunk lengths differ by
+// at most one.
+func generateDynamicChunks(totalSteps int, numCycles int) []chunk {
```
maxsmythe, line 87: side comment (no need to implement)

If we wanted we could just generate the next chunk size dynamically during iteration and avoid creating a list. Probably not worth doing at the expected user counts and chunk counts.

internal/benchmarking/boomer/sweperf/sweperf.go
```diff
@@ -0,0 +1,699 @@
...
+		} else if val := os.Getenv("WORKLOAD_REPO"); val != "" {
+			template = val
+		} else {
+			template = defaultSweperfTemplate
+		}
+	}
+
+	totalSteps := dyn.SweperfTotalSteps
+	if totalSteps <= 0 {
+		if val := os.Getenv("TOTAL_STEPS"); val != "" {
```
maxsmythe: Why are we inheriting config from either environment or dynamic config?

internal/benchmarking/boomer/sweperf/sweperf.go
```diff
@@ -0,0 +1,699 @@
...
+	if numCycles <= 0 {
+		numCycles = defaultSweperfNumCycles
+	}
+
+	return template, totalSteps, numCycles
+}
+
+// dynamicWait is the think time between cycles: a uniform draw from
+// [MinWait, MaxWait), or MinWait when the range is empty.
+func (r *sweperfRuntime) dynamicWait() time.Duration {
```
maxsmythe, line 163: Can we turn this into a shared function? I think every class uses a similar function.

internal/benchmarking/boomer/sweperf/sweperf.go
```diff
@@ -0,0 +1,699 @@
...
+// caller gets a session that is ready to take cycles. Resolves the config per
+// session, which lets a value changed mid-run apply to later sessions. On
+// failure it tears the actor down.
+func (r *sweperfRuntime) startUser(ctx context.Context) (*sweperfUser, error) {
+	tmpl, totalSteps, numCycles := r.resolveConfig()
+	chunks := generateDynamicChunks(totalSteps, numCycles)
+
+	u := &sweperfUser{
+		cfg:          r.cfg,
+		actorName:    "sb-" + uuid.NewString(),
```
maxsmythe, line 216: unrelated commentary: funny to me that this actorname prefix lives on. It was originally taken from an implementation of counter demo, IIRC. In any case, good to have an identifiable prefix to make logs more parseable.

internal/benchmarking/boomer/sweperf/sweperf.go
```diff
@@ -0,0 +1,699 @@
...
+
+// ref is the control-plane reference to this user's actor.
+func (u *sweperfUser) ref() *ateapipb.ObjectRef {
+	return &ateapipb.ObjectRef{Atespace: u.cfg.Atespace, Name: u.actorName}
+}
+
+// setActorRouting names the actor the atenet router should route this request
+// to. The router selects the actor from this header alone, so the request's
+// own Host stays the router's; a request without it is refused with a 404.
+func (u *sweperfUser) setActorRouting(req *http.Request) {
```
maxsmythe, line 285: It strikes me that routing/request sending is also something that could be library-fied. Seems like too much work to throw into an unrelated PR though.

internal/benchmarking/boomer/sweperf/sweperf.go
```diff
@@ -0,0 +1,699 @@
...
+	CompletedStep int    `json:"completed_step"`
+	Error         string `json:"error"`
+}
+
+// pollJobCompletion waits for an asynchronous /execute job to reach a terminal
+// state, for up to two minutes. It returns an error when the job fails, when
+// it completes with a non-zero exit code, or when that budget runs out.
+func (u *sweperfUser) pollJobCompletion(ctx context.Context, jobID string, cycleNum int) error {
+	const maxRetries = 600
+	const retryInterval = 200 * time.Millisecond
```
maxsmythe, line 546: One benchmarking note -- this means we can only benchmark in 0.2 second increments, which may be fine depending on general duration of executions.

maxsmythe (commented):
Thank you for PR!

Biggest questions are about the repo we are drawing binaries from and how VU config values are sourced.

benchmarking/locust/tests/sweperf.py
pmandewalkar: true! will slightly reword it and point to the sweperf path

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: i think my idea was to have as many failsafes as possible? should i just stick to dynconfig

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: LOL i was a bit confused by the naming convention for this prefix, should i leave this untouched?

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: noted - i'm gonna move the dynwait to shared utils but from my understanding of your comment i shouldn't touch this yet right?

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: we will receive server-side CEL in the HTTP response and extending that and surfacing that metric will be our next step. after discussions with nishanth and haowei it seems that this polling rate is fine for our purpose, but we can discuss further if necessary!

internal/benchmarking/boomer/sweperf/sweperf.go
maxsmythe: We should try to be uniform here. I suspect dynconfig will need to be redone at some point, a uniform pattern will make that easier.

internal/benchmarking/boomer/sweperf/sweperf.go
maxsmythe: Whatever sparks joy :)

internal/benchmarking/boomer/sweperf/sweperf.go
maxsmythe: I mean, you can. Just felt mean to block this PR on it.

internal/benchmarking/boomer/sweperf/sweperf.go
maxsmythe: As long as the time resolution works for you.

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: i wanted to check scope before I do this. there are three copies rn: taskRuntime.dynamicWait + liveWait (lifecycle.go), durDirRuntime.dynamicWait (durdir.go), and this one. lifecycle.go already has a package-private
uniformWait(lo, hi) that durdir doesn't use.

Options I see:
1. Move uniformWait to boomerutil.UniformWait(lo, hi). Each runtime would then have a 2-line dynamicWait() that picks MinWait/MaxWait (liveWait picks MinLive/MaxLive). this shares the math function but leaves 3 wrappers.
2. same, but delete the wrappers and call boomerutil.UniformWait(dyn.MinWait, dyn.MaxWait) at the 6 sleep sites. this has no wrappers, but the which window choice gets repeated at each call site.
3. we could accessors on dynconfig.Config [cfg.DrawWait() / cfg.DrawLive()] over one shared helper. this gets rid of the copies and keeps the field choice next to the fields, which helps if dynconfig gets redone like you were suggesting.

I think 3 makes sense, or 1 if you'd rather keep dynconfig free of actions. what do u think?

Also, these would touch glutton. i'm happy to do it here, or I can put it in a separate PR with the shared setActorRouting change so this one stays unblocked and sweperf-only, up to you!

benchmarking/workloads/manifests/swebench-astropy-7336-template.yaml.tmpl
pmandewalkar: Agreed. Waiting until the SWE-Perf image repo is public, then I'll repoint this to the official location.

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: fixed, pulled only from dynconfig now to match how the other users do it

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: leaving untouched!

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: discussed, will put shared function extraction in a diff pr

benchmarking/workloads/manifests/swebench-astropy-7336-template.yaml.tmpl
pmandewalkar: Public now, added the repo

internal/benchmarking/boomer/sweperf/sweperf.go
```diff
@@ -0,0 +1,679 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+// templateNS is the namespace holding the actor templates this workload
+// instantiates from.
+const (
+	templateNS = "benchmark-workloads"
```
roycaihw, line 55: Shall we randomize this namespace / make it configurable? Or is this (`benchmark-workloads`) a common pattern shared by other benchmarks?

internal/benchmarking/boomer/sweperf/sweperf.go
```diff
@@ -0,0 +1,679 @@
...
+// chunk partitions totalSteps instructions into numCycles slices.
+type chunk struct {
+	start int
+	end   int
+}
+
+// generateDynamicChunks splits totalSteps into numCycles contiguous chunks.
+// The remainder is spread over the earliest cycles, so chunk lengths differ by
+// at most one.
+func generateDynamicChunks(totalSteps int, numCycles int) []chunk {
```
roycaihw, line 87: nit: the chunking logic can be put in a separate util.go file and be unit-tested there

internal/benchmarking/boomer/sweperf/sweperf.go
```diff
@@ -0,0 +1,679 @@
...
+
+	slog.Info("Completed sweperf cycle",
+		slog.String("actor", u.actorName),
+		slog.Int("cycle", cycleNum),
+	)
+
+	u.cycleIndex++
+}
+
+// executeRequest is the /execute body: the 1-based, inclusive range of trace
```
roycaihw, line 511: nit: ideally the sweperf client library can also live in its own separate client.go file

roycaihw (approved):
LGTM. Thanks!

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: it's shared! mirror's glutton's convention.

internal/benchmarking/boomer/sweperf/sweperf.go
pmandewalkar: makes sense! i think Max also suggested a bit of library-fying, so ideally that will go in a separate PR and I can address your two comments there. I want to get the workload itself to Nishanth as soon as I can and will handle the refactoring after :D

maxsmythe: Why would we have the chunking logic in a util.go file? Are we expecting to reuse the chunks for other code? Seems pretty use-case specific to me.

maxsmythe (approved):
LGTM

PR #1702, opened 2026-09-17
ateapi: [Part 1] Runtime Authorization check for Atespace CRUD

SHRUTI6991:

Implement Runtime OpenFGA Authorization for Atespace RPCs


SHRUTI6991:
Category | Scenario | Expected Behavior
-- | -- | --
Startup & Bootstrap | Initial admin bootstrapping (--authz-bootstrap-owners) | Specified principals (including Kubernetes ServiceAccounts with special characters like :) are idempotently granted owner on global:root on startup.
Role Hierarchy & Inheritance | Global Admin vs. Space-Scoped User vs. Unprivileged User | • Global Admins inherit full access (get, list, create, delete) across all atespaces.• Space-Scoped Users can only access/list the specific atespaces they are granted roles on.• Unprivileged Users are denied all operations (PermissionDenied).
Existence Probing Protection | Accessing or deleting a nonexistent atespace | • Unprivileged Users receive PermissionDenied (cannot probe whether an atespace exists).• Global Admins pass authz fallback and receive NotFound from the database.
Scoped Pagination | Listing atespaces with PageSize when earlier DB pages contain unauthorized spaces | Server loops through database pages until PageSize authorized items are collected (or EOF), ensuring authorized spaces on later pages are not dropped.
Create Failure & Rollback | Database creation succeeds, but OpenFGA tuple write fails | The newly created database row is automatically rolled back (deleted) and an Internal error is returned so no orphaned/unlinked atespace remains.
Idempotent Duplicate Creates | Creating an atespace that already exists (AlreadyExists) | Self-heals the parent_global link without deleting or resetting any existing user role bindings on the live atespace.
Delete & Orphan Cleanup | Deleting an atespace (and handling partial delete failures) | • Deletes both the database row and all OpenFGA tuples (parent_global + user roles).• If tuple deletion fails mid-flight, retrying DeleteAtespace sweeps away the leftover tuples even when the DB row is already gone (NotFound).
Delete & Recreate Safety | Deleting an atespace and creating a new one with the same name | Purges any leftover tuples before creating the new space, preventing former members of the old atespace from inheriting access to the new one.
Internal Reconciler Bypass | Background control-plane loops (e.g., ActorTemplate golden snapshot reconciler) | Uses WithBypass context so internal system operations succeed without requiring an external user JWT.

internal/authz/model.fga
SHRUTI6991, line 65: We dont have update RPC for atespace, I will clean this up.

SHRUTI6991:
Not ready for review yet.

internal/authz/interceptor.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Checker evaluates runtime OpenFGA permissions for gRPC requests.
+type Checker interface {
+	CheckPermission(ctx context.Context, relation, object string) error
+}
+
+// UnaryServerInterceptor returns a gRPC unary interceptor that enforces per-RPC
+// permissions registered in defaultRPCPermissions using checker.
+func UnaryServerInterceptor(checker Checker) grpc.UnaryServerInterceptor {
```
SHRUTI6991, line 56: Not sure if there is much value of having this interceptor now. 

Initially when I started I was thinking of putting all the authz logic in grpc interceptor before it even reached handler. But now that we have to resolve uid, I wonder if handler is a better place for authz logic. It prevents the UID context pass as well.

cmd/ateapi/internal/store/atepg/atepg.go
```diff
@@ -455,8 +452,23 @@ func (p *Persistence) ListAtespaces(ctx context.Context, opts store.ListOptions)
...
+	var existingUID string
 	var protoBytes []byte
-	err := p.pool.QueryRow(ctx, `DELETE FROM atespaces WHERE name = $1 RETURNING proto`, name).Scan(&protoBytes)
+	err := p.pool.QueryRow(ctx, `
+		WITH existing AS (
+			SELECT uid FROM atespaces WHERE name = $1 FOR UPDATE
+		),
+		deleted AS (
+			DELETE FROM atespaces
+			WHERE name = $1 AND ($2 = '' OR uid = $2)
```
thockin, line 464: Aren't we supposed to have UID and version preconditions on every delete?  @juli4n

cmd/ateapi/main.go
```diff
@@ -75,6 +76,8 @@ var (
 	authenticationConfigFile = pflag.String("authentication-config", "", "YAML file configuring trusted JWT providers.")
 	postgresConnectionString = pflag.String("postgres-connection-string", "", "PostgreSQL connection string (libpq DSN or URI).")
 	postgresSchema           = pflag.String("postgres-schema", "public", "PostgreSQL schema for Substrate tables. This overrides a search_path connection parameter.")
+	enableAuthz              = pflag.Bool("enable-authz", false, "Enable OpenFGA authorization checks.")
```
thockin, line 79: Do we really need this to be optional?  If so, can we default to true?

SHRUTI6991: I wanted some feature flag mechanism since it's going to touch multiple RPCs until its ready (specifically read path checks like this PR + governance API to insert tuples). Once the governance API + impl goes in we can enable this to true. 

So just used boolean, not sure if there is substrate level feature flag mechanism.

cmd/ateapi/internal/store/atepg/atepg.go
SHRUTI6991: This logic is now completely changed.

Now that we can store resource name in OpenFGA tuples table, we don't this complex logic. DeleteAtespace just deletes atespace resource and tuples in 1 TX after this change: https://github.com/agent-substrate/substrate/pull/1839/changes#top

PR #1704, opened 2026-09-17
microvm: drop the kata-config asset

### LINKED ISSUE #1693: Drop the kata-config asset from the microvm SandboxConfig
Today, the microvm `SandboxConfig` includes an asset for the kata config:

```yaml
kata-config:
  url: "gs://${BUCKET_NAME}/kata-assets/configuration-clh.toml"
  sha256: "20f5de4f705423ddd1ee93318c784c6aede30c07c78dc6b311a771cc2f6859c7"
```

I don't think we need this config. In Kata Containers, it is read by the Go kata shim
(`containerd-shim-kata-v2`) and the `kata-runtime` to configure Cloud Hypervisor. In Substrate, `ateom-microvm` runs neither, only the `kata-agent`. IIUC, it reads this file to set defaults: `default_vcpus`, `default_memory` and `kernel_params`. Looking into history in Kata Containers, the first 2 never changed and `kernel_params` has changed once.

Also, from 4.1.0 and newer, kata-static no longer ships this file.

@BenTheElder Thoughts?

hdp617:
Today, `ateom` fetches the `kata-config` asset (configuration-clh.toml) to read only 3 values `default_memory`, `default_vcpus` and `kernel_params`. The first 2 are redundant because (1) the values never change and (2) ateom has the same defaults. Only `kernel_params` is relevant for the `kata-agent`; its value also changed once in the Kata project history.

`ateom` now owns all three values, which also allows us to tune them specifically for Substrate.

Verified e2e with a GKE cluster.

Fixes #1693

- [x] Tests pass


hdp617:
cc @BenTheElder

BenTheElder:
This appears to be failing tests ..

hdp617:
It's a flake; I don't have permission to rerun.

cmd/ateom-microvm/internal/kata/config.go
```diff
@@ -15,71 +15,29 @@
...
 // WithDebugConsole appends the kata-agent debug-console kernel parameters so the
 // guest agent binds a root debug shell on vsock port 1026, which DebugConsoleDump
 // connects to for in-guest diagnostics. Both params are required: agent.debug_console
 // enables the console and agent.debug_console_vport=1026 makes the agent bind it on
 // the vsock port (the agent only binds a vsock listener when the vport is > 0).
 // Idempotent.
-func WithDebugConsole(kernelParams string) string {
-	return appendKernelParams(kernelParams, "agent.debug_console_vport",
+func WithDebugConsole() string {
+	return appendKernelParams(baseKernelParams, "agent.debug_console_vport",
```
BenTheElder: 🤖 **nit** 🟢 – With `baseKernelParams` now a constant, this is effectively a constant too, and the dedupe it asks for is dead code: `appendKernelParams`'s marker check can never match, because a compile-time constant cannot contain `agent.debug_console_vport`. The "Idempotent." doc line above is vacuous for the same reason — there is no input to re-apply.

That also makes the second assertion in `config_test.go:28` a tautology: a nullary deterministic function trivially returns the same string twice, so the debug-console dedupe path is untested. `WithAgentDebug` is the real case, and its test at line 39 correctly feeds `got` back in.

Making this a `const` would say what it is and drop both the dead branch and the assertion that cannot fail.

hack/install-microvm-deps.sh
```diff
@@ -142,7 +142,7 @@ OUT="${OUT:-${ROOT}/bin/microvm-assets/$ARCH}"
 
 # --- 1. assets: assemble (if missing or stale) -----------------------------
 need_assemble=false
-for f in cloud-hypervisor virtiofsd vmlinux rootfs.img configuration-clh.toml; do
+for f in cloud-hypervisor virtiofsd vmlinux rootfs.img; do
```
BenTheElder, line 145: 🤖 **nit** 🟢 – The five-to-four sweep missed three comments that still describe the old asset set:

- line 151 here, "The five filenames don't change when a version pin moves"
- line 188, which still lists `config` among the downloaded assets with committed shas
- `hack/microvm-assets/assemble.sh:67`, also "the five filenames stay the same"

The stamp logic they describe is unchanged and still correct; only the counts and the asset list are stale.

cmd/ateom-microvm/internal/kata/config.go
hdp617: I made it a constant and removed the test.

hack/install-microvm-deps.sh
hdp617: Updated.

hdp617:
Looks like a flake.

BenTheElder (approved):

PR #1706, opened 2026-09-17
Reclaim actor node state when a worker pod goes away

### LINKED ISSUE #1688: atelet: actor state directories are never reclaimed when a worker pod goes away
### What happened?

Nothing removes `/var/lib/ateom-gvisor/actors/<actor-UID>/` when the worker pod hosting the actor goes away. The directory holds the actor's durable-dir contents, checkpoint/restore state and bundle overlay — for a durable-dir workload that is gigabytes per actor, not the few idle bytes of #1677's socket dirs. The directories persist for the life of the node.

This surfaced as node disk climbing during a durable-dir benchmark and never falling afterwards, eventually reaching ~90% and getting atelet evicted.

Measured on a 3-node dev cluster (`c3-standard-4`, 100 GB pd-balanced) against `a0d306ee`. Workload: 15 concurrent clients, each driving an actor that writes a 1 GiB file into its durable dir and then suspends and resumes repeatedly, for 10 minutes.

| Metric | Measurement / Impact |
|---|---|
| orphaned actor dirs | **23** (19 with payload, 4 empty template actors) |
| retained under `actors/` | **27.10 GB** |
| per actor | **~1.4 GB** |
| reclaimed by scaling the WorkerPool to 0 | **0 bytes** |
| reclaimed by `--delete-all` (full uninstall) | **0 bytes** |
| reclaimed by a subsequent reinstall | **0 bytes** — fresh atelets start on top of it |

Reproduced on both sandbox classes; they share `ateompath.BasePath`, so the layout and the leak are identical. An equivalent micro-VM run left 18.33 GB across 16 dirs, likewise unchanged by scaling to 0.

> [!NOTE]
> #1654 ("Remove an actor's directory when atelet terminates it") does **not** cover this. All figures above are from a build containing it. It removes the directory *when `Terminate` runs*; the defect is that `Terminate` is never reached once the pod is gone.

No delete is required to trigger this — an actor being de-scheduled from a node is enough.

### Mechanism

As far as I can tell, two independent gaps compose:

1. **The worker-delete path never contacts the node.** `WorkerPoolSyncer.reconcile` notices the pod is gone and calls Control API `DeleteWorker`. `WorkerWorkflow.ensureBoundActorsReleased` → `releaseBoundActor` moves each bound actor to `CRASHED` and clears its pod pointers, but nothing asks atelet to terminate it. After the worker record is deleted, the actor names no worker, no worker names a node, and nothing ever revisits the actor UID — the directory is orphaned permanently.

2. **The actor-delete path gives up when the pod is gone.** `ActorWorkflow.ensureAteletTerminated` dials `DialForWorker(ns, pod)` and returns nil on `ErrWorkerPodNotFound`. That is the correct shortcut for a pod that legitimately vanished, but it is also exactly how the state comes to be orphaned.

Even when a `Terminate` is issued, it fails before reaching the directory cleanup: the teardown dials the ateom socket first, and a vanished ateom (`connect: no such file or directory`, `connection refused`) aborts the call ahead of the unmounts and the directory removal.

### Expected Behavior

Actor node state is reclaimed once no actor can use it again. Two complementary mechanisms, mirroring the split proposed in #1677:

1. **Prompt reclaim on the delete path.** The worker-delete workflow is the last moment anything still knows which node holds the state, so it should tell that node to reclaim before releasing the record. Reaching the atelet requires dialing **by node** rather than by pod, since the pod is precisely what has disappeared. This must be best-effort: an unreachable node cannot be allowed to wedge worker deregistration.

   `Terminate` also needs to tolerate a vanished ateom — skip the sandbox teardown and still run the unmounts and directory cleanup — otherwise the call fails before doing the part that matters. A *reachable* ateom that rejects the call should still fail: that sandbox is live.

2. **A sweep as backstop.** Evictions, node crashes and uninstalls produce orphans that no delete path will ever visit. A periodic atelet sweep reconciling `actors/` against the actors the control plane places on that node covers them. It must fail closed — if the live set cannot be read in full, delete nothing — and must skip actors holding a **local pause snapshot**, which is node-pinned state whose deletion is unrecoverable.

Neither covers the graceful path. Reclaiming during `Checkpoint` as well would keep steady state near zero and shrink the sweep's caseload; that is a separate change.

### Steps to Reproduce

1. Install ate-system and deploy a WorkerPool with 15 workers.
2. Record the node state:
   ```bash
   du -sb /var/lib/ateom-gvisor/* && ls -1 /var/lib/ateom-gvisor/actors | wc -l
   ```
3. Drive 15 concurrent actors for ~10 minutes, each writing a 1 GiB file into its durable dir and suspending/resuming between iterations.
4. Record the node state again — expect several GB per node under `actors/`.
5. Scale the WorkerPool to 0, wait for the pods to drain, and record again.

Every actor directory and every byte is still present. The same holds after `hack/install-ate.sh --delete-all` and after a fresh reinstall.

### Sandbox Runtime

Both / Runtime Agnostic — gVisor and micro-VM share `ateompath.BasePath`.

`main` @ `a0d306ee`

### Additional Context

**Not fixed by #1678.** That PR adds `defer os.RemoveAll(ateomDir)` to the two ateom runtime mains, which covers `ateoms/<pod-UID>/` only. `actors/<uid>/` is written and owned by atelet, not by ateom, and is not touched by that defer — nor could it be, since the ungraceful exits that strand actor state are exactly the cases where a defer does not run. Same pattern, different directory, very different blast radius:

|            | #1677 / #1678                                      | this issue                                           |
|------------|----------------------------------------------------|------------------------------------------------------|
| path       | `ateoms/<worker-pod-UID>/`                         | `actors/<actor-UID>/`                                |
| written by | ateom                                              | atelet                                               |
| contents   | one dead `ateom.sock`                              | durable dir, checkpoints, bundle overlay             |
| size       | a few idle bytes                                   | ~1.4 GB per actor                                    |
| harm       | dial+probe per stats sweep, `cachedPools` retention | node disk pressure, atelet eviction                  |
| trigger    | any pool rollout                                   | worker pod deleted or de-scheduled with actors bound |

#1677's proposed janitor (`cmd/atelet/ateomgc.go`) is close in shape to the sweep described above — both modeled on `imagegc.go`, both node-scoped, fail-closed, with a min-age. If both land it is worth deciding whether they should be one loop.

**Documentation gap.** There is no written description of atelet's on-disk layout or how it does bookkeeping, which is part of why the ownership and lifetime of these directories is unclear.

### Confirmation

sairajp-rewind:
Fixes #1688 

Nothing removes `actors/<actor-UID>/` once the worker pod hosting the actor is gone. The directory holds the actor's durable-dir contents, checkpoint/restore state and bundle overlay, so it is as large as the workload made it and it persists for the life of the node. This adds reclaim on the delete path, a sweep for the cases no delete path visits, and a wipe on uninstall.

Three commits, reviewable in order:

| commit                                                                | what                            |
|-----------------------------------------------------------------------|---------------------------------|
| `ateapi,atelet: reclaim actor node state when a worker's pod is gone` | the delete path                 |
| `atelet: sweep orphaned actor state directories`                      | the backstop                    |
| `hack: wipe the node state hostPath on uninstall`                     | uninstall leaves nothing behind |

### 1. Reclaim on the worker-delete path

`WorkerWorkflow.releaseBoundActor` now calls `Terminate` before releasing each bound actor — the last moment anything still knows which node holds the state. It dials **by node** (`DialForAteletOnNode`), since the pod it would otherwise dial through is what just disappeared, and is best-effort: an unreachable node must not wedge worker deregistration.

`Terminate` needed fixing too. It tore the workload down first and returned early on failure, never reaching the unmounts or the directory removal. A **gone** ateom (missing socket, no sandbox record, `UNAVAILABLE`) now skips that step and still reclaims. A **reachable** ateom that rejects the call still fails — that sandbox is live.

### 2. Sweep as a backstop

Evictions, node crashes and uninstalls leave directories no delete path will ever visit. `cmd/atelet/actorgc.go` runs at startup and on a period, like the existing `imagegc.go`, and compares `actors/` against the actors the control plane says belong on this node.

It does not delete:

- anything, if that list cannot be read in full — it fails closed;
- actors this atelet is currently running;
- directories younger than `--actor-gc-min-age`;
- directories with a **local pause snapshot**. A PAUSED actor has one, and has no worker assignment to show up in the list by.

It renames a directory before deleting it, so a crash mid-delete leaves debris the next pass cleans up instead of a half-empty directory that still looks live.

Flags, all in `cmd/atelet/actorgc.go`: `--actor-gc-period` (default 5m; 0 disables the periodic pass, startup still runs), `--actor-gc-min-age` (default 10m), `--actor-gc-dry-run` (default false).

### 3. Wipe on uninstall

`--delete-all` tore down the control plane but never touched the hostPath it wrote to, so the data outlived the software — and fresh atelets would start on top of it. `delete_ate_system` now runs a short-lived DaemonSet that empties the state directory on every node, sequenced after the atelet DaemonSet is gone so nothing is writing to it. Readiness is the wipe having finished, so a completed rollout means every node is clean. `--keep-node-state` opts out.

### Verification

3-node cluster, durable-dir workload: 15 concurrent clients each writing a 1 GiB file into its durable dir and suspending/resuming for 10 minutes.

|                                                     | before                      | after                                        |
|-----------------------------------------------------|-----------------------------|----------------------------------------------|
| **delete path** — scale the WorkerPool to 0         | `+0.00 GB`, 12 dirs remain  | **26.23 GB → 0**, 0 dirs                     |
| **sweep** — start atelets over pre-existing orphans | untouched                   | **23 dirs / 29.1 GB reclaimed**, 0 failures  |
| **uninstall** — `--delete-all`                      | `+0.00 GB`, 16 dirs remain  | **26.24 GB → 0** on every node               |

On the delete path, 12 reclaim calls produced 12 matching `Terminate`s that took the gone-ateom branch — a per-node match against the directories present (5/5/2). The sweep's first pass ran before ateapi was serving and correctly reclaimed nothing (`reading the live actor set failed; skipping this pass`).

Reproduced and fixed on both sandbox classes; they share `ateompath.BasePath`, so the layout and the leak are identical.

### Relation to #1677 / #1678

Same pattern, different directory. #1678 removes `ateoms/<pod-UID>/` on graceful ateom shutdown; this is `actors/<actor-UID>/`, written by atelet, and the exits that strand it are exactly the ungraceful ones a defer does not cover.

The sweep here and the janitor proposed in #1677 landed on the same shape — both modeled on `imagegc.go`, node-scoped, fail-closed, min-age. If both land, worth deciding whether they should be one loop.

### Not covered

The graceful path. Reclaiming during `Checkpoint` would keep steady state near zero and shrink the sweep's caseload; that is a separate change.

`Terminate` idempotency generally. Only one case improves here: `Terminate` with no live sandbox now succeeds instead of erroring.

- [x] Tests pass


sairajp-rewind:
cc: @shrutiyam-glitch @juli4n

igooch:
@sairajp-rewind this is the remaining half of #641: #788 and #1654 covered the live-worker path, and what was left was reaching the node when no worker can be dialed. Would you be up for taking #641 into this PR's scope?

The one piece it would add: the sweep skips directories with a local pause snapshot on the premise that "a later delete still reclaims" them, but today it doesn't. `ensureAteletTerminated` returns early with no worker assignment (`workflow_delete.go:128`) and pause clears the assignment (`workflow_pause.go:289`), so a PAUSED actor deleted with any-state is never reclaimed by either half. The delete workflow does know the node via `LocalSnapshotInfo.NodeVmsWithLocalSnapshots`, and with this PR's dial-by-node and gone-ateom tolerance it could Terminate there. Same shape for the revert TODO at `workflow_revert.go:91`.

If you'd rather keep this PR as is, that's fine too. I'd just ask that the skip's comment say it is permanent for now, and I'll do the delete-path piece as a follow-up.

PR #1708, opened 2026-09-17
microvm: bump kata assets to 4.1.0

### LINKED ISSUE #1695: Bump Kata assets to 4.1.0
[Kata 4.1.0](https://github.com/kata-containers/kata-containers/releases/tag/4.1.0) ships with virtiofsd 1.14.0. It removes the need to pin virtiofsd and to build it from source in arm64.

https://github.com/agent-substrate/substrate/blob/85ce8ed5313f64aa4d8d5fb616ce450abf1e95e0/hack/microvm-assets/assemble.sh#L129-L142

hdp617:
Stacked on #1704, which removes kata-config.

Kata 4.1.0 bundles virtiofsd 1.14.0 (required by Substrate). This simplifies the dev process on arm64 because it removes the need to build virtiofsd from source.

Verified on an arm64 KVM host (Lima + kind).

Fixes #1695

- [x] Tests pass


hdp617:
cc @BenTheElder

cmd/ateom-microvm/internal/kata/config.go
```diff
@@ -15,62 +15,19 @@
...
-	Hypervisor struct {
-		CLH struct {
-			DefaultMemory int    `toml:"default_memory"`
-			DefaultVCPUs  int    `toml:"default_vcpus"`
-			KernelParams  string `toml:"kernel_params"`
-		} `toml:"clh"`
-	} `toml:"hypervisor"`
-}
+const (
+	// DefaultMemoryMiB is the default guest memory size (MiB).
```
BenTheElder, line 23: Can we add a TODO to revisit this? 2GB is probably not an appropriate default for substrate, we just inherited it from kata and never changed it (versus setting defaults on all our examples).

I think we should perhaps set it rather low by default and require explicit sizing up.

We shouldn't do that in this PR, but a clear TODO that suggests we're not planning to keep this as-is would be good.

cmd/ateom-microvm/internal/kata/config.go
```diff
@@ -15,62 +15,19 @@
...
+	DefaultMemoryMiB = 2048
+	// DefaultVCPUs is the default guest vCPU count.
+	DefaultVCPUs = 1
```
BenTheElder, line 26: Same here

cmd/ateom-microvm/internal/kata/config.go
```diff
@@ -15,62 +15,19 @@
...
-		cfg.MemoryMiB = memDefault
-	}
-	if cfg.VCPUs <= 0 {
-		cfg.VCPUs = vcpuDefault
-	}
-	return cfg, nil
-}
+// BaseKernelParams is the guest kernel command line parameters ateom boots with;
+// there is no kata shim to inject them.
+const BaseKernelParams = "cgroup_no_v1=all systemd.unified_cgroup_hierarchy=1"
```
BenTheElder: Does this need to be exported?

cmd/ateom-microvm/internal/kata/config_test.go
```diff
@@ -19,57 +19,27 @@ import (
...
-	}
-	if want := "agent.foo=bar systemd.unit=kata-containers.target"; cfg.KernelParams != want {
-		t.Errorf("KernelParams = %q, want %q", cfg.KernelParams, want)
+// TestBaseKernelParams pins the guest command line to the bytes kata 4.0.0 built into
+// configuration-clh.toml, the asset ateom used to fetch and parse. Dropping that asset
+// must not drop these params, so assert them literally: a guest that keeps a cgroup v1
+// hierarchy available is a silent behavior change, not a loud failure.
+func TestBaseKernelParams(t *testing.T) {
+	want := "cgroup_no_v1=all systemd.unified_cgroup_hierarchy=1"
+	if BaseKernelParams != want {
```
BenTheElder: This seems like a bizzare test.

cmd/ateom-microvm/internal/kata/config_test.go
```diff
@@ -19,57 +19,27 @@ import (
...
-				t.Fatalf("ParseConfig: %v", err)
-			}
-			if cfg.MemoryMiB != 2048 {
-				t.Errorf("MemoryMiB = %d, want default 2048", cfg.MemoryMiB)
-			}
-			if cfg.VCPUs != 1 {
-				t.Errorf("VCPUs = %d, want default 1", cfg.VCPUs)
-			}
-		})
+// TestGuestDefaultsMatchShippedKataConfig pins the sizing defaults to the values kata
```
BenTheElder: Same here. what's the value of this test?

cmd/ateom-microvm/internal/kata/config.go
hdp617: SGTM. Added a TODO for #1724.

cmd/ateom-microvm/internal/kata/config.go
hdp617: It was used once outside of this package; I've made it private.

cmd/ateom-microvm/internal/kata/config_test.go
hdp617: Yes, this is kinda silly. I missed it. Removed.

cmd/ateom-microvm/internal/kata/config_test.go
hdp617: Removed.

docs/dev/microvm-local.md
```diff
@@ -137,26 +137,25 @@ worker, and `vzNAT` gives the VM outbound networking under the vz VM type.
 
 ### 3. Assemble the arm64 assets inside the Lima VM
 
-`hack/microvm-assets/assemble.sh` must run on a Linux host of the target
-architecture (on arm64 it builds `virtiofsd` from source with cargo against
-Linux-only libraries), so run it in the Lima guest, not on macOS:
+Step 4 runs on macOS, where `assemble.sh` cannot finish: it ends by
+checksumming the assets with `sha256sum`, which macOS does not ship. Assemble
```
BenTheElder: 🤖 **nit** 🟢 – The instruction is right but the reason isn't. macOS does ship `sha256sum` (`/sbin/sha256sum`, verified on 15.7.9), so the script would not fail there. The real host-side blocker comes earlier, at `assemble.sh:98`: `tar --zstd -xf` makes bsdtar shell out to an external `zstd` binary, which stock macOS does not have (`tar: Can't launch external program: zstd`).

Suggest something like "…cannot finish: unpacking the kata tarball needs a `zstd` binary that macOS does not ship."

hack/microvm-assets/README.md
```diff
@@ -5,13 +5,12 @@ toolchain at runtime — nothing kata-specific is baked into the worker image. a
 the kata-agent directly (no kata shim, no containerd). Each actor container's rootfs is an
 overlay of a read-only lower (the OCI image, served into the guest over virtio-fs by
 `virtiofsd`) and a writable upper on a guest tmpfs, so `virtiofsd` is part of the asset
-set. The asset set is five files:
+set. The asset set is four files:
 
 - `cloud-hypervisor` — the VMM binary (fetched from its release)
-- `virtiofsd` — the virtio-fs daemon serving the RO lower (built from source; see `assemble.sh`)
+- `virtiofsd` — the virtio-fs daemon serving the RO lower (from kata-static)
```
BenTheElder, line 11: 🤖 **should-fix** 🟡 – Step 1 further down (lines 31-32) still says to paste the printed sha256 sums into `demos/counter/counter-microvm.yaml.tmpl`. That file has no `spec.assets` and no `sha256` at head, and this commit points `assemble.sh`'s own closing message at `manifests/microvm/sandboxconfig-microvm.yaml.tmpl` instead.

So the README and the script now disagree. Following the README on the next bump leaves the pins stale, which surfaces only as an actor stuck in `STATUS_RESUMING`. Updating step 1 to name the manifest would fix it.

docs/dev/microvm-local.md
BenTheElder: 🤖 **nit** 🟢 – The instruction is right but the reason isn't. macOS does ship `sha256sum` (`/sbin/sha256sum`, verified on 15.7.9), so the script would not fail there. The real host-side blocker comes earlier, at `assemble.sh:98`: `tar --zstd -xf` makes bsdtar shell out to an external `zstd` binary, which stock macOS does not have (`tar: Can't launch external program: zstd`).

Suggest something like "…cannot finish: unpacking the kata tarball needs a `zstd` binary that macOS does not ship."

BenTheElder: (Also ideally we just make this macOS compatible, we can use docker run with a digest-pinned container to run tools that are not on the host if we need)

hack/microvm-assets/README.md
hdp617: I've removed that part in the README. It's only relevant when bumping asset versions, which is not frequent currently. The `assemble.sh` script should update the manifest automatically in this case. It also avoids script and doc divergence like this. Will do in a follow-up.

docs/dev/microvm-local.md
hdp617: Right, will do in a follow-up!

manifests/microvm/sandboxconfig-microvm.yaml.tmpl
```diff
@@ -20,15 +20,15 @@
...
-# assets are 4.0.0.
+# The sandbox binaries (cloud-hypervisor, virtiofsd, guest kernel, guest rootfs)
+# are FETCHED at runtime from the cluster object store bucket ${BUCKET_NAME}
+# under kata-assets/ (rustfs on kind, GCS on GKE) - NOT baked into the worker
+# image. ateom boots cloud-hypervisor itself and gives the actor an overlay
+# rootfs (virtio-fs RO lower + guest-tmpfs upper), so it fetches virtiofsd too.
+# The kata containerd shim is NOT fetched (ateom drives the kata-agent
+# directly). kata assets are 4.1.0, which bundles virtiofsd v1.14.0 - the first
+# release with the vhost-0.16 snapshot/restore fix; the v1.13.3 kata bundled
+# through 4.0.0 hangs restore.
```
BenTheElder: we can trim this comment but that should come later, this is mostly just correcting it.

BenTheElder (dismissed):

manifests/microvm/sandboxconfig-microvm.yaml.tmpl
hdp617: Needed to rebase so I removed the comment here.

BenTheElder (approved):

PR #1711, opened 2026-09-17
docs: Document RevertActor and drop "terminal" from CRASHED

shrutiyam-glitch:
`RevertActor` returns a RUNNING, PAUSED, or CRASHED actor to SUSPENDED at its last external snapshot, so CRASHED is no longer a dead end that only `DeleteActor` can clear. Docs and code comments still described it as terminal and told operators to delete and recreate the actor, losing its state.

Update the api-guide, architecture, upgrade guide, and kubectl-ate README to cover the new verb, and correct the comments that justified keeping a partial external snapshot by naming actor deletion as the only remaining collector -- revert collects it too.

Follow up for the PR - #1675 
Issue - #1556 

- [x] Tests pass


cmd/ateapi/internal/controlapi/workflow_delete_test.go
```diff
@@ -447,7 +448,7 @@ func TestDeleteActor_CollectsSnapshotsAfterWorkerDelete(t *testing.T) {
 				t.Fatalf("DeleteWorker: %v", err)
 			}
 
-			// The actor is CRASHED and can only be deleted from here.
+			// The actor is CRASHED; delete reclaims its partial snapshot.
```
zoez7: > delete reclaims its partial snapshot.

what does this mean?

cmd/ateapi/internal/controlapi/workflow_delete_test.go
```diff
@@ -364,8 +364,9 @@ func TestDeleteActor_CollectsInFlightSnapshotWithoutTemplate(t *testing.T) {
 
 // TestDeleteActor_CollectsSnapshotsAfterWorkerDelete verifies that
 // deleting an actor whose suspend a worker delete crashed mid-finalize reclaims
-// every object that suspend wrote. CRASHED is terminal, so the actor delete is
-// the only collector left: whatever it cannot name is leaked for good.
+// every object that suspend wrote. When an actor crashes mid-suspend, only
+// delete or revert can collect the partial snapshot: whatever they cannot name
```
zoez7: > only delete or revert can collect the partial snapshot
we are not collecting partial snapshot anywhere right?

shrutiyam-glitch: by collecting, I meant only those two actions can get rid of the in progress snapshots. 
will reword it.

zoez7 (approved):

PR #1713, opened 2026-09-17
feat(ateapi): track actor_template_uid per snapshot on ExternalSnapshot

shrutiyam-glitch:
Follow up on the comment - https://github.com/agent-substrate/substrate/pull/1675#discussion_r4031619230 

Stamps `actor_template_uid` onto `ExternalSnapshot` so template provenance travels with the snapshot itself, rather than being inferred from the single `ActorStatus.current_actor_template_uid`.

#### Motivation

`current_actor_template_uid` records the template the **last sprint booted with** (`finalizeRunning` stamps it on every resume). It does *not* record the template the **snapshot** was captured under. Those two diverge as soon as an actor runs a sprint that does not produce a new durable snapshot, and `loadActorForResume` was using the former to decide whether to force a `DATA`-only restore instead of `FULL`.

Walking an actor through a repoint and a crash:

| Step | Action | `current_…_uid` | `ExternalSnapshot` | 
|---|---|---|---|
| **a** | Suspend on template **A** | A | captured on **A** | 
| **b** | `UpdateActor` repoints the spec to **B** | A | captured on **A** | 
| **c** | Resume → `A != B`, restores `DATA`, sprint boots on **B** | **B** | still on **A** | 
| **d** | Actor crashes (or is reverted) — no new snapshot taken | B | still on **A** | 
| **e** | Resume → `current == target == B`, so **no repoint is detected** | B | still on **A** | 

At step **e** the old check reports "template not replaced" and restores the external snapshot in `FULL` — replaying a memory image captured under **A**'s sandbox on **B**. That is exactly the case the guard exists to prevent; it was silently defeated by the sprint at step **c** advancing `current_actor_template_uid` past the snapshot.

*(Note: Paused actors do not face this divergence. Because `UpdateActor` is only allowed in the `SUSPENDED` state, a paused actor cannot have its template updated. Therefore, local checkpoints do not need separate template provenance tracking).*

#### Changes

- **Data Model Updates:** Added the `actor_template_uid` field to `ExternalSnapshot`.
- **Snapshot Stamping:** Updated the capture logic to stamp external snapshots with the active sandbox's `ActorTemplate` UID when suspending an actor or creating an actor from a tag.
- **Resume Evaluation Logic:** Modified the external resume logic to compare the target template UID against the specific snapshot's UID (`ExternalSnapshot.actor_template_uid`) rather than the actor's overall status. Local restores bypass this check entirely since templates cannot change during a pause.

**Tests**
 - `TestResumeActor_AteletWireRequest` (unit): snapshot fixtures now carry `actor_template_uid`, since provenance is read from the snapshot rather than from actor status.
 - Functional expectations for create/update/resume/pause updated for the new field on the external snapshot golden files.
 - `TestUpdateTemplateLifecycle` (e2e): asserts `external_snapshot.actor_template_uid` after suspend and re-suspend.

A `TODO` marks the `resume → crash/revert → resume` e2e sequence to be added once `RevertActor` lands — that is the path described in the table above.

- [ ] Tests pass


internal/e2e/suites/demo/updatetemplate_test.go
```diff
@@ -187,6 +190,57 @@ func runUpdateTemplateTestCase(t *testing.T, onCommit ateapipb.SnapshotContentSc
 		t.Errorf("[after template update] expected %q (template B validating the preserved file), got response: %s", want, resp)
 	}
 
+	//
+	// Pause under template B while the external snapshot is still from
```
zoez7: > 	// Pause under template B while the external snapshot is still from
>	// template A: LocalSnapshotInfo records template B while ExternalSnapshot
>	// retains template A.

Is that true? After pause, the ExternalSnapshot entry should be cleared from the Actor.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -48,9 +48,13 @@ type resumeSnapshotSource struct {
...
 	GoldenSnapshotURI resources.SnapshotURI
-	// TemplateReplaced is true when the snapshot's recorded template UID
-	// differs from the actor's current template.
-	TemplateReplaced bool
+	// LocalTemplateReplaced and ExternalTemplateReplaced are true when the
+	// actor's local pause checkpoint, respectively its external snapshot, was
+	// captured under a template other than the actor's current one. They are
+	// tracked separately because the two can have been captured under
+	// different templates, and each restore path reads only its own.
+	LocalTemplateReplaced    bool
```
zoez7: LocalTemplateReplaced should not be possible, updating an actor template is only allowed for suspended actors, this might be an existing bug.

shrutiyam-glitch: my bad, I went through the flow again, updating an actor's template is only possible in suspended state. 
whenever an actor resumes from a local pause checkpoint (LocalSnapshotInfo != nil), its template could **not** have been changed via UpdateActor since the pause checkpoint was taken. So LocalTemplateReplaced is not required, removed and update the title and pr desc.

internal/e2e/suites/demo/updatetemplate_test.go
shrutiyam-glitch: `ensurePausedFinalized` sets LocalSnapshotInfo and clears WorkerAssignment, leaving ExternalSnapshot intact.

cmd/ateapi/internal/controlapi/workflow_suspend.go
```diff
@@ -384,8 +384,9 @@ func (w *ActorWorkflow) ensureSuspendedFinalized(ctx context.Context, actorRef r
 	externalSnapshot := latestActor.GetStatus().GetExternalSnapshot()
 	if inProgressSnapshotURI != "" {
 		externalSnapshot = &ateapipb.ExternalSnapshot{
-			SnapshotUri:  inProgressSnapshotURI,
-			ContentScope: commitSnapshotScope(actorRef.Atespace, actorTemplate),
+			SnapshotUri:      inProgressSnapshotURI,
+			ContentScope:     commitSnapshotScope(actorRef.Atespace, actorTemplate),
+			ActorTemplateUid: latestActor.GetStatus().GetCurrentActorTemplateUid(),
```
zoez7: Can we delete the actor.status.current_actor_template_uid? I don't think we need it in addition to the ExternalSnapshot.actor_template_uid

shrutiyam-glitch: done in [5e11f06](https://github.com/agent-substrate/substrate/pull/1713/commits/5e11f06b1d9c0e5e798ff22503d5e575ac01d2ba)

zoez7 (approved):

Issue #1686, opened 2026-09-16
System Upgrade GA


HavenXia:
The epic for GA's plan and implementation of Substrate Upgrade. 

The GA plan is at https://docs.google.com/document/d/1hw0fGm6ZqtjylX-ZzxJFbBhfpsVwLA_JFQBSVBWFpwg/edit?usp=sharing.

HavenXia:
# Substrate Upgrade GA

High-level outline for the substrate upgrade within a release window.

## Decision

By GA atelet and ateom are decoupled, so a dataplane upgrade is an edit of the WorkerPool's `workerImage`.

The pool's Deployment rolls the edit: it deletes old worker pods in batches (25% by default), and each worker pod goes through the eviction path. The worker gets SIGTERM and forwards it to every actor; actors have X minutes (30 today) to be suspended, then ateom kills the rest. They resume on any worker with room.

Compared with the public preview upgrade:

- Upgrade order is still dataplane, then control plane.
- No node labels, no cloned pool, no manual drain, no version reads.

The alternative, a custom controller that starts a new worker alongside the old worker and moves capacity (CPU and RAM, by in-place pod resizing) one actor at a time as the old one drains, is post-GA. It defines the WorkerPool update semantic; switching to it later is a one-time change for users.

## Scope

Regular release upgrades within a release window: order, what each step does to running actors, rollback.

Out of scope: sandbox runtime upgrades (gVisor, microVM images), store engine upgrades, Kubernetes and node OS upgrades.

### Assumptions

- Upgrades stay within a release window (v0.x.y to v0.x.z). Across windows: reinstall.
- Within a window, ateom RPC changes are backward compatible. A new atelet must work against the previous ateom; the reverse is not supported.
- Control plane compatibility (API server, store schema) is covered by the release policy.

## Design

### Upgrade flow

Five applies, each followed by a wait:

1. Apply CRDs.
2. Upgrade ate-controller. Wait for Ready.
3. Upgrade the atelet DaemonSet. Wait until every node runs the new atelet.
4. Edit each WorkerPool `workerImage`. Wait for the rollout to complete on every pool.
5. Upgrade the rest of the control plane: ate-api-server, then atenet.

Rollback is the same list in reverse with the old images.

Progress UX: `kubectl rollout status` on the pool's Deployment and `kubectl get workerpools`.

### Actor contract

[...]

PR #1721, opened 2026-09-17
ate-setup: keep an existing SandboxConfig on deploy

HavenXia:
Deploy `ate-system` re-applied `sandboxconfig-gvisor.yaml` on every run, so a system upgrade could overwrite the user's `SandboxConfig` and change the sandbox runtime with it. 

For GA a system upgrade does not touch the sandbox runtime, so the default is now created only when missing.

Part of #1686.

- [ ] Tests pass


HavenXia:
Offline discussed with @zoez7:

The system upgrade does not touch `SandboxConfig`. The installation ships a default `SandboxConfig`, named with the VERSION tag (e.g. `substrate-gvisor-<VERSION>`) so it won't overwrite the existing config from the last version.  

The sandbox runtime version belongs to the user and changes only when they manually created a new `ActorTemplate` to use the new `SandboxConfig`. 

Therefore this PR is not necessary anymore.

PR #1723, opened 2026-09-17
benchmarking/locust: record cluster hardware facts and density frontiers

Nishanth29:
Part of #1590

First of two PRs. This one covers Kubernetes-side hardware discovery; the follow-up adds the Prometheus harvest.

## What this PR does

The locust runner records no information about the hardware it runs on, so trial results cannot be normalized across machine types or node counts. This adds cluster capacity discovery and derives actor density frontiers from it, persisting both the raw readings and the derived ratios to `stats.jsonl`.

### Cluster hardware discovery

New module `benchmarking/locust/cluster_facts.py`, kept separate from `runner.py`. It reads node count, machine type, allocatable cores and allocatable RAM, and the worker pod count via the official Kubernetes Python client, which resolves in-cluster and local kubeconfig auth without a kubectl subprocess.

Node and pod listing is proportional to cluster size, so `--no-cluster-facts` disables discovery. Discovery is best effort in either case: an unreachable API server or missing RBAC leaves the facts null and does not fail the run.

No value is defaulted or inferred. Unmeasured fields are `null`; a measured zero is `0`. The two remain distinguishable to any consumer.

### Density frontiers

Added to the `trial_summary` row in `stats.jsonl`:

| Field | Meaning |
|---|---|
| `actors_per_node` | peak actors over node count |
| `actors_per_vcpu` | peak actors over allocatable cores |
| `actors_per_gb_ram` | peak actors over allocatable RAM |
| `actors_per_pod_p50` / `_p90` / `_p99` | distribution of actors per worker pod |
| `aggregate_failure_ratio` | failures over requests, all RPCs |
| `<operation>_failure_ratio` | one per operation Locust reported, e.g. `dur_dir_write_failure_ratio` |

Actors per pod is reported as a distribution across the run's time samples (`p50` / `p90` / `p99`) rather than one average, because ramp-up and custom load shapes have no single user count to call steady.

The raw readings (`machine_type`, `node_count`, `allocatable_cores`, `allocatable_ram_gb`, `worker_pod_count`) are written alongside the derived ratios so they can be re-derived without re-running the trial.

### RBAC

Nodes are cluster scoped and require a ClusterRole with `list` on `nodes`. Pods require only a namespaced Role with `list` in `benchmark-workloads`. `locust.yaml` and `runner-job.yaml.tmpl` each define their own separately named pair.

Note that `locust.yaml` now binds a Role in `benchmark-workloads`, so that namespace must exist before the manifest is applied. `benchmarking/workloads/deploy.sh` creates it.

`status.json` is unchanged.

## How this was tested

11 unit tests in `test_cluster_facts.py` covering percentile boundaries, capacity scoped to only the nodes running worker pods, zero worker pods treated as a valid reading, unreadable facts returning null, and flag behavior.

Verified against a live GKE cluster. Discovery returned `c3d-standard-8`, 1 node, 7.91 allocatable cores, 27.73 GB, 5 worker pods, all matching `kubectl`. The emitted frontiers were re-derived by hand from `stats_history.csv` and matched. `--no-cluster-facts` returns all nulls.

Both manifests pass `kubectl apply --dry-run=client`, with no cluster-scoped name collisions between them.

## References

[Agent Substrate: Actor Density Benchmark Specs](https://docs.google.com/document/d/1sv5aBXvGOQ69iaqFxdPZ-d6tNbh5AMuwKODDyYjzYQw/edit?tab=t.0)

- [x] Tests pass


benchmarking/automation/manifests/runner-job.yaml.tmpl
```diff
@@ -41,6 +41,62 @@ roleRef:
   name: atelet-endpointslices
   apiGroup: rbac.authorization.k8s.io
 ---
+# Hardware discovery for the runner: runner.py reads node capacity and worker
```
maxsmythe: Probably too much explanation, can just say "named to avoid collision with ClusterRole defined by ..."

Or nothing at all is acceptable too :)

benchmarking/locust/cluster_facts.py
```diff
@@ -0,0 +1,276 @@
...
+        ratios = sorted(round(u / pod_count, 4) for u in observed)
+        n = len(ratios)
+        actors_per_pod_p50 = round(ratios[int(n * 0.50)], 2)
+        actors_per_pod_p90 = round(ratios[min(int(n * 0.90), n - 1)], 2)
+        actors_per_pod_p99 = round(ratios[min(int(n * 0.99), n - 1)], 2)
+
+    # Locust's Aggregated row provides the total directly. Resume (including
+    # cold starts) and Suspend have separate SLOs, and their CSV Name values
+    # lack the grpc_ prefix added in stats.jsonl.
+    wanted = ("Aggregated", "ResumeActor", "ResumeActorColdStart",
```
maxsmythe: These names can vary per-test (e.g. durdir test may look different from ping)

You can probably just do this for all measurements -- let the consumer identify if looking at a ratio makes sense.

benchmarking/automation/manifests/runner-job.yaml.tmpl
Nishanth29: gotcha, will prolly remove them then.

benchmarking/locust/cluster_facts.py
Nishanth29: sg, will drop wanted.. ig emitting `<name>_failure_ratio` dynamically for every row in stats.csv should work

benchmarking/README.md
```diff
@@ -114,6 +121,52 @@ You must have enabled otel tracing for your cluster to view traces.
...
+  own CSV output.
+* `logs.txt`, `traces.txt`: the runner log, and the trace IDs seen during the run.
+* `stats.jsonl`: one JSON object per line, one per metric. Every row carries
+  the same five keys: `timestamp`, `tag`, `test_name`, `metric`, and a flat
+  `measurements` map holding that metric's numbers.
+
+### Density frontiers
+
+With cluster discovery enabled, `stats.jsonl` gains a `trial_summary` row
+describing how densely actors packed onto the hardware. Its `measurements`
```
maxsmythe: nit: densely actors *are* packed

benchmarking/locust/cluster_facts.py
```diff
@@ -0,0 +1,266 @@
...
+
+    # The flag stands in only when no sample was read at all.
+    peak_users = max(observed) if observed else args.users
+
+    node_count = facts.get("node_count")
+    cores = facts.get("allocatable_cores")
+    ram_gb = facts.get("allocatable_ram_gb")
+    pod_count = facts.get("worker_pod_count")
+
+    actors_per_node = round(peak_users / node_count, 2) if node_count else None
```
maxsmythe: this assumes actors == users, which is not always true:

https://github.com/agent-substrate/substrate/blob/cdac9baef81dd319b46086d695266e6161e9e592/benchmarking/locust/runner.py#L100-L110

We either need to have some way of measuring actors directly or call this "users-per-node"

benchmarking/locust/manifests/locust.yaml
```diff
@@ -202,3 +202,50 @@ roleRef:
   kind: Role
   name: atelet-endpointslices
   apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
maxsmythe: AI found this one: when running locust directly, cluster facts is never invoked, making these permission sets moot.

benchmarking/locust/cluster_facts.py
```diff
@@ -0,0 +1,266 @@
...
+            failure_ratios = {"aggregate_failure_ratio": None}
+    else:
+        _log(logs, f"Notice: {stats_csv} not found; failure ratios unknown")
+
+    summary_entry = {
+        "timestamp": data_ts,
+        "tag": args.tag,
+        "test_name": args.name,
+        "metric": "trial_summary",
+        "measurements": {
```
maxsmythe, line 268: all measurement values are strings. Uniform types helps with ingestion pipelines.

maxsmythe (changes requested):
Getting closer, a bit more feedback.

benchmarking/locust/manifests/locust.yaml
Nishanth29: just confirming which way you want it. should i drop these rules, or add cluster facts collection to the web ui path so they're actually used?

benchmarking/locust/cluster_facts.py
Nishanth29: done, strings now. kept nulls as null since there's no string form of "not measured" that doesn't read as data.

benchmarking/locust/cluster_facts.py
Nishanth29: tryin to keep the actor naming and fixing the math instead: peak_users * (actors_per_user or 1) for all six keys, percentiles included. nothing changes for runs today since the flag is unset.

i did look at counting actors directly. the only actor gauge is ate.actor.stats.sampled_actors, but it only counts actors with a live resource measurement so suspended ones drop out. counting them properly would need a new server side gauge, which feels like its own change rather than part of this one.

would that work for you?

maxsmythe: I think this works?

A lot of this is fairly fragile -- for instance, node counts are taken at the end, so calculated density mid-run may be affected by autoscaling.

I think for now it's fine to do this and iterate, but we need to be mindful that these numbers are making assumptions about test shape and highlight that somewhere to avoid confusion down the road.

Nishanth29: agreed on the fragility. node count is the weakest link, it's one reading taken at the end, so if the cluster autoscales mid run the numerator and denominator are from different points in time.

i'll write the assumptions next to the keys in the benchmarking README:
- numerator is peak concurrent users times actors-per-user, so derived rather than observed
- denominator is a single end of run reading
- assumes every actor is alive at the peak

on the naming, sticking with actors_* since these are meant to answer how many actors fit on the hardware. users_per_node would describe the load generator config instead of the cluster....

benchmarking/locust/manifests/locust.yaml
Nishanth29: dropped (discussed offline)

benchmarking/locust/cluster_facts.py
```diff
@@ -0,0 +1,275 @@
...
+    peak_actors = peak_users * per_user
+
+    node_count = facts.get("node_count")
+    cores = facts.get("allocatable_cores")
+    ram_gb = facts.get("allocatable_ram_gb")
+    pod_count = facts.get("worker_pod_count")
+
+    actors_per_node = round(peak_actors / node_count, 2) if node_count else None
+    actors_per_vcpu = round(peak_actors / cores, 2) if cores else None
+    actors_per_gb_ram = round(peak_actors / ram_gb, 2) if ram_gb else None
```
roycaihw: Are these active actors or total actors?

benchmarking/README.md
```diff
@@ -114,6 +121,65 @@ You must have enabled otel tracing for your cluster to view traces.
...
+  Capacity covers the nodes the worker pods are running on rather than the
+  whole cluster, so a separate infrastructure pool is not counted. They are
+  recorded so the ratios below can be re-derived later, or recomputed against
+  a different denominator.
+* `actors_per_node`, `actors_per_vcpu`, `actors_per_gb_ram`: the most actors
+  Locust reported running, over the matching capacity. The `-u` flag only
+  stands in when no sample was read.
+* `actors_per_pod_p50`, `actors_per_pod_p90`, `actors_per_pod_p99`: actors per
+  worker pod across the run. Reported as a distribution rather than one
+  average, and it spans ramp-up too, because a custom load shape has no
```
roycaihw: Shall we measure steady state only instead of including ramp-up to reduce the confusion?

roycaihw (approved):
LGTM

benchmarking/locust/cluster_facts.py
Nishanth29: its total...( peak virtual users times --actors-per-user) so suspended actors count too. The active number is server side in #1725, cluster_packing.assigned_workers.

benchmarking/README.md
Nishanth29: so for fixed-user runs (like sweperf or glutton), p50/p90/p99 already reflect steady state since ramp-up is <10% of the run and stays below p10. 

kept the full window here so step-ladder / custom LoadTestShape runs work too (since they don't have a single users / spawn_rate cutoff), while #1725 slices to steady state (start + users/spawn_rate) on the server side.

maxsmythe (approved):
LGTM

PR #1726, opened 2026-09-17
Atunnel <> PEP contract documentation

keithmattix:
As we're getting closer to GA, it's important to start documenting the network contracts, especially where we want things to be pluggable. This PR takes a stab at documenting the contract between atunnel and the egress PEP. It doesn't have a policy section since that's evolving, but this aims to be a good first step.

/cc @EItanya @thockin @bowei @LiorLieberman 

- [X] Tests pass


docs/network-egress.md
```diff
@@ -0,0 +1,67 @@
...
+
+Last updated: 09/17/2026
+
+## Overview
+
+Agent Substrate enforces that all outbound traffic from an actor MUST go through an egress policy enforcement point (PEP). That PEP can (optionally) inspect, filter, or redirect the traffic according to substrate-native egress policies. This document describes the contract between substrate and egress PEPs.
+
+## Transport
+
+Actor traffic exits the sandbox through [`atunnel`](./glossary.md#components). At present, TCP traffic is forwarded on to the PEP while all other protocols are blocked; the one exception is that UDP with destination port 53 is allowed to bypass the PEP for DNS. `atunnel` is configured with the PEP's network address and port at startup and uses that address to open an mTLS HTTP/1.1 CONNECT tunnel to the PEP. The CONNECT request target and `Host` header are both set to the original destination IP and port that `atunnel` derived from the actor's TCP connection. Actor networking is currently IPv4-only.
```
bowei: TCP 53 DNS?

bowei: Do we require HTTP/1.1 CONNECT or HTTP/2 is ok as well?

docs/network-egress.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Trust Boundary
+
+In substrate, we consider everything originating from the actor to be untrusted. The egress PEP MUST NOT trust any decision or attestation coming from the actor itself; it MUST instead trust only information carried through substrate-controlled channels (e.g. the actor certificate and the CONNECT request produced by atunnel). For example, the PEP should not interrogate the actor's HTTP request to determine the original destination port; it should instead use the CONNECT request target set by atunnel. PEPs that are implementing substrate egress policy MAY use actor traffic characteristics to match egress policies or as a hint of actor intent.
```
bowei: Isn't the port actually ok to use? I think we should use the hostname example as that is the case that is critical.

docs/network-egress.md
```diff
@@ -0,0 +1,67 @@
...
+
+In the default Kubernetes deployment, the PEP obtains its certificate and private key through a projected `podCertificate` volume using substrate's Service DNS signer (`servicedns.podcert.ate.dev/identity`). The signer derives the certificate's DNS SANs from the Services that select the PEP Pod, so the PEP MUST be selected by a Service and the hostname configured in `atunnel` MUST match one of those Service DNS names. For the default PEP, that name is `atenet-egress.ate-system.svc`. The projected `credential-bundle.pem` contains the serving certificate chain and private key, while the `ClusterTrustBundle` for the same signer supplies `atunnel` with the corresponding trust anchors.
+
+## Client Authorization
+
+`atunnel` MUST present an actor-specific client certificate when connecting to the PEP. At actor activation, `atunnel` generates a private key and requests a short-lived certificate from substrate through the node-local atelet. The private key remains in `atunnel`. The certificate identifies the actor by its atespace, name, and UID and is scoped to the `atunnel` purpose.
```
bowei, line 25: Does atunnel deal with all the key stuff or it's all through atelet?

docs/network-egress.md
```diff
@@ -0,0 +1,67 @@
...
+
+The PEP MUST require and verify the client certificate before accepting a CONNECT request. It MUST verify the certificate chain and lifetime against the Actor Identity CA, require the client authentication usage, and require exactly one valid `ActorIdentity` extension with OID `1.3.6.1.4.1.11129.2.12.2` whose atespace, actor name, actor UID, and purpose are present. The purpose MUST be `atunnel`, and the certificate's actor URI SAN MUST identify the same actor as the extension. The PEP MUST then confirm with ate-api-server that the actor still exists, that its current UID matches the certificate, and that it is running. The PEP MUST reject the CONNECT request if any of these checks fail.
```
bowei: note about OID value this will change, prefer just say "substrate OID"

bowei (commented):
thanks for writing this up

docs/network-egress.md
keithmattix: Oh good point; forgot tcp DNS skips the PEP; will fix. Right now, according to the code, atunnel always sends h1 CONNECT. To your point though, I do think we should allow h2 (probably even prefer it)

docs/network-egress.md
keithmattix: I chose the port because there was a nice contrast; you should use the port from the CONNECT Host, NOT the port from the inner request. I can switch to hostname though

docs/network-egress.md
keithmattix: From what I could tell, atunnel is just a server routine within ateom, not an actual executable. I just use the atunnel name here because we're referring to the networking functionality

docs/network-egress.md
keithmattix: Oh didn't realize that it was changing. FWIW though, I do think there's value in documenting the current value somewhere outside of the code.

bowei: OID is currently Google corp. It will change when CNCF donation completes.

docs/network-egress.md
keithmattix: I double checked: DNS does not currently skip the PEP, but I made it do so in this PR based on slack conversation. I also think we should tackle h2 in a follow up

docs/network-egress.md
keithmattix: Used hostname as the example

docs/network-egress.md
keithmattix: I kept the current OID but noted that it will change at the donation

bowei (approved):

docs/network-egress.md
LiorLieberman: Doesn't #1689 changing this as well? wrt to DNS

keithmattix: That's one reason I dated this; we'll need an update once that changes

PR #1728, opened 2026-09-17
Revert "e2e: copy the source pool's worker resources"

ahmedtd:
Reverts agent-substrate/substrate#1691 - e2e tests are failing at head.


BenTheElder (approved):
let's unbreak main and I'll reintroduce

basically we're too tight on the resources in the runner.

but we do need the parking tests to stop using an unsized pool ...

PR #1729, opened 2026-09-18
CSI volume support: follow-up improvements

### LINKED ISSUE #1077: CSI volume support: follow-up improvements from PR #552

anishgang:
Resolves #1077

This PR implements the follow up improvements from PR #552 across 5 commits:

1. **volume: factor out generic LookupPlugin helper**: Deduplicates plugin lookup logic between `cmd/ateapi` and `cmd/atelet`.
2. **csi: harden CSIDriverConfig ControllerEndpoint validation**: Adds OpenAPI regex/length constraints and runtime URI/port validation.
3. **volume/csi: query and cache driver capabilities at plugin init**: Caches capabilities during plugin initialization and avoids calling unsupported endpoints (example: hostpath attach/stage).
4. **atelet: replace CSIDriverConfig shared informer with direct client**: Replaces the cluster-wide shared informer watch with direct client `Get` cached per worker.
5. **volume: support configurable access modes and propagate PublishContext**: Adds `VolumeAccessMode` enum and propagates `PublishContext` from `ControllerPublishVolume` to node mounts.

### Testing
- All repository unit tests passed (`make test`)
- All linters and verifiers passed (`make verify`, `kube-api-linter`, `metrics.sh`)


internal/volume/csi/client.go
```diff
@@ -40,16 +43,45 @@ func parseEndpoint(endpoint string) (string, string, error) {
 	}
 	switch u.Scheme {
 	case "unix":
-		if u.Path == "" {
-			return "", "", fmt.Errorf("unix endpoint missing path: %s", endpoint)
+		if !strings.HasPrefix(endpoint, "unix:///") || strings.HasPrefix(endpoint, "unix:////") || u.Host != "" || u.Path == "" {
```
Sneha-at, line 46: @hajiler I am unsure if we want to add some regex to perform the validation here. Do we have any reference of such validation anywhere in this or other repo? I am wondering if regex it the best option here.
Issue:
>Current behavior: validated only by +kubebuilder:validation:Pattern= + "^(tcp|dns)://.+$", which accepts effectively any string after the scheme and has no length bound.
Proposed fix: tighten the CEL/regex validation to a real host:port (or dns:/// target) form, add a MaxLength, and consider a validating webhook for anything the schema can't express. This field determines where the control plane dials, so a permissive value is a request-forgery vector.

anishgang: I think the regex is only doing a rough first pass though. It's
in the CRD schema, and OpenAPI doesn't give us anything else for strings, so it's
really just there to catch typos at apply time I am pretty sure. The actual validation should be this function, and it doesn't use regex at all. It parses with url.Parse, net.SplitHostPort and strconv.Atoi, mostly because IPv6 like tcp://[::1]:9000 and a port range of 1 to 65535 are both pretty difficult to do in
regex.
For precedent, driverName on this same CRD already had a Pattern before this PR, and SandboxConfig and WorkerPool have them too.
If we wanted we could maybe swap the schema regex for CEL, though CEL has no
URL parser so it comes down to splitting on :, which might break IPv6 (not fully sure, would need to look into it)? Happy to add it if you think it's worth it
One thing I noticed because of your comment: bare unix:/// looks like it slips
through here even though the CRD rejects it, since url.Parse gives Path == "/" rather than "". Small fix if you think this is necessary

cmd/ateapi/internal/controlapi/volumes.go
```diff
@@ -109,13 +112,16 @@ func createActorVolumes(ctx context.Context, registry VolumePluginRegistry, scLi
...
-		plugin, err := registry.GetPlugin(ctx, vol.GetVolumeType())
+		plugin, err := volume.LookupPlugin(ctx, registry.GetPlugin, vol.GetVolumeType())
 		if err != nil {
 			return resultVolumes, status.Errorf(codes.FailedPrecondition, "failed to get volume plugin for driver %q (StorageClass %q): %v", sc.Provisioner, scName, err)
 		}
 
-		storageVolumeID, volCtx, volErr := plugin.CreateVolume(ctx, actVolID, specVol.GetExternalVolumeTemplate().GetCapacity(), sc.Provisioner, sc.Parameters)
+		storageVolumeID, volCtx, volErr := plugin.CreateVolume(ctx, actVolID, specVol.GetExternalVolumeTemplate().GetCapacity(), sc.Provisioner, sc.Parameters, vol.GetAccessMode())
 		if volErr != nil {
+			if status.Code(volErr) == codes.InvalidArgument {
```
Sneha-at: A lot of other error codes are also not considered internal like `ResourceExhausted`, `PermissionDenied` etc. If we are filtering the Invalid Argument, we should also look at other user error codes.

cmd/ateapi/internal/controlapi/volumes.go
```diff
@@ -235,24 +243,56 @@ func detachActorVolumes(ctx context.Context, st detachActorVolumesStore, registr
...
+		plugin, err := volume.LookupPlugin(ctx, registry.GetPlugin, vol.GetVolumeType())
 		if err != nil {
-			errs = append(errs, fmt.Errorf("failed to get volume plugin for %q: %w", vol.GetVolumeType(), err))
+			errs = append(errs, err)
 			continue
 		}
 		if err := plugin.DetachVolume(ctx, vol.GetStorageVolumeId(), node); err != nil {
 			if status.Code(err) == codes.NotFound {
 				slog.WarnContext(ctx, "Volume not found during detach, assuming already detached", slog.String("volume_id", vol.GetStorageVolumeId()), slog.String("node", node))
+				vol.PublishContext = nil
```
Sneha-at: We don't use this volume parameters anywhere below, we set teh publishContext and the ContextNode explicitly
```
if toVol.GetVolumeName() == cVol.GetVolumeName() {
						toVol.PublishContext = nil
						toVol.PublishContextNode = ""
```

anishgang: You're right, I don't think the two lines do anything. The closure only uses Vol.GetVolumeName() and sets the fields explicitly, and on success we replace actor. Status with the stored one anyway. Switching clearedVols to a list of names.

cmd/ateapi/internal/controlapi/volumes.go
anishgang: Yeah makes sense. I only added InvalidArgument because that's what a driver returns for
an unsupported access mode, but the same applies to like OutOfRange, ResourceExhausted,
PermissionDenied and Unimplemented, which are all the caller's problem rather than
ours. I will probably be going with an allowlist rather than passing everything through, mainly so a driver's
NotFound doesn't surface as "actor not found" to clients. Also, I will be expanding the existing test
into a table so each code is covered.

cmd/ateapi/internal/controlapi/workflow_resume.go
```diff
@@ -623,27 +627,62 @@ func schedulingConstraints(actor *ateapipb.Actor, tmpl *ateapipb.ActorTemplate)
 // assigned worker's node. Attachment is idempotent, so a re-entered workflow
 // safely runs it again.
 // TODO replace re-execution with a proper check on the volumes' attach state.
-func (w *ActorWorkflow) ensureVolumesAttached(ctx context.Context, actor *ateapipb.Actor, worker *ateapipb.Worker, actorTemplate *ateapipb.ActorTemplate) (err error) {
+func (w *ActorWorkflow) ensureVolumesAttached(ctx context.Context, actorRef resources.ActorRef, actor *ateapipb.Actor, worker *ateapipb.Worker, actorTemplate *ateapipb.ActorTemplate) (_ *ateapipb.Actor, err error) {
```
Sneha-at: Is there any reason why we want to update the actor reference here to instance? I see it might be because of a test case added below TestAttachActorVolumes_PubContextChangeOnSameNode, is there any other way to test this? We can minimize code changes if there is no additional benefit

anishgang: Mostly consistency. ensureVolumesCreated and ensureWorkerAssigned right above it both
return the updated actor and have the caller reassign, so this just matches them. Nothing actually breaks without it though. The reason I'd keep it: without the return, the publish context only reaches the
restore step because getMountedActorVolumes hands back the same volume objects we
just mutated. So it works, but only because of how the memory happens to be shared.
If someone later changed that function to return copies, volumes would mount
without their device paths and no test would fail. Returning the actor makes that more so
"explicit" instead of accidental. Happy to rework the test to check the store instead if you want.

Issue #1604, opened 2026-09-10
Decouple and minimize atelet/ateom shared paths


HavenXia:
## Problem

Today ateom and atelet compute many per-actor directory paths from `actorUID` through the same `internal/ateompath` package. 

Theses directories are only created and wiped by atelet, and the RPC request between atelet/ateom only carries `actorUID` instead of these directory paths so ateom has to recompute the paths on its end.

## Proposal

atelet owns the `actorUID` related paths and passes it in the RPCs. Can wrap them in a message (e.g. `ActorDirs`) and pass it to atelet RPCs.
```
root_dir
bundle_dir
runsc_state_dir
pid_file_dir
checkpoint_dir
restore_dir
durable_dir_volume_mounts_dir
system_info_volume_roots_dir
volumes_dir
```
These directories are read or writes in ateom as a whole. For per-item paths like pid file ateom can still compute from directory+name, both from the request instead of shared path logic.

We still need to keep some shared paths for atelet/ateom: per-ateom directory and socket, the netns name, the credential broker socket and the OLTP socket.

HavenXia:
cc @BenTheElder

PR #1730, opened 2026-09-18
Add `ActorDirs` message to carry per-actor directories 

HavenXia:
Today `atelet` and `ateom` derive the per-actor directories (oci bundles, runsc state, pid files, checkpoint and restore state, durable-dir, system-info and volume roots) from the actor UID through the same package, `internal/ateompath`. 

Add an `ActorDirs` message to RunWorkloadRequest, RestoreWorkloadRequest, CheckpointWorkloadRequest and TerminateWorkloadRequest, and have atelet fill it with the directories it prepared.

This is the first part of #1604 to decouple atelet and ateom shared pathes.


BenTheElder (commented):
so for dirs that atelet feeds to ateom, let's split those off to a package internal to atelet so we know ateoms are not assuming it? (and vice versa, atelet should stop using ateompath)

there should probably be a third package for the socket paths they share

HavenXia:
> so for dirs that atelet feeds to ateom, let's split those off to a package internal to atelet so we know ateoms are not assuming it? (and vice versa, atelet should stop using ateompath)
> 
> there should probably be a third package for the socket paths they share

I agree on the shape for 3 packages: `atelet/internal/` for this PR's  pathes,  a shared path for shared pathes and an ateom specific path for ateoms. 

But this PR add only the proto and RPC, the follow up PRs will do the refactors, do you want me to have them together in this PR?

internal/proto/ateompb/ateom.proto
```diff
@@ -106,6 +106,30 @@ service Ateom {
   rpc TerminateWorkload(TerminateWorkloadRequest) returns (TerminateWorkloadResponse) {}
 }
 
+// ActorDirs is the set of per-actor directories atelet prepares on the volume
+// it shares with ateom. ateom takes them from the RPC instead of deriving
+// them from the actor UID.
+// Every directory is under root_dir.
+message ActorDirs {
+  string root_dir = 1;
```
juli4n, line 115: Document

BenTheElder:
> But this PR add only the proto and RPC, the follow up PRs will do the refactors, do you want me to have them together in this PR?

I think we want to reason about what these paths are and how they should be split more easily. If you don't have time a follow-up is OK, but to me that part is more important than cramming the paths into the RPCs. Otherwise you can't tell if the RPC change adequately captured the path knowledge.

internal/proto/ateompb/ateom.proto
HavenXia: added

HavenXia:
> > But this PR add only the proto and RPC, the follow up PRs will do the refactors, do you want me to have them together in this PR?
> 
> I think we want to reason about what these paths are and how they should be split more easily. If you don't have time a follow-up is OK, but to me that part is more important than cramming the paths into the RPCs. Otherwise you can't tell if the RPC change adequately captured the path knowledge.

Added the refactor part into this PR, the removing of ateom itself and using the ActorDirs message from RPC is put into next PR

HavenXia:
rebased over #1689, @BenTheElder PTAL

internal/nodepath/nodepath.go
```diff
@@ -0,0 +1,77 @@
+// Copyright 2026 Google LLC
```
BenTheElder: Aside: we will eventually export this package when atelet/ateom settle down.

tools/validate-image-cache/main.go
```diff
@@ -151,10 +152,13 @@ func (c runConfig) validate() error {
 // newStore opens the cache with the options both modes share: min-age
 // from --evict-idle, and bundle-spec rooting from the node's actors dir
 // (absent on a validation host, which InUse treats as an empty root set).
+// actorsDir is where atelet keeps the per-actor directories.
+var actorsDir = filepath.Join(nodepath.BasePath, "actors")
```
BenTheElder: 🤖 **nit** 🟢 – This `var` landed between `newStore`'s doc comment and the function, so godoc now attaches the `newStore` description to `actorsDir`, and `newStore` has none. Moving the var above the comment block fixes it.

Also, this copies `ateletpath.ActorsDir` as a literal because `tools/` can't import `cmd/atelet/internal`. The two can drift silently. Since more than one owner needs this path, `internal/nodepath` may be the better home for it.

internal/nodepath/nodepath.go
```diff
@@ -0,0 +1,77 @@
...
+
+// BasePath is the root shared folder on the host filesystem, mounted at the
+// same path into the atelet and ateom containers.
+const BasePath = "/var/lib/ateom-gvisor"
+
+// AteomSupportSocket is the node-local atelet socket used by atunnel
+// to request credentials for the worker's current actor assignment.
+var AteomSupportSocket = filepath.Join(BasePath, "ateom-support.sock")
+
+// AteletOTLPSocketPath is the node-scoped unix socket atelet serves the OTLP
```
BenTheElder: Note to self: we will drop the relay and just let ateom push directly following #1689 🎉

kubeconfig
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
BenTheElder: delete this file / purge it from git

internal/resources/snapshot.go
```diff
@@ -34,6 +34,13 @@ const (
 	tagsOwnerKind   = "tags"
 )
 
+// DurableDirTarFile is the snapshot file holding the tar of an actor's
```
BenTheElder: This is supposed to be opaque to the snapshot tooling it should not know about the specific files.

internal/ateompath/ateompath.go
```diff
@@ -175,54 +77,26 @@ func RunscDebugLogDir(actorUID, containerName string) string {
...
-	)
-}
-
-// LocalSnapshotDir is the directory holding one named local (pause) snapshot
-// of an actor: the checkpoint files plus their manifest.
-func LocalSnapshotDir(actorUID, snapshotName string) string {
-	return filepath.Join(LocalCheckpointsDir(actorUID), snapshotName)
-}
-
-// DurableDirTarFile is the snapshot file holding the tar of an
```
BenTheElder: OK so maybe it's not for this PR, but please file a TODO, this is a really bad API, atlet should just ask for a data snapshot if that's what it wants and we should not take a full snapshot to begin with, nor know about specific snapshot filenames.

Then atelet can upload whatever files happen to be produced by the snapshot, regardless of type.

@dberkov something we should roll up with any updated snapshot lifeycle.

tools/validate-image-cache/main.go
HavenXia: fixed the godoc issue, and move ActorDirs to internal/nodepath.

internal/nodepath/nodepath.go
HavenXia: ACK

internal/resources/snapshot.go
HavenXia: moved to a separete snapshotfiles.go next to it

internal/ateompath/ateompath.go
HavenXia: Added TODO

BenTheElder: This should probably be a tracking issue, beyond a code todo. This is a bug that will limit our options in the future.

FYI @msau42

internal/ateompath/ateompath.go
```diff
@@ -12,134 +12,42 @@
...
-func ActorNetNSPath(actorUID string) string {
-	return filepath.Join("/run/netns", ActorNetNSName(actorUID))
-}
-
-// ActorResolvConfPath is the resolver bind source outside the actor's rootfs.
-func ActorResolvConfPath(actorUID string) string {
-	return filepath.Join(ActorPath(actorUID), "resolv.conf")
-}
+// ActorsDir is the parent of the per-actor directories; see ActorPath.
+var ActorsDir = filepath.Join(nodepath.BasePath, "actors")
```
BenTheElder: 🤖 **nit** 🟢 – This is still a second copy of `nodepath.ActorsDir`, so the move to `internal/nodepath` stops short of the ateoms. atelet now builds actor directories from `nodepath.ActorsDir`, and the ateoms build them from this one. The values match today, but a change to either, or a test that overrides only one, would send the ateom looking under a different root than the one atelet prepared.

Nothing outside this package references it, so `ActorPath` can use `nodepath.ActorsDir` directly and this var can go. The package is transitional anyway, so it's cheap to close now.

internal/ateompath/ateompath.go
HavenXia: https://github.com/agent-substrate/substrate/issues/1846

internal/ateompath/ateompath.go
HavenXia: entire ateompath.go will be removed in the followup pr, but fixing it here to make it clear.

internal/proto/ateompb/ateom.pb.go
```diff
@@ -212,6 +212,127 @@ func (StatsSource) EnumDescriptor() ([]byte, []int) {
...
+// them from the actor UID.
+// Every directory is under root_dir.
+type ActorDirs struct {
+	state protoimpl.MessageState `protogen:"open.v1"`
+	// The actor's own directory.
+	RootDir string `protobuf:"bytes,1,opt,name=root_dir,json=rootDir,proto3" json:"root_dir,omitempty"`
+	// Holds one OCI bundle per container, named after the container.
+	OciBundleDir string `protobuf:"bytes,2,opt,name=oci_bundle_dir,json=ociBundleDir,proto3" json:"oci_bundle_dir,omitempty"`
+	// runsc --root for the actor's sandbox. gVisor only.
+	RunscStateDir string `protobuf:"bytes,3,opt,name=runsc_state_dir,json=runscStateDir,proto3" json:"runsc_state_dir,omitempty"`
```
BenTheElder: TODO: can we eliminate ateom knowing about gvisor vs ...? it mostly doesn't.

internal/proto/ateompb/ateom.pb.go
```diff
@@ -212,6 +212,127 @@ func (StatsSource) EnumDescriptor() ([]byte, []int) {
...
+	CheckpointDir string `protobuf:"bytes,5,opt,name=checkpoint_dir,json=checkpointDir,proto3" json:"checkpoint_dir,omitempty"`
+	// Where the snapshot files RestoreWorkload reads have been downloaded to.
+	RestoreDir string `protobuf:"bytes,6,opt,name=restore_dir,json=restoreDir,proto3" json:"restore_dir,omitempty"`
+	// Holds one subdirectory per durable-dir volume.
+	DurableDirVolumeMountsDir string `protobuf:"bytes,7,opt,name=durable_dir_volume_mounts_dir,json=durableDirVolumeMountsDir,proto3" json:"durable_dir_volume_mounts_dir,omitempty"`
+	// Holds one subdirectory per system-info volume.
+	SystemInfoVolumeRootsDir string `protobuf:"bytes,8,opt,name=system_info_volume_roots_dir,json=systemInfoVolumeRootsDir,proto3" json:"system_info_volume_roots_dir,omitempty"`
+	// Holds one mount point per CSI volume.
+	VolumesDir    string `protobuf:"bytes,9,opt,name=volumes_dir,json=volumesDir,proto3" json:"volumes_dir,omitempty"`
+	unknownFields protoimpl.UnknownFields
```
BenTheElder: hmmmm

internal/proto/ateompb/ateom.pb.go
BenTheElder: for similar APIs we got rid of atelet knowing about this vs ateom knowing it. need to trace this one to see why we have to pass it between them vs deriving it in an opaque way without atelet carrying this knowledge at all.

HavenXia: okay, they can be removed from ActorDir as they are not directly write by atelet but only read by ateom-gvisor.

internal/proto/ateompb/ateom.pb.go
HavenXia: These two should be kept since atelet does write to these folders and ateom-gvisor reads these folders to write.

PR #1731, opened 2026-09-18
workersync: make the list retry backoff per-syncer [flake fix]

BenTheElder:
The startup registered-worker scan read its backoff from a package-level var that tests overwrote. Syncers started by earlier tests outlive them and keep reading that var, so the write races the read and `go test -race` fails, in whichever test happens to do the write.

Hold the schedule on the syncer instead, so a test only ever shrinks the backoff of the syncer it owns.

Fixes a flake I encountered on one of my other PRs. I didn't see an open issue or fix but this is pretty straightforward.


juli4n (approved):
Thanks!

PR #1733, opened 2026-09-18
Consolidate subcommands for each resource on a single file.

juli4n:
As the number of resources in the Substrate API grows, having one file for each (resource x verb) is becoming hard to manage. We recently started consolidating all commands for a each resource in a single file (see tag.go), similar to how we organize RPC handlers in the ate-apiserver `controlapi` package.

This PR re-organizes the resources that are sill spread across multiple files.


bowei:
What about each resource in a separate package?

We should be careful if a file gets above 1kLoC

juli4n:
Sure, we can consider that if resources grow too much. I think we are ok for now?

laoj2 (approved):

PR #1737, opened 2026-09-18
Use deployment configuration for the router status Service IP

### LINKED ISSUE #1457: atenet router /statusz: Service IP panel is forbidden by the router's own shipped RBAC
## Summary

The atenet router's `/statusz` page (`--status-port`, default 4040) has a "Router Service IP" panel that tries to `get` the `atenet-router` Service via the Kubernetes API — but the RBAC the installer ships for the router's ServiceAccount never grants access to Services. The panel therefore renders a raw forbidden error on every install:

```
Router Service IP
Lookup Failed: services "atenet-router" is forbidden: User "system:serviceaccount:ate-system:atenet-router"
cannot get resource "services" in API group "" in the namespace "ate-system"
```

## Verification (live install, substrate rev 0d85e8572161 on GKE)

The ServiceAccount's only binding is the endpointslices role:

```
$ kubectl auth can-i get services -n ate-system --as=system:serviceaccount:ate-system:atenet-router
no

$ kubectl get role atenet-router-endpointslices -n ate-system -o jsonpath='{.rules}'
[{"apiGroups":["discovery.k8s.io"],"resources":["endpointslices"],"verbs":["get","watch","list"]}]
```

So the statusz code (`cmd/atenet/internal/router/status.go`) calls an API that its own deployment manifests forbid — the panel is dead on arrival, not a cluster-specific misconfiguration.

## Everything else on the page works

For context, the rest of `/statusz` renders fine and is genuinely useful (build rev, port allocation, request-parking state, dataplane / Kubernetes API / ATE API health checks with counters, full CLI args and resolved flags). This is only about the Service IP panel.

## Suggested fix (either)

1. Add `services` `get` (scoped to the router's namespace, arguably even `resourceNames: [atenet-router]`) to the `atenet-router-endpointslices` role — or a small dedicated role, since the existing one is named for endpointslices; or
2. Drop the API lookup: the page already knows the namespace, and the Service IP could be injected via env/downward-API-adjacent means at deploy time, avoiding the extra RBAC surface for a display-only field.

Option 2 keeps the router's RBAC minimal, which the current single-purpose role suggests was the intent.

robert-cronin:
Fixes #1457

The shipped router's `/statusz` handler tries to read a Service that its ServiceAccount cannot access. Read optional `ROUTER_SERVICE_IP` deployment configuration instead, mapped from Kubernetes' `ATENET_ROUTER_SERVICE_HOST`, and install the Service before the Deployment. The handler no longer makes the synchronous Service API request; the existing EndpointSlice-only Role is sufficient.

Label the address as a container-start snapshot, independent of `--namespace`. Missing or invalid configuration displays **Unavailable** in HTML and an empty `router_cluster_ip` string in JSON. Service recreation can leave the snapshot stale until restart. Document these semantics and cover IPv4, IPv6, missing/invalid values, HTML/JSON output, namespace independence, and zero Kubernetes API actions.

Validation: `make verify` and the router packages under `-race` passed. A full local Envoy installation passed the gVisor and micro-VM E2E lanes, including the separate MITM trust and egress-networking runs for both runtimes. Live HTTP checks covered fresh full/raw installs, missing/disabled injection, an unrelated namespace, egress, and readiness; real ServiceAccount checks denied Service access while preserving EndpointSlice permissions. The Kind volume tests used the repository-supported hostpath CSI driver because this host lacks NFS support.

- [x] Tests pass


keithmattix (commented):
If we're assuming a static service name, why not use the ATENET_ROUTER_SERVICE_HOST env var that gets auto injected

bowei (dismissed):

robert-cronin:
Switched this to `ROUTER_SERVICE_IP`, mapped from Kubernetes' `ATENET_ROUTER_SERVICE_HOST`, and removed the Service lookup and added RBAC grant. The manifest installs the Service first. `/statusz` labels the value as a container-start snapshot and shows **Unavailable** when injection is missing or invalid (an empty string in JSON); recreating the Service can leave it stale until restart.

Validated fresh full/raw installs, missing/disabled injection, namespace independence, egress, and readiness on Kind. `make verify` and both full Envoy gVisor/micro-VM E2E lanes passed, including the separate MITM trust and egress tests. Volume tests used hostpath CSI on the local cluster.

Issue #1538, opened 2026-09-08
Generalize EgressPolicy header modification with literal and secret-backed values


EItanya:
The current `inject_static_headers` API requires a credential reference for every header. Replace this credential-specific model with a generic request header modifier supporting `set`, `add`, and `remove`, with secrets as one possible value source.

```yaml
effects:
  requestHeaderModifier:
    set:
      - name: X-Client-Version
        value: "2026-09"
      - name: Authorization
        secret:
          uri: substrate-secret://example.com/provider/api-token
          prefix: "Bearer "
    add:
      - name: X-Client-Tag
        value: substrate
    remove:
      - X-Internal-Debug
```

Proposed types:

```proto
message RequestHeaderModifier {
  repeated HeaderValue set = 1;
  repeated HeaderValue add = 2;
  repeated string remove = 3;
}

message HeaderValue {
  string name = 1;

  oneof source {
    string value = 2;
    SecretHeaderValue secret = 3;
  }
}

message SecretHeaderValue {
  string uri = 1;
  string prefix = 2;
}
```

Expose `RequestHeaderModifier` through `EgressRuleEffects.request_header_modifier`. This follows Gateway API’s `RequestHeaderModifier` operation model while extending header values to support secret references.

**Behavior**

- `set` replaces all existing values for the named header.
- `add` appends a value while preserving existing values.
- `remove` deletes all values for the named header; an absent header is a no-op.
- Header names are case-insensitive. Each header may appear only once across the operations, making operation order irrelevant.
- Values are either literals or secret references. Secret-backed values are constructed as `prefix + resolved secret`, preserving support for `Bearer <token>` without a template language.
- The modifier applies only to requests authorized by the first matching hostname rule.

Related: #1317 — Add extension points to egress policy.

LiorLieberman:
Responded in #1751 but leaving here as well.

While its true that we _may_ eventually need to support a full blown model of header operations, I am worried we are trying to port more Gateway API features in a rush and bringing unnecessary complexity at the current stage of the project.

For this one specifically we discussed and decided we dont do that for GA

IMP we should revisit based on feedback

EItanya:
My issue is that the API we're adding is incredibly specific.  I'm fine with just adding secret/credential values in the header API we release, but having the whole API be called, `secret_credential_inject` means there's no room to move from there.

PR #1751, opened 2026-09-18
egress policy enhacements for GA

LiorLieberman:
**BREAKING** API changes prior in preparation for GA. 

Follow up 2 Egress dedicated meetings from the week, and the [enhancements doc](https://docs.google.com/document/d/1XLCo9SMROddQzwsyEhdBNWuLYFtWWZWfYvGlIO1BcUI/edit?resourcekey=0-st2SqKp-O68nv5LeIUfA-g&tab=t.0) that came out of it. 

This PR has proto-only change to the EgressPolicy API. No implementation; the gateway,
store contract, and e2e helpers stop compiling until the follow-up lands.

- `EgressRule` is now a union of protocol handlers: `http`, `https`,
  `tls_passthrough`. One ordered list, first match decides, deny by default.
- The `hostnames`, `cidrs`, and `all` rule kinds are removed. **No field numbers
  or names are reserved: we are pre-GA and existing policies must be recreated.**
- `http`: cleartext HTTP, matched per request on the authority and port. Carries `effects`.
- `https`: MITMed HTTPS that the gateway intercepts . [[Matched on SNI and port at the
  ClientHello, then per request on the authority;]]. Carries `effects`. MITM is off unless a name is
  listed here.
- `tls_passthrough`: TLS forwarded without decryption, matched once per
  connection on SNI and port. No effects.
- Name fields are `host_patterns` on `http` and `https`, `sni_patterns` on
  `tls_passthrough`. Same wildcard grammar as before, documented once on
  `HTTPRule.host_patterns`.
- `ports` added to all three handlers, 1 to 65535, empty means any port

- Credential injection is now conditional: a header is replaced only when the
  Actor's request already carries it, with any placeholder value. Requests
  without the header pass unchanged. Comments only, no new fields.

- Unsupported in v1 rules for CIDRRules, arbitrary TCP, non-TCP protocols.

- `google/protobuf/empty.proto` import dropped; nothing uses it now.

Questions

- how do we enable **all** SNIs -- e.g MITM all SNIs or inject header to all hosts? 

- The handler is called `https`, not `mitmHttps`. Everything in an `https`
  block is intercepted; the block name alone has to carry that. Is that clear
  enough, or do we want "intercepted" in the field name after all?
- CIDR rules are removed and not part of v1. We revisit them with the `tcp`
  handler. (TODO: lior to include meeting notes:).
- `inject_static_headers` and `CredentialHeaderInjection` still say "inject"
  while the behavior is replacement. Should we rename to `replace_headers` now, while we
  are already breaking the API?

follow-ups
- align egress implementation

- [x] Tests pass


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -399,40 +394,45 @@ message EgressPolicy {
...
-  // Matches when the original destination IP belongs to any configured prefix.
+  // HTTPS to the named hosts, intercepted (man-in-the-middle) so the gateway
+  // can authorize and modify each request.
   //
   // +k8s:optional
   // +k8s:unionMember
-  CIDRRule cidrs = 2;
+  HTTPSRule https = 2;
 
-  // Matches every destination.
```
haiyanmeng, line 427: What is the reason to remove `all`?

`all` can be useful if users want to use their own ext_proc on top of Substrate's ext_proc. They could set the EgressPolicy for every actor to `all` and then rely on their own ext_proc to do authz.

LiorLieberman: I think it was used only in cidr rules. But we should add something like that back. 

first open question in the description is `how do we enable all SNIs -- e.g MITM all SNIs or inject header to all hosts?`

haiyanmeng:
> how do we enable all SNIs -- e.g MITM all SNIs or inject header to all hosts?

Do we need to support this for GA?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -399,40 +394,45 @@ message EgressPolicy {
...
+  // TLS to the named servers, forwarded without decryption.
   //
   // +k8s:optional
   // +k8s:unionMember
-  google.protobuf.Empty all = 3;
+  TLSPassthroughRule tls_passthrough = 3;
```
haiyanmeng, line 429: Should `TLSPassthroughRule` be evaluated before `HTTPRule` and `HTTPSRule`?

If a `HTTPSRule` for a given hostname appears before a `TLSPassthroughRule` for the same hostname (i.e., SNI == hostname), what would happen? Should MITM happen or not?

pkg/proto/ateapipb/ateapi.proto
haiyanmeng, line 427: What is the reason to remove the rule `all`? Removing `all` leaves no way to express allow-everything.

`all` can be useful if users want to use their own ext_proc on top of Substrate's ext_proc. They could set the EgressPolicy for every actor to `all` and then rely on their own ext_proc to do authz.

@LiorLieberman , somehow this comment keeps landing on the wrong line.

haiyanmeng: Oops. I left the comment on the wrong line. Fixed it with a new comment: https://github.com/agent-substrate/substrate/pull/1751/changes#r4051062192

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -454,35 +454,94 @@ message HostnameRule {
...
+  // +k8s:optional
+  // +k8s:maxItems=16
+  // +k8s:listType=set
+  // +k8s:eachVal=+k8s:minimum=1
+  // +k8s:eachVal=+k8s:maximum=65535
+  repeated int32 ports = 2;
 }
 
-// EgressRuleEffects contains effects applied by a matching hostname rule.
+// EgressRuleEffects contains effects applied by a matching HTTP rule.
```
haiyanmeng, line 536: `a matching HTTP rule` -> `a matching HTTP or HTTPS rule`

pkg/proto/ateapipb/ateapi.proto
haiyanmeng: > I think it was used only in cidr rules. 

`all` was not under the `cidr` rules before. It was directly under `EgressRule`.

But I agree that an `all` rule at the top does not make sense anymore since only HTTP(S) traffic is supported for GA.

haiyanmeng: > But we should add something like that back.
> 
> first open question in the description is `how do we enable all SNIs -- e.g MITM all SNIs or inject header to all hosts?`

Can we add `all` under HTTPRule, HTTPSRule, and TLSPassthroughRule?

EItanya:
> inject_static_headers and CredentialHeaderInjection still say "inject"
while the behavior is replacement. Should we rename to replace_headers now, while we
are already breaking the API?

https://github.com/agent-substrate/substrate/issues/1538 fwiw

pkg/proto/ateapipb/ateapi.proto
haiyanmeng: I wonder if it is better to move `TLSPassthrough` out of `EgressRule`, and make it directly under `EgressPolicy`. Here are two reasons:

1. `TLSPassthroughRule` and `HTTP(S)Rule` do not share an evaluation point. A `HTTP(S)Rule` rule is defined to be evaluated per HTTP request, at the request checkpoint, against the request hostname (Host/:authority). `TLSPassthroughRule` is decided once per connection, from the ClientHello, against the SNI — before any request exists. 

2. It makes it easier to get all the TLSPassthrough SNI patterns.  Keeping `TLSPassthroughRule` under `EgressRule` means we need to scan through all the EgressRules in an EgressPolicy and merge all the `TLSPassthroughRule`.

LiorLieberman:
I thought explicitly talked on this header modifier (on Monday and yesterday) among other things and decided we are going to go with the most simple thing for now (that other projects are also doing). #1538 approach was an alternative in the doc as well that we ruled out for GA fore cred injection

@thockin @bowei

EItanya:
> I thought explicitly talked on this header modifier (on Monday and yesterday) among other things and decided we are going to go with the most simple thing for now (that other projects are also doing). #1538 approach was an alternative in the doc as well that we ruled out for GA fore cred injection
> 
> Tim Hockin (@thockin) Bowei Du (@bowei)

My issue is that the API we're adding is incredibly specific. I'm fine with just adding secret/credential values in the header API we release, but having the whole API be called, secret_credential_inject means there's no room to move from there.

bowei:
> > I thought explicitly talked on this header modifier (on Monday and yesterday) among other things and decided we are going to go with the most simple thing for now (that other projects are also doing). #1538 approach was an alternative in the doc as well that we ruled out for GA fore cred injection
> > Tim Hockin (Tim Hockin (@thockin)) Bowei Du (Bowei Du (@bowei))
> 
> My issue is that the API we're adding is incredibly specific. I'm fine with just adding secret/credential values in the header API we release, but having the whole API be called, secret_credential_inject means there's no room to move from there.

+1 on this -- if we can name things a bit more generically, it will give us easier future evolution.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -399,40 +394,45 @@ message EgressPolicy {
   // +k8s:customValidation # name must be "default"
   ResourceMetadata metadata = 1;
 
-  // Rules are evaluated in order. The first matching rule authorizes the
-  // request, only that rule's effects are applied, and evaluation stops even if
-  // later rules would also match. A request is denied when no rule matches.
+  // Rules are evaluated in order. The first matching rule decides, only that
```
bowei, line 397: I think imposing this ordering is a mistake, especially if we are going towards a model where the we have a lot of mechanisms that are helping identify the intent of the actor, specifically, the intent to send to a hostname or IP address.

We should think of the egress flow as composed of:

1. _Identification of Actor intent_ (which hostname it wants to send to, what protocol etc). This may be obtained through {sniffing, DNS address correlation (see tun2connect), direct use of HTTP_PROXY forward proxy var). How this identification is done is not a security boundary. We should assume from the purposes of policy that this identification happens in a separate phase from the egress PEP performing the egress action of the traffic.
2. _Modification of requests_ e.g. header manipulation for cred injection.
3. _Egress perimeter_. This is an API that should be expressed in terms of destinations and not how the identification of actor intent is done.

This API should be targeted towards 3, assuming step 1 has already done its job and identified either the host/IP that the request/connection is sent to.

haiyanmeng: +1. 

My [comment](https://github.com/agent-substrate/substrate/pull/1751#discussion_r4051051868) shares the same concern.

Evaluting `HTTP(S)Rules` in order makes sense. However, evaluating `HTTP(S)Rules` and `TLSPassTHroughRules` in order does not.

LiorLieberman: I am not certain how the order is impacting by the classification/flow you describe here.

>Egress perimeter. This is an API that should not be expressed in terms of destinations and not how the identification of actor intent is done.

I am confused, if not destinations, what is it instead?

>This API should be targeted towards 3, 

About bullet num 2, we wanted credential injection to be part of the this API. Are you suggesting we something else?

LiorLieberman:
>My issue is that the API we're adding is incredibly specific. I'm fine with just adding secret/credential values in the header API we release, but having the whole API be called, secret_credential_inject means there's no room to move from there.

Got it, I think this makes sense. Maybe we rename to "replace_secret" or something that leaves room to expand if we need. I am trying to balance not including apis we dont need to now, vs locking us for a decision.

pkg/proto/ateapipb/ateapi.proto
howardjohn: Here is what I was thinking that avoids ordering concerns

```
rules:
  github.com:
    http: {} # Allow http
  google.com:
    tls: {} # Allow tls only
  gitlab.com:
    # For tls, MITM. Parse as http and inject
    tls: {
      inspect: true,
      http: {
        inject:
          x-foo: bar
      }
    }
  *.wikipedia.com:
    # Allow HTTP and inject
    http: {
      inject:
        x-foo: bar
    }
  1.2.3.4/8:
    # Allow anything as opaque tcp
    tcp: {}
```
Semantics:

1. Match the connect authority to hostaname or IP. It will only be an IP if they do not use DNS (i.e. if its curl [github.com](http://github.com/), and [github.com](http://github.com/) is 1.2.3.4, it would not match the cidr rule)
  a. There is no priority needed - its either an IP or a hostname, it cannot be both.
  b. Exact hostname > wildcard hostname. More precise CIDR > less precise cidr if there is overlap. Nothing fancy

2. We then decide the protocol for the inner traffic. You need to specify if it can be http/tls/tcp.
  a.  If I just put one protocol we don't need sniffing; anything not matching that protocol is rejected. 
    tcp: treat as opaque stream (i.e. if it happened to be HTTP traffic its fine but we can't see it, and obviously cannot inject headers etc)
     tls: treat as TLS, we read the SNI only but do not terminate unless inspect=true (We could also have tls-passthrough vs https-terminate which is probably a nicer API)
     http: treat as HTTP, we read each request

   b. If you have multiple, we sniff to find which path to take. Note that given you already allowed the explicit host/IP, a mis-detect is only going to fail to apply to specific rules like header insertion, so its less-fatal than other designs.
Also note: sniff(HTTP | TLS) is trivial. Sniffing is only hard when you allow tcp as a possibility. Because we can reliably detect "Not TLS" in 1 byte and treat it as HTTP, while sniff(HTTP | TCP) has server first issues, and "Tcp that looks like HTTP" issues


3. We can/should probably do layer sanitization to avoid confusion attacks on the upstream. That is, if I allow [github.com](http://github.com/) I should ensure the SNI and Host headers of all requests actually match those as well. it could be an explicit policy if we needed I suppose

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: I dont think we need it. Like, why would folks not store "Bearer" or any other prefix in their secret. 

I dont feel super strongly but opening it for a mini discussion.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -399,40 +394,60 @@ message EgressPolicy {
...
 
-  // Rules are evaluated in order. The first matching rule authorizes the
-  // request, only that rule's effects are applied, and evaluation stops even if
-  // later rules would also match. A request is denied when no rule matches.
+  // Rules are an unordered set of allows. Traffic is denied when no rule
+  // matches. When more than one rule matches the same traffic, precedence
+  // MUST be given by the following criteria, in order, moving to the next one
+  // on ties:
+  //
+  //   * A matching pattern without a wildcard.
```
bowei: matching pattern => host_pattern (be more specific).

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -399,40 +394,60 @@ message EgressPolicy {
...
 message EgressRule {
-  // Matches when the request hostname matches any configured pattern. Effects
-  // are applied once when the rule matches.
+  // Cleartext HTTP to the named hosts. Only HTTP: a TLS connection to a name
+  // that matches no other protocol is refused.
   //
   // +k8s:optional
   // +k8s:unionMember
-  HostnameRule hostnames = 1;
+  HTTPRule http = 1;
```
bowei: This excludes having both http://foo.com:8000 and https://foo.com:8000, but I think that's ok.

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: should we just do "*"?

EItanya: I agree with @LiorLieberman here, `all` was probably my least favorite part of the initial design, (mea culpa on my part).

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: We removed any ordering right now. 

>If a HTTPSRule for a given hostname appears before a TLSPassthroughRule for the same hostname (i.e., SNI == hostname), what would happen? Should MITM happen or not?

We have defined conflict resolution rules, (added commetns in the proto) so there should not be a case where there is conflicts. If there is a tie, then we reject or publish a warning.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -454,49 +469,118 @@ message HostnameRule {
...
+  // +k8s:customValidation # format
+  repeated string ports = 2;
 
   // Effects do not authorize traffic. They are applied only when this is the
-  // first matching rule.
+  // deciding rule.
   //
   // +k8s:optional
-  EgressRuleEffects effects = 2;
+  EgressRuleEffects effects = 3;
```
yufan-su: Are we planning to do credential injection on cleartext HTTP?

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: I think I agree with all of your sematics here. I updated the language on semantics to reflect that. 

I did not add language to bullet 3 yet though. I think we want that, and we discussed that in the EgressCon meetings. We could only do it for MITMed traffic. Let me add it.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -399,40 +394,60 @@ message EgressPolicy {
...
-  // request, only that rule's effects are applied, and evaluation stops even if
-  // later rules would also match. A request is denied when no rule matches.
+  // Rules are an unordered set of allows. Traffic is denied when no rule
+  // matches. When more than one rule matches the same traffic, precedence
+  // MUST be given by the following criteria, in order, moving to the next one
+  // on ties:
+  //
+  //   * A matching pattern without a wildcard.
+  //   * A matching port other than "*".
+  //
```
haiyanmeng: Here is an example the tie break does not work:

```
- https: {host_patterns: ["*.example.com"],     ports: ["443"], effects: {…token A…}}
- https: {host_patterns: ["*.api.example.com"], ports: ["443"]}                       
```
For an egress policy with the rules listed above, traffic to `v1.api.example.com` matches both. Both patterns are wildcards; both ports are 443. The admission check doesn't fire, because it only rejects the same pattern on the same port and these differ. 

cc @bowei  @LiorLieberman

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -399,40 +394,60 @@ message EgressPolicy {
...
+  // Two rules with the same pattern on the same port are rejected at
```
haiyanmeng: `Two rules` -> `For a given protocol, two rules`.

This makes sure that the admission rule does not reject an egress policy like this:
```
- http:  {host_patterns: ["x.example.com"], ports: ["*"]}
- https: {host_patterns: ["x.example.com"], ports: ["*"]}
```

pkg/proto/ateapipb/ateapi.proto
haiyanmeng: I am okay with either option. But we should make sure including one in the API.

pkg/proto/ateapipb/ateapi.proto
haiyanmeng: @LiorLieberman , please fix this.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -399,40 +394,60 @@ message EgressPolicy {
...
-  // Rules are evaluated in order. The first matching rule authorizes the
-  // request, only that rule's effects are applied, and evaluation stops even if
-  // later rules would also match. A request is denied when no rule matches.
+  // Rules are an unordered set of allows. Traffic is denied when no rule
+  // matches. When more than one rule matches the same traffic, precedence
+  // MUST be given by the following criteria, in order, moving to the next one
+  // on ties:
+  //
+  //   * A matching pattern without a wildcard.
+  //   * A matching port other than "*".
```
bowei: [nit] Does the character "*" mean "any port". If not, is it more clear to say "any port" instead.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -454,49 +469,118 @@ message HostnameRule {
...
+  // Ports of the destination the Actor connected to. A port in a request's
+  // authority is neither matched nor dialed. Each entry is a decimal port
+  // number from 1 to 65535, or "*" as the only entry for any port. Empty
+  // means ["80"].
+  //
+  // +k8s:optional
+  // +k8s:maxItems=16
+  // +k8s:listType=set
+  // +k8s:customValidation # format
+  repeated string ports = 2;
```
EItanya: Why is port a string and not a uint? Is it just to accept "*"?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -454,49 +469,118 @@ message HostnameRule {
...
+  repeated string host_patterns = 1;
+
+  // Ports of the destination the Actor connected to, in the format of
+  // HTTPRule.ports. Empty means ["443"].
+  //
+  // +k8s:optional
+  // +k8s:maxItems=16
+  // +k8s:listType=set
+  // +k8s:customValidation # format
+  repeated string ports = 2;
```
EItanya: Same question as above

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -399,40 +394,60 @@ message EgressPolicy {
...
   // +k8s:maxItems=256
-  // +k8s:listType=atomic # rule order matters
+  // +k8s:listType=atomic
+  // +k8s:customValidation # no two rules tie on a name and port
   repeated EgressRule rules = 2;
 }
 
-// EgressRule authorizes a request using exactly one destination matcher.
+// EgressRule allows traffic using exactly one protocol. The protocol fixes
+// what the gateway matches on and when: a request, or a TLS connection at its
```
bowei: suggest:

```
EgressRule allows traffic for a given protocol and destination (e.g. host name or IP, port).

The protocol will be identified by inspection of the egress traffic:

* Plaintext HTTP request (e.g. `GET /about HTTP/1.1\r\n`)
* TLS ClientHello
* HTTPS (TLS + HTTP)

Any traffic that cannot be identified to a specific protocol will be dropped. Protocols where the server sends first after connect are also not supported.

Users should expect additional protocols to be supported in the future.
```

pkg/proto/ateapipb/ateapi.proto
keithmattix: Idk, I've never seen anyone put "Bearer " in the secret since that's a header/network contract and not part of the actual secret value. If you're using the same secret in a client library, you won't want the bearer prefix. I vote we keep this in

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: `*.api.example.com` is more specific, therefore it wins

LiorLieberman: this kinda goes without saying, but probably its better to be more specific in the language beyond ` A matching pattern without a wildcard.`

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: Not for a given protocols. it is actually useful across protocols. Your example should be rejected.

Here is another example where we should reject across protocols.
```
- http:  {host: ["x.example.com"], ports: ["222"]}
- tlsPassthrough: {hosts: ["x.example.com"], ports: ["222"]}
```

Should be rejected.

pkg/proto/ateapipb/ateapi.proto
EItanya: I left a comment about this field, do we want to make port matching explicit always?

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: yes

EItanya: I'm not sure I agree here. Until now port has been default empty, allow all. Do we need to change the valence?

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: I am not sure about "The protocol will be identified" - this reads as "sniffing" to me. In most cases, we dont sniff, we try to apply whats in the policy and if it does not match we reject with an error.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -454,49 +469,118 @@ message HostnameRule {
   // +k8s:maxItems=256
   // +k8s:listType=set
   // +k8s:customValidation # format
-  repeated string patterns = 1;
+  repeated string host_patterns = 1;
+
+  // Ports of the destination the Actor connected to. A port in a request's
+  // authority is neither matched nor dialed. Each entry is a decimal port
+  // number from 1 to 65535, or "*" as the only entry for any port. Empty
+  // means ["80"].
```
haiyanmeng: `Empty means ["80"]`: does this mean PostgreSQL store `["80"]`? or PostgreSQL store empty but egress gateway should treat an empty `ports` field as `["80"]`?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -454,49 +469,118 @@ message HostnameRule {
...
   //
   // +k8s:required
   // +k8s:maxItems=256
   // +k8s:listType=set
   // +k8s:customValidation # format
-  repeated string cidrs = 1;
+  repeated string host_patterns = 1;
+
+  // Ports of the destination the Actor connected to, in the format of
+  // HTTPRule.ports. Empty means ["443"].
```
haiyanmeng: `Empty means ["443"]`: does this mean PostgreSQL store `["443"]`? or PostgreSQL store empty but egress gateway should treat an empty `ports` field as `["443"]`?

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: FWIW - what we "had" until now had no ports at all. So "empty allows all" was never a choice (I know its a choice in istio and other projects).

If empty means any, every https rule that means 443 is /silently/ open on all ports unless the author/harness remember to type it. If empty means 443, in the rare case when all ports are needed you type "*" and its a more explicit choice. 

"any port" is also more problematic with http. with https, if you open all ports, the server destination still need to have a certificate, so its harder for you to speak to something you did not intended to. With http you can reach every plain-HTTP service on the host, which is not necessarily intended. 

no port == any feels like a usabilty preference rather than a design choice and I would prefer sticking with the more secure one, esp given that harnesses are the ones that author it.

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: lets resolve this comment for now in favor of the above comment.

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: "*" does mean any port

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: discussed offline with john but for posterity:  updated the semantics to clarify and align with his proposal.
Keeping the list shape rather than a map keyed by destination, for two reasons:

1. The host_pattern as a key is a bit weird, and is harder to append validation, or expand. Kubernetes API conventions also rule out/discourage maps of subobject - (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps).

2. One key means one pattern. A rule that allows twenty hosts with the same credential becomes twenty entries, each repeating the effects. With host_patterns as a list it is one rule.

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: This is the last part I am no sure what to put in the api comments. I think we are all in agreement on the semantics, incl sniffing (although its an impl detail). But a lot of it depended on us having that named connect.

* protocol decisions (http, tls, tcp) are scoped per (hostname, port), not only per port. 
* https and tls_passthrough on the **same host and port** is rejected at admission.
* http and https (or http and tls) on **the same host and port** should be rejected at admission too. It avoids sniffing on all traffic to that host and port.
* Wildcard ports don't change the rule. The http plus https on ports "*" example falls under the reject-at-admission agreement.
* Open future question: http and tcp on the same **host and port** - allowed and requires sniffing OR not allowed - for the case where actor dialed an ip.

I am torn between (1) putting something on it (if we sniff and how) in the api comment plus a shoutout to how this is changes in the future (2) put something more light without anything about the future (3) dont say anything about how protocol is identified and leave it to be impl specific.

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: added host (space) pattern - that its also valid for sni_patteerns.

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: done

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: good point, I dont think we need to support that. leaving the comment open in case someone have different opinions

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: resolved. Feel free to re-open if you still have quesitons

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: AFAICT there is no way right not to store server side defaults. so postgress does not store it.

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: resolve this in favor of the other comment you have on the same subject.

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -454,49 +471,118 @@ message HostnameRule {
...
+  // be unique case-insensitively.
   //
   // +k8s:optional
   // +k8s:maxItems=16
   // +k8s:listType=map
   // +k8s:listMapKey=header
   // +k8s:customUnique # case-insensitive
   // +k8s:customValidation # for duplicate headers
-  repeated CredentialHeaderInjection inject_static_headers = 1;
+  repeated CredentialHeaderInjection replace_headers = 1;
```
haiyanmeng: The message name `CredentialHeaderInjection` seems incorrect now since all it does is to replace the value of a header.

pkg/proto/ateapipb/ateapi.proto
haiyanmeng: header replacement may not apply to cleartext HTTP. However, we may add other effects in the future applicable to both HTTPS and cleartext HTTP.

pkg/proto/ateapipb/ateapi.proto
haiyanmeng: If we don't have server side defaults today, it seems that we are forced to take the position that `PostgreSQL store empty but egress gateway should treat an empty field as ["80"]`. We should make this clear in the comment.

If Substrate supports server side defaults in the future, switching to it for this API field may be a breaking change.

pkg/proto/ateapipb/ateapi.proto
bowei: we can remove the specifics for now (the bullet list) and just say it can change to be more specific in the future.

bowei:
I am LGTM for this. Marking approve but others should mark approval before merging.

bowei (dismissed):

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -454,49 +471,118 @@ message HostnameRule {
...
+  // +k8s:listType=set
+  // +k8s:customValidation # format
+  repeated string ports = 2;
 }
 
-// EgressRuleEffects contains effects applied by a matching hostname rule.
+// EgressRuleEffects contains effects applied by a matching HTTP or HTTPS rule.
 //
 // +k8s:customValidation # for "at least one" and duplicate headers
 message EgressRuleEffects {
```
EItanya: Should we rename this `HttpRuleEffects`?

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: We may need to leave it

cc @EItanya

bowei: I think we leave it for now. It is a hint to the system that http from actor == http outbound, but it doesn't have to be that way. Question we should ask ourselves is whether or not this presents a lever that is too easy to mess up accidentally. It definitely feels like a potential open manhole.

LiorLieberman: I should add a follwowup to document the risks per https://cloud-native.slack.com/archives/C0B6M3E2J3D/p1790276208463309?thread_ts=1790272497.090439&cid=C0B6M3E2J3D

pkg/proto/ateapipb/ateapi.proto
LiorLieberman: renamed

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -463,49 +484,118 @@ message HostnameRule {
...
 }
 
-// EgressRuleEffects contains effects applied by a matching hostname rule.
+// HttpRuleEffects contains effects applied by a matching HTTP or HTTPS rule.
 //
 // +k8s:customValidation # for "at least one" and duplicate headers
-message EgressRuleEffects {
-  // Injects values retrieved from credential providers into request headers.
-  // Header names must be unique case-insensitively.
+message HttpRuleEffects {
```
haiyanmeng: Should this be `HTTPRuleEffects` to be consistent with other message names like `HTTPRule`?

haiyanmeng:
> > how do we enable all SNIs -- e.g MITM all SNIs or inject header to all hosts?
> 
> Do we need to support this for GA?

@LiorLieberman , @EItanya , to allow all egress http(s) traffic, the egress policy should be like this. Right?

```
- http:  {ports: ["*"]}
- https: {ports: ["*"]}
```

PR #1752, opened 2026-09-18
Separate PostgresQL runtime and schema ownership

### LINKED ISSUE #997: Define least-privilege role design for Postgres

iplay88keys:
Fixes #997

## Summary

Separate ateapi's normal database traffic from schema ownership and maintenance by supporting distinct runtime and DDL connections.

The runtime connection handles requests and watch polling. The optional DDL connection handles schema creation, Goose migrations, grants, and outbox partition maintenance. Omitting the DDL connection preserves single-connection deployments.

This also adds refreshable PostgreSQL connection sources so Kubernetes Secret rotation can take effect without restarting ateapi.

## What changed

- Added a dedicated owner/maintenance pool for:
  - schema creation
  - Goose migrations
  - runtime privilege grants
  - outbox partition creation, cleanup, and retention
- Kept request handling and watch polling on runtime connections.
- Limited the owner pool to two connections, sufficient for Goose's session lock and migration work.
- Granted the runtime role DML access to tables and sequences after migrations while excluding the migration ledger.
- Preserved single-connection operation by making the DDL connection default to the runtime connection source.
- Accepted either a literal PostgreSQL connection string or `@file:/absolute/path` for runtime and DDL connections.
- Used pgx `BeforeConnect` to re-read file-backed credentials and TLS material before every new physical connection.
- Rejected file updates that change the database identity—host, port, fallback hosts, database, or user—with a redacted restart-required error.
- Added `--postgres-max-conn-lifetime` to bound how long existing runtime, watch, and DDL connections can retain old credentials.

The bundled PostgreSQL deployment now:

- provisions separate `ateapi_runtime` and `ateapi_ddl` roles
- generates and persists independent passwords
- restricts TCP access to those roles and the `atepg` database
- requires both SCRAM authentication and a trusted client certificate
- assigns schema and object ownership to the DDL role
- disables SCRAM channel binding in generated bundled DSNs because pgx cannot derive channel-binding data from the Ed25519-signed service certificate; verified TLS, client certificates, and SCRAM remain enabled

Installer changes include:

- storing credential-bearing DSNs in a Secret
- keeping the schema name in the ConfigMap
- rejecting a DDL DSN without an explicit runtime DSN
- hashing the effective database configuration into the pod template so installer-managed credential or schema changes roll ateapi
- redacting DSNs from logs and errors
- applying and waiting for bundled PostgreSQL when deploying ateapi independently

Documentation recommends a dedicated Substrate schema for split-role deployments because runtime grants cover every table and sequence in that schema.

## Rotation behavior

For an `@file:` source, ateapi re-reads the file whenever pgx opens a new physical connection. Existing connections remain valid until they are retired, so operators should set `--postgres-max-conn-lifetime` to their desired rotation window and keep old and new credentials valid during that overlap.

Literal connection strings remain supported. Their password is static for the process lifetime, while referenced TLS files can still be refreshed for new connections.

## Security considerations

This prevents normal request paths from performing migrations or outbox partition DDL and provides database-level separation and auditing.

Both credentials still exist inside the serving ateapi process, so this is not a security boundary against complete process compromise. Moving migrations and partition maintenance to a separate process remains a possible future hardening step.

## Testing

- `go test ./cmd/ateapi/internal/store/atepg ./cmd/ateapi ./cmd/ate-setup/internal/config ./cmd/ate-setup/internal/steps`
- focused race tests for `cmd/ateapi` and `cmd/ateapi/internal/store/atepg`
- repository race-test phase
- GolangCI, shellcheck, boilerplate, and `bash -n` verification
- `git diff --check`

Coverage includes split-role permissions, single-connection operation, owner-pool routing, credential and TLS rotation, stable-identity validation, redaction, runtime/DDL source independence, watch-pool copying, connection-lifetime propagation, installer credential reuse, DSN validation, and configuration-hash rollouts.


cmd/ate-setup/internal/steps/apienv.go
```diff
@@ -48,40 +49,47 @@ func (e *Env) CreateAPIServerEnvVars(ctx context.Context) error {
...
+		ownerDSN = readWriteDSN
+	}
+	readWriteDSN = withPoolMaxConns(readWriteDSN, e.Cfg.PostgresPoolMaxConns, readWriteFromOperator)
+	log.Infof("POSTGRES_READ_WRITE_CONNECTION_STRING: %s", redactDSN(readWriteDSN))
+	log.Infof("POSTGRES_OWNER_CONNECTION_STRING: %s", redactDSN(ownerDSN))
 
-	if err := e.Kube.ApplyConfigMap(ctx, e.Namespace(), ConfigMapAPIEnvVars, cloudSQLEnvVars(cloudsql)); err != nil {
+	configVars := cloudSQLEnvVars(cloudsql)
+	configVars["ATE_API_POSTGRES_READ_WRITE_ROLE"] = e.Cfg.PostgresReadWriteRole
+	configVars["ATE_API_POSTGRES_OWNER_ROLE"] = e.Cfg.PostgresOwnerRole
```
EItanya: 1. **Preserve custom roles when adopting Cloud SQL configuration.**

If an installation uses `tenant_owner` / `tenant_readwrite` and is later redeployed from a shell without those environment variables, the installer adopts the recorded Cloud SQL endpoint and DSNs but overwrites the role names with `substrate_owner` / `substrate_readwrite`. The restarted API then fails at `SET ROLE` if those default roles were never provisioned.

I reproduced the overwrite with the Kubernetes fake client. Please preserve the recorded role names when no explicit override was supplied, using the defaults only when no recorded value exists.

cmd/ate-setup/internal/steps/apienv.go
```diff
@@ -48,40 +49,47 @@ func (e *Env) CreateAPIServerEnvVars(ctx context.Context) error {
...
+				return err
+			}
 		}
 	}
-	dsn = withPoolMaxConns(dsn, e.Cfg.PostgresPoolMaxConns, dsnFromOperator)
-	log.Infof("POSTGRES_CONNECTION_STRING: %s", redactDSN(dsn))
+	if ownerDSN == "" {
+		ownerDSN = readWriteDSN
+	}
+	readWriteDSN = withPoolMaxConns(readWriteDSN, e.Cfg.PostgresPoolMaxConns, readWriteFromOperator)
```
EItanya: 2. **Keep credential-file references intact when applying pool sizing.**

With `ATE_API_POSTGRES_READ_WRITE_CONNECTION_STRING=@file:/run/postgres/readwrite-dsn` and `ATE_API_POSTGRES_POOL_MAX_CONNS=20`, this produces `@file:/run/postgres/readwrite-dsn pool_max_conns=20`. I reproduced that output; the new file reader treats the entire suffix as the filename, so the API cannot read the credential file and startup fails.

Please avoid appending DSN parameters to `@file:` references. Pool sizing can live in the referenced DSN or be applied separately to the parsed pool configuration.

cmd/ateapi/internal/store/atepg/bootstrap.go
```diff
@@ -0,0 +1,139 @@
...
+
+	source, err := newConnectionStringSource(cfg.EndpointSource)
+	if err != nil {
+		return err
+	}
+	dsn, err := source()
+	if err != nil {
+		return err
+	}
+	connConfig, err := pgx.ParseConfig(dsn)
```
EItanya: 3. **Strip pool-only settings before opening the bootstrap connection.**

With bootstrap enabled and `pool_max_conns=4` in the owner DSN, `pgx.ParseConfig` leaves the option in the PostgreSQL startup parameters. I reproduced the resulting failure against PostgreSQL 18: `FATAL: unrecognized configuration parameter "pool_max_conns" (SQLSTATE 42704)`. Startup stops before identity creation or migrations, although this is a valid DSN for the normal application pools.

Please parse with `pgxpool.ParseConfig` and use its `.ConnConfig` for the direct administrator connection, so pool-only options are consumed locally.

EItanya (commented):
🤖 AI-generated review.

I found three reproducible configuration issues, detailed inline. The five affected package suites passed, and an additional PostgreSQL 18 probe passed the actual OpenFGA migrations and initialization with the restricted roles. Cluster E2E was not run locally.

iplay88keys:
I have local changes that add a second Cloud SQL Auth Proxy sidecar to `ate-api-server` for the owner GSA, which I have also tested manually. Under the current Workload Identity setup, identity is configured at the pod level: the pod’s Kubernetes service account maps to the runtime GSA. The owner proxy therefore uses that runtime identity to impersonate the owner GSA, which is a documented approach [here](https://github.com/GoogleCloudPlatform/cloud-sql-proxy#configuring-service-account-impersonation).

This provides separate PostgreSQL logins for runtime operations and owner operations, although it is not a hard security boundary still because the runtime identity retains permission to impersonate the owner GSA. I also considered the other way around and having the pod's KSA be the owner, which would impersonate the runtime GSA, but that would make the more privileged owner GSA the pod's ambient identity. The least-privileged option there was to have the runtime KSA.

Achieving stronger separation would require moving owner operations out of `ate-api-server`: migrations would need to run through a separate Job or migrate command and outbox partition maintenance (which currently uses the owner connection) would also need a separately identified workload or a different privilege model.

@juli4n @EItanya @jpbetz, I’m happy to include the two-proxy approach in this PR, but I think the broader separation should be handled in a follow-up PR with an agreed-upon design and scope.

PR #1753, opened 2026-09-18
Add executor runtime lease reclamation

brayzn:
## Summary

- add a durable actor runtime lease with token/generation fencing and heartbeat renewal
- reclaim expired executor-owned actors natively before releasing their workers
- preserve the router's read-only ResumeActor fast path and use an explicit executor lease claim
- keep golden actor reconciliation lease-aware and preserve durable snapshots on crash

## Recovery contract

The executor renews every 30 seconds. A 120-second lease expires without heartbeats; the ate-api reaper claims the expired row, terminates the live workload, releases the worker, marks the actor CRASHED, and removes the lease. A running command is not resumed after executor death; its durable workspace snapshot remains the recovery input.

## Validation

- go test ./cmd/ateapi
- go test -count=1 ./cmd/ateapi/internal/controlapi ./cmd/ateapi/internal/store/atepg ./cmd/atenet/internal/router/ingress
- go test ./... (all relevant packages pass; existing non-TTY color tests fail in internal/e2e)


juli4n:
Thanks for the contribution. This is a large change, which introduces a new API surface, changes to the schema, etc. and as such needs careful design and review.

I would recommend you first file a bug describing the issue you are trying to solve, include repro steps so we can discuss and align on the solution. There are multiple work streams a people working on actor and worker lifecycle, and we need to make sure everyone get on the page.

I'll close this PR for now. Please file a bug so we can discuss it.

PR #1754, opened 2026-09-18
CI: Fail on missing preconditions and exceeded boundaries

### LINKED ISSUE #1747: Testing preconditions are not a hard-stop for some test causing low-fidelity signals in the CI/testing infra.
Testing preconditions, such as docker missing, can cause some tests to skip or silently pass. Others leave resources hanging waiting on something that will never happen but the root cause of the issue is unclear.

When a test that requires a precondition runs without said precondition being met, the test should fail when running in CI and produce a concise summary pinpointing the root cause.

[area/tests][area/dev-infra]

stshrive:
This change addresses six ways CI pipeline could report false success or hang on broken infrastructure. Importantly, this change does two things to reduce load on infrastructure:

1. It prevents jobs from running for [the default action limit of 360m](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idtimeout-minutes) by pinning the timeout to `45m`.
2. It limits pull_request jobs from continuing execution once superseded by new changes.

- **Silent container test skipping**: Tests now explicitly fail if `CI` or `REQUIRE_DOCKER` is set (while still skipping on local machines lacking Docker), with the check implemented in `dockerenv` to avoid an import cycle between `storetest` and `atepg`.
- **Unbounded trust bundle wait**: Enforced a shared 120-second timeout across both bundles (overridable via `ATE_INSTALL_TRUST_BUNDLE_TIMEOUT`) and added diagnostic dumping of bundles, controller pods, and logs before returning a non-zero exit code.
- **Missing sandbox preflight validation**: Added early preflight checks for `/dev/kvm` and `SandboxConfig/microvm` that fail fast and print actionable remediation instructions.
- **Skipped migration checks on main**: Configured the migration immutability check to run on pushes to main to catch modified migrations at the point of merge.
- **Pre-merge vulnerability scanning**: Configured `govulncheck` to run on pull requests to catch vulnerable dependencies prior to merging into main.
- **Missing job timeouts and concurrency limits**: Defined explicit timeout-minutes (45m and 120m bounds) and added concurrency groups that automatically cancel superseded pull request runs without canceling runs on main.

Fixes #1747

- [x] Tests pass
  - Silent container skipping, unbounded trust bundles, and missing sandbox were all forced locally and confirmed to exist with changes here resolving each.
  - The latter half of the scenarios exist in CI only due to being GH Action trigger issues.


BenTheElder:
>Post-merge vulnerability scanning: Configured govulncheck to run on pull requests to catch vulnerable dependencies prior to merging into main.

We intentionally did not do this. Vulns are point-in-time X sources, not just sources. 
We don't want to stop merging other PRs when there's a vuln in main's deps.

stshrive:
> We intentionally did not do this. Vulns are point-in-time X sources, not just sources.
> We don't want to stop merging other PRs when there's a vuln in main's deps.

There are two vectors of getting a vuln though.

The first is a new dependency that is _not_ in main and has not been scanned yet gets added in a PR without being scanned.

The second is a new point-in-time vuln being discovered in a dependency already in main.

I think it is worth manually overriding the latter to prevent the former from being inadvertently added to main.

stshrive:
I removed the update to the vuln check execution pattern. This should remedy any concerns about new point-in-time vulns in main blocking PRs.

annapendleton (approved):

cmd/ateapi/internal/store/dockerenv/dockerenv.go
```diff
@@ -60,3 +60,11 @@ func Configure(ctx context.Context) error {
 	}
 	return nil
 }
+
+// Required reports whether an unavailable container must fail the test rather
+// than skip it. Skipping is right on a workstation without Docker and wrong in
```
BenTheElder, line 65: hmm, I wonder if we should just require it ... something to consider in a follow-up, maybe add a TODO?

detecting if docker is available is also a fragile concept vs just spawning the container (let alone getting into topics like podman support)

skipping locally can also give a false sense of security when these packages are tested, and we pretty much require docker to develop substrate anyhow

internal/e2e/preflight.go
```diff
@@ -61,5 +63,30 @@ func PreflightChecks() error {
 		return fmt.Errorf("ListActors RPC failed: %v", err)
 	}
 
+	// The micro-VM class needs a SandboxConfig and a node advertising /dev/kvm.
+	// Without either, worker pods stay Pending and every suite fails on its own
+	// timeout minutes later, naming neither cause.
+	if IsMicroVM() {
+		if _, err := clients.SubstrateK8s.ApiV1alpha1().SandboxConfigs().Get(ctx, SandboxClassMicroVM, metav1.GetOptions{}); err != nil {
+			return fmt.Errorf("E2E_SANDBOX_CLASS=%s but SandboxConfig/%s is missing (apply manifests/microvm/sandboxconfig-microvm.yaml.tmpl): %w",
```
BenTheElder: we should probably give a more useful instruction here, you can't apply the tmpl file directly.

internal/e2e/preflight.go
```diff
@@ -61,5 +63,30 @@ func PreflightChecks() error {
...
+				SandboxClassMicroVM, SandboxClassMicroVM, err)
+		}
+
+		nodes, err := clients.K8s.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
BenTheElder: this won't work in large clusters, the results will be paginated, and it's possible the first page has none of the uVM nodes. it probably makes more sense IMHO to just let these things be runtime test failures, it only takes marginally longer to start real tests and fail when we can't schedule the pod / actor, and this knowledge of KVM is going to get out of sync with the real implementation.

consider eg in the future we might have /dev/mshv on azure

internal/e2e/preflight.go
BenTheElder: @zoez7 this will need updating with sandboxconfig redesign.

BenTheElder: We should also start including the uVM manifest by default when we no longer need to build binaries on the fly, that is https://github.com/agent-substrate/substrate/pull/1708

internal/e2e/preflight.go
stshrive: I don't think this will paginate based on the references for `ListOptions` which requires the `limit` value be set.

- https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.37/#listoptions-v1-meta
- https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks

Anyway, your point about `/dev/mshv` stands and I will push the specific device failure modes to runtime as suggested. Doing this removes the specific list call as well.

cmd/ateapi/internal/store/dockerenv/dockerenv.go
stshrive: I have added a TODO for now.

internal/e2e/preflight.go
stshrive: I have updated the recommendation in the message. I used a reference to the README which should survive a rename/redesign to the underlying system for uVM setup. Once the manifest is included by default, this check should just be a no-op.

internal/e2e/preflight.go
stshrive: I've pushed the error to runtime detection. IMO there is still a missing element which is to get the diagnostics out upon failure. I added another TODO for that but it is in a separate file, `resume.go`

maxsmythe (approved):
LGTM

Issue #1740, opened 2026-09-18
Avoid forwarding the client cert from Envoy to ext_proc through a header


haiyanmeng:
Currently, Envoy forwards the client certificate to ext_proc through the `x-forwarded-client-cert` header. 

- `atunnel` sends a bare `CONNECT` request to the egress gateway. The request has `Method`, `URL`, and `Host` and no `Header` map at all.
- The egress filter chain sets `forward_client_cert_details: SANITIZE_SET`, so whatever a client sends under this name is discarded and replaced by Envoy's own value.

We should avoid forwarding the client cert from Envoy to ext_proc through a header. The separation of `trusted data from Envoy` from `potential user-supplied data` makes it easier to prove the system is secure. 

cc @ahmedtd  @bowei

swiftdiaries:
I'll have some time today after the egress call to work on this. 
Not sure if I can do /assign @swiftdiaries ?

haiyanmeng:
Thanks for picking this up, @swiftdiaries . I assigned it to you.

PR #1758, opened 2026-09-19
Preserve peer certificate chains for egress authentication

swiftdiaries:
## Context
Addresses the egress certificate authentication issue in #1740.

## Changes
Pass the peer certificate chain through private Envoy metadata and use it for actor authentication.

```mermaid
flowchart LR
  subgraph Before
    A1["Actor / atunnel mTLS CONNECT"] --> E1["Envoy validates peer certificate"]
    E1 --> X["SANITIZE_SET writes Envoy-produced XFCC"]
    X --> P1["ext_proc / atenet parses XFCC"]
    P1 --> V1["Go verifies chain and extracts ActorIdentity"]
  end
  subgraph After
    A2["Actor / atunnel mTLS CONNECT"] --> E2["Envoy validates peer certificate"]
    E2 --> L["Lua reads downstreamSslConnection full chain"]
    L --> M["Private stream metadata: dev.ate.egress.peer_certificate"]
    M --> P2["ext_proc metadata_context"]
    P2 --> V2["Go decodes, verifies chain, extracts ActorIdentity"]
  end
  V1 --> G["GetActor UID and RUNNING check"]
  V2 --> G
  G --> Q["Egress policy"]
```

The previous XFCC value was Envoy-produced and sanitized. Private metadata removes the certificate from ordinary-header visibility and propagation, and removes XFCC parsing from atenet.

## Validation
`make verify` passed locally. VM non-root unit-race and apitool-race, root-race, and Envoy E2E on gVisor and microVM passed, including auth, leaf-only mutation/restoration, and MITM. AgentGateway E2E remains unverified.


cmd/atenet/internal/router/egress/egress.go
```diff
@@ -321,22 +309,26 @@ func (h *Handler) validateActor(ctx context.Context, identity *substratex509.Act
 // on the request into a verified ActorIdentity, or an error describing why it
 // cannot be trusted.
 func (h *Handler) authenticateActorCertificate(md *extproc.RequestMetadata) (*substratex509.ActorIdentity, error) {
+	if peer, present := md.DynamicMetadata[envoyClientCertificateMetadataNamespace]; present {
```
swiftdiaries: Review note:
With the changes in this PR: Envoy sends the full peer certificate chain through private dynamic metadata. AgentGateway sends one peer certificate in `source.certificate`. Do we want to make this explicit and not implicit?

And I wanted to check if the intended flow, decision ladder works:
- If the Envoy metadata namespace is present, `atenet` must use that certificate chain.
- If the chain is empty or invalid, `atenet` must deny the CONNECT request.
- `atenet` must not fall back to `source.certificate` after it finds the Envoy namespace.

cc @ahmedtd @bowei @keithmattix (for the agentgateway parts)

PR #1759, opened 2026-09-19
glutton: replace stuck actors on resume/hibernate failure

chw120:
## Summary
Replace broken/stuck `gluttonActor` instances in place when `resume()` or `hibernate()` fails during benchmark iterations.

## Problem
When `SuspendActor` fails (e.g., due to transient GCS `ResourceExhausted` errors during cold-bucket ramp-up), `ate-api-server` can leave the actor permanently stuck in `ACTOR_STATE_SUSPENDING`. 

Previously, `gluttonUser` kept the broken `gluttonActor` in its `u.actors` slice for the remainder of the benchmark run:
- `hibernate()` ignored `SuspendActor` / `PauseActor` errors.
- Subsequent `ResumeActor` calls against that slot repeatedly failed with `got: ACTOR_STATE_SUSPENDING` for the rest of the test, reducing effective concurrency and generating continuous failure noise.

## Solution
- Have `hibernate()`, `pause()`, and `suspend()` return `bool` indicating whether the operation succeeded.
- Add `(*gluttonUser).replaceActor(ctx, broken)` to delete the broken actor (`AnyState: true`) and replace its slot in `u.actors` with a freshly created actor (`sb-<uuid>`).
- Invoke `user.replaceActor(ctx, actor)` in `iterate()` when either `actor.resume(ctx)` or `actor.hibernate(ctx)` fails.
- Add unit test coverage in `lifecycle_test.go` (`TestGluttonIterate_ReplacesActorOnResumeFailure`).

- [x] Tests pass


internal/benchmarking/boomer/glutton/lifecycle.go
```diff
@@ -408,32 +435,33 @@ func isConcurrentUpdateConflict(err error) bool {
...
-	_ = u.tracedCall(ctx, "PauseActor", func(callCtx context.Context, tr *metadata.MD) error {
+func (u *gluttonActor) pause(ctx context.Context) bool {
+	err := u.tracedCall(ctx, "PauseActor", func(callCtx context.Context, tr *metadata.MD) error {
 		_, err := u.cfg.APIStub.PauseActor(callCtx, &ateapipb.PauseActorRequest{
 			Actor: u.ref(),
 		}, grpc.Trailer(tr))
 		return err
 	})
 	u.actorRunning = false
+	return err == nil
```
maxsmythe, line 453: Do we know that all errors are non-recoverable?

maxsmythe (commented):
One comment about whether we need more complex error classification, otherwise LGTM

internal/benchmarking/boomer/glutton/lifecycle.go
chw120: Good question. Nope. Updated the PR to classifiy on — replaceNow for NotFound/DataLoss/FailedPrecondition/crashed, retryLater for the cluster-wide and misconfigured-run codes, and anything unrecognized counts toward a 3-strikes replacement. WDYT?

maxsmythe (approved):
LGTM

internal/benchmarking/boomer/glutton/lifecycle.go
maxsmythe: I like it! This is the kind of thing that I think we should probably turn into a library to share across tests.

PR #1760, opened 2026-09-19
benchmarking: spread actors over several worker pools

chw120:
### Motivation
Large-scale benchmarks that span multiple node pools / machine families (e.g. `n4`, `c4`, `c3`) require two layers of placement constraints:
1. **Kubernetes layer (`spec.template.nodeSelector`)**: each `WorkerPool`'s worker pods must be scheduled onto the matching node pool.
2. **Actor placement layer (`Actor.worker_selector`)**: a suspended micro-VM memory snapshot captures the CPU features observed by the guest and cannot be resumed on an incompatible CPU model. While the Substrate scheduler already supports per-actor `Actor.worker_selector`, the benchmark harness previously only deployed a single `WorkerPool` and left `Actor.worker_selector` unset on every actor.

This PR updates both sides of the benchmark harness:
- **Workload deployment (`deploy.sh`)**: creates one `WorkerPool` per entry with `metadata.labels.pool=<name>` and optional `spec.template.nodeSelector`.
- **Load generator (`boomer`)**: assigns each actor a pool at creation (weighted by pool capacity) and populates `Actor.worker_selector: {pool: <name>}` so the actor stays within that pool across its entire `create` → `suspend` → `resume` lifecycle.

### Changes

- **`internal/benchmarking/boomer/userclass/pools.go`**: Add `PoolPicker` to parse `name:weight[,...]` specs, draw a pool weighted by capacity (`Pick()`), and build `Actor.worker_selector` (`SelectorFor()` matching `pool=<name>`). All methods are nil-safe so single-pool runs incur no branching.
- **`internal/benchmarking/boomer/glutton/{lifecycle,durdir}.go` & `cmd/benchmarking/boomer-worker/main.go`**: Add `--worker-pools`, draw one pool per actor when minted in `startUser()`, and attach `Actor.worker_selector` on `CreateActor`.
- **`benchmarking/workloads/deploy.sh` & `manifests/workerpool.yaml.tmpl`**: Accept `--worker-pools name:weight[:nodeSelectorKey=value]`, split `--worker-count` across pools proportionally by weight, and render one `WorkerPool` per entry with `metadata.labels.pool=<name>` and optional `spec.template.nodeSelector`.
- **`benchmarking/deploy_locust.sh`, `benchmarking/locust/*`, `hack/install-ate.sh`**: Forward `--worker-pools` to both the workload deployer (full spec with `nodeSelector`) and the boomer workers (`name:weight` list) from a single flag so the two halves cannot drift.

### Backwards Compatibility

When `--worker-pools` is omitted (default):
- `PoolPicker` is `nil` and `Actor.worker_selector` remains unset (`nil`).
- `workloads/deploy.sh` deploys a single `WorkerPool` named `benchmark-ateom` with `replicas: ${WORKER_COUNT}` and no `nodeSelector`.
- All existing single-pool benchmarks and CI jobs behave identically to before.

### Testing

- Unit tests for pool parsing, validation, nil safety, and weighted distribution (`internal/benchmarking/boomer/userclass/pools_test.go`).
- Wiring tests asserting the `CreateActorRequest` protobuf payloads for `glutton` and `durdir` (`internal/benchmarking/boomer/glutton/pools_wiring_test.go`).
- Validated rendered single-pool and multi-pool manifests against the `WorkerPool` CRD with `kubectl apply --validate=strict`.

- [x] Tests pass


benchmarking/workloads/deploy.sh
```diff
@@ -30,19 +30,27 @@ if [[ -z "${BUCKET_NAME:-}" ]]; then
...
+    exit 1
+  fi
+done
 
 WORKER_COUNT=1
+# Worker pools to create, as name:weight[:nodeSelectorKey=value] entries. Empty
+# means the single pool named benchmark-ateom that this script has always
+# created. Use the same weights the boomer workers run with, so each pool gets
+# the actors its workers can hold.
+WORKER_POOLS=""
```
supahero1, line 53: Can this be `WORKER_POOLS="${WORKER_POOLS:-}"`? Alternatively pass `--worker-pools` in `orchestrator.py` to `deploy_workloads()` and `teardown_workloads()`

benchmarking/workloads/manifests/workerpool.yaml.tmpl
```diff
@@ -0,0 +1,42 @@
...
+# Keeping an actor in one class is a correctness requirement, not tuning, once
+# the classes differ in machine type: a micro-VM snapshot carries the CPU
+# features the guest observed, and nothing masks them to a common baseline on
+# resume, so an actor that moves between CPU models fails to restore.
+apiVersion: ate.dev/v1alpha1
+kind: WorkerPool
+metadata:
+  name: ${POOL_WORKERPOOL_NAME}
+  namespace: benchmark-workloads
+  labels:
```
maxsmythe: Per comments, this is for sharding the workers across several different types of nodes... what constrains a specific worker pool to a given node type?

benchmarking/workloads/deploy.sh
```diff
@@ -136,6 +152,126 @@ substitute() {
...
+  for weight in "${POOL_WEIGHTS[@]}"; do
+    total=$((total + weight))
+  done
+
+  # Round each pool but the last to nearest and give the last the remainder, so
+  # the counts sum to WORKER_COUNT exactly. A share that rounds to zero still
+  # gets one worker: a pool with none can never serve the actors pinned to it.
+  local i assigned=0 count
+  for (( i = 0; i < ${#POOL_NAMES[@]} - 1; i++ )); do
+    count=$(( (WORKER_COUNT * POOL_WEIGHTS[i] + total / 2) / total ))
```
maxsmythe: Shouldn't this be `WORKER_COUNT * POOL_WEIGHTS[i] / total`? Not sure where `total / 2` is coming from?

benchmarking/README.md
```diff
@@ -26,10 +26,48 @@ image, then deploys the Locust workers:
...
+That run puts 50 workers on `n4d` nodes and 50 on `c4`, and sends half the
+actors to each.
+
+Pinning is a correctness requirement once the pools differ in machine type, not
+a tuning knob. A suspended actor's memory snapshot records the CPU features the
+guest saw, and nothing masks them to a common baseline on resume, so an actor
+that moves between CPU models fails to restore. Pinning is also how a run
+measures one machine type against another in the same test.
+
+A pool name is a class of interchangeable workers, not one `WorkerPool`: it
```
maxsmythe, line 56: nit: not one `worker`... a pool name is a WorkerPool (which is a collection of workers)

benchmarking/workloads/deploy.sh
maxsmythe, line 229: Rounding alters the ratios, which means that the actor distribution will be weighted differently from the worker distribution. Instead of doing math here, maybe work with absolute worker counts and feed those into the load generator to calculate distribution?

As a bonus, that means no dealing with rounding.

benchmarking/workloads/deploy.sh
maxsmythe: +1, orchestrator.py should know the composition of clusters it's deploying to.

hack/install-ate.sh
```diff
@@ -150,7 +150,10 @@ function usage() {
   echo ""
   echo "  --deploy-benchmarks                    Deploy workloads + locust load test stack"
   echo "  --delete-benchmarks                    Delete the locust stack and workloads"
-  echo "  --benchmark-worker-count N             Number of WorkerPool replicas (default: 1)"
```
maxsmythe, line 153: I think this script might be (or may have already) gone away in favor of golang installer?

maxsmythe (changes requested):
Very cool. Biggest question is how pool sizes should be specified and also wiring the data into orchestration.

PR #1771, opened 2026-09-21
(chore): Write both copies of an actor lifecycle event from one call

### LINKED ISSUE #1744: Build a lifecycle event once, then consume it in stdout and OTLP
## Problem

A lifecycle event is emitted twice today, once to stdout and once over OTLP:

```go
slog.LogAttrs(ctx, slog.LevelInfo, actorevent.StateChangedBody, attrs...)
actorevent.Emit(ctx, actorevent.StateChanged, attrs)
```

The attrs slice is shared and the body comes from a shared constant, so those can't drift. Two things still can:

- Severity. `slog.LevelInfo` lives at the call site, `log.SeverityInfo` lives on the `Event`. 
- The pairing. Nothing makes you call `Emit` at all. Add a third record with only `slog.LogAttrs` and it has no OTLP copy, and no test fails.

Proposal

Have `actorevent` own one emission helper that takes the event and the attrs and produces both outputs, with the slog level derived from Event.Severity rather than stored twice. One call per record, nothing left to keep in sync by hand.

From @JeffLuoo's review on #1658.

krisztianfekete:
An actor lifecycle record went to stdout and to OTLP in two separate calls. The attributes were shared, but the severity was not, so the slog level sat at the call site and the OTel severity sat on the `Event`. Nothing made a caller write both copies either, so a new record could reach stdout only and no test would notice.

Since this PR `actorevent.Log` now writes both copies. The level comes from `Event.Severity`, so it is stored once. The OTLP only `Emit` and the exported body constants are gone, so the dual write is the only way out of the package.

Also normalizes `ate.actor.operation.name` on the crash event, which the state change event already did.

Fixes #1744 

Testing

 - Unit tests for the level mapping, the dual write, and the registry check.
 - End to end on a fresh kind cluster. Both event names arrive with the right severity (9 and 17), the right attributes, and trace context on the record fields. The stdout copies match record for record.

- [x] Tests pass


internal/actorevent/actorevent.go
```diff
@@ -142,9 +157,19 @@ func NewEmitter(lp log.LoggerProvider) *Emitter {
...
-func (e *Emitter) Emit(ctx context.Context, ev Event, attrs []slog.Attr) {
+// Log writes both copies of ev from one call. Both read attrs; neither changes it.
+//
+// stdout goes first and unconditionally, because it is the copy that survives an
+// ungraceful exit. The two are not gated alike though: --log-level=warn silences
+// the stdout copy of an info event while the OTLP copy still ships.
+func (e *Emitter) Log(ctx context.Context, ev Event, attrs []slog.Attr) {
+	slog.LogAttrs(ctx, ev.Level(), ev.Body, attrs...)
+	e.emit(ctx, ev, attrs)
+}
```
JeffLuoo, line 177: The two copies still take two **different timestamps**. slog.LogAttrs stamps the stdout record with its own time.Now(), and emit calls time.Now() again at BuildRecord.

The package comment now says that nothing about a record is kept in step by hand, but a consumer that joins the stdout copy to the OTLP copy on an exact timestamp finds no match, because the two differ by microseconds.

Let's also align on the timestamps.

JeffLuoo (changes requested):

internal/actorevent/actorevent.go
krisztianfekete: This is now done.

JeffLuoo (approved):

PR #1772, opened 2026-09-21
Add DV for TagStatus + drop unused TagStatus.source_actor_uid field

laoj2:
TagStatus.source_actor_uid is unused, we can add this back if we ever need it. 

We were also missing some of the DV for TagStatus fields:

* Added maxLength=2048 requirement for ExternalSnapshot.snapshot_uri (matching in_progress_snapshot_uri)
* Added validation for actor_template_uid
* Added validation for storage_location, matching what we have in SnapshotsConfig.storage_location


pkg/proto/ateapipb/ateapi.proto
```diff
@@ -550,7 +550,6 @@ message ActorStatus {
   // snapshot the Actor is currently taking.
   //
   // +k8s:optional
-  // +k8s:maxLength=2048
```
juli4n, line 553: All string fields should have either a maxLength or a format so we can have a well-known size limit.

laoj2: SG. Added maxLength back here and to ExternalSnapshot.snapshot_uri fields

juli4n (dismissed):

laoj2:
Rebased

juli4n (approved):

Issue #483, opened 2026-07-21
Allow actors to self-suspend


juli4n:
Substrate's value comes from suspending idle actors and multiplexing many of them onto few workers. But today nothing decides when an actor is idle: suspend is only triggered by an external SuspendActor/PauseActor API call, presumably a higher order system.

We discussed an automatic idle-detection, but doesn't exist yet and I'm not sure if it can realistically work for session-based protocols such as websockets or grpc (with stream).

For some use cases, actors reliably know when they are idle, "I finished my turn / my streams are dormant / checkpoint me."

We should consider a cooperative model where actors *yield* by writing to a well-known file (e.g. `/var/ate/state/suspend`).

EItanya:
This feels like a dupe of https://github.com/agent-substrate/substrate/issues/257?

juli4n:
Yes, both seem to propose the same feature, they differ on the mechanism to accomplish it, but better to merge and have a single discussion.

ronlv10:
+1 for this feature.

I'd like to add that calling `SuspendActor` doesn't reliably work right now. When the actor calls the ate-api, the actor gets suspended, and upon resuming, an HTTP error occurs because the (previous) connection dropped (I talked about this issue [here](https://github.com/agent-substrate/substrate/issues/465#issuecomment-5027144693)). This leaves no good way to verify if the suspension actually succeeded. I think writing to a file instead would make the process more atomic and robust.

sfunkenhauser:
# Actor idleness probe

I propose adding an idleness probe, instead of a self-suspend action.

## Why not self-suspend

The obvious way to let an actor give its worker slot back is to let it ask: the actor tells ateom it has nothing left to do, and ateom proxies the suspend. The practical problem is that there is no actor→host channel today, and adding one is not cheap: HTTP means configuring the actor's egress and pinning a static endpoint IP:port into the sandbox, vsock exists only on some runtimes and would need a UDS on gVisor, and either way the actor carries client code and a protocol it did not previously need. That makes the actor non-portable — the workload has to know it is running on Substrate  (and potentially which sandbox class).

The deeper problem is that self-suspend asserts a transition rather than a fact, and a transition is a lifecycle the actor then has to own and manage: stop serving, request the suspend, wait to be woken, resume serving — along with whatever edge cases fall out of the request being asynchronous, since the actor goes on running for an indeterminate time after asking.

[...]

PR #1779, opened 2026-09-21
Add plumbing to allow ateom to suspend actors

sfunkenhauser:
Part of #483 

- [ x ] Tests pass


cmd/ateapi/internal/workerservice/suspend.go
```diff
@@ -0,0 +1,157 @@
...
+		return nil, err
+	}
+	// Workers are global-scoped, so the reference carries no atespace.
+	if errs := resources.ValidateGlobalObjectRef(req.GetWorker(), field.NewPath("worker")); len(errs) > 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid worker: %v", errs.ToAggregate())
+	}
+	if errs := validateActorRef(req.GetActor(), field.NewPath("actor")); len(errs) > 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid actor: %v", errs.ToAggregate())
+	}
+	if req.GetActorUid() == "" {
```
juli4n: Should we use the generated DV validations instead? Can we follow the same pattern we follow in the control API of having a validate<RequestMessage> method that calls the generated DV code and call it from here?

pkg/proto/ateapipb/ateapi.proto
```diff
@@ -2016,6 +2016,24 @@ service WorkerService {
...
+  //
+  // This is a request, not a command. It runs the same suspend workflow
+  // Control.SuspendActor runs, behind the same preconditions, so a proposal
+  // that races a resume, pause, or delete loses to it.
+  //
+  // atelet calls this with its own client certificate, as it does for
+  // SetWorkerCapacity, naming the Worker its caller proved itself to be. The
+  // control plane serves it only for an Actor that is assigned to that Worker.
+  //
+  // Idempotent in the same sense SuspendActor is: the workflow fast-forwards
```
juli4n: Is this true? It looks like if the actor is suspended, the method returns NOT_FOUND.

cmd/ateapi/internal/workerservice/suspend.go
sfunkenhauser: I don't want `workerserve` to depend on `controlapi` which is where the generated DV code currently lives. We should probably move the generated DV code to a neutral `apivalidation` package which can be imported into both api surfaces.

This would be a pretty significant refactor, can we do this as a follow up PR?

juli4n: Agree. We can do it in a follow up. I wanted to move the validations to a separate package (under `cmd/ateapi/internal`) for other reasons.

juli4n: Maybe add a TODO comment in the code to remove the hand-written validations and use DV?

sfunkenhauser: Done

pkg/proto/ateapipb/ateapi.proto
sfunkenhauser: Updated the comment to reflect the state of the code. We don't want this method to be idempotent as:
1) An already suspended actor no longer belongs to this worker, and this worker shouldn't be able to know it exists or make calls against it.
2) When moving to FGA for validation, depending on the actor state will make things more complicated.

cmd/ateapi/internal/workerservice/suspend.go
```diff
@@ -0,0 +1,165 @@
...
+	// package holds the only +k8s:validation-gen marker, and this package must
+	// not depend on the Control service. Once the marker moves to a neutral
+	// package both can import, call the generated validator here and drop
+	// validateActorRef with it.
+	//
+	// Workers are global-scoped, so the reference carries no atespace.
+	if errs := resources.ValidateGlobalObjectRef(req.GetWorker(), field.NewPath("worker")); len(errs) > 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid worker: %v", errs.ToAggregate())
+	}
+	if errs := validateActorRef(req.GetActor(), field.NewPath("actor")); len(errs) > 0 {
```
laoj2: what about also checking that the actor is not a golden actor? Otherwise, we may be allowing ateoms to suspend them, when this is a control plane decision.

sfunkenhauser: Good catch. Done.

laoj2 (approved):

